PR #14195, opened 2020-03-31
Update Makefile

wiggin77:
#### Summary

`make searchengine-mocks` was failing due to not finding `mockery`, although it was installed.

Updated this task to use GOBIN, like the others.


wiggin77:
Simple change to makefile - no QA needed.

PR #14196, opened 2020-03-31
MM-23747 fix Sprintf linter errors from new gosimple

wiggin77:
#### Summary

This PR fixes errors from `make check-style` due to new `gosimple`.

The upgrade to Go 1.14.x, triggered VSCode to update tools, which included an updated `gosimple`.  The linter is now complaining about 
```
S1038: should use fmt.Printf instead of fmt.Println(fmt.Sprintf(...)) (but don't forget the newline) (gosimple)
```

- replace all instances of `fmt.Sprintf("whatever")` with simply "whatever".
- replace all instances of `fmt.Println(fmt.Sprintf("whatever %d", i))` with `fmt.Printf("whatever %d\n", i)`

#### Ticket Link

https://mattermost.atlassian.net/browse/MM-23747

### DIFF
--- a/app/channel_test.go
+++ b/app/channel_test.go
@@ -811,8 +811,8 @@ func TestGetChannelMembersTimezones(t *testing.T) {
 func TestGetChannelsForUser(t *testing.T) {
 	th := Setup(t).InitBasic()
 	channel := &model.Channel{
-		DisplayName: fmt.Sprintf("Public"),
-		Name:        fmt.Sprintf("public"),
+		DisplayName: "Public",
+		Name:        "public",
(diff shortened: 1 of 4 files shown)


wiggin77:
@lindy65 , I don't think QA is necessary here as there are no functional changes.  This simply stops wrapping plain strings in Sprintf calls, and replaces with equivalent strings.  Same for embedded Sprintf calls.    Do you agree?

wiggin77:
> Maybe we also bump the version of `golangci-lint` to make sure that new failures don't happen?

Agreed.  @agnivade is doing that.

PR #14202, opened 2020-04-01
MM-23395: Add websockets support for update group

catalintomai:
#### Summary
Ensuring that when we delete or update a MM group, the websocket event is triggered and we are updating the referenceable groups (et al) in Redux for group-mentions

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-23395 Note: this addresses items(events) 1-9 of the ticket - see ticket comments.

#### Related Pull Requests
redux: https://github.com/mattermost/mattermost-redux/pull/1113
webapp: https://github.com/mattermost/mattermost-webapp/pull/5245

Note: removed the mobile changes: mattermost/mattermost-mobile#4116 from this PR since they require extra consideration and are not blocking for the current set for webapp.


app/group.go
```diff
@@ -32,11 +32,27 @@ func (a *App) CreateGroup(group *model.Group) (*model.Group, *model.AppError) {
...
+		messageWs.Add("group", newGroup.ToJson())
+		a.Publish(messageWs)
+	}
+
+	return newGroup, err
 }
 
 func (a *App) DeleteGroup(groupID string) (*model.Group, *model.AppError) {
-	return a.Srv().Store.Group().Delete(groupID)
+	newGroup, err := a.Srv().Store.Group().Delete(groupID)
```
mkraft, line 47:
```suggestion
	deletedGroup, err := a.Srv().Store.Group().Delete(groupID)
```

app/group.go
```diff
@@ -32,11 +32,27 @@ func (a *App) CreateGroup(group *model.Group) (*model.Group, *model.AppError) {
...
+
+	if err == nil {
+		messageWs := model.NewWebSocketEvent(model.WEBSOCKET_EVENT_RECEIVED_GROUP, "", "", "", nil)
+		messageWs.Add("group", newGroup.ToJson())
+		a.Publish(messageWs)
+	}
+
+	return newGroup, err
```
mkraft, line 55:
```suggestion
	return deletedGroup, err
```
...or `nil`.

app/group.go
```diff
@@ -32,11 +32,27 @@ func (a *App) CreateGroup(group *model.Group) (*model.Group, *model.AppError) {
...
 }
 
 func (a *App) DeleteGroup(groupID string) (*model.Group, *model.AppError) {
-	return a.Srv().Store.Group().Delete(groupID)
+	newGroup, err := a.Srv().Store.Group().Delete(groupID)
+
+	if err == nil {
+		messageWs := model.NewWebSocketEvent(model.WEBSOCKET_EVENT_RECEIVED_GROUP, "", "", "", nil)
```
mkraft, line 50: Should this be the same same websocket event as the used to update a group?

app/group.go
catalintomai: updated.

app/group.go
catalintomai: when we remove/delete a group we are actually update its delete_at timestamp, so once can look at it as the update of one of its properties. As we were discussing some time ago, the GetGroup* call(s) return the group if it exists in the table, "removed" or not, so the redux state is populated with both "removed" and not-removed groups. If the redux state would've only contain the non-removed ones, then I think that we would've need to add a specific DELETE event to actually remove the entry from the redux state when it is "removed" from the UserGroups table, with the contract that the redux state contains only non-removed groups.

srkgupta:
QA to validate if all the points mentioned in the ticket https://mattermost.atlassian.net/browse/MM-23395 is working fine or not.

srkgupta:
@catalintomai can you please resolve the merge conflicts so that this could be tested on top of the latest master.

catalintomai:
> @catalintomai can you please resolve the merge conflicts so that this could be tested on top of the latest master.

@srkgupta , they are merged now. Thanks.

srkgupta:
@catalintomai please ignore the issue #1 mentioned above. Actually I just tried this again, and editing a Group mention is working fine now and changes are reflected immediately via the Redux event. And also points #10 to #13 are also working fine. So the only thing which is not working is this:

A group is associated to a team
A group is disassociated to a team
A group is associated to a channel
A group is disassociated from a channel

These are also required to be handled, because, a team/channel can already be group synced. And then the sysadmin can decide to add one more group to that team/channel. In this case, the users in the team/channel don't see the list of group mentions correctly unless they manually refresh the browser/desktop.

catalintomai:
> @catalintomai please ignore the issue #1 mentioned above. Actually I just tried this again, and editing a Group mention is working fine now and changes are reflected immediately via the Redux event. And also points #10 to #13 are also working fine. So the only thing which is not working is this:
> 
> A group is associated to a team
> A group is disassociated to a team
> A group is associated to a channel
> A group is disassociated from a channel
> 
> These are also required to be handled, because, a team/channel can already be group synced. And then the sysadmin can decide to add one more group to that team/channel. In this case, the users in the team/channel don't see the list of group mentions correctly unless they manually refresh the browser/desktop.

@srkgupta , you are correct, I failed to consider the case in which the channel/team is already group-synced and we are further linking groups to it. The latest update should contain the fix for the issue. ( have to run "npm run dev" in the redux folder to build the node_modules).

api4/channel.go
```diff
@@ -322,7 +322,6 @@ func patchChannel(c *Context, w http.ResponseWriter, r *http.Request) {
 	if c.Err != nil {
 		return
 	}
-
```
hahmadia, line 325: nit: we can probably add this back in

PR #14214, opened 2020-04-01
MM-23489 Fix poor Unicode handling in display names

wiggin77:
#### Summary

This PR implements a blacklist of Unicode characters that should not appear in names.  Instances of these characters are filtered out based on https://www.w3.org/TR/unicode-xml/#Charlist

Fields filtered:

- user:  first name, last name, nickname, bot description, username
(Edit added username)

- team: name, display name, description, company name

- channel:  name, display name

#### Ticket Link

https://mattermost.atlassian.net/browse/MM-23489


model/utils_test.go
```diff
@@ -736,3 +737,32 @@ func checkNowhereNil(t *testing.T, name string, value interface{}) bool {
...
+
+	tests := []struct {
+		name string
+		arg  string
+		want string
+	}{
+		{name: "empty string", arg: "", want: ""},
+		{name: "ascii only", arg: "Hello There", want: "Hello There"},
+		{name: "allowed unicode", arg: "Ādam likes Iñtërnâtiônàližætiøn", want: "Ādam likes Iñtërnâtiônàližætiøn"},
+		{name: "blacklist char, don't reverse string", arg: "\u202E2resu", want: "2resu"},
```
isacikgoz, line 759: Should we add a allowed `unicode` escaped character to the test cases?

wiggin77: Ok, I added that.  It's really just testing the compiler's interpretation of different unicode syntax.  https://github.com/mattermost/mattermost-server/pull/14214/commits/9546f681cccc6783e35f30b204ddf5c887b24951

wiggin77:
> Is there any specific reason for instead of adding new test cases you added new test methods?

No solid reason other than the tests were adapted from the existing "create", "update", "patch" tests which did not follow the test case pattern.  I'm 0/5 on it.  If you are >0 then I'll change it.

wiggin77:
> @wiggin77, To me, it's more clean to have one actual method per one test method. And, it would be more consistent with the rest of the codebase. Therefore, I am >0 on it 

Combined here: https://github.com/mattermost/mattermost-server/pull/14214/commits/71ee6a01ebc40fbc1b6d732ae46e384357075d1e

lindy65:
Hi @wiggin77 :)

Just checking that I am testing this correctly... I followed the repro steps in the [Jira ticket](https://mattermost.atlassian.net/browse/MM-23489) but I am not sure whether for the expected of: 

> The RLO character and other similar characters should not be allowed in display names.

means that I should **not** be able to SAVE such a username OR whether the username is saved in Account Settings but then just displayed in a post as per screenshot?

image

Can you confirm please? Thanks!

wiggin77:
> means that I should **not** be able to SAVE such a username OR whether the username is saved in Account Settings but then just displayed in a post as per screenshot?

Hi @lindy65 , the ticket talked about display names (via real first name, real last name), not usernames, but now that you mention it maybe user names should get the same treatment. 

To answer your question, the expected result is that the name is accepted but displayed as if the offending character(s) were never present.  In fact the characters are stripped out when saving.

lindy65:
Thanks @wiggin77 :)

Apologies, when I spoke about username above I was actually referring to first/last name but I'm pleased my error made you think about sanitizing usernames as well!

If I put `<U+202E>2resu` as the first name of the user and save, it saves the first name as exactly that and, if I mention `@user-1` with my display set to first/last name, I see:

image

I apologise for the questions but I'm not sure I'm understanding correctly because you say:

> the expected result is that the name is accepted but displayed as if the offending character(s) were never present. In fact the characters are stripped out when saving.

This is not the case when I use these characters for first or last name. They are saved and offending character(s) are not stripped when saving...

I **do** get an error when trying to use `<U+202E>2resu` as a username (as per screenshot below)

image

wiggin77:
Hi @lindy65  What's happening there is no Unicode character is being entered and it is just using the literal `<U+202E>2resu` which has no Unicode.  The username error is generated by pre-existing code; you can't have brackets in a username.

How to actually enter unicode differs between OS's.  I will ask the reporter @jupenur how they entered the characters when generating the steps to reproduce.

lindy65:
Ah, thank you for explaining @wiggin77 - I was following the steps in the Jira ticket but obviously need more help with the unicode ;)

> The username error is generated by pre-existing code; you can't have brackets in a username.

Yep, I was just confirming that I am still seeing this error (as expected)

wiggin77:
With Linux, using Firefox (and probably Chrome) you can enter the Unicode by pressing Shift-Ctrl-U then typing 202e followed by a space.  I'm trying to figure out how to do that in Window in a browser.

wiggin77:
The only way I've figured out how to enter it in Chrome + Windows is to open Word (or another Unicode editor; notepad won't work) and hold down Alt while typing 8238 on the number pad.  Release Alt and type your name.  It will be reversed.  Ctrl-A to highlight the whole thing and cut & paste into Mattermost.

lindy65:
Yeah, I found this for entering unicode in Win10 but it doesn't work directly into Mattermost:
Press and hold down the Alt key.
Press the + (plus) key on the numeric keypad .
Type the hexidecimal unicode value.
Release the Alt key. 

I will try in MS Word and copy/paste into MM

wiggin77:
FWIW I added this to a test plan in the ticket.  I should have done that in the first place.

lindy65:
OK! we're getting somewhere :)

This is working as expected! Using your steps for MS Word, I copy/pasted the unicode (my name backwards) and pasted it into the firstname field in Mattermost and, upon saving, my name is displayed correctly.

When mentioning my user, my name is also displayed correctly.

Let me know if this is all that needs testing and I'll pass this through from QA 👍

wiggin77:
> Let me know if this is all that needs testing and I'll pass this through from QA

That's it.  I don't think you want to test every Unicode character in the blacklist - there is a unit test covering that.

PR #14230, opened 2020-04-03
[MM-23727] Make channel validation consistent on the server

fmunshi:
#### Summary
- ~Channel creation enforces that the channel URL must not contain `__` no matter which characters surround it so this PR changes the update and rename logic to match the create channel validations~

- This PR now removes the strict unnecessary validation (must not contain `__`) at the server level and ensures that it is consistent no matter where in the server a rename is done from.

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-23727

#### QA Test steps
- Attempt to rename a channel using the rename channel modal (or the CLI) and set the url to something like `channel__name`
- Ensure that it does not save

### DIFF
--- a/app/channel.go
+++ b/app/channel.go
@@ -144,10 +144,6 @@ func (a *App) CreateChannelWithUser(channel *model.Channel, userId string) (*mod
 		return nil, model.NewAppError("CreateChannelWithUser", "api.channel.create_channel.direct_channel.app_error", nil, "", http.StatusBadRequest)
 	}
 
-	if strings.Index(channel.Name, "__") > 0 {
-		return nil, model.NewAppError("CreateChannelWithUser", "api.channel.create_channel.invalid_character.app_error", nil, "", http.StatusBadRequest)
-	}
-
 	if len(channel.TeamId) == 0 {
 		return nil, model.NewAppError("CreateChannelWithUser", "app.channel.create_channel.no_team_id.app_error", nil, "", http.StatusBadRequest)
 	}
@@ -464,14 +460,6 @@ func (a *App) GetGroupChannel(userIds []string) (*model.Channel, *model.AppError
 
 // UpdateChannel updates a given channel by its Id. It also publishes the CHANNEL_UPDATED event.
 func (a *App) UpdateChannel(channel *model.Channel) (*model.Channel, *model.AppError) {
-	userIds := strings.Split(channel.Name, "__")
-	if channel.Type != model.CHANNEL_DIRECT &&
-		len(userIds) == 2 &&
-		model.IsValidId(userIds[0]) &&
-		model.IsValidId(userIds[1]) {
-		return nil, model.NewAppError("UpdateChannel", "api.channel.update_channel.invalid_character.app_error", nil, "", http.StatusBadRequest)
-	}
-
 	_, err := a.Srv().Store.Channel().Update(channel)
 	if err != nil {
 		return nil, err
--- a/i18n/en.json
+++ b/i18n/en.json
@@ -4500,7 +4500,7 @@
   },
   {
     "id": "model.channel.is_valid.name.app_error",
-    "translation": "Invalid name. User ids are not permitted in channel name for non-direct message channels."
+    "translation": "Invalid channel name. User ids are not permitted in channel name for non-direct message channels."
   },
   {
     "id": "model.channel.is_valid.purpose.app_error",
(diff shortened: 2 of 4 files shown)


fm2munsh:
@agnivade I agree that the validation is unnecessarily strict in this case - maybe this PR should keep the validation to only `userId__userId` and instead modify the `CreateChannel` method to also validate the same way.

The main issue that I noticed was that the server was checking for `__` on create and then checking `userID__user_ID` on rename so the validation was inconsistent as seen here: https://github.com/mattermost/mattermost-server/blob/master/app/channel.go#L147

And I will talk about the webapp validation with a PM and see what their thoughts on it are.

agnivade:
Yep, validation is inconsistent. See my comment here: https://github.com/mattermost/mattermost-server/pull/13441#issuecomment-568152388. The translations would also need to be changed. Or perhaps add a new error code with the same message in English. I should have thought of that in that PR itself :P

Yes, it would be great if we could make both webapp and server have the same validation code. And also make it consistent for createChannel and renameChannel.

Thanks !

jespino:
Yes, my feeling here is creating or updating channels should follow the same validation, what validation I think not allowing `__` as part of the name is a bit restrictive and is not necessary (I think). Can we make the creation validation less restrictive (checking the ids too)?

i18n/en.json
```diff
@@ -255,10 +255,6 @@
     "id": "api.channel.create_channel.direct_channel.app_error",
     "translation": "Must use createDirectChannel API service for direct message channel creation."
   },
-  {
-    "id": "api.channel.create_channel.invalid_character.app_error",
```
agnivade, line 259: Please also remove the translations in other files.

agnivade:
PM review for the error message.

i18n/en.json
fmunshi: My assumption was that we only modify `en` translations and all other translations are updated in the community translation PRs as per @jespino 's comment here https://community.mattermost.com/core/pl/jt8sdya5pjfh7bgbr5389gs9ac

agnivade: Hmm .. I am not really aware of this "other process". My main concern was that we are deleting a main translation. So there's not much use of keeping the other translation strings. They are just lying there. If there's another "process" which takes care of deleting those, then I'm fine.

jespino: Yes, exactly, the translations are added and removed only to the `en.json` file, the rest of the files are updated through our i18n process.

app/channel.go
```diff
@@ -464,14 +460,6 @@ func (a *App) GetGroupChannel(userIds []string) (*model.Channel, *model.AppError
 
 // UpdateChannel updates a given channel by its Id. It also publishes the CHANNEL_UPDATED event.
 func (a *App) UpdateChannel(channel *model.Channel) (*model.Channel, *model.AppError) {
-	userIds := strings.Split(channel.Name, "__")
```
jespino, line 467: I would expect to see the `IsValid` here. This function is called from multiple places and is hard to track what are not validating if we remove it from here.

app/channel.go
```diff
@@ -144,10 +144,6 @@ func (a *App) CreateChannelWithUser(channel *model.Channel, userId string) (*mod
 		return nil, model.NewAppError("CreateChannelWithUser", "api.channel.create_channel.direct_channel.app_error", nil, "", http.StatusBadRequest)
 	}
 
-	if strings.Index(channel.Name, "__") > 0 {
```
jespino, line 147: Same here, I would expect to have an `IsValid` here, or may one in the `CreateChannel` method, this way we have the creation and update properly verified.

app/channel.go
fmunshi: Does it make sense to add it here when the store validates it anyway and the error is propagated up?

jespino: Yes, that is a good point, I didn't realize that. Yes, if there is already validated at the store works for me. :)

i18n/en.json
```diff
@@ -4502,6 +4494,10 @@
     "id": "model.channel.is_valid.id.app_error",
     "translation": "Invalid Id."
   },
+  {
+    "id": "model.channel.is_valid.name.app_error",
+    "translation": "Invalid name. User ids are not permitted in channel name for non-direct message channels."
```
wiersgallak, line 4499: Recommend using what we had prior (includes the word "channel" before "name"):  "Invalid channel name. User ids are not permitted in channel name for non-direct message channels."

fmunshi: @wiersgallak I have updated the wording here now

PR #14231, opened 2020-04-03
MM-23817: Use auto-vendoring

agnivade:
#### Summary

With Go 1.14 (https://tip.golang.org/doc/go1.14#go-command),
it now automatically checks for the existence of a vendor directory.

So there is no need to explicitly set the `-mod=vendor` flag.

#### Ticket Link

https://mattermost.atlassian.net/browse/MM-23817

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -52,7 +52,7 @@ else
 endif
 
 # Go Flags
-GOFLAGS ?= $(GOFLAGS:) -mod=vendor
+GOFLAGS ?= $(GOFLAGS:)
 # We need to export GOBIN to allow it to be set
 # for processes spawned from the Makefile
 export GOBIN ?= $(PWD)/bin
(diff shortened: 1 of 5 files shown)


agnivade:
Finally I get to do what I wrote here: https://community-daily.mattermost.com/core/channels/go/jfirtmd4stdi3cp3dkbrq1qpwe :)

reflog:
@agnivade - if I have 1.13 installed locally, how will it behave?

agnivade:
It will throw a warning that minimum version required is 1.14, but it will still go on. And since there's no `-mod=vendor` flag, it will take it from the mod cache.

agnivade:
@josephbaylon - Some light QA around Makefile commands should be sufficient. Just ensure that dependencies are not being downloaded. I have added some commands in the JIRA ticket.

Thanks.

PR #14234, opened 2020-04-03
MM-22056: Fix flaky test TestUploadFiles

agnivade:
#### Summary

This was an interesting synchronization problem that happens due
to an incorrect usage of io.Pipe. There is no apparent race condition
here prima facie. But as always with all flaky tests, there has to be
a race somewhere. Here it's in the filesystem. Essentially we are not waiting
until the image is fully written.

What happens is that in (*UploadFileTask).postprocessImage, we spawn 2 goroutines
to write the thumbnail and preview images concurrently. And in those goroutines,
there is a pipe. We encode the image in a PipeWriter, and at the same time,
read from the PipeReader to write the image. All is well until here. Because
a pipe is synchronous and there is no buffering -

> That is, each Write to the
    PipeWriter blocks until it has satisfied one or more Reads from the
    PipeReader that fully consume the written data. The data is copied directly
    from the Write to the corresponding Read (or Reads); there is no internal
    buffering.

But the problem happens inside the (*App).WriteFile method. The localstore
implementation uses an io.Copy to read from the PipeReader and write to a file handle.
And lo-and-behold, io.Copy uses a buffer ! And that too a 32k one. It's not apparent
from the docs of io.Copy, but io.Copy shares the same code with io.CopyBuffer
and the docs are clear there.

> CopyBuffer is identical to Copy except that it stages through the provided
    buffer (if one is required) rather than allocating a temporary one.

So now we get to the root of the problem. This means that, the moment when the
last chunk has been read from the PipeReader, jpeg.Encode is free to move forward.
Which means there is nothing stopping postProcess from finishing before the image
is actually written.

Therefore, in the rarest of rare situations, it would so happen that the last
chunk would still be unwritten, and the runtime, in a cruel twist of fate,
would complete executing all of the test code after uploading the file and would
finally get to reading the file. And now we have finally arrived at our race condition.
The file would not be completely written, before it is being read.

To fix this, we simply re-arrange the reader and writer ends of the pipe.
We write from inside the goroutine, so the reader has to read everything until
the write has called Close or CloseWithError. Yet another way to fix would be
to re-arrange the sync.Waitgroup to actually wait till the
writeFile is finished before returning. But the earlier approach seemed more
idiomatic as it's what is prescribed in the standard library example.

This also means that we can get rid of the
additional `time.Sleep(2 * time.Second)` in other tests also.

Bonus points:
- We replace math/rand with crypto/rand because more randomess !
- We explicitly add a blank import for the tiff decoder. This was
being implicitly added because we were importing the disintegration/imaging package.
But it's good to be explicit.
- We replace pointer to a waitgroup to a more idiomatic declaration of it.

#### Ticket Link

https://mattermost.atlassian.net/browse/MM-22056


agnivade:
@saturninoabril - pinging you because you added the tests which have the `time.Sleep(2 * time.Second)`. I removed them for the getthumbnail and getpreview handlers, because I know these would be fixed by this PR. But do you know if the sleeps in others (getFileInfo and getFileLink) were also due to this same issue ? If yes, then we can also get rid of them and shave off 6 more seconds from our tests.

saturninoabril:
@agnivade Thanks for checking. Yes, I believed those are of the same issue.

agnivade:
Awesome, thanks !

app/file.go
```diff
@@ -799,26 +800,25 @@ func (t *UploadFileTask) postprocessImage() {
 	writeJPEG := func(img image.Image, path string) {
 		r, w := io.Pipe()
 		go func() {
-			_, aerr := t.writeFile(r, path)
-			if aerr != nil {
-				mlog.Error("Unable to upload", mlog.String("path", path), mlog.Err(aerr))
-				return
+			err := jpeg.Encode(w, img, &jpeg.Options{Quality: 90})
```
gsagula, line 803: Can you make Quality value into some default constant and add a comment to it?

agnivade: done. And I also reverted the extra checks that were a TODO statement, which I forgot to remove.

agnivade:
Hi @lindy65 - This fixes a unit test. So I don't think there is anything to be tested here. Let me know your thoughts.

Thanks.

PR #14236, opened 2020-04-04
MM-23666: rewrite TestHTTPClient

lieut-data:
#### Summary
The existing `TestHTTPClient` actually tries to reach google.com, making it brittle to network hiccups (as unlikely as that might be). Rewrite the test to rely on a local server instead.

I found the existing implementation of `TestHTTPClient` extremely hard to understand, so I've rewritten it more explicitly, expanded coverage somewhat. In doing so, I found myself somewhat surprised by the current behaviour. If `allowHost` returns `true`, we ignore `allowIP` altogether. If `allowHost` returns `false`, we check with `allowIP`. There's a lack of symmetry here that the current filter names don't make clear. Perhaps `allowHost` should be renamed `skipIPCheckForHost`?

#### Ticket Link
Fixes: https://mattermost.atlassian.net/browse/MM-23666

### DIFF
--- a/services/httpservice/client_test.go
+++ b/services/httpservice/client_test.go
@@ -48,8 +48,7 @@ func TestHTTPClient(t *testing.T) {
 		for _, testCase := range testCases {
 			t.Run(testCase.description, func(t *testing.T) {
 				c := NewHTTPClient(NewTransport(testCase.enableInsecureConnections, nil, nil))
-				_, err := c.Get(testCase.url)
-				if testCase.expectedAllowed {
+				if _, err := c.Get(testCase.url); testCase.expectedAllowed {
 					require.NoError(t, err)
 				} else {
 					require.Error(t, err)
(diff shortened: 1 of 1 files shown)


jupenur:
If I'm understanding this correctly, before this change the test covered a case with both an external host and a valid SSL certificate -- the `google.com` one. After this change it no longer does. Is that an acceptable compromise?

Otherwise looks great.

lieut-data:
> We give priority to the hostname check. If that does not pass (or it's not passed in) we rely on IP addresses. That's how I read it.

@streamer45, that's a reasonable interpretation. I think I was just surprised that I could not /reject/ a host using `allowHost`, but can using `allowIP`. Superficially, the nomenclature suggests they should be parallel.

@jupenur, you're absolutely right that we're no longer testing `TestHTTPClient` against a server with a valid SSL certificate. I /think/ the compromise is ok, if not better, since we're now testing that a self-signed certificate is actually rejected or accepted based on the requested configuration. My gut says that if we were to regress and not notice, it would be this latter case, and not our ability to connected to a trusted endpoint.

Thoughts?

jupenur:
Makes sense. FWIW, regarding `allowHost`, my interpretation was the same as @streamer45's, and I think it especially makes sense if you look at what it's used for currently: whitelisting either hosts or IPs based on a mixed list of both. Of course you shouldn't have to look at current use-cases to understand an API, but don't I think there's room for dangerously incorrect interpretations or anything like that.

lieut-data:
@prapti, heads up that there are no changes to the product here, just expanded / modified test coverage.

agnivade:
Would be nice to finally fix this. :+1:  This has been failing my PRs like 3 times out of 5. :(

PR #14239, opened 2020-04-06
MM-23876: Fix for patching channel moderations with a null team schem…

mkraft:
…e channel guest role.
#### Summary

Accounts for a blank `DefaultChannelGuestRole` on the channel moderation patch endpoint.

#### Ticket Link

https://mattermost.atlassian.net/browse/MM-23876

### DIFF
--- a/api4/channel_test.go
+++ b/api4/channel_test.go
@@ -3282,4 +3282,45 @@ func TestPatchChannelModerations(t *testing.T) {
 		require.NotEqual(t, scheme.DeleteAt, int64(0))
 	})
 
+	t.Run("Does not return an error if the team scheme has a blank DefaultChannelGuestRole field", func(t *testing.T) {
+		team := th.BasicTeam
+		scheme := th.SetupTeamScheme()
+		scheme.DefaultChannelGuestRole = ""
+
+		mockStore := mocks.Store{}
+		mockSchemeStore := mocks.SchemeStore{}
+		mockSchemeStore.On("Get", mock.Anything).Return(scheme, nil)
+		mockSchemeStore.On("Save", mock.Anything).Return(scheme, nil)
+		mockSchemeStore.On("Delete", mock.Anything).Return(scheme, nil)
+		mockStore.On("Scheme").Return(&mockSchemeStore)
+		mockStore.On("Team").Return(th.App.Srv().Store.Team())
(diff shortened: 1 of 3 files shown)


app/channel.go
```diff
@@ -726,7 +726,7 @@ func (a *App) GetChannelModerationsForChannel(channel *model.Channel) ([]*model.
 	}
 
 	var higherScopedGuestRole *model.Role
-	if len(guestRoleName) > 0 {
+	if len(higherScopedGuestRoleName) > 0 {
```
fm2munsh, line 729: oh good catch - that would have been bad

app/channel.go
```diff
@@ -772,19 +783,29 @@ func (a *App) PatchChannelModerationsForChannel(channel *model.Channel, channelM
...
+			return nil, err
+		}
+	}
+
 	memberRolePatch := memberRole.RolePatchFromChannelModerationsPatch(channelModerationsPatch, "members")
-	guestRolePatch := guestRole.RolePatchFromChannelModerationsPatch(channelModerationsPatch, "guests")
+	var guestRolePatch *model.RolePatch
+	if guestRole != nil {
+		guestRolePatch = guestRole.RolePatchFromChannelModerationsPatch(channelModerationsPatch, "guests")
+	}
```
fm2munsh, line 808: does `Role.PatchRole` handle a `nil` RolePatch gracefully?

fm2munsh: I guess it would just set the guest role to be a role with no permissions which makes sense in this case

api4/channel_test.go
```diff
@@ -3282,4 +3282,45 @@ func TestPatchChannelModerations(t *testing.T) {
...
+
+		patch := []*model.ChannelModerationPatch{
+			{
+				Name:  &createPosts,
+				Roles: &model.ChannelModeratedRolesPatch{Members: model.NewBool(true)},
+			},
+		}
+
+		_, res = th.SystemAdminClient.PatchChannelModerations(channel.Id, patch)
+		require.Nil(t, res.Error)
```
fm2munsh, line 3323: is it worth it to check the content of the response too as opposed to just checking that it did not error?

app/channel.go
mkraft: Yup, I added that bug on my last PR.

PR #14240, opened 2020-04-06
MM-23131 Include HTTP status code in the metrics

ethervoid:
#### Summary

Right now we're only including handler and method. But to get better graphics like RED ones, for example, we need the status code returned so this PR includes:

- Wrapper for http.ResponseWriter that stores the status code set by the `WriteHeader` method
- A new version of the `ObserveApiEndpointDuration` method that includes the status code

#### Ticket Link

https://mattermost.atlassian.net/browse/MM-23131


ethervoid:
@reflog yep, I've have been searching and is the only way I've found. Regarding performance Impact I don't think it affects but I can do some benchmarking

agnivade:
Yeah .. status codes are only available after the response has been written. So the only way is through a middleware.

Perf impact should be minimal, but I am yet to do a review yet.

web/handlers.go
```diff
@@ -79,6 +80,7 @@ type Handler struct {
 }
 
 func (h Handler) ServeHTTP(w http.ResponseWriter, r *http.Request) {
+	newWriter := BuildResponseWriter(w)
```
agnivade, line 83: Is there a need to create a new variable ? Since we are wrapping it, it maybe better to just overwrite w with this and then reuse w throughout. And we can avoid this code churn.

Also, there is no need to expose `BuildResponseWriter`. Idiomatically, it should be `newWrappedWriter(w)`.

web/response_writer_wrapper.go
```diff
@@ -0,0 +1,62 @@
...
+package web
+
+import (
+	"bufio"
+	"errors"
+	"net"
+	"net/http"
+)
+
+type ResponseWriterWrapper struct {
```
agnivade, line 13: No need to expose the struct.

web/response_writer_wrapper.go
```diff
@@ -0,0 +1,62 @@
...
+	http.ResponseWriter
+	statusCode        int
+	statusCodeWritten bool
+	hijacker          http.Hijacker
```
agnivade, line 17: Is there a need to store the hijacker and the flusher ? From what I see here: https://github.com/urfave/negroni/blob/f4316798d5d3acd39eb6784301b19f27f471415f/response_writer.go#L31, we can just cast the response to the appropriate interface and get the job done.

web/response_writer_wrapper_test.go
```diff
@@ -0,0 +1,88 @@
...
+}
+
+func createTestHandler(wrapper *ResponseWriterWrapper, testFunc func(w http.ResponseWriter, r *http.Request)) http.Handler {
+	return &TestHandler{
+		Wrapper:  wrapper,
+		TestFunc: testFunc,
+	}
+}
+
+type ResposeRecorderHijack struct {
```
agnivade, line 32: No need to expose these.

web/handlers.go
ethervoid: I don't like to modify the current object to avoid side-effects or if we need the original one. I'd rather prefer to create a new variable.

web/response_writer_wrapper.go
ethervoid: Yes, given that we're storing the interface for `http.ResponseWriter` either the `Flush` and `Hijack` methods are not available from the wrapper so you have to get them from the original writer as you can see in [this play example](https://play.golang.com/p/C-WDdmHvn33). In case you want to have it right you have to do something like [this](https://play.golang.com/p/GA7ISH3Gf8q)

I'm pretty sure the Negroni implementation is not working right in the `Hijack` case

agnivade: Ah yes, the negroni implementation is doing exactly that. What do you think about embedding the Flusher interface too ? It makes it cleaner IMO because we are just composing interfaces and not storing any concrete state.

ethervoid: In that case, we have to embed the Hijack interface too. I don't have a strong opinion on this so I can do it without problem

agnivade: Yes, I think that's better.

web/handlers.go
ethervoid: In offline conversation with @agnivade we have agreed to shadow the ResponseWriter variable coming in the `ServerHTTP` method and cast it to the wrapper when needed, for example calling the `StatusCode` method

ethervoid:
@agnivade All the changes uploaded

web/response_writer_wrapper.go
ethervoid: Finally given that we don't want a new Interface we let this part as it is (talked offline)

ethervoid:
@ogi-m gentle ping :)

saturninoabril:
@ethervoid Do you have corresponding PR to [enterprise](https://github.com/mattermost/enterprise/blob/master/metrics/metrics.go#L720)?

PR #14245, opened 2020-04-07
[MM-23535] Add disable database search flag and return empty posts results if set

mgdelacroix:
#### Summary
Adds a configuration flag to disable database search in case. This is useful for environments that would prefer not to obtain search results rather than querying a big database.

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-23535

#### Related Pull Requests
- [Has enterprise changes](https://github.com/mattermost/enterprise/pull/608)
- [Has webapp changes](https://github.com/mattermost/mattermost-webapp/pull/5326)


store/searchlayer/layer.go
```diff
@@ -17,12 +17,14 @@ type SearchStore struct {
...
 	post         *SearchPostStore
+	config       *model.Config
 }
 
-func NewSearchLayer(baseStore store.Store, searchEngine *searchengine.Broker) SearchStore {
+func NewSearchLayer(baseStore store.Store, searchEngine *searchengine.Broker, cfg *model.Config) SearchStore {
 	searchStore := SearchStore{
 		Store:        baseStore,
 		searchEngine: searchEngine,
+		config:       cfg,
```
jespino, line 27: This doesn't handle config changes, I think, maybe another approach is to receive a config service here, and request the config each time. Other option is to add a "SetConfig" method and connect a ConfigListener to update this config every single time.

mgdelacroix: Thanks @jespino! Added a config listener to update the configuration for the `SearchStore` in https://github.com/mattermost/mattermost-server/pull/14245/commits/9e3184fcee82772e556a175bf08f87b5e217f2a8. I'm happy to explore other approaches like the config service one if you think they might fit better

PR #14247, opened 2020-04-07
Revert "Updating to use a different network cidr for the mm-test network"

metanerd:
Reverts mattermost/mattermost-server#14223

This PR resulted into jenkins build issues.


agnivade:
This is an emergency quick fix that was needed to fix the build pipeline. Hence no QA is required.

PR #14252, opened 2020-04-07
MM-23734

reflog:
#### Summary

This PR introduces SidebarCategories and SidebarChannels tables as described in the spec https://docs.google.com/document/d/1thbPWPd9Wnz2GiMJq0Fm9mFzXmL1t66UHX6WNVNAAhc/edit

To test the performance of the migration, the following sample data was generated (based on numbers received from Cloud team to use our community server as baseline)
```go
	$(GO) run $(GOFLAGS) -ldflags '$(LDFLAGS)' $(PLATFORM_FILES) config set TeamSettings.MaxUsersPerTeam 76000
	$(GO) run $(GOFLAGS) -ldflags '$(LDFLAGS)' $(PLATFORM_FILES) config set TeamSettings.MaxChannelsPerTeam 4000
	$(GO) run $(GOFLAGS) -ldflags '$(LDFLAGS)' $(PLATFORM_FILES) sampledata -t 5 --channels-per-team 3274 -w 4 -u 6000 --direct-channels 19162 --group-channels 4511
```
On my laptop it took 19 seconds to perform migration.

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-23733


store/sqlstore/upgrade.go
```diff
@@ -784,3 +786,14 @@ func upgradeDatabaseToVersion522(sqlStore SqlStore) {
 		saveSchemaVersion(sqlStore, VERSION_5_22_0)
 	}
 }
+
+func upgradeDatabaseToVersion523(sqlStore SqlStore) {
+	//if shouldPerformUpgrade(sqlStore, VERSION_5_22_0, VERSION_5_23_0) {
+	if err := sqlStore.Channel().MigrateChannelsToSidebar(); err != nil {
```
hmhealey, line 792: This can happen in 5.24. Also, this isn't doing the lazy migration as discussed before. Do we think that the impact of this won't be significant?

store/sqlstore/channel_store.go
```diff
@@ -416,6 +445,188 @@ func (s SqlChannelStore) createIndexesIfNotExists() {
...
+	if _, err := s.GetMaster().Exec(sql, args...); err != nil {
+		return err
+	}
+
+	return nil
+}
+
+// migrateDirectMessagesToSidebarChannels populates direct messages category with DMs and GMs
+func (s SqlChannelStore) migrateDirectMessagesToSidebarChannels() error {
+	sb := s.
```
hmhealey, line 540: If we're populating the DM categories, this needs to look at the Preferences table since that's used to hide DMs when the user removes them from their sidebar.

Alternatively, since the DMs channel needs to be sorted dynamically, we could populate it dynamically. The Direct Messages category only offers automatic sorting unlike the other categories that are all manually sorted.

store/sqlstore/channel_store.go
```diff
@@ -334,6 +334,30 @@ type publicChannel struct {
 	Purpose     string `json:"purpose"`
 }
 
+type sidebarCategoryType string
+
+const (
+	SidebarCategoryChannels       sidebarCategoryType = "C"
+	SidebarCategoryDirectMessages sidebarCategoryType = "D"
+	SidebarCategoryFavorites      sidebarCategoryType = "F"
```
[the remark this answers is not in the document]
reflog: I wanted to keep them as short as possible to avoid unneeded long strings in JSON responses... We do the same with channel types (O/P/F/G/D).
@hmhealey  Thoughts?

store/sqlstore/channel_store.go
```diff
@@ -416,6 +445,188 @@ func (s SqlChannelStore) createIndexesIfNotExists() {
...
+	}
+	if err := transaction.Commit(); err != nil {
+		return model.NewAppError("SqlChannelStore.Save", "store.sql_channel.save.commit_transaction.app_error", nil, err.Error(), http.StatusInternalServerError)
+	}
+	return nil
+}
+
+// migrateFavoriteChannels copies favorites from Preferences table to SidebarChannels
+func (s SqlChannelStore) migrateFavoriteChannelToSidebarChannels() error {
+	sb := s.
```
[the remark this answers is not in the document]
reflog: @hmhealey  Unless I am mistaken, this should take care of GMs - they will get picked up from the Preferences table, same as the rest. Or am I missing something?

store/sqlstore/channel_store.go
```diff
@@ -416,6 +445,188 @@ func (s SqlChannelStore) createIndexesIfNotExists() {
...
+		getQueryBuilder().
+		Select("Preferences.UserId", "Preferences.Name AS ChannelId", "SidebarCategories.Id AS CategoryId", "0 AS `Order`").
+		From("Preferences").
+		Where(sq.And{
+			sq.Eq{"Preferences.Category": "favorite_channel"},
```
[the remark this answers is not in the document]
reflog: The whole SQL is one big string... Would adding a constant here really help? 🤔

grundleborg:
Some comments/questions on the migration, as that's the high risk part of this PR. 

To clearly state the rationale for why I consider this to be high risk: large customers use Mattermost as part of their critical infrastructure, and therefore have strict SLAs about outages/down time. We have established a convention that in general only "trivial" schema changes (i.e. to quick for the human eye to notice - less than a couple of seconds) happen in a blocking way on server startup. We have broken this convention occasionally, but only *intentionally* when the migration can actually be run out of band - i.e. the SQL can be extracted and run on it's own before doing the Mattermost upgrade without breaking functionality for the older version of Mattermost.

So, to conclude the premise: we need to be **very confident** that we won't cause unacceptable down time for customers with this migration.

#### What is "very confident"

We believe that across the size-spectrum of customer installs that we are aware of there won't be issues.

#### What is "unacceptable down time"

We don't have a clear definition for this. Perhaps that's something we need to fix more generally so we can set explicit rules for making these decisions? Anyway, based on what I know at the moment, anything much over about a minute is "unacceptable". Whether the threshold needs to be down as low as "a couple of seconds" I'm not sure. We can loop in the PMs to discuss this in detail if needed.

#### How to we evidence this?

In order to approve this PR, I'd want to see a clear methodology and working as to how we've verified the impact of this migration. I think all the information is probably here already, but it needs to be presented all together so it is totally clear to the reviewer what the rationale is. Here's where I'd start:

1. A list of what database data is being migrated.
2. A numerical basis for what we expect the amount of data for each item in (1) to be during the migration. In this case, it looks like we're using evidence from community.mattermost.com, but we should also calculate that scaled up appropriately to a typical "large" MM customer install.
3. Note on the test data generated, and if there are any aspects of the migration for which test data was not generated.
4. Numbers as per the test. Details and justification for any extrapolation needed to get the final "expected" numbers for a future migration.
5. If the migration is slow, is it backwards compatible? If the DB admin extracts the SQL and runs it in advance of the MM upgrade, will anything break in the old version of Mattermost? Or will continuing to run the old version of MM cause unmigrated data to end up in the database?

#### Reversibility and Multi-Version Clusters

This is another topic, aside from the performance of the migration, which we need to consider thoroughly.

1. Is this migration backwards compatible?
2. If not, can we provide a comment with SQL to manually reverse the migration if the sysadmin has to revert an MM upgrade?
3. What happens on Community with the different versions of MM sharing a database?

reflog:
@grundleborg - thanks for the detailed feedback.
Let me address some of the questions posed here:
1. A list of what database data is being migrated - we are using Preferences table and ChannelMembers to populate the new SidebarChannels table and TeamMembers to populate the new SidebarCategories
2. Numerical basis for test - the `sampledata` command has been used using community as the baseline. Given more information about our customer installations - additional benchmarking can be done
3. Only favorite GMs are not supported by the `sampledata` command
4. Not sure about this question,can you elaborate?
5. The migration is one way only and is fully backwards compatible since it's new tables. In addition to the startup migration, "lazy migration" will be added to the REST calls that touch Preferences/ChannelMembers to keep SidebarChannels table properly populated

Reversability:
1. Yes, it's backwards compatible, see note 5 above
2. `DROP SidebarChannels; DROP SidebarCategories;` will be sufficient
3. Don't see the issue (see the note about lazy migrations)

store/sqlstore/channel_store.go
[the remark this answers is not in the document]
hmhealey: I don't actually know the origin of the shorter names for channel types, so I don't know if that was the original reason for using them. My main concern with using the single-letter names is that it means someone looking at the API response directly may have to look up the meaning of the individual letters later on, especially if we run into a case like the channel types where Public and Private both start with the same letter. I also don't think that the shorter names would noticeably impact the size of API responses, especially if we gzip them (although I can't remember if that actually got implemented or not).

That said, I'm personally biased towards more verbose names, so I'm 0/5 either way. We just need to remember to update the client-side versions of these constants.

store/sqlstore/channel_store.go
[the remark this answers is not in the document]
hmhealey: Currently, if you favourite a DM/GM, it shows up in your Favourites on each team. Is it creating a SidebarChannel entry for each of those teams?

Looking below, it looks like it'll actually create a Favourites category based on the Channel's `TeamId`, but that will be empty for DMs/GMs

store/sqlstore/channel_store.go
```diff
@@ -416,6 +421,260 @@ func (s SqlChannelStore) createIndexesIfNotExists() {
...
+			UserId:      u.UserId,
+			TeamId:      u.TeamId,
+			SortOrder:   10,
+			Type:        model.SidebarCategoryDirectMessages,
+		}); err != nil && !IsUniqueConstraintError(err, []string{"UserId"}) {
+			return nil, model.NewAppError("SqlChannelStore.MigrateSidebarCategories", "store.sql_channel.MigrateSidebarCategories.open_transaction.app_error", nil, err.Error(), http.StatusInternalServerError)
+		}
+
+		if err := transaction.Insert(&model.SidebarCategory{
+			DisplayName: "Channels",
```
grundleborg, line 469: Should these be localised?

store/sqlstore/channel_store.go
```diff
@@ -416,6 +421,260 @@ func (s SqlChannelStore) createIndexesIfNotExists() {
 	s.CreateIndexIfNotExists("idx_channels_scheme_id", "Channels", "SchemeId")
 }
 
+func (s SqlChannelStore) MigrateSidebarCategories(fromTeamId, fromUserId string) (map[string]string, *model.AppError) {
```
grundleborg, line 424: This method is quite complicated, I know we don't normally do this, but in cases like this I think it would be worthwhile adding a comment explaining what it does at a high level. Based on my understanding, it goes over every `TeamMember` and creates 3 default categories for it: for Favourites, Direct Messages and Channels. Is that correct?

store/sqlstore/channel_store.go
```diff
@@ -416,6 +421,260 @@ func (s SqlChannelStore) createIndexesIfNotExists() {
...
+		Select("Preferences.UserId", "Preferences.Name AS ChannelId", "SidebarCategories.Id AS CategoryId").
+		From("Preferences").
+		Where(sq.And{
+			sq.Eq{"Preferences.Category": model.PREFERENCE_CATEGORY_FAVORITE_CHANNEL},
+			sq.NotEq{"SidebarCategories.Id": nil},
+			sq.Gt{"Preferences.UserId": lastUserId},
+		}).
+		LeftJoin("Channels ON (Channels.Id=Preferences.Name)").
+		LeftJoin("SidebarCategories ON (SidebarCategories.UserId=Preferences.UserId AND SidebarCategories.TeamId=Channels.TeamId AND SidebarCategories.Type='F')").
+		OrderBy("Preferences.UserId")
```
grundleborg, line 508: Looks like each time this function is called we do a batch of 1 single user's favourites? I wonder if we should be a bit more aggressive and do say 100 in a transaction. We need to find a balance between this migration taking for ever and it blocking other queries by each transaction taking too long. 0/5 on this particular case though.

store/sqlstore/channel_store.go
```diff
@@ -416,6 +421,260 @@ func (s SqlChannelStore) createIndexesIfNotExists() {
...
+		getQueryBuilder().
+		Select("Channels.Id AS ChannelId", "ChannelMembers.UserId", "SidebarCategories.Id AS CategoryId").
+		From("Channels").
+		Where(sq.And{
+			sq.Or{
+				sq.Eq{"Channels.Type": model.CHANNEL_OPEN},
+				sq.Eq{"Channels.Type": model.CHANNEL_PRIVATE},
+			},
+			sq.NotEq{"SidebarCategories.Id": nil},
+			sq.Gt{"Channels.Id": lastChannelId, "ChannelMembers.UserId": lastUserId},
```
grundleborg, line 560: One invocation of this function migrates one DM channel? Seems like this might take a while? Again, maybe the batch size should be a bit larger?

store/sqlstore/channel_store.go
```diff
@@ -416,6 +421,260 @@ func (s SqlChannelStore) createIndexesIfNotExists() {
...
+	}
+
+	data := make(map[string]string)
+	data["UserId"] = userMemberships[len(userMemberships)-1].UserId
+	data["ChannelId"] = userMemberships[len(userMemberships)-1].ChannelId
+
+	return data, nil
+}
+
+func (s SqlChannelStore) MigrateDirectGroupMessagesToSidebarChannels(lastChannelId, lastUserId string) (map[string]string, *model.AppError) {
```
grundleborg, line 603: Same question here about batch size too.

reflog:
replaced by https://github.com/mattermost/mattermost-server/pull/14358

PR #14253, opened 2020-04-07
MM-23926: avoid setting last viewed for bots

lieut-data:
#### Summary
The current code path for `CreatePostAsUser` tries to update the `LastViewedAt` for bots, which logs a warning message if the bot isn't actually in the channel.

This pull request changes the semantics of bot posting to not update the bot's `LastViewedAt` timestamp, avoiding this log altogether. It matches the semantics of `from_webhook`, but notably makes bots slightly less like users in that they no longer "read" channels when they post. This seems reasonable, but I'm both looking for validation of this semantic change in addition to the code review.

#### Ticket Link
Fixes: https://mattermost.atlassian.net/browse/MM-23926


testlib/assertions.go
```diff
@@ -31,3 +31,24 @@ func AssertLog(t *testing.T, logs *bytes.Buffer, level, message string) {
 
 	t.Fatalf("failed to find %s log message: %s", level, message)
 }
+
+// AssertNoLog asserts that a JSON-encoded buffer of logs does not contains one with the given level and message.
+func AssertNoLog(t *testing.T, logs *bytes.Buffer, level, message string) {
```
hanzei, line 36: Nice helper method :+1:

PR #14254, opened 2020-04-07
Fixing system messages about non-visible users

jespino:
#### Summary
System messages about not visible users (like mention to users outside the
channel) were happening. This only notify the user who mention users visible to
them.

#### Ticket Link
[MM-23902](https://mattermost.atlassian.net/browse/MM-23902)


app/notification.go
```diff
@@ -428,7 +446,7 @@ func (a *App) filterOutOfChannelMentions(sender *model.User, post *model.Post, c
 		outOfChannelUsers = allUsers.FilterWithoutID(nonMemberIDs)
 		outOfGroupsUsers = allUsers.FilterByID(nonMemberIDs)
 	} else {
-		outOfChannelUsers = users
+		outOfChannelUsers = allUsers
```
jespino, line 449: @mkraft I think this was a bug, not related to the thing that I was trying to fix, but partially related.

PR #14258, opened 2020-04-07
[MM-23882] Include database server version in telemetry

ethervoid:
Also added a new query in the store to retrieve the database version

#### Summary

This PR includes the database version in telemetry

#### Ticket Link

https://mattermost.atlassian.net/browse/MM-23882


store/sqlstore/supplier_test.go
```diff
@@ -142,6 +144,29 @@ func TestGetReplica(t *testing.T) {
 	}
 }
 
+func TestGetDbVersion(t *testing.T) {
+	driverName := model.DATABASE_DRIVER_SQLITE
```
jespino, line 148: Why is this only tested for SQLITE?

ethervoid: Basically, because all the tests on that file are using only SQLite, even the `GetReplica` one` so I assumed that we was just testing SQLite but I've checked that the other tested methods don't have a different query per database engine sooooo I'm going to add it, thx for pointing it out :)

ethervoid: @jespino done I've made some refactor too

PR #14259, opened 2020-04-07
[MM-20684] plugin autocomplete implementation

iomodo:
#### Summary
This PR implements slash command autocomplete suggestions for plugins

#### Ticket Link
  Fixes 
https://mattermost.atlassian.net/browse/MM-21491
https://mattermost.atlassian.net/browse/MM-21493

#### Related PRs
webapp - https://github.com/mattermost/mattermost-webapp/pull/5289
redux - https://github.com/mattermost/mattermost-redux/pull/1121
demo-plugin - https://github.com/mattermost/mattermost-plugin-demo/pull/99


api4/command.go
```diff
@@ -362,6 +363,52 @@ func listAutocompleteCommands(c *Context, w http.ResponseWriter, r *http.Request
...
+		c.Err = err
+		return
+	}
+
+	suggestions := []model.AutocompleteSuggestion{}
+
+	autocompleteData := []*model.AutocompleteData{}
+	for _, command := range commands {
+		if command.AutocompleteData == nil {
+			if strings.HasPrefix(command.Trigger, userInput) {
```
levb, line 399: should this logic be in the `app` code?

app/command_autocomplete.go
```diff
@@ -4,109 +4,142 @@
...
 	suggestions := []model.AutocompleteSuggestion{}
-	index := strings.Index(userInput, " ")
-	if index == -1 { // no space in user input
+	index := strings.Index(inputToBeParsed, " ")
+	if index == -1 { // no space in input
 		for _, command := range commands {
-			if strings.HasPrefix(command.Trigger, userInput) && (command.RoleID == roleID || roleID == model.SYSTEM_ADMIN_ROLE_ID || roleID == "") {
-				suggestions = append(suggestions, model.AutocompleteSuggestion{Hint: command.Trigger, Description: command.HelpText})
+			if strings.HasPrefix(command.Trigger, inputToBeParsed) && (command.RoleID == roleID || roleID == model.SYSTEM_ADMIN_ROLE_ID || roleID == "") {
+				suggestion := inputParsed + command.Trigger
```
levb, line 25: nit/style: 0/5 suggest:
```go
    s := model.AutocompleteSuggestion{
        Suggestion: inputParsed + command.Trigger,
        Hint: command.Hint,
        Description: command.HelpText,
    }
    suggestions = append(suggestions, s)
```

model/command_autocomplete.go
```diff
@@ -75,16 +79,19 @@ type AutocompleteDynamicListArg struct {
 
 // AutocompleteSuggestion describes a single suggestion item sent to the frontend
 type AutocompleteSuggestion struct {
-	// Hint describes what user might want to input
+	// Suggestion describes what user might want to input
```
levb, line 82: An example in the comment here would go a long way... not 100% intention-revealing Suggestion/Hint/Description - what is what

model/command_autocomplete.go
```diff
@@ -352,61 +414,61 @@ func stringNotInSlice(a string, slice []string) bool {
 
 // CreateJiraAutocompleteData will create autocomplete data for jira plugin. For testing purposes only.
 func CreateJiraAutocompleteData() *AutocompleteData {
-	jira := NewAutocompleteData("jira", "Available commands: connect, assign, disconnect, create, transition, view, subscribe, settings, install cloud/server, uninstall cloud/server, help")
+	jira := NewAutocompleteData("jira", "[command]", "Available commands: connect, assign, disconnect, create, transition, view, subscribe, settings, install cloud/server, uninstall cloud/server, help")
```
levb, line 417: Why is this in the public package?

app/plugin_commands.go
```diff
@@ -18,7 +20,27 @@ type PluginCommand struct {
 
 func (a *App) RegisterPluginCommand(pluginId string, command *model.Command) error {
 	if command.Trigger == "" {
-		return fmt.Errorf("invalid command")
+		return errors.New("invalid command")
+	}
+	if command.AutocompleteData != nil && command.AutocompleteData.IsValid() != nil {
+		return errors.New("invalid autocomplete data in command")
```
hanzei, line 26: Why are you dropping the `error` returned from `IsValid()`? Mind just wrapping it?

app/plugin_commands.go
```diff
@@ -18,7 +20,27 @@ type PluginCommand struct {
...
+			siteURL = fmt.Sprintf("http://localhost%v", *a.Srv().Config().ServiceSettings.ListenAddress)
+		}
+		baseURL, err := url.Parse(siteURL + "/plugins/" + pluginId)
+		if err != nil {
+			return errors.Wrapf(err, "Can't parse url %s", siteURL+"/plugins/"+pluginId)
+		}
+		err = command.AutocompleteData.UpdateRelativeURLsForPluginCommands(baseURL)
+		if err != nil {
+			return errors.Wrap(err, "Can't update relative urls for plugin commands")
+		}
```
hanzei, line 43: Why are you dealing with the relativeURL vs. absoluteURL problem on the server side? It seams easier to me to just have the webapp build the request URL itself, if it starts with `/plugins/`.

hanzei:
A few pieces of feedback:

1. It looks weird how the box around `/` shrinks if the hint is long. I was expecting the box to grow with the length of the text. 
Screenshot from 2020-04-16 14-49-24
cc @asaadmahmood 

2. I seams confusing how a hint is shown in the same line as the input. It's not clear to me that `Error in the webapp` is a description, especially because it uses the same `[]` syntax as option field like `user`
Screenshot from 2020-04-16 14-52-20

3. When I type `/jira install cloud example.com`, I still see `[URL]` at the end of my input, which makes me thing that I still have to input an `URL`
Screenshot from 2020-04-16 14-58-50

4. Should the validation for `/jira install cloud` work?

5. When I type `/jira transition`, select entry and press `ENTER`, it skip the step where I can pick a state an just selects the first one. This doesn't happen if I use `TAB` instead.
Screenshot from 2020-04-16 15-01-35

6. Nit pick: I can submit something like `/jira transition mm-24493 sadv` from the webapp. Maybe we can generate an error in the webbapp instead?

7. When I type `/jira subscribe` and press `ENTER` or `TAB`, the popup dismisses. I suppose this is a bug in the demo plugin.

8. When I type `/jira connect ` and press `ENTER` or `TAB`, the popup dismisses and a space is entered. Why does the space get added?

9. If I scrolle with the keyboard to the last entry it's cut off
Screenshot from 2020-04-16 15-08-15

asaadmahmood:
@hanzei That first one should be fixed by this PR:
https://github.com/mattermost/mattermost-webapp/pull/5328

hanzei:
Thanks for the hint @asaadmahmood. I will re-test once `master` gets merged into the feature branch again.

app/plugin_commands.go
iomodo: I don't like it either. This method is not called from the webapp, it's called from the server itself to fetch dynamic arguments from the plugin.

hanzei: Got it. Does it make sense to move the handling to https://github.com/mattermost/mattermost-server/pull/14259/files#diff-02aa38f226a808932a2b1e34a8cc834fR116 and do something like https://github.com/mattermost/mattermost-server/blob/master/app/integration_action.go#L270-L273?

iomodo:
@hanzei 
1. will be fixed after merge
2. Developer can configure hint and description strings however they want. I probably did a bad job there :)
3. You might not be finished with the `URL` there, it will go away once you type space
4. Can you elaborate?
5. hmm.. I'll look into it.
6. checking for patterns as well as list arguments is not part of this PR, it might need quite substantial change in the webapp, but will be implemented later. Probably there should be a setting for each argument to allow arguments that are not in the list or not(same as standard combobox).
7. true
8. `jira connect` expects server url as an argument so space is added automatically to continue to input the argument.
9. design is not changed by this PR, same bug in on `daily` but not on `community`. @asaadmahmood FYI

hanzei:
2. Would you mind updating it as it's the "one true" example right now?
3. Similar request as for 2.
4. According to https://github.com/mattermost/mattermost-plugin-demo/blob/1a4fdde3c6d9fb74966c7add5fcad10c8c124008/server/command_hooks.go#L149-L158 only URLs should be allowed as input, but I can input what ever and no error is shown.
5. :+1: 
6. Should I open a ticket for this?
7. Moved to https://github.com/mattermost/mattermost-plugin-demo/pull/99#issuecomment-617071047
8. Does it expect `[server]`? Why can I as a user submit the command if `[server]` is not provided?
9. :+1:

iomodo:
@hanzei 
2. 👍 
3. 👍 
4. ahh it will be fixed with 6
6. I'll do it
7. 👍 
8. User can submit anything they want, right now autocomplete is just the suggestions and does not force anything. Also related to 6, where developer will decide to force or not the desired behavior.

model/command_autocomplete.go
```diff
@@ -300,21 +336,21 @@ func (a *AutocompleteArg) UnmarshalJSON(b []byte) error {
...
+				Item: item,
+				Hint: hint,
 			})
 		}
 		a.Data = &AutocompleteStaticListArg{PossibleArguments: possibleArguments}
 	} else if a.Type == AutocompleteArgTypeDynamicList {
 		m, ok := data.(map[string]interface{})
 		if !ok {
-			return errors.Errorf("Wrong Data type in the DynamicList argument %s", string(b))
+			return errors.Errorf("Wrong type type in the DynamicList argument %s", string(b))
```
larkox, line 353: Wrong type type -> Wrong type

iomodo:
@levb @larkox 
Additional commits address review issues, also added support of the "named" arguments and changed UX as suggested by @asaadmahmood

app/plugin_requests.go
```diff
@@ -201,5 +201,7 @@ func (a *App) servePluginRequest(w http.ResponseWriter, r *http.Request, handler
 	r.URL.RawQuery = newQuery.Encode()
 	r.URL.Path = strings.TrimPrefix(r.URL.Path, path.Join(subpath, "plugins", params["plugin_id"]))
 
+	println(fmt.Sprintf("r.URL3 = %v", r.URL))
```
hanzei, line 204: Debug code?

metanerd:
> Failed running enterprise tests. @mattermost/core-build-engineers have been notified. Error:
> `invalid character ':' looking for beginning of object key string`

I need to fix the destination branch pickup. This should be fine to merge for now into `MM-20684-Autocomplete-Initial-Implementation` and rerun EE tests from there.

api4/command.go
```diff
@@ -362,6 +363,39 @@ func listAutocompleteCommands(c *Context, w http.ResponseWriter, r *http.Request
...
+
+	commands, err := c.App.ListAutocompleteCommands(c.Params.TeamId, c.App.T)
+	if err != nil {
+		c.Err = err
+		return
+	}
+
+	suggestions := c.App.GetSuggestions(commands, userInput, roleId)
+
+	w.Write(model.AutocompleteSuggestionsToJSON(suggestions))
```
levb, line 396: not sure how it's handled in the server code generally, but shouldn't the `Content-Type` be set here?

app/command_autocomplete.go
```diff
@@ -4,109 +4,189 @@
...
 	"strings"
 
+	"github.com/mattermost/mattermost-server/v5/mlog"
 	"github.com/mattermost/mattermost-server/v5/model"
 )
 
 // GetSuggestions returns suggestions for user input.
-func (a *App) GetSuggestions(commands []*model.AutocompleteData, userInput, roleID string) []model.AutocompleteSuggestion {
+func (a *App) GetSuggestions(commands []*model.Command, userInput, roleID string) []model.AutocompleteSuggestion {
+	sort.Slice(commands, func(i, j int) bool {
```
levb, line 17: 1/5 cache (or normalize) the lowercase representations to avoid repetitive lowercasing. A separate ticket is ok.

app/integration_action.go
```diff
@@ -325,27 +325,34 @@ func (w *LocalResponseWriter) WriteHeader(statusCode int) {
 	w.status = statusCode
 }
 
-func (a *App) DoLocalRequest(rawURL string, body []byte) (*http.Response, *model.AppError) {
+func (a *App) doPluginRequest(method, rawURL string, values url.Values, body []byte) (*http.Response, *model.AppError) {
```
levb, line 328: 1/5 maybe as a separate ticket - I _think_ we have several places in the server code where plugin requests are composed; locate and consolidate?

levb:
@iomodo @hanzei @lieut-data - I wonder if we can roll this out as an "experimental" or "beta within beta" feature. I personally will not have confidence in its (interface) stability until we convert a few real plugins to use it. We should IMO discourage anyone else from relying on it until we stabilize.

lieut-data:
@levb, I'm fine with that, and would be happy to help test it as part of incident response.

iomodo:
@levb, I think we can merge this to the feature branch, meanwhile I'll start implementing autocomplete on couple of plugins and let's review it later(with plugin implementations), when we merge to master.

levb:
@iomodo I think we should be merging to master - we want to be using autocomplete on community servers ASAP; 0/5 on having a kill switch/feature flag. My only concern is that 3rd party developers don't start using it while the APIs are still changing, can be solved with documentation perhaps?

hanzei:
> My only concern is that 3rd party developers don't start using it while the APIs are still changing

I don't see this as a concern. The main motivation to merge now is that we can start trying it.

api4/command.go
iomodo: Content type is set by default for all REST API calls [here](https://github.com/mattermost/mattermost-server/blob/182c29b456eeb3df9a9c757dcc630995d537a44c/web/handlers.go#L158-L159)

app/command_autocomplete.go
iomodo: Normalizing automatically will not do, because if content is uppercase we want to output suggestions in uppercase(e.g. `/jira assign MM-12320 Lev`). We will still autocomplete even if user types in lowercase.

app/integration_action.go
iomodo: I'll create a ticket.

PR #14262, opened 2020-04-08
Fix community signup link

hmhealey:
The invite link in the readme is no longer valid, so people can't sign up for community. I changed it to point to the login screen since people can sign up from there anyway.


jasonblais:
Wondering where else the sign up link is broken

hmhealey:
I only checked the webapp repo (which points to pre-release) and the mobile repo (which points directly to the mobile channel), and both of those still work. I don't know where else we're using an actual invite link instead of just pointing to the server directly.

hmhealey:
The only reason I could see for an invite link is that it saves new users from having to find the sign up link and click on it, but that's about it. I think the risk of it breaking is a bigger issue since we may rotate keys at some point.

hanzei:
@hmhealey There a couple of places in the dev docs where the invite link is used. Are you interested in also fixing these places?

PR #14264, opened 2020-04-08
Add check-store-layers check to the CI and include their updates

mgdelacroix:
#### Summary
Adds a check in parallel with the `app-layers` check to check that the `store-layers` are updated.

image

### DIFF
--- a/store/opentracing_layer.go
+++ b/store/opentracing_layer.go
@@ -1588,6 +1588,24 @@ func (s *OpenTracingLayerChannelStore) RemoveMember(channelId string, userId str
 	return resultVar0
 }
 
+func (s *OpenTracingLayerChannelStore) RemoveMembers(channelId string, userIds []string) *model.AppError {
+	origCtx := s.Root.Store.Context()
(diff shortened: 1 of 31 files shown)


store/opentracing_layer.go
```diff
@@ -5431,6 +5449,42 @@ func (s *OpenTracingLayerRoleStore) Save(role *model.Role) (*model.Role, *model.
 	return resultVar0, resultVar1
 }
 
+func (s *OpenTracingLayerSchemeStore) CountByScope(scope string) (int64, *model.AppError) {
```
ethervoid, line 5452: No need to add these changes, I've made a PR for it [here](https://github.com/mattermost/mattermost-server/pull/14261)

mgdelacroix: Great! Updated the PR to remove everything but the check itself then :)

mgdelacroix:
Marking this PR as blocked by https://github.com/mattermost/mattermost-server/pull/14261 as the `check-store-layers` will fail until that PR is merged

PR #14268, opened 2020-04-10
MM-23800: remove goroutineID and stack printing

agnivade:
#### Summary

Each hub has a goroutineID which is calculated with a known hack.
The FAQ clearly explains why goroutines don't have an id:
https://golang.org/doc/faq#no_goroutine_id.

We only added that because sometimes the hub would be deadlocked and
having the goroutineID would be useful when getting the stack trace.
This is also problematic in stress tests because the hubs would
frequently get overloaded and the logs would unnecessarily have stack traces.

But that was in the past, and we have done extensive testing with
load tests and fuzz testing to smooth any rough edges remaining.
Including adding additional metrics for hub buffer size.

Monitoring the metrics is a better way to approach this problem.
Therefore, we remove these kludges from the code.

#### Ticket Link

https://mattermost.atlassian.net/browse/MM-23800


agnivade:
I think the original deadlock was fixed, and we just kept this code in case any future ones happen again. The whole objective of the testing by covering extra paths and fuzz testing was to figure out exactly this. Whether there is actually any deadlock, or are we just afraid to touch the code.

So far, I haven't seen any deadlock, and with Mario's new metrics, we have a much better picture now. I think it's time to get rid of this.

agnivade:
@ogi-m - I have marked QA testing area to be "websockets". Anything covering that area should be tested. There are 2 more PRs covering websockets going to come your way in the future. Just a heads up :)

agnivade:
ping @ogi-m

PR #14272, opened 2020-04-10
Ignore VSCode .code-workspace files.

mkraft:
#### Summary

Ignores VSCode's .code-workspace files.

#### Ticket Link

n/a

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -36,7 +36,10 @@ imports/imports.go
 # Folders
 _obj
 _test
+
+# VSCode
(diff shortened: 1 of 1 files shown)


mkraft:
> This seems to be needed for multi-root workspaces. Is this for the server and enterprise repos ? The enterprise repo is technically at the same level at the server directory.

It's for any multiple repos, but after opening this I realized it should probably just be in my personal global gitignore, so I'm going to close this.

PR #14275, opened 2020-04-12
MM-23935 extend session expiry on user activity

wiggin77:
#### Summary

This PR adds a configuration option to extend session timeouts based on user activity (typing, switching channels).

- add `ServiceSettings.ExtendSessionLengthWithActivity` (bool) config option

- if user types anything before a session expires the session will be extended to now + session length

- if user switches channels before a session expires the session will be extended to now + session length

- session revocation via sessionIdleTimeout is disabled when `ServiceSettings.ExtendSessionLengthWithActivity` is enabled

- new session store func for updating session ExpiresAt

- includes unit tests for store layer and app.

- session length defaults for mobile and web/ldap changed from 180 days to 30 days per spec

Specs:  https://docs.google.com/document/d/1r-Eyucw5piSHAv_O7JTifcLkt8ax9tyXxroa4LMr7SE/edit

#### Ticket Link

https://mattermost.atlassian.net/browse/MM-23935


esethna:
Thanks @wiggin77, does this include the default on for existing TE servers and all new installs? https://docs.google.com/document/d/1r-Eyucw5piSHAv_O7JTifcLkt8ax9tyXxroa4LMr7SE/edit#heading=h.7mu1883pmoxo

wiggin77:
> Thanks @wiggin77, does this include the default on for existing TE servers and all new installs? 

@esethna  Definitely on for new installs. I **think** it will be on for existing installs after upgrade in that the property is missing from existing config.json and in code it defaults to on.  We'll have to test to make sure.

esethna:
@wiggin77 what about existing enterprise installs? See expected in the spec: https://docs.google.com/document/d/1r-Eyucw5piSHAv_O7JTifcLkt8ax9tyXxroa4LMr7SE/edit#heading=h.xvb25mxplen7

esethna:
@wiggin77 Not seeing the option on this page yet?

image

esethna:
@wiggin77 will a follow-up PR be coming for the webapp changes to the sys console?

wiggin77:
> @wiggin77 what about existing enterprise installs? See expected in the spec:

Enterprise installs default to `ExtendSessionLengthWithActivity=false` and team defaults to `ExtendSessionLengthWithActivity=true`

wiggin77:
> @wiggin77 will a follow-up PR be coming for the webapp changes to the sys console?

@esethna I assumed a front-end developer would be making the UI changes.  I could take a crack at it but my React Native is rusty as an old nail.

esethna:
> Enterprise installs default to `ExtendSessionLengthWithActivity=false` and team defaults to `ExtendSessionLengthWithActivity=true`

@wiggin77 sorry, just to be clear, can you verify the expected is:

1. Existing TE servers get `ExtendSessionLengthWithActivity=true`
2. Existing EE servers get `ExtendSessionLengthWithActivity=false`
3. All **new servers** (TE and EE) get `ExtendSessionLengthWithActivity=true`

wiggin77:
> @wiggin77 sorry, just to be clear, can you verify the expected is:
> 
> 1. Existing TE servers get `ExtendSessionLengthWithActivity=true`
> 2. Existing EE servers get `ExtendSessionLengthWithActivity=false`
> 3. All **new servers** (TE and EE) get `ExtendSessionLengthWithActivity=true`

I'm glad you asked. I printed the spec before I started on it and the "New Installs" header was under the staple fold.  That said, I don't think I can easily determine a new install from an upgrade wrt config defaults.  So, the new EE installs default to `ExtendSessionLengthWithActivity=false` at the moment.  I will look into it.

esethna:
@wiggin77 can you confirm the following:

1. ExtendSessionLengthWithActivity is enabled by default for new servers?
2. ExtendSessionLengthWithActivity is disabled by default for existing servers
3. Defaults for session length are unchanged from their current defaults, ie 180 days for web and mobile, 30 days for SSO

@DSchalla is there a reason SSO session length is 30 days, or can we increase to 180 to be consistent with other session lengths and improve UX?

wiggin77:
> 1. ExtendSessionLengthWithActivity is enabled by default for new servers?

confirmed

> 2. ExtendSessionLengthWithActivity is disabled by default for existing servers

confirmed

> 3. Defaults for session length are unchanged from their current defaults, ie 180 days for web and mobile, 30 days for SSO
> 

Confirmed.  This PR does not change those defaults.

DSchalla:
FYI, will check after QA review.

lindy65:
Thanks @wiggin77 - I'll need a couple of days to test this on the spinwick from mobile and web. Is there a way you can set up SSO on the spinwick for me so I can test that option too?

FYI cc// @DSchalla

lindy65:
FYI, I am in the process of testing session expires on mobile and webapp - need to wait 24 hours...

lindy65:
Hi @wiggin77 - I have tested session expiry on webapp and mobile apps and this is working as expected.

I have not yet tested on SSO but will try to get GitLab set up on the spinwick later today

lindy65:
OK, GitLab SSO is set up on the Spinwick and I am testing session expiry for the next 24 hours 👍

lindy65:
Hi @wiggin77,

I logged in to the spinwick via GitLab yesterday and made a post at 12:12pm and have not visited that web page since.  Session expiry in the System Console for SSO is set at 1 day.

image

I went to check now (12:55pm) and the user has not been logged out.

Looking at the help text in the Sys Console, it seems the user might have automatically been logged back in so I am going to log out now and re-test over the next 24 hours.

image

lindy65:
Hi @wiggin77,

Apologies for the delayed testing due to having to wait a few days for sessions to expire...

As per your test steps in the Jira ticket, I have tested that sessions extend and sessions expire after 1 day on webapp, mobile and SSO (GitLab).  However, I am not able to test:

> - test that "ServiceSettings.ExtendSessionLengthWithActivity" is `false` for upgrade to existing install

> - test that "ServiceSettings.ExtendSessionLengthWithActivity" is `true` for new installation

Is this sufficient?

//cc @DSchalla

DSchalla:
@esethna I had to dig a little bit for it, but it was changed in March 2019 only for Web and Mobile by @crspeller here: https://github.com/mattermost/mattermost-server/pull/10453 - With this feature now in place, I would advocate that we can change everything back to 30 days since this feature is a more effective fix than just increasing the timeout. If a user is away for more than 30 days on all devices, I suspect a log out is reasonable. The old increase was mainly to make log outs even for active users less frequent, which this PR fixes.

crspeller:
@wiggin77 Don't have much context into this, but for my own interest, how is this different then the `SessionIdleTimeout` setting?

wiggin77:
> @wiggin77 Don't have much context into this, but for my own interest, how is this different then the `SessionIdleTimeout` setting?

Hi @crspeller , the `SessionIdleTimeout` only forces a check of the `ExpiresAt` time versus current time.  It does not log out or prevent being logged out itself.  The `ExpiresAt` time is always what determines when the logout will happen so that is what is manipulated when extending the session.

wiggin77:
> However, I am not able to test:
> 
> > * test that "ServiceSettings.ExtendSessionLengthWithActivity" is `false` for upgrade to existing install
> 
> > * test that "ServiceSettings.ExtendSessionLengthWithActivity" is `true` for new installation
> 
> Is this sufficient?

I think this is critical to verify so we'll have to figure out a way to test it.  It can be tested locally, by installing fresh and checking the default; clean the machine by deleting server, then installing 5.20 and upgrading to check the default again.

crspeller:
@wiggin77 It does though: https://github.com/mattermost/mattermost-server/blob/master/app/session.go#L75

It checks if the time since last activity is greater than the value given and revokes the session if it is. How is what is implemented here different then just setting the idle timeout the the required value and setting the actual timeout to a large number?

wiggin77:
> @wiggin77 It does though: https://github.com/mattermost/mattermost-server/blob/master/app/session.go#L75
> 
> It checks if the time since last activity is greater than the value given and revokes the session if it is. How is what is implemented here different then just setting the idle timeout the the required value and setting the actual timeout to a large number?

@crspeller I'm going from memory, and can't look at the code for another half hour, but `SessionIdleTimeout` is updated for any websocket request regardless if triggered by user activity or automatically by client.  Even if a websocket request is never generated without user activity (as defined by the spec) today, that leaves a brittle adherence to spec as a client could be changed to generate traffic periodically.

wiggin77:
... adding on to previous comment having reviewed the code again.  `SessionIdleTimeout` is very much a transport monitor, checking for no traffic on a socket.  What we are trying to achieve with this PR is to have explicit, well defined rules that determine when a session can be extended versus revoked.  Those rules may evolve over time and IMO should not be tied to transport.

In fact, I don't think `SessionIdleTimeout` is needed any more, and I would remove it except for the need to maintain current behaviour for existing installs.

esethna:
@wiggin77 correct, session idle timeout is not needed when extending sessions is enabled. That's why the spec for the system console UI states to hide SessionIdleTimeout if extending sessions is enabled.  

image

wiggin77:
> correct, session idle timeout is not needed when extending sessions is enabled.

Agreed.  The code in this PR actually disables the code @crspeller linked above when `ExtendSessionLengthWithActivity` is enabled.  We can't remove the code though, at least not yet.

crspeller:
@wiggin77 If it's intended as a replacement that's fine. We should queue a removal of SessionIdleTimeout for 6.0! 

Are we updating the cookies anywhere? Otherwise the sessions will expire on the client anyway.

wiggin77:
> Are we updating the cookies anywhere? Otherwise the sessions will expire on the client anyway.

Thanks @crspeller .  Good point.  I'll check on this.

esethna:
> With this feature now in place, I would advocate that we can change everything back to 30 days since this feature is a more effective fix than just increasing the timeout. If a user is away for more than 30 days on all devices, I suspect a log out is reasonable. The old increase was mainly to make log outs even for active users less frequent, which this PR fixes.

@DSchalla The motivation behind this entire feature is to improve ux for users by not logging them out as frequently. This includes frequent users and infrequent users. A lot of users use Mattermost the check their work schedules, for example, and I'd be concerned that reducing the session lengths back to 30 days would decrease the effectiveness of this feature for infrequent Mattermost users. 

I would strongly lean toward keeping the existing defaults unless this is a requirement from sec team.

lindy65:
> I think this is critical to verify so we'll have to figure out a way to test it

Thanks @wiggin77 - I will ask one of the SDETs on the QA team to help test.

DSchalla:
Thanks for the feedback @esethna!

Do we have precise feedback in this case? Is more than 180 days, which was the default for over a year, called in this case frequently? Are we certain that the feedback is not coming from users where their administrators had to configure lower session duration by compliance regulations?

I fear I am struggling to see many scenarios where people open Mattermost less frequently than 30 days, even in the example of work schedules. If you open and type/switch a channel in Mattermost at least once a month you'll never be logged out with this feature, previously this was not the case so I could see the 180 days session duration being a band-aid until this was in place.

If we reduce the inactivity timeout to 30 days for all sessions, it would be consistent and decrease the likelihood that an attacker would get access to a session token of an inactive user and being able to abuse it. Increasing SSO sessions to 180 days would be consistent as well, but having session duration that last at least half a year *and* are auto extended if you are active are quite uncommon. This is not a hard veto, but I am hesitant to reduce the session security further without providing any value gain.

Aside from that, we should investigate whether the logouts are actually caused by either server misconfiguration or the mobile app logging people out because it received a status code 401 back. This change won't help in those cases.

esethna:
Thanks @DSchalla, I'm afraid we don't have any concrete data aside from qualitative feedback and user stories. But given the previous defaults (which the majority of servers are using) were 180 days for web/mobile, and the **idle timeout is 30 days**, I would assume users were in fact hitting the 30 day idle timeout value and getting logged out. 

If we decrease the default session length to 30 days, even with the auto extending sessions feature, this would mean no change to user experience for these infrequent users.

wiggin77:
Per channel discussion @esethna & @DSchalla 
 https://community.mattermost.com/core/pl/9d463yu64pn4m8fy1wi1qrbbfy I've modified the new install session length defaults to 30 days for web /ldap, mobile and SSO.

I also added code to extend the max age of session cookie stored on web clients whenever the session expired is extended per @crspeller 's comments above.

Unfortunately this invalidates the testing for default configs.  Apologies @prapti .

prapti:
Ah! No worries, @wiggin77! Earlier today, I checked the config in the cloud server once it was up and the default value for ExtendSessionLengthWithActivity was false for new install. I was thinking of continuing more testing just now (was dealing with a flat tire issue for the past couple hours).
Does that mean we'll need to check for different values in the server config or wait on testing those for now?

wiggin77:
> Does that mean we'll need to check for different values in the server config or wait on testing those for now?

The defaults are updated in this PR now.  Can be tested any time.  Here are the expected values for config.json:

Upgrade existing install:
```
ServiceSettings.ExtendSessionLengthWithActivity: false
ServiceSettings.SessionLengthWebInDays: 180
SessionLengthWebInDays.SessionLengthMobileInDays: 180
SessionLengthMobileInDays.SessionLengthSSOInDays: 30
```

New install:
```
ServiceSettings.ExtendSessionLengthWithActivity: true
ServiceSettings.SessionLengthWebInDays: 30
SessionLengthWebInDays.SessionLengthMobileInDays: 30
SessionLengthMobileInDays.SessionLengthSSOInDays: 30
```

esethna:
@DSchalla can you review and approve if this PR is ready to merge?

prapti:
I can't get a cloud instance to spin up for the latest commit (hash: 83386d6). Wondering if the ee/circleci check should pass too before the cloud plugin can recognize the commit hash. Does anyone know how to fix that check?

wiggin77:
> Does anyone know how to fix that check?

I did an update branch - hopefully that will clear it up.

wiggin77:
> I can't get a cloud instance to spin up for the latest commit (hash: [83386d6](https://github.com/mattermost/mattermost-server/commit/83386d62734707d02b51851a902f36f516e2f2f4)).

@prapti , is it working now?  CI checks are green now.

PR #14277, opened 2020-04-13
MM-23805: Refactor web_hub

agnivade:
#### Summary

This is a beginning of the refactoring of the websocket code.

To start off with, we unexport some methods and constants which did not
need to be exported. There are more remaining but some are out of scope for this PR.

The main chunk of refactor is to unexport the webconn send channel
which was the main cause of panics. Since we were directly sending
to the connection from various parts of the codebase, it would be possible
that the send channel would be closed and we could still send a message.
This would crash the server.

To fix this, we refactor the code to centralize all sending from the main
hub goroutine. This means we can leverage the connections map to check
if the connection exists or not, and only then send the message.

#### Ticket Link

https://mattermost.atlassian.net/browse/MM-23805


app/cluster.go
```diff
@@ -30,3 +30,117 @@ func (s *Server) InvokeClusterLeaderChangedListeners() {
 		})
 	})
 }
+
+func (a *App) Publish(message *model.WebSocketEvent) {
```
jespino, line 34: This is not strictly cluster relate, this is related to sending thing through the websockets, and, because we support clustering, if we have a cluster configured, we also propagate that event to the cluster.

Maybe we can separate it and rename it a bit, something like "PublishMessageInLocalHub" and then create other for the "PublishMessageInCluster" or something like that.

Anyway, I think moving this methods here is not appropriate because is not the cluster code responsability, is more in the area of the hub responsability. IMO.

agnivade: Alright, no worries. I'll bring those back in.

agnivade: done

app/web_hub.go
```diff
@@ -103,6 +156,7 @@ func (a *App) GetHubForUserId(userId string) *Hub {
 	return hub
 }
 
+// HubRegister registers a connection to a hub.
 func (a *App) HubRegister(webConn *WebConn) {
```
streamer45, line 160: Should we check if `webConn` is `nil`?

app/web_hub.go
```diff
@@ -113,6 +167,7 @@ func (a *App) HubRegister(webConn *WebConn) {
 	}
 }
 
+// HubUnregister unregisters a connection from a hub.
 func (a *App) HubUnregister(webConn *WebConn) {
```
streamer45, line 171: Same as above, should we check if `webConn` is `nil`?

app/web_hub.go
```diff
@@ -395,10 +445,10 @@ func (h *Hub) Start() {
 				for _, webCon := range candidates {
 					if webCon.ShouldSendEvent(msg) {
 						select {
-						case webCon.Send <- msg:
+						case webCon.send <- msg:
 						default:
 							mlog.Error("webhub.broadcast: cannot send, closing websocket for user", mlog.String("user_id", webCon.UserId))
-							close(webCon.Send)
+							close(webCon.send)
```
streamer45, line 451: I see we close the `send` channel in two separate places. Here and in the `directMsg` case above.
Are we safe to do so? Should we add here a check with `connections.Has` like we do above?

agnivade: Yes, definitely. Good point.

app/web_hub.go
```diff
@@ -250,85 +292,82 @@ func (a *App) invalidateCacheForUserTeams(userId string) {
...
 func (h *Hub) Unregister(webConn *WebConn) {
 	select {
 	case h.unregister <- webConn:
 	case <-h.stop:
 	}
 }
 
+// Broadcast broadcasts the message to all connections in the hub.
 func (h *Hub) Broadcast(message *model.WebSocketEvent) {
 	if h != nil && h.broadcast != nil && message != nil {
```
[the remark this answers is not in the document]
agnivade: Will check .. and if not needed, will remove.

app/web_hub.go
agnivade: Maybe a bit unnecessary. It is only called from api4/websocket and app/websocket_router. Both of them cannot deal with nil connections.

streamer45: Alright. As long as we make sure from the calling side then it's okay.

app/web_hub.go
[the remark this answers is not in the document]
agnivade: done

agnivade:
1. ack
2. ack
3. I am aware of this, and I have refrained from doing it in this PR because it falls under performance optimizations. There is also the matter of the hub count becoming stale by the time we get to the 2nd line using that value. I will look into it separately.
4. nice catch.
5. mmm .. you mean there is a chance we can bump up the metric if Broadcast is called after Hub is stopped ?

streamer45:
> 3. I am aware of this, and I have refrained from doing it in this PR because it falls under performance optimizations. There is also the matter of the hub count becoming stale by the time we get to the 2nd line using that value. I will look into it separately.

Sure. Just commenting what I see :)

> 5. mmm .. you mean there is a chance we can bump up the metric if Broadcast is called after Hub is stopped ?

Yes. No big deal just thought it was worth mentioning it.

agnivade:
1. Fair enough.
2. Wow yeah, that's a good one. I think we can also get rid of the `hubConnectionIndexIndexes` then. It's sole purpose is to give a reverse index of the slice given a webconn, to the other 2 slices. I am wondering whether there's even any impact with iterating the `connectionsByUserId` slice, given the fact that 90% of the time, it will be just one entry. I am noting it down, and will do it after the webconn refactor is also merged.

By now, I will get nasty merge conflicts for my webconn PR :P

agnivade:
@lindy65 - I have outlined the testing strategy in the jira ticket. Basically this is a big websocket refactor. Please test whatever you can related to websockets. We would want to ensure that all the websocket broadcast events work fine. And also user statuses work fine. Feel free to test any additional areas that you feel appropriate.

Thanks !

agnivade:
@lindy65 - Hopefully you haven't started testing this. I just merged the other PR which was part of the same PR chain after this one. So you don't have to test that PR again separately. It's all doing the same thing, and requires the same testing steps. So nothing changes from your end.

agnivade:
The CI is failing because it's not up to date with the latest master. If the spinwicks fail to build in that case, then this needs to wait until @ogi-m is done testing #14268.

lindy65:
Hi @agnivade - I haven't actually started testing this yet.  Ogi and I are syncing up later to discuss the test plan for websockets so I'll start after that...

agnivade:
Alright, sounds good. Thanks !

agnivade:
There seems to be a valid failure. Looking into it.

app/web_hub.go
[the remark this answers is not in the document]
agnivade: Well .. surprise surprise. This was needed after all :)


The hub nil check is required because of the wonderful way we setup our tests :) We call `s, err := app.NewServer(options...)` which returns a server, but only after that, we call `wsapi.Init(th.App, th.App.Srv().WebSocketRouter)` through our FakeApp adapter to initialize the hub. But in the `NewServer` call itself, we call `RunOldAppInitialization` which directly proceeds to broadcast some messages happily. 

```
message.Add("config", s.FakeApp().ClientConfigWithComputed())
s.Go(func() {
	s.FakeApp().Publish(message)
})
```

Therefore, to make this work, we incur a nil check for every single message broadcast. Just found this out because CI failed on hub Broadcast crash.

streamer45:
Ran a couple of load-test, everything looks good :+1:

agnivade:
Awesome, thanks !

PR #14291, opened 2020-04-14
Fix broken master build by updating store

fmunshi:
#### Summary
Looks like this https://github.com/mattermost/mattermost-server/pull/14010 was merged without running `make store-layers` maybe before the ci check was added? 

#### Ticket Link
- none dev only changes


streamer45:
@fmunshi May also need `make store-mocks`.

fmunshi:
done @streamer45

PR #14297, opened 2020-04-16
MM-23724 add RestoreTeam endpoint

wiggin77:
#### Summary

This PR implements a new API endpoint which restores a soft deleted team.

Includes units tests.

API docs here: https://github.com/mattermost/mattermost-api-reference/pull/512

#### Ticket Link

https://mattermost.atlassian.net/browse/MM-23724


api4/team.go
```diff
@@ -231,6 +232,40 @@ func patchTeam(c *Context, w http.ResponseWriter, r *http.Request) {
...
+
+	// Return the restored team to be consistent with RestoreChannel.
+	team, err := c.App.GetTeam(c.Params.TeamId)
+	if err != nil {
+		c.Err = err
+		return
+	}
+
+	auditRec.AddMeta("team", team)
+	auditRec.Success()
```
saturninoabril, line 264: Probably not related to this PR but asking for reference. Do we have docs/guides on how/where to put audit record/log? Like does the order matter? In other places, I'm seeing `Success` is called first before `AddMeta`.

wiggin77: The order doesn't matter.  I'll look at writing a how-to doc.

api4/team_test.go
```diff
@@ -427,6 +427,74 @@ func TestPatchTeam(t *testing.T) {
...
+			AllowOpenInvite: (teamType == model.TEAM_OPEN),
+			InviteId:        model.NewId(),
+			Name:            "aa-" + model.NewRandomTeamName() + "zz",
+			Email:           "success+" + model.NewId() + "@simulator.amazonses.com",
+			Type:            teamType,
+		}
+		team, _ = Client.CreateTeam(team)
+		require.NotNil(t, team)
+		if deleted {
+			th.SystemAdminClient.SoftDeleteTeam(team.Id)
```
ali-farooq0, line 450: Should assert success on the return values here.

wiggin77: Thanks.  Check added https://github.com/mattermost/mattermost-server/pull/14297/commits/1937849dca14adb08e33634cbc30802f4fbbaa5d

PR #14299, opened 2020-04-16
MM-24135: Migrate AppError from SaveChannel/channel_store.go

agnivade:
#### Summary

This is the first POC of migration of store app errors to plain error.

We create a few basic error types in the store package and use
them to return the errors from store methods. In the app layer,
we inspect the error and re-create the exact app errors. This lets
us preserve the same error content, but yet move to plain errors.

Since this is a gradual migration, this means that the error inspection
code will be duplicated across the app layer whenever a store method
is invoked. But all of that should go away once we start propagating
the errors higher up the hierarchy.

There has been a significant amount of changes in the storetest and searchtest
layer, primarily because we have to rename the err variable now that it is of
a different type.

#### Ticket Link

https://mattermost.atlassian.net/browse/MM-24135


agnivade:
The eagle has landed :)

My main concern was that the error handling code is duplicated for the same store method in the app layer. But I realised that once we propagate the error higher up the stack the error handling will be de-duplicated, and it will be only in the api layer. So I guess that's what we have to live with for now.

The main chunk of the code is in app/channel.go and store/sqlstore/channel_store.go. Rest are all fixes to make everything work.

app/channel.go
```diff
@@ -411,12 +451,49 @@ func (a *App) createGroupChannel(userIds []string, creatorId string) (*model.Cha
...
+			switch {
+			case invErr.Entity == "Channel" && invErr.Field == "DeleteAt":
+				return nil, model.NewAppError("CreateChannel", "store.sql_channel.save.archived_channel.app_error", nil, "", http.StatusBadRequest)
+			case invErr.Entity == "Channel" && invErr.Field == "Type":
+				return nil, model.NewAppError("CreateChannel", "store.sql_channel.save.direct_channel.app_error", nil, "", http.StatusBadRequest)
+			case invErr.Entity == "Channel" && invErr.Field == "Id":
+				return nil, model.NewAppError("SqlChannelStore.Save", "store.sql_channel.save_channel.existing.app_error", nil, "id="+invErr.Value.(string), http.StatusBadRequest)
+			}
+		case errors.As(nErr, &intErr):
+			switch intErr.Location {
```
jespino, line 472: I don't see the value of differentiating the different error types here. I would say, this is the case were I would return a regular `errors.New` instance, and here I would just return a generic message for this endpoint something like `model.NewAppError("CreateChannel", "app.channel.create_channel.internal_error", nil, intErr.Error(), http.StatusInternalServerError)` where the string for `app.channel.create_channel.internal_error` is something like "Unable to save the new channel"

app/channel.go
```diff
@@ -411,12 +451,49 @@ func (a *App) createGroupChannel(userIds []string, creatorId string) (*model.Cha
...
+				return nil, model.NewAppError("CreateChannel", "store.sql_channel.save.upsert_public_channel.app_error", nil, intErr.Error(), http.StatusInternalServerError)
+			case "commit_transaction":
+				return nil, model.NewAppError("SqlChannelStore.Save", "store.sql_channel.save.commit_transaction.app_error", nil, intErr.Error(), http.StatusInternalServerError)
+			}
+		case errors.As(nErr, &cErr):
+			if cErr.Resource == "Channel" {
+				return channel, model.NewAppError("CreateChannel", store.CHANNEL_EXISTS_ERROR, nil, cErr.Error(), http.StatusBadRequest)
+			}
+		case errors.As(nErr, &ltErr):
+			if ltErr.What == "channels_per_team" {
```
jespino, line 489: If we only have one possible limit to exceed there, I would remove the check entirely

app/channel.go
```diff
@@ -411,12 +451,49 @@ func (a *App) createGroupChannel(userIds []string, creatorId string) (*model.Cha
...
+				return nil, model.NewAppError("CreateChannel", "store.sql_channel.save_channel.save.app_error", nil, intErr.Error(), http.StatusInternalServerError)
+			case "save_channel_count":
+				return nil, model.NewAppError("CreateChannel", "store.sql_channel.save_channel.current_count.app_error", nil, intErr.Error(), http.StatusInternalServerError)
+			case "upsert_public_channel":
+				return nil, model.NewAppError("CreateChannel", "store.sql_channel.save.upsert_public_channel.app_error", nil, intErr.Error(), http.StatusInternalServerError)
+			case "commit_transaction":
+				return nil, model.NewAppError("SqlChannelStore.Save", "store.sql_channel.save.commit_transaction.app_error", nil, intErr.Error(), http.StatusInternalServerError)
+			}
+		case errors.As(nErr, &cErr):
+			if cErr.Resource == "Channel" {
```
jespino, line 485: Same here, If i only can return a specific error, I would save the check

app/channel.go
```diff
@@ -411,12 +451,49 @@ func (a *App) createGroupChannel(userIds []string, creatorId string) (*model.Cha
...
+		var appErr *model.AppError
+		switch {
+		case errors.As(nErr, &invErr):
+			switch {
+			case invErr.Entity == "Channel" && invErr.Field == "DeleteAt":
+				return nil, model.NewAppError("CreateChannel", "store.sql_channel.save.archived_channel.app_error", nil, "", http.StatusBadRequest)
+			case invErr.Entity == "Channel" && invErr.Field == "Type":
+				return nil, model.NewAppError("CreateChannel", "store.sql_channel.save.direct_channel.app_error", nil, "", http.StatusBadRequest)
+			case invErr.Entity == "Channel" && invErr.Field == "Id":
+				return nil, model.NewAppError("SqlChannelStore.Save", "store.sql_channel.save_channel.existing.app_error", nil, "id="+invErr.Value.(string), http.StatusBadRequest)
```
jespino, line 469: Probably we can set all the wheres here as `SqlChannelStore.Save` not only this one, because the errors is returned from there.

app/channel.go
```diff
@@ -411,12 +451,49 @@ func (a *App) createGroupChannel(userIds []string, creatorId string) (*model.Cha
...
+			}
+		case errors.As(nErr, &cErr):
+			if cErr.Resource == "Channel" {
+				return channel, model.NewAppError("CreateChannel", store.CHANNEL_EXISTS_ERROR, nil, cErr.Error(), http.StatusBadRequest)
+			}
+		case errors.As(nErr, &ltErr):
+			if ltErr.What == "channels_per_team" {
+				return nil, model.NewAppError("CreateChannel", "store.sql_channel.save_channel.limit.app_error", nil, ltErr.Error(), http.StatusBadRequest)
+			}
+		case errors.As(nErr, &appErr): // in case we haven't converted to plain error.
```
jespino, line 492: Is this case possible?

agnivade: Yes, it's possible. Because  `channel.IsValid()` still returns an AppError and we cannot change that until v6. So just casting it to normal error and re-casting it back again here.

app/channel.go
agnivade: Fair enough.

app/channel.go
agnivade: Clarified offline. Going to make this case fall back to the default case.

app/channel.go
agnivade: Hmm .. actually I made a mistake. IMO we should be setting the `where` as the origin point from where the AppError was instantiated. And the AppError was instantiated from `CreateChannel`. That itself contains another error internally which contains detailed info. But we should not set `where` as the origin point of the embedded error. Let me know what you think.

jespino: Yes, lets keep it as `CreateChannel` because is where is generated, and is like that for the rest of the source code, and I don't want to include more changes in this PR. I think what you proposed is the most consistent approach with the rest of the code.

app/channel.go
jespino: Good point, but what about capturing the `IsValid` error from there (handling is a a regular error) there and returning the specific error type from the store? that way we can remove entirely any reference to `model.AppError` leaking from the store. I'm not 100% sure if that is the correct approach, what do you think?

agnivade: Resolved offline. We are going to keep as is for now. Because it will create a mess of string matching from the caller.

store/errors.go
```diff
@@ -0,0 +1,73 @@
...
+
+func (e *ErrInvalidInput) Error() string {
+	return fmt.Sprintf("invalid input: entity: %s field: %s value: %s", e.Entity, e.Field, e.Value)
+}
+
+// ErrLimitExceeded indicates an error that has occured because some value exceeded a limit.
+type ErrLimitExceeded struct {
+	What  string // What was the object that exceeded.
+	Count int    // The value of the object.
+	meta  string // Any additional metadata.
```
mkraft, line 33: What's an example of `meta` being used? I'm wondering if it makes more sense as a `string` or as a `map[string]interface{}`.

store/errors.go
```diff
@@ -0,0 +1,73 @@
...
+		Count: count,
+		meta:  meta,
+	}
+}
+
+func (e *ErrLimitExceeded) Error() string {
+	return fmt.Sprintf("limit exceeded: what: %s count: %d metadata: %s", e.What, e.Count, e.meta)
+}
+
+// ErrConflict indicates a conflict that occured.
```
mkraft, line 48: What's the reason behind `ErrConflict` vs, say, `RecordExists`? What kind of other conflicts type be used for?

store/errors.go
agnivade: That's a good point. I was trying to keep it consistent with AppError, because this is how it initially looked: `model.NewAppError("SqlChannelStore.Save", "store.sql_channel.save_channel.limit.app_error", nil, "teamId="+channel.TeamId, http.StatusBadRequest)`. 

Is there a reason why we are not passing the `teamId` in the `params` map, but rather in the `details` string ?

I am totally open to changing this.

store/errors.go
agnivade: Right now, it's very hard to say. Because this is just the first PR. That's why I kept it generic for now. But 0/5 on changing it right now.

store/errors.go
mkraft: > Is there a reason why we are not passing the teamId in the params map, but rather in the details string ?

I have no clue, I don't know the history. I guess this can come up organically in the future if necessary.

agnivade:
Wonderful, much appreciated @lindy65 !

PR #14033, opened 2020-03-10
Dependency updates.

grundleborg:
#### Summary
Dependency Updates

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-21581

PR #14304, opened 2020-04-16
Revert "Dependency updates. (#14033)"

lieut-data:
#### Summary
This reverts commit f14c79f170d3456ae9db7739f88563b37183e9a3. With @stylianosrigas's help, I've proven it resolves https://mattermost.atlassian.net/browse/MM-24165, and expect it to be related to https://mattermost.atlassian.net/browse/MM-24166 and https://mattermost.atlassian.net/browse/MM-24118 as well.

I don't currently have a detailed explanation of which dependency broke this behaviour, but @agnivade and I suspected something to do with `memberlist`. I had some trouble reverting just a subset of the dependencies, so opted to just rip it out altogether and punt the issue back to the server team for detailed analysis.

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-24165


lieut-data:
@agnivade, that's correct -- it reproduced 100% of the time against `-daily` until this commit (after 1-3 attempts), and 0% of the time against `-release` and the other servers.

lieut-data:
@lindy65, this one is "hard to test" outside of just merging it to community... which it already technically is since we've temporarily pointed `-daily` at this branch. So if you just gave `community-daily` a smoke test, I think that would suffice for QA testing.

lieut-data:
Also, it's important to note that the issue was `HA` specific, so if we do test using a spinmint, we'll need an HA equipped one too for maximum coverage.

agnivade:
@lieut-data - Got it. I'll look into it.

lindy65:
Thanks @lieut-data - appreciate the info.  I'll smoke test around the tickets listed in this PR 👍

PR #14305, opened 2020-04-16
[MM-23605] Create new team scheme updated socket event

fmunshi:
#### Summary
- Adds a new websocket event to be triggered when a scheme is assigned to a team 
- `update_team` which was previously being used only reloads the team in the state and not the memberships, so I created a new websocket event for a scheme being added/removed from a team

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-23605

### DIFF
--- a/app/team.go
+++ b/app/team.go
@@ -188,7 +188,7 @@ func (a *App) UpdateTeamScheme(team *model.Team) (*model.Team, *model.AppError) 
 		return nil, err
 	}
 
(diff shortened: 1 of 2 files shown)


agnivade:
@fmunshi - Is the webapp/mobile updated with this new event ? Or will that happen later ?

fmunshi:
Webapp is PR is up https://github.com/mattermost/mattermost-webapp/pull/5344, mobile will come just havent got around to it.

PR #14312, opened 2020-04-17
Fixing reply count on new posts

jespino:
#### Summary
Fixing reply count on new posts

#### Ticket Link
[MM-17468](https://mattermost.atlassian.net/browse/MM-17468)


jespino:
@reflog sorry, I made some changes to fix another problem.

store/sqlstore/post_store.go
```diff
@@ -703,11 +703,9 @@ func (s *SqlPostStore) getPostsAround(before bool, options model.GetPostsOptions
 		direction = ">"
 		sort = "ASC"
 	}
-	replyCountSubQuery := s.getQueryBuilder().Select("COUNT(Posts.Id)").From("Posts").Where(sq.Expr("p.RootId = '' AND RootId = p.Id AND DeleteAt = 0"))
+	replyCountSubQuery := s.getQueryBuilder().Select("COUNT(Posts.Id)").From("Posts").Where(sq.Expr("Posts.RootId = (CASE WHEN p.RootId = '' THEN p.Id ELSE p.RootId END) AND Posts.DeleteAt = 0"))
```
reflog, line 706: shouldn't it be:
`(CASE WHEN Posts.RootId = '' THEN Posts.Id ELSE Posts.RootId END) `

jespino: I don't think so, the `p.RootId` and `p.Id` is the post from outside the subquery.

PR #14317, opened 2020-04-19
Change Dockerfile HEALTHCHECK intervals

cadavre:
We should not use such a long `HEALTHCHECK` `--interval` as first task check if happening after given time, and before healthcheck returns `0` – **task won't be available** to the Docker networks – meaning that Mattermost won't work until then even if it deployed correctly!

And in Docker envs it happens often – that DB is available all the time, and just MM app is being updated. In this case even if we are healthy since 1s – we need to wait 5min until it goes online.


teresa-novoa:
Hi @cadavre I see that @hanzei requested a review on your side, let us know if you need any help. Thanks!

hanzei:
@cpanato @metanerd Could you please provide QA test steps for this PR?

teresa-novoa:
@cpanato can you help to sync and merge this PR? Thanks!

teresa-novoa:
Thanks @cpanato!

PR #14318, opened 2020-04-20
MM-24132: Migrate AppError from SaveDirectChannel/channel_store.go

agnivade:
#### Summary

This PR migrates 2 new methods SaveDirectChannel and CreateDirectChannel
to return error instead of AppError.

We also need to handle the error internally in SaveMultipleMember for now
until that is migrated too.

#### Ticket Link

https://mattermost.atlassian.net/browse/MM-24132


store/sqlstore/channel_store.go
```diff
@@ -559,81 +559,42 @@ func (s SqlChannelStore) CreateDirectChannel(user *model.User, otherUser *model.
...
-
-	if memberSaveErr != nil {
-		return nil, model.NewAppError("SqlChannelStore.SaveDirectChannel", "store.sql_channel.save_direct_channel.add_members.app_error", nil, memberSaveErr.Error(), http.StatusInternalServerError)
+	if err != nil {
+		return nil, err
 	}
 
 	if err := transaction.Commit(); err != nil {
-		return nil, model.NewAppError("SqlChannelStore.SaveDirectChannel", "store.sql_channel.save_direct_channel.commit.app_error", nil, err.Error(), http.StatusInternalServerError)
+		return nil, store.NewErrInternal("commit_transaction", err, "")
```
jespino, line 597: There is any advantage between `NewErrInternal` and use directly errors.New?

agnivade: I wanted to keep the Location of the error as a field so that we don't lose track of what is important. Otherwise we would just have a plain string of location + metadata.

We can change it to errors.New/errors.Wrap if you think keeping the structure is not important. I know that's what was decided on initially. Let me know.

agnivade:
@jespino - I will wait for the 1st one to be merged, and then cherry-pick this one on master. Apart from the ErrInternal, are there any other concerns that need to be addressed ?

store/sqlstore/channel_store.go
```diff
@@ -1352,9 +1313,35 @@ func (s SqlChannelStore) SaveMultipleMembers(members []*model.ChannelMember) ([]
...
+		case errors.As(err, &cErr):
+			switch cErr.Resource {
+			case "ChannelMembers":
+				return nil, model.NewAppError("CreateChannel", "store.sql_channel.save_member.exists.app_error", nil, cErr.Error(), http.StatusBadRequest)
+			}
+		case errors.As(err, &appErr): // in case we haven't converted to plain error.
+			return nil, appErr
+		case errors.As(err, &intErr):
+			switch {
+			case intErr.Location == "channel_roles_tosql":
```
mgdelacroix, line 1331: nit: can we change this `switch` to be a `switch intErr.Location` to keep consistency with the previous one?

agnivade: sounds good. Will do it after I merge the first one and fix the conflicts.

agnivade:
@jespino - Merged latest master, removed internal error, fixed layer generation. Requesting one final look.

agnivade:
Hi @lindy65 - same steps apply as https://mattermost.atlassian.net/browse/MM-24135. Thanks !

PR #14334, opened 2020-04-20
MM-24217 MFA secret invalidate cache

wiggin77:
#### Summary

This PR fixes a bug where activating MFA fails due to old secret being fetched from user cache.  MFA functions (`GenerateMfaSecret`, `ActivateMfa`, `DeactivateMfa`) were writing to database but not invalidating user caches.  

The bug was bisected to PR https://github.com/mattermost/mattermost-server/pull/13838 and does not exist in v5.22.  It was working accidentally before that despite not invalidating cache properly. 

#### Ticket Link

https://mattermost.atlassian.net/browse/MM-24217


wiggin77:
> Alright, so for adding any cache invalidation logic, everything should go through the cache layer.

Thanks @agnivade for looking at this.  While I agree these changes belong in the cache layer, this is not the PR for that.  There are 6 or 7 other cache related invalidations in app/user alone, and they should get refactored via a separate PR.  Some may require special attention such as `invalidateUserCacheAndPublish`.

This PR is just to fix broken daily.

agnivade:
> This PR is just to fix broken daily.

Yes I see. I was just trying to suggest a cleaner fix IMO. It wasn't a separate refactoring in my mind.

Feel free to go ahead if there is urgency around it. Did you also want to add the tests as part of a separate PR or this PR ?

wiggin77:
> Feel free to go ahead if there is urgency around it. Did you also want to add the tests as part of a separate PR or this PR ?

I'd prefer new unit tests be part of the PR to move all these to cache layer.  The unit test would be very different there as we won't be unit testing each App function for cache correctness.

wiggin77:
Hi @wiersgallak , it was recommended to me that PM Review is not needed for a simple bug fix so I've removed the label.

PR #14338, opened 2020-04-21
MM-23816: Group Mentions: Add ability to rename group names

catalintomai:
#### Summary
Covers reqs #9, #10, #11 in https://mattermost.atlassian.net/wiki/spaces/EN/pages/389710117/EE+Group+Mentions+MVP

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-23816

#### Related PRs
redux: https://github.com/mattermost/mattermost-redux/pull/1135
webapp: https://github.com/mattermost/mattermost-webapp/pull/5362


model/group.go
```diff
@@ -157,6 +159,20 @@ func (group *Group) IsValidForUpdate() *AppError {
...
 
+var validGroupnameChars = regexp.MustCompile(`^[a-z0-9\.\-_]+$`)
+
+func (group *Group) IsValidName() *AppError {
+	if l := len(group.Name); l == 0 || l > GroupNameMaxLength {
+		return NewAppError("Group.IsValidName", "model.group.name.app_error", map[string]interface{}{"GroupNameMaxLength": GroupNameMaxLength}, "", http.StatusBadRequest)
+	}
+
+	if !validGroupnameChars.MatchString(group.Name) {
+		return NewAppError("Group.IsValidName", "model.group.name.app_error", nil, "", http.StatusBadRequest)
```
sbishel, line 170: Add more detail to this error. Such as "Contains invalid characters" or a separate error.

catalintomai: added separate error.

catalintomai:
> A couple of questions: am I reading it correctly that `(*Group)IsValidName` is only invoked on create and not update? What's the reason for not using it for both.

Actually, IsValidForUpdate() calls IsValidForCreate(), so Update is covered as well.

mkraft:
> > A couple of questions: am I reading it correctly that `(*Group)IsValidName` is only invoked on create and not update? What's the reason for not using it for both.
> 
> Actually, IsValidForUpdate() calls IsValidForCreate(), so Update is covered as well.

Ah OK. So it is possible that the API `patchGroup` function can generate a `Name` field that can't be saved because it doesn't pass validation because they're using different logic (downcase & replace spaces with dashes vs the `^[a-z0-9\.\-_]+$` regex)?

catalintomai:
> IsValidForCreate

The regex is the same used for validating the username
```validUsernameChars = regexp.MustCompile(`^[a-z0-9\.\-_]+$`)```

File: api4/group.go [hunk 2]
```diff
@@ -135,6 +135,26 @@ func patchGroup(c *Context, w http.ResponseWriter, r *http.Request) {
 		if groupPatch.Name == nil {
 			tmp := strings.ReplaceAll(strings.ToLower(group.DisplayName), " ", "-")
 			groupPatch.Name = &tmp
+		} else {
+			if *groupPatch.Name == "all" || *groupPatch.Name == "channel" || *groupPatch.Name == "here" {
```
sbishel, line 139: Seems like we should already have constants for these?

catalintomai:
> MM-25015

@srkgupta , fixed MM-25015, re-submitted for review.

PR #14339, opened 2020-04-21
MM-24133: Migrate AppError from bot_store.go

agnivade:
#### Ticket Link

https://mattermost.atlassian.net/browse/MM-24133


store/opentracing_layer.go
```diff
@@ -465,11 +445,6 @@ func (s *OpenTracingLayerBotStore) Update(bot *model.Bot) (*model.Bot, *model.Ap
 
 	defer span.Finish()
 	resultVar0, resultVar1 := s.BotStore.Update(bot)
-	if resultVar1 != nil {
```
jespino, line 468: This is a symptom that the opentracing generator is not capturing errors, only `model.AppError`, probably we need to modify the generator to tackle that. Is it correct @reflog?

store/timer_layer.go
```diff
@@ -377,79 +377,79 @@ func (s *TimerLayerAuditStore) Save(audit *model.Audit) *model.AppError {
...
-func (s *TimerLayerBotStore) Get(userId string, includeDeleted bool) (*model.Bot, *model.AppError) {
+func (s *TimerLayerBotStore) Get(userId string, includeDeleted bool) (*model.Bot, error) {
 	start := timemodule.Now()
 
 	resultVar0, resultVar1 := s.BotStore.Get(userId, includeDeleted)
 
 	elapsed := float64(timemodule.Since(start)) / float64(timemodule.Second)
 	if s.Root.Metrics != nil {
 		success := "false"
-		if resultVar1 == nil {
```
jespino, line 388: Same here, we are not handling the errors in the timer layer generator, we should change that.

store/opentracing_layer.go
reflog: @jespino - I'll take a look and report back

reflog: yet. happens due to:
```go

		"errorPresent": func(results []string) bool {
			for _, typeName := range results {
				if strings.Contains(typeName, APP_ERROR_TYPE) {
					return true
				}
			}
			return false
		},
```
I'll submit the PR to fix

agnivade: Thanks @reflog !

agnivade:
@jespino - Done the same here too. PTAL.

PR #14340, opened 2020-04-21
MM-23896: Fix clearing of batched emails on user activity

agnivade:
On user activity, we were clearing the job.pendingNotifications map.
But we had already created a copy of the notifications slice while
iterating the map. Therefore, if we pass the copied slice, it would
still have the old notifications which were originally deleted.

The unit tests would not catch this because it was testing the
job.pendingNotifications map and not actually checking if the email
handler was being called or not. We fix that now.

### DIFF
--- a/app/email_batching_test.go
+++ b/app/email_batching_test.go
@@ -4,6 +4,7 @@
 package app
 
 import (
+	"sync/atomic"
 	"testing"
 	"time"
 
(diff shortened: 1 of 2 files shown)


app/email_batching_test.go
```diff
@@ -115,7 +116,28 @@ func TestCheckPendingNotifications(t *testing.T) {
...
+
+	var wasCalled int32
+	job.checkPendingNotifications(time.Unix(10050, 0), func(string, []*batchedNotification) {
+		atomic.StoreInt32(&wasCalled, int32(1))
+	})
+
+	// A hack to check whether the handler was called.
+	// It's not straightforward to just wait for it using a channel because the test should
+	// NOT call the handler, and it will be called only if the test fails.
+	time.Sleep(1 * time.Second)
```
ali-farooq0, line 136: This is to give `job.server.Go(...)` time to make sure it has had a chance to run the handler if the right conditions have met? I wonder if we can check if `job.pendingNotifications` still contains the `userId` instead? as that will guarantee that it didn't get into the new code block, yes?

agnivade: Well, this is a `require` call. So if it fails, the next statements won't be executed.

ali-farooq0: no i meant check if a pending notifications exists for a user first, invoke `checkPendingNotifications` and if it exists then you know the batch handler was never invoked. That way you can get rid of this sleep.

agnivade: Ah okay I get you now. Yes, that _would_ work with _this_ version of the code. But it won't work if my fix isn't there. And this is why this test failed in catching the bug for so long.

So if I just check for `job.pendingNotifications` which is already done 1 line below, and there is a regression later, the test will not be able to catch the failure. Hence, I did it this way.

ali-farooq0: Ahh I see. Yeah this is a tricky one. Only way we could know for sure if the go routine wasn't invoked is if we mock out `job.server.Go(...)`. Alright this'll do for now 👍

agnivade: Yeah, believe me I hate putting in `time.Sleep`s more than anything. But recently I removed 3 * `time.Sleep(2 * time.Second)` by fixing a flaky test. So I call that a win :)

agnivade:
@jespino - Friendly ping as this is slated for 5.23.

cpanato:
I'm going to test the test server fix, some spam is expected sorry

PR #14348, opened 2020-04-21
[MM-24177] Adds props to post for front end to know if highlight or not

hahmadia:
#### Summary

 This PR adds props to post if channel mentions and group mentions for that post should be highlighted or not

#### Ticket Link

https://mattermost.atlassian.net/browse/MM-24177

**Unblocks:**
https://github.com/mattermost/mattermost-webapp/pull/5372


app/post.go
```diff
@@ -393,6 +393,14 @@ func (a *App) FillInPostProps(post *model.Post, channel *model.Channel) *model.A
 		post.DelProp("channel_mentions")
 	}
 
+	if !a.HasPermissionToChannel(post.UserId, post.ChannelId, model.PERMISSION_USE_CHANNEL_MENTIONS) {
+		post.AddProp("disable_channel_highlight", true)
+	}
```
fmunshi, line 398: There is existing logic that handles a prop for channel mentions it is called `POST_PROPS_MENTION_HIGHLIGHT_DISABLED`

app/post.go
```diff
@@ -393,6 +393,14 @@ func (a *App) FillInPostProps(post *model.Post, channel *model.Channel) *model.A
...
+
+	if !a.HasPermissionToChannel(post.UserId, post.ChannelId, model.PERMISSION_USE_GROUP_MENTIONS) {
+		post.AddProp("disable_group_highlight", true)
```
fmunshi, line 401: these props should be constants

app/post.go
```diff
@@ -393,6 +393,14 @@ func (a *App) FillInPostProps(post *model.Post, channel *model.Channel) *model.A
 		post.DelProp("channel_mentions")
 	}
 
+	if !a.HasPermissionToChannel(post.UserId, post.ChannelId, model.PERMISSION_USE_CHANNEL_MENTIONS) {
+		post.AddProp(model.POST_PROPS_MENTION_HIGHLIGHT_DISABLED, true)
+	}
```
fmunshi, line 398: you actually dont need this here the prop is added in the create post function via `post.DisableMentionHighlights()`

app/post.go
```diff
@@ -393,6 +393,14 @@ func (a *App) FillInPostProps(post *model.Post, channel *model.Channel) *model.A
...
+
+	if !a.HasPermissionToChannel(post.UserId, post.ChannelId, model.PERMISSION_USE_GROUP_MENTIONS) {
+		post.AddProp(model.POST_PROPS_GROUP_HIGHLIGHT_DISABLED, true)
+	}
```
fmunshi, line 402: can we make this check a bit smarter? maybe we can do a regex check for `\B@` and only if that exists we check the permission and set the prop?

app/post.go
hahmadia: True that friend! Thanks for pointing it out. I didn't pay close attention to notice that.

app/post.go
hahmadia: Done! + added tests for this add property portion of the code.

app/post.go
```diff
@@ -396,6 +397,11 @@ func (a *App) FillInPostProps(post *model.Post, channel *model.Channel) *model.A
 		post.DelProp("channel_mentions")
 	}
 
+	matched, err := regexp.MatchString(`\B@`, post.Message)
```
hmhealey, line 400: Won't this detect any mention and most regular usage of the at sign? Would it be worth it to explicitly check for group mentions

fmunshi: Yea it would result in this prop being set anytime there is a single `@` in the message body - do you think delaying post creation and querying `groups` by `name` for all matched mentions in the post body and then applying the prop only if it matches an existing ldap group?

hahmadia: ^^ That's what I was thinking. We could explicitly check for a group mention but it means more computation and possibly a delay in the call. I'm not sure if it's worth it ... But any thoughts on that Harrison?

hmhealey: Farhan and I chatted about this, and I'm now in agreement that this is the better way to do it. On one hand, it's adding extra data to every post containing a mention that's made by a user that doesn't have this permission, but at the same time, that's likely a lot less work than we'd require to verify the message actually contains a group mention.

model/post.go
```diff
@@ -62,7 +62,8 @@ const (
 	POST_PROPS_OVERRIDE_ICON_URL   = "override_icon_url"
 	POST_PROPS_OVERRIDE_ICON_EMOJI = "override_icon_emoji"
 
-	POST_PROPS_MENTION_HIGHLIGHT_DISABLED = "mentionHighlightDisabled"
+	POST_PROPS_MENTION_HIGHLIGHT_DISABLED = "disable_channel_highlight"
```
hmhealey, line 65: As per my comment on the other PR, this should be reverted to camel case since even though it's inconsistent, changing this breaks compatibility.

app/post.go
```diff
@@ -396,6 +397,11 @@ func (a *App) FillInPostProps(post *model.Post, channel *model.Channel) *model.A
...
+	if err == nil && matched && !a.HasPermissionToChannel(post.UserId, post.ChannelId, model.PERMISSION_USE_GROUP_MENTIONS) {
```
fmunshi, line 401: A license check here makes sense i think so that this group highlight disabled prop is only added when the server has ldap groups

model/post.go
hahmadia: Done! Thanks

app/post.go
hahmadia: Done! Thanks

app/post_test.go
```diff
@@ -1753,3 +1753,58 @@ func TestCountMentionsFromPost(t *testing.T) {
 		assert.Equal(t, numPosts, count)
 	})
 }
+
+func TestFillInPostProps(t *testing.T) {
```
fmunshi, line 1757: I would add a test case that ensures that the prop is not added if the license does not exist

app/post.go
agnivade: How frequently is this going to be called ? If it's not rare, then it might be a good idea to pre-compile the regex and store it somewhere for better performance.

hahmadia: @agnivade 
Thanks for the suggestion. I've implemented this suggestion! 😄

PR #14358, opened 2020-04-22
MM-23734 && MM-23733: Channel Sidebar Improvements Backend

reflog:
#### Summary

This PR introduces SidebarCategories and SidebarChannels tables as described in the spec https://docs.google.com/document/d/1thbPWPd9Wnz2GiMJq0Fm9mFzXmL1t66UHX6WNVNAAhc/edit

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-23733

Replaces https://github.com/mattermost/mattermost-server/pull/14252


reflog:
@agnivade and @jespino - I know you weren't involved with this PR, but it's a large one and I'd appreciate additional eyes on it. Thanks in advance

store/sqlstore/channel_store.go
```diff
@@ -416,6 +422,276 @@ func (s SqlChannelStore) createIndexesIfNotExists() {
...
+		getQueryBuilder().
+		Select("Preferences.UserId", "Preferences.Name AS ChannelId", "SidebarCategories.Id AS CategoryId").
+		From("Preferences").
+		Where(sq.And{
+			sq.Eq{"Preferences.Category": model.PREFERENCE_CATEGORY_FAVORITE_CHANNEL},
+			sq.NotEq{"SidebarCategories.Id": nil},
+			sq.Gt{"Preferences.UserId": lastUserId},
+		}).
+		LeftJoin("Channels ON (Channels.Id=Preferences.Name)").
+		LeftJoin("SidebarCategories ON (SidebarCategories.UserId=Preferences.UserId AND SidebarCategories.TeamId=Channels.TeamId AND SidebarCategories.Type='F')").
```
hmhealey, line 512: Does this migrate favourited DMs/GMs properly? They aren't assigned to a team, so they're going to need different handling than other channels

store/sqlstore/channel_store.go
```diff
@@ -3100,3 +3376,215 @@ func (s SqlChannelStore) GroupSyncedChannelCount() (int64, *model.AppError) {
...
+		return nil, model.NewAppError("SqlChannelStore.UpdateSidebarCategory", "store.sql_channel.sidebar_categories.commit_transaction.app_error", nil, err.Error(), http.StatusInternalServerError)
+	}
+
+	return &model.SidebarCategoryWithChannels{
+		SidebarCategory: *category,
+		Channels:        channelIDs,
+	}, nil
+}
+
+func (s SqlChannelStore) UpdateSidebarChannelByPreference(preference *model.Preference) *model.AppError {
```
hmhealey, line 3561: Maybe I'm missing it, but do we have logic to update the Favorites preferences based off of when the Favorites caetgory is updated? We need to make sure that changes that are done with the new sidebar are reflected on the mobile app where this isn't going to be used yet.

api4/channel.go
```diff
@@ -33,6 +33,13 @@ func (api *API) InitChannel() {
...
 	api.BaseRoutes.User.Handle("/teams/{team_id:[A-Za-z0-9]+}/channels", api.ApiSessionRequired(getChannelsForTeamForUser)).Methods("GET")
 
+	api.BaseRoutes.User.Handle("/teams/{team_id:[A-Za-z0-9]+}/channels/categories", api.ApiSessionRequired(getCategoriesForTeamForUser)).Methods("GET")
+	api.BaseRoutes.User.Handle("/teams/{team_id:[A-Za-z0-9]+}/channels/categories", api.ApiSessionRequired(createCategoryForTeamForUser)).Methods("POST")
+	api.BaseRoutes.User.Handle("/teams/{team_id:[A-Za-z0-9]+}/channels/categories/order", api.ApiSessionRequired(getCategoryOrderForTeamForUser)).Methods("GET")
+	api.BaseRoutes.User.Handle("/teams/{team_id:[A-Za-z0-9]+}/channels/categories/order", api.ApiSessionRequired(updateCategoryOrderForTeamForUser)).Methods("PUT")
+	api.BaseRoutes.User.Handle("/teams/{team_id:[A-Za-z0-9]+}/channels/categories/{category_id:[A-Za-z0-9]+}", api.ApiSessionRequired(getCategoryForTeamForUser)).Methods("GET")
+	api.BaseRoutes.User.Handle("/teams/{team_id:[A-Za-z0-9]+}/channels/categories/{category_id:[A-Za-z0-9]+}", api.ApiSessionRequired(updateCategoryForTeamForUser)).Methods("PUT")
+
```
grundleborg, line 42: 0/5 but you could create a common base for these `api.BaseRoutes.ChannelCategories` or similar.

api4/channel.go
```diff
@@ -1719,3 +1726,114 @@ func patchChannelModerations(c *Context, w http.ResponseWriter, r *http.Request)
 	auditRec.Success()
 	w.Write(b)
 }
+
+func getCategoriesForTeamForUser(c *Context, w http.ResponseWriter, r *http.Request) {
+	c.RequireUserId().RequireTeamId()
+	if c.Err != nil {
+		return
+	}
+	auditRec := c.MakeAuditRecord("getCategoriesForTeamForUser", audit.Fail)
```
grundleborg, line 1735: Do we need to add audit records to GET requests that don't mutate anything?

api4/channel.go
```diff
@@ -1719,3 +1726,114 @@ func patchChannelModerations(c *Context, w http.ResponseWriter, r *http.Request)
...
+	defer c.LogAuditRec(auditRec)
+
+	categories, err := c.App.GetSidebarCategories(c.Params.UserId, c.Params.TeamId)
```
grundleborg, line 1738: Permissions check?

api4/channel.go
```diff
@@ -1719,3 +1726,114 @@ func patchChannelModerations(c *Context, w http.ResponseWriter, r *http.Request)
...
+
+func createCategoryForTeamForUser(c *Context, w http.ResponseWriter, r *http.Request) {
+	c.RequireUserId().RequireTeamId()
+	if c.Err != nil {
+		return
+	}
+	auditRec := c.MakeAuditRecord("createCategoryForTeamForUser", audit.Fail)
+	defer c.LogAuditRec(auditRec)
+	categoryCreateRequest := model.SidebarCategoryFromJson(r.Body)
+	category, appErr := c.App.CreateSidebarCategory(c.Params.UserId, c.Params.TeamId, categoryCreateRequest.DisplayName, categoryCreateRequest.Channels)
```
grundleborg, line 1756: Permissions check?

api4/channel.go
```diff
@@ -1719,3 +1726,114 @@ func patchChannelModerations(c *Context, w http.ResponseWriter, r *http.Request)
...
+}
+
+func createCategoryForTeamForUser(c *Context, w http.ResponseWriter, r *http.Request) {
+	c.RequireUserId().RequireTeamId()
+	if c.Err != nil {
+		return
+	}
+	auditRec := c.MakeAuditRecord("createCategoryForTeamForUser", audit.Fail)
+	defer c.LogAuditRec(auditRec)
+	categoryCreateRequest := model.SidebarCategoryFromJson(r.Body)
```
grundleborg, line 1755: What happens if this fails?

api4/channel.go
```diff
@@ -1719,3 +1726,114 @@ func patchChannelModerations(c *Context, w http.ResponseWriter, r *http.Request)
...
+	w.Write(category.ToJson())
+}
+
+func getCategoryOrderForTeamForUser(c *Context, w http.ResponseWriter, r *http.Request) {
+	c.RequireUserId().RequireTeamId()
+	if c.Err != nil {
+		return
+	}
+	auditRec := c.MakeAuditRecord("getCategoriesForTeamForUser", audit.Fail)
+	defer c.LogAuditRec(auditRec)
```
grundleborg, line 1772: Do we need audit check here?

api4/channel.go
```diff
@@ -1719,3 +1726,114 @@ func patchChannelModerations(c *Context, w http.ResponseWriter, r *http.Request)
...
+
+	order, err := c.App.GetSidebarCategoryOrder(c.Params.UserId, c.Params.TeamId)
```
grundleborg, line 1774: Permissions check?

api4/channel.go
```diff
@@ -1719,3 +1726,114 @@ func patchChannelModerations(c *Context, w http.ResponseWriter, r *http.Request)
...
+func updateCategoryOrderForTeamForUser(c *Context, w http.ResponseWriter, r *http.Request) {
+	c.RequireUserId().RequireTeamId()
+	if c.Err != nil {
+		return
+	}
+	auditRec := c.MakeAuditRecord("updateCategoryOrderForTeamForUser", audit.Fail)
+	defer c.LogAuditRec(auditRec)
+
+	categoryOrder := model.ArrayFromJson(r.Body)
+	err := c.App.UpdateSidebarCategoryOrder(c.Params.UserId, c.Params.TeamId, categoryOrder)
```
grundleborg, line 1793: Permissions check?

api4/channel.go
```diff
@@ -1719,3 +1726,114 @@ func patchChannelModerations(c *Context, w http.ResponseWriter, r *http.Request)
...
+
+func updateCategoryOrderForTeamForUser(c *Context, w http.ResponseWriter, r *http.Request) {
+	c.RequireUserId().RequireTeamId()
+	if c.Err != nil {
+		return
+	}
+	auditRec := c.MakeAuditRecord("updateCategoryOrderForTeamForUser", audit.Fail)
+	defer c.LogAuditRec(auditRec)
+
+	categoryOrder := model.ArrayFromJson(r.Body)
```
grundleborg, line 1792: What happens if the data here is invalid or modifies another user's category?

api4/channel.go
```diff
@@ -1719,3 +1726,114 @@ func patchChannelModerations(c *Context, w http.ResponseWriter, r *http.Request)
...
+	auditRec.Success()
+	w.Write([]byte(model.ArrayToJson(categoryOrder)))
+}
+
+func getCategoryForTeamForUser(c *Context, w http.ResponseWriter, r *http.Request) {
+	c.RequireUserId().RequireTeamId().RequireCategoryId()
+	if c.Err != nil {
+		return
+	}
+	auditRec := c.MakeAuditRecord("getCategoriesForTeamForUser", audit.Fail)
```
grundleborg, line 1808: No audit check needed here I think.

api4/channel.go
```diff
@@ -1719,3 +1726,114 @@ func patchChannelModerations(c *Context, w http.ResponseWriter, r *http.Request)
...
+	defer c.LogAuditRec(auditRec)
+
+	categories, err := c.App.GetSidebarCategory(c.Params.UserId, c.Params.TeamId, c.Params.CategoryId)
```
grundleborg, line 1811: No permissions check.

api4/channel.go
```diff
@@ -1719,3 +1726,114 @@ func patchChannelModerations(c *Context, w http.ResponseWriter, r *http.Request)
...
+	c.RequireUserId().RequireTeamId().RequireCategoryId()
+	if c.Err != nil {
+		return
+	}
+	auditRec := c.MakeAuditRecord("updateCategoryForTeamForUser", audit.Fail)
+	defer c.LogAuditRec(auditRec)
+
+	categoryCreateRequest := model.SidebarCategoryFromJson(r.Body)
+
+	categories, appErr := c.App.UpdateSidebarCategory(c.Params.UserId, c.Params.TeamId, c.Params.CategoryId, categoryCreateRequest.DisplayName, categoryCreateRequest.Channels)
```
grundleborg, line 1831: Permissions check?

go.mod
```diff
@@ -44,7 +44,7 @@ require (
 	github.com/icrowley/fake v0.0.0-20180203215853-4178557ae428
 	github.com/jaytaylor/html2text v0.0.0-20190408195923-01ec452cbe43
 	github.com/jmoiron/sqlx v1.2.0
-	github.com/jonboulle/clockwork v0.1.0 // indirect
+	github.com/jonboulle/clockwork v0.1.0
```
grundleborg, line 47: No changes to dependencies so no need for any changes to this file.

store/sqlstore/upgrade.go
```diff
@@ -787,9 +788,7 @@ func upgradeDatabaseToVersion522(sqlStore SqlStore) {
 }
 
 func upgradeDatabaseToVersion523(sqlStore SqlStore) {
-	// TODO: uncomment when the time arrive to upgrade the DB for 5.23
-	// if shouldPerformUpgrade(sqlStore, VERSION_5_22_0, VERSION_5_23_0) {
-
+	//if shouldPerformUpgrade(sqlStore, VERSION_5_22_0, VERSION_5_23_0) {
```
grundleborg, line 791: I think these changes aren't needed here any more?

model/client4.go
```diff
@@ -5004,3 +5004,66 @@ func (c *Client4) PatchChannelModerations(channelID string, patch []*ChannelMode
...
+	r, err := c.DoApiGet(route, etag)
+	if err != nil {
+		return nil, BuildErrorResponse(r, err)
+	}
+	defer closeBody(r)
+	return OrderedSidebarCategoriesFromJson(r.Body), BuildResponse(r)
+}
+
+func (c *Client4) CreateSidebarCategoryForTeamForUser(userID, teamID string, category *SidebarCategoryWithChannels) (*SidebarCategoryWithChannels, *Response) {
+	payload, _ := json.Marshal(category)
```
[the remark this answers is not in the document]
reflog: Yeah, we don't check errors in marshaling most of the time... (see our ToJson methods in model)

api4/channel.go
```diff
@@ -1719,3 +1726,114 @@ func patchChannelModerations(c *Context, w http.ResponseWriter, r *http.Request)
...
+	if err != nil {
+		c.Err = err
+		return
+	}
+
+	auditRec.Success()
+	w.Write(categories.ToJson())
+}
+
+func updateCategoryForTeamForUser(c *Context, w http.ResponseWriter, r *http.Request) {
```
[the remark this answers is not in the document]
reflog: @hmhealey "Should this endpoint support patch instead of update operations" <- not sure what you mean. What's the difference between update and patch in terms of "changing an entry in the db" ?

I'll add support for multiple updates, will change the structure to accept array of Categories

store/sqlstore/channel_store.go
```diff
@@ -3100,3 +3377,242 @@ func (s SqlChannelStore) GroupSyncedChannelCount() (int64, *model.AppError) {
...
+		}
+		prevCategory.Channels = append(prevCategory.Channels, category.ChannelId)
+	}
+	return &oc, nil
+}
+
+func (s SqlChannelStore) GetSidebarCategoryOrder(userId, teamId string) ([]string, *model.AppError) {
+	var ids []string
+	if _, err := s.GetReplica().Select(&ids, "SELECT Id FROM SidebarCategories WHERE UserId=:userId AND TeamId=:teamId ORDER BY SortOrder ASC", map[string]interface{}{"userId": userId, "teamId": teamId}); err != nil {
+		return nil, model.NewAppError("SqlPostStore.GetSidebarCategoryOrder", "store.sql_channel.sidebar_categories.app_error", nil, err.Error(), http.StatusInternalServerError)
```
agnivade, line 3481: Do we want to check for a not found error and return a 404 ? A bad userid or teamId should not return a 500.

store/sqlstore/channel_store.go
```diff
@@ -3100,3 +3377,242 @@ func (s SqlChannelStore) GroupSyncedChannelCount() (int64, *model.AppError) {
...
+				sq.Eq{"ChannelId": oldCategory.Channels},
+				sq.Eq{"Preferences.Category": model.PREFERENCE_CATEGORY_FAVORITE_CHANNEL},
+			}).ToSql()
+
+			if _, err = transaction.Exec(sql, args...); err != nil {
+				return nil, model.NewAppError("SqlPostStore.UpdateSidebarCategory", "store.sql_channel.sidebar_categories.app_error", nil, err.Error(), http.StatusInternalServerError)
+			}
+		}
+		var channels []interface{}
+		var preferences []interface{}
```
agnivade, line 3551: Same as earlier.

store/sqlstore/channel_store.go
```diff
@@ -3100,3 +3377,242 @@ func (s SqlChannelStore) GroupSyncedChannelCount() (int64, *model.AppError) {
...
+
+	return &model.SidebarCategoryWithChannels{
+		SidebarCategory: *category,
+		Channels:        newCategory.Channels,
+	}, nil
+}
+
+func (s SqlChannelStore) GetSidebarCategory(userId, teamId, categoryId string) (*model.SidebarCategoryWithChannels, *model.AppError) {
+	var categories []*sidebarCategoryForJoin
+	if _, err := s.GetReplica().Select(&categories, "SELECT *, SidebarChannels.ChannelId FROM SidebarCategories, SidebarChannels WHERE SidebarCategories.UserId=:userId AND SidebarCategories.TeamId=:teamId AND SidebarChannels.CategoryId=SidebarCategories.Id AND SidebarCategories.Id=:categoryId ORDER BY SidebarChannels.SortOrder ASC", map[string]interface{}{"userId": userId, "teamId": teamId, "categoryId": categoryId}); err != nil {
```
agnivade, line 3433: Would it be possible to use squirrel for this and the next SQL query ? Might be cleaner if at least the read queries use squirrel.

api4/channel.go
```diff
@@ -1719,3 +1727,214 @@ func patchChannelModerations(c *Context, w http.ResponseWriter, r *http.Request)
...
+	if !c.App.SessionHasPermissionToUser(*c.App.Session(), c.Params.UserId) {
+		c.SetPermissionError(model.PERMISSION_EDIT_OTHER_USERS)
+		return
+	}
+
+	auditRec := c.MakeAuditRecord("updateCategoryForTeamForUser", audit.Fail)
+	defer c.LogAuditRec(auditRec)
+
+	categoryUpdateRequest, err := model.SidebarCategoryFromJson(r.Body)
+
```
agnivade, line 1921: Nit: remove gap. Error checks are usually immediately after their declaration.

store/sqlstore/channel_store.go
```diff
@@ -416,6 +422,277 @@ func (s SqlChannelStore) createIndexesIfNotExists() {
...
+	}
+
+	if len(userTeamMap) == 0 {
+		// No more team members in query result means that the migration has finished.
+		return nil, nil
+	}
+
+	for _, u := range userTeamMap {
+		if err := transaction.Insert(&model.SidebarCategory{
+			DisplayName: T("sidebar.category.favorites"),
```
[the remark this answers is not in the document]
reflog: @agnivade - how did you get the crash? Just started the server, or..?

store/sqlstore/channel_store.go
```diff
@@ -3100,3 +3377,242 @@ func (s SqlChannelStore) GroupSyncedChannelCount() (int64, *model.AppError) {
...
+		Id:          categoryId,
+		UserId:      userId,
+		TeamId:      teamId,
+		SortOrder:   0,
+		Type:        model.SidebarCategoryCustom,
+	}
+	if err = transaction.Insert(category); err != nil {
+		return nil, model.NewAppError("SqlPostStore.CreateSidebarCategory", "store.sql_channel.sidebar_categories.app_error", nil, err.Error(), http.StatusInternalServerError)
+	}
+	var channels []interface{}
```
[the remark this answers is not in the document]
reflog: Yep. Cannot convert []*Something to []interface{}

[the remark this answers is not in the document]
agnivade: My bad. Resolved offline.

store/sqlstore/channel_store.go
```diff
@@ -416,6 +422,277 @@ func (s SqlChannelStore) createIndexesIfNotExists() {
...
+	}
+
+	transaction, err := s.GetMaster().Begin()
+	if err != nil {
+		return nil, model.NewAppError("SqlChannelStore.MigrateSidebarCategories", "store.sql_channel.MigrateSidebarCategories.open_transaction.app_error", nil, err.Error(), http.StatusInternalServerError)
+	}
+
+	defer finalizeTransaction(transaction)
+
+	if _, err := transaction.Select(&userTeamMap, "SELECT TeamId, UserId FROM TeamMembers WHERE (TeamId, UserId) > (:FromTeamId, :FromUserId) ORDER BY TeamId, UserId LIMIT 100", map[string]interface{}{"FromTeamId": fromTeamId, "FromUserId": fromUserId}); err != nil {
```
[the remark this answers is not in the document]
reflog: as we've discussed offline, LIMIT is applied for items that come after the previous userId, teamId and will continue to run until no source records are left

store/sqlstore/channel_store.go
```diff
@@ -3100,3 +3352,272 @@ func (s SqlChannelStore) GroupSyncedChannelCount() (int64, *model.AppError) {
...
+}
+
+func (s SqlChannelStore) GetSidebarCategory(userId, teamId, categoryId string) (*model.SidebarCategoryWithChannels, *model.AppError) {
+	var categories []*sidebarCategoryForJoin
+	sql, args, _ := s.getQueryBuilder().
+		Select("*", "SidebarChannels.ChannelId").
+		From("SidebarCategories, SidebarChannels").
+		Where(sq.And{
+			sq.Eq{"SidebarCategories.UserId": userId},
+			sq.Eq{"SidebarCategories.TeamId": teamId},
```
grundleborg, line 3413: I'm a bit confused by this - can a sidebar category be shared by multiple users, or by multiple teams of the same user? It looks like not from the model structs, but maybe I misunderstood the schema? I guess I just don't really understand what's going on with all these IDs on the category object.

api4/channel.go
```diff
@@ -1719,3 +1727,214 @@ func patchChannelModerations(c *Context, w http.ResponseWriter, r *http.Request)
...
+	if c.Err != nil {
+		return
+	}
+
+	if !c.App.SessionHasPermissionToUser(*c.App.Session(), c.Params.UserId) {
+		c.SetPermissionError(model.PERMISSION_EDIT_OTHER_USERS)
+		return
+	}
+
+	categories, err := c.App.GetSidebarCategory(c.Params.UserId, c.Params.TeamId, c.Params.CategoryId)
```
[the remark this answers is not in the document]
grundleborg: Looks like a no, because in the SQL query we check all of UserId TeamId and CategoryId match the database record, but is that actually an intentional defence against this risk, or completely unrelated?

store/sqlstore/channel_store.go
```diff
@@ -3100,3 +3352,272 @@ func (s SqlChannelStore) GroupSyncedChannelCount() (int64, *model.AppError) {
...
+
+	defer finalizeTransaction(transaction)
+
+	categoryId := model.NewId()
+	category := &model.SidebarCategory{
+		DisplayName: newCategory.DisplayName,
+		Id:          categoryId,
+		UserId:      userId,
+		TeamId:      teamId,
+		SortOrder:   0,
```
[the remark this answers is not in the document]
reflog: Nope. A bug. Will fix

store/sqlstore/channel_store.go
```diff
@@ -3100,3 +3352,272 @@ func (s SqlChannelStore) GroupSyncedChannelCount() (int64, *model.AppError) {
...
+	}
+
+	defer finalizeTransaction(transaction)
+
+	categoryId := model.NewId()
+	category := &model.SidebarCategory{
+		DisplayName: newCategory.DisplayName,
+		Id:          categoryId,
+		UserId:      userId,
+		TeamId:      teamId,
```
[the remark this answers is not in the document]
reflog: added validation in the api

store/sqlstore/channel_store.go
```diff
@@ -3100,3 +3352,272 @@ func (s SqlChannelStore) GroupSyncedChannelCount() (int64, *model.AppError) {
...
+		return nil, model.NewAppError("SqlPostStore.CreateSidebarCategory", "store.sql_channel.sidebar_categories.app_error", nil, err.Error(), http.StatusInternalServerError)
+	}
+	var channels []interface{}
+	runningOrder := 0
+	for _, channelID := range newCategory.Channels {
+		channels = append(channels, &model.SidebarChannel{
+			ChannelId:  channelID,
+			CategoryId: categoryId,
+			SortOrder:  int64(runningOrder),
+			UserId:     userId,
```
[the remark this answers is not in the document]
reflog: yup. cleanup was needed. adding

store/sqlstore/channel_store.go
reflog: will fix

api4/channel.go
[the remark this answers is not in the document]
hmhealey: I wasn't sure if we wanted to support patch requests for categories like we have with some other endpoints since I think you might've mentioned something about them at some point, but I can't remember if that was actually the case. Regardless, we can work with either of them from our end.

Thanks for adding the batched endpoint!

store/sqlstore/channel_store.go
```diff
@@ -3100,3 +3352,272 @@ func (s SqlChannelStore) GroupSyncedChannelCount() (int64, *model.AppError) {
...
+		Type:        model.SidebarCategoryCustom,
+	}
+	if err = transaction.Insert(category); err != nil {
+		return nil, model.NewAppError("SqlPostStore.CreateSidebarCategory", "store.sql_channel.sidebar_categories.app_error", nil, err.Error(), http.StatusInternalServerError)
+	}
+	var channels []interface{}
+	runningOrder := 0
+	for _, channelID := range newCategory.Channels {
+		channels = append(channels, &model.SidebarChannel{
+			ChannelId:  channelID,
```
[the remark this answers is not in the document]
hmhealey: The move channel should presumably take care of that. We've been running into quite a few issues lately (around things like unreads and which channels are visible) where the server and app get out of sync with how they expect data to look, so I'd prefer for the server to provide more definitive data for things like that.

store/sqlstore/channel_store.go
```diff
@@ -3132,3 +3401,287 @@ func (s SqlChannelStore) GroupSyncedChannelCount() (int64, *model.AppError) {
...
+				sq.Eq{"ChannelId": oldCategory.Channels},
+				sq.Eq{"Preferences.Category": model.PREFERENCE_CATEGORY_FAVORITE_CHANNEL},
+			}).ToSql()
+
+			if _, err = transaction.Exec(sql, args...); err != nil {
+				return nil, model.NewAppError("SqlPostStore.UpdateSidebarCategory", "store.sql_channel.sidebar_categories.app_error", nil, err.Error(), http.StatusInternalServerError)
+			}
+		}
+		var channels []interface{}
+		var preferences []interface{}
```
hmhealey, line 3615: Shouldn't these be given a concrete type?

store/sqlstore/channel_store.go
```diff
@@ -3132,3 +3401,287 @@ func (s SqlChannelStore) GroupSyncedChannelCount() (int64, *model.AppError) {
...
+
+func (s SqlChannelStore) UpdateSidebarCategories(userId, teamId string, categories []*model.SidebarCategoryWithChannels) ([]*model.SidebarCategoryWithChannels, *model.AppError) {
+
+	transaction, err := s.GetMaster().Begin()
+	if err != nil {
+		return nil, model.NewAppError("SqlChannelStore.UpdateSidebarCategory", "store.sql_channel.sidebar_categories.open_transaction.app_error", nil, err.Error(), http.StatusInternalServerError)
+	}
+
+	defer finalizeTransaction(transaction)
+	for _, oldCategory := range categories {
```
hmhealey, line 3583: Are these names correct? `categories` would've come from the API, so why would we get `oldCategory` from it?

store/sqlstore/channel_store.go
```diff
@@ -3132,3 +3401,287 @@ func (s SqlChannelStore) GroupSyncedChannelCount() (int64, *model.AppError) {
...
+		sql, args, _ := s.getQueryBuilder().Delete("SidebarChannels").Where(sq.Eq{"ChannelId": oldCategory.Channels}).ToSql()
+
+		if _, err = transaction.Exec(sql, args...); err != nil {
+			return nil, model.NewAppError("SqlPostStore.UpdateSidebarCategory", "store.sql_channel.sidebar_categories.app_error", nil, err.Error(), http.StatusInternalServerError)
+		}
+
+		if category.Type == model.SidebarCategoryFavorites {
+			// clean previous preference favorites (to support bi-directional sync)
+			sql, args, _ = s.getQueryBuilder().Delete("Preferences").Where(sq.And{
+				sq.Eq{"ChannelId": oldCategory.Channels},
```
hmhealey, line 3606: Related to above, shouldn't we be deleting the preferences corresponding to the channel IDs in the original category? This looks like it's deleting the preferences for the new channel IDs since it's using `oldCategory`

store/sqlstore/channel_store.go
```diff
@@ -3132,3 +3401,287 @@ func (s SqlChannelStore) GroupSyncedChannelCount() (int64, *model.AppError) {
...
+			}
+		}
+		var channels []interface{}
+		var preferences []interface{}
+		runningOrder := 0
+		for _, channelID := range oldCategory.Channels {
+			channels = append(channels, &model.SidebarChannel{
+				ChannelId:  channelID,
+				CategoryId: category.Id,
+				SortOrder:  int64(runningOrder),
```
hmhealey, line 3621: If we're just recreating the sorting keys each time we run this, I suppose the sorting distance isn't working out as I'd hoped since I hoped we'd be able to only update the rows that were actually changing. There's nothing wrong with keeping the extra sorting distance, but I guess this isn't working out like I had originally hoped where that would make future ordering changes easier.

store/sqlstore/channel_store.go
```diff
@@ -3132,3 +3401,287 @@ func (s SqlChannelStore) GroupSyncedChannelCount() (int64, *model.AppError) {
...
+}
+
+// UpdateSidebarChannelByPreference is called when the Preference table is being updated to keep SidebarCategories in sync
+func (s SqlChannelStore) UpdateSidebarChannelByPreference(preference *model.Preference) *model.AppError {
+	if preference.Category != model.PREFERENCE_CATEGORY_DIRECT_CHANNEL_SHOW && preference.Category != model.PREFERENCE_CATEGORY_GROUP_CHANNEL_SHOW && preference.Category != model.PREFERENCE_CATEGORY_FAVORITE_CHANNEL {
+		return nil
+	}
+	params := map[string]interface{}{
+		"UserId":       preference.UserId,
+		"ChannelId":    preference.Name,
```
hmhealey, line 3659: For DMs, the preference name isn't the channel ID of the DM unfortunately, it's the ID of the teammate that the DM is with.

store/sqlstore/channel_store.go
```diff
@@ -3132,3 +3401,287 @@ func (s SqlChannelStore) GroupSyncedChannelCount() (int64, *model.AppError) {
...
+		"CategoryType": "",
+	}
+
+	switch preference.Category {
+	case model.PREFERENCE_CATEGORY_FAVORITE_CHANNEL:
+		params["CategoryType"] = model.SidebarCategoryFavorites
+	case model.PREFERENCE_CATEGORY_DIRECT_CHANNEL_SHOW:
+	case model.PREFERENCE_CATEGORY_GROUP_CHANNEL_SHOW:
+		params["CategoryType"] = model.SidebarCategoryDirectMessages
```
hmhealey, line 3668: Was this here before? I don't know if having the DM/GM preferences update the SidebarChannel entries for the DMs category works since there's unfortunately ways for a DM/GM to become visible without updating this preference.

I think we might want to remove this part for now and just focus on keeping the favourites in sync between preferences and the sidebar for now since there's still more that goes into the visibility of DMs and GMs.

store/sqlstore/channel_store.go
```diff
@@ -3132,3 +3401,287 @@ func (s SqlChannelStore) GroupSyncedChannelCount() (int64, *model.AppError) {
...
+
+	if err = transaction.Commit(); err != nil {
+		return nil, model.NewAppError("SqlChannelStore.UpdateSidebarCategory", "store.sql_channel.sidebar_categories.commit_transaction.app_error", nil, err.Error(), http.StatusInternalServerError)
+	}
+
+	return categories, nil
+}
+
+// UpdateSidebarChannelByPreference is called when the Preference table is being updated to keep SidebarCategories in sync
+func (s SqlChannelStore) UpdateSidebarChannelByPreference(preference *model.Preference) *model.AppError {
```
hmhealey, line 3653: Since all of the linking between preferences and SidebarChannels is being done in the store layer, are we going to be able to emit the proper websocket events when one causes the other to change?

store/sqlstore/channel_store.go
agnivade: Heh .. no. You can't assign `[]*type` to `[]interface{}`. I fell in the same trap too.

store/sqlstore/channel_store.go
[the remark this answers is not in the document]
reflog: this will be handled by UpdateSidebarChannelCategoryOnMove method in channel store

store/sqlstore/channel_store.go
reflog: categoryIdToUpdate will be better :)

store/sqlstore/channel_store.go
reflog: there's no "original" category. Category ID is the same.
When update comes in, we use the Id inside the request to delete old channels and add the new channels that are specified in the request

store/sqlstore/channel_store.go
reflog: wat? :)
so there is not Channel with the Id of preference.Name for DMs? What about GMs?

store/sqlstore/channel_store.go
reflog: you are right. i'll reroute it through App layer and add the events

store/sqlstore/channel_store.go
reflog: @hmhealey  I am not sure I understand. Why should we remove this? What ways are there for DMs to become visible without this part?
If we remove this - how do you propose we keep them in sync?

store/sqlstore/channel_store.go
reflog: @hmhealey  There seems to be a misunderstanding here. The API proposed in the document receives the new full list of channels. I am assuming - it's ordered by the user. So we insert them in the database and re-index their SortOrder. 
What seems to be the issue?

api4/channel.go
```diff
@@ -1753,3 +1761,214 @@ func patchChannelModerations(c *Context, w http.ResponseWriter, r *http.Request)
...
+		return
+	}
+
+	auditRec := c.MakeAuditRecord("createCategoryForTeamForUser", audit.Fail)
+	defer c.LogAuditRec(auditRec)
+	categoryCreateRequest, err := model.SidebarCategoryFromJson(r.Body)
+	if err != nil {
+		c.SetInvalidParam("category")
+		return
+	}
```
grundleborg, line 1802: I think we need to add a check here to make sure the `c.Params.UserId == categoryCreateRequest.UserId` and `c.Params.TeamId == categoryCreateRequest.TeamId` or else we have possible security vulnerabilities through injecting one value in the request params to pass the permissions check and then using a totally different value to get saved into the database as part of the category.

api4/channel.go
```diff
@@ -1753,3 +1761,214 @@ func patchChannelModerations(c *Context, w http.ResponseWriter, r *http.Request)
...
+	w.Write([]byte(model.ArrayToJson(order)))
+}
+
+func updateCategoryOrderForTeamForUser(c *Context, w http.ResponseWriter, r *http.Request) {
+	c.RequireUserId().RequireTeamId()
+	if c.Err != nil {
+		return
+	}
+
+	if !c.App.SessionHasPermissionToUser(*c.App.Session(), c.Params.UserId) || !c.App.SessionHasPermissionToCategory(*c.App.Session(), c.Params.TeamId, c.Params.CategoryId) {
```
grundleborg, line 1843: The second part of this permissions check looks odd. Do we have a `CategoryId` in this request?

api4/channel.go
grundleborg: OK, so this handler just modifies the *order*. The team and user come through the params now, an no category ID is provided. So that means this comment is no longer relevant.

api4/channel.go
```diff
@@ -1753,3 +1761,214 @@ func patchChannelModerations(c *Context, w http.ResponseWriter, r *http.Request)
...
+	w.Write([]byte(model.ArrayToJson(categoryOrder)))
+}
+
+func getCategoryForTeamForUser(c *Context, w http.ResponseWriter, r *http.Request) {
+	c.RequireUserId().RequireTeamId().RequireCategoryId()
+	if c.Err != nil {
+		return
+	}
+
+	if !c.App.SessionHasPermissionToUser(*c.App.Session(), c.Params.UserId) || !c.App.SessionHasPermissionToCategory(*c.App.Session(), c.Params.TeamId, c.Params.CategoryId) {
```
grundleborg, line 1868: Do we need to check both these separately? `SessionHasPermissionToCategory` seems to cover both bases as far as I can see.

api4/channel.go
```diff
@@ -1753,3 +1761,214 @@ func patchChannelModerations(c *Context, w http.ResponseWriter, r *http.Request)
...
+		c.SetPermissionError(model.PERMISSION_EDIT_OTHER_USERS)
+		return
+	}
+
+	categories, err := c.App.GetSidebarCategory(c.Params.UserId, c.Params.TeamId, c.Params.CategoryId)
```
grundleborg, line 1873: So, not important, but just for my peace of mind, is there a reason why `GetSidebarCategory` can't just receive a single param of `c.Params.CategoryId` now?

api4/channel.go
```diff
@@ -1753,3 +1761,214 @@ func patchChannelModerations(c *Context, w http.ResponseWriter, r *http.Request)
...
+	w.Write(categories.ToJson())
+}
+
+func updateCategoriesForTeamForUser(c *Context, w http.ResponseWriter, r *http.Request) {
+	c.RequireUserId().RequireTeamId().RequireCategoryId()
+	if c.Err != nil {
+		return
+	}
+
+	if !c.App.SessionHasPermissionToUser(*c.App.Session(), c.Params.UserId) || !c.App.SessionHasPermissionToCategory(*c.App.Session(), c.Params.TeamId, c.Params.CategoryId) {
```
grundleborg, line 1888: As above, not sure we need both of these.

api4/channel.go
```diff
@@ -1753,3 +1761,214 @@ func patchChannelModerations(c *Context, w http.ResponseWriter, r *http.Request)
...
+	if err != nil {
+		c.Err = err
+		return
+	}
+
+	w.Write(categories.ToJson())
+}
+
+func updateCategoriesForTeamForUser(c *Context, w http.ResponseWriter, r *http.Request) {
+	c.RequireUserId().RequireTeamId().RequireCategoryId()
```
grundleborg, line 1883: Looks like this now handles multiple categories so shouldn't `RequireCategoryId()` any more?

api4/channel.go
```diff
@@ -1753,3 +1761,214 @@ func patchChannelModerations(c *Context, w http.ResponseWriter, r *http.Request)
...
+	}
+
+	auditRec := c.MakeAuditRecord("updateCategoriesForTeamForUser", audit.Fail)
+	defer c.LogAuditRec(auditRec)
+
+	categoriesUpdateRequest, err := model.SidebarCategoriesFromJson(r.Body)
+	if err != nil {
+		c.SetInvalidParam("category")
+		return
+	}
```
grundleborg, line 1900: As above, need to check equality of the TeamId and UserId between the URL params and the body JSON.

api4/channel.go
```diff
@@ -1753,3 +1761,214 @@ func patchChannelModerations(c *Context, w http.ResponseWriter, r *http.Request)
...
+	return nil
+}
+
+func updateCategoryForTeamForUser(c *Context, w http.ResponseWriter, r *http.Request) {
+	c.RequireUserId().RequireTeamId().RequireCategoryId()
+	if c.Err != nil {
+		return
+	}
+
+	if !c.App.SessionHasPermissionToUser(*c.App.Session(), c.Params.UserId) || !c.App.SessionHasPermissionToCategory(*c.App.Session(), c.Params.TeamId, c.Params.CategoryId) {
```
grundleborg, line 1946: As above, only the second one needed I think?

api4/channel.go
```diff
@@ -1753,3 +1761,214 @@ func patchChannelModerations(c *Context, w http.ResponseWriter, r *http.Request)
...
+	}
+
+	auditRec := c.MakeAuditRecord("updateCategoryForTeamForUser", audit.Fail)
+	defer c.LogAuditRec(auditRec)
+
+	categoryUpdateRequest, err := model.SidebarCategoryFromJson(r.Body)
+	if err != nil {
+		c.SetInvalidParam("category")
+		return
+	}
```
grundleborg, line 1958: As above, need to check userId, teamId and categoryId are equal in body payload and in URL parameters.

api4/channel.go
reflog: I think we need both, SessionHasPermissionToUser validates users/session.UserId and the category one validates proper category request

api4/channel.go
reflog: nope. no reason :) will change

api4/channel.go
grundleborg: But all that does is:

```
func (a *App) HasPermissionToUser(askingUserId string, userId string) bool {
	if askingUserId == userId {
		return true
	}

	if a.HasPermissionTo(askingUserId, model.PERMISSION_EDIT_OTHER_USERS) {
		return true
	}

	return false
}
```

Which seems to be covered by the other check anyway?

store/sqlstore/channel_store.go
hmhealey: The preference's name is the channel ID for GMs, but it's the ID of the other user for DMs. If you have the ID of the current user, you can get the other user's ID  from the channel using [`Channel.GetOtherUserIdForDM`](https://github.com/mattermost/mattermost-server/blob/master/model/channel.go#L310).

I swear making it the ID of the other user made sense back when I added Preferences for showing/hiding DMs, but it definitely doesn't make sense looking back on it, especially since GMs were added

store/sqlstore/channel_store.go
hmhealey: I was imagining some sort of transformation to avoid modifying extra rows in the database, although that's obviously a lot more complicated.

With the way you're doing it, I don't think separating the indices by 10 is actually necessary, but I don't see a reason to change that now since it doesn't hurt to do it.

store/sqlstore/channel_store.go
hmhealey: DMs can unfortunately become visible on the client when you receive a message in one, but the Preference isn't sent back to the server until the user actually views the channel. This is currently all controlled on the client-side I forgot about that when writing the spec for this, and I meant to mention it to you, but I forgot.

I think doing it the way you're doing it makes sense in the long run, but we might need to compensate for this somewhere (either on the client by adding to the received categories in this case or on the server by updating the preferences when a message is received in a DM), so I'll file a followup to just ensure that that behaviour works before this is shipped.

store/sqlstore/channel_store.go
agnivade: Hey @reflog .. looks like we were wrong. The `transaction.Insert` call takes `interface{}` so it should be possible to use types for these. When you DM'd me, you said we cannot pass to `[]interface{}` and I didn't take a look at the code again. But actually it takes `interface{}`, not `[]interface{}`. Is there a reason this won't work ?

store/sqlstore/channel_store.go
reflog: ok, discussed this with @hmhealey offline, for now this function will only handle favorites

store/sqlstore/channel_store.go
reflog: ok, discussed this with @hmhealey offline, for now this function will only handle favorites

store/sqlstore/channel_store.go
```diff
@@ -416,6 +422,268 @@ func (s SqlChannelStore) createIndexesIfNotExists() {
...
+		}); err != nil && !IsUniqueConstraintError(err, []string{"UserId"}) {
+			return nil, model.NewAppError("SqlChannelStore.MigrateSidebarCategories", "store.sql_channel.MigrateSidebarCategories.app_error", nil, err.Error(), http.StatusInternalServerError)
+		}
+
+		if err := transaction.Insert(&model.SidebarCategory{
+			DisplayName: T("sidebar.category.dm"),
+			Id:          model.NewId(),
+			UserId:      u.UserId,
+			TeamId:      u.TeamId,
+			Sorting:     model.SidebarCategorySortAlphabetical,
```
hmhealey, line 485: Sorry, late change since I forgot about this earlier, but can you make the DMs category default to recency sorting instead?
```suggestion
			Sorting:     model.SidebarCategorySortRecent,
```

reflog: No problem. Will change

api4/channel.go
```diff
@@ -1754,3 +1762,227 @@ func patchChannelModerations(c *Context, w http.ResponseWriter, r *http.Request)
...
+		c.SetPermissionError(model.PERMISSION_EDIT_OTHER_USERS)
+		return
+	}
+
+	auditRec := c.MakeAuditRecord("updateCategoryOrderForTeamForUser", audit.Fail)
+	defer c.LogAuditRec(auditRec)
+
+	categoryOrder := model.ArrayFromJson(r.Body)
+
+	for _, categoryId := range categoryOrder {
```
grundleborg, line 1855: As discussed out of band, can you fetch the category here and check `c.Params.UserId == category.UserId && c.Params.TeamId == category.TeamId` just to be explicit.

api4/channel.go
```diff
@@ -1754,3 +1762,227 @@ func patchChannelModerations(c *Context, w http.ResponseWriter, r *http.Request)
...
+	defer c.LogAuditRec(auditRec)
+
+	categoriesUpdateRequest, err := model.SidebarCategoriesFromJson(r.Body)
+	if err != nil {
+		c.SetInvalidParam("category")
+		return
+	}
+	var channelsToCheck []string
+	for _, category := range categoriesUpdateRequest {
+		if !c.App.SessionHasPermissionToCategory(*c.App.Session(), c.Params.TeamId, category.Id) {
```
grundleborg, line 1913: Can you also add the check `category.UserId == c.Params.UserId && category.TeamId == c.Params.TeamId` here to be explicit?

api4/channel.go
```diff
@@ -1754,3 +1762,227 @@ func patchChannelModerations(c *Context, w http.ResponseWriter, r *http.Request)
...
+			c.SetInvalidParam("category")
+			return
+		}
+		channelsToCheck = append(channelsToCheck, category.Channels...)
+	}
+	if appErr := validateUserChannels("updateCategoriesForTeamForUser", c, c.Params.TeamId, c.Params.UserId, channelsToCheck); appErr != nil {
```
grundleborg, line 1919: Should we enforce that certain fields on the category cannot be updated?

app/authorization.go
```diff
@@ -80,6 +80,15 @@ func (a *App) SessionHasPermissionToChannelByPost(session model.Session, postId
 	return a.SessionHasPermissionTo(session, permission)
 }
 
+func (a *App) SessionHasPermissionToCategory(session model.Session, teamId, categoryId string) bool {
```
grundleborg, line 83:
```suggestion
func (a *App) SessionHasPermissionToCategory(session model.Session, categoryId string) bool {
```

api4/channel.go
reflog: @grundleborg  we do, inside of the Store. Only Channels, DisplayName and Sorting fields are updated

grundleborg: Perfect.

model/client4.go
```diff
@@ -5005,6 +5005,87 @@ func (c *Client4) PatchChannelModerations(channelID string, patch []*ChannelMode
 	return ChannelModerationsFromJson(r.Body), BuildResponse(r)
 }
 
+func (c *Client4) GetSidebarCategoriesForTeamForUser(userID, teamID, etag string) (*OrderedSidebarCategories, *Response) {
+	route := fmt.Sprintf(c.GetUserRoute(userID) + c.GetTeamRoute(teamID) + "/channels/categories")
```
isacikgoz, line 5009: Maybe we can use a similar approaches for other endpoints here and create a method like `c.GetCategoryRoute`.

model/client4.go
```diff
@@ -5005,6 +5005,87 @@ func (c *Client4) PatchChannelModerations(channelID string, patch []*ChannelMode
...
+	payload, _ := json.Marshal(category)
+	route := fmt.Sprintf(c.GetUserRoute(userID) + c.GetTeamRoute(teamID) + "/channels/categories")
+	r, appErr := c.doApiPostBytes(route, payload)
+	if appErr != nil {
+		return nil, BuildErrorResponse(r, appErr)
+	}
+	defer closeBody(r)
+	cat, err := SidebarCategoryFromJson(r.Body)
+	if err != nil {
+		return nil, &Response{StatusCode: http.StatusBadRequest, Error: NewAppError(c.GetUserRoute(userID), "model.client.connecting.app_error", nil, err.Error(), http.StatusForbidden)}
```
isacikgoz, line 5032: We can also wrap `CreateSidebarCategoryForTeamForUser`, `GetSidebarCategoryForTeamForUser` and `UpdateSidebarCategoryForTeamForUser` methods with a new method to avoid code repetition here :)

```suggestion
	if err != nil {
		return nil,BuildErrorResponse(r, NewAppError("Client4.CreateSidebarCategoryForTeamForUser", "model.utils.decode_json.app_error", nil, err.Error(), r.StatusCode))
```

store/sqlstore/channel_store.go
```diff
@@ -416,6 +422,268 @@ func (s SqlChannelStore) createIndexesIfNotExists() {
...
+		LeftJoin("ChannelMembers ON ChannelMembers.ChannelId=Channels.Id").
+		LeftJoin("SidebarCategories ON (SidebarCategories.UserId=ChannelMembers.UserId AND SidebarCategories.Type='C' AND SidebarCategories.TeamId=Channels.TeamId)").
+		OrderBy("Channels.Id", "ChannelMembers.UserId", "Channels.Name DESC").
+		Limit(100)
+
+	transaction, err := s.GetMaster().Begin()
+	if err != nil {
+		return nil, model.NewAppError("SqlChannelStore.MigrateChannelsToSidebarChannels", "store.sql_channel.migrate_channels_sidebar.open_transaction.app_error", nil, err.Error(), http.StatusInternalServerError)
+	}
+	defer finalizeTransaction(transaction)
```
isacikgoz, line 602: Could we use this in the beginning of this method? And same with others.

store/sqlstore/channel_store.go
```diff
@@ -3132,3 +3415,298 @@ func (s SqlChannelStore) GroupSyncedChannelCount() (int64, *model.AppError) {
...
+		if _, err = transaction.Exec(sql, args...); err != nil {
+			return nil, model.NewAppError("SqlPostStore.UpdateSidebarCategory", "store.sql_channel.sidebar_categories.app_error", nil, err.Error(), http.StatusInternalServerError)
+		}
+
+		if category.Type == model.SidebarCategoryFavorites {
+			// clean previous preference favorites (to support bi-directional sync)
+			sql, args, _ = s.getQueryBuilder().Delete("Preferences").Where(sq.And{
+				sq.Eq{"ChannelId": categoryToUpdate.Channels},
+				sq.Eq{"Preferences.Category": model.PREFERENCE_CATEGORY_FAVORITE_CHANNEL},
+			}).ToSql()
```
isacikgoz, line 3625: Could we use something like this?
```suggestion
			sql, args, _ = s.getQueryBuilder().Delete("Preferences").Where(
				sq.Eq{
					"ChannelId":            categoryToUpdate.Channels,
					"Preferences.Category": model.PREFERENCE_CATEGORY_FAVORITE_CHANNEL,
				},
			).ToSql()
```

reflog:
thanks for the review @isacikgoz , I've applied your suggestions.
regarding the error on start - that because I've just recently added another field to the db struct and it's not present on your machine, and no 'upgrade.go' code is present. please drop SidebarCategories table and restart, it should be fine.

hmhealey:
@reflog I just ran this locally, and I got a panic
```
DEBU[2020-05-07T10:10:08.0864909-04:00] Worker received a new candidate job.          caller="migrations/worker.go:57" worker=Migrations
DEBU[2020-05-07T10:10:08.898164-04:00] Worker finished                               caller="migrations/worker.go:47" worker=Migrations
panic: interface conversion: interface {} is nil, not int64

goroutine 4026 [running]:
github.com/mattermost/mattermost-server/v5/migrations.(*Worker).runSidebarCategoriesPhase2Migration(0xc003b96540, 0x0, 0x0, 0xc004862a01, 0x1014bee, 0xc001351b70, 0x2bf67c9)
	/Users/harrison/go/src/github.com/mattermost/mattermost-server/migrations/sidebar_categories_phase_2.go:125 +0xe09
github.com/mattermost/mattermost-server/v5/migrations.(*Worker).runMigration(0xc003b96540, 0xc0048d2150, 0x24, 0x0, 0x0, 0xc003c4c220, 0x1a, 0xc0048629c0, 0x40a8040)
	/Users/harrison/go/src/github.com/mattermost/mattermost-server/migrations/worker.go:154 +0x3a5
github.com/mattermost/mattermost-server/v5/migrations.(*Worker).DoJob(0xc003b96540, 0xc003c57b60)
	/Users/harrison/go/src/github.com/mattermost/mattermost-server/migrations/worker.go:103 +0x349
github.com/mattermost/mattermost-server/v5/migrations.(*Worker).Run(0xc003b96540)
	/Users/harrison/go/src/github.com/mattermost/mattermost-server/migrations/worker.go:58 +0x1bb
created by github.com/mattermost/mattermost-server/v5/jobs.(*Workers).Start.func1
	/Users/harrison/go/src/github.com/mattermost/mattermost-server/jobs/workers.go:92 +0x120
exit status 2
```

hmhealey:
No, it didn't run successfully. It created 300 entries (so it presumably did 100 users) and then it panicked and crashed.

reflog:
@hmhealey apologies, should be fixed now

hmhealey:
Do we have a way to test this on an actual server before merging it to master? We don't want to bring down community-daily with this

hmhealey:
And what rows in the database do I need to delete to reset the migration? I'm getting a different nil pointer now
```
DEBU[2020-05-07T11:10:22.9962502-04:00] Worker received a new candidate job.          caller="migrations/worker.go:57" worker=Migrations
DEBU[2020-05-07T11:10:25.1593611-04:00] Worker finished                               caller="migrations/worker.go:47" worker=Migrations
panic: interface conversion: interface {} is nil, not string

goroutine 4024 [running]:
github.com/mattermost/mattermost-server/v5/migrations.(*Worker).runSidebarCategoriesPhase2Migration(0xc003ae9680, 0xc0049a6750, 0xcc, 0xc0049e0401, 0x1014bee, 0xc001ed21a0, 0x2bf6769)
	/Users/harrison/go/src/github.com/mattermost/mattermost-server/migrations/sidebar_categories_phase_2.go:119 +0xe0c
github.com/mattermost/mattermost-server/v5/migrations.(*Worker).runMigration(0xc003ae9680, 0xc001de4120, 0x24, 0xc0049a6750, 0xcc, 0xc00179f000, 0xc0049a6750, 0xcc, 0x0)
	/Users/harrison/go/src/github.com/mattermost/mattermost-server/migrations/worker.go:154 +0x3a5
github.com/mattermost/mattermost-server/v5/migrations.(*Worker).DoJob(0xc003ae9680, 0xc0049e0420)
	/Users/harrison/go/src/github.com/mattermost/mattermost-server/migrations/worker.go:103 +0x349
github.com/mattermost/mattermost-server/v5/migrations.(*Worker).Run(0xc003ae9680)
	/Users/harrison/go/src/github.com/mattermost/mattermost-server/migrations/worker.go:58 +0x1bb
created by github.com/mattermost/mattermost-server/v5/jobs.(*Workers).Start.func1
	/Users/harrison/go/src/github.com/mattermost/mattermost-server/jobs/workers.go:92 +0x120
exit status 2
```

api4/channel.go
```diff
@@ -1754,3 +1762,227 @@ func patchChannelModerations(c *Context, w http.ResponseWriter, r *http.Request)
...
+	defer c.LogAuditRec(auditRec)
+
+	categoriesUpdateRequest, err := model.SidebarCategoriesFromJson(r.Body)
+	if err != nil {
+		c.SetInvalidParam("category")
+		return
+	}
+	var channelsToCheck []string
+	for _, category := range categoriesUpdateRequest {
+		if !c.App.SessionHasPermissionToCategory(*c.App.Session(), c.Params.UserId, c.Params.TeamId, category.Id) {
```
hmhealey, line 1913: Should we confirm that the Team ID parameter and the category's TeamID field match here? I just realized that this API could theoretically be used to update categories across multiple teams at once.

reflog: @hmhealey the check is done inside the SessionHasAccessToCategory

Makefile
```diff
@@ -266,7 +266,7 @@ ifeq ($(BUILD_ENTERPRISE_READY),true)
 		$(GO) test $(GOFLAGS) -race -run=$(TESTS) -c $$package; \
 		if [ -f $$(basename $$package).test ]; then \
 			echo "Testing "$$package; \
-			./$$(basename $$package).test -test.timeout=2000s || exit 1; \
+			./$$(basename $$package).test -test.timeout=4000s || exit 1; \
```
agnivade, line 269: I am wondering why we are bumping the timeout in EE packages ?

PR #14365, opened 2020-04-23
MM-24500 Fix API ping endpoint race condition

gabrieljackson:
This corrects an issue found when running multiple clustered
Mattermost servers and using the optional `get_server_status`
check on the ping API endpoint. This is done by allowing each
server to write and read from its own health check key in the
system table.

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-24500


gabrieljackson:
I couldn't find a great way to ensure we get the desired non-race behavior with unit tests, so let me know if you have any ideas on that.

Also, this implementation could "leak" a key in theory if for some reason it can write to the system table, but can't delete that key. I am not sure if it is worth adding extra complexity to deal with this or not.

hmhealey:
> Also, this implementation could "leak" a key in theory if for some reason it can write to the system table, but can't delete that key. I am not sure if it is worth adding extra complexity to deal with this or not.

I'd say that that shouldn't happen, but if I say that, then it will probably happen.

gabrieljackson:
@saturninoabril, I believe there may have been an issue with the SpinWick itself. I just recreated the server and it seems to be performing correctly again.

PR #14368, opened 2020-04-24
MM-24397: Reusing the read buffer while reading messages from websockets

agnivade:
#### Summary

The core problem was that conn.ReadMessage allocated a buffer every time it was read.
This created heavy amount of allocations every single time we read a message from the websocket.

To avoid this, we bypass the ReadMessage which was more of a helper method,
and actually call the NextReader which returns a reader object.
We can then reuse a single byte.Buffer instance to read the object unmarshal
into a WebSocketEvent object. This gets rid of the allocation in the read path completly
and allows GC more time to do other tasks.

As we can see from the alloc_space profile here:

Before:
baseheap

After:
reusedheap

There were 2 big boxes in the original profile. Not there's just one. So effectively 10GB of allocation is gone.

#### Ticket link

https://mattermost.atlassian.net/browse/MM-24397

### DIFF
--- a/model/websocket_client.go
+++ b/model/websocket_client.go
@@ -175,7 +175,6 @@ func (wsc *WebSocketClient) Listen() {
 		for {
 			// Reset buffer.
 			buf.Reset()
-			var rawMsg json.RawMessage
 			_, r, err := wsc.Conn.NextReader()
 			if err != nil {
 				if !websocket.IsCloseError(err, websocket.CloseNormalClosure, websocket.CloseNoStatusReceived) {
@@ -193,7 +192,6 @@ func (wsc *WebSocketClient) Listen() {
 				return
 			}
 
-			rawMsg = buf.Bytes()
 			event := WebSocketEventFromJson(&buf)
 			if event == nil {
 				continue
@@ -204,7 +202,7 @@ func (wsc *WebSocketClient) Listen() {
 			}
 
 			var response WebSocketResponse
-			if err := json.Unmarshal(rawMsg, &response); err == nil && response.IsValid() {
+			if err := json.Unmarshal(buf.Bytes(), &response); err == nil && response.IsValid() {
 				wsc.ResponseChannel <- &response
 				continue
 			}
--- a/model/websocket_client.go
+++ b/model/websocket_client.go
@@ -173,6 +173,8 @@ func (wsc *WebSocketClient) Listen() {
 		buf.Grow(avgReadMsgSizeBytes)
 
 		for {
+			// Reset buffer.
+			buf.Reset()
 			var rawMsg json.RawMessage
 			_, r, err := wsc.Conn.NextReader()
 			if err != nil {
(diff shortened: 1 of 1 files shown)


model/websocket_client.go
```diff
@@ -167,19 +170,34 @@ func (wsc *WebSocketClient) Listen() {
...
+			if err != nil {
+				// This should use a different error ID, but en.json is not imported anyways.
+				// It's a different bug altogether but we let it be for now.
+				// See MM-24520.
+				wsc.ListenError = NewAppError("NewWebSocketClient", "model.websocket_client.connect_fail.app_error", nil, err.Error(), http.StatusInternalServerError)
 				return
 			}
 
+			rawMsg = buf.Bytes()
 			event := WebSocketEventFromJson(bytes.NewReader(rawMsg))
```
streamer45, line 198: Can we just pass `&buf` here?

model/websocket_client.go
```diff
@@ -167,19 +170,34 @@ func (wsc *WebSocketClient) Listen() {
...
 			close(wsc.quitPingWatchdog)
 		}()
+		var buf bytes.Buffer
+		buf.Grow(avgReadMsgSizeBytes)
 
 		for {
 			var rawMsg json.RawMessage
 			var err error
-			if _, rawMsg, err = wsc.Conn.ReadMessage(); err != nil {
+			var r io.Reader
```
streamer45, line 179: Not sure we really need to declare these.

model/websocket_client.go
```diff
@@ -195,7 +192,7 @@ func (wsc *WebSocketClient) Listen() {
 			}
 
 			rawMsg = buf.Bytes()
-			event := WebSocketEventFromJson(bytes.NewReader(rawMsg))
+			event := WebSocketEventFromJson(&buf)
 			// Reset buffer.
 			buf.Reset()
```
streamer45, line 197: I am wondering if it's safe to use `rawMsg` (which points to the data in the buffer) after this call.

agnivade: Good catch. Unfortunately now it's going to make things ugly, because this is in a loop so I can't use defer. I _could_ use a function, but I am wondering about how readable it's going to make things.

streamer45: Can we reset once at the top of the loop?

agnivade: That's perfect :ok_hand:

model/websocket_client.go
```diff
@@ -193,8 +195,6 @@ func (wsc *WebSocketClient) Listen() {
 
 			rawMsg = buf.Bytes()
```
streamer45, line 196: I would maybe move this right before we use it (and possibly remove the var declaration on top) or directly pass `buf.Bytes()` to `json.Unmarshal()`.

agnivade:
Great, let me update the load test one. You can test it.

agnivade:
> If I've understood correctly, what we want to avoid here is the call to ReadAll inside ReadMessage, which in turns allocates a new buffer every time, right?

Right on. :)

agnivade:
@lindy65 - Sending another websocket PR your way. Apologies for so many PRs. Let me know if I should offload this one to @ogi-m.

lindy65:
No problem @agnivade - can you help add test steps here or in the Jira ticket please?  Or is general websocket testing sufficient?

agnivade:
Apologies. General websocket testing is sufficient. Thanks !

lindy65:
Perfect - thanks!

PR #14374, opened 2020-04-26
MM-23093  Implement Server Setup telemetry - server configuration

reflog:
#### Summary
Added advanced first day diagnostics reporting.
Previously diagnostics were reported on boot and then every 24 hours.
This PR changes this to: 
* Send once every 10 minutes for the first hour
* Send once every hour thereafter for the first 12 hours
* Send at the 24 hour mark and every 24 hours after

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-23093


app/server.go
```diff
@@ -727,11 +735,27 @@ func runSecurityJob(s *Server) {
 	}, time.Hour*4)
 }
 
+func doDiagnosticsIfNeeded(s *Server) {
```
hanzei, line 738: I'm confused. This methods is never called. Is there an EE PR?

model/config.go
```diff
@@ -300,6 +300,7 @@ type ServiceSettings struct {
 	DEPRECATED_DO_NOT_USE_RestrictPostDelete          *string `json:"RestrictPostDelete" mapstructure:"RestrictPostDelete"`                   // This field is deprecated and must not be used.
 	DEPRECATED_DO_NOT_USE_AllowEditPost               *string `json:"AllowEditPost" mapstructure:"AllowEditPost"`                             // This field is deprecated and must not be used.
 	PostEditTimeLimit                                 *int
+	ServerFirstRun                                    *int64 `json:"ServerFirstRun"`
```
hanzei, line 303: Why is this part of the config and not the DB?

app/server.go
```diff
@@ -727,11 +735,27 @@ func runSecurityJob(s *Server) {
...
+	hoursSinceFirstServerRun := time.Since(time.Unix(*s.Config().ServiceSettings.ServerFirstRun, 0)).Hours()
+	// Send once every 10 minutes for the first hour
```
hanzei, line 740: I'm worried that admins might be surprised that data is send to a third party so quickly after setting up a server.

In the past there have been a lot of reports by community members complaining about our aggressive collecting of data. (https://github.com/mattermost/mattermost-server/issues/10595, https://github.com/mattermost/mattermost-server/issues/9466) With this changes, Admins who care about there data, won't be able to disable telemetry in time.

cc @adamjclarkson

reflog: @adamjclarkson , @grundleborg  thoughts?

amyblais: cc @wiersgallak

app/server.go
reflog: no, just a dumb typo :(
thanks for catching it. fixed

model/config.go
reflog: My thinking was - we need to start reporting ASAP, maybe even before DB connection is made.
cc @grundleborg  thoughts?

amyblais:
@adamjclarkson @wiersgallak @grundleborg Do we still want this for v5.23? If yes are the concerns being discussed (tomorrow is code complete for v5.23)?

app/server.go
adamjclarkson: The analytics we are collecting are not changing, it's only the frequency that we're collecting them. So there is no encroachment here on privacy. Admins are still able to turn off telemetry and opt out at any time. The data itself gives us clarity on what Admins are doing during the first 24 hours in order to improve the setup process, guiding admins more in the future through configuration.

adamjclarkson:
Yes we definitely need this 5.23. I understand the concerns but this change is not a substantial change to our approach too analytics, we're only increasing frequency of data collection in first 24 hours.

model/config.go
grundleborg: I don't think we should be using the config like this. In HA setups where orchestration deploys the config files (e.g. they are kept in Git and deployed by Chef or Ansible or the like), it will get overwritten every time they deploy. This should be in the "Systems" table in the database. I don't think there's any need to send telemetry before the database is up and running.

reflog: Thanks. Will re-do using the database

app/server.go
```diff
@@ -727,11 +728,28 @@ func runSecurityJob(s *Server) {
...
+	} else if hoursSinceFirstServerRun > 12 && time.Since(s.timestampLastDiagnosticSent) >= 24*time.Hour {
+		doDiagnostics(s)
+	}
+}
+
 func runDiagnosticsJob(s *Server) {
+	// Send on boot
 	doDiagnostics(s)
+	// we ignore the error, since it's been ensured on server start
+	firstRun, _ := s.FakeApp().getFirstServerRunTimestamp()
```
grundleborg, line 749: Can still fail though to query the database though. What would the intended behaviour be in that case? I'd prefer to see the error handled and a sensible default value for `firstRun` be set based on that - it feels more explicit that way.

reflog: done.

app/server.go
```diff
@@ -727,11 +728,32 @@ func runSecurityJob(s *Server) {
 	}, time.Hour*4)
 }
 
+func doDiagnosticsIfNeeded(s *Server, firstRun time.Time) {
```
hanzei, line 731: Given the importance of correctly running diagnostics, I would really like to see tests for the change.

app/server.go
hanzei: The discussion isn't really about privacy, is about data ownership. Surely admins can opt out, but at the point they data already has been send, which is a bummer for some people. 

There are a lot of reports from admins who didn't know telemetry exist and discover to late that data has been send. With this change the amount of data that has been send will be even higher.

It's also unlikely that an admin will find out in the first 10 minutes that telemetry exists. Hence, this will surprise a lot of admins.

hanzei:
On other note: I'm worried about introducing such fundamental change just before feature complete - in a quality release.  I'm aware that getting telemetry data take a while and hence such changes should go in quickly, but the consequence of losing data would be remarkable.

app/server.go
adamjclarkson: I get it. From my end we desperately need more data to help us understand what Admins are doing (or not doing) in the first 24 hours to help us understand why 75% of servers don't last longer than 1 day. Improving 7 day server retention is one of the key objectives under our V2MOM and this data will provide critical information as we try to solve that problem. 
I am currently working on another project that is a redesign of the admin onboarding flow. In it we could introduce an opt-out for telemetry making clearer to users that we are collecting, link to the documentation and provide easy way for them to decline.  No surprises.

hanzei: It great to hear that the on boarding experience is planed to be redesigned.  :+1: 

With this in mind I have less concerns with merging this PR also because I see the need to improve the 7 day server retention.

app/server.go
reflog: @hanzei , will add tests.

grundleborg: We have tests for the diagnostics feature generally. I don't think it's necessary to insist on tests for the new timing calculation - a manual test should be sufficient here. Once we clean up and remove the Segment stuff for 5.24 or 5.26, we could look at a more general refactor to make more of this feature testable through automated tests.

amyblais:
@reflog Will this PR need QA review?

reflog:
@amyblais - yes

amyblais:
@reflog Let me know when this has been cherry picked

PR #14377, opened 2020-04-26
app: outdent skipSend condition

agnivade:
We invert the skipSend condition and outdent the remaining block
to make the code a bit more idiomatic.

While here, we also change the dropping message level from info to
warn because that's what it should be.

### DIFF
--- a/app/web_conn.go
+++ b/app/web_conn.go
@@ -187,7 +187,7 @@ func (wc *WebConn) writePump() {
 				case model.WEBSOCKET_EVENT_TYPING,
 					model.WEBSOCKET_EVENT_STATUS_CHANGE,
 					model.WEBSOCKET_EVENT_CHANNEL_VIEWED:
-					mlog.Info(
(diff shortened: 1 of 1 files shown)


agnivade:
@lindy65 - This is a super minor change. Maybe just do a very brief sanity checks around websockets. That oughta suffice.

lindy65:
Thanks @agnivade - will do... An FYI that there are some issues with spinwicks at the moment so will test as soon as these are resolved.

Also, is there a Jira ticket for this PR?

agnivade:
> Also, is there a Jira ticket for this PR?

There isn't. It was just a spontaneous PR of 0.5 Mana worth. But let me know if you want one.

lindy65:
Yes please @agnivade as I will need to test on a resolved ticket post-merge :)

agnivade:
Here you go: https://mattermost.atlassian.net/browse/MM-24551

PR #14378, opened 2020-04-27
MM-24310: Recycle DB connections properly

agnivade:
#### Summary

We set the MaxConnLifetime to let database/sql automatically
reap connections which have exceeded a duration. This lets
us gradually close old connections as they exceed their lifetime.

The previous method instantiated a new store altogether that was
costly and error-prone. There is no need to instantiate application
level structures when we can just set the maxconnlifetime.

And after that's done, we reset the lifetime to their original value.

While we are here, we also fix a bug in the DoesTableExist method
for sqlite DB.

#### Ticket Link

https://mattermost.atlassian.net/browse/MM-24310


store/sqlstore/supplier.go
```diff
@@ -1004,6 +1004,23 @@ func (ss *SqlSupplier) GetAllConns() []*gorp.DbMap {
...
+// to d, and then resets them back to their original duration.
+func (ss *SqlSupplier) RecycleDBConnections(d time.Duration) {
+	// Get old time.
+	originalDuration := time.Duration(*ss.settings.ConnMaxLifetimeMilliseconds) * time.Millisecond
+	// Set the max lifetimes for all connections.
+	for _, conn := range ss.GetAllConns() {
+		conn.Db.SetConnMaxLifetime(d)
+	}
+	// Wait for that period with an additional 2 seconds of scheduling delay.
+	time.Sleep(d + 2*time.Second)
```
lieut-data, line 1017: Would it makes sense to hard-code the various durations entirely within this method? Not sure the caller has the context necessary to provide a meaningful value for `d`.

(If this was just added to enable a shorter unit test, then perhaps we can export the recommended duration value from this package for use by the `app` layer?)

store/sqlstore/supplier_test.go
```diff
@@ -151,6 +153,45 @@ func TestGetDbVersion(t *testing.T) {
...
+			settings := makeSqlSettings(driver)
+			supplier := sqlstore.NewSqlSupplier(*settings, nil)
+
+			var wg sync.WaitGroup
+			tables := []string{"Posts", "Channels", "Users"}
+			for _, table := range tables {
+				wg.Add(1)
+				go func(table string) {
+					defer wg.Done()
+					res := supplier.DoesTableExist(table)
```
lieut-data, line 177: It's not super obvious why this test is testing table existence? Should we just run a hard-coded, manual query if we're trying to open a connection? (The current code suggests that the recycling is somehow tied to table existence.)

store/sqlstore/supplier.go
agnivade: > (If this was just added to enable a shorter unit test, then perhaps we can export the recommended duration value from this package for use by the app layer?)

mmm .. I would prefer not to export a package level constant only to be overriden by the tests.

To me, it gives a certain amount of control to the caller. The context is the amount of grace period that can be allowed for connections to be reaped. This wasn't the case originally, and we were sleeping just for a random period of time. In future, we can even add a query param to the API call and pass that here.

That was my reasoning to pass the value from upper layer. I will add a comment to explain this.

 It is an important knob because it gives us control over how fast/slow to reap connections. This all depends on what the customer wants to do / the kind of workload.

store/sqlstore/supplier_test.go
agnivade: > Should we just run a hard-coded, manual query if we're trying to open a connection? 

Yeah, I think that makes sense.

store/sqlstore/supplier_test.go
```diff
@@ -174,8 +174,9 @@ func TestRecycleDBConns(t *testing.T) {
 				wg.Add(1)
 				go func(table string) {
 					defer wg.Done()
-					res := supplier.DoesTableExist(table)
-					assert.True(t, res)
+					query := `SELECT count(*) FROM ` + table
+					_, err := supplier.GetMaster().SelectInt(query)
+					assert.Nil(t, err)
```
lieut-data, line 179: Since `SelectInt` returns an `error`, we could use `assert.NoError(t, err)` here instead of the purely-nil check required for `*model.AppError`.

agnivade: Good one. We should probably have a vet check to catch other cases like this throughout the codebase.

agnivade:
gentle ping @ogi-m

PR #14385, opened 2020-04-27
MM-24395: Set response header to be attachment for SVG images.

agnivade:
We check the file extension and appropriately set the response header.
SVG images without a file extension aren't proxied at all. So there's no
problem with that.

#### Ticket link

https://mattermost.atlassian.net/browse/MM-24395

### DIFF
--- a/services/imageproxy/local.go
+++ b/services/imageproxy/local.go
@@ -7,9 +7,12 @@ import (
 	"errors"
 	"io"
 	"io/ioutil"
+	"mime"
 	"net/http"
 	"net/http/httptest"
 	"net/url"
+	"path/filepath"
+	"strings"
 
 	"github.com/mattermost/mattermost-server/v5/mlog"
 	"github.com/mattermost/mattermost-server/v5/services/httpservice"
@@ -76,6 +79,19 @@ func (backend *LocalBackend) GetImage(w http.ResponseWriter, r *http.Request, im
 		w.WriteHeader(http.StatusBadRequest)
 		w.Write([]byte{})
 		return
+	}
+
+	u, err := url.Parse(imageURL)
+	if err != nil {
+		mlog.Error("Failed to parse URL for proxied image", mlog.String("url", imageURL), mlog.Err(err))
+		w.WriteHeader(http.StatusBadRequest)
+		w.Write([]byte{})
+		return
+	}
+
+	extension := strings.ToLower(filepath.Ext(u.Path))
+	if mime.TypeByExtension(extension) == "image/svg+xml" {
+		w.Header().Set("Content-Disposition", "attachment;filename=\""+filepath.Base(u.Path)+"\"")
 	}
 
 	w.Header().Set("X-Frame-Options", "deny")
--- a/services/imageproxy/local.go
+++ b/services/imageproxy/local.go
@@ -5,8 +5,10 @@ package imageproxy
 
 import (
 	"errors"
+	"fmt"
 	"io"
 	"io/ioutil"
+	"mime"
 	"net/http"
 	"net/http/httptest"
 	"net/url"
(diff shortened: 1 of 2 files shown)


services/imageproxy/local.go
```diff
@@ -78,6 +81,19 @@ func (backend *LocalBackend) GetImage(w http.ResponseWriter, r *http.Request, im
...
+	u, err := url.Parse(imageURL)
+	if err != nil {
+		mlog.Error("Failed to parse URL for proxied image", mlog.String("url", imageURL), mlog.Err(err))
+		w.WriteHeader(http.StatusBadRequest)
+		w.Write([]byte{})
+		return
+	}
+
+	extension := strings.ToLower(filepath.Ext(u.Path))
+	if mime.TypeByExtension(extension) == "image/svg+xml" {
```
jupenur, line 93: The proxy backend primarily uses the `Content-Type` response header for MIME type comparisons, so there's a potential mismatch here: a malicious or otherwise misconfigured image server can serve `/image.jpg` with `Content-Type: image/svg+xml`.

A few alternative ideas on how to fix it:

 - Use the `GetImageDirect` implementation instead of `GetImage` -- likely comes with a performance hit since responses would then be buffered
 - Wrap the `http.ResponseWriter` in a custom implementation of the interface that allows us to conditionally set a `Content-Disposition` at the same time as the `Content-Type` value is being copied over
 - Fork the proxy backend and add the `Content-Disposition` header there
 - Somehow enforce the content type detected here and overwrite the one detected by the backend

I'm 0/5 on how feasible any of these are, though. If setting the `Content-Disposition` header conditionally turns out too complicated, we might want to set it on all responses instead, but that probably requires explicit PM approval.

agnivade: I have been thinking about 2. But the problem seems to be that the ResponseWriter wrapper only allows us to hook into WriteHeader and Write calls. I'll try and see if I can get the value of the Header within these calls, and if so, modify it.

I think 1 would not be very ideal and I would consider it after all the other alternatives.

agnivade: Nvm .. found it. PTAL.

services/imageproxy/local.go
```diff
@@ -66,6 +67,20 @@ func makeLocalBackend(proxy *ImageProxy) *LocalBackend {
...
 }
 
+type contentTypeRecorder struct {
+	http.ResponseWriter
+	filename string
+}
+
+func (rec *contentTypeRecorder) WriteHeader(code int) {
+	hdr := rec.ResponseWriter.Header()
+	if hdr.Get("Content-Type") == "image/svg+xml" {
```
jupenur, line 77: Matching directly against the header value won't work since it could contain additional params. Parse with [`mime.ParseMediaType`](https://golang.org/pkg/mime/#ParseMediaType) first.

services/imageproxy/local.go
```diff
@@ -66,6 +67,20 @@ func makeLocalBackend(proxy *ImageProxy) *LocalBackend {
...
+		hdr.Set("Content-Disposition", "attachment;filename=\""+rec.filename+"\"")
```
jupenur, line 78: Do we know `rec.filename` won't contain double-quotes? If not, we should replace them with underscores.

agnivade: Used %q to do the quoting. Let me know if that doesn't work.

services/imageproxy/local.go
agnivade: done

services/imageproxy/local.go
```diff
@@ -66,6 +69,26 @@ func makeLocalBackend(proxy *ImageProxy) *LocalBackend {
...
+	defer rec.ResponseWriter.WriteHeader(code)
+
+	hdr := rec.ResponseWriter.Header()
+	contentType := hdr.Get("Content-Type")
+	mediaType, _, err := mime.ParseMediaType(contentType)
+	if err != nil {
+		mlog.Error("error in decoding media type", mlog.Err(err))
+		return
+	}
+	if mediaType == "image/svg+xml" {
```
jupenur, line 87: The error case should either set `Content-Disposition` or somehow bubble up the error to the client: We don't know what the MIME type is or how the client will interpret it, so it could still be SVG.
```suggestion
	if err != nil || mediaType == "image/svg+xml" {
```

agnivade: It might need more work to bubble up the error to the client since it's a middleware. Do you not want to log the error at all ?

jupenur: Logging it doesn't seem strictly necessary since this is an error caused by malformed input from the external image server -- and it's not causing anything to actually fail. But I'm not opposed to logging the error either. Bubbling up probably doesn't make sense, I agree.

agnivade: Alright, I'll make the changes.

PR #14391, opened 2020-04-28
MM-24312: Fix Dependency updates

agnivade:
A bug in hashicorp memberlist caused us to revert the
dependency updates. The bug is now identified to be:
https://github.com/hashicorp/memberlist/issues/219.

It was a regression introduced in memberlist 0.2.0.

A simple revert of the revert was causing a lot of merge conflicts,
so we just upgrade the dependencies afresh, and pin down the memberlist
version to 0.1.5.

The upgrade also caused an incompatibility in the rudder API which needed
to be fixed.

Once memberlist bug is fixed, we need to upgrade it again.

#### Ticket link

https://mattermost.atlassian.net/browse/MM-24312


agnivade:
@grundleborg - please take a look at the rudder changes. FWIW - their master branch looks completely empty and they seem to have deleted a lot of code. Don't know what's going on but it looks very fishy.

app/server.go
```diff
@@ -885,7 +885,7 @@ func (s *Server) initRudder(endpoint string) {
 			config.Verbose = true
 			config.BatchSize = 1
 		}
-		client, err := rudder.NewWithConfig(RUDDER_KEY, config)
+		client, err := rudder.NewWithConfig(RUDDER_KEY, RUDDER_DATAPLANE_URL, config)
```
lieut-data, line 888: Since we're passing `RUDDER_DATAPLANE_URL` directly now, is the `config.Endpoint` setting still relevant?

agnivade: I'll leave it to @grundleborg to take a call on that. I'd like to keep this PR _just_ about updating dependencies with minimal disruption in the codebase.

agnivade:
@lindy65 - Giving it to you since you handled the original PR. Steps are outlined in the JIRA ticket. Thanks !

cpanato:
there are some issues, im fixing, sorry

PR #14395, opened 2020-04-28
IS-255: Remove code foree testing workflow

metanerd:
#### Summary
Awaiting enterprise PR.


metanerd:
@agnivade Jenkinsfile.branch should be still in use.

PR #14399, opened 2020-04-28
MM-24611: Fix flaky test TestDeletePreferencesWebsocket 

agnivade:
After registering the conn in the hub, we proceeded to send a
direct message to the user. We had changed it to send the direct message
in the same hub goroutine that handles the registration. This was the correct
behavior and fixes chances of having panics due to sending to closed channels.

However, often fixing something unearths some deeper underlying bug. This was
such a case :)

The issue was that register channel had a buffer size of 1. And we were sending
a direct message after registration. In the code to send direct message, we
were checking if the user has been registered or not, and if not, then skip it.

Therefore, since the register channel buffer was 1, it could very well be that
the select case would pick up the direct message send case first - in which
case it would not have been registered, and therefore no hello message would be sent.

The fix is to unbuffer the register and unregister channels. There does not seem
to be a valid reason to make these buffered channels. They are meant to be
synchronous operations, because the code following them assumes that the user
has been registered.

While here, we also remove all the time.Sleeps before waiting on the Response channel
because they are not required at all. Waiting on a channel is already blocking.

#### Ticket link

https://mattermost.atlassian.net/browse/MM-24611


cpanato:
thanks @agnivade for opening this temp pr to test. all good now :)

cpanato:
@metanerd i think everything is ok, we can improve more, but the basic feature is available again

agnivade:
Well .. I opened it for my own testing to investigate a flaky test. But if it has been useful, then all the more glad :)

agnivade:
@crspeller - Added you to check whether there was any explicit intention behind adding the buffer of 1.

agnivade:
> If I had to guess it was probably an attempted performance optimization at some point. 

A performance optimization by adding a buffer of 1 ? Hmm .. :thinking: 

Preventing a deadlock sounds more sensible. Running a load test makes total sense. I will leave it in the safe hands of @streamer45.

crspeller:
> A performance optimization by adding a buffer of 1 ?

I didn't say it made sense. I said that's what my guess was 😆

streamer45:
Ran a couple of load-tests. Everything looked good :+1:

agnivade:
@ogi-m - This is a bit hard to verify with QA since it fixes a flaky test. I think the flaky test being fixed is one level in itself. And on top of it, the load tests passing also ensure that no regression have happened.

Let me know your thoughts. Thanks.

PR #14405, opened 2020-04-29
MM-20951 Add sentry support to mattermost-server

reflog:
#### Summary
This PR introduces Sentry support to the server.
It's enabled by default, unless Diagnostics have been previously disabled.
It will have mattermost DSN (i.e. where to report to) in the configuration, but should be possible to reconfigure to use a hosted sentry instance.

#### Ticket Link

https://mattermost.atlassian.net/browse/MM-20951


app/server.go
```diff
@@ -510,6 +527,14 @@ func (s *Server) Start() error {
 	mlog.Info("Starting Server...")
 
 	var handler http.Handler = s.RootRouter
+
+	if *s.Config().LogSettings.EnableSentry {
+		sentryHandler := sentryhttp.New(sentryhttp.Options{
+			Repanic: true,
+		})
+		handler = sentryHandler.Handle(handler)
```
jespino, line 535: This is only handling the panics, right? Do we need to create our own handler wrapper to store the 500 errors too?

reflog: Yes, this is only for panics. Logging handled errors needs to be discussed a little more. (Privacy concerns?)

reflog:
@wiersgallak - there's no GUI for this setting, as it's a rather 'internal' and obscure one. It can easily be added later though.
The default setting will be taken from the EnableDiagnostics as we've discussed.

reflog:
@grundleborg - what do you think about possibility of merging this? Also, had another idea - it's probably good to disable Sentry while developing. I can disable it if `BuildHash=dev`. Thoughts?

reflog:
@grundleborg - the DSN has a default value pointing to Mattermost Sentry instance.
The users can override it to point to their local one

jespino:
@reflog I don't see the point on having your own sentry for this unless you are forking mattermost, and in that cases, you can modify directly the sentry DSN anyway.

reflog:
@jespino @grundleborg I'm 0/5 on this, if you think it should just be hardcoded, I have no issue with it. Less code :)

jespino:
@jespino I'm 4/5 here about, I only see the setting to Enable/Disable the sentry reports, due privacy concerns (people that don't want any data leaving their systems). But changing the DSN is not necessary (IMO).

reflog:
@jespino @grundleborg made DSN hardcoded similar to Rudder. I assume we set the constant during build somewhere?

model/config.go
```diff
@@ -88,18 +88,18 @@ const (
...
+	SERVICE_SETTINGS_DEFAULT_TLS_KEY_FILE          = ""
+	SERVICE_SETTINGS_DEFAULT_READ_TIMEOUT          = 300
+	SERVICE_SETTINGS_DEFAULT_WRITE_TIMEOUT         = 300
+	SERVICE_SETTINGS_DEFAULT_IDLE_TIMEOUT          = 60
+	SERVICE_SETTINGS_DEFAULT_MAX_LOGIN_ATTEMPTS    = 10
+	SERVICE_SETTINGS_DEFAULT_ALLOW_CORS_FROM       = ""
+	SERVICE_SETTINGS_DEFAULT_LISTEN_AND_ADDRESS    = ":8065"
+	SERVICE_SETTINGS_DEFAULT_GFYCAT_API_KEY        = "2_KtH_W5"
+	SERVICE_SETTINGS_DEFAULT_GFYCAT_API_SECRET     = "3wLVZPiswc3DnaiaFoLkDvB4X0IV6CpMkj4tf2inJRsBY6-FnkT08zGmppWFgeof"
+	SERVICE_SETTINGS_DEFAULT_SENTRY_DSN            = "https://9d7c9cccf549479799f880bcf4f26323@o94110.ingest.sentry.io/5212327"
```
grundleborg, line 102:
```suggestion
```

app/server.go
```diff
@@ -193,6 +196,22 @@ func NewServer(options ...Option) (*Server, error) {
 	// Use this app logger as the global logger (eventually remove all instances of global logging)
 	mlog.InitGlobalLogger(s.Log)
 
+	if *s.Config().LogSettings.EnableSentry {
+		if strings.Contains(SENTRY_DSN, "placeholder") {
+			mlog.Warn("Sentry reporting is enabled, but SENTRY_DSN is not set. Disabling reporting.")
+		} else {
+			if model.BuildHash == "dev" {
```
grundleborg, line 203: If we're only injecting the DSN as part of the release process, we don't need this check. Or we could leave this check, hard code the DSN into the source and not have to do any injection in the release process. 0/5 on which we choose.

reflog: Good point. I think it's cleaner to inject the DSN. I'll remove the check

grundleborg:
> made DSN hardcoded similar to Rudder. I assume we set the constant during build somewhere?

To do the injection, take a look at the similar PR here for when Rudder was added: https://github.com/mattermost/mattermost-shared-pipelines/pull/66 and also https://github.com/mattermost/mattermost-shared-pipelines/pull/68 because you'll have the same bug as Rudder otherwise :)

reflog:
thanks @grundleborg , added PR for the pipeline as well.

model/config.go
```diff
@@ -88,18 +88,17 @@ const (
...
-	SERVICE_SETTINGS_DEFAULT_READ_TIMEOUT       = 300
-	SERVICE_SETTINGS_DEFAULT_WRITE_TIMEOUT      = 300
-	SERVICE_SETTINGS_DEFAULT_IDLE_TIMEOUT       = 60
-	SERVICE_SETTINGS_DEFAULT_MAX_LOGIN_ATTEMPTS = 10
-	SERVICE_SETTINGS_DEFAULT_ALLOW_CORS_FROM    = ""
-	SERVICE_SETTINGS_DEFAULT_LISTEN_AND_ADDRESS = ":8065"
-	SERVICE_SETTINGS_DEFAULT_GFYCAT_API_KEY     = "2_KtH_W5"
-	SERVICE_SETTINGS_DEFAULT_GFYCAT_API_SECRET  = "3wLVZPiswc3DnaiaFoLkDvB4X0IV6CpMkj4tf2inJRsBY6-FnkT08zGmppWFgeof"
-
+	SERVICE_SETTINGS_DEFAULT_SITE_URL              = "http://localhost:8065"
```
jespino, line 91: probably gofmt would reduce the number of spaces at the end and probably remove this whole bulk of changes.

vendor/willnorris.com/go/imageproxy/imageproxy.go
```diff
@@ -221,7 +221,8 @@ func (p *Proxy) serveImage(w http.ResponseWriter, r *http.Request) {
 		contentType = peekContentType(b)
 	}
 	if resp.ContentLength != 0 && !contentTypeMatches(p.ContentTypes, contentType) {
-		p.logf("content-type not allowed: %q", contentType)
```
jespino, line 224: nit: looks like this is not related to the changes.

reflog:
@grundleborg @wiersgallak - all yours to review! :)

harshilsharma63:
Just curious, how are you preventing user's data from being sent to Sentry?

reflog:
@harshilsharma63 - the only thing logged is the crash callstack. No personal data is transmitted.

harshilsharma63:
@reflog yes I understand that. What I meant was that if anyone has tweaked any configuration on the Sentry side as an extra effort to prevent this accidentally. I'm asking this because I'm planning to do the same in one of my plugins.

wiersgallak:
> should we check diagnostics are enabled and sentry is enabled to send sentry responses? Otherwise when people turn off diagnostics sentry data will still be sent to us.

Yes, I agree we should check both.

reflog:
@grundleborg @wiersgallak - done

reflog:
> @reflog yes I understand that. What I meant was that if anyone has tweaked any configuration on the Sentry side as an extra effort to prevent this accidentally. I'm asking this because I'm planning to do the same in one of my plugins.

@harshilsharma63 - there's no tweaking done server side, atleast not at the moment

reflog:
@lindy65 - there isn't much for QA to test here, it's a crash reporting feature, server side only. Full functionality is covered by tests.

reflog:
@wiersgallak - do we need to document this anywhere?

wiersgallak:
@reflog Yeah, we will add this to our docs for the release.  Added a tag and assigned to me so we ensure it is picked up in that process.

reflog:
@wiersgallak so it's safe to merge?

PR #14406, opened 2020-04-29
MM-24547: Fix writer leak when connection closes

agnivade:
When the connection is closed, the exit path does not
shut down the writer goroutine. In which case, it will keep spinning forever.

Since we already have the CAS mechanism now, we can use that to close the writer
and guard from a double-close.

The problem with refactoring the writer closing to a common
function was that we needed to wait for the reader to exit
before closing the EventChannel and ResponseChannel.

But then there is another problem that the API can be used in such
a way that the client is liable to call Close without even calling
Listen. In that case, we cannot wait for Listen to quit.

So from Close, we can only close the connection. And therefore
we need to duplicate the writer closing in the read loop's
defer block.

This makes closing the websocket client idempotent from both perspective -
- Explicitly closing.
- Closing due to connection tear down.

There are still 2 races left:
- Using the exported Conn to directly write messages. We cannot do anything about
that as long as clients directly using that.
- Setting the wsc.pingTimeoutTimer field in a separate goroutine when calling
.Connect(). This will need to be seen later.

### Ticket link

https://mattermost.atlassian.net/browse/MM-24547

### DIFF
--- a/model/websocket_client.go
+++ b/model/websocket_client.go
@@ -183,10 +183,10 @@ func (wsc *WebSocketClient) Listen() {
 	// But if (1) happens, we set the closed bit, and close the rest of the stuff.
 	go func() {
 		defer func() {
-			// We CAS to 1 and proceed.
 			close(wsc.EventChannel)
 			close(wsc.ResponseChannel)
 			close(wsc.quitPingWatchdog)
+			// We CAS to 1 and proceed.
 			if !atomic.CompareAndSwapInt32(&wsc.closed, 0, 1) {
 				return
 			}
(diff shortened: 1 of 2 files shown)


agnivade:
@streamer45 - This is ready to be tested now.

agnivade:
@lindy65 - This might be a bit hard to test as this requires using the websocket client code to be used somehow. This is not related to the mattermost server in any way.

FWIW, the PR adds unit tests for these conditions. And Claudio has run tests for nearly 2 hours on this PR and verified that there are no regressions and it fixes what it intends to fix.

PR #14410, opened 2020-04-29
MM-24575 Error in logs when sending DM to same account

wiggin77:
#### Summary

This PR fixes the issue where an error was appearing in the logs any time a user sends a DM to themself.  `GetUser` was being called with a blank id.

#### Ticket Link

https://mattermost.atlassian.net/browse/MM-24575

### DIFF
--- a/app/auto_responder.go
+++ b/app/auto_responder.go
@@ -18,6 +18,10 @@ func (a *App) SendAutoResponseIfNecessary(channel *model.Channel, sender *model.
 	}
 
 	receiverId := channel.GetOtherUserIdForDM(sender.Id)
+	if receiverId == "" {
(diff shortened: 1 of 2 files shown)


wiggin77:
> Is this something we could adds tests for? I know anything using a go func isn't exactly easy to test

Agreed, I thought about this.  I can't think of a useful test that adds value without a large amount of work to deal with the async call.  I decided it wasn't worth it to test a single `if` statement.

If you can think of a relatively simple test that adds value I'm happy to implement it.

hmhealey:
The only thing I could think of is wrapping the contents of the go func in another function, but that's basically what `SendAutoResponseIfNecessary` does already, so it'd feel unnecessary. You'd also be just testing an if statement which isn't particularly useful.

PR #14420, opened 2020-04-29
Check license expiration, disable license and notify administrator

catalintomai:
#### Summary
- once a day, checks the license expiration and notifies the admin that the license was disabled (not physically removed) if the grace period expired.
- reduces the license expiration grace period from 15 days to 10 days.
- adds new template for admin email notification

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-24661

#### Related PRs
webapp: https://github.com/mattermost/mattermost-webapp/pull/5418

#### Screenshots

Screen Shot 2020-04-29 at 12 22 11 PM


app/server.go
```diff
@@ -781,6 +791,46 @@ func doSessionCleanup(s *Server) {
...
+	}
+
+	users, err := s.Store.User().GetSystemAdminProfiles()
+	if err != nil {
+		mlog.Error("Failed to get system admins for license expired message from Mattermost.")
+		return
+	}
+
+	//send email to admin(s)
+	for _, user := range users {
```
jespino, line 815: We are going to send this message every day?

app/server.go
```diff
@@ -781,6 +791,46 @@ func doSessionCleanup(s *Server) {
 	s.Store.Session().Cleanup(model.GetMillis(), SESSIONS_CLEANUP_BATCH_SIZE)
 }
 
+func doLicenseExpirationCheck(s *Server) {
+	s.FakeApp().LoadLicense()
+	license := s.License()
+
+	if license == nil {
+		mlog.Info("License cannot be found.")
```
jespino, line 799: I'm not sure about logging this cases and the next one. For me that 2 cases can be "business as usual".

app/server.go
```diff
@@ -781,6 +791,46 @@ func doSessionCleanup(s *Server) {
...
+
+		mlog.Info("Sending license expired email.", mlog.String("user_email", user.Email))
+		s.Go(func() {
+			if err := s.FakeApp().SendRemoveExpiredLicenseEmail(user.Email, user.Locale, *s.Config().ServiceSettings.SiteURL, license.Id); err != nil {
+				mlog.Error("Error while sending the license expired email.", mlog.String("user_email", user.Email), mlog.Err(err))
+			}
+		})
+	}
+
+	//remove the license
```
jespino, line 830: Probably adding here a Warning or Info log here about removing the license would be great.

app/server.go
```diff
@@ -781,6 +791,46 @@ func doSessionCleanup(s *Server) {
...
+
+	//send email to admin(s)
+	for _, user := range users {
+		user := user
+		if user.Email == "" {
+			mlog.Error("Invalid system admin email.", mlog.String("user_email", user.Email))
+			continue
+		}
+
+		mlog.Info("Sending license expired email.", mlog.String("user_email", user.Email))
```
jespino, line 822: If we warn about the fact that we are removing the license and we have notified the admins, probably not showing this log message is perfectly fine.

app/server.go
catalintomai: Only once, after the license removal, the first check returns.

app/server.go
catalintomai: sure, RemoveLicense() doesn't looks like it logs anything itself - thought it is a pretty heavy-duty operation.

model/license_test.go
```diff
@@ -116,6 +116,15 @@ func TestLicenseIsExpired(t *testing.T) {
 	assert.False(t, l1.IsExpired())
 }
 
+func TestLicenseIsPastGracePeriod(t *testing.T) {
+	l1 := License{}
+	l1.ExpiresAt = GetMillis() - LICENSE_GRACE_PERIOD - 1
+	assert.True(t, l1.IsPastGracePeriod())
+
+	l1.ExpiresAt = GetMillis() + 1
```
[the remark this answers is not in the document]
catalintomai: sure, we can change it, the thinking was that if one specifies a value somebody can question it later- like why that specific value, does it have a certain meaning - with "1" it is more clear (to me) that it is just for "going over/under limit" purpose, but I'm 0/5 on this.

app/server.go
catalintomai: The log entry is for us and is for a different operation (sending email) - the admin notification is for the user; same as above, we can remove it or make it Debug. Thoughts?

app/server.go
jespino: :facepalm: yes, you are totally right.

app/server.go
jespino: Yes, I think Debug is more appropriate here, because this information is not very useful for regular users without license, until it is deactivated, and when that happens, is properly logged.

app/server.go
jespino: Yes, clearly should be logged.

app/server.go
jespino: Yes, I think Debug would be more appropriate, but if you move the log of "RemoveLicense" to the "RemoveLicense" method, then, that log message will be generic, and then, probably makes sense to log it here.

model/license_test.go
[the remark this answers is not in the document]
jespino: yes, but 1ms is a very small amount of time and if the cpu switch the context between the two GetMillis, can fail the tests. I think GetMillis() + 1000 is pretty obvuious that you are saying "one second later" instead of "1 millisecond later".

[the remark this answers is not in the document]
catalintomai: agree, didn't think in terms of milliseconds/seconds.

app/server.go
catalintomai: yes, I was thinking of moving  it to RemoveLicense() as well, it makes more sense.

Issue #10327, opened 2019-02-20
Ignore configured database timeout during migration


esethna:
If you're interested please comment here and come [join our "Contributors" community channel](https://community.mattermost.com/core/channels/tickets) on our daily build server, where you can discuss questions with community members and the Mattermost core team. For technical advice or questions, please  [join our "Developers" community channel](https://community.mattermost.com/core/channels/developers).

New contributors please see our [Developer's Guide](https://developers.mattermost.com/contribute/getting-started/).

----

**Notes**: [Jira ticket](https://mattermost.atlassian.net/browse/MM-13799)

Mattermost instances on Postgres (and possibly MySQL) may face issues on upgrade if the required schema changes take longer than the configured timeout. It probably never makes sense to timeout the upgrade.

For this ticket, first verify that is the case by discussing the ticket in the ~Developers channel on community.mattermost.com. Both `@jesus.espino` and `@jesse.hallam` would be good to discuss it with.

If verified, then implement the changes in server store code.

joshuabezaleel:
May I take a shot on this?

hanzei:
Sure @joshuabezaleel, thanks for taking a look. :+1:

ConnorL33t:
I'd like to work on this!

hanzei:
Thanks @ConnorL33t :rocket: If you have questions, feel free to ask.

ConnorL33t:
I think I have a solution based off of the previous PR (and the comments) for this issue.

I have a couple of questions: 

1. If a migration is taking a long time for some reason would this be visible to the user? (server admin)

2. Right now there's some magic numbers due to driver differences (MySQL needs a "max" val, while Postgres will use 0 appropriately), is there a place that these values (ex: MYSQL_MAX_QUERY_TIMEOUT) could be made const for added "code documentation".

3. Does the test suite deal with migrations - does it need updates for this change?

hanzei:
@wiersgallak Can you help with 1.?

@lieut-data Can you help with 2. and 3.?

lieut-data:
@ConnorL33t:

2. These const values are unlikely to be used outside of the code setting them, so I'd just declare them as unexported values in that package (even in the same file where we migrate).
3. Ideally, we update the test suite to exercise a long-ish migration somehow, or at least the newly added code, to prove that they indeed don't time out.

hanzei:
@wiersgallak Gentle reminder about the above question

wiersgallak:
[...]

PR #14421, opened 2020-04-29
Increasing and restoring execution timeout when upgrading database

rvillablanca:
#### Summary

This PR modifes the `statements execution timeout` when upgrading database to a higher version. Currently these changes apply for `mysql` and `postgres` drivers but adding support for `sqlite` should be easy.

#### Ticket Link

Solves #10327


rvillablanca:
I need to know the best way to test it because currently upgrade operations are exiting without returning an error, so maybe I should test a simple query against some table and modify the function `getExecutionTimeoutAndSetUnlimited` to accept the timeout parameter for testing purposes. WDYT @jespino @lieut-data

rvillablanca:
I added unit tests but for mysql it seems there is no such option to stop a long running query/update. I have tested manually with an sql client for both postgres and mysql, the former is working fine but for mysql it seems there is no way to timeout a query/update in execution.

The issue #10327 describes the problem is actually in mattermost with postgres instances so, do we have evidence the problem is happening with mysql ?

lieut-data:
I'm not going to be able to help review this one. @jespino, when the time comes, perhaps another reviewer from the server team?

store/sqlstore/upgrade.go
```diff
@@ -180,6 +190,60 @@ func upgradeDatabase(sqlStore SqlStore, currentModelVersionString string) error
...
+
+	setMaxExecutionTime(sqlStore, value)
+
+	return currentExecutionTimeout
+}
+
+func setMaxExecutionTime(sqlStore SqlStore, value int64) {
+	var err error
+	if sqlStore.DriverName() == model.DATABASE_DRIVER_POSTGRES {
+		_, err = sqlStore.GetMaster().Exec(fmt.Sprintf("SET STATEMENT_TIMEOUT = %d", value))
```
jespino, line 230: Probably is a better option this:
```suggestion
		_, err = sqlStore.GetMaster().Exec("SET STATEMENT_TIMEOUT = $1", value)
```

store/sqlstore/upgrade.go
```diff
@@ -180,6 +190,60 @@ func upgradeDatabase(sqlStore SqlStore, currentModelVersionString string) error
 	return nil
 }
 
+func decideMaxExecutionTime(sqlStore SqlStore) int64 {
+	result := MAX_TIMEOUT_EXECUTION_TIME_MYSQL
+	if sqlStore.DriverName() == model.DATABASE_DRIVER_POSTGRES {
+		result = MAX_TIMEOUT_EXECUTION_TIME_POSTGRES
+	}
+	return result
```
jespino, line 198: maybe a clearer and simplest option would be
```suggestion
	if sqlStore.DriverName() == model.DATABASE_DRIVER_POSTGRES {
		return MAX_TIMEOUT_EXECUTION_TIME_POSTGRES
	}
	return MAX_TIMEOUT_EXECUTION_TIME_MYSQL
```

store/sqlstore/upgrade.go
```diff
@@ -126,6 +132,10 @@ func upgradeDatabase(sqlStore SqlStore, currentModelVersionString string) error
 		mlog.Warn("The database schema version and model versions do not match", mlog.String("schema_version", currentSchemaVersion.String()), mlog.String("model_version", currentModelVersion.String()))
 	}
 
+	// max execution time to execute upgrade statements
+	currentExecutionTimeout := getAndSetMaxExecutionTime(sqlStore, decideMaxExecutionTime(sqlStore))
```
jespino, line 136: Should we fail in case we aren't able to raise the timeout? I think is more safe to just accept this and try to upgrade anyway, if there is a timeout is going to be anyway. Probably a log message would be great, but exiting the process is too much. What do you think @rvillablanca?

store/sqlstore/upgrade.go
rvillablanca: Yeah, I like it

store/sqlstore/upgrade.go
rvillablanca: Yep, I agree with you that exiting could be too much, so if I change it by adding an error log then  the intention of this PR would be an **attempt** to set the timeout. I like that approach so I will make the changes soon @jespino

rvillablanca:
> I added unit tests but for mysql it seems there is no such option to stop a long running query/update. I have tested manually with an sql client for both postgres and mysql, the former is working fine but for mysql it seems there is no way to timeout a query/update in execution.
> 
> The issue #10327 describes the problem is actually in mattermost with postgres instances so, do we have evidence the problem is happening with mysql ?

What about this @jespino ? I'm worried about this because I couldn't change the timeout in mysql. I believe the option `SET SESSION WAIT_TIMEOUT` it is not what we need

jespino:
@rvillablanca I think the `SET SESSION WAIT_TIMEOUT` is changing the timeout for the session, what it means the connection, the problem is we have a connections poll so we need to set the timeout globally, reset all the connections from the poll, then do the migrations, and finally, set again the timeout globally and reset the connections of the poll again. See here for more information about the wait_timeout: https://stackoverflow.com/questions/4440336/mysql-wait-timeout-variable-global-vs-session

About the connections reset maybe we can add a function like `ResetConnections` with `sqlStore.Close()` + `sqlStore.initConnection()`.

rvillablanca:
I made some changes according your suggestions. First of all, since we are upgrading database in different sessions I created a `Reset()` method in `SqlStore` to re-initialize all conections with the new values and this is made when we change the timeout. 

For postgres this is working as expected, I had to change the query because the previous one was affecting only to the current session.

For mysql is a different story, first I had to change the configuration parameter from `WAIT_TIMEOUT` to  `MAX_EXECUTION_TIMEOUT`.

> WAIT_TIMEOUT: The number of seconds the server waits for activity on a noninteractive connection before closing it.

> The MAX_EXECUTION_TIME hint is permitted only for SELECT statements. It places a limit N (a timeout value in milliseconds) on how long a statement is permitted to execute before the server terminates it.

So I can't use `WAIT_TIMEOUT` because it is for idle connections, and for `MAX_EXECUTION_TIME` it only affects SELECT statements, also it requires super privileges.

As I said before, the issue describes the problem is happening in mattermost with postgres instances, but mysql. Maybe we're trying to solve a problem that is not ocurring in this DB. Are we getting timeout errors while upgrading in mysql?

teresa-novoa:
@streamer45 friendly reminder to review this PR, thanks!

streamer45:
Mainly waiting for the discussion to unfold. I don't see any problem with the PR per se but not sure what problem we are trying to solve here. Apparently these timeouts are not something enabled by default nor it's suggested to do so.

As far as I can see we usually enforce query timeouts at the application level. For example, [gorp](https://github.com/mattermost/gorp/blob/master/gorp.go) does that by using `context.WithTimeout` on the dbMap. 
Is there an actual need to do this at the db level? cc. @jespino

PR #14427, opened 2020-04-30
backport

agnivade:
- MM-24395: Set response header to be attachment for SVG images. (#14385)


agnivade:
This has failed on translations because backports on release branches always test with EE master instead of the correct release branch. I have verified this locally and there are no issues.

Going ahead with merging this.

PR #14440, opened 2020-04-30
MM-24547: Fix deadlock in pingTimeoutTimer

agnivade:
The read of the timer channel happened in 2 different goroutines
which created a race, and therefore if one was read before,
it would cause the other to deadlock.

To fix this, we use yet another channel to pass the message
and thereby make the reads happen in the same goroutine.

#### Ticket link

https://mattermost.atlassian.net/browse/MM-24685

### DIFF
--- a/model/websocket_client.go
+++ b/model/websocket_client.go
@@ -271,15 +271,18 @@ func (wsc *WebSocketClient) pingHandler(appData string) error {
 }
 
 func (wsc *WebSocketClient) pingWatchdog() {
-	select {
-	case <-wsc.resetTimerChan:
-		if !wsc.pingTimeoutTimer.Stop() {
-			<-wsc.pingTimeoutTimer.C
-		}
+	for {
+		select {
+		case <-wsc.resetTimerChan:
+			if !wsc.pingTimeoutTimer.Stop() {
+				<-wsc.pingTimeoutTimer.C
+			}
 
-		wsc.pingTimeoutTimer.Reset(time.Second * (60 + PING_TIMEOUT_BUFFER_SECONDS))
-	case <-wsc.pingTimeoutTimer.C:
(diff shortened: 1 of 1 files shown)


agnivade:
Recommended to be used on top of https://github.com/mattermost/mattermost-server/pull/14406. That has some additional safety hatches which makes things more bulletproof.

model/websocket_client.go
```diff
@@ -259,21 +263,27 @@ func (wsc *WebSocketClient) pingHandler(appData string) error {
...
+	for {
+		select {
+		case <-wsc.resetTimerChan:
+			if !wsc.pingTimeoutTimer.Stop() {
+				<-wsc.pingTimeoutTimer.C
+			}
+			wsc.pingTimeoutTimer.Reset(time.Second * (60 + PING_TIMEOUT_BUFFER_SECONDS))
+
+		case <-wsc.pingTimeoutTimer.C:
+			wsc.PingTimeoutChannel <- true
```
streamer45, line 283: Just noting that this could potentially block indefinitely if `PingTimeoutChannel` is not correctly handled by the user of the API.
Whether we should adopt a "better safe than sorry" approach and wrap this in a `select` or just make the expected usage explicit through documentation, I will leave it to you.

agnivade: This is complicated as is. I don't believe somebody is using the websocket client as heavily as us. And the API itself is error-prone which should be changed in v6. Let's just document this for now.

agnivade:
@ogi-m - This falls into the same category as https://github.com/mattermost/mattermost-server/pull/14406. It fixes a deadlock in the websocket client library which makes it a bit difficult to test. But it was causing issues in our load test and we have tested it sufficiently and confirmed that it fixes the problem.

Let me know if you know of a way to test this otherwise. Thanks.

PR #14442, opened 2020-04-30
MM-24707 fix bug when bulk importing post replies

wiggin77:
#### Summary

This PR fixes a bug where replies to posts were not importing during bulk import.

Root cause is an optimization added in 5.22 which batch adds the replies.  Unfortunately the logic is in error and mismatches the replies with the posts.  Occasionally this will result in no error but the reply will appear under the wrong post.  More often the reply will generate the error above, depending on the time stamp of the last post in every 1000 post batch.  

#### Ticket Link

https://mattermost.atlassian.net/browse/MM-24707

### DIFF
--- a/app/import_functions.go
+++ b/app/import_functions.go
@@ -1267,8 +1267,6 @@ func (a *App) importMultiplePosts(data []*PostImportData, dryRun bool) *model.Ap
 		return err
 	}
 
-	var lastPostWithData *postAndData
-	repliesBulk := []ReplyImportData{}
 	for _, postWithData := range postsWithData {
 		postWithData := postWithData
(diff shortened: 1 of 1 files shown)


wiggin77:
> LGTM but I don't have any background on this code. Can someone from platform do the review in place of me? @wiggin77

Sure @mickmister , I'll assign someone from platform.  The system suggested you  - I guess you worked on the code at some point.  Thanks.

wiggin77:
> @wiggin77 will add unit tests required for the changes through a different ticket, to allow this PR to get cherrypicked faster.

Ticket created:  https://mattermost.atlassian.net/browse/MM-24738

PR #14443, opened 2020-04-30
MM-24694: Add getGroupsByUserId to API layer

sbishel:
#### Summary
Adds a new endpoint to return all groups for a user

The code for the App and Store layer already existed, only the API layer code was necessary to add.

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-24694


crspeller:
@srkgupta FYI security reviews should be coordinated by @DSchalla now :)

api4/group.go
```diff
@@ -505,6 +509,37 @@ func getGroupMembers(c *Context, w http.ResponseWriter, r *http.Request) {
 	w.Write(b)
 }
 
+func getGroupsByUserId(c *Context, w http.ResponseWriter, r *http.Request) {
+	c.RequireUserId()
+	if c.Err != nil {
+		return
+	}
+
+	if c.App.Session().UserId != c.Params.UserId && !c.App.SessionHasPermissionTo(*c.App.Session(), model.PERMISSION_MANAGE_SYSTEM) {
```
hahmadia, line 518: Shouldn't this be a `||`? We would want to deny permission if the user requesting the groups is either not the user with the UserID in the params OR is not a system admin? So only grant access if we are a system admin or the user with the userID in the params.

sbishel: No, && is correct. 
If the contextual user is NOT the parameter user AND NOT a System Admin. - Deny Access.
I have unit tests that check all three scenarios.
Its actually the same permission check in `getTeamsForUser()`

hahmadia: @sbishel Sorry, you are right. I made a mistake and didn't think carefully enough about it. Thanks for the correction.

thefactremains:
docs were added here:
https://api.mattermost.com/#tag/groups/paths/~1users~1{user_id}~1groups/get

PR #14444, opened 2020-04-30
MM-24692: Add a since parameter to getGroups api

sbishel:
#### Summary
Adding parameter 'since' to getGroups api, similar to getUsers api

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-24692

#### Related PR
https://github.com/mattermost/mattermost-api-reference/pull/518


store/sqlstore/group_store.go
```diff
@@ -1198,6 +1198,12 @@ func (s *SqlGroupStore) GetGroups(page, perPage int, opts model.GroupSearchOpts)
 		`, opts.NotAssociatedToChannel)
 	}
 
+	if opts.Since > 0 {
+		groupsQuery = groupsQuery.Where(sq.Gt(map[string]interface{}{
+			"g.UpdateAt": opts.Since,
+		}))
```
fmunshi, line 1204: I think we will also either need to include deleted groups by default when passing `Since` or add a new option to `IncludeDeleted` since we need a way for the client to know which groups to remove from the state

sbishel: Updated to return deleted groups with "since". Since it will only return those deleted "since", not all deleted groups, it made sense to do this by default.

api4/group_test.go
```diff
@@ -911,4 +912,34 @@ func TestGetGroups(t *testing.T) {
...
+	opts.Since = model.GetMillis()
+	groups, response = th.Client.GetGroups(opts)
+	assert.Nil(t, response.Error)
+	assert.Empty(t, groups)
+
+	// make sure delete group is not returned without Since
+	opts.Since = 0
+	groups, response = th.Client.GetGroups(opts)
+	assert.Nil(t, response.Error)
+	//'Normal getGroups should not return delete groups
```
fmunshi, line 941:
```suggestion
	// Normal getGroups should not return delete groups
```

hahmadia:
Hey Scott! Looks great!
Just wanted to confirm something. When I pass in a since prameter, I'm expecting to receive all the groups that have been created, updated, and deleted since that time period. Is that what this returns? I'm just a little confused cause I see that it only checks against the `updatedAt` field but not `createdAt` or `deletedAt` so I just wanted to confirm before approving. Thanks!

sbishel:
Yep. When a group is either created or deleted, both update the `updatedAt` field. So that is the only one we need to check.  A great question and great thing to double check!
https://github.com/mattermost/mattermost-server/blob/master/store/sqlstore/group_store.go#L104
https://github.com/mattermost/mattermost-server/blob/master/store/sqlstore/group_store.go#L239

store/sqlstore/group_store.go
```diff
@@ -1144,9 +1144,16 @@ func (s *SqlGroupStore) GetGroups(page, perPage int, opts model.GroupSearchOpts)
 
 	groupsQuery = groupsQuery.
 		From("UserGroups g").
-		Where("g.DeleteAt = 0").
 		OrderBy("g.DisplayName")
 
+	if opts.Since > 0 {
+		groupsQuery = groupsQuery.Where(sq.Gt{
```
hahmadia, line 1150: Just a small relization! We may want to use `sq.GtOrEq` since `sq.Gt` returns all the rows that have a `updateAt` greater than the since but not equal too it!

sbishel:
@srkgupta, please see @fmunshi's comment above. We need to include deleted items in the returns because those items deleted "since" the last check will need removed from state.

Because you are using a "since" parameter of 1. It is returning everything because everything has been updated "since" then. But normally, the "since" parameter would be from perhaps a couple hours ago. If a group were deleted in the last couple hours, we need to return it.

store/sqlstore/group_store.go
sbishel: @hahmadia The other two places we have a "since" parameter, POSTS and USERS. They use GT. This was the pattern I was following. It depends on how the "since" value is created and stored. It can be done either way, but I would want a good reason to NOT follow the same pattern as POSTS and USERS.

PR #14445, opened 2020-05-01
MM-24484 allow base64 certs in config for TLS Syslog

wiggin77:
#### Summary

This PR allows cloud team to use base64 encoded certs in place of filespec for TLS cert in audit config for syslog.  The config property `ExperimentalAuditSettings.SysLogCert` can now specify a path to pem file or base64 encoded string containing the cert(s).

A unit test was written using a sample cert provided by @corey-robinson .  A new ticket will be generated to replace this cert with something that won't expire.

The package used to write to syslog via TLS has also been forked and the code now points to the fork.  Original package https://github.com/RackSec/srslog is now archived.  The fork was needed to add the ability to redial when the socket breaks for whatever reason.

The audit to syslog via TLS is an experimental feature which will only be used by cloud team until general release.

#### Ticket Link

https://mattermost.atlassian.net/browse/MM-24484   and   https://mattermost.atlassian.net/browse/MM-24416


audit/syslogtls.go
```diff
@@ -32,16 +33,7 @@ type SyslogParams struct {
...
-	if params.Cert != "" {
-		serverCert, err := ioutil.ReadFile(params.Cert)
-		if err != nil {
-			return nil, err
-		}
-		pool := x509.NewCertPool()
-		pool.AppendCertsFromPEM(serverCert)
-		config.RootCAs = pool
-	}
+	config.RootCAs = getCertPool(params.Cert)
```
lieut-data, line 36: I don't have context into who is using this, but is it feasible to distinguish a base-64 encoded certificate from a file as a separate params property instead of trying to test both possibilities?

wiggin77: Thanks for reviewing @lieut-data .  This is how cloud team asked for it, and as I recall is in line with other fields they use the same way.  @corey-robinson any thoughts on this?

On thing to consider is there would be extra logic needed for separate fields, such as erroring if both are used.

wiggin77:
> but also a general ask that we might review the fork changes as well before merging.

Going forward I will do any changes to deps as a PR.  This one was simply a commit after fork.  Commit is here:  https://github.com/wiggin77/srslog/commit/9e4fac0bf76e3aeaac11c81276f8d1f900cf3519

wiggin77:
Hi @lindy65 , this PR is a add-on to https://github.com/mattermost/mattermost-server/pull/13967 for the still experimental audit system.  QA passed on testing the first PR due to this only being used by cloud team for now.  Cloud team asked for a change in how certs are specified in the config as they are just starting to use the new auditing system.  Not sure if you will want to test or defer.  If you want to to test this I'll get a cert from cloud team to test with.  Thx.

wiggin77:
Thanks @lindy65 

> One question... Will someone on Cloud team be testing after this code is merged?

Yes.  Apparently you, via this ticket https://mattermost.atlassian.net/browse/MM-24416   :-)

lindy65:
Oh, ha ha @wiggin77 - I just added my name as QA to the two tickets on this PR so I can keep track of them post-merge ;)  I usually then check with a dev on the team who's going to test just to make sure things keep moving and nothing falls through the cracks...

lieut-data:
Thanks, @wiggin77 -- I've added some comments directly to https://github.com/wiggin77/srslog/commit/9e4fac0bf76e3aeaac11c81276f8d1f900cf3519.

audit/syslogtls.go
lieut-data: Ah, that makes sense, and seems reasonable. Thanks for the clarity, @wiggin77!

wiggin77:
> Thanks, @wiggin77 -- I've added some comments directly to [wiggin77/srslog@9e4fac0](https://github.com/wiggin77/srslog/commit/9e4fac0bf76e3aeaac11c81276f8d1f900cf3519).

Great review @lieut-data .  I've incorporated your suggestions here https://github.com/wiggin77/srslog/commit/5f255689df7a57cb0b72c28a9742363fa34a966b  and also replaced the expired unit test certs plus fixed a unit test.

audit/syslogtls.go
```diff
@@ -67,6 +59,31 @@ func (s *SyslogTLS) Shutdown(ctx context.Context) error {
 	return errs.ErrorOrNil()
 }
 
+// getCertPool returns a x509.CertPool containing the cert(s)
+// from `cert`, which can be a path to a .pem or .crt file,
+// or a base64 encoded cert.
+func getCertPool(cert string) *x509.CertPool {
```
mgdelacroix, line 65: Is it safe to just return `nil` on error? Shouldn't we return the error itself and manage it as part of `NewSyslogTLSTarget`?

wiggin77: Thanks @mgdelacroix 

The underlying Go API's for creating the pool have no specific error conditions so I cannot add any context other than worked or didn't.  Also, the TLS config expects a nil cert pool if one cannot be provided and it will try to connect without it so any error returned would be ignored.

I guess the underlying question is whether we should allow an invalid cert entry to still connect.  I could add a log warning about the invalid cert so it doesn't silently fail.  I'm not so sure about not having auditing for that server instance when it could have connected.  If it cannot connect because it really does need the cert then it will error in the logs.

Would a log warning suffice to address your concern?

wiggin77: Update:  I decide to error on invalid cert.  If you specify a cert, most likely you're going to need one.

mgdelacroix: You read my mind there, my concern was to successfully perform the connection without the explicitly configured cert, I think we should rather fail in those cases

wiggin77: Good call.

PR #14449, opened 2020-05-01
[MM-24700] Add api endpoint to GetGroupsByUserIds

fmunshi:
#### Summary
- Adds an endpoint to batch get all groups for a set of users to be used in the team and channel members block in the system console upon page load an action to load the groups for all the members loaded on the page will be dispatched. 

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-24700


fmunshi:
Closing this as per discussion with UX Review we will no longer be needing `Groups` as a column on the team / channel members list - If it is decided that we in fact _do_ want it I can re-open this.

PR #14450, opened 2020-05-02
MM-24759: fix order of calling SetstatusOnline

agnivade:
#### Summary

We would call SetStatusOnline in a goroutine before actually calling HubRegister.
This could cause the message not to be sent after all, because there's no guarantee
that Register would actually happen before it.

To fix it, we just change the order of things.

#### Ticket Link

https://mattermost.atlassian.net/browse/MM-24759


app/websocket_router.go
```diff
@@ -54,17 +54,17 @@ func (wr *WebSocketRouter) ServeWebSocket(conn *WebConn, r *model.WebSocketReque
...
 		conn.SetSession(session)
 		conn.SetSessionToken(session.Token)
 		conn.UserId = session.UserId
 
 		wr.app.HubRegister(conn)
 
+		wr.app.Srv().Go(func() {
+			wr.app.SetStatusOnline(session.UserId, false)
+			wr.app.UpdateLastActivityAtIfNeeded(*session)
+		})
```
streamer45, line 66: If I understood the problem correctly, there's a race between sending the status change message and registering the connection in the hub. If that's the case I am not sure how changing the order here can fully correct the issue.

As far as I can see we are still lacking any guarantee that the connection is going to get registered before we try to send the status change message.

agnivade: And that's why I set the "Do not merge" label. :wink:

streamer45: Yes, just though it was ready for review. I am guessing https://github.com/mattermost/mattermost-server/pull/14399 is the other piece of the puzzle.

agnivade: Yes, it _is_ ready for review. Considering #14399 is applied. Sorry I should have made that clear :P

streamer45: No worries :+1: , just making sure we are not missing anything. I could have guessed but @isacikgoz was totally unaware of that one and I fear he might have not been presented with the full picture.

PR #14463, opened 2020-05-04
ES now supports inactive users filtering

ethervoid:
#### Summary

Adding the delete_at let us filter users based in the logical deletion

#### Ticket Link

https://mattermost.atlassian.net/browse/MM-24333


metanerd:
Tests are failing. I am not sure why the trap isn't sprung.

ethervoid:
@metanerd look like we're running the tests using the `master` branch instead of this one `MM-24333`, as you can see in the details:

Screenshot from 2020-05-05 13-12-34

In the test env variables, you can see is using the master branch as well:

Screenshot from 2020-05-05 13-13-24

Also, I've seen some problems mixing other branches (that branch was pointing to this one):

Screenshot from 2020-05-05 13-15-22

This branch is `MM-24333` and is doing the checkout for `MM-24334`

metanerd:
This is correct. I will skip the tests for those branches in the future.

wiggin77:
Hi @ethervoid. Is this related to another PR?  I'm only seeing the unit tests here.

ethervoid:
@wiggin77 Yep, you're on that as well. They have the same branch name

PR #14471, opened 2020-05-05
MM-24805: websocket client can send after writer has closed

agnivade:
If the server closes the connection, and the client is trying to
write something from another goroutine, then it is possible
that it can try to write to the writeChan after it was closed.

To fix this we don't actually close the writeChan and just exit the writer.
And we do a non-blocking send from WriteMessage to just move forward
if the writer has exited

#### Ticket link

https://mattermost.atlassian.net/browse/MM-24805

### DIFF
--- a/model/websocket_client.go
+++ b/model/websocket_client.go
@@ -149,7 +149,6 @@ func (wsc *WebSocketClient) Close() {
 		return
 	}
 	wsc.quitWriterChan <- struct{}{}
-	close(wsc.writeChan)
 	// We close the connection, which breaks the reader loop.
 	// Then we let the defer block in the reader do further cleanup.
 	wsc.Conn.Close()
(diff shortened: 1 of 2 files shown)


model/websocket_client.go
```diff
@@ -249,10 +247,15 @@ func (wsc *WebSocketClient) SendMessage(action string, data map[string]interface
 	req.Data = data
 
 	wsc.Sequence++
-	wsc.writeChan <- writeMessage{
+	// we do a non-blocking send in case the writer has already exited.
+	select {
+	case wsc.writeChan <- writeMessage{
```
streamer45, line 252: Should we make `writeChan` buffered? My worry with this is that we might end up dropping messages during normal operations if the `writer` goroutine takes a bit longer to handle the message.

agnivade:
Closing this as this is buggy and we anyways decided to stop using this client.

PR #14475, opened 2020-05-05
[MM-24827] Disable group mentions on e10 or team edition

fmunshi:
#### Summary
- Disable all group mentions logic in the SendNotifications function when ldap groups license is present

#### Ticket Link
- https://mattermost.atlassian.net/browse/MM-24827


app/notification.go
```diff
@@ -740,6 +740,10 @@ func (a *App) allowChannelMentions(post *model.Post, numProfiles int) bool {
 
 // allowGroupMentions returns whether or not the group mentions are allowed for the given post.
 func (a *App) allowGroupMentions(post *model.Post) bool {
+	if license := a.License(); license == nil || !*license.Features.LDAPGroups {
+		return false
+	}
+
 	if !a.HasPermissionToChannel(post.UserId, post.ChannelId, model.PERMISSION_USE_GROUP_MENTIONS) {
```
hahmadia, line 747: Could we do this or do we need to do it in 2 different if statements?
```suggestion
	if !a.HasPermissionToChannel(post.UserId, post.ChannelId, model.PERMISSION_USE_GROUP_MENTIONS) || license == nil || !*license.Features.LDAPGroups {
```

fmunshi: I think it just looks a lot more readable in multiple if statements

fmunshi: Its also more consistent with the code style in the rest of the file for eg: lines 725-740
```
// allowChannelMentions returns whether or not the channel mentions are allowed for the given post.
func (a *App) allowChannelMentions(post *model.Post, numProfiles int) bool {
	if !a.HasPermissionToChannel(post.UserId, post.ChannelId, model.PERMISSION_USE_CHANNEL_MENTIONS) {
		return false
	}

	if post.Type == model.POST_HEADER_CHANGE || post.Type == model.POST_PURPOSE_CHANGE {
		return false
	}

	if int64(numProfiles) >= *a.Config().TeamSettings.MaxNotificationsPerChannel {
		return false
	}

	return true
}
```

hahmadia: Didn't notice that. In that case, lets follow the pattern already in place. 👍

PR #14477, opened 2020-05-05
[MM-24828] Take into account group synced state of team and channel when sending notifications

fmunshi:
#### Summary
- When a group is mentioned that is not tied to a group constrained team or channel users should not be notified for that group, this fixes that by limiting the groups queried to only ones that can be notified in that channel / team.

#### Example:
- the "marketing" group has mentions enabled
- the channel is constrained to "developers"
- a user is in both
- the "@ marketing" mention us used in the developer constrained channel
- the user should not receive notifications for the @ marketing message

#### Ticket Link
- https://mattermost.atlassian.net/browse/MM-24828


app/notification.go
```diff
@@ -751,6 +745,39 @@ func (a *App) allowGroupMentions(post *model.Post) bool {
...
+			groups, err = a.Srv().Store.Group().GetGroupsByTeam(team.Id, opts)
+		}
+		if err != nil {
+			return nil, err
+		}
+		for _, group := range groups {
+			groupsMap[group.Group.Name] = &group.Group
+		}
+		return groupsMap, nil
+	}
```
hahmadia, line 768:
```suggestion
	} else {
	   groups, err := a.Srv().Store.Group().GetGroups(0, 0, opts)
       }
```

app/notification.go
```diff
@@ -751,6 +745,39 @@ func (a *App) allowGroupMentions(post *model.Post) bool {
 	return true
 }
 
+// getGroupsAllowedForReferenceInChannel returns a map of groups allowed for reference in a given channel and team.
+func (a *App) getGroupsAllowedForReferenceInChannel(channel *model.Channel, team *model.Team) (map[string]*model.Group, *model.AppError) {
+	var err *model.AppError
+	groupsMap := make(map[string]*model.Group)
+	opts := model.GroupSearchOpts{FilterAllowReference: true}
+
```
hahmadia, line 753:
```suggestion
     var groups []*model.GroupWithSchemeAdmin
```

app/notification.go
```diff
@@ -751,6 +745,39 @@ func (a *App) allowGroupMentions(post *model.Post) bool {
...
+	if channel.IsGroupConstrained() || team.IsGroupConstrained() {
+		var groups []*model.GroupWithSchemeAdmin
```
hahmadia, line 755:
```suggestion
```

app/notification.go
```diff
@@ -751,6 +745,39 @@ func (a *App) allowGroupMentions(post *model.Post) bool {
...
+		} else {
+			groups, err = a.Srv().Store.Group().GetGroupsByTeam(team.Id, opts)
+		}
+		if err != nil {
+			return nil, err
+		}
+		for _, group := range groups {
+			groupsMap[group.Group.Name] = &group.Group
+		}
+		return groupsMap, nil
```
hahmadia, line 767: We can remove lines 761 to 767

app/notification.go
```diff
@@ -751,6 +745,39 @@ func (a *App) allowGroupMentions(post *model.Post) bool {
...
+	}
+
+	groups, err := a.Srv().Store.Group().GetGroups(0, 0, opts)
```
hahmadia, line 770:
```suggestion
```

fmunshi:
bump @catalintomai for dev review

app/notification_test.go
```diff
@@ -2410,3 +2410,106 @@ func TestInsertGroupMentions(t *testing.T) {
...
+		require.Nil(t, groupsMap[group3.Name])
+		require.Nil(t, groupsMap[group2.Name])
+		require.Equal(t, groupsMap[group1.Name], group1)
+	})
+
+	team.GroupConstrained = model.NewBool(false)
+	team, err = th.App.UpdateTeam(team)
+	require.Nil(t, err)
+
+	t.Run("should return all groups when team is not group constrained but channel that is group constrained exists on the team", func(t *testing.T) {
```
catalintomai, line 2507: why the "but channel that is group constrained exists on the team" ? How is it different from the second test case "should return only groups synced to channel if channel is group constrained" ?

fmunshi: In this test case there is a channel on the team that is group constrained, but the channel that we're passing as a parameter is different from the group constrained one

fmunshi: Do you think this test is not needed?

catalintomai: I guess the question is: does the "when team is not group constrained but channel that is group constrained exists on the team" matters in this case - or maybe we can rephrase as - "current channel is not group mentioned and the team it is part of is not group constrained either". Not clear if the the fact that there might be one or more channels group-constrained in the team matters.

fmunshi, line 2507:
```suggestion
	t.Run("should return all groups when team is and channel are not group constrained", func(t *testing.T) {
```

fmunshi: Done @catalintomai I have made the test description clearer

srkgupta:
@fmunshi can you please fix the circleci build errors

PR #14479, opened 2020-05-05
MM-23250: improve config.Migrate tests

lieut-data:
#### Summary
The config.Migrate tests were not running in isolation, but wrote over the "local" config. The test coverage for both directions was not exhaustive, and lacked coverage for the recently added plugin signature files.

Rewrite the unit tests as such, fixing a minor issue closing the stores after migration.

#### Ticket Link
Fixes: https://mattermost.atlassian.net/browse/MM-23250
Fixes: https://mattermost.atlassian.net/browse/MM-24823


config/migrate_test.go
```diff
@@ -1,80 +1,134 @@
...
 	}
-	return source
-}
-
-func TestMigrateDatabaseToFile(t *testing.T) {
-	if testing.Short() {
-		t.SkipNow()
+	filesData := make([]string, len(files))
+	for i := range files {
+		filesData[i] = model.NewId()
```
agarciamontoro, line 28: I understand that the `NewId()` call is just a way of creating random data for each file (please correct me if I'm wrong). If that's right, maybe we could add a comment stating that to make it clear.

config/migrate_test.go
```diff
@@ -1,80 +1,134 @@
...
-	config.SamlSettings.PrivateKeyFile = &files[2]
-	_, err = ds.Set(config)
-	require.NoError(t, err)
-
-	for _, file := range files {
-		err = ds.SetFile(file, data)
+
+	setup := func(t *testing.T) {
+		os.Clearenv()
+		t.Helper()
```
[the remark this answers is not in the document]
lieut-data: This is actually to prevent any developer-overridden `MM_*` settings from influencing the tests, since the `config` package consumes settings like `MM_SERVICESETTINGS_SITEURL=http://example.com` and such.

config/migrate_test.go
lieut-data: Yes, that's it! I'll clarify with a comment -- thanks :)

lieut-data:
Thanks for the feedback, @agarciamontoro! Changes applied :)

prapti:
@lieut-data I'm not familiar with config.Migrate. However I ran smoke tests, and ran 
`$ go test -v -run='Migrate' ./config` which ran successfully. Was this the correct test to run to verify the PR changes?

lieut-data:
Hey @prapti! No need to run the `go test` command -- that's verified by CI automatically. There's really very little outside of tests here to verify, but for completeness, I'd suggest using the mattermost CLI to verify the `config migrate` functionality to/from the database.

PR #14481, opened 2020-05-05
[GH-14434] fix search for emails in Posts in databases

fedealconada:
#### Summary
This PR fixes search for emails in Posts in databases

#### Ticket Link
Fixes https://github.com/mattermost/mattermost-server/issues/14434

### DIFF
--- a/store/sqlstore/post_store.go
+++ b/store/sqlstore/post_store.go
@@ -1206,6 +1206,16 @@ func (s *SqlPostStore) search(teamId string, userId string, params *model.Search
 		}
 	}
 
+	terms, containsEmails := wrapEmails(terms)
+	if containsEmails {
+		// Find and remove "@"
+		for i, v := range specialSearchChar {
+			if v == "@" {
+				specialSearchChar = append(specialSearchChar[:i], specialSearchChar[i+1:]...)
+				break
+			}
+		}
+	}
 	// these chars have special meaning and can be treated as spaces
 	for _, c := range specialSearchChar {
 		terms = strings.Replace(terms, c, " ", -1)
@@ -1779,4 +1789,35 @@ func (s *SqlPostStore) SearchPostsInTeamForUser(paramsList []*model.SearchParams
 	posts.SortByCreateAt()
 
 	return model.MakePostSearchResults(posts, nil), nil
+}
+
+func wrapEmails(terms string) (string, bool) {
+	emailRegex := regexp.MustCompile("^[a-zA-Z0-9.!#$%&'*+/=?^_`{|}~-]+@[a-zA-Z0-9](?:[a-zA-Z0-9-]{0,61}[a-zA-Z0-9])?(?:\\.[a-zA-Z0-9](?:[a-zA-Z0-9-]{0,61}[a-zA-Z0-9])?)*$")
+	var termsWithWrappedEmails []string
+	containsEmails := false
+	// remove quotes if it has
+	termsWithoutQuotes := terms
+	if len(termsWithoutQuotes) > 0 && termsWithoutQuotes[0] == '"' {
+		termsWithoutQuotes = termsWithoutQuotes[1:]
+	}
+	if len(termsWithoutQuotes) > 0 && termsWithoutQuotes[len(termsWithoutQuotes)-1] == '"' {
+		termsWithoutQuotes = termsWithoutQuotes[:len(termsWithoutQuotes)-1]
+	}
+
+	// split terms
+	s := strings.Split(termsWithoutQuotes, " ")
+	for i := range s {
+		if emailRegex.MatchString(s[i]) {
+			// wrap email with quotes
+			termsWithWrappedEmails = append(termsWithWrappedEmails, "\"" + s[i] + "\"")
+			containsEmails = true
+		} else {
+			termsWithWrappedEmails = append(termsWithWrappedEmails, s[i])
+		}
+	}
+	if !containsEmails {
+		return terms, false
+	}
+	return strings.Join(termsWithWrappedEmails[:], " "), containsEmails
+
 }
--- a/store/searchtest/post_layer.go
+++ b/store/searchtest/post_layer.go
@@ -31,7 +31,7 @@ var searchPostStoreTests = []searchTest{
 	{
 		Name: "Should be able to search for email addresses with or without quotes",
 		Fn:   testSearchEmailAddresses,
-		Tags: []string{ENGINE_ELASTICSEARCH},
+		Tags: []string{ENGINE_ALL},
 	},
 	{
 		Name: "Should be able to search when markdown underscores are applied",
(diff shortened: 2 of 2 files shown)


fedealconada:
I've tested this both on postgres and mysql and seems to fix it

store/sqlstore/post_store.go
```diff
@@ -1206,10 +1206,8 @@ func (s *SqlPostStore) search(teamId string, userId string, params *model.Search
 		}
 	}
 
-	// these chars have special meaning and can be treated as spaces
-	for _, c := range specialSearchChar {
```
fedealconada, line 1210: I'm not really sure why do you need to replace special characters with spaces. So, I'm basically removing this and just, when there's a special character in the terms, I wrap the term with quotes. Thoughts?

fedealconada: I believe that removing this will also fix #14437

ethervoid:
@fedealconada Could you please share the test you're doing just to know what kind of situations are you taking into account. 

Also, we have tests for this part that are deactivated so please go [here](https://github.com/mattermost/mattermost-server/blob/master/store/searchtest/post_layer.go#L34) and change `ENGINE_ELASTICSEARCH` for `ENGINE_ALL` and launch those tests using the following command:

```bash
go test ./store/sqlstore -mod=vendor -run=TestSearchPostStore
```

fedealconada:
@ethervoid Oh! I didn't know there were tests :) I just took at look at post_store_test.go and thought there were no tests. They way I previously "tested" was just via the frontend and searching an email.
I've just pushed a change and run the tests you mentioned and they pass...

store/sqlstore/post_store.go
ethervoid: The special characters part is a way to tokenize manually removing reserved characters both for postgres and mysql.

store/sqlstore/post_store.go
```diff
@@ -1205,11 +1205,9 @@ func (s *SqlPostStore) search(teamId string, userId string, params *model.Search
 			termMap[strings.ToUpper(term)] = true
 		}
 	}
-
-	// these chars have special meaning and can be treated as spaces
-	for _, c := range specialSearchChar {
-		terms = strings.Replace(terms, c, " ", -1)
-		excludedTerms = strings.Replace(excludedTerms, c, " ", -1)
+	wrappedByQuotes := strings.HasPrefix(terms, "\"") && strings.HasSuffix(terms, "\"")
+	if  !wrappedByQuotes &&  strings.ContainsAny(terms, strings.Join(specialSearchChar, "")) {
```
ethervoid, line 1209: You've removed the `excludedTerms` code from here so you should re-add it.

teresa-novoa:
@mgdelacroix friendly reminder to review this PR, thanks!

fedealconada:
@teresa-novoa, I still have to do some changes to this, sorry for the delay!

teresa-novoa:
Thanks @fedealconada!

fedealconada:
Following your suggestion,
> From my point of view, we should take a different approach and try to fix just once thing at a time. In this case, we're trying to solve the email problem so we need to find a way to be able to pass the entire term to the database and perform an accurate search

I've added some changes... let me know what do you think. I'm not 100% convinced, but I think it works. The idea is that, whenever the terms contain one or more email addresses, we should wrap them with quotes.

store/sqlstore/post_store.go
```diff
@@ -1780,3 +1790,34 @@ func (s *SqlPostStore) SearchPostsInTeamForUser(paramsList []*model.SearchParams
 
 	return model.MakePostSearchResults(posts, nil), nil
 }
+
+func wrapEmails(terms string) (string, bool) {
+	emailRegex := regexp.MustCompile("^[a-zA-Z0-9.!#$%&'*+/=?^_`{|}~-]+@[a-zA-Z0-9](?:[a-zA-Z0-9-]{0,61}[a-zA-Z0-9])?(?:\\.[a-zA-Z0-9](?:[a-zA-Z0-9-]{0,61}[a-zA-Z0-9])?)*$")
```
ethervoid, line 1795: No need to have a perfect regexp here, with the risk that implies, we are using a simpler one [here](https://github.com/mattermost/mattermost-server/blob/master/services/searchengine/utils.go#L13). You can use something similar

store/sqlstore/post_store.go
```diff
@@ -1780,3 +1790,34 @@ func (s *SqlPostStore) SearchPostsInTeamForUser(paramsList []*model.SearchParams
 
 	return model.MakePostSearchResults(posts, nil), nil
 }
+
+func wrapEmails(terms string) (string, bool) {
```
ethervoid, line 1794: Probably we want to rename this to something like `quoteEmailTerms`.

Also, I'd move this to [here](https://github.com/mattermost/mattermost-server/blob/master/store/sqlstore/utils.go)

store/sqlstore/post_store.go
```diff
@@ -1780,3 +1790,34 @@ func (s *SqlPostStore) SearchPostsInTeamForUser(paramsList []*model.SearchParams
...
+	emailRegex := regexp.MustCompile("^[a-zA-Z0-9.!#$%&'*+/=?^_`{|}~-]+@[a-zA-Z0-9](?:[a-zA-Z0-9-]{0,61}[a-zA-Z0-9])?(?:\\.[a-zA-Z0-9](?:[a-zA-Z0-9-]{0,61}[a-zA-Z0-9])?)*$")
+	var termsWithWrappedEmails []string
+	containsEmails := false
+	// remove quotes if it has
+	termsWithoutQuotes := terms
+	if len(termsWithoutQuotes) > 0 && termsWithoutQuotes[0] == '"' {
```
ethervoid, line 1800: Here are you checking for quotes for the whole string and for that you can use `strings.Trim`. With the simpler regexp, you can check for quoted emails without removing the quotes

store/sqlstore/post_store.go
```diff
@@ -1780,3 +1790,34 @@ func (s *SqlPostStore) SearchPostsInTeamForUser(paramsList []*model.SearchParams
...
+		termsWithoutQuotes = termsWithoutQuotes[1:]
+	}
+	if len(termsWithoutQuotes) > 0 && termsWithoutQuotes[len(termsWithoutQuotes)-1] == '"' {
+		termsWithoutQuotes = termsWithoutQuotes[:len(termsWithoutQuotes)-1]
+	}
+
+	// split terms
+	s := strings.Split(termsWithoutQuotes, " ")
+	for i := range s {
```
ethervoid, line 1809: Given that split doesn't return an error element you can do:
```go
for term := range strings.Split(termsWithoutQuotes, " ") {
}
```

store/sqlstore/post_store.go
```diff
@@ -1206,6 +1206,16 @@ func (s *SqlPostStore) search(teamId string, userId string, params *model.Search
 		}
 	}
 
+	terms, containsEmails := wrapEmails(terms)
+	if containsEmails {
+		// Find and remove "@"
+		for i, v := range specialSearchChar {
+			if v == "@" {
+				specialSearchChar = append(specialSearchChar[:i], specialSearchChar[i+1:]...)
```
ethervoid, line 1214: This will also remove that character for other terms where we may want to replace it, non-email terms for example

jasonblais:
Hi @fedealconada thanks for your contribution! Let us know if you have any questions about the feedback from @ethervoid above. Happy to help!

jfrerich:
Hi @fedealconada! I hope you are doing well. we would love to push merge your PR.  Do you need any assistance addressing the feedback form @ethervoid?

jasonblais:
Hi @fedealconada and big thanks for your pull request!

We haven't received an update for more than 60 days and will close this PR so someone else from the community can take it on.

If you're still interested and have questions, let us know! We'd be happy to help. :)

cc @jfrerich

PR #14483, opened 2020-05-06
Add metric warning support (announcement bar and DM)

catalintomai:
#### Summary
Add sysadmin warning support (announcement bar and DM) for multiple metrics (here - if the number of active users metric is above a certain limit)

TODO:
- add OpenAPI entries
- add server telemetry support

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-25042

#### Related PRs
redux: https://github.com/mattermost/mattermost-redux/pull/1143
webapp: https://github.com/mattermost/mattermost-webapp/pull/5447


app/server.go
```diff
@@ -800,6 +804,13 @@ func runLicenseExpirationCheckJob(s *Server) {
 	}, time.Hour*24)
 }
 
+func runStoreAndCheckNumberOfActiveUsersWarnMetricStatusJob(s *Server) {
+	doStoreAndCheckNumberOfActiveUsersWarnMetricStatus(s)
+	model.CreateRecurringTask("Store and Check Number Of Active Users Warn Metric Status", func() {
+		doStoreAndCheckNumberOfActiveUsersWarnMetricStatus(s)
+	}, time.Minute*1)
```
catalintomai, line 811: To change before final merge.

model/system.go
```diff
@@ -10,17 +10,20 @@ import (
...
+	SYSTEM_ACTIVE_LICENSE_ID                  = "ActiveLicenseId"
+	SYSTEM_LAST_COMPLIANCE_TIME               = "LastComplianceTime"
+	SYSTEM_ASYMMETRIC_SIGNING_KEY             = "AsymmetricSigningKey"
+	SYSTEM_POST_ACTION_COOKIE_SECRET          = "PostActionCookieSecret"
+	SYSTEM_INSTALLATION_DATE_KEY              = "InstallationDate"
+	SYSTEM_FIRST_SERVER_RUN_TIMESTAMP_KEY     = "FirstServerRunTimestamp"
+	SYSTEM_NUMBER_OF_ACTIVE_USERS_WARN_METRIC = "NumberOfActiveUsersWarnMetric"
 )
 
+const NUMBER_OF_ACTIVE_USERS_WARN_METRIC_LIMIT = 1
```
catalintomai, line 25: To change before final merge.

app/analytics.go
```diff
@@ -40,6 +40,7 @@ func (a *App) GetAnalytics(name string, teamId string) (model.AnalyticsRows, *mo
 		rows[8] = &model.AnalyticsRow{Name: "daily_active_users", Value: 0}
 		rows[9] = &model.AnalyticsRow{Name: "monthly_active_users", Value: 0}
 		rows[10] = &model.AnalyticsRow{Name: "inactive_user_count", Value: 0}
+		rows[11] = &model.AnalyticsRow{Name: "registered_users", Value: 0}
```
coreyhulen, line 43: Feels like a dup, isn't this already on the admin console page?

app/server.go
```diff
@@ -827,6 +838,59 @@ func doSessionCleanup(s *Server) {
...
+	mlog.Info("Number of active users", mlog.Int64("value", noActiveUsers))
+
+	if err = s.Store.System().SaveOrUpdate(&model.System{Name: model.SYSTEM_NUMBER_OF_ACTIVE_USERS_WARN_METRIC, Value: strconv.FormatInt(noActiveUsers, 10)}); err != nil {
+		mlog.Error("Unable to write to database.", mlog.Err(err))
+		return
+	}
+
+	if noActiveUsers > model.NUMBER_OF_ACTIVE_USERS_WARN_METRIC_LIMIT {
+		mlog.Info("Number of active users is greater than limit")
+		message := model.NewWebSocketEvent(model.WEBSOCKET_WARN_METRICS_STATUS, "", "", "", nil)
```
coreyhulen, line 883: Also certain why we need to send a websocket event, isn't posting the message good enough?

app/server.go
```diff
@@ -827,6 +838,59 @@ func doSessionCleanup(s *Server) {
 	s.Store.Session().Cleanup(model.GetMillis(), SESSIONS_CLEANUP_BATCH_SIZE)
 }
 
+func doStoreAndCheckNumberOfActiveUsersWarnMetricStatus(s *Server) {
+	// uncomment this if we want to have this check only for TE
+	// license := s.License()
+	// if license != nil {
+	// 	mlog.Debug("License is present, skip this check")
+	// 	return
+	// }
```
coreyhulen, line 847: We should only warn for unlicensed correct?

app/server.go
coreyhulen: What are we setting it to?  24 hrs?

app/server.go
coreyhulen: Sorry, I think I'm getting this confused with the other ticket.  Guessing this is needed for the banner bar to pop up.

app/analytics.go
catalintomai: @coreyhulen , when we send the telemetry, the "registered_users"/usercount IN TRACK_ACTIVITY count maps to :
```if count, err := a.Srv().Store.User().Count(model.UserCountOptions{IncludeDeleted: true}); err == nil {
		userCount = count
	}``` 
That is the definition I used, and I didn't see it the same for the admin console page display, but it could we get to the same number through some other query. I'll double-check.

File: app/server.go (hunk 4 above)
Comment (catalintomai) [reply app/server.go:883]: yes, both bar and DM are handled here.

File: app/server.go (hunk 5 above)
Comment (catalintomai) [reply app/server.go:847]: yes, I kept it as a reminder, since we might use the check for other metrics.

File: app/server.go (hunk 1 above)
Comment (catalintomai) [reply app/server.go:811]: That is the plan.

File: app/analytics.go (hunk 3 above)
Comment (coreyhulen) [reply app/analytics.go:43]: Hmm, it should be the metric on the system stats named `Total Active Users`. That's what we use to report for E10/E20 licenses.  It doesn't count deleted (I believe) and that's OK.

Comment (catalintomai) [reply app/analytics.go:43]: @coreyhulen , to be clear, the "registered_users" metric value here is extracted to be displayed in the message we send to acknowledge@.., not the one we use to determine to send/not-send the warning. 
In https://www.notion.so/Nag-ware-when-over-100-users-008505849eae45a98ae29e53a32b296a , we wanted to display: "Registered Users: XXXX", so I used the "registered_users" definition from the telemetry (the one with {`IncludeDeleted: true`}). 
I'll revert this part and use the `unique_user_count` value for the "Registered Users: XXXX" but then maybe we should then either 
1) change the text in the mail from "Registered Users XXXX" to smth. like "Active Users XXX" or 
2) change the "registered_users" from TRACK_ACTIVITY to smth. else (like `userCount` maybe).

File: api4/system.go [hunk 6]
```diff
@@ -545,3 +548,43 @@ func getServerBusyExpires(c *Context, w http.ResponseWriter, r *http.Request) {
...
+	}
+
+	w.Write([]byte(model.MapBoolToJson(status)))
+}
+
+func sendWarnMetricAckEmail(c *Context, w http.ResponseWriter, r *http.Request) {
+	auditRec := c.MakeAuditRecord("sendWarnMetricAckEmail", audit.Fail)
+	defer c.LogAuditRec(auditRec)
+
+	if !c.IsSystemAdmin() {
```
jespino, line 571: Why are checking here the `IsSystemAdmin` instead of the `PERMISSION_MANAGE_SYSTEM` (the effect 99.99% of the time is going to be the same, and we prefer to avoid personalizing in roles the right to do actions).

app/analytics.go
```diff
@@ -103,6 +103,13 @@ func (a *App) GetAnalytics(name string, teamId string) (model.AnalyticsRows, *mo
 			close(monthlyActiveChan)
 		}()
 
+		userRegisteredChan := make(chan store.StoreResult, 1)
```
jespino, line 106: You are requesting the users, but you are not using it anywhere here

app/app.go
```diff
@@ -154,6 +155,91 @@ func (a *App) getFirstServerRunTimestamp() (int64, *model.AppError) {
...
+		sysAdmins = append(sysAdmins, sysAdminsList...)
+
+		if len(sysAdminsList) < perPage {
+			mlog.Debug("Number of system admins is less than limit", mlog.Int("admin count", len(sysAdminsList)))
+			break
+		}
+	}
+
+	for _, sysAdmin := range sysAdmins {
+		channel, appErr := a.GetOrCreateDirectChannel(sysAdmin.Id, sysAdmin.Id)
```
jespino, line 216: It should ben some kind of "system bot" sending the message, this looks like you are receiving a message in your own private channel with yourself a message written by the system.

app/app.go
```diff
@@ -154,6 +155,91 @@ func (a *App) getFirstServerRunTimestamp() (int64, *model.AppError) {
...
+		if appErr != nil {
+			mlog.Error("Cannot create channel for system notifications!", mlog.String("Admin Id", sysAdmin.Id))
+			return appErr
+		}
+
+		post := &model.Post{
+			UserId:    sysAdmin.Id,
```
jespino, line 223: Here should be the `systemBot` created for this kind of tasks

app/email.go
```diff
@@ -513,6 +513,53 @@ func (a *App) SendDeactivateAccountEmail(email string, locale, siteURL string) *
...
+		return model.NewAppError("SendWarnMetricAckEmail", "api.email.send_warn_metric_ack.invalid_warn_metric.app_error", nil, "", http.StatusInternalServerError)
+	}
+
+	if err := mailservice.SendMailUsingConfig(model.MM_ACKNOWLEDGE_ADDRESS, subject, body, a.Config(), false, email); err != nil {
+		mlog.Error("Error while sending email", mlog.String("email", model.MM_ACKNOWLEDGE_ADDRESS), mlog.Err(err))
+		return model.NewAppError("SendWarnMetricAckEmail", "api.email.send_warn_metric_ack.failure.app_error", map[string]interface{}{"Error": err.Error()}, "", http.StatusInternalServerError)
+	}
+
+	mlog.Info("Disable the monitoring of warn metric", mlog.String("metric", warnMetricId))
+	err := a.SetWarnMetricStatus(warnMetricId)
```
jespino, line 555: because this is here, I think you are unable to ack the warning unless you have the email configured. I think the idea was, when you "ACK" the warning, we accept the ACK optimistically and then try to send the email.

app/server.go
```diff
@@ -827,6 +837,43 @@ func doSessionCleanup(s *Server) {
...
+	}
+
+	warnMetricId := model.SYSTEM_WARN_METRIC_NUMBER_OF_ACTIVE_USERS
+	data, err := s.Store.System().GetByName(warnMetricId)
+	if err == nil && data != nil && (data.Value == "true" || data.Value == "ack") {
+		mlog.Debug("This metric limit crossing has been acknowledged or detected already")
+		return
+	}
+
+	noActiveUsers, err := s.Store.User().Count(model.UserCountOptions{})
```
jespino, line 855: `noActiveUsers` sounds like "no active users" and I think is all the way around. Probably `numberActiveUsers` without any abbreviation here would be better.

catalintomai: sure, it it makes it clearer.

app/analytics.go
catalintomai: yes, this needs to be removed since we are using the existing definition of registered users.

model/system.go
catalintomai: changed to 100.

app/server.go
catalintomai: changed it to 24 hours.

app/analytics.go
catalintomai: removed the dup and reusing the active user count definition as surfaced in the SysConsole.

app/server.go
catalintomai: uncommented

api4/system.go
catalintomai: I agree that we should always check against the permission, though here IsSystemAdmin() is a wrapper around the PERMISSION_MANAGE_SYSTEM check, I think we should either remove it or rename it to smth. else. (better remove it imo). I removed the call to be consistent, but right now its presence feels like it causes confusion.

app/app.go
catalintomai: I updated the code to create/use a system bot instead, thanks for the suggestion = ).

app/app.go
catalintomai: see above.

app/email.go
catalintomai: The behavior, to my understanding is:
- click the "Acknowledgement" button in the UX, this tries to send the email and store the ack in the DB
- if sending the email fails (because it is not configured on the server or some other reasons), we are not storing the ack and instead we are showing the user a "MailTo" link in the modal. If they click it, we send the ack and open the mailto link in the browser. There is a functionality gap/hole right now in that we do not check if the user sent the mailto mail in the browser (we are instead acking the moment they click on the mailto link). I updated the logic to handle the forceAck flag.

srkgupta:
Hi @catalintomai 
I am beginning to test this now. Can you please resolve all the merge conflicts here. Thanks.

srkgupta:
@catalintomai, 
9) Another issue I am seeing is that user needs to reload their page in order for them to see the Announcement bar. Ideally this should appear automatically without requiring the user to reload the page.

jasonblais:
Thank you @srkgupta! 

> 1. If a user clicks on the "Contact Support" in the bot message, the button remains the same. User is allowed to click on it multiple times. Is this okay? Shouldn't the button be hidden?

We should update experience to a loading spinner with text `Contacting Support...` while the request is being made, similar to saving settings in the System Console (see below).

image


> 2. When someone clicks on the "Contact Support" from the Announcement bar, the Announcement bar is hidden for all the admins, as expected. However, the individual DMs are still open and admins can still contact multiple times. Is this fine? This might just cause multiple support tickets from the same MM instance to be created.

Good question. I think this is fine for this iteration, and we can adjust the experience in the future. At worst we'll receive multiple contacts from the same server and need to manage it accordingly on our side. But from the admin's perspective the experience will remain smooth.

(If we were to remove the "contact support" button, the bot message may be confusing to the admin, so it's better to leave it there)

> 3. Currently the link for the text "AAE-010-1010" leads to this page "https://mattermost.com/pl/default-aae-faq" which does not exist. Will this page be created in the future (or) should the link be changed?

Yes, there is a ticket to add a redirect for this. Product will own, and this will be added prior to the release.

> 4. The Bot sends out a warning alert to all the admins every 24 hours once. Is this expected? Or should the bot send the message only once. I did not see any references to what is the expected functionality for this?

We should only send it once, not once a day, given we are planning to send other adviser notifications in the future.

cc @esadur @lfbrock for visibility

srkgupta:
Thanks @jasonblais. @catalintomai can you please also fix the issues 1 and 4 as per the above comments from Jason.

catalintomai:
> Thanks @jasonblais. @catalintomai can you please also fix the issues 1 and 4 as per the above comments from Jason.

@jasonblais , @srkgupta Not clear about #4 though. In the current design, if the user does not ack the warning, we send him another warning after 24 hours. If he acks the warning, the bot message is not being sent anymore. Is the new design proposing that we send the bot warning once? What is the user does not ack it, we leave it on the announcement bar to warn the user? In the announcement bar,  if not acked, the warning is being present permanently (subject to close/refresh). So in the announcement bar we have it quasi-permanently and in the bot we have it once?

catalintomai:
> Thanks @jasonblais. @catalintomai can you please also fix the issues 1 and 4 as per the above comments from Jason. Similar to issue 4, if the Announcement bar was dismissed by admin once by clicking on the close icon, then Admin should not see the Announcement bar again the next day. @jasonblais please correct if I am wrong.

@srkgupta , @jasonblais , while this possibly can be done with (more) mana that budgeted for this, I think this complicates things with no clear gain imo. We would have to keep track of when all admins(or one of them) closed the announcement bar, store that action+timestamp, and on refresh check if we are inside the one day interval and not show the announcement bar + when the metrics job kicks in, check if that admin has closed the announcement bar in the last 24 hours and only then warn him, etc. Also, note also that the current announcement bar logic is same as for the other warnings (like license expiring) that are displayed in the announcement bar. 
If the end-goal is to have the admins contact support and provide the info we request asap, then the current simpler design suits that purpose imo (including sending bot message every 24 hours). 
If the goal is different (or has changed in the meantime) it would be great to have it clearly articulated.

srkgupta:
> If the end-goal is to have the admins contact support and provide the info we request asap, then the current simpler design suits that purpose imo (including sending bot message every 24 hours).

The problem with sending bot message every 24 hours is that this might annoy admins with messages too often. Generally such decisions of upgrading a MM instance cannot be taken within a short period of time and needs to be discussed company wide. Hence if we keep sending a bot message (which also sends out an email notification if user has allowed Email notifications), then it might be an overwhelming experience. And if there are more such alerts planned in the future, admins might just be getting too many system messages too often. I will let @jasonblais add his perspective as well.

catalintomai:
> > If the end-goal is to have the admins contact support and provide the info we request asap, then the current simpler design suits that purpose imo (including sending bot message every 24 hours).
> 
> The problem with sending bot message every 24 hours is that this might annoy admins with messages too often. Generally such decisions of upgrading a MM instance cannot be taken within a short period of time and needs to be discussed company wide. Hence if we keep sending a bot message (which also sends out an email notification if user has allowed Email notifications), then it might be an overwhelming experience. And if there are more such alerts planned in the future, admins might just be getting too many system messages too often. I will let @jasonblais add his perspective as well.

@srkgupta , the original goal of the announcement bar warning + bot message was not to immediately trigger upgrade (wrt to "upgrading a MM instance") though that is indeed the long-term goal, clicking on "Contact Support" does not achieve that, instead it just gathers info that we can use down the line.

jasonblais:
@catalintomai @srkgupta For now, let's not make changes to 4 until Friday. On Friday I have a call with a few members from product where this feature will be one point of discussion. So until then, let's leave point 4 as-is.

However, below are my thoughts on it:

A - If I take the admin's perspective, here is how I was envisioning the experience:

1. Mattermost server exceeds 500 users
2. System Admin receives a bot message recommending they consider activating user management controls. They can contact support for guidance, or choose to do nothing.
3. System Admin also receives the announcement banner recommending the same as 2. They can similarly dismiss it or contact support.

> We would have to keep track of when all admins(or one of them) closed the announcement bar, store that action+timestamp, and on refresh check if we are inside the one day interval and not show the announcement bar + when the metrics job kicks in, check if that admin has closed the announcement bar in the last 24 hours and only then warn him, etc.

Technically, I think we can keep it simple by keeping actions independent across admins. E.g if one admin contacts support, it won't affect what the other admin sees. Personally, as a System Admin, I would like to know of this notification.

So, the announcement banner behaviour can probably stay the same..although @catalintomai can you confirm it only reappears if the admin reloads their client? If so, that should be fine for this iteration, but we may in future not want to show the banner even after a page reload.

B - My main concern is sending the bot message every day. With the user threshold alone, we are sending a message when the system exceeds 500, 1K, etc users (I'm not sure if that's part of this PR or if this only has one trigger for 500 users - looks like this only has the one trigger).

We are also planning to send similar notifications if their database exceeds certain number of messages (as this might affect performance) or if they're using the test push notification service (as this is not recommended in production use).

This could mean 10-15 bot messages a day, if we send them every day. That is a really, really bad experience. If I experienced this, I would stop using Mattermost and move to another system.

catalintomai:
> @catalintomai @srkgupta For now, let's not make changes to 4 until Friday. On Friday I have a call with a few members from product where this feature will be one point of discussion. So until then, let's leave point 4 as-is.
> 
> However, below are my thoughts on it:
> 
> A - If I take the admin's perspective, here is how I was envisioning the experience:
> 
> 1. Mattermost server exceeds 500 users
> 2. System Admin receives a bot message recommending they consider activating user management controls. They can contact support for guidance, or choose to do nothing.
> 3. System Admin also receives the announcement banner recommending the same as 2. They can similarly dismiss it or contact support.
> 
> > We would have to keep track of when all admins(or one of them) closed the announcement bar, store that action+timestamp, and on refresh check if we are inside the one day interval and not show the announcement bar + when the metrics job kicks in, check if that admin has closed the announcement bar in the last 24 hours and only then warn him, etc.
> 
> Technically, I think we can keep it simple by keeping actions independent across admins. E.g if one admin contacts support, it won't affect what the other admin sees. Personally, as a System Admin, I would like to know of this notification.
> 
> So, the announcement banner behaviour can probably stay the same..although @catalintomai can you confirm it only reappears if the admin reloads their client? If so, that should be fine for this iteration, but we may in future not want to show the banner even after a page reload.
> 
> B - My main concern is sending the bot message every day. With the user threshold alone, we are sending a message when the system exceeds 500, 1K, etc users (I'm not sure if that's part of this PR or if this only has one trigger for 500 users - looks like this only has the one trigger).
> 
> We are also planning to send similar notifications if their database exceeds certain number of messages (as this might affect performance) or if they're using the test push notification service (as this is not recommended in production use).
> 
> This could mean 10-15 bot messages a day, if we send them every day. That is a really, really bad experience. If I experienced this, I would stop using Mattermost and move to another system.

@jasonblais , thanks for weighing in. Some observations:
- "can you confirm it only reappears if the admin reloads their client" - it reappears on page reload OR when the server metric warning job detects that the metric limit is exceeded and a websocket event is triggered (for this metric every 24 hours).
- "we can keep it simple by keeping actions independent across admins" - I believe the opposite, by keeping actions independent we increase the complexity - to me the metric warning is per system and the ack should also be per system, not per system admin user. 
- " if we send them every day" - depending on the metric we could possibly reduce/increase the frequency interval. 
- "I'm not sure if that's part of this PR " - this PR addresses one metric warning but establishes the framework for the other metrics as well. 
Also, the deployment (represented by any admin) can easily get out of the warning state by clicking on "Contact Support" - so they will not get warning anymore once they ack. It is still TBD (to me) if the ack should be per metric or per set of metrics (right now it is per metric) - like what is the value of contacting support (which is the target action) for every metric for both us and the admin.

jasonblais:
@catalintomai Thanks for the additional insight! Very helpful

catalintomai:
Hi @srkgupta 
#1 - updated the PR to add a "Contacting Support" loading throb
#5 - I believe you see this because the MM modal is squished to the left (maybe you have the browser debug opened) - can you try having it full screen? Irrespective of this, I updated the PR to not have it disappear and wrap instead in such conditions.
#6 - updated PR to add space
#7 - cannot repro this on my MailApp on Mac - can you verify that this happens on mailto links only for this PR? Can you try hooking the browser as the default mail opened for mailto and see if the same behavior is exhibited?
#8 - updated PR to remove the extra confirmation
#9 - "this should appear automatically" - the announcement bar warning is currently showing in the case the metric warning is "on" and not acked if 1) the user refreshes the page and 2) automatically, if 24 hours passed since the job last detected that the metric is in warning state and not acked.
If this is not what you see then there is a bug that I cannot repro now,  if your expectation is different, please specify. Thanks.

lfbrock:
@jasonblais - are the reviews complete from this? I thought from the deep dive with Chandar he wanted more UX involvement on this feature

jasonblais:
@lfbrock UX review hasn't been done on this

jasonblais:
@lfbrock Example of a UX defect found by QA: https://mattermost.atlassian.net/browse/MM-26474

model/config.go
```diff
@@ -47,6 +47,7 @@ const (
 	GENERIC_NO_CHANNEL_NOTIFICATION = "generic_no_channel"
 	GENERIC_NOTIFICATION            = "generic"
 	GENERIC_NOTIFICATION_SERVER     = "https://push-test.mattermost.com"
+	MM_SUPPORT_ADDRESS              = "catalin@mattermost.com" //support@mattermost.com CITOMAI
```
srkgupta, line 50: @catalintomai can you revert the support email address, as long as you are not willing to handle all the support requests :). Currently its sending the email to your email account.

srkgupta:
Hi @catalintomai 

Few issues observed on the latest changes. 

1. First time when an instance having more than 500 users is migrated to this PR, then the admin gets 3 notifications by the bot as shown in this screenshot:
Screenshot 2020-07-21 at 6 52 42 PM

Similarly if there are more than 400 users, then user receives 2 notifications. 
Screenshot 2020-07-21 at 6 51 00 PM

Expected: Only one notification should be sent to the bot, i.e. the highest one. 

2. The Banner is not responsive and does not wraps in the second line. Please check this screenshot. The link "Learn More" is not displayed. 
Screenshot 2020-07-21 at 6 55 55 PM

3. For AD/LDAP notification, currently the link is shown as "Learn More". As per the doc, it is expected to display: "Learn more about integrating with AD/LDAP". Similarly for SAML notification, instead of "Learn More", it should say "Learn more about integrating with SAML 2.0"
Screenshot 2020-07-21 at 6 51 00 PM

model/config.go
catalintomai: @srkgupta - sure, will change it before submission = ).

catalintomai:
Hi @srkgupta. Thankss for re-testing this.
for #1 - my understanding is that the triggers are independent of one another - they don't override one another cc: @esadur 
#2 - I'm using the current banner framework, so not sure this behavior is scoped top this PR but will check and see what can be done.
#3 - I'll update the strings.

esadur:
@srkgupta good catches. 

For issue 1 (multiple advisories firing at the same time):
- My ideal solution would be to have logic in the feature to stagger the notifications. Something like an advisory queue, sorted by priority (where each advisory has an associated priority. The top advisory is surfaced to the admin, and then there's a quiet period before another is surface (e.g. 72 hours).
- A simpler solution (I'm guessing) would be just to show the most important one (in our case, the one with the highest user count trigger) and suppress the others

For issue 2 (text in banner cut off on narrow window): cc @andrewbrown00, I defer to him here. Not sure what the preferred design solution is here. A few options as I see them:
- Have the entire banner be clickable (@srkgupta I think this is what you were talking about)
- Wrap the text somehow
- Implement the rectangle button in @andrewbrown00's original designs and fix that to the right side of the banner so that when the window is narrowed, the banner overlays the text.

srkgupta:
> * The top advisory is surfaced to the admin, and then there's a quiet period before another is surface (e.g. 72 hours).

FYI @esadur these notifications are sent once every 24 hours. I had raised a concern around this, but did not see any response yet. Are we fine with bugging admins every 24 hours asking them to upgrade?

catalintomai:
> > * The top advisory is surfaced to the admin, and then there's a quiet period before another is surface (e.g. 72 hours).
> 
> FYI @esadur these notifications are sent once every 24 hours. I had raised a concern around this, but did not see any response yet. Are we fine with bugging admins every 24 hours asking them to upgrade?

@srkgupta - the 200 and 400 bot messages should be shown only once, if they are shown multiple times it is a bug.

srkgupta:
> > > * The top advisory is surfaced to the admin, and then there's a quiet period before another is surface (e.g. 72 hours).
> > 
> > 
> > FYI @esadur these notifications are sent once every 24 hours. I had raised a concern around this, but did not see any response yet. Are we fine with bugging admins every 24 hours asking them to upgrade?
> 
> @srkgupta - the 200 and 400 bot messages should be shown only once, if they are shown multiple times it is a bug.

No, I am talking about 500 bot messages. Even showing this every 24 hours is extremely annoying as a sysadmin and even @jasonblais confirmed that they will stop using the product :(

esadur:
@srkgupta the 200- and 400-user advisories should only show up once. The 500-user advisor will recur daily until the user opts to contact support. @jasonblais is onboard with this approach.

justinegeffen:
Related forum post: https://forum.mattermost.org/t/new-admin-advisor-notifications-in-mattermost-v5-26/10263.

PR #14487, opened 2020-05-06
MM-24911: re-add the reactions table migration

agnivade:
Adding the line to migrate it in 5.24

#### Ticket link

https://mattermost.atlassian.net/browse/MM-24911


jespino:
One concern about this, how it would behave with a table with a million of reactions in it? Is this migration going to take a lot of time?

agnivade:
> One concern about this, how it would behave with a table with a million of reactions in it? Is this migration going to take a lot of time?

This was already tested by George and the answer is around 30 seconds IIRC. In the 5.20 release notes, this is explained in detail. It is suggested that for heavy tables, this query can be run in advance in a low-traffic time.

PR #14494, opened 2020-05-06
[MM-24453] Sanitize user term search

ethervoid:
#### Summary

Remove the @ at the beginning of the search term in order to
make it possible for ES to return results

#### Ticket Link

https://mattermost.atlassian.net/browse/MM-24453


store/searchlayer/user_layer.go
```diff
@@ -142,6 +145,10 @@ func (s *SearchUserStore) autocompleteUsersInChannelByEngine(engine searchengine
 	return autocomplete, nil
 }
 
+func (s *SearchUserStore) sanitizeTerm(term string) string {
+	return strings.TrimLeft(term, "@")
+}
```
mgdelacroix, line 150: nit: if we're not using the store as part of this function, we can move it to an external file like `utils.go` in the same package to be an independent method

ethervoid: Good point. On it

ethervoid: Done

ethervoid:
@isacikgoz I've been thinking about where to put this method and although we have sanitization in the app layer we need to include it in the store layer because the store methods can be used anywhere.

I think the search layer is a perfect for because it works for all the search engines and we don't have to duplicate it inside every engine code

PR #14496, opened 2020-05-07
provide and implement a new cache provider

liusy182:
#### Summary

This PR is a continuation from [14118](https://github.com/mattermost/mattermost-server/pull/14118#issuecomment-624668540). It provides and implements a cache provider under the interim package `cache2`

- For the ease of migration, the contract of the `cache2.Provider` remains the same as the existing `cache.Provider`

- The `name` param in `NewCacheWithParams` is not in use by the LRU cache, but I imagine this will be useful for redis cache in future as key prefixes.


services/cache2/provider.go
```diff
@@ -0,0 +1,50 @@
...
+package cache2
+
+import "time"
+
+// Provider is a provider for Cache
+type Provider interface {
+	// NewCache creates a new cache with given size.
+	NewCache(size int) Cache
+	// NewCacheWithParams creates a new cache with the given parameters.
+	NewCacheWithParams(size int, name string, defaultExpiryInSecs int, invalidateClusterEvent string) Cache
```
liusy182, line 13: on second thought, i am thinking to change this API a little bit, with a slight drawback of a dozens of more line updates. Please let me know what you think before I proceed:

```
// type ProviderOptions {
//   size int
//   defaultExpiry time.Duration
//   name string
//   invalidateClusterEvent string
// }

NewCacheWithParams(opt *ProviderOptions) Cache
```

the rationale is as follow:

1. Having a struct `ProviderOptions` so that we can keep the same interface in case a new field is required in the future. e.g. the `name` field is needed for a redis cache but not for a LRU cache. What if some other fields are needed for a new type of cache we support in future?

2. change `defaultExpiry` to a `time.Duration`. This one is pretty obvious.

agnivade: Thanks. This is on my list. I will take a look whenever I get some time.

agnivade: Ok, your changes to ProviderOptions look reasonable to me. Since we are changing the interface, let's change it to be better. I am okay with changing more code if that gives us more benefit in the long term. Let's also change the expiry to time.Duration.

Did you want to update the references from using the LRU implementation directly to going via the provider in another PR or this PR itself ? I am okay with both options.

jespino: Yes, I agree with this too. Is better to change it now, and with 4 parameters that are all options, I think is less error prone to have a data structure for that.

liusy182: Thanks. I will update this CR.

> Did you want to update the references from using the LRU implementation directly to going via the provider in another PR or this PR itself ? I am okay with both options.

My goal was to migrate usage as it is with minimum logical changes possible. I don't plan to migrate direct LRU references to cacheProvider. This PR contains all direct references to LRU and the rest of the code base all uses cacheProvider. 

I think we can do migration of direct LRU usage as a separate topic, or as a followup after the we are fully migrated to cache2 interface.

agnivade: Sorry could you explain your comment a bit ? I am not clear what you meant by these lines

> I don't plan to migrate direct LRU references to cacheProvider.

> I think we can do migration of direct LRU usage as a separate topic, 

These two seem to be conflicting to me.

liusy182: Sorry for being confusing. I meant to say that the first goal in my mind is to complete `cache2` package migration. Other improvements such as removing direct LRU references can be done at later stages, or do you think it is better to do it now?

agnivade: Got it. Yes, I agree with that.

services/cache2/provider.go
```diff
@@ -0,0 +1,49 @@
...
+	Name                   string
+	InvalidateClusterEvent string
+}
+
+// Provider is a provider for Cache
+type Provider interface {
+	// NewCache creates a new cache with given options.
+	NewCache(opts *CacheOptions) Cache
+	// Connect opens a new connection to the cache using specific provider parameters.
+	Connect()
```
agnivade, line 21: Let's also add `error` returns for both Connect and Close whenever we add external cache support later.

liusy182: make sense. will change that.

teresa-novoa:
@jespino friendly reminder to review this PR, thanks!

agnivade:
@lindy65 - Nothing to test here, since this just adds code which is not used anywhere.

teresa-novoa:
@agnivade can you help to merging this PR? Thanks!

agnivade:
My apologies. I should have assigned it to myself. Sorry for the delay @liusy182 !

PR #14509, opened 2020-05-08
WIP: Filter stop words when searching posts in mysql

rvillablanca:
#### Summary
This PR resolves partially #14439 by filtering stopwords from terms introduced in search field.

#### Ticket Link

Partially #14439 

JIRA: https://mattermost.atlassian.net/browse/MM-24523

**Notes**: It resolves only one point of three in the ticket, this is to remove all matching stopwords from the terms introduced by the user, for example having a post with the content `"where is the car"` and looking for `"the car"` will retrive the post. Before this changes when looking for a post with terms that containing a stopword, you always got empty list.

This is a WIP yet because currently stopwords are being retrived from Mysql in every search, and I think this could be stored in a variable in order to save an unnecessary and repeated query but not sure where to place that var.

Secondly, I think could be a bug when searching for a post with the terms `"where is my car"`, notice the `"my"` word, it is not an stopword but it is always getting empty results (assuming a post with `"where is my car"` already exists).

Finally for cases when looking with terms with hashtags that are stopwords (`#where`, `#the`, ...) I prefer to create another PR and analize it with more calm, I'm not sure to how to solve that, maybe using `REGEXP` as suggested by @ethervoid but I'm not sure about the performance of that.

### DIFF
--- a/i18n/en.json
+++ b/i18n/en.json
@@ -6671,6 +6671,10 @@
     "translation": "Searching has been disabled on this server. Please contact your System Administrator."
   },
   {
+    "id": "store.sql_post.search.app_error",
+    "translation": "We encountered an error searching posts."
+  },
+  {
     "id": "store.sql_post.update.app_error",
     "translation": "Unable to update the Post."
   },
--- a/store/searchtest/post_layer.go
+++ b/store/searchtest/post_layer.go
@@ -263,6 +263,11 @@ var searchPostStoreTests = []searchTest{
 		Fn:   testShouldBeAbleToFindPostsEvenIfStopWordsWhereGiven,
 		Tags: []string{ENGINE_MYSQL},
 	},
+	{
+		Name: "Should not go to database when searching with stopwords only",
+		Fn:   testShouldNotGoToDatabaseWhenSearchingWithStopWordsOnly,
+		Tags: []string{ENGINE_MYSQL},
+	},
 }
 
 func TestSearchPostStore(t *testing.T, s store.Store, testEngine *SearchTestEngine) {
@@ -1755,4 +1760,14 @@ func testShouldBeAbleToFindPostsEvenIfStopWordsWhereGiven(t *testing.T, th *Sear
 	require.Nil(t, apperr)
 	require.Len(t, posts.Posts, 1)
 	require.Equal(t, p1.Message, posts.Posts[p1.Id].Message)
+}
+
+func testShouldNotGoToDatabaseWhenSearchingWithStopWordsOnly(t *testing.T, th *SearchTestHelper) {
+	_, err := th.createPost(th.User.Id, th.ChannelBasic.Id, "where is the car?", "", model.POST_DEFAULT, 0, false)
+	require.Nil(t, err)
+	defer th.deleteUserPosts(th.User.Id)
+
+	posts, apperr := th.Store.Post().Search(th.Team.Id, th.User.Id, &model.SearchParams{Terms: "where is the"})
+	require.Nil(t, apperr)
+	require.Len(t, posts.Posts, 0)
 }
--- a/store/searchtest/post_layer.go
+++ b/store/searchtest/post_layer.go
@@ -258,6 +258,11 @@ var searchPostStoreTests = []searchTest{
 		Fn:   testShouldNotReturnLinksEmbeddedInMarkdown,
 		Tags: []string{ENGINE_POSTGRES, ENGINE_ELASTICSEARCH},
 	},
+	{
+		Name: "Should be able to find post even if stop word where given",
+		Fn:   testShouldBeAbleToFindPostsEvenIfStopWordsWhereGiven,
+		Tags: []string{ENGINE_MYSQL},
+	},
 }
 
 func TestSearchPostStore(t *testing.T, s store.Store, testEngine *SearchTestEngine) {
@@ -1739,4 +1744,15 @@ func testShouldNotReturnLinksEmbeddedInMarkdown(t *testing.T, th *SearchTestHelp
 	require.Nil(t, apperr)
 
 	require.Len(t, results.Posts, 0)
+}
+
+func testShouldBeAbleToFindPostsEvenIfStopWordsWhereGiven(t *testing.T, th *SearchTestHelper) {
+	p1, err := th.createPost(th.User.Id, th.ChannelBasic.Id, "where is the car?", "", model.POST_DEFAULT, 0, false)
+	require.Nil(t, err)
+	defer th.deleteUserPosts(th.User.Id)
+
+	posts, apperr := th.Store.Post().Search(th.Team.Id, th.User.Id, &model.SearchParams{Terms: "where is the car"})
+	require.Nil(t, apperr)
+	require.Len(t, posts.Posts, 1)
+	require.Equal(t, p1.Message, posts.Posts[p1.Id].Message)
 }
--- a/store/sqlstore/post_store.go
+++ b/store/sqlstore/post_store.go
@@ -6,6 +6,7 @@ package sqlstore
 import (
 	"database/sql"
 	"fmt"
+	"github.com/pkg/errors"
 	"net/http"
 	"regexp"
 	"strconv"
@@ -1236,6 +1237,24 @@ func (s *SqlPostStore) search(teamId string, userId string, params *model.Search
 		searchClause := fmt.Sprintf("AND to_tsvector('english', %s) @@  to_tsquery('english', :Terms)", searchType)
 		searchQuery = strings.Replace(searchQuery, "SEARCH_CLAUSE", searchClause, 1)
 	} else if s.DriverName() == model.DATABASE_DRIVER_MYSQL {
+		if searchType == "Message" {
+			stopWords, err := findStopWordsMysql(s)
+			if err != nil {
+				return nil, model.NewAppError("SqlPostStore.search", "store.sql_post.search.app_error", nil, err.Error(), http.StatusInternalServerError)
+			}
+
+			for i := 0; i < len(stopWords); i++ {
+				stopWords[i] = fmt.Sprintf(`\b%s\b`, stopWords[i])
+			}
+			re, err := regexp.Compile(strings.Join(stopWords, "|"))
+			if err != nil {
+				return nil, model.NewAppError("SqlPostStore.search", "store.sql_post.search.app_error", nil, err.Error(), http.StatusInternalServerError)
+			}
+
+			terms = re.ReplaceAllString(terms, " ")
+			terms = strings.TrimSpace(terms)
+		}
+
 		searchClause := fmt.Sprintf("AND MATCH (%s) AGAINST (:Terms IN BOOLEAN MODE)", searchType)
 		searchQuery = strings.Replace(searchQuery, "SEARCH_CLAUSE", searchClause, 1)
 
@@ -1279,6 +1298,17 @@ func (s *SqlPostStore) search(teamId string, userId string, params *model.Search
 	}
 	list.MakeNonNil()
 	return list, nil
+}
+
+func findStopWordsMysql(s *SqlPostStore) ([]string, error) {
+	query := "SELECT * FROM INFORMATION_SCHEMA.INNODB_FT_DEFAULT_STOPWORD"
+	var stopWords []string
+	_, err := s.GetReplica().Select(&stopWords, query)
+	if err != nil {
+		return nil, errors.Wrap(err, "could not find mysql stop words")
+	}
+
+	return stopWords, nil
 }
 
 func (s *SqlPostStore) AnalyticsUserCountsWithPostsByDay(teamId string) (model.AnalyticsRows, *model.AppError) {
--- a/store/sqlstore/post_store.go
+++ b/store/sqlstore/post_store.go
@@ -1253,6 +1253,10 @@ func (s *SqlPostStore) search(teamId string, userId string, params *model.Search
 
 			terms = re.ReplaceAllString(terms, " ")
 			terms = strings.TrimSpace(terms)
+
+			if terms == "" {
+				return list, nil
+			}
 		}
 
 		searchClause := fmt.Sprintf("AND MATCH (%s) AGAINST (:Terms IN BOOLEAN MODE)", searchType)


store/sqlstore/post_store.go
```diff
@@ -1236,6 +1237,28 @@ func (s *SqlPostStore) search(teamId string, userId string, params *model.Search
 		searchClause := fmt.Sprintf("AND to_tsvector('english', %s) @@  to_tsquery('english', :Terms)", searchType)
 		searchQuery = strings.Replace(searchQuery, "SEARCH_CLAUSE", searchClause, 1)
 	} else if s.DriverName() == model.DATABASE_DRIVER_MYSQL {
+		if searchType == "Message" {
+			stopWords, err := findStopWordsMysql(s)
```
ethervoid, line 1241: I agree with your point. This should come from a static list. I'd put that list in a file called `stop_word.go` in `searchlayer` folder/package

store/sqlstore/post_store.go
```diff
@@ -1236,6 +1237,28 @@ func (s *SqlPostStore) search(teamId string, userId string, params *model.Search
...
+			if err != nil {
```
ethervoid, line 1242: Getting that list makes this unnecessary

store/sqlstore/post_store.go
```diff
@@ -1236,6 +1237,28 @@ func (s *SqlPostStore) search(teamId string, userId string, params *model.Search
...
+				return nil, model.NewAppError("SqlPostStore.search", "store.sql_post.search.app_error", nil, err.Error(), http.StatusInternalServerError)
+			}
+
+			for i := 0; i < len(stopWords); i++ {
```
ethervoid, line 1246: Given that you'd have a slice of strings you don't need this

store/sqlstore/post_store.go
```diff
@@ -1236,6 +1237,28 @@ func (s *SqlPostStore) search(teamId string, userId string, params *model.Search
...
+				stopWords[i] = fmt.Sprintf(`\b%s\b`, stopWords[i])
+			}
+			re, err := regexp.Compile(strings.Join(stopWords, "|"))
```
ethervoid, line 1249: This approach removes part of words like `whereabouts`

I've made a simple test as you can see [here](https://play.golang.com/p/wNfSRsCAnfP) that could work as example

rvillablanca: I think the regexp is fine because is catching the **whole** word, but I made some tests to corroborate it :)

store/sqlstore/post_store.go
rvillablanca: Not sure about this, maybe I don't understand the point :sweat_smile:

rvillablanca:
With the current approach the list is static and we won't get any update on that list, maybe other approach could be something like:

```go
func mysqlStopWords() []string {
    if stopWords == nil {
        stopWorlds = retrieveFromDB()
    }

    return stopWords
}
```

Although I'm not sure if it's worth doing because I don't believe stop words are updated often.

/cc @ethervoid

ethervoid:
@rvillablanca Testing again, the `\b` splits also with the hyphen and for words like `father-in-law` is broken. I've updated [my test](https://play.golang.com/p/RRCflQTIs37) with both approaches

store/searchlayer/stop_word.go
```diff
@@ -0,0 +1,6 @@
+package searchlayer
+
+var (
+	MYSQL_STOP_WORDS = []string{"a", "about", "an", "are", "as", "at", "be", "by", "com", "de", "en", "for", "from", "how", "i", "in", "is", "it", "la", "of",
```
ethervoid, line 4: In the future, we're going to split the MySQL and the Postgres layer so this is useful because we can end up with a common list of stop words to do a prefilter

store/searchtest/post_layer.go
```diff
@@ -258,6 +258,11 @@ var searchPostStoreTests = []searchTest{
 		Fn:   testShouldNotReturnLinksEmbeddedInMarkdown,
 		Tags: []string{ENGINE_POSTGRES, ENGINE_ELASTICSEARCH},
 	},
+	{
+		Name: "Should not go to database when searching with stopwords only",
```
ethervoid, line 262: I'd include this test inside the `Should be able to ignore stop words` suite

store/sqlstore/post_store.go
```diff
@@ -1236,6 +1237,18 @@ func (s *SqlPostStore) search(teamId string, userId string, params *model.Search
 		searchClause := fmt.Sprintf("AND to_tsvector('english', %s) @@  to_tsquery('english', :Terms)", searchType)
 		searchQuery = strings.Replace(searchQuery, "SEARCH_CLAUSE", searchClause, 1)
 	} else if s.DriverName() == model.DATABASE_DRIVER_MYSQL {
+		if searchType == "Message" {
```
ethervoid, line 1240: Good way to avoid messing with the Hash :)

store/sqlstore/post_store.go
```diff
@@ -1281,6 +1294,22 @@ func (s *SqlPostStore) search(teamId string, userId string, params *model.Search
 	return list, nil
 }
 
+func removeMysqlStopWordsFromTerms(terms string) (string, error) {
```
ethervoid, line 1297: I like the extraction :)

store/searchlayer/stop_word.go
```diff
@@ -0,0 +1,6 @@
+package searchlayer
+
+var (
```
ethervoid, line 3: One question. Why do you wrap the variable instead of doing something like `var MYSQL...`?

store/sqlstore/post_store.go
```diff
@@ -1281,6 +1294,22 @@ func (s *SqlPostStore) search(teamId string, userId string, params *model.Search
...
+	stopWords := make([]string, len(searchlayer.MYSQL_STOP_WORDS))
+	copy(stopWords, searchlayer.MYSQL_STOP_WORDS)
+	for i := 0; i < len(stopWords); i++ {
+		stopWords[i] = fmt.Sprintf(`\b%s\b`, stopWords[i])
+	}
+	re, err := regexp.Compile(strings.Join(stopWords, "|"))
```
ethervoid, line 1303: Testing again, the \b splits also with the hyphen and for words like father-in-law is broken. I've updated [my test](https://play.golang.com/p/RRCflQTIs37) with both approaches

store/sqlstore/post_store_test.go
```diff
@@ -17,3 +18,23 @@ func TestPostStore(t *testing.T) {
 func TestSearchPostStore(t *testing.T) {
 	StoreTestWithSearchTestEngine(t, searchtest.TestSearchPostStore)
 }
+
+func TestMysqlStopWords(t *testing.T) {
```
ethervoid, line 22: Nice addition. You have to include hyphenated words :)

store/sqlstore/post_store.go
rvillablanca: Great!

store/sqlstore/post_store.go
rvillablanca: :smile:

store/searchlayer/stop_word.go
rvillablanca: Perfect

store/searchlayer/stop_word.go
rvillablanca: Maybe I am used to them, removing

store/sqlstore/post_store_test.go
rvillablanca: Ok

rvillablanca:
@ethervoid I made the suggestions but we need to adjust it to cover all cases. Having more time I'll do this

ethervoid:
@rvillablanca Sure, no rush. Take your time

store/sqlstore/post_store.go
```diff
@@ -1283,6 +1296,26 @@ func (s *SqlPostStore) search(teamId string, userId string, params *model.Search
 	return list, nil
 }
 
+func removeMysqlStopWordsFromTerms(terms string) (string, error) {
+	stopWords := make([]string, len(searchlayer.MYSQL_STOP_WORDS))
+	copy(stopWords, searchlayer.MYSQL_STOP_WORDS)
+	re, err := regexp.Compile(fmt.Sprintf(`^(%s)$`, strings.Join(stopWords, "|")))
```
ethervoid, line 1302: I've updated [the test](https://play.golang.com/p/NgkfNQkLFRN) and, if I'm not mistaken, the change is not working as expected

rvillablanca: yep, I had to change the approach because it couldn't find a regexp that could do all what I wanted, this is, a regexp that work with `"where about"` but not with `"where-about"` nor `"whereabout"`. I'm not the best with regexp but I thought I needed the lookbehind feature in golang regexp but it is not implemented, so I couldn't find a single regexp that solve these cases. After giving it some time, I decided to change the approch so now it is working good and it is covering all cases (at least those we talked about).

rvillablanca: Anyway the tests are passing now

https://github.com/rvillablanca/mattermost-server/blob/b42936f70941d44705b27f6b9f50bd8d152519ea/store/sqlstore/post_store_test.go#L22-L49

ethervoid: Cool, looks good to me :)

i18n/en.json
```diff
@@ -6790,6 +6790,10 @@
     "id": "store.sql_post.search.disabled",
     "translation": "Searching has been disabled on this server. Please contact your System Administrator."
   },
+  {
+    "id": "store.sql_post.search.app_error",
+    "translation": "We encountered an error searching posts."
```
ethervoid, line 6795:
```suggestion
    "translation": "Error searching posts"
```

store/sqlstore/post_store.go
```diff
@@ -1238,6 +1239,18 @@ func (s *SqlPostStore) search(teamId string, userId string, params *model.Search
 		searchClause := fmt.Sprintf("AND to_tsvector('english', %s) @@  to_tsquery('english', :Terms)", searchType)
 		searchQuery = strings.Replace(searchQuery, "SEARCH_CLAUSE", searchClause, 1)
 	} else if s.DriverName() == model.DATABASE_DRIVER_MYSQL {
+		if searchType == "Message" {
+			var err error
+			terms, err = removeMysqlStopWordsFromTerms(terms)
```
ethervoid, line 1244:
```suggestion
			terms, err := removeMysqlStopWordsFromTerms(terms)
```

store/sqlstore/post_store.go
```diff
@@ -1283,6 +1296,26 @@ func (s *SqlPostStore) search(teamId string, userId string, params *model.Search
...
+	}
+
+	newTerms := make([]string, 0)
+	separatedTerms := strings.Split(terms, " ")
+	for _, term := range separatedTerms {
+		term = strings.TrimSpace(term)
+		term = re.ReplaceAllString(term, "")
+		if term != "" {
+			newTerms = append(newTerms, term)
+		}
```
ethervoid, line 1314:
```suggestion
		if term = re.ReplaceAllString(term, ""); term != "" {
			newTerms = append(newTerms, term)
		}
```

store/sqlstore/post_store_test.go
```diff
@@ -17,3 +18,32 @@ func TestPostStore(t *testing.T) {
...
+
+	t.Run("Should not remove part of a word containg stop words separated by hyphens", func(t *testing.T) {
+		cases := []string{"where-about", "where-in", "to-this", "this-or-that", "was-when", "who-was", "in-the", "who-will", "the-www"}
+		for _, term := range cases {
+			got, err := removeMysqlStopWordsFromTerms(term)
+			require.NoError(t, err)
+			require.Equal(t, term, got)
+		}
+	})
+}
```
ethervoid, line 49:
```suggestion
func TestMysqlStopWords(t *testing.T) {
	mysqlStopWordsTests := []struct {
		Name     string
		Args     []string
		Expected []string
		Empty    bool
	}{
		{
			Name:     "Should remove only the stop words",
			Args:     []string{"where is my car", "so this is real", "test this-and-that is awesome"},
			Expected: []string{"my car", "so real", "test this-and-that awesome"},
		},
		{
			Name:     "Should not remove part of a word containg stop words",
			Args:     []string{"whereabouts", "wherein", "tothis", "thisorthat", "waswhen", "whowas", "inthe", "whowill", "thewww"},
			Expected: []string{"whereabouts", "wherein", "tothis", "thisorthat", "waswhen", "whowas", "inthe", "whowill", "thewww"},
		},
		{
			Name:  "Should remove all words from terms",
			Args:  []string{"where about", "where in", "to this", "this or that", "was when", "who was", "in the", "who will", "the www"},
			Empty: true,
		},
		{
			Name:     "Should not remove part of a word containg stop words separated by hyphens",
			Args:     []string{"where-about", "where-in", "to-this", "this-or-that", "was-when", "who-was", "in-the", "who-will", "the-www"},
			Expected: []string{"where-about", "where-in", "to-this", "this-or-that", "was-when", "who-was", "in-the", "who-will", "the-www"},
		},
	}

	for _, tc := range mysqlStopWordsTests {
		t.Run(tc.Name, func(t *testing.T) {
			for i, term := range tc.Args {
				got, err := removeMysqlStopWordsFromTerms(term)
				require.NoError(t, err)
				if tc.Empty {
					require.Empty(t, got)
				} else {
					require.Equal(t, tc.Expected[i], got)
				}
			}
		})
	}
}
```

ethervoid: This suggestion uses test cases struct and also I've added one more test case that removes stop words in the string

store/sqlstore/post_store.go
rvillablanca: This case we are shadowing the var `terms`, so we need to re-use the same var :)

store/sqlstore/post_store_test.go
rvillablanca: Great!

rvillablanca:
Ok we are done with all suggestions, thanks for the review @ethervoid :tada:

teresa-novoa:
@mgdelacroix friendly reminder to review this PR, thanks!

ethervoid:
@rvillablanca please follow [this checklist](https://developers.mattermost.com/contribute/getting-started/contribution-checklist/) because the check is saying you didn't sign the CLA :)

rvillablanca:
I think this is a mistake, this is not my first PR and I have several others approved and merged into master.

rvillablanca:
I just checked the list and I do appear on it

ethervoid:
@rvillablanca fixed. It was the check, we should take a look at it. Sorry for the noise

rvillablanca:
No problem @ethervoid, I forgot to merge master yerterday

store/sqlstore/post_store.go
```diff
@@ -1283,6 +1296,25 @@ func (s *SqlPostStore) search(teamId string, userId string, params *model.Search
...
+func removeMysqlStopWordsFromTerms(terms string) (string, error) {
+	stopWords := make([]string, len(searchlayer.MYSQL_STOP_WORDS))
+	copy(stopWords, searchlayer.MYSQL_STOP_WORDS)
+	re, err := regexp.Compile(fmt.Sprintf(`^(%s)$`, strings.Join(stopWords, "|")))
+	if err != nil {
+		return "", err
+	}
+
+	newTerms := make([]string, 0)
+	separatedTerms := strings.Split(terms, " ")
```
mgdelacroix, line 1308: nit: here we can use `Fields` instead and we will not need the `TrimSpace` in each term after:

```suggestion
	separatedTerms := strings.Fields(terms)
```

rvillablanca: Perfect!

ethervoid:
@ogi-m gentle reminder

teresa-novoa:
@ogi-m friendly reminder

jasonblais:
@ogi-m a friendly reminder with the QA review. Feel free to reassign to another QA if you are tight on time?

jasonblais:
@ogi-m can you help with QA review?

ogi-m (commented):
Tested with all the stop words and looks good, results are returned.

One question - this means there will be posts returned that don't contain all the search terms that the user entered (posts without the stopword). For example:
search

Normally the expected would be to return only the posts that match all the search terms.
Just checking if we're ok with the different behavior in these situations? @ethervoid

ethervoid:
@ogi-m hum would you please check that same example with ES to check the result?

ogi-m:
@ethervoid I've checked Postgres and ES with the same example:
1. Postgres - behaves completely the same as MySQL (same search results as in the screenshot above)
2. ES - also returns posts that don't contain all the search terms, but the behavior is slightly different as it doesn't return posts with numbers in that case:

ES

ethervoid:
Ok that should be good then. Probably we should take a look to know why messages with number don't appear in ES. I'll include @thefactremains to check this change of behavior but I think is ok to release this

PR #14514, opened 2020-05-08
[MM-24987] Dont call GetSchemeRolesForChannel and instead load the scheme from channel

fmunshi:
#### Summary
- Instead of always calling `GetSchemeRolesForChannel` modify this block to load use the scheme returned from create (if one is created) or call `GetScheme` directly using the channel struct already loaded into memory.
- This is required since there were cases where retrieving the channel from store would return a channel with an outdated scheme id resulting in the function `GetSchemeRolesForChannel` returning the team scheme roles (or system scheme roles) 

#### Ticket Link
- https://mattermost.atlassian.net/browse/MM-24987

#### Test steps
- Apply label `Setup HA Cloud Test Server`
- Create a team scheme and set all permissions to true
- Open channel moderation for a channel in that team and enable / disable create posts multiple times 
- Eventually it will say "Create posts is disabled in the team scheme" and you will be unable to add create posts back to the channel (as it gets disabled on the team level)

### DIFF
--- a/app/channel_test.go
+++ b/app/channel_test.go
@@ -1660,56 +1660,41 @@ func TestPatchChannelModerationsForChannel(t *testing.T) {
 	}
 
 	t.Run("Handles concurrent patch requests gracefully", func(t *testing.T) {
+		addCreatePosts := []*model.ChannelModerationPatch{
+			{
+				Name: &createPosts,
+				Roles: &model.ChannelModeratedRolesPatch{
+					Members: model.NewBool(false),
+					Guests:  model.NewBool(false),
+				},
+			},
+		}
+		removeCreatePosts := []*model.ChannelModerationPatch{
+			{
+				Name: &createPosts,
+				Roles: &model.ChannelModeratedRolesPatch{
+					Members: model.NewBool(false),
+					Guests:  model.NewBool(false),
+				},
+			},
+		}
+
 		wg := sync.WaitGroup{}
 		wg.Add(20)
 		for i := 0; i < 10; i++ {
 			go func() {
-				th.App.PatchChannelModerationsForChannel(channel, []*model.ChannelModerationPatch{
-					{
-						Name: &createPosts,
-						Roles: &model.ChannelModeratedRolesPatch{
-							Members: model.NewBool(true),
-							Guests:  model.NewBool(true),
-						},
-					},
-				})
-				th.App.PatchChannelModerationsForChannel(channel, []*model.ChannelModerationPatch{
-					{
-						Name: &createPosts,
-						Roles: &model.ChannelModeratedRolesPatch{
-							Members: model.NewBool(false),
-							Guests:  model.NewBool(false),
-						},
-					},
-				})
+				th.App.PatchChannelModerationsForChannel(channel, addCreatePosts)
+				th.App.PatchChannelModerationsForChannel(channel, removeCreatePosts)
 				wg.Done()
 			}()
 		}
-
 		for i := 0; i < 10; i++ {
 			go func() {
-				th.App.PatchChannelModerationsForChannel(channel, []*model.ChannelModerationPatch{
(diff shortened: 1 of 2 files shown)


agnivade:
First of all, sorry about the flaky test, I am trying to merge my [PR](https://github.com/mattermost/mattermost-server/pull/14399) ASAP.

Coming to the PR, I am trying to understand the root cause here. IIUC this is happening because `a.GetChannel` returns from cache. And the race is that some other server updates the channel scheme, and before the invalidation request hits this server, this code gets executed, which makes it return a nil scheme, and therefore it returns the team scheme.

So instead we call `GetScheme` directly. But the `GetScheme` call is itself cached. A `CreateChannelScheme` call makes 2 separate calls where first the scheme gets created and _then_ the channel gets updated with the scheme id. That's itself racy and non-atomic.

I am wondering whether there can still be a race where a scheme invalidation event hasn't reached the server, and `GetScheme` returns an old scheme. Can there still be issues with this ?

fmunshi:
Hi @agnivade 

The block containing `a.GetScheme(*channel.SchemeId)` will only be run if the channel already had a scheme associated with it - so i dont believe there could be any issues with this approach since even if somehow the channel was cached from a different request (two people editing the channel moderation at the same time) it would just return a nil scheme instead in that case, since the channel never directly references a team scheme 

The main issue with the other method (`GetSchemeRolesForChannel`) was that it would return the higher scoped scheme roles as a fallback (line 658 of app/channel.go `return a.GetTeamSchemeChannelRoles(channel.TeamId)`)

mkraft:
> First of all, sorry about the flaky test, I am trying to merge my [PR](https://github.com/mattermost/mattermost-server/pull/14399) ASAP.
> 
> Coming to the PR, I am trying to understand the root cause here. IIUC this is happening because `a.GetChannel` returns from cache. And the race is that some other server updates the channel scheme, and before the invalidation request hits this server, this code gets executed, which makes it return a nil scheme, and therefore it returns the team scheme.
> 
> So instead we call `GetScheme` directly. But the `GetScheme` call is itself cached. A `CreateChannelScheme` call makes 2 separate calls where first the scheme gets created and _then_ the channel gets updated with the scheme id. That's itself racy and non-atomic.
> 
> I am wondering whether there can still be a race where a scheme invalidation event hasn't reached the server, and `GetScheme` returns an old scheme. Can there still be issues with this ?

Just to add some more perspective, I validated the cache theory by testing this change, which also fixes it, but at the cost of an extra non-cached DB call: https://github.com/mattermost/mattermost-server/pull/14513/files.

I like @fmunshi more than this because it doesn't require the extra db call.

app/channel_test.go
```diff
@@ -1657,6 +1658,67 @@ func TestPatchChannelModerationsForChannel(t *testing.T) {
...
+			}()
+		}
+
+		wg.Wait()
+
+		higherScopedGuestRoleName, higherScopedMemberRoleName, _, _ := th.App.GetTeamSchemeChannelRoles(channel.TeamId)
+		higherScopedMemberRole, _ := th.App.GetRoleByName(higherScopedMemberRoleName)
+		higherScopedGuestRole, _ := th.App.GetRoleByName(higherScopedGuestRoleName)
+		assert.Contains(t, higherScopedMemberRole.Permissions, createPosts)
+		assert.Contains(t, higherScopedGuestRole.Permissions, createPosts)
```
fmunshi, line 1719: @agnivade I added this test and confirmed that it fails consistently on master (without the fix in this PR) and that once I switch to this PR it was passing consistently

fmunshi:
@srkgupta in order to test this it will need to be setup on a HA Cloud Test Server, I will add the label once the build passes

amyblais:
@fmunshi This can also be cherry-picked to v5.22 for a dot release.

PR #14516, opened 2020-05-08
MM-24939 expose ExtendSessionLengthWithActivity config to clients

wiggin77:
#### Summary

This PR exposes the `ServiceSettings.ExtendSessionLengthWithActivity` config property to clients.  This is needed by mobile to determine if/when to schedule local session expiry notifications.

Bonus feature:  `testfiles` added to `.gitignore` per agreement in platform team meeting.

#### Ticket Link

https://mattermost.atlassian.net/browse/MM-24939


wiggin77:
> 0/5, but perhaps the gitignore changes are better done in a separate PR, since it is unrelated to the current change.

Agreed.  But two one line changes is not a lot of review overhead and can be easily backed out.  Easiest review of the day.

wiggin77:
@lindy65 not sure if QA has any method to test this (or any desire to).

lindy65:
Thanks @wiggin77 - when you say "expose" - does this mean there is a setting in the System Console? If so, I can test that...

wiggin77:
> when you say "expose" - does this mean there is a setting in the System Console? If so, I can test that..

Thanks @lindy65 , the config setting is already in the system console via a different PR.  This is to provide the setting in the client config on client startup.

enahum:
A way to verify this is by inspecting the response of the config request in the console network tab of the browser.

The setting should not be present for users who have not login but should be there when the user is logged in. Right?

wiggin77:
> The setting should not be present for users who have not login but should be there when the user is logged in. Right?

Correct.  It should only be there for logged in users.

PR #14520, opened 2020-05-09
MM-25005: Remove HttpRequestsDuration bucket

agnivade:
#### Summary

The ApiTimesHistograms already captures what HttpRequestsDuration does.
And it's more granular than that. So there's no need to duplicate stuff.
Let's remove this one.

#### Ticket Link

https://mattermost.atlassian.net/browse/MM-25005


ethervoid:
What you you think about removing the HttpError counter as well. I think that is a duplicate metric now that we have the status code

agnivade:
I checked that. That uses a slightly different code path. It is logged when the `c.Err` field is set. While that eventually boils down to the same thing, but with the appError migration happening, the `err` field will be disconnected from the status code. So there's no "guarantee" that an application error is actually equivalent to a 5xx status code.

But I'm totally open to removing it if you think it's worth it.

ethervoid:
@agnivade, In that case, we shouldn't name it HttpError, would be something like ServerError. But we can manage that in the future with the App error revamp.

agnivade:
> In that case, we shouldn't name it HttpError, would be something like ServerError.

I agree. Filed https://mattermost.atlassian.net/browse/MM-25020.

agnivade:
@ogi-m - Not sure if there's a way, or even need to test this. This just removes a prometheus counter. The only way to verify would be to actually spin up a prometheus instance and verify that the counter does not appear. Up to you whether you want to verify or not.

Or alternatively, once this gets merged, you can verify it in our production Grafana panel also.

agnivade:
Special thanks to @stylianosrigas for checking the Grafana dashboards !

@amyblais - this needs some documentation in the release notes regarding some Prometheus metrics being removed to be replaced by better ones. Let me know and I can help with the text. Thanks !

PR #14521, opened 2020-05-09
MM-25006: Fix occurences of unclosed http bodies

agnivade:
And while here, we also use ioutil.Discard to read the remaining
body instead of reading with ioutil.ReadAll which allocates a separate
byte buffer.

### Ticket link

https://mattermost.atlassian.net/browse/MM-25006


agnivade:
This was mainly found from the bodyClose linter. But I didn't enable it because there were too many false positives, and it would turn out to be a nuisance.

api4/system.go
```diff
@@ -432,6 +434,10 @@ func getRedirectLocation(c *Context, w http.ResponseWriter, r *http.Request) {
 		w.Write([]byte(model.MapToJson(m)))
 		return
 	}
+	defer func() {
+		io.Copy(ioutil.Discard, res.Body)
```
jupenur, line 438: Potentially stupid question, but why is it necessary to read the full body before closing? What happens if we don't?

agnivade: Yes, I just got asked the same question earlier today :) It is a very common gotcha in Go, and one a lot of devs stumble upon. The answer is in the docs:

> // The http Client and Transport guarantee that Body is always
    // non-nil, even on responses without a body or responses with
    // a zero-length body. It is the caller's responsibility to
    // close Body. The default HTTP client's Transport may not
    // reuse HTTP/1.x "keep-alive" TCP connections if the Body is
    // not read to completion and closed.

https://golang.org/pkg/net/http/#Response

agnivade: Stress on "read to completion".

jupenur: We don't reuse the client, though, so I don't think keep-alive will work in either case?

app/post_metadata.go
```diff
@@ -402,7 +403,10 @@ func (a *App) getLinkMetadata(requestURL string, timestamp int64, isNewPost bool
 	}
 
 	if body != nil {
-		defer body.Close()
+		defer func() {
+			io.Copy(ioutil.Discard, body)
+			body.Close()
```
hanzei, line 408: 0/5 Maybe move this into a helper function similar to the one in `services/marketplace/client.go`?

services/marketplace/client.go
```diff
@@ -79,7 +80,7 @@ func (c *Client) GetPlugin(filter *model.MarketplacePluginFilter, pluginVersion
 // closeBody ensures the Body of an http.Response is properly closed.
 func closeBody(r *http.Response) {
 	if r.Body != nil {
-		_, _ = ioutil.ReadAll(r.Body)
+		_, _ = io.Copy(ioutil.Discard, r.Body)
```
hanzei, line 83: What is the difference between discarding and reading into `_`? I would assume that "it does the same".

app/admin.go
```diff
@@ -177,6 +178,10 @@ func (a *App) TestSiteURL(siteURL string) *model.AppError {
 	if err != nil || res.StatusCode != 200 {
 		return model.NewAppError("testSiteURL", "app.admin.test_site_url.failure", nil, "", http.StatusBadRequest)
 	}
+	defer func() {
+		io.Copy(ioutil.Discard, res.Body)
+		res.Body.Close()
```
hanzei, line 183: 1/5. By using 
```suggestion
		_, _ = ioutil.ReadAll(r.Body)
		_ = r.Body.Close()
```
the code would make clear that any errors are ignored on purpose.

api4/system.go
hanzei: Is there a cost on reading the whole `body`?

jupenur: @hanzei If the server is slow to respond, the read will block until we hit the client's timeout -- I think. It won't matter for the end-user since we use `defer`, but the file descriptor won't be freed immediately.

Of course in this specific case a well-behaving server shouldn't return any body at all, but that's not something we can rely on.

services/marketplace/client.go
jupenur: `io.Copy(ioutil.Discard, r.Body)` discards without allocating. `ReadAll` allocates a buffer.

api4/system.go
agnivade: Yeah, we don't reuse the client, which doesn't really make sense. We should. A client is a heavy object which maintains the connections, so initializing it every time has a cost.

It is generally _highly_ recommended to read the body to completion, and there are several articles around it. But let me know if it's not necessary here.

For non-GET cases, we anyways are reading the body, so the same question about misbehaving server applies there too. This just ensures that there is no remaining bytes left to read. Most of the cases, there will be none left to read, and this will usually turn out to be a noop.

services/marketplace/client.go
agnivade: Yes, please see: https://github.com/mattermost/mattermost-server/pull/14521#issue-415524611

agnivade:
> Are the false positive reports worth an issue there?

I am not sure, but I think it's generally very hard to detect these without false positives. The body can be closed totally somewhere else in a different function. And sometimes, they are not meant to be closed at all, like in websocket upgrade.

api4/system.go
jupenur: The only potential issue I see is regarding file descriptors as mentioned above, but I don't think even that's a problem since a malicious server could just as easily attempt to serve the response headers slowly. Marking this as resolved.

agnivade: I think that is a problem very specific to this HEAD case. In the general GET case, a server may send the rest of the body also slowly. But yes, even in the HEAD case, the headers can be sent slowly. :)

hanzei:
There are a couple of places where the body isn't read to completion e.g. https://github.com/mattermost/mattermost-server/blob/744ec281d2c8edb6fe24afc46dabd93f93d1a4ab/app/integration_action.go#L405. Is this worth a ticket?

agnivade:
Yes I think so. Most of the times, the body will already have been read. So this is more of a safety measure in case it isn't.

app/post_metadata.go
agnivade: This is slightly different because it works on the Body directly rather than from the request. So it will be a one-off function. Let's keep it here for now.

app/admin.go
agnivade: done

amyblais:
@agnivade Looks like the Jira ticket is for v5.24 but this PR hasn't been cherry-picked yet.

agnivade:
This is not necessary for 5.24. I will re-label it for 5.25.

hanzei:
@agnivade If this goes into `5.25.0`, maybe you should leave the `CherryPick/Approved` label to make clear, that the PR needs cherry-picking.

agnivade:
Woops my bad. You are right.

PR #14524, opened 2020-05-09
MM-25095: Handles static JS and CSS using Brotli encoding.

mkraft:
#### Summary

Requests static JS and CSS files in their Brotli format, as built by the webapp.

There's a PR open to merge in a newly added configuration to the gzip package: https://github.com/nytimes/gziphandler/pull/98

#### Ticket Link

https://mattermost.atlassian.net/browse/MM-25095

#### Dependency

https://github.com/mattermost/mattermost-webapp/pull/5466

#### Screenshots

Before:
before

After:
after


mkraft:
force-pushed to add "MM-25095:" prefix to commit messages.

web/static.go
```diff
@@ -30,11 +43,16 @@ func (w *Web) InitStatic() {
 
 		subpath, _ := utils.GetSubpathFromConfig(w.ConfigService.Config())
 
-		staticHandler := staticFilesHandler(http.StripPrefix(path.Join(subpath, "static"), http.FileServer(http.Dir(staticDir))))
+		staticHandler := brotliFilesHandler(staticFilesHandler(http.StripPrefix(path.Join(subpath, "static"), http.FileServer(http.Dir(staticDir)))))
 		pluginHandler := staticFilesHandler(http.StripPrefix(path.Join(subpath, "static", "plugins"), http.FileServer(http.Dir(*w.ConfigService.Config().PluginSettings.ClientDirectory))))
 
 		if *w.ConfigService.Config().ServiceSettings.WebserverMode == "gzip" {
-			staticHandler = gziphandler.GzipHandler(staticHandler)
+			everythingExceptBrotliGzipHandler, err := gziphandler.GzipHandlerWithOpts(gziphandler.ContentTypeExceptions(brotliContentTypes))
```
hmhealey, line 50: Maybe it would be easier to just construct `brotliContentTypes` here instead of having it done in `init`? 0/5

mkraft: Yes I like that more, then it's not initialized if not needed.

PR #14526, opened 2020-05-10
Layer generators error return type fix

reflog:
#### Summary
Fixed a bug in layer generators with handling of error return type (previously, only model.AppError were handled)

#### Ticket Link
Bug discovered during https://github.com/mattermost/mattermost-server/pull/14339


lindy65:
@reflog - can you help with test steps for this please?

reflog:
@lindy65 - unfortunately there isn't much that QA can test here, it's a build-time tool - if server builds work after this change (as in, complete without error) - then this change is working :)

PR #14527, opened 2020-05-10
MM-25011: Remove concurrent write to websocket connection

agnivade:
This test writes directly to a connection
which causes panics and more frustration in an already fragile CI.

Since this anyways checks an edge condition, and will anyways be
removed in v6, let's remove this for now and let CI be happy.

#### Ticket link

https://mattermost.atlassian.net/browse/MM-25011


agnivade:
@lindy65 - This does not really need any testing as it just fixes a unit test. Real code is not affected by this. Thanks.

PR #14539, opened 2020-05-11
Adding Upgrade to Enterprise version feature

jespino:
#### Summary
Adding Upgrade to Enterprise version feature

### DIFF
--- a/api4/system.go
+++ b/api4/system.go
@@ -50,6 +50,9 @@ func (api *API) InitSystem() {
 	api.BaseRoutes.ApiRoot.Handle("/server_busy", api.ApiSessionRequired(setServerBusy)).Methods("POST")
 	api.BaseRoutes.ApiRoot.Handle("/server_busy", api.ApiSessionRequired(getServerBusyExpires)).Methods("GET")
 	api.BaseRoutes.ApiRoot.Handle("/server_busy", api.ApiSessionRequired(clearServerBusy)).Methods("DELETE")
+	api.BaseRoutes.ApiRoot.Handle("/upgrade_to_enterprise", api.ApiSessionRequired(upgradeToEnterprise)).Methods("POST")
+	api.BaseRoutes.ApiRoot.Handle("/upgrade_to_enterprise/status", api.ApiSessionRequired(upgradeToEnterpriseStatus)).Methods("GET")
+	api.BaseRoutes.ApiRoot.Handle("/restart", api.ApiSessionRequired(restart)).Methods("POST")
 }
 
 func getSystemPing(c *Context, w http.ResponseWriter, r *http.Request) {
@@ -544,4 +547,62 @@ func getServerBusyExpires(c *Context, w http.ResponseWriter, r *http.Request) {
 		return
 	}
 	w.Write([]byte(c.App.Srv().Busy.ToJson()))
+}
+
+func upgradeToEnterprise(c *Context, w http.ResponseWriter, r *http.Request) {
+	if !c.App.SessionHasPermissionTo(*c.App.Session(), model.PERMISSION_MANAGE_SYSTEM) {
+		c.SetPermissionError(model.PERMISSION_MANAGE_SYSTEM)
+		return
+	}
+
+	auditRec := c.MakeAuditRecord("upgradeToEnterprise", audit.Fail)
+	defer c.LogAuditRec(auditRec)
+
+	percentage, _ := c.App.Srv().UpgradeToE0Status()
+	if percentage > 0 {
+		c.Err = model.NewAppError("upgradeToEnterprise", "api.upgrade_to_enterprise.app_error", nil, "", http.StatusTooManyRequests)
+		return
+	}
+
+	c.App.Srv().Go(func() {
+		c.App.Srv().UpgradeToE0()
+	})
+
+	auditRec.Success()
+	w.WriteHeader(http.StatusAccepted)
+	ReturnStatusOK(w)
+}
+
+func upgradeToEnterpriseStatus(c *Context, w http.ResponseWriter, r *http.Request) {
+	if !c.App.SessionHasPermissionTo(*c.App.Session(), model.PERMISSION_MANAGE_SYSTEM) {
+		c.SetPermissionError(model.PERMISSION_MANAGE_SYSTEM)
+		return
+	}
+
+	percentage, err := c.App.Srv().UpgradeToE0Status()
+	var s map[string]interface{}
+	if err != nil {
+		s = map[string]interface{}{"percentage": 0, "error": err.Error()}
+	} else {
+		s = map[string]interface{}{"percentage": percentage, "error": nil}
+	}
+
+	w.Write([]byte(model.StringInterfaceToJson(s)))
+}
+
+func restart(c *Context, w http.ResponseWriter, r *http.Request) {
+	if !c.App.SessionHasPermissionTo(*c.App.Session(), model.PERMISSION_MANAGE_SYSTEM) {
+		c.SetPermissionError(model.PERMISSION_MANAGE_SYSTEM)
+		return
+	}
+
+	auditRec := c.MakeAuditRecord("restartServer", audit.Fail)
+	auditRec.Success()
+	c.LogAuditRec(auditRec)
+	ReturnStatusOK(w)
+	time.Sleep(1 * time.Second)
+
+	go func() {
+		c.App.Srv().Restart()
+	}()
 }
--- a/services/upgrader/upgrader.go
+++ b/services/upgrader/upgrader.go
@@ -1,0 +1,176 @@
+package upgrader
+
+import (
+	"archive/tar"
+	"compress/gzip"
+	"errors"
+	"io"
+	"io/ioutil"
+	"net/http"
+	"os"
+	"path/filepath"
+	"runtime"
+	"sync/atomic"
+
+	"github.com/mattermost/mattermost-server/v5/model"
+)
+
+var upgradePercentage int64
+var upgradeError error
+var upgrading int32
+
+type writeCounter struct {
+	total  int64
+	readed int64
+}
+
+func (wc *writeCounter) Write(p []byte) (int, error) {
+	n := len(p)
+	wc.readed += int64(n)
+	percentage := (wc.readed * 100) / wc.total
+	if percentage == 0 {
+		upgradePercentage = 1
+	} else if percentage == 100 {
+		upgradePercentage = 99
+	} else {
+		upgradePercentage = percentage
+	}
+	return n, nil
+}
+
+func getUpgradeDirectory() (string, error) {
+	exePath, err := os.Executable()
+	if err != nil {
+		return "", err
+	}
+	return filepath.Dir(exePath), nil
+}
+
+func getCurrentVersionTgzUrl() string {
+	return "https://releases.mattermost.com/" + model.CurrentVersion + "/mattermost-" + model.CurrentVersion + "-linux-amd64.tar.gz"
+}
+
+func canIUpgrade() error {
+	if runtime.GOARCH != "amd64" {
+		return errors.New("Upgrades only supported for amd64 architectures")
+	}
+	return nil
+}
+
+func canIUpgradeToE0() error {
+	if err := canIUpgrade(); err != nil {
+		return err
+	}
+	if model.BuildEnterpriseReady == "true" {
+		return errors.New("You can't upgrade your code to enterprise because you are already in enterprise code")
+	}
+	return nil
+}
+
+func UpgradeToE0() error {
+	if !atomic.CompareAndSwapInt32(&upgrading, 0, 1) {
+		return errors.New("One upgrade is already running.")
+	}
+	defer atomic.CompareAndSwapInt32(&upgrading, 1, 0)
+
+	upgradePercentage = 1
+	upgradeError = nil
+
+	if err := canIUpgradeToE0(); err != nil {
+		upgradePercentage = 0
+		upgradeError = err
+		return err
+	}
+	filename, err := download(getCurrentVersionTgzUrl())
+	if err != nil {
+		if filename != "" {
+			os.Remove(filename)
+		}
+		upgradePercentage = 0
+		upgradeError = err
+		return err
+	}
+	defer os.Remove(filename)
+	executablePath, err := os.Executable()
+	if err != nil {
+		upgradePercentage = 0
+		upgradeError = err
+		return err
+	}
+
+	err = extractBinary(executablePath, filename)
+	if err != nil {
+		upgradePercentage = 0
+		upgradeError = err
+		return err
+	}
+	upgradePercentage = 100
+	return nil
+}
+
+func UpgradeToE0Status() (int64, error) {
+	return upgradePercentage, upgradeError
+}
+
+func download(url string) (string, error) {
+	resp, err := http.Get(url)
+	if err != nil {
+		return "", err
+	}
+	defer resp.Body.Close()
+
+	out, err := ioutil.TempFile("", "*_mattermost.tar.gz")
+	if err != nil {
+		return "", err
+	}
+	defer out.Close()
+
+	counter := &writeCounter{total: resp.ContentLength}
+	_, err = io.Copy(out, io.TeeReader(resp.Body, counter))
+	return out.Name(), err
+}
+
+func extractBinary(executablePath string, filename string) error {
+	gzipStream, err := os.Open(filename)
+	if err != nil {
+		return err
+	}
+
+	uncompressedStream, err := gzip.NewReader(gzipStream)
+	if err != nil {
+		return err
+	}
+
+	tarReader := tar.NewReader(uncompressedStream)
+
+	for true {
+		header, err := tarReader.Next()
+
+		if err == io.EOF {
+			return errors.New("Unable to find mattermost binary in the downloaded version")
+		}
+
+		if err != nil {
+			return err
+		}
+
+		if header.Typeflag == tar.TypeReg && header.Name == "mattermost/bin/mattermost" {
+			os.Rename(executablePath, executablePath+".bak")
+			outFile, err := os.Create(executablePath)
+			if err != nil {
+				os.Rename(executablePath+".bak", executablePath)
+				return err
+			}
+			defer outFile.Close()
+			if _, err := io.Copy(outFile, tarReader); err != nil {
+				os.Remove(executablePath)
+				os.Rename(executablePath+".bak", executablePath)
+				return err
+			}
+			os.Remove(executablePath + ".bak")
+			os.Chmod(executablePath, 0755)
+			break
+		}
+	}
+	return nil
+}
--- a/app/server.go
+++ b/app/server.go
@@ -12,9 +12,11 @@ import (
 	"net/http"
 	"net/url"
 	"os"
+	"os/exec"
 	"strings"
 	"sync"
 	"sync/atomic"
+	"syscall"
 	"time"
 
 	"github.com/gorilla/mux"
@@ -40,6 +42,7 @@ import (
 	"github.com/mattermost/mattermost-server/v5/services/searchengine"
 	"github.com/mattermost/mattermost-server/v5/services/timezones"
 	"github.com/mattermost/mattermost-server/v5/services/tracing"
+	"github.com/mattermost/mattermost-server/v5/services/upgrader"
 	"github.com/mattermost/mattermost-server/v5/store"
 	"github.com/mattermost/mattermost-server/v5/store/sqlstore"
 	"github.com/mattermost/mattermost-server/v5/utils"
@@ -470,6 +473,30 @@ func (s *Server) Shutdown() error {
 
 	mlog.Info("Server stopped")
 	return nil
+}
+
+func (s *Server) Restart() error {
+	s.Shutdown()
+
+	argv0, err := exec.LookPath(os.Args[0])
+	if err != nil {
+		return err
+	}
+
+	if _, err = os.Stat(argv0); err != nil {
+		return err
+	}
+
+	mlog.Info("Restarting server")
+	return syscall.Exec(argv0, os.Args, os.Environ())
+}
+
+func (s *Server) UpgradeToE0() error {
+	return upgrader.UpgradeToE0()
+}
+
+func (s *Server) UpgradeToE0Status() (int64, error) {
+	return upgrader.UpgradeToE0Status()
 }
 
 // Go creates a goroutine, but maintains a record of it to ensure that execution completes before


services/upgrader/upgrader.go
```diff
@@ -0,0 +1,176 @@
...
+	}
+	return filepath.Dir(exePath), nil
+}
+
+func getCurrentVersionTgzUrl() string {
+	return "https://releases.mattermost.com/" + model.CurrentVersion + "/mattermost-" + model.CurrentVersion + "-linux-amd64.tar.gz"
+}
+
+func canIUpgrade() error {
+	if runtime.GOARCH != "amd64" {
```
coreyhulen, line 54: Shouldn't we also check for `GOOS = linux` ?  What happens on mac ?

jespino: good point, I think this should work in all the operating system, but we should be getting the right build. I'll check it.

jespino: Ok, I think we only distribute the linux version for the server, so... yes, this feature should be only for linux.

jespino: Done

api4/system.go
```diff
@@ -545,3 +548,70 @@ func getServerBusyExpires(c *Context, w http.ResponseWriter, r *http.Request) {
...
 }
+
+func upgradeToEnterprise(c *Context, w http.ResponseWriter, r *http.Request) {
+	if !c.App.SessionHasPermissionTo(*c.App.Session(), model.PERMISSION_MANAGE_SYSTEM) {
+		c.SetPermissionError(model.PERMISSION_MANAGE_SYSTEM)
+		return
+	}
+
+	auditRec := c.MakeAuditRecord("upgradeToEnterprise", audit.Fail)
+	defer c.LogAuditRec(auditRec)
```
catalintomai, line 559: Is there a reason for auditing the request after the permission check?

api4/system.go
```diff
@@ -545,3 +548,70 @@ func getServerBusyExpires(c *Context, w http.ResponseWriter, r *http.Request) {
...
+
+	percentage, _ := c.App.Srv().UpgradeToE0Status()
```
catalintomai, line 561: maybe call the `UpgradeToE0Status()` after checking for `BuildEnterpriseReady == true` ?

services/upgrader/upgrader.go
```diff
@@ -0,0 +1,179 @@
...
+func UpgradeToE0() error {
+	if !atomic.CompareAndSwapInt32(&upgrading, 0, 1) {
+		return errors.New("One upgrade is already running.")
+	}
+	defer atomic.CompareAndSwapInt32(&upgrading, 1, 0)
+
+	upgradePercentage = 1
+	upgradeError = nil
+
+	if err := canIUpgradeToE0(); err != nil {
```
catalintomai, line 82: would it make sense to call this at the top of this method?

api4/system.go
```diff
@@ -545,3 +548,70 @@ func getServerBusyExpires(c *Context, w http.ResponseWriter, r *http.Request) {
...
+	if !c.App.SessionHasPermissionTo(*c.App.Session(), model.PERMISSION_MANAGE_SYSTEM) {
+		c.SetPermissionError(model.PERMISSION_MANAGE_SYSTEM)
+		return
+	}
+
+	percentage, err := c.App.Srv().UpgradeToE0Status()
+	var s map[string]interface{}
+	if err != nil {
+		appErr := model.NewAppError("upgradeToEnterpriseStatus", "api.upgrade_to_enterprise_status.app_error", nil, err.Error(), http.StatusTooManyRequests)
+		s = map[string]interface{}{"percentage": 0, "error": appErr.Error()}
```
catalintomai, line 594: don't know if it would be useful or not to capture/log/surface that there was an error at a certain percentage, maybe if we can correlate with the file size?

services/upgrader/upgrader.go
```diff
@@ -0,0 +1,179 @@
...
+		upgradePercentage = 0
+		upgradeError = err
+		return err
+	}
+	filename, err := download(getCurrentVersionTgzUrl())
```
catalintomai, line 87: If I understand correctly, 0-99 percentage is the download and the rest to 100 is the unzip/creating the final binary.  I guess that is an ok approx. , given that network download time normally takes the most, maybe in the future we can refine the info we give to the user and let him know that upgrade time is actually 99% download time.

services/upgrader/upgrader.go
jespino: yes, probably can be cleaner.

services/upgrader/upgrader.go
jespino: Yes, it is more about provide a more or less accurate percentage while he is waiting, but I don't see much value on adding correctness here, or extra info here, because is just a "progress bar/percentage of completion" that is not so important. It is better to have it correct, or to let the user know what is exactly going on, yes, but removing effort from other places to put it here, I don't see it as the best way to invest the effort.

api4/system.go
jespino: I think was based on existing code, and I've seen that there are `MakeAuditRecord` calls that are done before the permissions check and other after. I asked in the developers-server channel about it, but for now I going to move it before the permissions check.

api4/system.go
jespino: I can add that information inside the error when makes sense. For example, only as part of the "downloading error". (See the commit that I'm about to submit).

services/upgrader/upgrader.go
jespino: Done

api4/system.go
jespino: Done

api4/system.go
jespino: Done

api4/system.go
jespino: Done

services/upgrader/upgrader.go
catalintomai: Agree this is not a priority, it is more that it is a misleading message, if we call this upgrade progress and 99% is download (e.g. if it breaks, it might be important for them to know where it broke - in download or when deploying the binaries - they are different and should be handled differently.)

jespino: Actually, the correct error is reported as part of the status if there was an error. so this give you the information about what was the error about. In this case the error is about downloading, and if it fails extracting the binary is going to faile with the proper error.

jespino:
@srkgupta that looks like problems with the difference between enterprise and mattermost-server. Can you try again (getting the last version of the enterprise repo)?

srkgupta:
> @srkgupta that looks like problems with the difference between enterprise and mattermost-server. Can you try again (getting the last version of the enterprise repo)?

But in this case, we need to test on Team Edition, right. Hence ideally it would not have the enterprise folder itself. Else it will be a E0 instance and the workflow will be different, i.e. to request the trial. Hence this should work even without the enterprise folder.

jespino:
Yes, exactly, this should work without the enterprise folder, but be careful, the enterprise link to `../enterprise` is created automatically if you have the `../enterprise` directory. By the way, you need to modify too the `model/version.go` file to set the current version to an existing published stable version, because if not, you have no way to get the "enterprise" compiled version.

srkgupta:
> Yes, exactly, this should work without the enterprise folder, but be careful, the enterprise link to `../enterprise` is created automatically if you have the `../enterprise` directory. By the way, you need to modify too the `model/version.go` file to set the current version to an existing published stable version, because if not, you have no way to get the "enterprise" compiled version.

Ok thanks for the info. Let me recheck this.

jespino:
Adding @mgdelacroix to test it locally in a linux system.

i18n/en.json
```diff
@@ -2482,6 +2482,42 @@
...
+    "id": "api.upgrade_to_enterprise.invalid-user.app_error",
+    "translation": "Unable to write the executable file.\n\nPlease give permissions to the {{.MattermostUsername}} user to write the file {{.Path}}. To do that, you can execute the commands:\n\n```\nchown {{.MattermostUsername}} \"{{.Path}}\"\n```\n\n Remember to restore your previous existing permissions after the upgrade is complete executing:\n\n```\nchown {{.FileUsername}} \"{{.Path}}\"\n```"
+  },
+  {
+    "id": "api.upgrade_to_enterprise.system_not_supported.app_error",
+    "translation": "The system wasn't able to upgrade to enterprise, this feature is only available for linux-amd64 systems."
+  },
+  {
+    "id": "api.upgrade_to_enterprise_status.app_error",
+    "translation": "The system wasn't able to upgrade to enterprise, please try again."
```
esadur, line 2519:
```suggestion
    "translation": "Mattermost was unable to upgrade to Enterprise Edition."
```

i18n/en.json
```diff
@@ -2482,6 +2482,42 @@
...
+    "id": "api.upgrade_to_enterprise.generic_error.app_error",
+    "translation": "The system wasn't able to upgrade to enterprise."
+  },
+  {
+    "id": "api.upgrade_to_enterprise.invalid-permission.app_error",
+    "translation": "Unable to write the executable file.\n\nPlease give permissions to the {{.MattermostUsername}} user to write the file {{.Path}}. To do that, you can execute the commands:\n\n```\nchmod +w \"{{.Path}}\"\n```\n\n Remember to restore your previous existing permissions after the upgrade is complete executing:\n\n```\nchmod -w \"{{.Path}}\"\n```"
+  },
+  {
+    "id": "api.upgrade_to_enterprise.invalid-user-and-permission.app_error",
+    "translation": "Unable to write the executable file.\n\nPlease give permissions to the {{.MattermostUsername}} user to write the file {{.Path}}. To do that, you can execute the commands:\n\n```\nchown {{.MattermostUsername}} \"{{.Path}}\"\nchmod +w \"{{.Path}}\"\n```\n\n Remember to restore your previous existing permissions after the upgrade is complete executing:\n\n```\nchown {{.FileUsername}} \"{{.Path}}\"\nchmod -w \"{{.Path}}\"\n```"
```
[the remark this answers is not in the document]
jespino: Done

i18n/en.json
```diff
@@ -2482,6 +2482,42 @@
...
+  },
+  {
+    "id": "api.upgrade_to_enterprise.invalid-user.app_error",
+    "translation": "Unable to write the executable file.\n\nPlease give permissions to the {{.MattermostUsername}} user to write the file {{.Path}}. To do that, you can execute the commands:\n\n```\nchown {{.MattermostUsername}} \"{{.Path}}\"\n```\n\n Remember to restore your previous existing permissions after the upgrade is complete executing:\n\n```\nchown {{.FileUsername}} \"{{.Path}}\"\n```"
+  },
+  {
+    "id": "api.upgrade_to_enterprise.system_not_supported.app_error",
+    "translation": "The system wasn't able to upgrade to enterprise, this feature is only available for linux-amd64 systems."
```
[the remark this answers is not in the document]
jespino: Done

i18n/en.json
```diff
@@ -2482,6 +2482,42 @@
...
+  },
+  {
+    "id": "api.upgrade_to_enterprise.invalid-user-and-permission.app_error",
+    "translation": "Unable to write the executable file.\n\nPlease give permissions to the {{.MattermostUsername}} user to write the file {{.Path}}. To do that, you can execute the commands:\n\n```\nchown {{.MattermostUsername}} \"{{.Path}}\"\nchmod +w \"{{.Path}}\"\n```\n\n Remember to restore your previous existing permissions after the upgrade is complete executing:\n\n```\nchown {{.FileUsername}} \"{{.Path}}\"\nchmod -w \"{{.Path}}\"\n```"
+  },
+  {
+    "id": "api.upgrade_to_enterprise.invalid-user.app_error",
+    "translation": "Unable to write the executable file.\n\nPlease give permissions to the {{.MattermostUsername}} user to write the file {{.Path}}. To do that, you can execute the commands:\n\n```\nchown {{.MattermostUsername}} \"{{.Path}}\"\n```\n\n Remember to restore your previous existing permissions after the upgrade is complete executing:\n\n```\nchown {{.FileUsername}} \"{{.Path}}\"\n```"
```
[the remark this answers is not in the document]
jespino: Done

i18n/en.json
```diff
@@ -2482,6 +2482,42 @@
...
+    "id": "api.upgrade_to_enterprise.app_error",
+    "translation": "There is an upgrade to enterprise already running, only one upgrade is possible at the same time."
+  },
+  {
+    "id": "api.upgrade_to_enterprise.generic_error.app_error",
+    "translation": "The system wasn't able to upgrade to enterprise."
+  },
+  {
+    "id": "api.upgrade_to_enterprise.invalid-permission.app_error",
+    "translation": "Unable to write the executable file.\n\nPlease give permissions to the {{.MattermostUsername}} user to write the file {{.Path}}. To do that, you can execute the commands:\n\n```\nchmod +w \"{{.Path}}\"\n```\n\n Remember to restore your previous existing permissions after the upgrade is complete executing:\n\n```\nchmod -w \"{{.Path}}\"\n```"
```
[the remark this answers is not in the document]
jespino: Done

amyblais:
@mgdelacroix Kind reminder to review this

jespino:
@DSchalla can you take a look into this?

services/upgrader/upgrader.go
```diff
@@ -0,0 +1,250 @@
...
+			os.Remove(filename)
+		}
+		upgradeError = fmt.Errorf("error downloading the new version (percentage: %d)", upgradePercentage)
+		mlog.Error("Unable to download the mattermost server version", mlog.Int64("percentage", upgradePercentage), mlog.String("url", getCurrentVersionTgzUrl()), mlog.Err(err))
+		upgradePercentage = 0
+		return err
+	}
+	defer os.Remove(filename)
+
+	err = extractBinary(executablePath, filename)
```
DSchalla, line 174: After downloading, how do we ensure this is an archive from Mattermost?  After downloading, we should ensure that two things are correct:

1. The SHA256 hash
2. The GPG signature

Otherwise we are opening ourselves up to any MitM attacks.

services/upgrader/upgrader.go
```diff
@@ -0,0 +1,250 @@
...
+	defer resp.Body.Close()
+
+	out, err := ioutil.TempFile("", "*_mattermost.tar.gz")
+	if err != nil {
+		return "", err
+	}
+	defer out.Close()
+
+	counter := &writeCounter{total: resp.ContentLength}
+	_, err = io.Copy(out, io.TeeReader(resp.Body, counter))
```
DSchalla, line 203: Should we wrap the downloaded file in a https://golang.org/pkg/io/#LimitedReader to avoid a potential DoS attack?

services/upgrader/upgrader.go
```diff
@@ -0,0 +1,250 @@
...
+				return err
+			}
+			defer outFile.Close()
+			if _, err := io.Copy(outFile, tarReader); err != nil {
+				os.Remove(executablePath)
+				os.Rename(executablePath+".bak", executablePath)
+				return err
+			}
+			os.Remove(executablePath + ".bak")
+			os.Chmod(executablePath, 0755)
```
DSchalla, line 245: System administrators may use a different config than `0755` for their binary, can we read the permissions before and restore it for the new file?

services/upgrader/upgrader.go
```diff
@@ -0,0 +1,250 @@
...
+		if err == io.EOF {
+			return errors.New("Unable to find mattermost binary in the downloaded version")
+		}
+
+		if err != nil {
+			return err
+		}
+
+		if header.Typeflag == tar.TypeReg && header.Name == "mattermost/bin/mattermost" {
+			os.Rename(executablePath, executablePath+".bak")
```
DSchalla, line 232: What if this errors (e.g. file already exists)?

services/upgrader/upgrader.go
```diff
@@ -0,0 +1,250 @@
...
+			outFile, err := os.Create(executablePath)
+			if err != nil {
+				os.Rename(executablePath+".bak", executablePath)
+				return err
+			}
+			defer outFile.Close()
+			if _, err := io.Copy(outFile, tarReader); err != nil {
+				os.Remove(executablePath)
```
DSchalla, line 240: See above (but doesn't exist).

services/upgrader/upgrader.go
```diff
@@ -0,0 +1,250 @@
...
+				os.Rename(executablePath+".bak", executablePath)
```
DSchalla, line 241: See above.

services/upgrader/upgrader.go
jespino: Good point

services/upgrader/upgrader.go
jespino: I guess, so, I'll take a look.

services/upgrader/upgrader.go
jespino: Yes, sure.

services/upgrader/upgrader.go
jespino: I'll take a look.

services/upgrader/upgrader.go
jespino: If doesn't exists, shouldn't be a problem, but any other error can be a problem. Let me check.

services/upgrader/upgrader.go
jespino: I'll check it

services/upgrader/upgrader.go
jespino: @DSchalla shouldn't be enough the GPG signature? I mean, if the signature is ok, the binary is generated by us, that verifies 2 things, the data integrity and the author.

DSchalla: @jespino Yes, you are absolutely correct since we sign our actual archives. :+1:

jespino: This is now done.

services/upgrader/upgrader.go
jespino: There is only one upgrader running at the same time, and the upgrader here is the one that is requesting the url, always a mattermost server version tar.gz file, that is going to be a big file, but unless you contaminate the DNS or do a man in the middle attack, you can't point that to other file. Anyway I can limit it to something like 300MB (enough for make our binary grow in the future without suddenly break it)

jespino: Done, I added a limit of 300MB for the tar file and 1K for the signature, more than enough in both cases.

services/upgrader/upgrader.go
jespino: Actually, this should have the right permissions from the tar file.

services/upgrader/upgrader.go
jespino: I changed the whole approach here, to save the bytes of the binary in memory, and then, restore it if the process fails. The probability of this failing in the middle of the upgrade is really really low. And maybe we can mitigate this a bit more, but I don't see the value.

jespino: Ok, I moved back to my previous approach. But I have to check it some other things.

jespino:
I still have to include permissions detection for the parent directory to be able to create the new file and change the permissions for that file.

services/upgrader/upgrader.go
```diff
@@ -0,0 +1,372 @@
...
+
+		if header.Typeflag == tar.TypeReg && header.Name == "mattermost/bin/mattermost" {
+			permissions := getFilePermissionsOrDefault(executablePath, 0755)
+			tmpFile, err := ioutil.TempFile("", "*")
+			if err != nil {
+				return err
+			}
+			tmpFileName := tmpFile.Name()
+			os.Remove(tmpFileName)
+			err = os.Rename(executablePath, tmpFileName)
```
jupenur, line 332: This looks like a race. Not sure how winnable, but nevertheless. Consider the following:

 - An attacker monitors `/tmp` as the unprivileged user `evil`
 - The upgrader creates `tmpFile` and sets `tmpFileName` to `/tmp/6b2453a7-4351-4c73-8d6d-4614c9711c4d`
 - Immediately after the `os.Remove` call, `evil` creates a *world-writable directory* at `/tmp/6b2453a7-4351-4c73-8d6d-4614c9711c4d`
 - The upgrader proceeds to the `os.Rename` call, which succeeds because instead of renaming the file it ends up moving the file to the attacker-created directory
 - Now `evil` controls the backup and if the upgrade fails, a malicious copy may be restored

jespino: @jupenur changing it to use a temp file in the same mattermost directory instead of the tmp directory would address your concerns?

jupenur: It would, yeah. Another alternative would be to use `ioutl.TempDir` instead of `ioutil.TempFile` so that the `os.Remove` call can be done away with.

jespino:
@DSchalla @esadur I need your explicit approval to get this merged

services/upgrader/upgrader_linux.go
```diff
@@ -0,0 +1,332 @@
...
+}
+
+func CanIUpgradeToE0() error {
+	if err := canIUpgrade(); err != nil {
+		mlog.Error("Unable to upgrade from TE to E0", mlog.Err(err))
+		return err
+	}
+	if model.BuildEnterpriseReady == "true" {
+		mlog.Warn("Unable to upgrade from TE to E0 because the server is already in E0")
+		return errors.New("You can't upgrade your code to enterprise because you are already in enterprise code")
```
esadur, line 159:
```suggestion
		return errors.New("You cannot upgrade your server from TE to E0 because you are already running Mattermost Enterprise Edition.")
```

services/upgrader/upgrader_linux.go
```diff
@@ -0,0 +1,332 @@
...
+	}
+	return nil
+}
+
+func UpgradeToE0() error {
+	if !atomic.CompareAndSwapInt32(&upgrading, 0, 1) {
+		mlog.Warn("Trying to upgrade while other upgrade is running")
```
esadur, line 166:
```suggestion
		mlog.Warn("Trying to upgrade while another upgrade is running")
```

services/upgrader/upgrader_linux.go
```diff
@@ -0,0 +1,332 @@
...
+		return errors.New("One upgrade is already running.")
```
esadur, line 167:
```suggestion
		return errors.New("Another upgrade is already running.")
```

services/upgrader/upgrader_linux.go
```diff
@@ -0,0 +1,332 @@
...
+	defer atomic.CompareAndSwapInt32(&upgrading, 1, 0)
+
+	upgradePercentage = 1
+	upgradeError = nil
+
+	executablePath, err := os.Executable()
+	if err != nil {
+		upgradePercentage = 0
+		upgradeError = errors.New("error getting the executable path")
+		mlog.Error("Unable to get the mattermost executable path", mlog.Err(err))
```
esadur, line 178:
```suggestion
		mlog.Error("Unable to get the path of the Mattermost executable", mlog.Err(err))
```

services/upgrader/upgrader_linux.go
```diff
@@ -0,0 +1,332 @@
...
+	}
+	defer atomic.CompareAndSwapInt32(&upgrading, 1, 0)
+
+	upgradePercentage = 1
+	upgradeError = nil
+
+	executablePath, err := os.Executable()
+	if err != nil {
+		upgradePercentage = 0
+		upgradeError = errors.New("error getting the executable path")
```
esadur, line 177:
```suggestion
		upgradeError = errors.New("Error getting the path of the executable")
```

services/upgrader/upgrader_linux.go
```diff
@@ -0,0 +1,332 @@
...
+		return err
+	}
+
+	filename, err := download(getCurrentVersionTgzUrl(), 1024*1024*300)
+	if err != nil {
+		if filename != "" {
+			os.Remove(filename)
+		}
+		upgradeError = fmt.Errorf("error downloading the new version (percentage: %d)", upgradePercentage)
+		mlog.Error("Unable to download the mattermost server version", mlog.Int64("percentage", upgradePercentage), mlog.String("url", getCurrentVersionTgzUrl()), mlog.Err(err))
```
esadur, line 188:
```suggestion
		mlog.Error("Unable to download the Mattermost server binary file", mlog.Int64("percentage", upgradePercentage), mlog.String("url", getCurrentVersionTgzUrl()), mlog.Err(err))
```

services/upgrader/upgrader_linux.go
```diff
@@ -0,0 +1,332 @@
...
+		mlog.Error("Unable to get the mattermost executable path", mlog.Err(err))
+		return err
+	}
+
+	filename, err := download(getCurrentVersionTgzUrl(), 1024*1024*300)
+	if err != nil {
+		if filename != "" {
+			os.Remove(filename)
+		}
+		upgradeError = fmt.Errorf("error downloading the new version (percentage: %d)", upgradePercentage)
```
esadur, line 187:
```suggestion
		upgradeError = fmt.Errorf("Error downloading the new Mattermost server binary file (percentage: %d)", upgradePercentage)
```

services/upgrader/upgrader_linux.go
```diff
@@ -0,0 +1,332 @@
...
+		upgradePercentage = 0
+		return err
+	}
+	defer os.Remove(filename)
+	sigfilename, err := download(getCurrentVersionTgzUrl()+".sig", 1024)
+	if err != nil {
+		if sigfilename != "" {
+			os.Remove(sigfilename)
+		}
+		upgradeError = errors.New("error downloading the new version signature file")
```
esadur, line 198:
```suggestion
		upgradeError = errors.New("Error downloading the signature file of the new server")
```

services/upgrader/upgrader_linux.go
```diff
@@ -0,0 +1,332 @@
...
+		mlog.Error("Unable to download the mattermost server version signature file", mlog.String("url", getCurrentVersionTgzUrl()+".sig"), mlog.Err(err))
```
esadur, line 199:
```suggestion
		mlog.Error("Unable to download the signature file of the new Mattermost server", mlog.String("url", getCurrentVersionTgzUrl()+".sig"), mlog.Err(err))
```

services/upgrader/upgrader_linux.go
```diff
@@ -0,0 +1,332 @@
...
+		upgradePercentage = 0
+		return err
+	}
+	defer os.Remove(sigfilename)
+
+	err = verifySignature(filename, sigfilename, mattermostBuildPublicKey)
+	if err != nil {
+		upgradePercentage = 0
+		upgradeError = errors.New("Unable to verify downloaded file signature")
```
esadur, line 208:
```suggestion
		upgradeError = errors.New("Unable to verify the signature of the downloaded file")
```

services/upgrader/upgrader_linux.go
```diff
@@ -0,0 +1,332 @@
...
+		mlog.Error("Unable to verify downloaded file signature", mlog.Err(err))
```
esadur, line 209:
```suggestion
		mlog.Error("Unable to verify the signature of the downloaded file", mlog.Err(err))
```

services/upgrader/upgrader_linux.go
```diff
@@ -0,0 +1,332 @@
...
+		return err
+	}
+
+	tarReader := tar.NewReader(uncompressedStream)
+
+	for {
+		header, err := tarReader.Next()
+
+		if err == io.EOF {
+			return errors.New("Unable to find mattermost binary in the downloaded version")
```
esadur, line 277:
```suggestion
			return errors.New("Unable to find the Mattermost binary in the downloaded version")
```

What does "version" mean here? Would "file" be better? "Resource"? ".tar file"?

services/upgrader/upgrader_linux.go
```diff
@@ -0,0 +1,332 @@
...
+			os.Remove(tmpFileName)
+			err = os.Rename(executablePath, tmpFileName)
+			if err != nil {
+				return err
+			}
+			outFile, err := os.Create(executablePath)
+			if err != nil {
+				err2 := os.Rename(tmpFileName, executablePath)
+				if err2 != nil {
+					mlog.Critical("Unable to restore the executable backup. Restore the executable manually.")
```
esadur, line 300:
```suggestion
					mlog.Critical("Unable to restore the backup of the executable file. Restore the executable file manually.")
```

services/upgrader/upgrader_linux.go
```diff
@@ -0,0 +1,332 @@
...
+					return errors.Wrap(err2, "critical error: unable to upgrade the binary or restore the old binary version. Please restore it manually")
```
esadur, line 301:
```suggestion
					return errors.Wrap(err2, "Critical error: unable to upgrade the binary or restore the old binary version. Please restore it manually.")
```

Can we add more details or instructions for such a critical error?

services/upgrader/upgrader_linux.go
```diff
@@ -0,0 +1,332 @@
...
+				}
+				return err
+			}
+			defer outFile.Close()
+			if _, err = io.Copy(outFile, tarReader); err != nil {
+				err2 := os.Remove(executablePath)
+				if err2 != nil {
+					mlog.Critical("Unable to restore the executable backup. Restore the executable manually.")
```
esadur, line 309:
```suggestion
					mlog.Critical("Unable to restore the backup of the executable file. Restore the executable file manually.")
```

services/upgrader/upgrader_linux.go
```diff
@@ -0,0 +1,332 @@
...
+					return errors.Wrap(err2, "critical error: unable to upgrade the binary or restore the old binary version. Please restore it manually")
```
esadur, line 310:
```suggestion
					return errors.Wrap(err2, "Critical error: unable to upgrade the binary or restore the old binary version. Please restore it manually.")
```

services/upgrader/upgrader_linux.go
```diff
@@ -0,0 +1,332 @@
...
+				}
+
+				err2 = os.Rename(tmpFileName, executablePath)
+				if err2 != nil {
+					mlog.Critical("Unable to restore the executable backup. Restore the executable manually.")
```
esadur, line 315:
```suggestion
					mlog.Critical("Unable to restore the backup of the executable file. Restore the executable file manually.")
```

services/upgrader/upgrader_linux.go
```diff
@@ -0,0 +1,332 @@
...
+					return errors.Wrap(err2, "critical error: unable to upgrade the binary or restore the old binary version. Please restore it manually")
```
esadur, line 316:
```suggestion
					return errors.Wrap(err2, "Critical error: unable to upgrade the binary or restore the old binary version. Please restore it manually.")
```

services/upgrader/upgrader_linux.go
```diff
@@ -0,0 +1,332 @@
...
+				}
+				return err
+			}
+			err = os.Remove(tmpFileName)
+			if err != nil {
+				mlog.Warn("Unable to unable to clean up the binary backup file.", mlog.Err(err))
```
esadur, line 322:
```suggestion
				mlog.Warn("Unable to clean up the binary backup file.", mlog.Err(err))
```

services/upgrader/upgrader_linux.go
```diff
@@ -0,0 +1,332 @@
...
+			}
+			err = os.Chmod(executablePath, permissions)
+			if err != nil {
+				mlog.Warn("Unable to unable to set the right permissions to the file.", mlog.Err(err))
```
esadur, line 326:
```suggestion
				mlog.Warn("Unable to set the correct permissions for the file.", mlog.Err(err))
```

When does this error get hit? What are the correct permissions?

api4/system.go
```diff
@@ -548,6 +553,108 @@ func getServerBusyExpires(c *Context, w http.ResponseWriter, r *http.Request) {
...
+		return
+	}
+
+	percentage, err := c.App.Srv().UpgradeToE0Status()
+	var s map[string]interface{}
+	if err != nil {
+		var isErr *upgrader.InvalidSignature
+		switch {
+		case errors.As(err, &isErr):
+			appErr := model.NewAppError("upgradeToEnterpriseStatus", "api.upgrade_to_enterprise_status.app_error", nil, err.Error(), http.StatusBadRequest)
```
esadur, line 627: The red error text has the prefix "upgradeToEnterpriseStatus", which isn't appropriate formatting for an end user. The string shouldn't be included. See image.
upgradetoenterprisestatus text

I'm not sure if this is being set here.

services/upgrader/errors.go
```diff
@@ -0,0 +1,49 @@
...
+
+// InvalidArch indicates that the current operating system or cpu architecture doesn't support upgrades
+type InvalidArch struct{}
+
+func NewInvalidArch() *InvalidArch {
+	return &InvalidArch{}
+}
+
+func (e *InvalidArch) Error() string {
+	return "invalid operating system or processor architecture"
```
esadur, line 16: The red error text we're currently generating isn't concatenated correctly. See image.
error-incorrect-concatenation

i18n/en.json
```diff
@@ -2618,6 +2618,46 @@
...
+    "id": "api.upgrade_to_enterprise.system_not_supported.app_error",
+    "translation": "Mattermost was unable to upgrade to Enterprise Edition. This feature will only work on Linux systems with x86-64 architecture."
+  },
+  {
+    "id": "api.upgrade_to_enterprise_status.app_error",
+    "translation": "Mattermost was unable to upgrade to Enterprise Edition."
+  },
+  {
+    "id": "api.upgrade_to_enterprise_status.signature.app_error",
+    "translation": "The file signature for the downloaded mattermost doesn't match."
```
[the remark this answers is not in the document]
jespino: This is about the verification of the digital signature of the file. That means if the file doesn't match the signature, is not an officially generated valid mattermost binary file.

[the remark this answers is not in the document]
jespino: This is to ensure that nobody is can intercept the petition and replace it with a custom binary.

services/upgrader/upgrader_linux.go
```diff
@@ -0,0 +1,332 @@
...
+		mlog.Error("Unable to verify the mattermost file signature", mlog.Err(err))
+		return NewInvalidSignature()
+	}
+	return nil
+}
+
+func canIWriteTheExecutable() error {
+	executablePath, err := os.Executable()
+	if err != nil {
+		return errors.New("error getting the executable path")
```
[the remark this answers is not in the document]
jespino: It is the standard approach to not capitalize error strings on golang.

services/upgrader/upgrader_linux.go
```diff
@@ -0,0 +1,332 @@
...
+	}
+	executableInfo, err := os.Stat(path.Dir(executablePath))
+	if err != nil {
+		return errors.New("error getting the executable info")
+	}
+	stat, ok := executableInfo.Sys().(*syscall.Stat_t)
+	if !ok {
+		return errors.New("error getting the executable info")
```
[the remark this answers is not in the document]
jespino: It is the standard approach to not capitalize error strings on golang.

services/upgrader/upgrader_linux.go
```diff
@@ -0,0 +1,332 @@
...
+	fileUID := int(stat.Uid)
+	fileUser, err := user.LookupId(strconv.Itoa(fileUID))
+	if err != nil {
+		return errors.New("error getting the executable info")
+	}
+
+	mattermostUID := os.Getuid()
+	mattermostUser, err := user.LookupId(strconv.Itoa(mattermostUID))
+	if err != nil {
+		return errors.New("error getting the executable info")
```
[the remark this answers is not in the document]
jespino: It is the standard approach to not capitalize error strings on golang.

services/upgrader/upgrader_linux.go
```diff
@@ -0,0 +1,332 @@
...
+	}
+	stat, ok := executableInfo.Sys().(*syscall.Stat_t)
+	if !ok {
+		return errors.New("error getting the executable info")
+	}
+	fileUID := int(stat.Uid)
+	fileUser, err := user.LookupId(strconv.Itoa(fileUID))
+	if err != nil {
+		return errors.New("error getting the executable info")
```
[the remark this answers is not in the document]
jespino: It is the standard approach to not capitalize error strings on golang.

services/upgrader/upgrader_linux.go
```diff
@@ -0,0 +1,332 @@
...
+}
+
+func canIWriteTheExecutable() error {
+	executablePath, err := os.Executable()
+	if err != nil {
+		return errors.New("error getting the executable path")
+	}
+	executableInfo, err := os.Stat(path.Dir(executablePath))
+	if err != nil {
+		return errors.New("error getting the executable info")
```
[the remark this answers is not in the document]
jespino: The information about the executable file, metadata of the file. This error shouldn't happen unless there is a problem accessing the filesystem, something that is going to be really rare.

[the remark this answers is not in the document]
jespino: It is the standard approach to not capitalize error strings on golang.

services/upgrader/upgrader_linux.go
jespino: It is the standard approach to not capitalize error strings on golang. I'll change the rest text, but not the `Error` string.

services/upgrader/upgrader_linux.go
jespino: It is the standard approach to not capitalize error strings on golang. I'll change the rest of the text.

services/upgrader/upgrader_linux.go
jespino: It is the standard approach to not capitalize error strings on golang. I'll change the rest of the text.

services/upgrader/upgrader_linux.go
jespino: It is talking about the Mattermost executable file (the binary) in the downloaded .tar.gz file.

services/upgrader/upgrader_linux.go
jespino: It is the standard approach to not capitalize error strings on golang.

services/upgrader/upgrader_linux.go
jespino: It is the standard approach to not capitalize error strings on golang.

services/upgrader/upgrader_linux.go
jespino: It is the standard approach to not capitalize error strings on golang.

i18n/en.json
```diff
@@ -7370,6 +7410,10 @@
     "id": "store.sql_post.search.disabled",
     "translation": "Searching has been disabled on this server. Please contact your System Administrator."
   },
+  {
+    "id": "store.sql_system.get_by_name.app_error",
+    "translation": "Unable to get system data."
```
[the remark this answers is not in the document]
jespino: Yes it could. I'll add the changes.

services/upgrader/errors.go
jespino: I going to address all this in the frontend side, just adding the translations in the client instead of using the server messages. This way we no longer need to care about server side generated errors.

i18n/en.json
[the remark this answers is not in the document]
jespino: Actually, not I have to remove it from the file because is not longer part of this PR.

api4/system.go
jespino: I going to address all this errors from the frontend side instead of using the server side errors, so this PR can be merge without the need of this changes.

services/upgrader/errors.go
jespino: Finally I decide to fix it in the server side, because fix it in the frontend side have other complications.

jespino: Now should be fixed.

api4/system.go
jespino: Now should be fixed.

PR #14540, opened 2020-05-11
MM24459: Implement ldap picture sync

sbishel:
#### Summary
When updating the ldap picture, that go routine starts before the current session had connected its websocket. Therefore, we need to check and loop while waiting for the websocket to connect. The loop will try a fixed number of times before continuing and placing a message on the websocket. It is necessary to handle this update via the current user's websocket because the client caches the users image. Therefore even a refresh won't update the photo until the cache expires (24 hours).
Also split out the adjustments made when saving an image, so that they can be applied to an image being used to compare to the existing image. 

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-24459

#### Related PRs
https://github.com/mattermost/enterprise/pull/650


app/user.go
```diff
@@ -854,12 +854,11 @@ func (a *App) SetProfileImageFromMultiPartFile(userId string, file multipart.Fil
 	return a.SetProfileImageFromFile(userId, file)
 }
 
-func (a *App) SetProfileImageFromFile(userId string, file io.Reader) *model.AppError {
-
+func (a *App) AdjustImage(file io.Reader) ([]byte, *model.AppError) {
```
mkraft, line 857:
```suggestion
func (a *App) AdjustImage(file io.Reader) (*bytes.Buffer, *model.AppError) {
```

app/user.go
```diff
@@ -872,12 +871,20 @@ func (a *App) SetProfileImageFromFile(userId string, file io.Reader) *model.AppE
 	buf := new(bytes.Buffer)
 	err = png.Encode(buf, img)
 	if err != nil {
-		return model.NewAppError("SetProfileImage", "api.user.upload_profile_user.encode.app_error", nil, err.Error(), http.StatusInternalServerError)
+		return nil, model.NewAppError("SetProfileImage", "api.user.upload_profile_user.encode.app_error", nil, err.Error(), http.StatusInternalServerError)
 	}
+	return buf.Bytes(), nil
```
mkraft, line 876:
```suggestion
	return buf, nil
```

app/user.go
```diff
@@ -872,12 +871,20 @@ func (a *App) SetProfileImageFromFile(userId string, file io.Reader) *model.AppE
...
+func (a *App) SetProfileImageFromFile(userId string, file io.Reader) *model.AppError {
+
+	buf, err := a.AdjustImage(file)
+	if err != nil {
+		return err
+	}
 	path := "users/" + userId + "/profile.png"
 
-	if _, err := a.WriteFile(buf, path); err != nil {
+	if _, err := a.WriteFile(bytes.NewReader(buf), path); err != nil {
```
mkraft, line 887:
```suggestion
	if _, err := a.WriteFile(buf, path); err != nil {
```

I think with the above suggested change to `(*App).AdjustImage` you can use the `Buffer` directly without having to create a new reader.

app/web_hub_test.go
```diff
@@ -27,9 +30,11 @@ func dummyWebsocketHandler(t *testing.T) http.HandlerFunc {
 		for err == nil {
 			_, _, err = conn.ReadMessage()
 		}
+		mlog.Debug(err.Error())
```
fmunshi, line 33: do we want to keep this?

app/web_hub_test.go
```diff
@@ -27,9 +30,11 @@ func dummyWebsocketHandler(t *testing.T) http.HandlerFunc {
...
 		if _, ok := err.(*websocket.CloseError); !ok {
 			require.NoError(t, err)
 		}
+		mlog.Debug("Made IT")
```
fmunshi, line 37: same with this log?

app/web_hub_test.go
sbishel: Nope, thanks.

PR #14542, opened 2020-05-12
MM-24644 fix infinite loop splitting post attachments

wiggin77:
#### Summary

This PR fixes an infinite loop when splitting post attachments into multiple posts.  This bug was due to performing an assignment copy of a model.Post instead of using model.Post.clone().  The Post.Props map was being inadvertently carried over to the copy.

Includes new unit tests for this case and a few others.

This is meant to be the minimum fix as it is likely to be cherry picked for a dot release.  I plan to revisit this implementation as it generates more posts than necessary when the message text and attachments need to be split.

#### Ticket Link

https://mattermost.atlassian.net/browse/MM-24644

### DIFF
--- a/app/webhook_test.go
+++ b/app/webhook_test.go
@@ -456,6 +456,86 @@ func TestSplitWebhookPost(t *testing.T) {
 	}
 }
 
+func makePost(message int, attachments []int) *model.Post {
+	var props model.StringInterface
+	if len(attachments) > 0 {
+		sa := make([]*model.SlackAttachment, 0, len(attachments))
+		for _, a := range attachments {
+			attach := &model.SlackAttachment{
+				Text: strings.Repeat("那", a),
+			}
+			sa = append(sa, attach)
+		}
+		props = map[string]interface{}{"attachments": sa}
+	}
+	post := &model.Post{
+		Message: strings.Repeat("那", message),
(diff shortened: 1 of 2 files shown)


app/webhook_test.go
```diff
@@ -456,6 +456,86 @@ func TestSplitWebhookPost(t *testing.T) {
...
+
+func TestSplitWebhookPostAttachments(t *testing.T) {
+	maxPostSize := 10000
+	testCases := []struct {
+		name     string
+		post     *model.Post
+		expected []*model.Post
+	}{
+		{
+			// makePost(messageLength, []int{attachmentLength, ...})
```
agnivade, line 486: Let's remove this ?

wiggin77: I left it there intentionally as documentation for the next person who is wondering what the numbers represent.  I saves them looking for the makePost func.

wiggin77: Thanks for the review.  Any thoughts on whether the algorithm should be corrected now, so that the attachments are included starting with the first split?

agnivade: Ah I see. Yes that makes sense. Re: algorithm, I think it may be good to discuss this with the team. I am not sure on the historical context on the splitting decision here.

PR #14547, opened 2020-05-12
[MM-24604] Migrate get channel and get channel by name endpoints to local mode

mgdelacroix:
#### Ticket Link
https://mattermost.atlassian.net/browse/MM-24603
https://mattermost.atlassian.net/browse/MM-24604


api4/channel_test.go
```diff
@@ -1680,12 +1680,8 @@ func TestGetChannelMembers(t *testing.T) {
 		CheckBadRequestStatus(t, resp)
 
 		_, resp = client.GetChannelMembers("", 0, 60, "")
-		// NOTE: for some reason, while using the LocalClient, the route /channels//members is not handled
```
ashishbhate, line 1679: Same for `/channels//posts`, which is why I needed this PR for `getPostsForChannel`

There are some "hidden" dependencies in our routes

mgdelacroix: Yep, it definitely looks like it. If this happens again and knowing this now, what we can do is directly migrate all the subrouters to `LocalRouter` and then build the API on top of them, instead of migrating the ones we need for each PR. Plenty are migrated already, so hope this is the last we see of the issue

PR #14557, opened 2020-05-13
[MM-20684] Slash Command Autocomplete

iomodo:
#### Summary
This PR implements slash command autocomplete suggestions for plugins

#### Ticket Link 
Fixes https://mattermost.atlassian.net/browse/MM-18098
Fixes https://mattermost.atlassian.net/browse/MM-21485
Fixes https://mattermost.atlassian.net/browse/MM-21486
Fixes https://mattermost.atlassian.net/browse/MM-21487
Fixes https://mattermost.atlassian.net/browse/MM-21488
Fixes https://mattermost.atlassian.net/browse/MM-21489
Fixes https://mattermost.atlassian.net/browse/MM-21490
Fixes https://mattermost.atlassian.net/browse/MM-21491
Fixes https://mattermost.atlassian.net/browse/MM-21493
Fixes https://mattermost.atlassian.net/browse/MM-21494
Fixes https://mattermost.atlassian.net/browse/MM-24766
Fixes https://mattermost.atlassian.net/browse/MM-24767

#### Related PRs
webapp - https://github.com/mattermost/mattermost-webapp/pull/5499
redux - https://github.com/mattermost/mattermost-redux/pull/1121
demo-plugin - https://github.com/mattermost/mattermost-plugin-demo/pull/99
api-reference - https://github.com/mattermost/mattermost-api-reference/pull/522


model/command_autocomplete.go
```diff
@@ -0,0 +1,451 @@
...
+		HelpText: helpText,
+		Type:     AutocompleteArgTypeText,
+		Required: required,
+		Data:     &AutocompleteTextArg{Hint: hint, Pattern: pattern},
+	}
+	ad.Arguments = append(ad.Arguments, &argument)
+}
+
+// AddStaticListArgument adds positional AutocompleteArgTypeStaticList argument to the command.
+func (ad *AutocompleteData) AddStaticListArgument(helpText string, items []AutocompleteListItem, required bool) {
```
hanzei, line 138: Subjective nit pick: I seams to be that slice arguments usually are the last ones past to a function. Hence, I would go with:
```suggestion
func (ad *AutocompleteData) AddStaticListArgument(helpText string, required bool, items []AutocompleteListItem) {
```

model/command_autocomplete.go
```diff
@@ -0,0 +1,451 @@
...
+	ad.AddNamedStaticListArgument("", helpText, items, required)
+}
+
+// AddNamedStaticListArgument adds named AutocompleteArgTypeStaticList argument to the command.
+func (ad *AutocompleteData) AddNamedStaticListArgument(name, helpText string, items []AutocompleteListItem, required bool) {
```
hanzei, line 143: Same here

plugin/client_rpc.go
```diff
@@ -99,6 +99,9 @@ func init() {
 	gob.Register(&model.AppError{})
 	gob.Register(&ErrorString{})
 	gob.Register(&opengraph.OpenGraph{})
+	gob.Register(&model.AutocompleteDynamicListArg{})
+	gob.Register(&model.AutocompleteStaticListArg{})
+	gob.Register(&model.AutocompleteTextArg{})
```
hanzei, line 104: Should `AutocompleteData` also be added to the list?

model/command.go
```diff
@@ -18,23 +18,25 @@ const (
...
+	Username             string            `json:"username"`
+	IconURL              string            `json:"icon_url"`
+	AutoComplete         bool              `json:"auto_complete"`
+	AutoCompleteDesc     string            `json:"auto_complete_desc"`
+	AutoCompleteHint     string            `json:"auto_complete_hint"`
+	DisplayName          string            `json:"display_name"`
+	Description          string            `json:"description"`
+	URL                  string            `json:"url"`
+	AutocompleteData     *AutocompleteData `db:"-" json:"autocomplete_data,omitempty"`
+	AutocompleteIconData string            `db:"-" json:"autocomplete_icon_data,omitempty"`
```
hanzei, line 39: What is the data type here and can a comment make that clear?

app/command_channel_rename.go
```diff
@@ -25,12 +25,15 @@ func (me *RenameProvider) GetTrigger() string {
 }
 
 func (me *RenameProvider) GetCommand(a *App, T goi18n.TranslateFunc) *model.Command {
+	renameAutocompleteData := model.NewAutocompleteData(CMD_RENAME, T("api.command_channel_rename.hint"), T("api.command_channel_rename.desc"))
+	renameAutocompleteData.AddTextArgument(T("api.command_channel_rename.hint"), "[text]", "")
```
hanzei, line 29: There are also other commands like `code`that need to be updated. Is there a ticket to do that?

app/command_autocomplete.go
```diff
@@ -0,0 +1,253 @@
...
+	// input with a single word
+	index := strings.Index(in, " ")
+	if index == -1 { // typing of the single word argument is not finished
+		return true, parsed + toBeParsed, "", model.AutocompleteSuggestion{Complete: parsed + toBeParsed, Suggestion: "", Hint: a.Hint, Description: arg.HelpText}
+	}
+	// single word argument already typed
+	return false, parsed + in[:index+1], in[index+1:], model.AutocompleteSuggestion{}
+}
+
+func parseStaticListArgument(arg *model.AutocompleteArg, parsed, toBeParsed string) (found bool, alreadyParsed string, yetToBeParsed string, suggestions []model.AutocompleteSuggestion) {
```
hanzei, line 215: Are the names return values needed? Maybe it make the code easier to read if they get dropped, given that the method signature is already quite long. The same comment applies to all other methods in `app/command_autocomplete.go`.

app/command_autocomplete.go
```diff
@@ -0,0 +1,253 @@
...
+	a := arg.Data.(*model.AutocompleteStaticListArg)
+	return parseListItems(a.PossibleArguments, parsed, toBeParsed)
+}
+
+func (a *App) parseDynamicListArgument(arg *model.AutocompleteArg, parsed, toBeParsed string) (found bool, alreadyParsed string, yetToBeParsed string, suggestions []model.AutocompleteSuggestion) {
+	dynamicArg := arg.Data.(*model.AutocompleteDynamicListArg)
+	params := url.Values{}
+	params.Add("user_input", parsed+toBeParsed)
+	params.Add("parsed", parsed)
+	resp, err := a.doPluginRequest("GET", dynamicArg.FetchURL, params, nil)
```
hanzei, line 225: Maybe it's just the function name that is confusing me, but normal integrations will be able to register for a dynamic list, right?

app/command_autocomplete.go
```diff
@@ -0,0 +1,253 @@
...
+	// single word argument already typed
+	return false, parsed + in[:index+1], in[index+1:], model.AutocompleteSuggestion{}
+}
+
+func parseStaticListArgument(arg *model.AutocompleteArg, parsed, toBeParsed string) (found bool, alreadyParsed string, yetToBeParsed string, suggestions []model.AutocompleteSuggestion) {
+	a := arg.Data.(*model.AutocompleteStaticListArg)
+	return parseListItems(a.PossibleArguments, parsed, toBeParsed)
+}
+
+func (a *App) parseDynamicListArgument(arg *model.AutocompleteArg, parsed, toBeParsed string) (found bool, alreadyParsed string, yetToBeParsed string, suggestions []model.AutocompleteSuggestion) {
```
hanzei, line 220: 0/5 I doesn't seam like there is much parsing
```suggestion
func (a *App) getDynamicListArgument(arg *model.AutocompleteArg, parsed, toBeParsed string) (found bool, alreadyParsed string, yetToBeParsed string, suggestions []model.AutocompleteSuggestion) {
```

model/command_autocomplete.go
```diff
@@ -0,0 +1,451 @@
...
+
+	"github.com/pkg/errors"
+)
+
+// Argument types
+const (
+	AutocompleteArgTypeText        = "TextInput"
+	AutocompleteArgTypeStaticList  = "StaticList"
+	AutocompleteArgTypeDynamicList = "DynamicList"
+)
```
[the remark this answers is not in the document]
iomodo: Yes it is idiomatic, but I don't see any advantages in introducing type which is never used.

model/command_autocomplete.go
```diff
@@ -0,0 +1,451 @@
...
+	// Complete describes completed suggestion
+	Complete string
+	// Suggestion describes what user might want to input next
+	Suggestion string
+	// Hint describes a hint about the suggested input
+	Hint string
+	// Description of the command or a suggestion
+	Description string
+	// icon
+	IconData string
```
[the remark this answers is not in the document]
iomodo: yes base64 encoded svgs

model/command_autocomplete.go
```diff
@@ -0,0 +1,451 @@
...
+		Hint:        hint,
+		HelpText:    helpText,
+		RoleID:      SYSTEM_USER_ROLE_ID,
+		Arguments:   []*AutocompleteArg{},
+		SubCommands: []*AutocompleteData{},
+	}
+}
+
+// AddCommand adds a subcommand to the autocomplete data.
+func (ad *AutocompleteData) AddCommand(command *AutocompleteData) {
```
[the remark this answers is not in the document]
iomodo: `AddCommand` is idiomatic here, as is `addTree`(not addSubTree). For example you can see `cobra` [implementation](https://github.com/spf13/cobra/blob/94a87a7b83bda98e918cf7a4793f41372a5e4139/command.go#L1102)

model/command_autocomplete.go
```diff
@@ -0,0 +1,451 @@
...
+	IconData string
+}
+
+// NewAutocompleteData returns new Autocomplete data.
+func NewAutocompleteData(trigger, hint, helpText string) *AutocompleteData {
+	return &AutocompleteData{
+		Trigger:     trigger,
+		Hint:        hint,
+		HelpText:    helpText,
+		RoleID:      SYSTEM_USER_ROLE_ID,
```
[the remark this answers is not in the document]
iomodo: There is not through the API, I think we should add that after we finalize roles and permissions.

model/command_autocomplete.go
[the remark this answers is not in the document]
iomodo: I tried that, but API was not good. Added `nil` checks are in the `IsValid` method instead.

plugin/client_rpc.go
iomodo: No, this is needed only for objects which are provided using `interface{}` type

app/command_channel_rename.go
iomodo: Yes https://mattermost.atlassian.net/browse/MM-21499. We should make them HW.

app/command_autocomplete.go
iomodo: They are needed to understand what those returned strings will be, for readability.

app/command_autocomplete.go
iomodo: They will be in the future, it is not implemented yet.

model/command_autocomplete.go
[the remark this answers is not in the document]
hanzei: Could you please make this more clear in the comment?

api4/command.go
```diff
@@ -369,6 +370,39 @@ func listAutocompleteCommands(c *Context, w http.ResponseWriter, r *http.Request
...
+
+	commands, err := c.App.ListAutocompleteCommands(c.Params.TeamId, c.App.T)
+	if err != nil {
+		c.Err = err
+		return
+	}
+
+	suggestions := c.App.GetSuggestions(commands, userInput, roleId)
+
+	w.Write(model.AutocompleteSuggestionsToJSON(suggestions))
```
levb, line 403: set the content-type?

app/integration_action.go
```diff
@@ -325,27 +325,34 @@ func (w *LocalResponseWriter) WriteHeader(statusCode int) {
 	w.status = statusCode
 }
 
-func (a *App) DoLocalRequest(rawURL string, body []byte) (*http.Response, *model.AppError) {
+func (a *App) doPluginRequest(method, rawURL string, values url.Values, body []byte) (*http.Response, *model.AppError) {
```
levb, line 328: I am a little confused - is this a relevant change, or a rebase/merge issue? This file doesn't seem relevant to autocomplete.

model/command_autocomplete.go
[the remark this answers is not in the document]
levb: perhaps just `s/IconData/IconSVG[Base64]/`?

api4/command.go
iomodo: `json` is the default content-type as implemented in the handler. So no need to add it here.

app/integration_action.go
iomodo: It is relevant for downloading dynamic data from plugins. I just reorganized code so we could have different http methods `GET` , `POST`..

model/command_autocomplete.go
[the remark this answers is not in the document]
iomodo: We already use `IconData` in other parts of the code so I'll stick to the name and add a comment.

model/command_autocomplete.go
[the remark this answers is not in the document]
hanzei: I'm fine with the current implementation.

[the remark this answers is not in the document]
hanzei: Could you elaborate what issue the API without a pointer is?

model/command_autocomplete.go
```diff
@@ -0,0 +1,451 @@
...
+	// Hint of a command
+	Hint string
+	// Text displayed to the user to help with the autocomplete description
+	HelpText string
+	// Role of the user who should be able to see the autocomplete info of this command
+	RoleID string
+	// Arguments of the command. Arguments can be named or positional.
+	// If they are positional order in the list matters, if they are named order does not matter.
+	// All arguments should be either named or positional, no mixing allowed.
+	Arguments []*AutocompleteArg
```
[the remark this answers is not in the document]
hanzei: I'm still think this a valid point, but I'm also fine with addressing it in a separate ticket.

app/command_autocomplete.go
hanzei: Well, I think comes down to personal preference. I'm fine with keeping the named returns.

DHaussermann:
Regression tested this for backwards compatibility 
- Ran through release testing of all system slash commands 
- Tested main input and RHS 
-  Regression tested other auto-completes such such as channel and @ mentions as a precaution
- Regression tested a couple plugins for backwards compatibility. No issues found passing parameters (string and bool)
-  Regression tested a custom slash command for backwards compatibility
- Confirmed with Shota web traffic on auto-complete looks as expected
- Confirmed with Shota that parsing the commands should not cause any noticeable performance deterioration 
LGTM! Please merge.

mickmister:
@iomodo Looks like tests are failing here

aaronrothschild:
@iomodo Let's discuss Documentation during our next 1:1 so it's ready for release.

PR #14558, opened 2020-05-13
MM-24867: local mode for inviteUsersToTeam

ashishbhate:
#### Summary

- local mode support for inviteUsersToTeam
#### Ticket Link

- https://mattermost.atlassian.net/browse/MM-24867


ashishbhate:
The "normal" handler for inviteUsersToTeam passes the userID from the session to the app layer that has most of the invite logic. The app layer checks allowed domains. 

We can't use the app layer methods directly since they require the userID from the session for setting the inviting user. The allowed domain helper functions aren't exposed in the Public API.

I think it would be better to expose the allowed domains helpers, rather than reimplement most of it for local mode. 

For now, the local mode handler doesn't check allowed domains. The mattermost cli behaves the same way. 
I've also marked graceful mode as not implemented, because the correct implementation requires the ability to check allowed domains.

ashishbhate:
I've worked around the issue I mentioned in the previous comment.

PR #14565, opened 2020-05-13
MM-24386/MM-24465 Enable link previews and full push notifications by default

hmhealey:
We're changing the default values for both of these settings to improve user experience in Mattermost for new installs when the user hasn't changed any of these settings yet. These changes will only affect new installs and not existing users since we don't want to surprise server owners by turning on settings that they may have disabled.

@ashishbhate For a brief explanation of how this works, `SetDefaults` is called on all settings structs after the server loads the config from the disk. Any fields that are `nil` are missing from the config which generally happens when either the config is empty on a new install or the setting was just added in a new server version. It can also happen if the user deletes the corresponding lines from their config.json, but we don't really care about that case.

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-24386
https://mattermost.atlassian.net/browse/MM-24465


esethna:
@prapti we'll need to test defaults again here (new servers vs existing servers)

prapti:
@esethna Cloud test server creation is failing for the most recent commit for this PR (hash: aea9a8c). 3/5 that the failing checks are causing it. Is there any way someone could update so the checks here are passing? Then I can give it another try.

hmhealey:
Good catch, Ashish. That test should be fixed now.

prapti:
Testing in progress.

prapti:
Test in progress.

prapti:
@esethna @hmhealey I verified that the default are `"ServiceSettings.EanbleLinkPreviews":true` & `"EmailSettings.PushNotificationContents": "full"` for new install, and `false` and `generic` for server upgrades.
Was there any other values that I needed to check?

app/notification_push_test.go
```diff
@@ -1296,6 +1296,7 @@ func TestAllPushNotifications(t *testing.T) {
 	defer pushServer.Close()
 
 	th.App.UpdateConfig(func(cfg *model.Config) {
+		*cfg.EmailSettings.PushNotificationContents = model.GENERIC_NOTIFICATION
```
esethna, line 1299: @hmhealey what are we changing here regarding email notifications? This ticket is for push notifications

esethna:
@prapti we should be testing for mobile push notifications, not email notifications

app/notification_push_test.go
hmhealey: Yeah, that's for mobile push notifications

hmhealey:
Manually cherry-picked onto the release branch

model/config.go
```diff
@@ -1456,7 +1456,7 @@ func (s *EmailSettings) SetDefaults(isUpdate bool) {
 	}
 
 	if s.PushNotificationContents == nil {
-		s.PushNotificationContents = NewString(GENERIC_NOTIFICATION)
+		s.PushNotificationContents = NewString(FULL_NOTIFICATION)
```
ThiefMaster, line 1459: This change does not appear in the changelog...

hmhealey: That seems like it might've been an oversight, but it would only affect new install of Mattermost. These defaults only come into affect if that setting is completely missing from the config.

ThiefMaster: You are right. My deployment is a big unusual (default config, gets populated with values I set, so anything I didn't set explicitly keeps its default). I should probably change that at some point :)

esethna: cc// @amyblais re this being missed in the changelog ^

hmhealey: Ah, I see. Assuming you have some way to persist the config.json, you could consider setting any specific values you want using environment variables if you want to configure them from outside Mattermost. I think there's also an option to store the configuration in the database, but I haven't tried that out myself.

PR #14570, opened 2020-05-14
MM-25158 Add user to initial sidebar categories when joining team

hmhealey:
This pulls out the code to create the initial Favorites/Channels/Direct Messages categories from the migration logic and puts it into its own method so that it can be called when joining a team.

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-25158


app/team.go
```diff
@@ -632,6 +632,15 @@ func (a *App) JoinUserToTeam(team *model.Team, user *model.User, userRequestorId
 		return err
 	}
 
+	if err := a.createInitialSidebarCategories(user, team); err != nil {
+		mlog.Error(
+			"Encountered an issue creating default sidebar categories.",
+			mlog.String("user_id", user.Id),
+			mlog.String("team_id", team.Id),
+			mlog.Err(err),
+		)
```
agnivade, line 641: Should we return from here or not ?

reflog: Good question. @hmhealey - pretty sure we can't allow to join the team if something fails here...

hmhealey: We unfortunately don't have a good pattern of handling errors in side effects like this. Right below this, it tries to join the Town Square and Off Topic channels, but if that fails, it just logs an error and continues.

Since we don't return for other more-critical errors, I don't think we should return here either.

PR #14575, opened 2020-05-14
MM-25126 Add a separate default sorting method for categories

hmhealey:
Just in case we decide to change how categories are sorted by default, having separate values for the default sorting method and for manual sorting lets us keep track of whether or not a user has actually used manual ordering. The client will treat them the same for now.

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-25126


hmhealey:
Merging this despite the failing checks since they relate to missing API documentation, and that's not necessary since this is on a feature branch.

PR #14577, opened 2020-05-14
Add ?src=docker query parameter to identify download source

jasonblais:
#### Context:

Right now, multiple sources point to the server binaries on releases.mattermost.com, including the download page, upgrade instructions, Helm charts, version archive, and others.

We can identify the download source by adding a querystring to the url, i.e. https://releases.mattermost.com/5.22.1/mattermost-5.22.1-linux-amd64.tar.gz?src=docker

The intent is that we can then identify sources of downloads, identify those that typically lead to an unsuccessful server activation, and improve the experience for the developer/administrator.

This PR adds the "?src=docker" querysting tag to download source for the Docker app, so we know when someone downloaded it from this source

See https://github.com/mattermost/docs/pull/3596 for a similar change made to download sources on the version archive page.


cpanato:
@jasonblais you will need to add this change as well in the release pipeline, because there we set the package we will download. This change here is for people that is building the image and not changing anything in the arguments.


for the release pipeline change, see: https://github.com/mattermost/mattermost-shared-pipelines/blob/master/vars/platform_release.groovy#L158 and https://github.com/mattermost/mattermost-shared-pipelines/blob/master/vars/platform_release.groovy#L164

jasonblais:
Thank you! I submitted a PR for those changes https://github.com/mattermost/mattermost-shared-pipelines/pull/70

PR #14580, opened 2020-05-15
MM-25238: Fix system ping check to read from master

agnivade:
The code was writing to master and immediately after that reading
from a replica causing it to fail intermittently.

Since this is not a very high-traffic table, it should be safe to read
from master always.

#### Ticket link

https://mattermost.atlassian.net/browse/MM-25238

### DIFF
--- a/api4/system.go
+++ b/api4/system.go
@@ -86,23 +86,23 @@ func getSystemPing(c *Context, w http.ResponseWriter, r *http.Request) {
 			Value: currentTime,
 		})
 		if writeErr != nil {
-			mlog.Debug("Unable to write to database.", mlog.Err(writeErr))
+			mlog.Warn("Unable to write to database.", mlog.Err(writeErr))
 			s[dbStatusKey] = model.STATUS_UNHEALTHY
 			s[model.STATUS] = model.STATUS_UNHEALTHY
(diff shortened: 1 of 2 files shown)


streamer45:
Wondering if https://github.com/mattermost/mattermost-server/pull/14365 was related to this problem.

agnivade:
Heh, that's a slightly different problem. That will happen in the condition where it tries to match the time of the DB row. In that case, setting a unique cluster key solves the problem. 

In our case, the error was happening at this line:
```
mlog.Warn("Unable to read from database.", mlog.Err(readErr))
```
which means that it was unable to find the row at all. That can only happen if you are reading from a replica. It's expected because the queries are executed consecutively, so there's a high probability this will happen.

agnivade:
@saturninoabril - assigning you because you recently tested this. The steps are the same here. Just that getting a failure is slightly harder. We have seen this in our load tests.

I think if you try a lot of times, it is bound to fail at some point. This PR should make them pass 100% of times.

agnivade:
Yes. Though technically, this PR doesn't fix an HA issue. The issue was to do with DB read-replicas. But in any case, it would be good to have more coverage around HA.

PR #14583, opened 2020-05-15
MM-25256 don't logout mobile users via Session Idle Timeout

wiggin77:
#### Summary

This PR fixes a bug where mobile clients were being logged out via Session Idle Timeout.  This setting only applies to desktop app and browsers.  See system console text.

session_idle_tmeout

Original ticket with "doesn't apply to mobile" here:  https://mattermost.atlassian.net/browse/MM-7481

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-25256

### DIFF
--- a/app/session.go
+++ b/app/session.go
@@ -75,7 +75,7 @@ func (a *App) GetSession(token string) (*model.Session, *model.AppError) {
 	}
 
 	if *a.Config().ServiceSettings.SessionIdleTimeoutInMinutes > 0 &&
-		!session.IsOAuth &&
+		!session.IsOAuth && !session.IsMobileApp() &&
 		session.Props[model.SESSION_PROP_TYPE] != model.SESSION_TYPE_USER_ACCESS_TOKEN &&
 		!*a.Config().ServiceSettings.ExtendSessionLengthWithActivity {
 


enahum:
Fyi, @josephbaylon

amyblais:
Do we know if this a recent regression and which PR caused this bug?

wiggin77:
> Do we know if this a recent regression and which PR caused this bug?

Was broken via this commit (March 15, 2019): https://github.com/mattermost/mattermost-server/commit/5dae08761cc36d2ba625533089772afec91224fa

wiggin77:
@esethna , I've added `do not merge` tag until it is decided whether this timeout should apply to mobile.

enahum:
The timeout was explicitly excluded from the initial implementation as having the app in the background for that period of time will log you out, and it is just terrible UX

enahum:
Interestingly the decision to include the idle timeout to mobile was taken without anyone from the mobile/apps team involvement? 🤔

wiggin77:
> Interestingly the decision to include the idle timeout to mobile was taken without anyone from the mobile/apps team involvement?

Digging deeper I see where it was switched back (to not apply to mobile) but PR comment suggests it was only done for release branch.

enahum:
I still think it should not apply to mobile

josephbaylon:
I set the Session Idle Timeout to 5 minutes and waited some time after 5 minutes, I am no longer logged out due to session expiry. I'll leave it to @lindy65 for final verification once development decision has been made.

lindy65:
Hi @enahum @wiggin77 

Is is confirmed that `Session Idle Timeout` will NOT apply to mobile apps?  If so, then does https://mattermost.atlassian.net/browse/MM-16111 need to be closed?

Also, this PR is missing dev review from @crspeller 

Like Joseph, I also tested setting the Session Idle Timeout to 5 minutes and was not logged out after that time period was reached (and quite some time after).

enahum:
@lindy65 I would say keep the ticket opened so the discussion can continue, as of now the session idle timeout should not apply to mobile as it should have not been re-introduced without having been prepared for it.

wiggin77:
> Is is confirmed that `Session Idle Timeout` will NOT apply to mobile apps?

A decision was made to not have it apply to mobile here https://mattermost.atlassian.net/browse/MM-15971 ; but it was only applied to the release branch and didn't make it to master.  As such I think this PR is valid and should be merged as was the original intention.  Any objections?  cc @esethna @wiersgallak

esethna:
I agree ^, this PR is valid. See comment here: https://mattermost.atlassian.net/browse/MM-25256?focusedCommentId=88287&oldIssueView=true&page=com.atlassian.jira.plugin.system.issuetabpanels%3Acomment-tabpanel#comment-88287

esethna:
Docs already reflected this so no changes needed

PR #14586, opened 2020-05-16
[MM-25284] Include license type in Marketplace request

hanzei:
#### Summary
The Marketplace might hide certain EE only plugins in the future from being listed in the Marketplace modal. Please note that there is some ongoing discussion on how exactly the UX would be, but the gist is that the Marketplace needs to know the license type of a given Mattermost instance.

Because this kind of changes to the interaction between the marketplace and the server take a lot of time to apply, this change should go in as early as possible e.i. `v5.24.0`.

~~I did opt for using `License.SkuShortName` to identify the type of license. From the license files I have seen it seams like the most robust approach. But admittedly I have little knowable about the license type. @mkraft What do you think about this approach? Another more robust idea would be to send the full list of features.~~

Based on the feedback from @lieut-data the server also sends if the installation is EE-ready, e.i. if it's compiled with the enterprise code.

Please note that this is not a "security" feature. There will be additional checks inside the code of EE-only plugins that ensures it's only run on a EE10/EE20 server.

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-25284


app/plugin.go
```diff
@@ -497,7 +497,7 @@ func (a *App) getRemoteMarketplacePlugin(pluginId, version string) (*model.BaseM
 	return plugin, nil
 }
 
-func (a *App) getRemotePlugins(filter *model.MarketplacePluginFilter) (map[string]*model.MarketplacePlugin, *model.AppError) {
+func (a *App) getRemotePlugins(_ *model.MarketplacePluginFilter) (map[string]*model.MarketplacePlugin, *model.AppError) {
```
hanzei, line 500: I'm not sure why `*model.MarketplacePluginFilter` is passed at all, but I just omitted the name to make clear it's not used.

model/marketplace_plugin.go
```diff
@@ -90,6 +91,7 @@ func (filter *MarketplacePluginFilter) ApplyToURL(u *url.URL) {
 	}
 	q.Add("filter", filter.Filter)
 	q.Add("server_version", filter.ServerVersion)
+	q.Add("license_sku_short_name", filter.LicenseSkuShortName)
```
hanzei, line 94: 0/5 if the URL value should be omitted if `filter.LicenseSkuShortName` is empty.

app/plugin.go
lieut-data: Probably a relic from when we were going to pass through filtering settings upstream. I'm fine with removing the parameter altogether.

app/plugin.go
```diff
@@ -514,10 +514,15 @@ func (a *App) getRemotePlugins(filter *model.MarketplacePluginFilter) (map[strin
...
 
 	// Fetch all plugins from marketplace.
-	marketplacePlugins, err := marketplaceClient.GetPlugins(&model.MarketplacePluginFilter{
+	filter := &model.MarketplacePluginFilter{
 		PerPage:       -1,
 		ServerVersion: model.CurrentVersion,
-	})
+	}
+	if a.License() != nil {
+		filter.LicenseSkuShortName = a.License().SkuShortName
```
lieut-data, line 522: I'm not sure this is the right approach -- it's too bad we can't just serialize the entire license and send that up.

Separately, and while I know the UX is being sorted out, I'd be very surprised if we /hide/ enterprise plugins altogether. I wonder if the "bit" of interest is whether or not the server binary was compiled with enterprise code or not (regardless of whether it's licensed)?

hanzei: > it's too bad we can't just serialize the entire license and send that up.

0/5 Include it into the body? I'm aware that a `GET` with a `body` is strange, but that might just fulfill the needs here.

@lieut-data The desired UX is, that EE-only plugins should only be shown to server that can actually run them. Otherwise admins might be frustrated by downloading a plugin that doesn't run on there server. What does an EE ready server indicate that might help here?

lieut-data: > The desired UX is, that EE-only plugins should only be shown to server that can actually run them. Otherwise admins might be frustrated by downloading a plugin that doesn't run on there server. What does an EE ready server indicate that might help here?

The nuance here is that a server compiled with EE, but without a license, is /capable/ of running the plugins. We are actively pursuing UX that provides opportunities to prompt such an upgrade. By contrast, a server compiled without EE must be "physically replaced" even with a license in hand, and so can never run such a plugin. TLDR; I think showing enterprise-only plugins to an unlicensed, enterprise installation is desired at this time.

hanzei: @aaronrothschild  and I synced up offline:

The main goal of this change is to allow us to ship ee-only plugins via the Marketplace in the next three month, while not confusing admins who have don't have a license installed.

Due to the complexity in differentiating between E10 and E20, all EE-only plugin will work on both E10 and E20. There are no plans to add E20-only plugins for now.

While advertising enterprise-only plugins is an interesting idea, it's not feasible to implement for `v5.24.0`.

With all of that in mind the proposal is to have a `is_licensed` URL param. The Marketplace will omit any ee-only plugins if `is_licensed == false`.


@lieut-data Please let me know what you think about this.

aaronrothschild: @hanzei 
> Due to the complexity in differentiating between E10 and E20, all EE-only plugin will work on both E10 and E20. There are no plans to add E20-only plugins for now.

@hanzei Small change to this based on an offline discussion yesterday: These will be E20 only plugins.  They should not work on E10 licences.  

Could we for now do a check on the "LDAP" feature for example - which is only in E20, then run the plugin based on that feature being licenced? 

Over tie we should add a feature called "E20 only plugins" to licences, and that is only available on E20 licences.

levb: 1/5 - in the current (non-interactive) experience of installing plugins, exposing plugins that do not work seems like the wrong thing to do. I think it's fine if we currently only expose the enterprise plugins to _licensed_ customers.

In the future, we should make the installation of plugins more interactive, and at that point I agree it'd be benefitial to expose all available plugins; and if a user attempts to install one that requires a license - warn then. cc @aaronrothschild

hanzei: Based on the discussion above and offline I updated the code to:
- Send a `license_type` field that is `""`, `"E10` or `"E20"`. `"E20"` is send if the LDAP feature is enabled. That means if we change the way we map features to license types substantially, the information might be incorrect. But I don't see a better way to implement that check. @mkraft Thoughts?
- Send a `ee_ready` field that is `true` if the server was complied using the enterprise code. @lieut-data suggested adding this flag as it given us flexibility if we e.g. decide to show EE plugins to only `ee_ready` servers.


This approach given us maximum feasibility to decide later which plugins we want to show for which servers.

app/plugin.go
hanzei: Removed it :+1:

app/plugin.go
```diff
@@ -514,10 +514,24 @@ func (a *App) getRemotePlugins(filter *model.MarketplacePluginFilter) (map[strin
...
 	// Fetch all plugins from marketplace.
-	marketplacePlugins, err := marketplaceClient.GetPlugins(&model.MarketplacePluginFilter{
+	filter := &model.MarketplacePluginFilter{
 		PerPage:       -1,
 		ServerVersion: model.CurrentVersion,
-	})
+	}
+	license := a.License()
+	if license != nil {
+		if *license.Features.LDAP {
```
mkraft, line 523: I think you may be able to check for `FutureFeatures` rather than `LDAP`?

mkraft:
> but the gist is that the Marketplace needs to know the license type of a given Mattermost instance.

I don't quite understand this part. The reason I ask for clarification is checking the license type is—as far as know—a new paradigm. Typically we check one of:

1. Licensed or not.
2. `FutureFeatures`, meaning highest license level possible until we decided whether to add a feature flag.
3. A specific feature flag.

... but the concept of the license type (ie E10, E20) is purposely kept out of code to empower the PMs to make licensing decisions with less code changes.

levb:
I don't understand the code and the licensing consideration well enough to have an opinion on the specifics here, but a couple of considerations:
1) 2/5 it is IMO _usually_ better to have the client send the "raw", undigested data that it has (including its own identification and metadata) and have the logic on how to interpret it on the server, since updating the servers is usually cheaper than the clients.
2) In the short term - we really just need 1 bit to differentiate "enterprise" from "free" - we just don't know any better WRT/plugin packaging. Designing for the future may (or may not) be premature.

hanzei:
Based on a discussion with @mkraft I added a `EnterprisePlugins` feature flag to the license. This will allows us to decouple that flag from the E20 license. For now it's only set for E20 licenses.

@mkraft @lieut-data Sorry for the back and forth. Please take a look at the PR. That was the last change from my side.

model/license.go
```diff
@@ -64,6 +64,7 @@ type Features struct {
 	GuestAccountsPermissions  *bool `json:"guest_accounts_permissions"`
 	IDLoadedPushNotifications *bool `json:"id_loaded"`
 	LockTeammateNameDisplay   *bool `json:"lock_teammate_name_display"`
+	EnterprisePlugins         *bool `json:"enterprise_plugins"`
```
lieut-data, line 67: Not going to hold up the review for this, but I wonder if we shouldn't just send up the `future_features` flag. Adding a field here makes sense for  changes /in/ the server we might license discretely in the future, but this bucket is just way too wide to be useful for licensing enterprise plugins. Since we don't have a packaging story, should we just key off the existing "default" bit?

model/marketplace_plugin.go
```diff
@@ -74,11 +74,13 @@ func (plugin *BaseMarketplacePlugin) DecodeSignature() (io.ReadSeeker, error) {
...
-	Page          int
-	PerPage       int
-	Filter        string
-	ServerVersion string
-	LocalOnly     bool
+	Page              int
+	PerPage           int
+	Filter            string
+	ServerVersion     string
+	EEReady           bool
```
lieut-data, line 81: Can we use `BuildEnterpriseReady` to match `model.BuildEnterpriseReady`?

hanzei: That makes sense, thanks for the suggestion :+1:

model/license.go
hanzei: Adding a new feature flag has almost now cost. Adding a flag right now given us a bit flexibility if there is the decision to package enterprise plugins into a new SKU. Also, if this flag is considered obsolete in the future, we can just drop it without compatibility issues.

PR #14587, opened 2020-05-16
Match Users Even When SAML (+other) Contains Mixed Case

justledbetter:
#### Summary
Fixes issues where SAML assertions with mixed-case username/email fields don't match on active users (since 'first.last@example.com' != 'First.Last@example.com' in the query string).

Usernames/E-mail addresses are stored in the database in lowercase always, so we should be able to safely assume that all lookups of those fields should also be lowercase.  Rather than add a bunch of strings.ToLower() calls before we get to this point, it's hopefully most efficient to do it at the point where we interface with SQL.

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-25282


teresa-novoa:
@mkraft @sbishel friendly reminder to review this PR, thanks!

jasonblais:
@justledbetter Deep apologies for such a long reply time. This PR slipped our queue and our `mattermod` reminders for stale PRs was broken. I'm queuing this PR for review now. Huge thank you for your contribution 🙏

PR #14588, opened 2020-05-18
MM-24170: Allow mysql to choose the right index

agnivade:
We use the same optimization used in MM-23369 to prevent
mysql from using the index in the sort query.

#### Ticket Link

https://mattermost.atlassian.net/browse/MM-24170


agnivade:
@amyblais - This is review complete, and just needs to go through QA. A customer is expecting this to be available in 5.24. Will it be possible to include it ?

amyblais:
@agnivade Yes but preferably if it gets merged by today.

agnivade:
I will leave it to @ogi-m then to prioritize as necessary.

PR #14589, opened 2020-05-18
Only upgrade necessary steps to resource_class xlarge.

metanerd:
#### Summary
Resource xlarge was added to every step. Now only the very intense build steps get xlarge resource.


metanerd:
Resolved otherwise.

PR #14591, opened 2020-05-18
MM-22706: pass along set_online flag in websocket response

agnivade:
To let the client know whether a user has created a post without
being online or not, we get the set_online query param and
pass it down to the websocket event being passed down to the client.

With this PR, the "data" field of the `posted` event will contain
a `set_online` boolean field set to true/false depending on the
query_param set_online value set in the createPost call.

#### Ticket Link

https://mattermost.atlassian.net/browse/MM-22706


app/auto_responder.go
```diff
@@ -52,7 +52,7 @@ func (a *App) SendAutoResponse(channel *model.Channel, receiver *model.User) (bo
 		UserId:    receiver.Id,
 	}
 
-	if _, err := a.CreatePost(autoResponderPost, channel, false); err != nil {
+	if _, err := a.CreatePost(autoResponderPost, channel, false, true); err != nil {
```
hmhealey, line 55: Shouldn't this be `false`? If the post is coming from the auto responder, that doesn't mean that the user is online.

I think there might already be clientside handling to prevent that, but it's probably best to be consistent here with what we tell the client.

agnivade: Sure, I just went with the current behavior to avoid breaking anything, and set it to false only for the creation of post. But will change this. Thanks.

agnivade:
Gentle ping @jespino

PR #14600, opened 2020-05-19
MM-25276  Remove categories when leaving a team

reflog:
#### Summary
When a user leaves a team, we should remove the corresponding entries from the SidebarCategories and SidebarChannels tables in the database.

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-25276

### DIFF
--- a/app/team.go
+++ b/app/team.go
@@ -948,6 +948,10 @@ func (a *App) RemoveTeamMemberFromTeam(teamMember *model.TeamMember, requestorId
 		return err
 	}
 
+	if err = a.Srv().Store.Channel().ClearSidebarOnTeamLeave(user.Id, teamMember.TeamId); err != nil {
+		return err
+	}
+
 	// delete the preferences that set the last channel used in the team and other team specific preferences
 	if err := a.Srv().Store.Preference().DeleteCategory(user.Id, teamMember.TeamId); err != nil {
 		return err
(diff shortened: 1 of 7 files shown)


app/team.go
```diff
@@ -1002,8 +1002,11 @@ func (a *App) LeaveTeam(team *model.Team, user *model.User, requestorId string)
 		}
 	}
 
-	err = a.RemoveTeamMemberFromTeam(teamMember, requestorId)
-	if err != nil {
+	if err = a.Srv().Store.Channel().UpdateSidebarChannelsCategoriesOnLeave(channel, team.Id); err != nil {
```
hmhealey, line 1005: Should this maybe be done in `RemoveTeamMemberFromTeam`? Is this called when the user is removed from a team instead of them just leaving a team?

i18n/en.json
```diff
@@ -3892,7 +3892,7 @@
   },
   {
     "id": "ent.compliance.csv.metadata.json.zipfile.appError",
-    "translation": "Unable to create the zip file"
+    "translation": ""
```
hmhealey, line 3895: Not sure where this translation string went

app/team.go
```diff
@@ -1002,8 +1006,7 @@ func (a *App) LeaveTeam(team *model.Team, user *model.User, requestorId string)
 		}
 	}
 
-	err = a.RemoveTeamMemberFromTeam(teamMember, requestorId)
-	if err != nil {
+	if err = a.RemoveTeamMemberFromTeam(teamMember, requestorId); err != nil {
```
agnivade, line 1009: `:=`

i18n/en.json
agnivade: This is because this PR uses an outdated enterprise codebase. I added it in b9a62905476fbbb4c9753bfb250e4a974023dbdd. Let's bring it back.

reflog: @agnivade - this is a PR to a feature branch. To solve this problem, branch needs to be synced with master. I'm not fixing this, will be fixed before merging the feature branch

app/team.go
reflog: @agnivade  why? there's no shadowing here

agnivade: Just to make it more idiomatic and consistent with the other if blocks. We are declaring a new scope with the if block. But reusing the `err` from the parent scope. All other if blocks like this do the same.

PR #14603, opened 2020-05-19
Add trial license api

jespino:
#### Summary
Add trial license api

#### Ticket Link
[MM-25341](https://mattermost.atlassian.net/browse/MM-25341)

#### Related PRs
Has webapp changes: mattermost/mattermost-webapp#5532


app/license.go
```diff
@@ -214,3 +217,22 @@ func (a *App) GetSanitizedClientLicense() map[string]string {
 
 	return sanitizedLicense
 }
+
+// RequestTrialLicense request a trial license from the mattermost offical license server
```
catalintomai, line 221: official instead.

i18n/en.json
```diff
@@ -5702,6 +5710,14 @@
     "id": "model.websocket_client.connect_fail.app_error",
     "translation": "Unable to connect to the WebSocket server."
   },
+  {
+    "id": "mysqlsearchengine.purge-indexes.app-error",
```
catalintomai, line 5714: needed?

app/license.go
```diff
@@ -12,6 +13,8 @@ import (
 	"github.com/mattermost/mattermost-server/v5/utils"
 )
 
+const requestTrialURL = "https://portal.test.cloud.mattermost.com/api/v1/trials"
```
catalintomai, line 16: not sure if there is a convention or not in MM for consts to be upper case (or maybe only if they are used in more than one place?)

api4/license.go
```diff
@@ -138,3 +141,59 @@ func removeLicense(c *Context, w http.ResponseWriter, r *http.Request) {
...
+
+	b, readErr := ioutil.ReadAll(r.Body)
+	if readErr != nil {
+		c.Err = model.NewAppError("removeLicense", "api.license.request-trial.bad-request", nil, "", http.StatusBadRequest)
+		return
+	}
+	json.Unmarshal(b, &usersNumber)
+	if usersNumber.Users == 0 {
+		c.Err = model.NewAppError("removeLicense", "api.license.request-trial.bad-request", nil, "", http.StatusBadRequest)
+		return
```
catalintomai, line 172: I've seen these blocks factored-out in model as smth. like usersNumber := model.GetUsersNumberFromJson(r.Body), I get that the type here is local/not reused, still personally like that approach better - 2/5

app/license.go
coreyhulen: Guess is for testing?  do we have a live server?

utils/license.go
```diff
@@ -22,13 +22,13 @@ import (
...
 
 var publicKey []byte = []byte(`-----BEGIN PUBLIC KEY-----
-MIIBIjANBgkqhkiG9w0BAQEFAAOCAQ8AMIIBCgKCAQEAyZmShlU8Z8HdG0IWSZ8r
-tSyzyxrXkJjsFUf0Ke7bm/TLtIggRdqOcUF3XEWqQk5RGD5vuq7Rlg1zZqMEBk8N
-EZeRhkxyaZW8pLjxwuBUOnXfJew31+gsTNdKZzRjrvPumKr3EtkleuoxNdoatu4E
-HrKmR/4Yi71EqAvkhk7ZjQFuF0osSWJMEEGGCSUYQnTEqUzcZSh1BhVpkIkeu8Kk
-1wCtptODixvEujgqVe+SrE3UlZjBmPjC/CL+3cYmufpSNgcEJm2mwsdaXp2OPpfn
-a0v85XL6i9ote2P+fLZ3wX9EoioHzgdgB7arOxY50QRJO7OyCqpKFKv6lRWTXuSt
-hwIDAQAB
+MIIBIjANBgkqhkiG9w0BAQEFAAOCAQ8AMIIBCgKCAQEAwVMaUtQMFtQsRoa2FQd+
```
coreyhulen, line 25: Why did this change?  it should have correct?  Doesn't this public key need to match the private key?

i18n/en.json
jespino: Actually is not related at all, is related to local only files.

app/license.go
jespino: The testing server is up and running and the plan is to change this urls after QA tests.

The constans case styling is going to be changed for 6.0 for sure. I'm ok moving it back to uppercase.

api4/license.go
jespino: Yes, probably makes sense, I'll do it :)

utils/license.go
jespino: Yes, for testing purpose I have change the license public key to match with the test environment license private key. The plan is to change it back after QA.

jespino:
@srkgupta yes, that is the expected behavior because I changed in this PR the public key used, that is because I'm using the test-environment of for generate the trial license. The plan is to revert that part of the change before the PR is merged. (Change the Public Key with the real one, and changing the URL for requesting the license with the real one). If you want I can do it right now and test this last thing again.

jasonblais:
@srkgupta Let us know if you have any questions or blockers to test the changes in this PR?

srkgupta:
Hi @jespino 
I tested the basic workflow to request the Trial license and it worked fine. However I had some clarifications regarding the scope and had some observations as well. 

1. How many number of times can an instance request a trial? Currently I had requested a trial once, and then removed the license and requested trial again and this worked indefinitely. 
2. I wasn't able to test a scenario where the trial license period expired. Tried to modify the CreateAt data in LICENSES table and restarted the server but had no effect. 
3. On what basis is the LICENSE information like Name, Company Name populated when we request the trial. For me it defaulted to "Kenneth Moreno" and localhost instance URL respectively. 
4. Currently when we request a trial license, we do not display in the License page that the current license is a Trial license. Shouldn't this information be displayed somewhere in the screen?
Screenshot 2020-06-02 at 7 02 50 PM

5. The expected behavior to increase the user limit by 20% worked fine when a trial license was requested. However there were no restrictions applied. I was able to create way more number of users than the allowed limit. 
Screenshot 2020-06-02 at 7 02 34 PM

jespino:
@srkgupta I answer your questions inline:

> Hi @jespino
> I tested the basic workflow to request the Trial license and it worked fine. However I had some clarifications regarding the scope and had some observations as well.
> 
>     1. How many number of times can an instance request a trial? Currently I had requested a trial once, and then removed the license and requested trial again and this worked indefinitely.
Now there is no limit. That was decide in that way, in the future we can limit that behavior at the license server if we want. For now we expect low abuse of the feature.

>     2. I wasn't able to test a scenario where the trial license period expired. Tried to modify the CreateAt data in LICENSES table and restarted the server but had no effect.
You shouldn't be able to modify the date of the license, maybe you can modify the date of your machine. Anyway, that is not related to this feature, because the license validation is done in other (not related) part of the code.

>     3. On what basis is the LICENSE information like Name, Company Name populated when we request the trial. For me it defaulted to "Kenneth Moreno" and localhost instance URL respectively.
The name and the instance url is taken from your system configuration, if you are using the sampledata, the sysadmin name is "Kenneth Moreno", and the host is localhost:8065.

>     4. Currently when we request a trial license, we do not display in the License page that the current license is a Trial license. Shouldn't this information be displayed somewhere in the screen?
I think so, but again is not part of this feature, the trial license exists before this feature is implemented, so if it should show that there, would be another ticket maybe one PM can answer that question better than me.

> 
> Screenshot 2020-06-02 at 7 02 50 PM
> 
>     1. The expected behavior to increase the user limit by 20% worked fine when a trial license was requested. However there were no restrictions applied. I was able to create way more number of users than the allowed limit.

Yes, the main restriction applied here is the timeframe, the number of users, for now is not restricted, if we want to restrict that number we can change the license server API to disallow trial versions requests greater than X. I'll bring it up during our next meeting today to confirm it, but I think is the expected behavior.
> 
> 
> Screenshot 2020-06-02 at 7 02 34 PM

jespino:
@srkgupta Great, thanks! I'll confirm the absense of limit in the number of users, and if everything is ok, I'll merge it :)

jasonblais:
@justinegeffen would this be something you'd be open to documenting? This may need
- an addition to our API docs (api.mattermost.com)
- update to our existing docs around trial licenses at docs.mattermost.com

justinegeffen:
Thanks for the tag, @jasonblais. I've added myself as an assignee and will revert with an update before the end of the week.

justinegeffen:
Following up on this @jasonblais, it looks like @jespino has merged the required API doc update [here](https://github.com/mattermost/mattermost-api-reference/pull/524). Let me know if there was something else you had in mind on the API side. :)

jasonblais:
@justinegeffen Looks good from my side :+1: 

And I assume https://github.com/mattermost/docs/pull/3742 is the update about trial licenses at docs.mattermost.com?

justinegeffen:
Yep, and https://github.com/mattermost/docs/pull/3739. :)

PR #14607, opened 2020-05-20
MM-25362: Enable brotli only when TLS is enabled

agnivade:
Without this, local browsers fail to render js/css if they don't have an updated version of webapp. It may be safer to only enable it conditionally if there are caching issues where an older webapp requests a newer server version.

#### Ticket Link

https://mattermost.atlassian.net/browse/MM-25362


agnivade:
Not needed.

PR #14609, opened 2020-05-20
MM-25281  Place new categories in the correct position

reflog:
#### Summary

When a new category is created, it should be placed as follows:
1. If the Favorites category is first, the new category should be placed after it
2. Otherwise, the new category should be placed first.
#### Ticket Link

https://mattermost.atlassian.net/browse/MM-25281

### DIFF
--- a/store/sqlstore/channel_store.go
+++ b/store/sqlstore/channel_store.go
@@ -3440,26 +3440,35 @@ func (s SqlChannelStore) CreateSidebarCategory(userId, teamId string, newCategor
 
 	defer finalizeTransaction(transaction)
 
-	sql, args, _ := s.getQueryBuilder().
-		Select("MAX(SortOrder)").
-		From("SidebarCategories").
-		GroupBy("SortOrder").
-		Where(sq.And{
-			sq.Eq{"TeamId": teamId},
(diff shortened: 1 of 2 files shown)


store/sqlstore/channel_store.go
```diff
@@ -3440,26 +3440,35 @@ func (s SqlChannelStore) CreateSidebarCategory(userId, teamId string, newCategor
...
-		}).ToSql()
-	maxOrder, err := transaction.SelectInt(sql, args...)
-	if err != nil {
-		return nil, model.NewAppError("SqlPostStore.CreateSidebarCategory", "store.sql_channel.sidebar_categories.app_error", nil, err.Error(), http.StatusInternalServerError)
+	categoriesWithOrder, appErr := s.getSidebarCategoriesT(transaction, userId, teamId)
+	if appErr != nil {
+		return nil, appErr
 	}
-	categoryId := model.NewId()
+	if len(categoriesWithOrder.Categories) < 1 {
```
hmhealey, line 3447: I was going to ask why we'd error out when you have no categories, but I guess this'll make it easier to not get into a weird cases by hitting this API before migration has occurred.

PR #14610, opened 2020-05-20
MM-25154: Fix data race in InvokeClusterLeaderChangedListener

agnivade:
The order of booting up a server should be
Job scheduler -> Cluster.

And shutdown should be the opposite. This is needed because the job scheduler
initializes certain data structures that are later called by ClusterLeaderChanged
event handlers. And the event handlers run in a separate goroutine.

Therefore, if the cluster initialization happens before the job scheduler,
then a race happens between accessing the jobs variable and setting it.

To prevent this race, we fix the order of startup, and add comments in both
places to prevent things from regressing again.

#### Ticket Link

https://mattermost.atlassian.net/browse/MM-25154


saturninoabril:
@agnivade Do you know what's the fix version? Please update the ticket accordingly.  Assigning to you.  Thanks

PR #14614, opened 2020-05-20
[MM-25385] Add group_count_with_allow_reference telemetry

fmunshi:
#### Summary
- Adds `group_count_with_allow_reference` to telemetry

#### Ticket Link
- https://mattermost.atlassian.net/browse/MM-25385


store/sqlstore/group_store.go
```diff
@@ -1443,15 +1443,23 @@ func (s *SqlGroupStore) GroupMemberCount() (int64, *model.AppError) {
...
-	return s.countTableWithSelect("COUNT(*)", tableName)
+	return s.countTableWithSelectAndWhere("COUNT(*)", tableName, nil)
 }
 
-func (s *SqlGroupStore) countTableWithSelect(selectStr, tableName string) (int64, *model.AppError) {
-	query := s.getQueryBuilder().Select(selectStr).From(tableName).Where(sq.Eq{"DeleteAt": 0})
+func (s *SqlGroupStore) countTableWithSelectAndWhere(selectStr, tableName string, whereStmt map[string]interface{}) (int64, *model.AppError) {
+	if whereStmt == nil {
+		whereStmt = sq.Eq{"DeleteAt": 0}
+	}
```
sbishel, line 1460: I'm 0/5 about changing this. It actually provides more versatile functionality and all the callers are right here. 
But shouldn't "DeleteAt" always be added? Based on previous code, it was always added. The caller shouldn't need to add it. Also, if caller passed in an empty (not nil) map. "DeleteAt" wouldn't get added.

fmunshi: Yea I was thinking about doing that too but then that removes the possibility of having this function ever return deleted entries (if we ever want that in the future) - I guess I could check if the key exists for `DeleteAt` and only set `DeleteAt = 0` if the key does not exist in the passed where statement

PR #14617, opened 2020-05-21
MM-25014:set group name when group is created

sbishel:
#### Summary
When a group link gets created it is being created with a "NewId()" as the name value. That value is then used in the WebApp as the token or slug. This PR creates the slug name when the group is created.

Although group mentions may never be turned on for that group, the Name field is indexed and must be unique. Therefore, we may append a duplicate breaker. When a group is patched, we must also check for uniqueness. The patch results in a thrown error for the user to modify. When it is being created, we must have a unique identifier. If a unique name base on Display Name cannot be found, default is the NewId().

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-25014

### DIFF
--- a/api4/ldap.go
+++ b/api4/ldap.go
@@ -203,7 +203,7 @@ func linkLdapGroup(c *Context, w http.ResponseWriter, r *http.Request) {
 		// Only try a finite number of times, then give up.
 		baseName := strings.ReplaceAll(strings.ToLower(displayName), " ", "-")
 		validName := ""
-		for i := 0; i < 100; i++ {
+		for i := 0; i < 10; i++ {
 			checkName := baseName
 			if i > 0 {
 				checkName = baseName + "-" + strconv.Itoa(i)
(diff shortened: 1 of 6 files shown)


api4/ldap.go
```diff
@@ -194,11 +197,35 @@ func linkLdapGroup(c *Context, w http.ResponseWriter, r *http.Request) {
...
-		// TODO: In a future phase of LDAP groups sync `Name` will be used for at-mentions and will be editable on
-		// the front-end so it will not have an initial value of `model.NewId()` but rather a slugified version of
-		// the LDAP group name with an appended duplicate-breaker.
+		// `Name` will be used for at-mentions and will be editable on the front-end so
+		// create name based on display name and make sure it is unique.
+		// Append a duplicate-breaker if necessary.
+		// Only try a finite number of times, then give up.
+		baseName := strings.ReplaceAll(strings.ToLower(displayName), " ", "-")
+		validName := ""
+		for i := 0; i < 10; i++ {
```
sbishel, line 206: Why 10? The number itself is arbitrary. However, the calls made by ValidateGroupName do not distinguish between a db error and an item not found error. Therefore, if we don't find a valid name in 10 attempts, we will use the "NewId()" string. If the 10 was valid because of duplicate names, the admin would want to change the name anyway from "Group_10" to something more understandable. More than likely it is due to a db connection error.

sbishel:
Changing implementation to set to null until AllowReference is set.  Closing in favor of this https://github.com/mattermost/mattermost-server/pull/14651

PR #14619, opened 2020-05-21
Mm 25040 group constrained team change

mkraft:
#### Summary

* Filters get groups when the parent team is group-constrained.
* Prevents associating a group to a channel if the parent team is group-constrained and the group is not already associated to the parent team.

#### Ticket Link

https://mattermost.atlassian.net/browse/MM-25040


model/client4.go
```diff
@@ -354,7 +354,7 @@ func (c *Client4) GetElasticsearchRoute() string {
 }
 
 func (c *Client4) GetBleveRoute() string {
-	return fmt.Sprintf("/bleve")
+	return "/bleve"
```
mkraft, line 357: Unrelated linting fix.

Willyfrog:
@mkraft just to make sure I'm getting this right, a group constrained team is one that is synchronizing with LDAP groups?

also, you might want to assign someone who knows that part of the code? not sure a full apps team review is the best for this PR.

api4/group.go
```diff
@@ -767,6 +773,28 @@ func getGroups(c *Context, w http.ResponseWriter, r *http.Request) {
 		return
 	}
 
+	// if the team is group-constrained, only display the intersect of the teams
+	// groups and the groups returned by the query.
+	if parentTeam != nil && parentTeam.IsGroupConstrained() {
```
cpoile, line 778: A thought: because we do the paging logic prior to the intersection, are we sure the result will be as expected? Eg, if you have 20 groups, Page = 0, PerPage = 10, but only one group is in the intersection, you'll get one group in the response, even though there may be more in the next page -- the user will think "ok, I guess only one group in total" not thinking to call for Page = 1.

app/group.go
```diff
@@ -86,30 +88,51 @@ func (a *App) UpsertGroupSyncable(groupSyncable *model.GroupSyncable) (*model.Gr
 		return nil, err
 	}
 
-	if gs == nil {
-		gs, err = a.Srv().Store.Group().CreateGroupSyncable(groupSyncable)
+	// reject the syncable creation if the group isn't already associated to the parent team
+	if groupSyncable.Type == model.GroupSyncableTypeChannel {
+		var channel *model.Channel
+		channel, err = a.Srv().Store.Channel().Get(groupSyncable.SyncableId, true)
```
cpoile, line 94: Sorry, I'm not sure if I can review this section of the code because I don't really understand what's happening. :)

app/group.go
```diff
@@ -372,3 +396,26 @@ func (a *App) UserIsInAdminRoleGroup(userID, syncableID string, syncableType mod
...
+	page := 0
+	var teamGroups []*model.Group
+
+	for batchCount > 0 {
+		batch, err := a.Srv().Store.Group().GetGroupsByTeam(teamID, model.GroupSearchOpts{PageOpts: &model.PageOpts{Page: page, PerPage: 100}})
+		if err != nil {
+			return nil, err
+		}
+
+		batchCount = len(batch)
```
cpoile, line 412: Maybe I don't understand it fully, but: Does this mean it will stop if there are no groups in a page, even though that may be because there was no intersection (even though there may be more in latter pages)? In other words: we may have a batchCount of 0, but we're not done yet?

app/group.go
```diff
@@ -86,30 +88,51 @@ func (a *App) UpsertGroupSyncable(groupSyncable *model.GroupSyncable) (*model.Gr
...
+				if teamGroup.Id == groupSyncable.GroupId {
+					permittedGroup = true
+					break
+				}
+			}
+			if !permittedGroup {
+				return nil, model.NewAppError("App.UpsertGroupSyncable", "group_not_associated_to_synced_team", nil, "", http.StatusBadRequest)
+			}
+		} else {
+			_, err = a.UpsertGroupSyncable(model.NewGroupTeam(groupSyncable.GroupId, team.Id, groupSyncable.AutoAdd))
```
Willyfrog, line 121: I don't understand why are we recurring here. since it is not a `GroupSyncableTypeChannel` it will ignore the first part of the `if`, and all the other has already been done, so it would only be iterating.

or maybe I'm missing some pieces, as I don't know this part of code.

app/group.go
```diff
@@ -372,3 +396,26 @@ func (a *App) UserIsInAdminRoleGroup(userID, syncableID string, syncableType mod
...
 	return true, nil
 }
+
+// GetAllGroupsByTeam iterates all pages of groups associated to the given team and returns them all.
+func (a *App) GetAllGroupsByTeam(teamID string) ([]*model.Group, *model.AppError) {
+	batchCount := 1
+	page := 0
+	var teamGroups []*model.Group
+
+	for batchCount > 0 {
```
Willyfrog, line 406: if this is an internal call (not a public facing api), why does it need to go through pages? wouldn't it make sense to ask for all the groups directly to the database?

also, glancing throught the code, couldn't that be acomplished by passing nil as the `pageOpts` field? (I might have misunderstood the code for it)

app/group.go
mkraft: I think this whole block needs a refactor given it's confusing to read.

Line 92 branches this entire diff, meaning that all of these changes related only to a "group channel" (aka `GroupSyncableTypeChannel`) upsert request.

The requirements can be summarized like this:

* Existing requirement: if a parent team is not group-constrained, then create the group channel *and* the group-team.
* New requirement (that was labelled a bug, but really it's a requirement change): if the parent team is group-constrained don't create anything unless the team is already associated to the group.

api4/group.go
mkraft: You're right. This change is breaking pagination. I hadn't noticed, so thank you. This either needs to incorporate the change into the data tier or request the next page(s) of results—by making further DB requests—to fill in any results removed so that the client can know whether there will be more results in the next page or not. I'm going to make one of those updates.

mkraft: @cpoile I'm going to switch to this: https://github.com/mattermost/mattermost-server/pull/14619/files#diff-2c6106afe8477623894c02707bffe06dR1219-R1238. I'll request a re-review once all of the additional tests—including covering pagination—are in place.

app/group.go
```diff
@@ -102,14 +102,14 @@ func (a *App) UpsertGroupSyncable(groupSyncable *model.GroupSyncable) (*model.Gr
 			return nil, err
 		}
 		if team.IsGroupConstrained() {
-			var teamGroups []*model.Group
-			teamGroups, err = a.GetAllGroupsByTeam(team.Id)
+			var teamGroups []*model.GroupWithSchemeAdmin
+			teamGroups, err = a.Srv().Store.Group().GetGroupsByTeam(channel.TeamId, model.GroupSearchOpts{})
```
cpoile, line 106: nit: we could use `:=` here, and remove the var declaration, but up to you (since its your code)

mkraft: I think this was one of the "shadow declaration" changes.

PR #14623, opened 2020-05-21
MM-25413: Fix empty localization string

agnivade:
#### Ticket link

https://mattermost.atlassian.net/browse/MM-25413


lindy65:
Hi @agnivade - can you help with test steps for this PR please?  Thanks!

agnivade:
Hi @lindy65 - Unfortunately I don't think you can test this. The code path is a safety check which won't happen as of now. So there is no way to trigger this AFAI can see.

PR #14624, opened 2020-05-21
Add database read retry to enhanced ping endpoint

gabrieljackson:
The enhanced ping endpoint can return an unhealthy database read
result if the database read replicas have a high-enough replication
delay.

This adds basic retry logic to the ping endpoint to re-attempt
database reads up to a given "unreasonable" replication delay.


gabrieljackson:
I wanted some initial input and agree with @crspeller's comments in another channel.

Basically, we are in agreement that although this check is useful, it should perhaps be separated out into a second check type. Going to remove read DB check instead for now. I will do a follow-up PR for that.

PR #14625, opened 2020-05-21
Remove database read check from enhanced ping endpoint

gabrieljackson:
The enhanced ping endpoint can return an unhealthy database read
result if the database read replicas have a high-enough replication
delay. Instead of adding the complexity and delay of retry logic,
we are opting to remove the read check for now and will re-add it
in the future in a separate health check flow.

https://mattermost.atlassian.net/browse/MM-25428

### DIFF
--- a/api4/system.go
+++ b/api4/system.go
@@ -79,7 +79,7 @@ func getSystemPing(c *Context, w http.ResponseWriter, r *http.Request) {
 		dbStatusKey := "database_status"
 		s[dbStatusKey] = model.STATUS_OK
 
-		// Database Write/Read Check
+		// Database Write Check
 		currentTime := fmt.Sprintf("%d", time.Now().Unix())
 		healthCheckKey := fmt.Sprintf("health_check_%s", c.App.GetClusterId())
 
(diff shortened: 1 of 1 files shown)


gabrieljackson:
I will skip QA on this one as the only way to ensure the fix is to have a badly-performing server and database, which is not really possible to replicate easily without a complicated setup.

gabrieljackson:
Manually cherry-picked this due to auto-picking failing.

agnivade:
This is funny. I am wondering whether the issue was reproduced on master before attempting this fix ? Because I had already fixed it via - https://github.com/mattermost/mattermost-server/commit/21af1f49f10d7890f3b51d5be99efcbb71c33edd. :)

Was that fix still not sufficient ?

gabrieljackson:
Haha, sorry, I didn't actually notice that change @agnivade, as I had been working through various iterations of the what to do with that ping option and getting feedback from other engineers.

Thanks for bringing it to my attention. I still think this change makes sense to have for now though. Here is some of the backstory of the endpoint and the reasoning for this PR:

`get_server_status` was originally intended to exercise various crucial external services/components and perform some basic, but important, validation that the server could properly use them. One of these checks was to make sure that the server could use the read replicas if those were configured. During my various PoC fixes for this issue that we both independently found, I tried everything from read retries to writing and reading from master like you implemented. Read retries was too much overhead and complexity and after talking about just reading from master, a few of us basically agreed that if you can write to master, reading from master is probably a near guarantee so might as well just move along.

The long-term plan is to now expose a new endpoint or option to perform a read check on the DB read replicas along with other checks as well that may not fit in the scope of a quick health check endpoint.

Anyway, let me know if you have any thoughts and thanks for the heads-up on your PR!

agnivade:
Ah no worries. I was just wondering whether there was a bug in my fix or not. I am fine with whatever you think is appropriate.

PR #14628, opened 2020-05-22
MM-25439: Skip SuddenClose flaky test

agnivade:
This fails in SuddenClose because we check for closing the writeChan
only after waiting the closure of Event and Response channels.

Therefore, it is still racy and can panic.
There is no use chasing this again because it will be completely overhauled in v6.

#### Ticket link

https://mattermost.atlassian.net/browse/MM-25439


agnivade:
@lindy65 - I don't believe there is anything to test here. This just skips a test.

PR #14635, opened 2020-05-22
MM-24517: Add newly created channel admin role to all clients' redux state.

mkraft:
#### Summary

Makes all clients aware of newly created channel role.

#### Ticket Link

https://mattermost.atlassian.net/browse/MM-24517


mkraft:
> Looks good but do you know where the update role event is sent for the other roles? `ChannelUser` and `ChannelGuest`?
> Wouldnt those need to be updated too? or am i missing something why did it work for users and guests but not for admins?

The `(*App).UpdateRole` method sends the message for the other two roles.

fmunshi:
> > Looks good but do you know where the update role event is sent for the other roles? `ChannelUser` and `ChannelGuest`?
> > Wouldnt those need to be updated too? or am i missing something why did it work for users and guests but not for admins?
> 
> The `(*App).UpdateRole` method sends the message for the other two roles.

Ah and because we're never actually updating the channel admin role this role never sends an update even, makes sense - do you think its worth having a comment explaining why its needed?

model/client4.go
```diff
@@ -388,7 +388,7 @@ func (c *Client4) GetElasticsearchRoute() string {
 }
 
 func (c *Client4) GetBleveRoute() string {
-	return fmt.Sprintf("/bleve")
+	return "/bleve"
```
mkraft, line 391: Unrelated lint error fix.

PR #14636, opened 2020-05-23
Check for nil error in ErrConflict

agnivade:
We check if the err is nil or not before calling the Error() method.


agnivade:
@lindy65 - This just covers an edge case in the code, which is not testable from outside.

PR #14640, opened 2020-05-24
Enable gossip encryption

agnivade:
### Ticket link

https://mattermost.atlassian.net/browse/MM-25050


agnivade:
I had thought about that, but it might need some work to implement it in a cluster upgrade scenario. Because now all servers will generate their own keys. So to prevent that, they would need a distributed lock on the DB to agree on what is the correct key. It will be something like - "check if key exists, else insert, return" wrapped in a transaction.

I went for the simpler approach, but totally open to do it this way too.

DSchalla:
Yeah that makes a lot of sense. I thought whether we could utilize the leader election mechanism, but that won't work since they would need to rely on Gossip.. ;-) I would personally be fine to push the auto generation back to a later point in time. We should have this before making the not experimental anymore just because adoption will otherwise be tough.

agnivade:
Ok, in that case, I think it might be better to do it in this PR itself.

store/sqlstore/system_store.go
```diff
@@ -86,3 +87,35 @@ func (s SqlSystemStore) PermanentDeleteByName(name string) (*model.System, *mode
 
 	return &system, nil
 }
+
+// InsertIfExists inserts a given system value if it does not already exist. If a value
+// already exists, it returns the old one, else returns the new one.
+func (s SqlSystemStore) InsertIfExists(system *model.System) (*model.System, *model.AppError) {
```
streamer45, line 93: Nice, you really like your idempotence :)

store/sqlstore/system_store.go
```diff
@@ -86,3 +87,35 @@ func (s SqlSystemStore) PermanentDeleteByName(name string) (*model.System, *mode
...
+		return nil, model.NewAppError("SqlSystemStore.InsertIfExists", "store.sql_system.get_by_name.app_error", nil, err.Error(), http.StatusInternalServerError)
+	}
+
+	if origSystem.Value != "" {
+		// Already a value exists, return that.
+		return &origSystem, nil
+	}
+
+	// Key does not exist, need to insert.
+	if err := tx.Insert(system); err != nil {
```
streamer45, line 113: Can multiple transactions of this type run at the same time? 
Is it possible to get here and have all but one of them fail? 
As far as I understand only one should succeed in writing but all of them should return something.

agnivade: That's a great point ! I read a bit about transaction isolation levels. And it seems that the default isolation levels in most DBs is "read-committed". What we need here is "serializable" isolation level. But unfortunately, our gorp fork does not have [BeginTx](https://golang.org/pkg/database/sql/#DB.BeginTx) which allows you to specify the isolation level.

So I'd need to patch our fork, and then re-vendor the library again.

Thanks for catching this btw. Would have been a nasty bug.

store/sqlstore/system_store.go
```diff
@@ -86,3 +88,37 @@ func (s SqlSystemStore) PermanentDeleteByName(name string) (*model.System, *mode
 
 	return &system, nil
 }
+
+// InsertIfExists inserts a given system value if it does not already exist. If a value
+// already exists, it returns the old one, else returns the new one.
+func (s SqlSystemStore) InsertIfExists(system *model.System) (*model.System, *model.AppError) {
```
streamer45, line 94: Should we add a test for this? Maybe spawning a few goroutines and make sure the results are consistent.

store/sqlstore/system_store.go
```diff
@@ -86,3 +88,37 @@ func (s SqlSystemStore) PermanentDeleteByName(name string) (*model.System, *mode
...
+		// Already a value exists, return that.
+		return &origSystem, nil
+	}
+
+	// Key does not exist, need to insert.
+	if err := tx.Insert(system); err != nil {
+		return nil, model.NewAppError("SqlSystemStore.InsertIfExists", "store.sql_system.save.app_error", nil, err.Error(), http.StatusInternalServerError)
+	}
+
+	if err := tx.Commit(); err != nil {
```
streamer45, line 120: Do we have/need a strategy in case the transaction were to fail? 
If this is something needed to have the server operating we might require some kind of re-trying mechanism.

store/sqlstore/system_store.go
agnivade: You get full marks for maintaining "high standards". I like it :)

store/sqlstore/system_store.go
agnivade: Discussed offline. Since this is a relatively simple get-update query, there's very low chances of the transaction failing. And in general, retrying failed transactions are something not done in our codebase as of now. Skipping this.

agnivade:
@saturninoabril - This is a combination of 3 PRs from webapp, server and EE codebase. I have assigned them to you. But it's essentially a single feature.

The test steps are detailed in https://mattermost.atlassian.net/browse/MM-25050. 

It's a big feature, so there are quite a few things to verify. Would be good to have your own test cluster to be able to play around with. Let me know if you have any questions. Thanks !

saturninoabril:
Noted. I've setup recently an [HA cluster for QA testing](http://qa-ha-test.dev.spinmint.com/). This PR is now deployed there using `mattermost-enterprise-edition:2cb450d`.

agnivade:
> Just curious, is there an option to reset/regenerate the secret key in case wanted to do so?

It's not exposed via any API, but if the user wishes, they can directly edit the DB entry to a new key. The design decision that we took was that key rotation would be a fairly rare event, so for now manual editing would suffice. We can add such an API later in the future should the need arise.

agnivade:
Release notes:

This feature adds an `EnableExperimentalGossipEncryption` option under `ClusterSettings`. If this is set to true, and `UseExperimentalGossip` is also `true`, all communication through the cluster using the gossip protocol would be encrypted.

The encryption uses AES-256 by default, and it is not kept configurable by design. However, if one wishes, they can set the value in `Systems` table manually for the `ClusterEncryptionKey` row. A key is a byte array converted to base64. It should be either 16, 24, or 32 bytes to select AES-128, AES-192, or AES-256.

To update the key, one can execute:
`UPDATE Systems SET Value='<value>' WHERE Name='ClusterEncryptionKey';` in MySQL and
`UPDATE systems SET value='<value>' WHERE name='ClusterEncryptionKey'` for Postgres.

Lastly, for any change this this config setting to take effect, the whole cluster must be shutdown first. Then the config change made, and then restart. In a cluster, all servers either will completely use encryption or not. There cannot be any partial usage.

amyblais:
@agnivade Was this added to ``app/diagnostics.go`` (if needed)?

agnivade:
I will add it.

PR #14643, opened 2020-05-24
MM-25476: Migrate AppError from command_store.go

joshuabezaleel:
#### Summary
Migrate AppError to idiomatic error on commandStore.

#### Ticket Link
Fixes: #14637 
Jira: https://mattermost.atlassian.net/browse/MM-25476


joshuabezaleel:
Hi @hanzei , the tests have all passed and I hope it should already be ready to review. Looking forward to it! :slightly_smiling_face:

joshuabezaleel:
I accidentally closed this, so sorry.

hanzei:
Awesome @joshuabezaleel :tada: I request a dev review to look into this

store/sqlstore/command_store.go
```diff
@@ -153,7 +160,7 @@ func (s SqlCommandStore) AnalyticsCommandCount(teamId string) (int64, *model.App
 
 	c, err := s.GetReplica().SelectInt(query, map[string]interface{}{"TeamId": teamId})
 	if err != nil {
-		return 0, model.NewAppError("SqlCommandStore.AnalyticsCommandCount", "store.sql_command.analytics_command_count.app_error", nil, err.Error(), http.StatusInternalServerError)
+		return 0, fmt.Errorf("unable to count the commands: team_id=%s", teamId)
```
agnivade, line 163: Any reason we are not wrapping the error here ?

store/sqlstore/command_store.go
```diff
@@ -91,54 +96,56 @@ func (s SqlCommandStore) GetByTrigger(teamId string, trigger string) (*model.Com
...
 
 	return nil
 }
 
-func (s SqlCommandStore) PermanentDeleteByTeam(teamId string) *model.AppError {
+func (s SqlCommandStore) PermanentDeleteByTeam(teamId string) error {
 	_, err := s.GetMaster().Exec("DELETE FROM Commands WHERE TeamId = :TeamId", map[string]interface{}{"TeamId": teamId})
 	if err != nil {
-		return model.NewAppError("SqlCommandStore.DeleteByTeam", "store.sql_command.save.delete_perm.app_error", nil, "id="+teamId+", err="+err.Error(), http.StatusInternalServerError)
+		return store.NewErrInvalidInput("Command", "TeamId", teamId)
```
agnivade, line 120: We are not differentiating between invalid teamId and an actual error in the sql command. It may not be ideal, but I'd like to keep that consistency in this PR. That just means we can use `errors.Wrapf` to wrap and return.

store/sqlstore/command_store.go
```diff
@@ -91,54 +96,56 @@ func (s SqlCommandStore) GetByTrigger(teamId string, trigger string) (*model.Com
...
 	}
 	return nil
 }
 
-func (s SqlCommandStore) PermanentDeleteByUser(userId string) *model.AppError {
+func (s SqlCommandStore) PermanentDeleteByUser(userId string) error {
 	_, err := s.GetMaster().Exec("DELETE FROM Commands WHERE CreatorId = :UserId", map[string]interface{}{"UserId": userId})
 	if err != nil {
-		return model.NewAppError("SqlCommandStore.DeleteByUser", "store.sql_command.save.delete_perm.app_error", nil, "id="+userId+", err="+err.Error(), http.StatusInternalServerError)
+		return store.NewErrInvalidInput("Command", "userId", userId)
```
agnivade, line 128: Same as above. Let's just wrap and return.

store/sqlstore/command_store.go
```diff
@@ -44,9 +47,9 @@ func (s SqlCommandStore) createIndexesIfNotExists() {
 	s.CreateIndexIfNotExists("idx_command_delete_at", "Commands", "DeleteAt")
 }
 
-func (s SqlCommandStore) Save(command *model.Command) (*model.Command, *model.AppError) {
+func (s SqlCommandStore) Save(command *model.Command) (*model.Command, error) {
 	if len(command.Id) > 0 {
-		return nil, model.NewAppError("SqlCommandStore.Save", "store.sql_command.save.saving_overwrite.app_error", nil, "id="+command.Id, http.StatusBadRequest)
+		return nil, fmt.Errorf("cannot overwrite existing command: command_id=%s", command.Id)
```
agnivade, line 52: I'd make this an invalid input.

store/sqlstore/command_store.go
```diff
@@ -91,54 +96,56 @@ func (s SqlCommandStore) GetByTrigger(teamId string, trigger string) (*model.Com
 		query = "SELECT * FROM Commands WHERE TeamId = :TeamId AND \"trigger\" = :Trigger AND DeleteAt = 0"
 	}
 
-	if err := s.GetReplica().SelectOne(&command, query, map[string]interface{}{"TeamId": teamId, "Trigger": trigger}); err != nil {
-		return nil, model.NewAppError("SqlCommandStore.GetByTrigger", "store.sql_command.get_by_trigger.app_error", nil, "teamId="+teamId+", trigger="+trigger+", err="+err.Error(), http.StatusInternalServerError)
+	if err := s.GetReplica().SelectOne(&command, query, map[string]interface{}{"TeamId": teamId, "Trigger": trigger}); err == sql.ErrNoRows {
+		return nil, store.NewErrNotFound("Command", trigger)
```
agnivade, line 100: Let's also add the `teamId` while we are here. Perhaps make a string concatenation to create the final string like done in the original code.

store/sqlstore/command_store.go
```diff
@@ -55,33 +58,35 @@ func (s SqlCommandStore) Save(command *model.Command) (*model.Command, *model.Ap
...
 	return &command, nil
 }
 
-func (s SqlCommandStore) GetByTeam(teamId string) ([]*model.Command, *model.AppError) {
+func (s SqlCommandStore) GetByTeam(teamId string) ([]*model.Command, error) {
 	var commands []*model.Command
 
 	if _, err := s.GetReplica().Select(&commands, "SELECT * FROM Commands WHERE TeamId = :TeamId AND DeleteAt = 0", map[string]interface{}{"TeamId": teamId}); err != nil {
-		return nil, model.NewAppError("SqlCommandStore.GetByTeam", "store.sql_command.save.get_team.app_error", nil, "teamId="+teamId+", err="+err.Error(), http.StatusInternalServerError)
+		return nil, errors.Wrapf(err, "selectbyteam: team_id=%s", teamId)
```
agnivade, line 83: The convention used is to just prefix by the Gorp method, which would be "select: ".

store/sqlstore/command_store.go
```diff
@@ -91,54 +96,56 @@ func (s SqlCommandStore) GetByTrigger(teamId string, trigger string) (*model.Com
 		query = "SELECT * FROM Commands WHERE TeamId = :TeamId AND \"trigger\" = :Trigger AND DeleteAt = 0"
 	}
 
-	if err := s.GetReplica().SelectOne(&command, query, map[string]interface{}{"TeamId": teamId, "Trigger": trigger}); err != nil {
-		return nil, model.NewAppError("SqlCommandStore.GetByTrigger", "store.sql_command.get_by_trigger.app_error", nil, "teamId="+teamId+", trigger="+trigger+", err="+err.Error(), http.StatusInternalServerError)
+	if err := s.GetReplica().SelectOne(&command, query, map[string]interface{}{"TeamId": teamId, "Trigger": trigger}); err == sql.ErrNoRows {
+		return nil, store.NewErrNotFound("Command", trigger)
+	} else if err != nil {
+		return nil, errors.Wrapf(err, "selectbytrigger: team_id=%s, trigger=%s", teamId, trigger)
```
agnivade, line 102: "selectone: " is more consistent than "selectbytrigger: ".

i18n/en.json
```diff
@@ -2998,6 +2998,46 @@
     "id": "app.cluster.404.app_error",
     "translation": "Cluster API endpoint not found."
   },
+  {
```
agnivade, line 3001: Let's also remove the keys from the store layer which got removed. They are not required anymore.

app/analytics.go
```diff
@@ -244,7 +247,15 @@ func (a *App) GetAnalytics(name string, teamId string) (model.AnalyticsRows, *mo
 		commandChan := make(chan store.StoreResult, 1)
 		go func() {
 			c, err := a.Srv().Store.Command().AnalyticsCommandCount(teamId)
-			commandChan <- store.StoreResult{Data: c, Err: err}
+			var appErr *model.AppError
+			if err != nil {
+				switch {
```
agnivade, line 252: For the store methods which don't actually return an AppError, there is no need to do check for this. If it just returns a normal error, we can directly create an AppError and return it / or pass to the channel as done here.

This comment is also applicable to all the other methods in the app layer. From what I see, only Save and Update return an AppError (Save would also return an InvalidInput from my comment below). And Get and GetByTrigger return a NotFound. So only in those cases, we need a switch-case. For the others, we can just directly return.

joshuabezaleel: Hi @agnivade , first of all thank you for the really thorough and thoughtful review! Truly appreciate it :slightly_smiling_face: 
1. I think I can change this switch-case on other part like `app/team.go` or `app/user.go` but on this particular case in `app/analytics.go` not using a switch-case got me an error panic of `invalid memory address or nil pointer dereference` from the `TestGetAnalyticsOld` test so that's why I used switch-case on this one. What do you think?

store/sqlstore/command_store.go
joshuabezaleel: Also I am still not sure on this one even though I've put up a patch. Looking forward to hearing on this more :slightly_smiling_face:

app/analytics.go
agnivade: Hi @joshuabezaleel - so sorry for the delay here, I somehow completely missed your notification. Regarding this issue, now the `StoreResult` has a `NErr` field added for the purposes of this migration. So you need not check for the error inside the goroutine, but just assign the error to `NErr` than `Err` and check for the value in the caller.

agnivade:
Hi @joshuabezaleel - Sorry for the delay, I replied to your comment. Please let me know if there's anything else I can help with. Thanks for your time !

app/analytics.go
joshuabezaleel: No worries at all! Thank you for the pointing it out, @agnivade ! Sorry for missing and not noticing that there is a new field of `NErr`.

joshuabezaleel:
I am should be the one thanking though, @agnivade :slightly_smiling_face: Thank you for the really helpful review! That should do the trick. PTAL :smile:

agnivade:
Is this ready for review ? It is still in the draft stage, so wanted to confirm.

joshuabezaleel:
Sorry that I forgot to change the status of the PR from draft, @agnivade . Yes, it is ready, all tests should be okay by now :slightly_smiling_face:

app/analytics.go
```diff
@@ -244,7 +244,7 @@ func (a *App) GetAnalytics(name string, teamId string) (model.AnalyticsRows, *mo
 		commandChan := make(chan store.StoreResult, 1)
 		go func() {
 			c, err := a.Srv().Store.Command().AnalyticsCommandCount(teamId)
-			commandChan <- store.StoreResult{Data: c, Err: err}
+			commandChan <- store.StoreResult{Data: c, NErr: err}
```
agnivade, line 247: We have missed actually checking this error below. In L310, we should be checking for NErr rather than Err.

app/command.go
```diff
@@ -615,21 +620,36 @@ func (a *App) CreateCommand(cmd *model.Command) (*model.Command, *model.AppError
...
+	return command, nil
 }
 
 func (a *App) GetCommand(commandId string) (*model.Command, *model.AppError) {
 	if !*a.Config().ServiceSettings.EnableCommands {
 		return nil, model.NewAppError("GetCommand", "api.command.disabled.app_error", nil, "", http.StatusNotImplemented)
 	}
 
-	cmd, err := a.Srv().Store.Command().Get(commandId)
+	command, err := a.Srv().Store.Command().Get(commandId)
```
agnivade, line 642: Let's keep this as `cmd` to avoid any unncessary changes. Go also prefers short names, so `cmd` is preferred.

app/command.go
```diff
@@ -646,15 +666,38 @@ func (a *App) UpdateCommand(oldCmd, updatedCmd *model.Command) (*model.Command,
 	updatedCmd.CreatorId = oldCmd.CreatorId
 	updatedCmd.TeamId = oldCmd.TeamId
 
-	return a.Srv().Store.Command().Update(updatedCmd)
+	command, err := a.Srv().Store.Command().Update(updatedCmd)
+	if err != nil {
+		var nfErr *store.ErrNotFound
+		var appErr *model.AppError
+		switch {
+		case errors.As(err, &nfErr):
```
agnivade, line 674: I don't see the `Update` method returning a NotFound error. But perhaps I have missed something ?

agnivade:
That's just a flaky step. When this happens, just go to the CircleCI UI and restart it.

agnivade:
@joshuabezaleel - Oh sorry about (2), I just checked the code. That's not required because we are just migrating errors from command_store. Those goroutines get data from other stores. So in short, only errors which are migrated will need to use `nErr` and set the `NErr` field in StoreResult. All others will use `Err`. If you have shadowing issues, you can use `err2` for those variables which are not migrated, but still show as shadow errors.

Sorry for the confusion.

joshuabezaleel:
Should be okay by now. Thank you so much for the great help, @agnivade ! Truly appreciate it :slightly_smiling_face:

agnivade:
@ogi-m - All testing related to slash commands is required here. Thanks.

PR #14646, opened 2020-05-25
Migrate tool dependencies to go.tools.mod

### LINKED ISSUE #13088: Migrate tool dependencies to be tracked in go.mod
### Problem

The mattermost codebase uses several tools like `megacheck`, `mgotool`, `mockery` etc. to achieve various tasks. However, the versions of these tools are not tracked in go.mod and directly fetched from the tip of their respective repos. This is a source of discrepancy and possible bugs because one single update from the tool developer might break things in the codebase.

It is also not straightforward to track the dependencies because for it to appear in go.mod, it has to be part of the repo codebase. Fortunately, we have an official recommendation on how to approach this: https://github.com/golang/go/wiki/Modules#how-can-i-track-tool-dependencies-for-a-module.

#### Approach:

The core idea is to use a tools.go file under a separate build tag so that it is ignored when building the repo. But the dependencies inside the file will get tracked in go.mod because commands like `go  mod tidy` looks into _all_ files under all build tags.

However, there is a slight complication in our case because all of our dependencies are vendored. But we don't want the tools also to be vendored, just their versions tracked.

An effective way around this is to have a sub-module with a separate go.mod file under a `tools` directory. While sub-modules within a single repo are discouraged, but in this case, there is no extra code in that module and its sole purpose is to track the tool dependencies. And as a concrete example, [golangci-lint](https://github.com/golangci/golangci-lint/tree/master/tools) also does the same.

This solves both of our problems of tracking tool dependencies as well as not getting them vendored.

EDIT: We will also be installing the tools in a `bin` directory inside the repo root so as to not pollute any global namespace with one repo's tools. As different repos can use different versions of a tool. And it is good to keep one repo's tools localized within the repo. Similar to `node_modules/bin` directory.

#### Future solutions:

https://github.com/golang/go/issues/34506#issue-497899555 is the correct solution to this where we can have separate go.mod files - one for the repo, and one for the tools. For eg. `go.mod` and `go.tools.mod`. And we can use `go.tools.mod` when working with tools. But this will only probably be available in 1.15.

Until then, the above approach is good enough.

I will be sending a PR tagging this issue as a reference.

/cc @hanzei @reflog

agnivade:
Go 1.14 introduces usage of alternate go.mod files to track
other dependencies which are not related to the main app.
We use this to track all tool dependencies so that everybody uses
the same version of all tools, including CI.

This will prevent version conflicts due to everybody using different
versions of the tools.
And it will not try to upgrade the tool version, every single time,
one runs the tool command.

While here, we also re-generate some filestore mocks which weren't up to date.

Fixes #13088


agnivade:
Waiting on https://github.com/mattermost/enterprise/pull/665 to be merged.

agnivade:
@lindy65 - This is verified by CI itself. I don't think you could do much QA here.

PR #14648, opened 2020-05-25
MM-24481: Use atomics for load/store of client configs

agnivade:
There are 3 client config related variables which are set in regenerateClientConfig
but can be read from public methods in the App struct.

We use the same approach as the license fields in the App struct and use atomic.Value
for modifying them.

### Ticket link

https://mattermost.atlassian.net/browse/MM-24481

### DIFF
--- a/app/config.go
+++ b/app/config.go
@@ -72,15 +72,15 @@ func (a *App) ReloadConfig() error {
 }
 
 func (a *App) ClientConfig() map[string]string {
-	return a.Srv().clientConfig
+	return a.Srv().clientConfig.Load().(map[string]string)
 }
 
 func (a *App) ClientConfigHash() string {
-	return a.Srv().clientConfigHash
+	return a.Srv().clientConfigHash.Load().(string)
 }
 
 func (a *App) LimitedClientConfig() map[string]string {
-	return a.Srv().limitedClientConfig
+	return a.Srv().limitedClientConfig.Load().(map[string]string)
 }
 
 // Registers a function with a given listener to be called when the config is reloaded and may have changed. The function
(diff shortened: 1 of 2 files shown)


app/config.go
```diff
@@ -319,9 +319,9 @@ func (a *App) regenerateClientConfig() {
 	}
 
 	clientConfigJSON, _ := json.Marshal(clientConfig)
-	a.Srv().clientConfig = clientConfig
-	a.Srv().limitedClientConfig = limitedClientConfig
-	a.Srv().clientConfigHash = fmt.Sprintf("%x", md5.Sum(clientConfigJSON))
+	a.Srv().clientConfig.Store(clientConfig)
+	a.Srv().limitedClientConfig.Store(limitedClientConfig)
+	a.Srv().clientConfigHash.Store(fmt.Sprintf("%x", md5.Sum(clientConfigJSON)))
```
streamer45, line 324: Just want to confirm that we are making sure that this is always going to execute before any call that loads the value (e.g. `ClientConfig()`).

agnivade: Yes, this gets called during server startup from `RunOldAppInitialization` using our wonderful FakeApp adapter.

agnivade:
This will be hard for Lindy to test. I will request Claudio to do some peer testing and confirm that it's fixed, since it was caught in the load test originally.

lindy65:
Thanks @agnivade - I've added the QA: Review Done label and will pass this through from QA's perspective 👍 

Will @streamer45 be peer testing here on the PR as well as post-merge on the resolved ticket?

streamer45:
> Will @streamer45 be peer testing here on the PR as well as post-merge on the resolved ticket?

Yes. I can take care of this.

PR #14650, opened 2020-05-25
[MM-24088] Use Raw SQL query for PluginStore.Get instead of GORP

ethervoid:
#### Summary

As part of performance improvements, we're moving the top used queries to avoid using gorp and get rid of the overhead caused by the reflection usage

#### Ticket Link

https://mattermost.atlassian.net/browse/MM-24088


agnivade:
This would be hard to load test because our tests don't hit this endpoint yet. We can go with another micro-benchmark and just do a relative comparison. Would you like to write one ?

store/sqlstore/plugin_store.go
```diff
@@ -9,6 +9,7 @@ import (
 	"fmt"
 	"net/http"
 
+	sq "github.com/Masterminds/squirrel"
```
agnivade, line 12: Nit: this should be in a separate group.

store/sqlstore/plugin_store.go
```diff
@@ -219,16 +220,39 @@ func (ps SqlPluginStore) SetWithOptions(pluginId string, key string, value []byt
...
+		)
+	}
+
+	query := ps.getQueryBuilder().Select("PluginId, PKey, PValue, ExpireAt").
+		From("PluginKeyValueStore").
+		Where(sq.Eq{"PluginId": pluginId}).
+		Where(sq.Eq{"PKey": key}).
+		Where(sq.Or{sq.Eq{"ExpireAt": 0}, sq.Gt{"ExpireAt": currentTime}})
+	queryString, args, err := query.ToSql()
+
```
agnivade, line 241: Nit: this line can be removed.

store/sqlstore/plugin_store.go
```diff
@@ -219,16 +220,39 @@ func (ps SqlPluginStore) SetWithOptions(pluginId string, key string, value []byt
...
+	if err != nil {
+		return failure(err, http.StatusInternalServerError)
+	}
+	row := ps.GetReplica().Db.QueryRow(queryString, args...)
+
+	var kv model.PluginKeyValue
```
agnivade, line 247: Nit:
```suggestion

	row := ps.GetReplica().Db.QueryRow(queryString, args...)
	var kv model.PluginKeyValue
```

store/sqlstore/plugin_store.go
```diff
@@ -219,16 +220,39 @@ func (ps SqlPluginStore) SetWithOptions(pluginId string, key string, value []byt
...
+	if err := row.Scan(&kv.PluginId, &kv.Key, &kv.Value, &kv.ExpireAt); err != nil {
 		if err == sql.ErrNoRows {
-			return nil, model.NewAppError("SqlPluginStore.Get", "store.sql_plugin_store.get.app_error", nil, fmt.Sprintf("plugin_id=%v, key=%v, err=%v", pluginId, key, err.Error()), http.StatusNotFound)
+			return failure(err, http.StatusNotFound)
```
lieut-data, line 250: nit: it feels slightly more obvious to write:
```
return nil, failure(err, http.StatusNotFound)
```

instead of having `failure` always return a `nil` `*model.PluginKeyValue`. 0/5

ethervoid:
Well, very disappointing results:

### Microbenchmark (16 cores have a really bad delta :\)

```
$ benchstat benchmark_plugin_get_old.log benchmark_plugin_get_new.log 
name                       old time/op    new time/op    delta
PluginStoreGetMySQL           171µs ± 5%     176µs ± 5%   +2.72%  (p=0.014 n=9+9)
PluginStoreGetPostgres        166µs ± 3%     179µs ± 4%   +7.49%  (p=0.000 n=10+10)
PluginStoreGetMySQL-2         186µs ± 5%     175µs ± 6%   -6.13%  (p=0.000 n=10+10)
PluginStoreGetPostgres-2      174µs ± 3%     175µs ± 6%     ~     (p=0.971 n=10+10)
PluginStoreGetMySQL-4         182µs ± 4%     173µs ± 3%   -5.24%  (p=0.000 n=10+10)
PluginStoreGetPostgres-4      175µs ± 4%     175µs ± 6%     ~     (p=0.796 n=10+10)
PluginStoreGetMySQL-8         188µs ± 5%     180µs ± 8%   -4.04%  (p=0.029 n=10+10)
PluginStoreGetPostgres-8      174µs ± 6%     175µs ± 4%     ~     (p=0.853 n=10+10)
PluginStoreGetMySQL-16        187µs ± 7%     173µs ± 5%   -7.45%  (p=0.000 n=10+10)
PluginStoreGetPostgres-16     208µs ±48%     171µs ± 7%  -17.87%  (p=0.003 n=9+10)

name                       old alloc/op   new alloc/op   delta
PluginStoreGetMySQL          3.98kB ± 0%    7.18kB ± 0%  +80.68%  (p=0.000 n=10+10)
PluginStoreGetPostgres       4.04kB ± 0%    7.80kB ± 0%  +93.07%  (p=0.000 n=10+10)
PluginStoreGetMySQL-2        3.98kB ± 0%    7.18kB ± 0%  +80.68%  (p=0.000 n=9+10)
PluginStoreGetPostgres-2     4.04kB ± 0%    7.80kB ± 0%  +93.07%  (p=0.000 n=8+10)
PluginStoreGetMySQL-4        3.98kB ± 0%    7.18kB ± 0%  +80.68%  (p=0.000 n=9+10)
PluginStoreGetPostgres-4     4.04kB ± 0%    7.80kB ± 0%  +93.01%  (p=0.000 n=9+10)
PluginStoreGetMySQL-8        4.01kB ± 0%    7.18kB ± 0%  +79.37%  (p=0.000 n=10+9)
PluginStoreGetPostgres-8     4.08kB ± 0%    7.80kB ± 0%  +91.33%  (p=0.000 n=9+9)
PluginStoreGetMySQL-16       4.01kB ± 1%    7.18kB ± 0%  +79.29%  (p=0.000 n=9+10)
PluginStoreGetPostgres-16    4.07kB ± 1%    7.80kB ± 0%  +91.56%  (p=0.000 n=10+10)

name                       old allocs/op  new allocs/op  delta
PluginStoreGetMySQL             128 ± 0%       152 ± 0%  +18.75%  (p=0.000 n=10+10)
PluginStoreGetPostgres          134 ± 0%       171 ± 0%  +27.61%  (p=0.000 n=10+10)
PluginStoreGetMySQL-2           128 ± 0%       152 ± 0%  +18.75%  (p=0.000 n=10+10)
PluginStoreGetPostgres-2        134 ± 0%       171 ± 0%  +27.61%  (p=0.000 n=10+10)
PluginStoreGetMySQL-4           128 ± 0%       152 ± 0%  +18.75%  (p=0.000 n=10+10)
PluginStoreGetPostgres-4        134 ± 0%       171 ± 0%  +27.61%  (p=0.000 n=10+10)
PluginStoreGetMySQL-8           128 ± 0%       152 ± 0%  +18.75%  (p=0.000 n=10+10)
PluginStoreGetPostgres-8        134 ± 0%       171 ± 0%  +27.61%  (p=0.000 n=10+10)
PluginStoreGetMySQL-16          128 ± 0%       152 ± 0%  +18.75%  (p=0.000 n=10+10)
PluginStoreGetPostgres-16       134 ± 0%       171 ± 0%  +27.61%  (p=0.000 n=10+10)
```

### Load tester

I've installed the [remind plugin](https://github.com/scottleedavis/mattermost-plugin-remind) in the load tester server. I've used the reminder plugin because is simple and is calling the `KVGet` method [every second](https://github.com/scottleedavis/mattermost-plugin-remind/blob/master/server/scheduler.go#L316) with a scheduler.

Also, I've added the necessary code to execute commands by the users so every user was creating reminders.

Finally, I've created two panels in grafana to compare and the result was a bit disappointing

Screenshot from 2020-05-25 21-20-14

Sorry for the popup in the image (didn't realize it was there :sweat: ). Here is the image of the hidden part:
Screenshot from 2020-05-25 20-50-31

As you can see the gain is averaging about 4-5ms but we're adding more memory/allocation overhead.

agnivade:
We are abusing the microbenchmark to test a DB improvement, so it is expected that the numbers aren't what we expect. Check the p-values in some cases, it ranges from 0.7-0.9. So there is a bit of variance in the system because you are going to the DB.

The Grafana graphs are more interesting. I am wondering how are we getting 20ms as the time taken ? Is the app and the DB so far apart ? Usually it should take around 4-5ms. Can you install the DB and the app in the same machine and run the test again ? I am trying to magnify just the impact made by removing Gorp by reducing the DB roundtrip time as much as possible.

ethervoid:
Here is the same benchmark but using a local database. No great improvement to be honest

mm_24088

agnivade:
Alright, thanks for the work @ethervoid ! I don't see any harm in merging this. If my guess is correct, since this just selects one row, the performance gap is very small. But other queries scan multiple rows which gives us a larger window to reduce.

Good to go from my side.

ethervoid:
@agnivade agree!

agnivade:
Gentle ping again @ogi-m.

PR #14651, opened 2020-05-25
MM-25014 - Default Group Name to null, until AllowGroupReference is set

sbishel:
#### Summary
Converts the Name field to a pointer, so that the field can be null in the database. Most of the changes are in test files, but there ended up being a bunch of code to change.
#### Ticket Link
https://mattermost.atlassian.net/browse/MM-25014
https://mattermost.atlassian.net/browse/MM-25446

#### Related PR
https://github.com/mattermost/mattermost-webapp/pull/5569
https://github.com/mattermost/enterprise/pull/666


app/notification.go
```diff
@@ -658,7 +658,7 @@ func (m *ExplicitMentions) addGroupMention(word string, groups map[string]*model
 		m.GroupMentions = make(map[string]*model.Group)
 	}
 
-	m.GroupMentions[group.Name] = group
+	m.GroupMentions[*group.Name] = group
```
mkraft, line 661: Wouldn't this—and any other instances—require a `nil` check to avoid a nil pointer dereference?

sbishel: I had convinced myself it wasn't necessary. But let me double check.

sbishel: Agree.  I wrapped these in a check.

sbishel:
@srkgupta - When testing this, the database upgrade must run for v5.24. Otherwise the Name field will not be nulled and you will continue to see the alpha/numeric name. Also, it will reset AllowReference = False. So, if you have existing group mentions they will be turned off. This was to keep Name and AllowReference in sync. (null and false).

PR #14654, opened 2020-05-25
[MM-24123] Fix flaky TestGetLinkMetadata

streamer45:
#### Summary

PR fixes flaky `TestGetLinkMetadata`.
The main issue was relying on [strings.Contains()](https://golang.org/pkg/strings/#Contains) as a method to check the error's type.
Unfortunately `net/http` package can return several different errors for the same underlying reason (in this case, a connection timeout).

How to reproduce:
```go
package main

import (
	"fmt"
	"net/http"
	"strings"
	"time"
)

type CustomTransport struct {
	Transport http.RoundTripper
}

func (t *CustomTransport) RoundTrip(req *http.Request) (*http.Response, error) {
	return t.Transport.RoundTrip(req)
}

func main() {
	// Using a non-routable host to generate timeouts.
	requestURL := "http://10.255.255.1"

	// It's important to use a custom roundtripper to activate the internal
	// codepath that generates the racy behaviour.
	transport := &CustomTransport{
		http.DefaultTransport,
	}

	client := &http.Client{
		Timeout:   100 * time.Millisecond,
		Transport: transport,
	}

	for {
		req, err := http.NewRequest("GET", requestURL, nil)
		if err != nil {
			return
		}
		if _, err := client.Do(req); err != nil {
			fmt.Println(err.Error())
			if !strings.Contains(err.Error(), "Client.Timeout") {
				panic("")
			}
		}
	}
}
```

Internally [transport.getConn()](https://github.com/golang/go/blob/f65ad0dda7ffef9397d1aaa47259ad4d4f12474f/src/net/http/transport.go#L1316-L1317) and [client.SetRequestCancel()](https://github.com/golang/go/blob/f65ad0dda7ffef9397d1aaa47259ad4d4f12474f/src/net/http/client.go#L402-L403) were racing to return the timeout error.

#### Ticket

https://mattermost.atlassian.net/browse/MM-24123


streamer45:
@ogi Not sure there's a reliable way to test this. I was able (at times) to reproduce the failure running:

```
go test -v -run="TestGetLinkMetadata/should_cache_timeout_errors" -count=100 -failfast ./app
```
In any case, it's a test only modification which doesn't affect the product.

streamer45:
After some testing I was able to reliably use [os.IsTimeout()](https://golang.org/pkg/os/#IsTimeout) to detect timeouts so I added a check for it.

PR #14664, opened 2020-05-26
[MM-25538] Adds the termOperator and takes it into account for terms and hashtag queries

mgdelacroix:
#### Summary
Adds the term operator for both term and hashtag queries and for normal and excluded terms

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-25538

### DIFF
--- a/services/searchengine/bleveengine/search.go
+++ b/services/searchengine/bleveengine/search.go
@@ -43,6 +43,11 @@ func (b *BleveEngine) SearchPosts(channels *model.ChannelList, searchParams []*m
 	filters = append(filters, typeQ)
 
 	for i, params := range searchParams {
+		var termOperator query.MatchQueryOperator = query.MatchQueryOperatorAnd
+		if searchParams[0].OrTerms {
+			termOperator = query.MatchQueryOperatorOr
+		}
+
 		// Date, channels and FromUsers filters come in all
 		// searchParams iteration, and as they are global to the
 		// query, we only need to process them once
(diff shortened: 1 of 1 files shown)


services/searchengine/bleveengine/search.go
```diff
@@ -43,6 +43,11 @@ func (b *BleveEngine) SearchPosts(channels *model.ChannelList, searchParams []*m
 	filters = append(filters, typeQ)
 
 	for i, params := range searchParams {
+		var termOperator query.MatchQueryOperator = query.MatchQueryOperatorAnd
```
ethervoid, line 46: Question. Why don't do `termOperator := query.MatchQueryOperatorAnd`?

mgdelacroix: I'm not really sure why, but the type inference system doesn't play well with that assignment particularly:

```
# github.com/mattermost/mattermost-server/v5/services/searchengine/bleveengine
services/searchengine/bleveengine/search.go:149:25: cannot use termOperator (type int) as type query.MatchQueryOperator in argument to hashtagQ.SetOperator
services/searchengine/bleveengine/search.go:154:25: cannot use termOperator (type int) as type query.MatchQueryOperator in argument to hashtagQ.SetOperator
services/searchengine/bleveengine/search.go:161:25: cannot use termOperator (type int) as type query.MatchQueryOperator in argument to messageQ.SetOperator
services/searchengine/bleveengine/search.go:168:25: cannot use termOperator (type int) as type query.MatchQueryOperator in argument to messageQ.SetOperator
```

ethervoid: Yes, is a problem with how the constant is created. The type is `MatchOperator` but the constant is being created as int as you can see [here](https://github.com/blevesearch/bleve/blob/master/search/query/match.go#L40-L42).

I've sent [a PR](https://github.com/blevesearch/bleve/pull/1410) in order to fix that in the upcoming versions if it's merged

mgdelacroix: Awesome! Thanks for taking the time to send the PR!! :heart:

PR #14667, opened 2020-05-27
MM-25115- migrate post, session and status cache to cache2

liusy182:
#### Summary

This CR is a continuation from [#14496](https://github.com/mattermost/mattermost-server/pull/14496). 

- It migrates post, session and status cache to cache2 package, by using the new cache `Provider` interface. 

- This CR also provides a new field called `CacheProvider2` in struct `app/server`. This facilitates further migration as there are many existing usages of the `server.CacheProvider` which are based on the old cache interface.

#### Ticket Link

https://mattermost.atlassian.net/browse/MM-25115

### DIFF
--- a/app/post.go
+++ b/app/post.go
@@ -6,6 +6,7 @@ package app
 import (
 	"encoding/json"
 	"fmt"
+	"github.com/mattermost/mattermost-server/v5/services/cache2"
 	"net/http"
 	"strings"
 	"sync"
@@ -111,15 +112,17 @@ func (a *App) deduplicateCreatePost(post *model.Post) (foundPost *model.Post, er
 
 	// Query the cache atomically for the given pending post id, saving a record if
 	// it hasn't previously been seen.
-	value, loaded := a.Srv().seenPendingPostIdsCache.GetOrAdd(post.PendingPostId, unknownPostId, PENDING_POST_IDS_CACHE_TTL)
+	var postId string
+	errCache := a.Srv().seenPendingPostIdsCache.Get(post.PendingPostId, &postId)
 
-	// If we were the first thread to save this pending post id into the cache,
-	// proceed with create post normally.
-	if !loaded {
+	if errCache == cache2.ErrKeyNotFound {
+		a.Srv().seenPendingPostIdsCache.SetWithExpiry(post.PendingPostId, unknownPostId, PENDING_POST_IDS_CACHE_TTL)
 		return nil, nil
 	}
 
-	postId := value.(string)
+	if errCache != nil {
+		return nil, model.NewAppError("errorGetPostId", "api.post.error_get_post_id.pending", nil, "", http.StatusInternalServerError)
+	}
 
 	// If another thread saved the cache record, but hasn't yet updated it with the actual post
 	// id (because it's still saving), notify the client with an error. Ideally, we'd wait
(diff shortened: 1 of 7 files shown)


jespino:
@liusy182 I added the i18n strings :)

liusy182:
> @liusy182 I added the i18n strings :)

got it. thanks for fixing!

app/post.go
```diff
@@ -111,15 +112,17 @@ func (a *App) deduplicateCreatePost(post *model.Post) (foundPost *model.Post, er
 
 	// Query the cache atomically for the given pending post id, saving a record if
 	// it hasn't previously been seen.
-	value, loaded := a.Srv().seenPendingPostIdsCache.GetOrAdd(post.PendingPostId, unknownPostId, PENDING_POST_IDS_CACHE_TTL)
+	var postId string
+	errCache := a.Srv().seenPendingPostIdsCache.Get(post.PendingPostId, &postId)
```
agnivade, line 116: The convention we have been using is `nErr` for `error` types.

app/post.go
```diff
@@ -111,15 +112,17 @@ func (a *App) deduplicateCreatePost(post *model.Post) (foundPost *model.Post, er
...
```
agnivade, line 117: Nit: let's remove the empty line.

agnivade:
Definitely some regression in p99. But okay in average. Kinda what is expected.

Original:
original

PR:
serialized

/cc @grundleborg

agnivade:
@lindy65  - Some testing around posts, sessions and statuses is needed. It is recommended to test in an HA setup to see if there are any edge-cases.

Look out for any sort of unexpected behavior - like getting stale data, or things not behaving correctly.

grundleborg:
@agnivade one other question related to the load test - is there any discernable change in CPU usage on the app servers before/after this change?

agnivade:
@grundleborg - Yes.

Original:
original


PR:
PR

It is expected because we are using gob serialization which is probably the lowest performant of encoders out there. But like we discussed in the server meeting, we decided on focus on getting this feature in first, and then focus on performance improvements. This should not be the same once we move to a better serialization algorithm, which should be done before 5.26.

grundleborg:
OK, so about a 25% increase in CPU usage. I think we need to be careful then, as this means some servers that are "close to capacity" now will not be over capacity after this is merged.

I've no objection to it being merged at this stage, but we should have the following in place:
1. A plan for what we're going to do to reduce that CPU usage penalty *before* 5.26
2. A plan for an emergency temporary revert if, as we approach the feature freeze, we haven't got (1) sorted, so we can pull this out of 5.26 and reapply to 5.28.

agnivade:
Totally agree.

1. The plan is to switch to a better serialization algorithm. I can do it any time.
2. I don't believe 2 will be necessary. But in case it is, I think the simplest approach would be to revert all the commits that made this change. One concern might be the code churn produced by reverting multiple commits. But it's not too bad, we are just changing one interface with another. We could also have an internal constant flag which determines whether to switch to serialized cache or not, in which case, we need to bring back the old code again.

@liusy182 - Do you have an idea as to how many more commits would be required to switch over the rest of the caches to cache2 ? I think this is the 2nd commit right ? It might be better to combine the rest in a single commit so that it becomes easy to revert later (in case).

liusy182:
I already have the [3rd commit](https://github.com/mattermost/mattermost-server/pull/14691) and I was planning to have 10~15 smaller commits dependent on that, with each one only migrating 1 cache user.

The comments totally make sense. To summarize, we have 2 options:

1. make 1 single giant commit to do the migration so that it is easier to revert back later. It will modify 20+ files or even more.

2. have a config option to allow switching back to the old cache implementation. This will ease the urgency of doing performance optimization in the same release.

I have some thoughts on No.2. We might be able to have the `cache2.LRU` wraps `cache/lru.LRU`. Internally, `cache2.LRU` reads a config and decides whether to use the new serializable implementation or the old `cache/lru.LRU` implementation. Do we want to explore this option or No.1 ?

@agnivade @grundleborg

agnivade:
Personally, I think 1 is fine. I don't see any issue with a big commit. It's essentially replacing interface names and changing some input params. A big commit will also help me get the full performance impact with a single commit.

5.26 release is still far away and I can take care of the performance improvement part.

liusy182:
sounds good. I will incorporate the rest of the changes into #14691 @agnivade

agnivade:
Awesome, thanks @liusy182 ! Excited to see the full PR.

PR #14671, opened 2020-05-27
[MM-24560] Clear the team cache after a team scheme is deleted

fmunshi:
#### Summary
- When a team is deleted the schemeId on a team is set to `''` however the team cache is never cleared so intermittently deleting a team scheme will result in users not being able to access anything in the team that the deleted scheme was assigned to. 
- There was already code to clear channel channel caches if the scheme was of scope `channel` but in the case of scope `team` no cache was cleared, this PR fixes that

#### Ticket Link
- https://mattermost.atlassian.net/browse/MM-24560


amyblais:
@fmunshi Kind reminder to cherry-pick to v5.25.

PR #14673, opened 2020-05-27
MM-24674 Update channel members by group to look at distinct timezones correctly

fmunshi:
#### Summary
- Timezone is stored as text in the format of a map in the database so doing a `distinct` on just the column would return more unique entries than expected for ex the following three timezones would previously be treated as three different timezones, now they will be counted as a single timezone in timezone count: 
`{"automaticTimezone":"America/Toronto","manualTimezone":"Asia/Kuching","useAutomaticTimezone":"true"}`
`{"automaticTimezone":"America/Toronto","manualTimezone":"","useAutomaticTimezone":"true"}`
`{"automaticTimezone":"","manualTimezone":"America/Toronto","useAutomaticTimezone":"false"}`

- This query is quite ugly now and I would appreciate any comments on a way to make it better

#### Ticket Link
- https://mattermost.atlassian.net/browse/MM-24674


store/sqlstore/channel_store.go
```diff
@@ -1793,7 +1793,41 @@ func (s SqlChannelStore) GetMemberCountsByGroup(channelID string, includeTimezon
 	selectStr := "GroupMembers.GroupId, COUNT(ChannelMembers.UserId) AS ChannelMemberCount"
 
 	if includeTimezones {
-		selectStr = "GroupMembers.GroupId, COUNT(ChannelMembers.UserId) AS ChannelMemberCount, COUNT( DISTINCT Users.Timezone ) AS ChannelMemberTimezonesCount"
+		endOfFirstKey := `LOCATE(':', Users.Timezone) + 2`
+		lengthOfFirstValue := `LOCATE(',', Users.Timezone) - LOCATE(':', Users.Timezone) - 3`
+		endOfSecondKey := `LOCATE(',', Users.Timezone) + 19`
+		lengthOfSecondValue := `LOCATE('useAutomaticTimezone', Users.Timezone) - 22 - LOCATE(',', Users.Timezone)`
```
fmunshi, line 1799: this query is now tightly coupled with the order of the keys in this map... not sure if theres a way to avoid while still computing everything in a single query. On the bright side if someone modifies this column at all these tests will break, but then this query will need to be modified again.

fmunshi:
Moving this to work in progress as we have found a better way to do this

fmunshi:
@mkraft @sbishel this is good for review now

fmunshi:
Hey @srkgupta,

The issue you were seeing was actually slightly different - the issue was that I was not ignoring users that do not have timezone set at all so there would always be +1 timezone count.

I modified it now to ignore users without a timezone set similar to how @ all is handled

fmunshi:
@srkgupta are you sure it is the same issue?

I just tried it again and it works as expected for me.

Can you pull the branch down and try again and ensure that the server is restarted before trying.

fmunshi:
I created a video showing what I see https://share.vidyard.com/watch/m1QrRefdecao5pKraqVBy4?

srkgupta:
Hi @fmunshi 
I tried to pull the latest again and restarted the server and reloaded the pages on the browsers. But I still see the same issue as mentioned in my previous comment, i.e. if 1 user has a different timezone, then the group mentions shows it as 1 timezone instead of 2 timezones. Similarly if 2 users have different timezone, then it shows as 2 timezone instead of 3 timezones. This works absolutely fine for `@all` mentions though. The group mention always shows 1 less.

PR #14674, opened 2020-05-27
SqlLinkMetadataStore migrates to plain errors

rvillablanca:
#### Summary

This is a proposal PR to help to migrate the store layer to use idiomatic go errors so this case is for `SqlLinkMetadataStore`.

Fixes https://github.com/mattermost/mattermost-server/issues/14677


rvillablanca:
@agnivade You are leading this kind of changes so you could help with the review I thing :smile:

agnivade:
This is great. Let me create some tickets so that we can track this.

store/sqlstore/link_metadata_store.go
```diff
@@ -47,33 +47,33 @@ func (s SqlLinkMetadataStore) Save(metadata *model.LinkMetadata) (*model.LinkMet
...
+		return nil, errors.Wrap(err, "could not create query with querybuilder")
 	}
 	err = s.GetReplica().SelectOne(&metadata, query, args...)
 	if err != nil {
 		if err == sql.ErrNoRows {
-			return nil, model.NewAppError("SqlLinkMetadataStore.Get", "store.sql_link_metadata.get.app_error", nil, "url="+url+", "+err.Error(), http.StatusNotFound)
+			return nil, store.NewErrNotFound("LinkMetadata", "url="+url)
 		}
-		return nil, model.NewAppError("SqlLinkMetadataStore.Get", "store.sql_link_metadata.get.app_error", nil, "url="+url+", "+err.Error(), http.StatusInternalServerError)
+		return nil, errors.Wrapf(err, "could not get metadata with url=%s", url)
```
agnivade, line 71: For SQL query errors, let's just prefix with the query method, in this case, "selectone". So maybe something like "selectone: url=%s"

rvillablanca: Done

rvillablanca:
> Strange, we are not using this table anywhere in the code.

Not sure if I understand well, but I saw you are actually using this Store in `app/post_metadata.go`

agnivade:
> Not sure if I understood well, but I saw you are actually using this Store in app/post_metadata.go

Ah I see now. Yes indeed. But I was wondering if we are using it, how come we are not touching the code in this PR. Now I see that we are not returning an AppError in the app layer at all. Well, good for us because we have to touch less code.

srkgupta:
Hi @agnivade 
Can you please update the QA Test Steps on the ticket and add details on what screens/functionalities are affected and needs to be tested?

https://mattermost.atlassian.net/browse/MM-25597

agnivade:
@srkgupta - Some testing around embedded links in posts (using markdown or normal) should suffice. Basically we are trying to test the metadata we store for links in posts.

rvillablanca:
@agnivade ready to be integrated :)

PR #14675, opened 2020-05-27
SqlTermsOfService migrates to plain errors

rvillablanca:
#### Summary

This is a proposal PR to help to migrate the store layer to use idiomatic go errors so this case is for `SqlTermsOfService`.

Fixes https://github.com/mattermost/mattermost-server/issues/14678

### DIFF
--- a/i18n/en.json
+++ b/i18n/en.json
@@ -7315,7 +7315,7 @@
     "translation": "No terms of service found."
   },
   {
-    "id": "store.sql_terms_of_service_store.save.existing.app_error",
+    "id": "app.terms_of_service.create.existing.app_error",
     "translation": "Must not call save for existing terms of service."
   },
   {
--- a/i18n/en.json
+++ b/i18n/en.json
@@ -1693,6 +1693,10 @@
     "translation": "Action integration error."
   },
   {
+    "id": "api.post.error_get_post_id.pending",
+    "translation": "Unable to get the pending post."
+  },
+  {
     "id": "api.post.get_message_for_notification.files_sent",
     "translation": {
       "one": "{{.Count}} file sent: {{.Filenames}}",
@@ -2993,6 +2997,14 @@
   {
     "id": "app.channel.post_update_channel_purpose_message.updated_to",
     "translation": "%s updated the channel purpose to: %s"
+  },
+  {
+    "id": "app.channel.update.bad_id",
+    "translation": "Unable to update the channel."
+  },
+  {
+    "id": "app.channel.update_channel.internal_error",
+    "translation": "Unable to update channel."
   },
   {
     "id": "app.cluster.404.app_error",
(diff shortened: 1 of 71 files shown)


rvillablanca:
I need to order the message keys in en.json and other languages alphabetically but I think this is already automated by some `make` task

agnivade:
> I need to order the message keys in en.json and other languages alphabetically but I think this is already automated by some make task

Yes, you need to use `make i18n-extract`.

rvillablanca:
Thanks :)

agnivade:
@rvillablanca - A lot of translation keys from en.json seem to be removed. I'm wondering why. Could you merge with latest master, and run the i18n-extract again ?

This is failing the CI.

rvillablanca:
Yep, they were removed after running `i18n-extract`

rvillablanca:
> @rvillablanca - A lot of translation keys from en.json seem to be removed. I'm wondering why. Could you merge with latest master, and run the i18n-extract again ?
> 
> This is failing the CI.

Ok

rvillablanca:
The branch is already up to date

rvillablanca:
@agnivade What if we restore to commit before doing `i18n-extract`? I noticed after running the command that several keys were removed but I thought it was like a purge of unused keys.

If I restore, I'll have to order the keys alphabetically manually ?

agnivade:
> If I restore, I'll have to order the keys alphabetically manually ?

If you don't use the command, then yes you'll have to do it manually. The keys are there, it's just not in this repo, but in our enterprise repo. But it's not missing all the enterprise keys, which is odd. I can take a look into it after the review is done.

i18n/es.json
```diff
@@ -6511,19 +6511,19 @@
     "translation": "No se puede actualizar la fecha de la última actualización del icono del equipo"
   },
   {
-    "id": "store.sql_terms_of_service.save.app_error",
+    "id": "app.terms_of_service.create.app_error",
```
jespino, line 6514: The expectation here is to only update the `en.json` file

i18n/en.json
```diff
@@ -3827,8 +3819,20 @@
     "translation": "Unable to rename the team, the name is already in use."
   },
   {
-    "id": "app.user.complete_switch_with_oauth.blank_email.app_error",
-    "translation": "Unable to complete SAML login with an empty email address."
+    "id": "app.terms_of_service.create.app_error",
```
jespino, line 3822: Because you have autogenerated and you don't have the enterprise repository you removed a lot of valid enterprise strings. You have to edit it by hand, or let us regenerate it for you.

agnivade: I'm wondering whether this should be considered a bug ? This essentially means a contributor cannot run `i18n-extract`. We should fix that to selectively check either just TE repo or both TE and EE.

jespino: Yes, is a bug in the design, probably we can just ensure that an `ent` prefix is used for enterprise strings (with the same i18n-extract tool) and then, make the TE generation aware of no `ent` strings must be altered.

agnivade: Created https://mattermost.atlassian.net/browse/MM-25619

i18n/es.json
rvillablanca: Ok @jespino I will remove all changes in other language files but, what will happen with the thoses file containing outdated keys ? Because here I'm not adding new ones, but modifying.

agnivade:
@lindy65 - All testing related to terms of service is required here. Not sure if there are any though. :confused:

PR #14679, opened 2020-05-28
[MM-25607] Correctly log if plugin icon is invalid

hanzei:
#### Summary
TIL: `errors.Wrapf(nil, "something")` is `nil`. https://github.com/pkg/errors/blob/614d223910a179a466c1767a985424175c39b465/errors.go#L202-L204

#### Ticket Link
Fixes https://mattermost.atlassian.net/browse/MM-25607


prapti:
@hanzei Assuming the error should be printed in the server logs, I don't see the error (`icon is not svg %s`) from server when the icon file in github's asset folder is modified as follows:

1. icon.svg is replaced by a text file "icon.txt"
2. icon.svg is force-renamed "icon.txt"
3. icon.svg is replaced by a text file force-named "icon.svg"

Wondering if I'm looking for the error in the right place or I need a specific github PR version too to accompany the changes in this PR?

hanzei:
@prapti The QA test steps I provided where unclear. The error should only be logged when the Marketplace gets opened. I update the ticket with better testing steps.

Issue #14493, opened 2020-05-06
Addd arm64 management for plugins


protheusfr:
Mattermost work well on arm / arm64 (thanks to Go), but they are artificially locked to adm64 arch for plugin management.

Is it possible to update the following code on `model/manifest.go:193 `:
```
type ManifestExecutables struct {
	// LinuxAmd64 is the path to your executable binary for the corresponding platform
	LinuxAmd64 string `json:"linux-amd64,omitempty" yaml:"linux-amd64,omitempty"`
	// DarwinAmd64 is the path to your executable binary for the corresponding platform
	DarwinAmd64 string `json:"darwin-amd64,omitempty" yaml:"darwin-amd64,omitempty"`
	// WindowsAmd64 is the path to your executable binary for the corresponding platform
	// This file must have a ".exe" extension
	WindowsAmd64 string `json:"windows-amd64,omitempty" yaml:"windows-amd64,omitempty"`
}
```

By 

```
type ManifestExecutables struct {
	// LinuxAmd64 is the path to your executable binary for the corresponding platform
	LinuxAmd64 string `json:"linux-amd64,omitempty" yaml:"linux-amd64,omitempty"`
	// DarwinAmd64 is the path to your executable binary for the corresponding platform
	DarwinAmd64 string `json:"darwin-amd64,omitempty" yaml:"darwin-amd64,omitempty"`
	// WindowsAmd64 is the path to your executable binary for the corresponding platform
	// This file must have a ".exe" extension
	WindowsAmd64 string `json:"windows-amd64,omitempty" yaml:"windows-amd64,omitempty"`
	// LinuxArm64 is the path to your executable binary for the corresponding platform
	LinuxArm64 string `json:"linux-arm64,omitempty" yaml:"linux-arm64,omitempty"`
}
```

 ?

amyblais:
@protheusfr I can ask our team about this. What is the reason behind the request?

protheusfr:
Hello @amyblais ,

We use Mattersmost in an ARM64 server and would like to use some existing plugins.
Most of them can be easily recompiled for ARM64 but the artificial limitation in `model/manifest.go` (see my initial message) avoid to achieve that.

harshilsharma63:
@protheusfr just out of curiosity, are you able to run Mattermost fine on ARM64 machine?

protheusfr:
Yes, without any issue, in version 5.22.0.

harshilsharma63:
[...]

PR #14682, opened 2020-05-28
[GH-14493] Accept any GOOS and GOARCH combination in manifest

yukiisbored:
#### Summary

This pull request allows GOOS and GOARCH combination for plugin manifests by replacing `ManifestExecutables` which hardcodes GOOS/GOARCH combinations to a generic map of string with the GOOS/GOARCH combination as key.  It preserves compatibility with the already established plugin manifest format.

#### Ticket Link
Closes https://github.com/mattermost/mattermost-server/issues/14493

#### Release Note

```release-note
Enable experimental support for arm64 plugins by allowing any matching GOOS-GOOARCH combination in the plugin manifest.
```


model/manifest.go
```diff
@@ -288,12 +278,11 @@ func (m *Manifest) GetExecutableForRuntime(goOs, goArch string) string {
...
-			executable = server.Executables.LinuxAmd64
-		} else if goOs == "darwin" && goArch == "amd64" {
-			executable = server.Executables.DarwinAmd64
-		} else if goOs == "windows" && goArch == "amd64" {
-			executable = server.Executables.WindowsAmd64
+		comb := fmt.Sprintf("%s-%s", goOs, goArch)
+		val, exist := server.Executables[comb]
+
+		if exist {
+			executable = val
```
hanzei, line 285: What is the UX for an admin like if no executable was found? Does the error message make that clear?

lieut-data: Tracing the code to the supervisor, it looks like the message might be pretty confusing. I'd love to see an explicit call to `GetExecutableForRuntime` inside `newSupervisor` that returns an explicit error when it fails to find an executable for the current architecture.

model/manifest.go
```diff
@@ -198,7 +198,7 @@ type Manifest struct {
 type ManifestServer struct {
 	// Executables are the paths to your executable binaries, specifying multiple entry points
 	// for different platforms when bundled together in a single plugin.
-	Executables *ManifestExecutables `json:"executables,omitempty" yaml:"executables,omitempty"`
+	Executables map[string]string `json:"executables,omitempty" yaml:"executables,omitempty"`
```
lieut-data, line 201: Ah, this is something I did not anticipate: we apply SemVer to our `model` package, and this is unfortunately a backwards incompatible change to the signature of this model object.

To preserve SemVer, could we customize the deserialization to both populate `Executables` and a new, private field (`otherExecutables map[string]string`?) that is used within `GetExecutableForRuntime` to ultimately achieve the same effect?

yukiisbored: It appears that Go doesn't have a mechanism to handle "unknown fields" via its `Encoder` nor `Unmarshal` for both `encoding/json` and `gopkg.in/yaml.v2`. We could add special logic inside `FindManifest` which will handle this specific case but I'm not sure if that's a good idea and it feels really hacky.

To retain values of "unknown fields", the simplest way is to unmarshal/decode it into a generic `map[string]interface` and take the values from there and apply it to the created struct variable.

Another issue with this approach is when you want to encode/marshal it into a json/yaml file. `encoding/json` and `gopkg.in/yaml.v2` doesn't have anything to manipulate json/yaml data directly and would mean a special function to just handle encoding for that part.

model/manifest.go
yukiisbored: That would mean changing the function return type from `string` to `(string, error)` which also mean breaking semver :/

model/manifest.go
teresa-novoa: Thanks Muhammad! @lieut-data  can you take a look? thanks!

model/manifest.go
lieut-data: In this case, I think having `GetExecutableForRuntime` in the `model` package was probably unnecessary to begin with. It's not really "shared" code in the sense that we intend anyone other than the server's `plugin` package to execute same. I'd propose copying and implementing a private `getExecutableForRuntime` within the `plugin` package that has the desired return semantics.

model/manifest.go
lieut-data: Thanks @yukiisbored! Preserving SemVer is a hard requirement until the next major version bump. Could we define an unexported `manifestServer` of this `struct` with the desired `map[string]string`, define `UnmarshalJSON` on `ManifestServer` to unmarshal to an instance of `manifestServer` and then copy over the fields in question, including to the new `otherExecutables`?

yukiisbored: Due to how [`reflect`](https://golang.org/pkg/reflect/) works in Go, `encoding/json` can only work if the field is exported. It won't handle unexported fields: https://play.golang.org/p/uoXRQFwmesY

Even if we decide to export that field, it would result in another issue: You cannot have multiple fields with the same json tag. Doing so it would result in both fields being ignored: https://play.golang.org/p/5KPU9OuDsP1

I also tried experimenting with embedding structs which might allow us to have two versions of `Manifest` or `ManifestServer` but it ended up with the same roadblock as before: https://play.golang.org/p/ofeuXSTvdzH

 I'm not sure if there's a way to do this without importing another library just to handle this specific case or have a separate copy of `Manifest` which uses the "new" `ManifestServer` structure. 

Another point to consider is the fact that I don't see `ManifestServer` being used directly anywhere outside of `manifest.go`, so I'm not sure if introducing a change which would break how plugin manifest data is handled to be worth it.

yukiisbored: On the "importing another library" case, we could import a library like [jsonparser](https://github.com/buger/jsonparser) or [fastjson](https://github.com/valyala/fastjson) which allows parsing JSON without a schema.

With it we can go to the original unexported fields route, I'm just unsure whether it's a good idea to import another library just for this case and how the call to `GetExecutableForRuntime` might look like since the original JSON data is required for this to work, might need to see how the manifest itself is loaded for this very specific case.

lieut-data:
@yukiisbored, I've pushed a change that attempts to preserve SemVer while still introducing a new `AllExecutables` `map[string]string` to support any `GOOS` and `GOARCH` combination. Let me know your thoughts and if this still works for you?

lieut-data:
(Removing myself as a reviewer since I'm contributing code.)

model/manifest.go
```diff
@@ -196,9 +196,21 @@ type Manifest struct {
...
+	// Executables is a legacy field populated with a subset of supported platform executables.
+	// When unmarshalling, Executables is authoritative for the platform executable paths it
+	// contains, overriding any values in AllExecutables. When marshalling, AllExecutables
+	// is authoritative.
+	//
+	// Code duplication is avoided when (un)marshalling by leveraging type aliases in the
+	// various (Un)Marshal(JSON|YAML) methods, since aliases don't inherit the aliased type's
+	// methods.
+	//
+	// In v6.0, we should remove this field and rename AllExecutables back to Executables.
```
crspeller, line 212: Ticket for this?

jasonblais:
@hanzei friendly reminder to help with dev review, thanks!

model/manifest.go
lieut-data: https://mattermost.atlassian.net/browse/MM-27593

prapti:
@hanzei @crspeller Does this PR need a general sanity test around plugins? I'm not very clear on what I'm testing.

crspeller:
@prapti Ya that sounds good. Run some plugins with these changes.

hanzei:
@crspeller Should we test the functionality added here i.e. run a plugin on previously not supported platform like `arm64`?

crspeller:
@hanzei Ya, that would be great. Go ahead and do that.

lieut-data:
@prapti, can you help identify any blockers to merging this PR?

prapti (approved):
Apologies about the long wait on this PR. The main issue was finding a proper GOARCH test setup to run a full set of tests against this PR and verify that nothing is unacceptably broken. I've run a set of smoke tests locally on my dev box however, to make sure that at least what was working before still works as expected, and so far everything looks good, hence approving this PR.
Thank you for the patience everyone!
cc @lieut-data

lieut-data:
Thanks, @prapti! I've kicked off ee tests, and will merge this once they pass. The changes should be included in v5.31.

amyblais:
> I've kicked off ee tests, and will merge this once they pass. The changes should be included in v5.31.

@lieut-data Do you have more context for whether this is still needed for v5.31 or can be pushed to v5.32?

lieut-data:
@amyblais, it should be fine to just merge this to master. While Apple Silicon makes this change more interesting, it's still primarily for developers at this point.

PR #14684, opened 2020-05-28
Add Jitsi plugin to diagnostics

jasonblais:
Jitsi plugin: https://github.com/mattermost/mattermost-plugin-jitsi

@amyblais wondering if we can add to 5.24? This adds one telemetry, low risk.

cc @aaronrothschild FYI

### DIFF
--- a/app/diagnostics.go
+++ b/app/diagnostics.go
@@ -674,6 +674,7 @@ func (a *App) trackConfig() {
 		"enable_gitlab":                 pluginActivated(cfg.PluginSettings.PluginStates, "com.github.manland.mattermost-plugin-gitlab"),
 		"enable_jenkins":                pluginActivated(cfg.PluginSettings.PluginStates, "jenkins"),
 		"enable_jira":                   pluginActivated(cfg.PluginSettings.PluginStates, "jira"),
+		"enable_jitsi":                  pluginActivated(cfg.PluginSettings.PluginStates, "jitsi"),
 		"enable_nps":                    pluginActivated(cfg.PluginSettings.PluginStates, "com.mattermost.nps"),
 		"enable_webex":                  pluginActivated(cfg.PluginSettings.PluginStates, "com.mattermost.webex"),
 		"enable_welcome_bot":            pluginActivated(cfg.PluginSettings.PluginStates, "com.mattermost.welcomebot"),


app/diagnostics.go
```diff
@@ -674,6 +674,7 @@ func (a *App) trackConfig() {
 		"enable_gitlab":                 pluginActivated(cfg.PluginSettings.PluginStates, "com.github.manland.mattermost-plugin-gitlab"),
 		"enable_jenkins":                pluginActivated(cfg.PluginSettings.PluginStates, "jenkins"),
 		"enable_jira":                   pluginActivated(cfg.PluginSettings.PluginStates, "jira"),
+		"enable_jitsi":                  pluginActivated(cfg.PluginSettings.PluginStates, "jitsi"),
```
hanzei, line 677: Not blocking for this PR, but I'm wondering if `jitsi` might not be a good choice for a plugin id. It's encouraged to use a unique id like `com.mattermost.jitsi`. If there is any good time to change a plugin id, then it's right before a plugin goes to the Marketplace.

hanzei: cc @jespino :point_up:

jespino: I'm not 100% sure about this, if we want to go down that path, probably we should track both here, `jitsi` and `com.mattermost.jitsi`. What is the policy applied to other plugins like jira, github, or zoom?

hanzei: Other plugins won't get there `id` changed because changing the `id` remove exciting configuration and that is not acceptable for Marketplace plugins.

jespino: There are already people out there using the old plugin and using the old config, there is not too much to config in the 1.3, version of the plugin, but still there have information like the JWT APP ID and JWT Secret.

jasonblais: @hanzei I see the benefit of using a unique ID, but there are likely many systems using the plugin already (the Jitsi plugin has the most GitHub stars of any plugins in the Mattermost org). So I'd prefer not to cause a breaking change for them.

It won't have a big impact on metrics, if someone uploads a custom-built plugin with id `jitsi`.

If you feel strongly about this, it can be considered. But I'm 3/5 not to make the change.

PR #14687, opened 2020-05-28
Migrate Update method from ChannelStore to return idiomatic plain error

rvillablanca:
#### Summary

This is a proposal PR that migrates the method `ChannelStore.Update` to return idiomatic go error interface (when it applies).

Fixes https://github.com/mattermost/mattermost-server/issues/14689

### DIFF
--- a/store/sqlstore/channel_store.go
+++ b/store/sqlstore/channel_store.go
@@ -631,10 +631,10 @@ func (s SqlChannelStore) saveChannelT(transaction *gorp.Transaction, channel *mo
 }
 
 // Update writes the updated channel to the database.
-func (s SqlChannelStore) Update(channel *model.Channel) (*model.Channel, *model.AppError) {
+func (s SqlChannelStore) Update(channel *model.Channel) (*model.Channel, error) {
 	transaction, err := s.GetMaster().Begin()
 	if err != nil {
-		return nil, model.NewAppError("SqlChannelStore.Update", "store.sql_channel.update.open_transaction.app_error", nil, err.Error(), http.StatusInternalServerError)
+		return nil, errors.Wrap(err, "begin_transaction")
 	}
 	defer finalizeTransaction(transaction)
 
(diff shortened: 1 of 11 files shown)


rvillablanca:
Hey maintainers @jespino @agnivade this is another PR that helps in the campaign to migrate the all stores to return plain error interface when it is required, replacing `*model.AppError`.

I pretend to help you in this campaing by migrating litle by litle some methods of big stores like `ChannelStore` or completely small ones like `ChannelMemberHistoryStore` as long as time allows me. I know you have plans to migrate even app layer and service layer from `*model.AppError` according what Agniva said me, so this is a big effort where I can help.

In the past I helped with the migration of operations from an async approach to sync by default, and my feeling was that I really could help and I was a contribution in that campaing (together with Jesus) so this time I would like to help in this big campaign, but I need to know what is the best approach so we can quickly release several PR without taking so mucho of your time.

This is a small and fine grained PR and I think this is the best and fastest way you can approve the changes instead of big PR but I'm open to your suggestions.

agnivade:
@rvillablanca - This is exactly what I was thinking of doing for big stores. Thank you so much ! Keep'em coming.

If you plan to migrate entire stores at a time, let me know and I will create those tickets so that we can track them.

rvillablanca:
What about simple methods migration like this one, is it necessary to create issues ?

agnivade:
Yes, it helps us track the changes during QA. There's no manual step involved. We use a script to automatically create all of this. Just let me know what methods you are going to migrate and I will create tickets for those.

store/sqlstore/channel_store.go
```diff
@@ -631,10 +631,10 @@ func (s SqlChannelStore) saveChannelT(transaction *gorp.Transaction, channel *mo
...
 
 // Update writes the updated channel to the database.
-func (s SqlChannelStore) Update(channel *model.Channel) (*model.Channel, *model.AppError) {
+func (s SqlChannelStore) Update(channel *model.Channel) (*model.Channel, error) {
 	transaction, err := s.GetMaster().Begin()
 	if err != nil {
-		return nil, model.NewAppError("SqlChannelStore.Update", "store.sql_channel.update.open_transaction.app_error", nil, err.Error(), http.StatusInternalServerError)
+		return nil, errors.Wrap(err, "begin_transaction")
 	}
 	defer finalizeTransaction(transaction)
```
agnivade, line 640: Let's change `updateChannelT` also to return `error` ? Only the 
```
if err := channel.IsValid(); err != nil {
	return nil, err
}
```
part would still return an AppError as is. But we can convert the rest of them to the proper errors.

store/sqlstore/channel_store.go
```diff
@@ -645,11 +645,11 @@ func (s SqlChannelStore) Update(channel *model.Channel) (*model.Channel, *model.
 
 	// Additionally propagate the write to the PublicChannels table.
 	if err := s.upsertPublicChannelT(transaction, updatedChannel); err != nil {
-		return nil, model.NewAppError("SqlChannelStore.Update", "store.sql_channel.update.upsert_public_channel.app_error", nil, err.Error(), http.StatusInternalServerError)
+		return nil, errors.Wrap(err, "failed to update public channels")
```
agnivade, line 648: Let's add a prefix of "upsertPublicChannelT: " in the error message. Also, it would be "failed to upsert .. "

store/sqlstore/channel_store.go
rvillablanca: Ok this is done

store/sqlstore/channel_store.go
rvillablanca: Done

agnivade:
@lindy65 - Testing channel update scenarios are sufficient here.

PR #14688, opened 2020-05-29
Migrate Get/GetFromMaster methods from ChannelStore to return error interface

rvillablanca:
#### Summary

This is a proposal PR that migrates the method `ChannelStore.Get` and `ChannelStore.GetFromMaster` to return idiomatic go error interface (when it applies).

Fixes https://github.com/mattermost/mattermost-server/issues/14690

### DIFF
--- a/store/store.go
+++ b/store/store.go
@@ -13,9 +13,9 @@ import (
 )
 
 type StoreResult struct {
-	Data   interface{}
-	Err    *model.AppError
-	NewErr error
+	Data interface{}
+	Err  *model.AppError
+	NErr error
 }
 
 type Store interface {
--- a/store/store.go
+++ b/store/store.go
@@ -15,6 +15,8 @@ import (
 type StoreResult struct {
 	Data interface{}
 	Err  *model.AppError
+
+	// NErr a temporary field used by the new code for the AppError migration. This will later become Err when the entire store is migrated.
 	NErr error
 }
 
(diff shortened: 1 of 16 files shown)


rvillablanca:
**Note for reviewers**: Since we are migrating from `*model.AppErro` to `error` it was necessary to add a new Field in the `StoreResult` struct, so we can write to and check from it for this new errors when using this migrations in goroutines.

rvillablanca:
I have to rename the message keys in `en.json` so let's keep that suggestiont out of the review :joy:

store/store.go
```diff
@@ -15,6 +15,7 @@ import (
 type StoreResult struct {
 	Data interface{}
 	Err  *model.AppError
+	NErr error
```
agnivade, line 18: Let's add a comment here to clarify the purpose of this field.

```
NErr error
```

rvillablanca: No problem! :+1:

jespino:
@rvillablanca there are some tests failing, probably related to the type conversions and maybe the absense of renaming properly the error variable, can you take a look?

rvillablanca:
Ready, tests passing @jespino

lindy65:
Hi @agnivade - can you help with test steps for this PR please? Thanks!

agnivade:
They should mainly focus on all actions that fetch channels.

PR #14691, opened 2020-05-29
MM-25115 - make cacheProvider2 accessible to localcachelayer

liusy182:
#### Summary

This PR is a continuation from [#14496](https://github.com/mattermost/mattermost-server/pull/14496). It makes the new `cacheProvider2` accessible to the `localcachelayer.NewLocalCacheLayer` method. 

`NewLocalCacheLayer(...)` creates multiple cache storages underneath. Future PRs will migrate the usage from `cacheProvider` to `cacheProvider2`

This PR partially overlaps with #14667 , but do not depend on each other. Both PRs modify the same file `app/server.go`.

#### Ticket Link

https://mattermost.atlassian.net/browse/MM-25115


liusy182:
WIP. Will incorporate the rest of the migration work into this PR.

store/storetest/user_store.go
```diff
@@ -1056,9 +1056,12 @@ func testUserStoreGetAllProfilesInChannel(t *testing.T, ss store.Store) {
...
 		require.Nil(t, err)
-		assert.Equal(t, map[string]*model.User{
-			u1.Id: sanitized(u1),
-		}, profiles)
+		require.Equal(t, 1, len(profiles))
+		profile, ok := profiles[u1.Id]
+		require.True(t, ok)
+		require.True(t, profile.AuthData == nil || *profile.AuthData == "")
+		profile.AuthData = model.NewString("")
+		require.Equal(t, sanitized(u1), profile)
```
liusy182, line 1064: This block of change is to workaround [an issue in gob deserialization](https://stackoverflow.com/questions/42675054/golang-gob-converts-pointer-to-0-into-nil-pointer) as it does not distinguish between `nil` and `""` for data type `*string` unfortunately. 

I have scanned through the code base and it looks like `AuthData == nil` and `AuthData == ""` represent the same meaning throughout the code base.

liusy182:
ready for review.

agnivade:
Amazing work @liusy182. I will get to reviewing it soon.

agnivade:
As expected, there are some regressions in performance.

Across all 3 metrics - CPU, mem, API response.

cpu

mem

api

There is a slight dip in the middle of the first test because the API time exceeded the threshold and therefore the coordinator had to reduce the number of users. In master, everything happens smoothly.

The difference in API times is not very clear due to the big spikes during logouts. But basically, with the new cache, it mostly hovers around the limit of 200ms. And in master it's 9ms.

/cc @grundleborg - Are you okay with pushing this to daily, or would you prefer to have the performance fix go in first ? Both are independant of each other as they operate in different layers.

agnivade:
That's a good question @liusy182, and as I mentioned earlier, I don't have a direct answer to that. I think we would want to have a good cost-benefit ratio. i.e. try to have as much performance possible without investing too much effort into it.

With that said, I already have sent out the PR to improve performance. It gives us a decent improvement without requiring us to do a lot of code changes. Optimizing further would require spending more time and effort. So once that goes in, I am happy to merge this.

Please also take a look at the conflicts.

Thanks.

liusy182:
Cool thanks I see you had #14826 :)
I will resolve conflicts after this PR is merged as they both touch the same files.

agnivade:
@liusy182 - That PR is now merged. You are welcome to fix the conflicts. Thanks!

store/storetest/user_store.go
agnivade: Wondering if this is still needed?

agnivade:
@liusy182 - left a quick comment [here](https://github.com/mattermost/mattermost-server/pull/14691#discussion_r443096153).

store/storetest/user_store.go
liusy182: just tried we dont. i am reverting this back.

agnivade:
I am happy to announce that the new performance optimizations have made the difference to be negligible at a high level. 

Here are the results:

avg-store-times
cpu-utilization
heap-in-use
rps
stack-in-use


@grundleborg - The above is a quick taste of what you would get when comparing 2 load tests with the new tool.

liusy182:
Cool. I am really glad to see the result of the performance improvement you did. @agnivade

agnivade:
@lindy65 - This is a very important and potentially breaking change. We should test this in an HA setup and do a full smoke test of all features by logging in from multiple users and testing for data consistency among them.

For example, things like user statuses, emoji reaction count, channel unreads between different users. If a user marks a channel as unread, and then logs out and logs in again, it should still remain unread. Things like that.

lindy65:
Thanks for the testing info @agnivade - I'm going to try to spin up an HA spinwick but have had trouble doing so over the past few days.  Hopefully it'll work on this PR!

PR #14692, opened 2020-05-29
[MM-25606] Make plugin API errors encodable

iomodo:
#### Summary
This PR will make errors returned by plugin API encodable

#### Ticket Link
Fixes - https://mattermost.atlassian.net/browse/MM-25606


plugin/interface_generator/main.go
```diff
@@ -310,6 +310,7 @@ func (s *apiRPCServer) {{.Name}}(args *{{.Name | obscure}}Args, returns *{{.Name
 		{{.Name}}{{funcStyle .Params}} {{funcStyle .Return}}
 	}); ok {
 		{{if .Return}}{{destruct "returns." .Return}} = {{end}}hook.{{.Name}}({{destruct "args." .Params}})
+		{{if .Return}}{{encodeErrors "returns." .Return}}{{end}}
```
hanzei, line 313: A little request for better readable code: Using the tool from https://golang.org/pkg/text/template/#hdr-Text_and_spaces It's possible to omit the empty line and reduce the `diff`.

iomodo:
@hanzei kind reminder

amyblais:
Moved the ticket to Queued for now as this is not a regression and we're past code complete,

iomodo:
@DHaussermann here are the easiest steps:
1. Create plugin with non-valid autocomplete data(for example with `trigger` been empty string "").
2. Upload and enable plugin
3. Observe an error in the logs stating that autocomplete data was not valid.

DHaussermann:
@iomodo I'd like to make this change locally from the PR on this branch https://github.com/mattermost/mattermost-plugin-agenda/tree/add-autocomplete to make a custom version of a Agenda to test this with. But I'm not sure where the maker the change for autocomplete `trigger` to be blank. 

Can you share the file path where this change would be needed.

hanzei:
@iomodo Looks like you have to do this manually.

PR #14698, opened 2020-05-29
[MM-25268] Implement prefix/suffix searching on teams and channel page in system console

hahmadia:
#### Summary

Implements prefix/suffix searching for team and channel page in the system console

#### Ticket Link

https://mattermost.atlassian.net/browse/MM-25268

**Screenshots**
image

image

image


store/sqlstore/team_store.go
```diff
@@ -364,7 +364,7 @@ func (s SqlTeamStore) SearchAll(term string) ([]*model.Team, *model.AppError) {
 
 	term = sanitizeSearchTerm(term, "\\")
 
-	if _, err := s.GetReplica().Select(&teams, "SELECT * FROM Teams WHERE Name LIKE :Term OR DisplayName LIKE :Term", map[string]interface{}{"Term": term + "%"}); err != nil {
+	if _, err := s.GetReplica().Select(&teams, "SELECT * FROM Teams WHERE Name LIKE :Term OR DisplayName LIKE :Term", map[string]interface{}{"Term": "%" + term + "%"}); err != nil {
```
sbishel, line 367: I would build this in a small function. ie. wildcardSearchTerm(term)

mkraft:
@hahmadia Are these queries case insensitive? The queries don't look it and I wouldn't think functionally we want them to be case sensitive.

store/sqlstore/team_store.go
hahmadia: Good call. I have turned that into a function and added tests for both channel and teams.

hahmadia:
> @hahmadia Are these queries case insensitive? The queries don't look it and I wouldn't think functionally we want them to be case sensitive.

I added a strings.ToLower(%term%) and added tests on both channels and teams to ensure case insensitive. Thanks for the point out!

api4/channel_test.go
```diff
@@ -1150,11 +1150,53 @@ func TestSearchAllChannels(t *testing.T) {
...
+
+	search = &model.ChannelSearch{Term: strings.ToUpper(th.BasicChannel.Name)}
+
+	channels, resp = th.SystemAdminClient.SearchAllChannels(search)
+	CheckNoError(t, resp)
+
+	assert.Len(t, *channels, 1)
+	assert.Equal(t, th.BasicChannel.Id, (*channels)[0].Id)
+
+	search = &model.ChannelSearch{Term: th.BasicChannel.Name[0:2] + strings.ToUpper(th.BasicChannel.Name[2:5]) + th.BasicChannel.Name[5:]}
```
sbishel, line 1187: You may want to throw in a comment here, just mentioning what you're testing. 
// Test Mixed Case.

api4/channel_test.go
```diff
@@ -1150,11 +1150,53 @@ func TestSearchAllChannels(t *testing.T) {
...
+	CheckNoError(t, resp)
+
+	assert.Len(t, *channels, 1)
+	assert.Equal(t, th.BasicChannel.Id, (*channels)[0].Id)
+
+	search = &model.ChannelSearch{Term: th.BasicChannel.Name}
+
+	channels, resp = th.SystemAdminClient.SearchAllChannels(search)
+	CheckNoError(t, resp)
+
```
sbishel, line 1199: Add a couple more tests for thoroughness.
(ie, starts with, ends with, search for display name.)
 "what", "ever"

api4/team_test.go
```diff
@@ -1220,18 +1220,50 @@ func TestSearchAllTeamsPaged(t *testing.T) {
...
 		Search             *model.TeamSearch
 		ExpectedTeams      []string
 		ExpectedTotalCount int64
 	}{
+		{
+			Name:               "Get foobar channel",
+			Search:             &model.TeamSearch{Term: "oob", Page: model.NewInt(0), PerPage: model.NewInt(100)},
+			ExpectedTeams:      []string{foobarTeam.Id},
+			ExpectedTotalCount: 1,
+		},
```
sbishel, line 1242: Same here, add search for "what" and "ever".

hahmadia:
> I would just a couple more tests for completeness.
> 
> Also, this may be handled in the web app. But do we only search after 2 or 3 characters are entered? Do we search on a single character?

We search whenever the user presses `Enter` on their keyboard.

store/sqlstore/team_store.go
```diff
@@ -374,8 +378,14 @@ func (s SqlTeamStore) SearchAll(term string) ([]*model.Team, *model.AppError) {
 	var teams []*model.Team
 
 	term = sanitizeSearchTerm(term, "\\")
+	term = wildcardSearchTerm(term)
 
-	if _, err := s.GetReplica().Select(&teams, "SELECT * FROM Teams WHERE Name LIKE :Term OR DisplayName LIKE :Term", map[string]interface{}{"Term": term + "%"}); err != nil {
+	operatorKeyword := "ILIKE"
```
catalintomai, line 383: I'd rather that we do not use Postgres (or any SQL DB provider) SQL extensions and stick as much as possible to vanilla SQL, otherwise porting the SQL code to any other provider (e.g. MSSQL) becomes an even more difficult task. Also, we should use/factor out `buildLIKEClause` and `buildFulltextClause` for both channels and teams

api4/team_test.go
```diff
@@ -1220,18 +1220,74 @@ func TestSearchAllTeamsPaged(t *testing.T) {
...
+	require.Nil(t, err)
+
 	testCases := []struct {
 		Name               string
 		Search             *model.TeamSearch
 		ExpectedTeams      []string
 		ExpectedTotalCount int64
 	}{
+		{
+			Name:               "Get foobar channel",
```
catalintomai, line 1238: is it "team" instead of "channel"?

store/sqlstore/team_store.go
```diff
@@ -374,8 +378,14 @@ func (s SqlTeamStore) SearchAll(term string) ([]*model.Team, *model.AppError) {
...
+	if s.DriverName() == model.DATABASE_DRIVER_MYSQL {
+		operatorKeyword = "LIKE"
+	}
+	queryString := fmt.Sprintf("SELECT * FROM Teams WHERE Name %[1]s :Term OR DisplayName %[1]s :Term", operatorKeyword)
```
agnivade, line 387: This pattern seems to be repeated throughout the PR. If possible, I'd prefer it to be refactored out to a separate place.

And also, construct the query using Squirrel if possible. Makes for much more readable SQL queries.

api4/team_test.go
hahmadia: Ticket made https://mattermost.atlassian.net/browse/MM-26720

store/sqlstore/team_store.go
hahmadia: Ticket created to keep track https://mattermost.atlassian.net/browse/MM-26726

store/sqlstore/team_store.go
hahmadia: Ticket created to keep track https://mattermost.atlassian.net/browse/MM-26727

PR #14699, opened 2020-05-29
MM-25595 - Edits to bot username and display name fail to save

cpoile:
#### Summary
- The bot was correctly updated in the `SqlBotStore` by the end of the `PatchBot` function. 
- The user profile for the bot was correctly updated in the `SqlUserStore` by the end of the `PatchBot` function. 
- We then call `UpdateUserRoles` which loads the cached version of the bot. It then overwrites with that cached version. And that's how the updated bot gets reverted to the cached bot.

#### Ticket Link
- https://mattermost.atlassian.net/browse/MM-25595

### DIFF
--- a/app/user.go
+++ b/app/user.go
@@ -1375,8 +1375,6 @@ func (a *App) DeleteToken(token *model.Token) *model.AppError {
 }
 
 func (a *App) UpdateUserRoles(userId string, newRoles string, sendWebSocketEvent bool) (*model.User, *model.AppError) {
-	a.InvalidateCacheForUser(userId)
-
 	user, err := a.GetUser(userId)
 	if err != nil {
 		err.StatusCode = http.StatusBadRequest
--- a/app/user.go
+++ b/app/user.go
@@ -1413,6 +1413,7 @@ func (a *App) UpdateUserRoles(userId string, newRoles string, sendWebSocketEvent
 		mlog.Error("Failed during updating user roles", mlog.Err(result.Err))
 	}
 
+	a.InvalidateCacheForUser(userId)
 	a.ClearSessionCacheForUser(user.Id)
 
 	if sendWebSocketEvent {
--- a/app/user.go
+++ b/app/user.go
@@ -1397,8 +1397,8 @@ func (a *App) UpdateUserRoles(userId string, newRoles string, sendWebSocketEvent
 
 	schan := make(chan store.StoreResult, 1)
 	go func() {
-		userId, err := a.Srv().Store.Session().UpdateRoles(user.Id, newRoles)
-		schan <- store.StoreResult{Data: userId, Err: err}
+		id, err := a.Srv().Store.Session().UpdateRoles(user.Id, newRoles)
+		schan <- store.StoreResult{Data: id, Err: err}
 		close(schan)
 	}()
 
--- a/app/user.go
+++ b/app/user.go
@@ -1375,6 +1375,8 @@ func (a *App) DeleteToken(token *model.Token) *model.AppError {
 }
 
 func (a *App) UpdateUserRoles(userId string, newRoles string, sendWebSocketEvent bool) (*model.User, *model.AppError) {
+	a.InvalidateCacheForUser(userId)
+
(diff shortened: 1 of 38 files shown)


cpoile:
@agnivade thank you for the long comment -- that definitely helps!

Reverting #13838 does fix the problem. What do you think next steps should be for this, then? 

But, it might be better if you took charge of this, if you're willing? I'm wary of 'proxy coding' (making a PR that is essentially your code, but myself not fully understanding why).

agnivade:
I think it's fine if we just use the `InvalidateCacheForUser` method for now. I will raise it with our team on how to improve this situation.

Adding a test should be simple enough. We might need to do it from the api layer rather than the app layer to be able to trigger this. I think it should be something like create_bot -> patch_bot -> get_user ? We just need to look into the series of API calls in the browser which leads to the failure, and recreate them in a unit test.

cpoile:
@agnivade I found where the root problem was -- when updating a bot, we also update the bot's roles immediately after updating the bot's user. The problem is `UpdateUserRoles` was getting the user (the cached user), then saving the cached version. So, I'm now invalidating the cache for the user before getting the user.  I'm pretty sure this won't cause new problems, but let me know if it doesn't look right to you.

The test recreates the problem, but I added it to the existing patch bot test (or else I would have to cut and paste the entire test just for those added lines). Let me know if you would prefer to have a new test for those extra lines.

Thanks for the nudge in the right direction on this.

app/user.go
```diff
@@ -1375,6 +1375,8 @@ func (a *App) DeleteToken(token *model.Token) *model.AppError {
 }
 
 func (a *App) UpdateUserRoles(userId string, newRoles string, sendWebSocketEvent bool) (*model.User, *model.AppError) {
+	a.InvalidateCacheForUser(userId)
```
agnivade, line 1378: IMO, this does not look to be the right fix. IIUC, our objective is to invalidate the cache after `userUpdate, err := a.Srv().Store.User().Update(user, true)` is called in `PatchBot`. But with this change, what we are doing here is invalidating the cache every time we call `GetUser`. I don't think `GetUser` has an inherent problem in getting the cached value by itself. Otherwise we would not have cached the user at all.

I believe the right solution is to invalidate the cache _after_ updating the user in `PatchBot`. That is the right order because this is "supposed" to behave like a write-through cache (though it's not), so all writes must go through the cache first, and then the store. And that is how it is done in the rest of the codebase. Technically, there is still a bug here because there is a period of time after updating and before invalidating, the cache will return stale data, but this is by design.

I have tested this patch locally with your PR and tested that it passes:
```diff
diff --git a/app/bot.go b/app/bot.go
index 09f96d326..9b53a4e5d 100644
--- a/app/bot.go
+++ b/app/bot.go
@@ -89,6 +89,7 @@ func (a *App) PatchBot(botUserId string, botPatch *model.BotPatch) (*model.Bot,
        if err != nil {
                return nil, err
        }
+       a.InvalidateCacheForUser(user.Id)
 
        ruser := userUpdate.New
        a.sendUpdatedUserEvent(*ruser)
diff --git a/app/user.go b/app/user.go
index 0c6971239..7fe7ae111 100644
--- a/app/user.go
+++ b/app/user.go
@@ -1375,7 +1375,7 @@ func (a *App) DeleteToken(token *model.Token) *model.AppError {
 }
 
 func (a *App) UpdateUserRoles(userId string, newRoles string, sendWebSocketEvent bool) (*model.User, *model.AppError) {
-       a.InvalidateCacheForUser(userId)
 
        user, err := a.GetUser(userId)
        if err != nil {

```

Thank you so much for spending your time to dig into this ! And sorry for the multiple rounds of review on this.

cpoile: Ok, that makes sense. Thanks again Agniva!

amyblais:
@cpoile Does this need QA review?

cpoile:
@amyblais I'm 1/5 that it doesn't need it. But if you think it should: It only affects the edit bot screen in the integrations menu. It would be a quick test, and could follow the steps in the original ticket (from QA).

PR #14705, opened 2020-05-31
[MM-25677] Content-Type is optional

ethervoid:
#### Summary

mime.ParseMimeType returns and "no media type" error if the passed
string is empty.

Given that the Content-Type header is optional we shouldn't return error
in that case so we're fixing that allowing the users to call the webhook
without passing that header

#### Ticket Link

https://mattermost.atlassian.net/browse/MM-25677

### DIFF
--- a/web/webhook.go
+++ b/web/webhook.go
@@ -29,8 +29,10 @@ func incomingWebhook(c *Context, w http.ResponseWriter, r *http.Request) {
 
 	var err *model.AppError
 	incomingWebhookPayload := &model.IncomingWebhookRequest{}
-	mediaType, _, mimeErr := mime.ParseMediaType(r.Header.Get("Content-Type"))
-	if mimeErr != nil && mimeErr != mime.ErrInvalidMediaParameter {
+	contentType := r.Header.Get("Content-Type")
+	mediaType, _, mimeErr := mime.ParseMediaType(contentType)
+	// Having an empty content-type is fine given that is an optional header
+	if contentType != "" && mimeErr != nil && mimeErr != mime.ErrInvalidMediaParameter {
 		c.Err = model.NewAppError("incomingWebhook", "api.webhook.incoming.error", nil, mimeErr.Error(), http.StatusBadRequest)
 		return
 	}
--- a/web/webhook.go
+++ b/web/webhook.go
@@ -33,7 +33,12 @@ func incomingWebhook(c *Context, w http.ResponseWriter, r *http.Request) {
 	mediaType, _, mimeErr := mime.ParseMediaType(contentType)
 	// Having an empty content-type is fine given that is an optional header
 	if contentType != "" && mimeErr != nil && mimeErr != mime.ErrInvalidMediaParameter {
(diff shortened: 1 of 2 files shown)


web/webhook.go
```diff
@@ -29,9 +29,16 @@ func incomingWebhook(c *Context, w http.ResponseWriter, r *http.Request) {
 
 	var err *model.AppError
 	incomingWebhookPayload := &model.IncomingWebhookRequest{}
-	mediaType, _, mimeErr := mime.ParseMediaType(r.Header.Get("Content-Type"))
-	if mimeErr != nil && mimeErr != mime.ErrInvalidMediaParameter {
-		c.Err = model.NewAppError("incomingWebhook", "api.webhook.incoming.error", nil, mimeErr.Error(), http.StatusBadRequest)
+	contentType := r.Header.Get("Content-Type")
+	mediaType, _, mimeErr := mime.ParseMediaType(contentType)
+	// Having an empty content-type is fine given that is an optional header
+	if contentType != "" && mimeErr != nil && mimeErr != mime.ErrInvalidMediaParameter {
```
lieut-data, line 35: Can we clarify the logic here to make the `mediaType` parsing somewhat more explicit on the `contentType` not being empty?
```
var mediaType string
contentType := r.Header.Get("Content-Type")
if contentType != "" {
        mediaType, _, mimeErr := mime.ParseMediaType(contentType)
        if mimeErr != nil && mimeErr != mime.ErrInvalidMediaParameter {
                c.Err = ....
        }
}
```

web/webhook.go
```diff
@@ -29,9 +29,16 @@ func incomingWebhook(c *Context, w http.ResponseWriter, r *http.Request) {
...
+		c.Err = model.NewAppError("incomingWebhook",
+			"api.webhook.incoming.error",
+			nil,
+			"webhook_id="+id+" - error: "+mimeErr.Error(),
```
lieut-data, line 39: nit: can we join with commas vs. dashes?
```suggestion
 			"webhook_id="+id+", error: "+mimeErr.Error(), 
```

web/webhook.go
```diff
@@ -58,7 +65,12 @@ func incomingWebhook(c *Context, w http.ResponseWriter, r *http.Request) {
 		err := decoder.Decode(incomingWebhookPayload, r.PostForm)
 
 		if err != nil {
-			c.Err = model.NewAppError("incomingWebhook", "api.webhook.incoming.error", nil, err.Error(), http.StatusBadRequest)
+			c.Err = model.NewAppError("incomingWebhook",
+				"api.webhook.incoming.error",
+				nil,
+				"webhook_id="+id+" - error: "+err.Error(),
```
lieut-data, line 71:
```suggestion
 				"webhook_id="+id+", error: "+err.Error(), 
```

web/webhook.go
ethervoid: Yes, sure!

ethervoid:
@amyblais Is there a 5.23.1 branch to create the cherry-pick?

PR #14708, opened 2020-05-31
Migrate multiples methods from ChannelStore to return error interface

rvillablanca:
#### Summary

This PR migrates `GetByName`, `GetByNames`  and `GetByNameIncludeDeleted` methods.

Fixes https://github.com/mattermost/mattermost-server/issues/14718

Jira ticket: https://mattermost.atlassian.net/browse/MM-25693


rvillablanca:
@agnivade plese create this ticket so we can track it then :)

app/command_msg.go
```diff
@@ -9,6 +9,9 @@ import (
 	goi18n "github.com/mattermost/go-i18n/i18n"
 	"github.com/mattermost/mattermost-server/v5/mlog"
 	"github.com/mattermost/mattermost-server/v5/model"
+	"github.com/mattermost/mattermost-server/v5/store"
+
+	"github.com/pkg/errors"
```
agnivade, line 14: We can use the standard library for this.

store/sqlstore/channel_store.go
```diff
@@ -1193,7 +1193,10 @@ func (s SqlChannelStore) GetByNames(teamId string, names []string, allowFromCach
 
 		var dbChannels []*model.Channel
 		if _, err := s.GetReplica().Select(&dbChannels, query, props); err != nil && err != sql.ErrNoRows {
-			return nil, model.NewAppError("SqlChannelStore.GetByName", "store.sql_channel.get_by_name.existing.app_error", nil, "teamId="+teamId+", "+err.Error(), http.StatusInternalServerError)
+			if teamId != "" {
+				return nil, errors.Wrapf(err, "failed to get channels with teamId=%s and names=%v", teamId, names)
+			}
+			return nil, errors.Wrapf(err, "failed to get channels with names=%v", names)
```
agnivade, line 1199: Perhaps we can refactor this to:
```go
msg := fmt.Sprintf("failed to get channels with names=%v", names)
if teamId != "" {
  msg += fmt.Sprintf("teamId=%s", teamId)
}
return errors.Wrap(err, msg)
```

app/channel.go
```diff
@@ -77,7 +77,13 @@ func (a *App) JoinDefaultChannels(teamId string, user *model.User, shouldBeAdmin
 	for _, channelName := range a.DefaultChannelNames() {
 		channel, channelErr := a.Srv().Store.Channel().GetByName(teamId, channelName, true)
 		if channelErr != nil {
-			err = channelErr
+			var nfErr *store.ErrNotFound
+			switch {
+			case errors.As(err, &nfErr):
+				err = model.NewAppError("JoinDefaultChannels", store.MISSING_CHANNEL_ERROR, nil, nfErr.Error(), http.StatusNotFound)
```
agnivade, line 83: I think we can get rid of the `store.MISSING_CHANNEL_ERROR` altogether and just use the error key directly. It is anyways a bit misleading now because the error is an app layer error but it is in the store package. This would also need changing in 2 places in the api4 layer.

app/channel.go
```diff
@@ -1510,12 +1518,7 @@ func (a *App) GetChannelByName(channelName, teamId string, includeDeleted bool)
 func (a *App) GetChannelsByNames(channelNames []string, teamId string) ([]*model.Channel, *model.AppError) {
 	channels, err := a.Srv().Store.Channel().GetByNames(teamId, channelNames, true)
 	if err != nil {
-		if err.Id == "store.sql_channel.get_by_name.missing.app_error" {
```
agnivade, line 1513: This was interesting. Does this mean that this check was spurious ?

app/command_msg.go
rvillablanca: Yep, this was a mistake. I use goimports on save so I was added by him, even I didn't know that `github.com/pkg/errors` had the same methods `as` and `is`

store/sqlstore/channel_store.go
rvillablanca: Sure ;)

app/channel.go
rvillablanca: Yep, I thought about that but I forgot to mention it. I thing this part is a bit messy because some error keys are in constant and other simply hardcoded.

rvillablanca:
I'm done :)

app/channel.go
agnivade: Yeah totally agree. It's good that we are fixing these now :)

agnivade:
@lindy65 - The same testing that you do by testing all sorts of getting active+archived channels scenarios should work.

rvillablanca:
Resolving conflicts...

rvillablanca:
Conflicts resolved

PR #14712, opened 2020-06-01
MM-23574: Remove Segment.

grundleborg:
#### Summary
Delete Segment.

@agnivade requesting your review because `go mod tidy` did more than I was expecting...

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-23574


app/server.go
```diff
@@ -442,7 +440,7 @@ func (s *Server) Shutdown() error {
 		}
 	}
 
-	err := s.shutdownDiagnostics()
+	err := s.shutdownRudder()
```
reflog, line 443: i'm 1/4 on this, but I think we should lave the names as `shutdownDiagnostics` and `initDiagnostics`. Easier for people to read the code (not everyone knows what Rudder is)

grundleborg:
@reflog agreed - I've updated it.

app/server.go
```diff
@@ -125,9 +124,8 @@ type Server struct {
 	clientConfigHash    atomic.Value
 	limitedClientConfig atomic.Value
 
-	diagnosticId     string
-	diagnosticClient analytics.Client
-	rudderClient     rudder.Client
+	diagnosticId string
+	rudderClient rudder.Client
```
agnivade, line 128: Should we name it `diagnosticClient` to keep things same as previous ?

lindy65:
Hi @grundleborg - is basic smoke testing ok for testing this PR or is there anything specific I should be testing for? (other than Segment being removed that is)

grundleborg:
Quick smoke test should be fine. If it's reasonably straightforward, checking with Alex to make sure Server-side telemetry is still received via Rudder when you are testing this PR would also be great, but that can happen in master if needed.

lindy65:
Thanks! I believe Rudder can only be checked after merge so we can check that out on `master` 👍

PR #14720, opened 2020-06-01
Cherry pick 5.23 - [MM-25677] Content-Type is optional (#14705)

ethervoid:
#### Summary

* Content-Type is optional

mime.ParseMimeType returns and "no media type" error if the passed
string is empty.

Given that the Content-Type header is optional we shouldn't return an error
in that case, so we're fixing that allowing the users to call the webhook
without passing that header

* Include webhook id in the error message

Given that the number of webhooks could be big the user could
need the id to check which one of the multiple webhooks are failing
so include the id aids in that part

#### Ticket Link

https://mattermost.atlassian.net/browse/MM-25677

### DIFF
--- a/web/webhook.go
+++ b/web/webhook.go
@@ -28,11 +28,22 @@ func incomingWebhook(c *Context, w http.ResponseWriter, r *http.Request) {
 	r.ParseForm()
 
 	var err *model.AppError
+	var mediaType string
 	incomingWebhookPayload := &model.IncomingWebhookRequest{}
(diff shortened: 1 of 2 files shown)


amyblais:
Do we know why the build is taking so long?

ethervoid:
@amyblais it failed but didn't show the error in the checker. I've restarted it

amyblais:
May we need help from DevOps team? I know it's getting late for you and I would like to aim to get this done by today.

PR #14725, opened 2020-06-01
MM-24467- Use new ServiceProviderIdentifier in AuthnRequest

sbishel:
#### Summary
The tag in the AuthnRequest is being created incorrectly. This PR uses a new configuration setting to set the <Issuer>. The new configuration setting is getting set to the prior value so that existing installations will continue to work.

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-24467

#### Related PRs
https://github.com/mattermost/enterprise/pull/674
https://github.com/mattermost/mattermost-webapp/pull/5683


amyblais:
@sbishel Was this added to diagnostics https://github.com/mattermost/mattermost-server/blob/master/app/diagnostics.go?

sbishel:
@amyblais No, but it shouldn't need to be. It is a required field and if SAML is turned on, it must be filled in. There are 3 or 4 other fields in this section that fall into that category as well.(ie. IdpUrl, IdpDescriptorUrl, IdpMetadataUrl, AssertionConsumerServiceUrl).

PR #14726, opened 2020-06-02
Migrate channelbynamecache storelayer

rvillablanca:
#### Summary

This migrates the `channelByNameCache` from `ChannelStore` to `LocalCacheChannelStore`.

#### Ticket Link

#13100

### DIFF
--- a/store/localcachelayer/channel_layer.go
+++ b/store/localcachelayer/channel_layer.go
@@ -4,6 +4,11 @@
 package localcachelayer
 
 import (
+	"database/sql"
+	"fmt"
+	"net/http"
+	"strings"
+
 	"github.com/mattermost/mattermost-server/v5/model"
 	"github.com/mattermost/mattermost-server/v5/store"
 )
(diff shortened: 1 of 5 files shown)


rvillablanca:
@jespino you could be interested in this change, although I still need to add the tests

rvillablanca:
At first the tests are passing but I think this could need more tests, I have to more for `GetByNames`

agnivade:
This is okay, but ideally, we need to intercept _all_ the store methods in the cache layer and move all the invalidations from the app layer to the cache layer. I am 0/5 on whether to do it in this PR or later, but for your understanding, this is the ideal way to do it.

rvillablanca:
I have some conflicts here but I'm not sure if it is worth fixing, I know that there are some bugs around cache and you are taking another approach to improve the caching process.

When I get feedback from you I'll make the changes only if is necessary :)

agnivade:
The conflicts are for the migration to the cache2 interface. It is an orthogonal thing to the cache invalidation bugs that we want to address. You can take a look at this PR https://github.com/mattermost/mattermost-server/pull/14691 which migrates everything else to cache2.

PR #14733, opened 2020-06-02
[MM-25263] Add group_id parameter to existing users endpoints

fmunshi:
#### Summary
- Even though we had an existing `GetMemberUsers` endpoint for groups I think it makes more sense to treat groups similar to channels and teams and allow the users api to handle getting and searching group users when a group Id is passed as a parameter

- Also adds a dedicated group/stats api to get member count.

#### Todo: 
- [x] Add documentation for new endpoint

#### Ticket Link
- https://mattermost.atlassian.net/browse/MM-25263


fmunshi:
Related Docs PR https://github.com/mattermost/mattermost-api-reference

PR #14735, opened 2020-06-03
MM-24876: local mode support for getPost

agnivade:
This PR adds an handler for getPost which can be used
in local mode.


agnivade:
woops, branched off of master.

api4/post_test.go
```diff
@@ -1855,48 +1855,66 @@ func TestGetPostsForChannelAroundLastUnread(t *testing.T) {
...
 	th := Setup(t).InitBasic()
 	defer th.TearDown()
+	// TODO: migrate this entirely to the subtest's client
+	// once the other methods are migrated too.
 	Client := th.Client
 
-	post, resp := Client.GetPost(th.BasicPost.Id, "")
-	CheckNoError(t, resp)
+	var privatePost *model.Post
+	th.TestForClientAndLocal(t, func(t *testing.T, c *model.Client4) {
```
ashishbhate, line 1863: I think we can use `TestForAllClients` here. The system admin and local mode clients should behave the same. We can then drop the check that's on line 1914

agnivade: Done. Left the utility function though. In case someone needs it.

agnivade:
Gentle ping @mgdelacroix

api4/apitestlib.go
```diff
@@ -704,6 +704,22 @@ func (me *TestHelper) CreateGroup() *model.Group {
 	return group
 }
 
+// TestForClientAndLocal runs a test function for the normal client and as a local client.
+func (me *TestHelper) TestForClientAndLocal(t *testing.T, f func(*testing.T, *model.Client4), name ...string) {
```
mgdelacroix, line 708: nonblocking: the local client actions are supposed to be closer to a sysadmin than to an unprivileged client, so it's difficult that this helper ends being used. I'm happy leaving it here tho 1/5

agnivade: Ah I see. That makes sense. Will remove.

PR #14736, opened 2020-06-03
MM-24876: add support for getPost in local mode

agnivade:
This PR adds an handler for getPost which can be used
in local mode.

#### Ticket link
https://mattermost.atlassian.net/browse/MM-24876


agnivade:
Created a PR against this branch as per instructions in the ticket. But this will create conflicts against master because the code in app/authorization.go is already there in master with some other PRs.

agnivade:
I need to make it against master. Closing again :(

PR #14746, opened 2020-06-03
MM-25737: Fix for groups list not returning results.

mkraft:
#### Summary

Fixes incorrect SQL query that filters groups even if the parent team is not group-constrained.

#### Ticket Link

https://mattermost.atlassian.net/browse/MM-25737


amyblais:
@mkraft Just to confirm, is this a regression in v5.24?

mkraft:
> @mkraft Just to confirm, is this a regression in v5.24?

Yes, it was introduced in https://github.com/mattermost/mattermost-server/pull/14619.

PR #14756, opened 2020-06-04
MM-24914 Various fixes for sidebar channel handling

hmhealey:
Fixes for some bugs I'm coming across while attaching the redux and server code for the new sidebar. Most are fairly trivial, but I do want to call out a couple:
1. I removed the unique constraint on both sidebar tables. I had intended for that constraint to remove the possibility that we somehow end up with two items having the same SortOrder value, but I ran into issues because gorp updates rows one at a time, so something like rearranging items A, B, and C to A, C, B would cause a constraint error.
2. I added some stuff to the handling of orphaned channels since I was running into issues where channels from the wrong team and in custom categories were being added to the Channels category.

I'm also aware of another issue with channels not being unfavourited properly in the preferences, but that can wait for a separate PR.

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-24914

### DIFF
--- a/store/sqlstore/channel_store.go
+++ b/store/sqlstore/channel_store.go
@@ -393,8 +393,18 @@ func newSqlChannelStore(sqlStore SqlStore, metrics einterfaces.MetricsInterface)
 		tablePublicChannels.ColMap("Purpose").SetMaxSize(250)
 
 		tableSidebarCategories := db.AddTableWithName(model.SidebarCategory{}, "SidebarCategories").SetKeys(false, "Id")
+		tableSidebarCategories.ColMap("Id").SetMaxSize(26)
+		tableSidebarCategories.ColMap("UserId").SetMaxSize(26)
+		tableSidebarCategories.ColMap("TeamId").SetMaxSize(26)
+		tableSidebarCategories.ColMap("Sorting").SetMaxSize(64)
+		tableSidebarCategories.ColMap("Type").SetMaxSize(64)
+		tableSidebarCategories.ColMap("DisplayName").SetMaxSize(64)
+
 		tableSidebarChannels := db.AddTableWithName(model.SidebarChannel{}, "SidebarChannels").SetKeys(false, "ChannelId", "UserId", "CategoryId")
 		tableSidebarChannels.SetUniqueTogether("CategoryId", "SortOrder")
+		tableSidebarChannels.ColMap("ChannelId").SetMaxSize(26)
+		tableSidebarChannels.ColMap("UserId").SetMaxSize(26)
+		tableSidebarChannels.ColMap("CategoryId").SetMaxSize(26)
 	}
 
 	return s
--- a/store/sqlstore/channel_store.go
+++ b/store/sqlstore/channel_store.go
@@ -3659,13 +3659,6 @@ func (s SqlChannelStore) UpdateSidebarCategories(userId, teamId string, categori
 			}
 			category.DisplayName = categoryToUpdate.DisplayName
 		}
-		// we can update sort method only on DM category
-		if category.Sorting != categoryToUpdate.Sorting {
-			if category.Type != model.SidebarCategoryDirectMessages {
-				return nil, model.NewAppError("SqlPostStore.UpdateSidebarCategory", "store.sql_channel.sidebar_categories.invalid_update_fields", map[string]interface{}{"field": "Sorting"}, err.Error(), http.StatusBadRequest)
-			}
-			category.Sorting = categoryToUpdate.Sorting
-		}
 
 		if _, err = transaction.UpdateColumns(func(col *gorp.ColumnMap) bool {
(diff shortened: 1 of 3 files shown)


store/storetest/channel_store.go
```diff
@@ -6617,13 +6618,287 @@ func testSidebarChannelsMigration(t *testing.T, ss store.Store) {
...
+func testGetSidebarCategory(t *testing.T, ss store.Store, s SqlSupplier) {
+	t.Run("should return a custom category with its Channels field set", func(t *testing.T) {
+		user := &model.User{Id: model.NewId()}
+		teamId := model.NewId()
+
+		channelId1 := model.NewId()
+		channelId2 := model.NewId()
+		channelId3 := model.NewId()
+
+		appErr := ss.Channel().CreateInitialSidebarCategories(user, teamId)
```
agnivade, line 6630: The convention that we are following now (with the appError migration) is that AppErrors are declared as err, and errors are declared as nErr. 0/5 if you want to do it here or not, just wanted to let you know.

store/sqlstore/channel_store.go
```diff
@@ -3640,23 +3656,18 @@ func (s SqlChannelStore) UpdateSidebarCategories(userId, teamId string, categori
...
+
 		// we can update display name only on custom categories
 		if category.DisplayName != categoryToUpdate.DisplayName {
 			if category.Type != model.SidebarCategoryCustom {
-				return nil, model.NewAppError("SqlPostStore.UpdateSidebarCategory", "store.sql_channel.sidebar_categories.invalid_update_fields", map[string]interface{}{"field": "DisplayName"}, err.Error(), http.StatusBadRequest)
+				return nil, model.NewAppError("SqlPostStore.UpdateSidebarCategory", "store.sql_channel.sidebar_categories.invalid_update_fields", map[string]interface{}{"field": "DisplayName"}, "", http.StatusBadRequest)
 			}
 			category.DisplayName = categoryToUpdate.DisplayName
 		}
-		// we can update sort method only on DM category
```
reflog, line 3653: was this requirement changed?

store/storetest/channel_store.go
```diff
@@ -6617,13 +6618,287 @@ func testSidebarChannelsMigration(t *testing.T, ss store.Store) {
...
+}
+
 func testCreateInitialSidebarCategories(t *testing.T, ss store.Store) {
 	t.Run("should create initial favorites/channels/DMs categories", func(t *testing.T) {
 		user := &model.User{Id: model.NewId()}
 		teamId := model.NewId()
 
 		err := ss.Channel().CreateInitialSidebarCategories(user, teamId)
-		require.Nil(t, err)
+		assert.Nil(t, err)
```
reflog, line 6901: any reason for the change?

store/sqlstore/channel_store.go
hmhealey: I should've changed it with my manual vs default sort method PR. I'm setting the sort order to manual when someone rearranges their channels manually. It could make it weird if someone uses the API to set their sort order to something different using the API, but I think everything should still work fine regardless. Sorry about making you add it when I just ended up removing it.

store/storetest/channel_store.go
hmhealey: It's just something I missed writing the tests. I tend to use `require` for test set up and `assert` for checking things that I actually care about in the test

store/storetest/channel_store.go
hmhealey: Yeah, I can change that. Most of the old tests use err/appErr as names, but I'm good using the new way of doing it.

store/sqlstore/channel_store.go
```diff
@@ -3721,10 +3712,47 @@ func (s SqlChannelStore) UpdateSidebarCategories(userId, teamId string, categori
...
+			sql, args, _ = s.getQueryBuilder().Delete("Preferences").Where(
+				sq.Eq{
+					"Name":     originalCategory.Channels,
+					"Category": model.PREFERENCE_CATEGORY_FAVORITE_CHANNEL,
+				},
+			).ToSql()
+
+			if _, err = transaction.Exec(sql, args...); err != nil {
+				return nil, model.NewAppError("SqlPostStore.UpdateSidebarCategory", "store.sql_channel.sidebar_categories.app_error", nil, err.Error(), http.StatusInternalServerError)
+			}
```
agnivade, line 3727: Could we perhaps refactor this deletion code from the if and the else conditions into a single one ? Only the `Name` column needs to change. And then add the new ones if type is `SidebarCategoryFavorites`.

hmhealey: I don't think there's much we could do to make that shorter since `Name` pulls from the two different arrays there. The best we could do is to make a helper function to delete multiple preferences by category and name, but I don't think that would do a ton to make it more readable.

Also, I do kind of like having two clear paths for Favorites and non-Favorites categories since the steps for updating the preferences in each case are perfectly clear with this.

PR #14758, opened 2020-06-05
MM-25700 : Use a counting semaphore for push notifications hub

agnivade:
The ordering of push notifications for a single user is anyways not maintained
because the push proxy runs behind a load-balanced setup. Therefore, there is
no guarantee that notifications for a single user will be sent serially.

With that restriction lifted, we can simplify the architecture and get
rid of our hardcoded set of workers with a simple counting semaphore
with a bounded concurrency.

#### Ticket link

https://mattermost.atlassian.net/browse/MM-25700


app/notification_push.go
```diff
@@ -245,63 +239,73 @@ func (a *App) updateMobileAppBadgeSync(userId string) *model.AppError {
...
-			err = a.updateMobileAppBadgeSync(notification.userId)
-		default:
-			mlog.Error("Invalid notification type", mlog.String("notification_type", string(notification.notificationType)))
-		}
-
-		if err != nil {
-			mlog.Error("Unable to send push notification", mlog.String("notification_type", string(notification.notificationType)), mlog.Err(err))
-		}
+func (hub *PushNotificationsHub) start() {
+	hub.sema = make(chan struct{}, runtime.NumCPU()*8) // numCPU * 8 is a good amount of concurrency.
```
jespino, line 259: I'm not 100% sure about this part, what should be a sensible amount of parallelism? What happen if the push proxy stop working or start to reply slowly (the default connection timeout is 3 seconds and the default request timeout is 30 seconds)? for example, each rountrip taking 200ms, that means 5 * 8 notifications per CPU, with a 8 CPUs maching is 320 notifications per second, sounds like a lot, but after that, anything that needs to notify is going to perceive the contention there. Considering how cheap are the coroutines and the mainly IO bound tasks, makes sense to increase the per-cpu concurrency here?

agnivade: That's a good point. To add a data point, I checked our graphs and at peaks, we are serving about 200 notifications/second. 

> What happen if the push proxy stop working or start to reply slowly

Then it would apply backpressure to the sender. This is one of the features of having a bounded semaphore like this. That it can apply backpressure rather than overloading the upstream server.

> Considering how cheap are the coroutines and the mainly IO bound tasks, makes sense to increase the per-cpu concurrency here?

One thing to note is that this is for a single server, so in a clustered setup, it would get multiplied by the number of servers. At the end of the day, it's a heuristic and there's no right value. If we keep it too low, it would reduce concurrency, if we make it too high, then it would overload the push proxy, and also create high goroutine contention in the server. I am happy to increase it to 16 if you prefer. Let me know.

jespino: My concert is a spike in notifications can freeze post creation during certain amount of tiem.

I'm not 100% sure, but I think the notifications are sent from the node that receive the post itself, not distribured through all the cluster nodes, so 1 post that emits multiple notifications, has to send all the notifications from the same server.

My concern is about a situation like Town Square, with 20000 users, some of them connected, some of them not, some of them with multiple devices, some of them with non, then, somebody mention `@all` (because he can, or because really want to notify something to the whole company). Then, you have lets say 10.000 notifications to send, even if you send 1000 notifications per second, you still have to wait 9 seconds until any of the coroutines that really send notifications are free, until then, any post that needs to notify anybody is going to be blocked. That is my concern about this approach. I'm not sure if our previous solution was more reliable in that scenarios, but I think was at least enqueue more data.

Maybe we can change the notifications buffer size dynamically using a config setting or something like that. Allowing that scenarios to be handle in big deployments, and allow to save memory in small/memory-sensitive environments.

agnivade: That seems reasonable to me.

agnivade:
@jespino - I had to use the fakeapp again here. I could nearly migrate everything except for `a.ClearSessionCacheForUser` which involved session related things which then went on to cache invalidation and web_hub. Perhaps this can be done after the session refactor you have planned.

As a benefit, this exercise also brought out the things needed to refactor push notification to a separate service. In short, the only thing that's needed is an app-independant way to invalidate caches.

agnivade:
Gentle ping on this @ogi-m

amyblais:
@agnivade Was this added to ``app/diagnostics.go`` (if needed)?

agnivade:
It's not needed.

PR #14759, opened 2020-06-05
MM-25701: Reuse HTTP client in sendAckToPushProxy

agnivade:
We reuse the http client made for push notifications.
This leads to reuse of TCP connections rather than
creating a new client every time.

#### Ticket link

https://mattermost.atlassian.net/browse/MM-25701


agnivade:
Gentle ping @ogi-m

PR #14762, opened 2020-06-05
[MM-25767] Quick switch users search is always falling back to the database

ethervoid:
#### Summary

This PR fixes multiple problems we have when filtering the list of allowed channels in different situations: with or with team id, with view restrictions, etc

#### Ticket Link

https://mattermost.atlassian.net/browse/MM-25767


ethervoid:
@mgdelacroix and @jespino please take a look at this draft. I need to include more coverage and test different options (API, manually testing, etc) but I want to know your opinion because this is a sensitive change

api4/user_test.go
```diff
@@ -997,6 +997,7 @@ func TestAutocompleteUsersInChannel(t *testing.T) {
 		Username        string
 		ExpectedResults int
 		MoreThan        bool
+		Exception       bool
```
jespino, line 1000: `Exception`? maybe you mean `Error` or `ShouldFail` or something like that?

store/searchlayer/user_layer.go
```diff
@@ -146,45 +145,48 @@ func (s *SearchUserStore) autocompleteUsersInChannelByEngine(engine searchengine
 	return autocomplete, nil
 }
 
+// We're contemplating different scenarios in this method:
```
jespino, line 148: In godoc is a good practice to use the name of the function as the beginning of the explanation.

ethervoid:
@jespino done! @mgdelacroix I've check if the guest user is only returning the channel they're in and is working as expected

store/searchlayer/user_layer.go
```diff
@@ -146,45 +145,48 @@ func (s *SearchUserStore) autocompleteUsersInChannelByEngine(engine searchengine
 	return autocomplete, nil
 }
 
+// getListOfAllowedChannelsForTeam We're contemplating different scenarios in this method:
```
jespino, line 148: actually is more a explanation, something like `getListOfAllowedChannelsForTeam return the list of allowed channels contemplating different scenarios:`

api4/user.go
```diff
@@ -852,9 +852,18 @@ func autocompleteUsers(c *Context, w http.ResponseWriter, r *http.Request) {
 	}
 
 	if len(channelId) > 0 {
-		// Applying the provided teamId here is useful for DMs and GMs which don't belong
-		// to a team. Applying it when the channel does belong to a team makes less sense,
-		// but the permissions are checked above regardless.
+		// We're using the channelId to search for users inside that channel and the team
```
mgdelacroix, line 855: :heart:

api4/user.go
```diff
@@ -852,9 +852,18 @@ func autocompleteUsers(c *Context, w http.ResponseWriter, r *http.Request) {
...
+		// to get the not in channel list. Also we want to include the DM and GM users for
+		// that team which could only be obtained having the team id.
+		if len(teamId) == 0 {
+			c.Err = model.NewAppError("autocompleteUser",
+				"api.user.autocomplete_users.missing_team_id.app_error",
+				nil,
+				"channelId="+channelId+", teamId="+teamId,
```
mgdelacroix, line 862: Having the teamId here seems unnecessary:

```suggestion
				"channelId="+channelId",
```

PR #14764, opened 2020-06-07
MM-25779 - Update apiTimerLayerTemplate and hooksTimerLayerTemplate to record error values

### LINKED ISSUE #14754: Update apiTimerLayerTemplate and hooksTimerLayerTemplate to record error return values
When recording prometheus metrics for all RPC-Plugin-API calls we currently record *all* return values as success, even when the API returns an `AppError`/`error` value.

Proposal is to update [`apiTimerLayerTemplate`](https://github.com/mattermost/mattermost-server/blob/bd9d674006b34de12fc8ef6296c56adf8bcb2696/plugin/interface_generator/main.go#L321) and [`hooksTimerLayerTemplate `](https://github.com/mattermost/mattermost-server/blob/bd9d674006b34de12fc8ef6296c56adf8bcb2696/plugin/interface_generator/main.go#L363) to record `success=false` for error return values.

Example: for `OnActivate` hook we currently always record all calls as success as per current implementation:
```
func (hooks *hooksTimerLayer) OnActivate() error {
	startTime := timePkg.Now()
	_returnsA := hooks.hooksImpl.OnActivate()
	hooks.recordTime(startTime, "OnActivate", true /* always success */)
	return _returnsA
}
```

Proposal is to update both templates such that it generated structs that check for any error return values and record the result, such as:
```
func (hooks *hooksTimerLayer) OnActivate() error {
	startTime := timePkg.Now()
	_returnsA := hooks.hooksImpl.OnActivate()
	hooks.recordTime(startTime, "OnActivate", _returnsA == nil /* check for nil err */)
	return _returnsA
}
```
```
func (api *apiTimerLayer) GetBundlePath() (string, error) {
	startTime := timePkg.Now()
	*returnsA, *returnsB := api.apiImpl.GetBundlePath()
	api.recordTime(startTime, "GetBundlePath", _returnsB == nil /* check for nil err */)
	return *returnsA, *returnsB
}
```

The templates are written using Go's text/templating system. Code traversal is done through go/ast. See more resources below.

Run following to generate `apiTimerLayer` and `hooksTimerLayer` from templates.

```
mattermost-server> make pluginapi
```

Helpful resources:
1. https://mattermost.com/blog/instrumenting-go-code-via-ast/
2. https://golang.org/pkg/text/template/

----

JIRA: https://mattermost.atlassian.net/browse/MM-25779

Ashniu123:
#### Summary
Template update to check for error return value to be `nil` if it exists otherwise defaults to `true`.

#### Ticket Link

Closes #14754 on Github
Closes JIRA https://mattermost.atlassian.net/browse/MM-25779

### DIFF
--- a/plugin/interface_generator/main.go
+++ b/plugin/interface_generator/main.go
@@ -141,7 +141,7 @@ func FieldListDestruct(structPrefix string, fieldList *ast.FieldList, fileset *t
 	return strings.Join(result, ", ")
 }
 
-func FieldListToRecordError(structPrefix string, fieldList *ast.FieldList, fileset *token.FileSet) string {
+func FieldListToRecordSuccess(structPrefix string, fieldList *ast.FieldList, fileset *token.FileSet) string {
 	if fieldList == nil || len(fieldList.List) == 0 {
 		return ""
 	}
@@ -380,7 +380,7 @@ func (api *apiTimerLayer) recordTime(startTime timePkg.Time, name string, succes
 func (api *apiTimerLayer) {{.Name}}{{funcStyle .Params}} {{funcStyle .Return}} {
 	startTime := timePkg.Now()
 	{{ if .Return }} {{destruct "_returns" .Return}} := {{ end }} api.apiImpl.{{.Name}}({{valuesOnly .Params}})
-	api.recordTime(startTime, "{{.Name}}", {{ if .Return }} {{ destructErrors "_returns" .Return }} {{ else }} true {{ end }})
+	api.recordTime(startTime, "{{.Name}}", {{ if .Return }} {{ shouldRecordSuccess "_returns" .Return }} {{ else }} true {{ end }})
 	{{ if .Return }} return {{destruct "_returns" .Return}} {{ end -}}
 }
 
@@ -422,7 +422,7 @@ func (hooks *hooksTimerLayer) recordTime(startTime timePkg.Time, name string, su
 func (hooks *hooksTimerLayer) {{.Name}}{{funcStyle .Params}} {{funcStyle .Return}} {
 	startTime := timePkg.Now()
 	{{ if .Return }} {{destruct "_returns" .Return}} := {{ end }} hooks.hooksImpl.{{.Name}}({{valuesOnly .Params}})
-	hooks.recordTime(startTime, "{{.Name}}", {{ if .Return }} {{ destructErrors "_returns" .Return }} {{ else }} true {{ end }})
+	hooks.recordTime(startTime, "{{.Name}}", {{ if .Return }} {{ shouldRecordSuccess "_returns" .Return }} {{ else }} true {{ end }})
 	{{ if .Return }} return {{destruct "_returns" .Return}} {{end -}}
 }
 
@@ -451,8 +451,8 @@ func generateHooksGlue(info *PluginInterfaceInfo) {
 		"destruct": func(structPrefix string, fields *ast.FieldList) string {
 			return FieldListDestruct(structPrefix, fields, info.FileSet)
 		},
-		"destructErrors": func(structPrefix string, fields *ast.FieldList) string {
-			return FieldListToRecordError(structPrefix, fields, info.FileSet)
+		"shouldRecordSuccess": func(structPrefix string, fields *ast.FieldList) string {
+			return FieldListToRecordSuccess(structPrefix, fields, info.FileSet)
 		},
 		"obscure": func(name string) string {
 			return "Z_" + name
@@ -500,8 +500,8 @@ func generatePluginTimerLayer(info *PluginInterfaceInfo) {
 		"destruct": func(structPrefix string, fields *ast.FieldList) string {
 			return FieldListDestruct(structPrefix, fields, info.FileSet)
 		},
-		"destructErrors": func(structPrefix string, fields *ast.FieldList) string {
-			return FieldListToRecordError(structPrefix, fields, info.FileSet)
+		"shouldRecordSuccess": func(structPrefix string, fields *ast.FieldList) string {
+			return FieldListToRecordSuccess(structPrefix, fields, info.FileSet)
 		},
 	}
 
--- a/plugin/interface_generator/main.go
+++ b/plugin/interface_generator/main.go
@@ -149,12 +149,7 @@ func FieldListToRecordSuccess(structPrefix string, fieldList *ast.FieldList, fil
 	result := ""
 	nextLetter := 'A'
 	for _, field := range fieldList.List {
-		typeNameBuffer := &bytes.Buffer{}
-		err := printer.Fprint(typeNameBuffer, fileset, field.Type)
-		if err != nil {
-			panic(err)
-		}
-		typeName := typeNameBuffer.String()
+		typeName := baseTypeName(field.Type)
 		if strings.Contains(typeName, "error") || strings.Contains(typeName, "Error") {
 			result = structPrefix + string(nextLetter)
 			break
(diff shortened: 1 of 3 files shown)


plugin/interface_generator/main.go
```diff
@@ -141,6 +141,33 @@ func FieldListDestruct(structPrefix string, fieldList *ast.FieldList, fileset *t
...
+		typeName := typeNameBuffer.String()
+		if strings.Contains(typeName, "error") || strings.Contains(typeName, "Error") {
+			result = structPrefix + string(nextLetter)
+			break
+		}
+		nextLetter++
+	}
+
+	if result == "" {
+		return "true"
```
ali-farooq0, line 166: since we found no error variables, lets return a `""` here and handle the else case where this is being called, otherwise we'll have two else paths.

plugin/interface_generator/main.go
```diff
@@ -141,6 +141,33 @@ func FieldListDestruct(structPrefix string, fieldList *ast.FieldList, fileset *t
 	return strings.Join(result, ", ")
 }
 
+func FieldListDestructErrors(structPrefix string, fieldList *ast.FieldList, fileset *token.FileSet) string {
```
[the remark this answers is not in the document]
ali-farooq0: On a second thought maybe lets keep the `nil` check here and call this `FieldListToRecordError` instead and that will help with the method's context maybe.

plugin/interface_generator/main.go
```diff
@@ -141,6 +141,33 @@ func FieldListDestruct(structPrefix string, fieldList *ast.FieldList, fileset *t
...
+			result = structPrefix + string(nextLetter)
+			break
+		}
+		nextLetter++
+	}
+
+	if result == "" {
+		return "true"
+	}
+	return fmt.Sprintf("%s == nil", result)
```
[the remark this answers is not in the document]
Ashniu123: It will clash when `true` is returned.

Ashniu123:
@ali-farooq0 What about the `destructErrors` name. Is that fine or should I change it to something else?

plugin/interface_generator/main.go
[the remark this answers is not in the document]
Ashniu123: The main reason I have done it this way is because some functions return ` *model.AppError` or `error` type where what you are saying is works to check if the error is `nil`. 

But in some places there is `*http.Response` or `bool` or some other _non-error_ return value. There directly `true` should be returned even when there are return fields. Hence, I have done it this way. 

And finally, there are some functions which don't return any values and `true` should be returned. This is handled by the `... {{ else }} ...` part.

What do you think?

ali-farooq0:
@Ashniu123 good question -- lets call it `shouldRecordSuccess`. Lets also rename the other one to `FieldListToRecordSuccess` for consistency as well -- apologies for the naming churn lol really trying to think of better names. Hopefully this will help the next dev coming in to work on this :)

plugin/interface_generator/main.go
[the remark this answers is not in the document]
ali-farooq0: Ah I see. Thanks for the explanation. Could we just merge case 1 and case 3? and just return `true` for both from this function. That will help eliminate the `{{ else }}` from the calling part and keep all the logic here. What do you think? @Ashniu123

plugin/interface_generator/main.go
```diff
@@ -141,6 +141,33 @@ func FieldListDestruct(structPrefix string, fieldList *ast.FieldList, fileset *t
...
+
+	result := ""
+	nextLetter := 'A'
+	for _, field := range fieldList.List {
+		typeNameBuffer := &bytes.Buffer{}
+		err := printer.Fprint(typeNameBuffer, fileset, field.Type)
+		if err != nil {
+			panic(err)
+		}
+		typeName := typeNameBuffer.String()
```
agnivade, line 157: This is not really required. You can get the base type from `field.Type` through some recursion:

```go
func baseTypeName(x ast.Expr) string {
	switch t := x.(type) {
	case *ast.Ident:
		return t.Name
	case *ast.SelectorExpr:
		if _, ok := t.X.(*ast.Ident); ok {
			// only possible for qualified type names;
			// assume type is imported
			return t.Sel.Name
		}
	case *ast.ParenExpr, *ast.StarExpr:
		return baseTypeName(t.X)
	}
	return ""
}
```

And then use `baseTypeName(typeName) == "error" || baseTypeName(typeName) == "*model.AppError"`. I haven't tested this though. This is just code copied from the standard library.

plugin/interface_generator/main.go
```diff
@@ -353,7 +380,7 @@ func (api *apiTimerLayer) recordTime(startTime timePkg.Time, name string, succes
 func (api *apiTimerLayer) {{.Name}}{{funcStyle .Params}} {{funcStyle .Return}} {
 	startTime := timePkg.Now()
 	{{ if .Return }} {{destruct "_returns" .Return}} := {{ end }} api.apiImpl.{{.Name}}({{valuesOnly .Params}})
-	api.recordTime(startTime, "{{.Name}}", true)
+	api.recordTime(startTime, "{{.Name}}", {{ if .Return }} {{ shouldRecordSuccess "_returns" .Return }} {{ else }} true {{ end }})
```
agnivade, line 383: I was wondering if we could refactor this whole if-else to another helper function ? Seems like it's completely duplicated to me.

plugin/interface_generator/main.go
Ashniu123: Doesn't it reduce readability of code?

agnivade: IMO, this is the right way to get the type of an expression. What we are doing right now is a bit hackish, by printing the expression and then comparing the string.

It's just code which might be a bit difficult to understand at first, but it doesn't look unreadable to me.

Ashniu123: Oh. I didn't know that. I was just following the way it was done in other functions.

So should I change them all to reflect it this way?

agnivade: Oh my indeed ! I didn't notice that there are other places where we are doing this. There's no need to change other stuff in this PR. But maybe add this function and use it in this PR, and then send another PR migrating the other functions ?

Thanks.

Ashniu123: Could you share the link to where this code is, in the standard library?

Currently this throws error
```
❯ make pluginapi
go generate  ./plugin
# command-line-arguments
interface_generator/main.go:207:24: t.X undefined (type ast.Expr has no field or method X)
plugin/client_rpc.go:4: running "go": exit status 2
make: *** [Makefile:241: pluginapi] Error 1
```

agnivade: That is strange, all of those expressions have `X` in their field.

It is from here:
https://github.com/golang/go/blob/2a4e71c752875f84dc0061f75fdcbfc96ae8d5f2/src/go/doc/reader.go#L100-L116

plugin/interface_generator/main.go
```diff
@@ -141,6 +141,28 @@ func FieldListDestruct(structPrefix string, fieldList *ast.FieldList, fileset *t
...
+func FieldListToRecordSuccess(structPrefix string, fieldList *ast.FieldList, fileset *token.FileSet) string {
+	if fieldList == nil || len(fieldList.List) == 0 {
+		return "true"
+	}
+
+	result := ""
+	nextLetter := 'A'
+	for _, field := range fieldList.List {
+		typeName := baseTypeName(field.Type)
+		if strings.Contains(typeName, "error") || strings.Contains(typeName, "Error") {
```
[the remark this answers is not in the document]
Ashniu123: No need for "Sorry". I think all queries are valid and I like these thorough discussions. :smile: 

I'll just cross-check if all functions' "error" return type are in fact either `error` or `model.AppError` and then update it asap. :+1:

ali-farooq0:
Started a conversation to see if this can be tested: https://community-daily.mattermost.com/core/pl/nmjyjry6itd1mr9bmrcq896wgc

ali-farooq0:
Was able to successfully confirm that prometheus is recording both `true` and `false` `success` values.

#### plugin_api
image


#### plugin_hooks
image

image


Thanks @angeloskyratzakos for your help with this! 🎉

PR #14765, opened 2020-06-08
Migrate GetAllChannels and GetAllChannelsCount from ChannelStore to r…

rvillablanca:
#### Summary

Migrate `GetAllChannels` and `GetAllChannelsCount` to return plain errors.

Fixes https://github.com/mattermost/mattermost-server/issues/14766


rvillablanca:
Ticket is required to track this :+1: 
/cc @agnivade

lindy65:
Hi @agnivade - will testing around creating, renaming, deleting, archiving, making private be sufficient/suitable?

rvillablanca:
@jespino could you tell me what problem there is with the translations so I can fix them before creating the PR?

jespino:
@rvillablanca sure, the expectation is to have every single translation in the file ordered alphabetically by id, so... you don't have only to rename the key, you have to move that translations to the correct place in the file.

rvillablanca:
Ah sure, are you doing that with the command or manually? In another PR there were many changes and that was a little tricky to do it manually so I tried using the command (don't remember which one) bit it removed some EE translations

jespino:
Yes, I'm sure, but I use the command for simplicity. But the changes genrated are exactly that, move the strings changed to the correct position in the file.

Anyway, I don't mind to execute the command for you when is needed in this PRs.

agnivade:
@lindy65 - sorry about the delay. Whatever testing you did for https://github.com/mattermost/mattermost-server/pull/14688 can be applied here. There is no need to test for deleting/updating channels, just anything around getting channels.

lindy65:
Sounds good to me! Thanks @agnivade :)

rvillablanca:
Great, thank you all of you :)

PR #14768, opened 2020-06-08
MM-25507: Remove old migration state marker before running migrations

agnivade:
While resetting permissions, we were not removing old migration state residue
for the EMOJIS_PERMISSIONS_MIGRATION_KEY and GUEST_ROLES_CREATION_MIGRATION_KEY.
Therefore, during their migration, they got skipped because the code checks
if a migration key is already present or not.

To fix this, we remove the migration key just as we do for ADVANCED_PERMISSIONS_MIGRATION_KEY.

#### Ticket Link

https://mattermost.atlassian.net/browse/MM-25507
Fixes https://github.com/mattermost/mattermost-server/issues/14642


agnivade:
Nice catch @srkgupta. I took a look and the fix is somewhere else (related to MakeDefaultRoles). This fix is specifically related to tracking the migration state in the system table. I would like to keep them separate to prevent 2 different things being in the same PR.

Could you file another ticket for this and assign it to me?

agnivade:
Nvm, I made https://mattermost.atlassian.net/browse/MM-26206.

agnivade:
ping @srkgupta

PR #14771, opened 2020-06-08
Docker compose network

metanerd:
#### Summary

Since build.mattermost.com is gone, we can try again, if CIDRs still overlap. 
Link to original PR: https://github.com/mattermost/mattermost-server/pull/14223


metanerd:
And revert.

PR #14777, opened 2020-06-08
MM-25762 - Moved password settings to client limited configuration

marianunez:
#### Summary
- Moved password settings to be passed down also in the client limited configuration when the user is no yet logged in so the webApp can properly indicate password requirements at account creation. 
- Removed setting defaults given that `PasswordSettings` is not a licensed feature that need to have a pre-set value.

#### Ticket Link
[MM-25762](https://mattermost.atlassian.net/browse/MM-25762)


config/client.go
```diff
@@ -114,11 +114,6 @@ func GenerateClientConfig(c *model.Config, diagnosticID string, license *model.L
 	props["SamlPositionAttributeSet"] = "false"
 	props["EnableCluster"] = "false"
 	props["EnableMetrics"] = "false"
-	props["PasswordMinimumLength"] = "0"
-	props["PasswordRequireLowercase"] = "false"
-	props["PasswordRequireUppercase"] = "false"
-	props["PasswordRequireNumber"] = "false"
-	props["PasswordRequireSymbol"] = "false"
```
jupenur, line 121: Why was this here in the first place?

marianunez: I'm not sure, it sets defaults values for features that are EE. Maybe password settings used to be one of them in the past? 🤷‍♀️

PR #14785, opened 2020-06-09
Document SqlTeamStore delete methods

FlaviaBastos:
#### Summary
Document SqlTeamStore delete methods

#### Ticket Link
Fixes https://github.com/mattermost/mattermost-server/issues/14005

JIRA: [https://mattermost.atlassian.net/browse/MM-23001](https://mattermost.atlassian.net/browse/MM-23001)

### DIFF
--- a/store/sqlstore/team_store.go
+++ b/store/sqlstore/team_store.go
@@ -559,7 +559,8 @@ func (s SqlTeamStore) GetAllTeamPageListing(offset int, limit int) ([]*model.Tea
 	return teams, nil
 }
 
-// PermanentDelete deletes from the database the team entry that matches the teamId passed as parameter.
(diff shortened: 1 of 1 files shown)


store/sqlstore/team_store.go
```diff
@@ -559,6 +559,7 @@ func (s SqlTeamStore) GetAllTeamPageListing(offset int, limit int) ([]*model.Tea
 	return teams, nil
 }
 
+// PermanentDelete deletes from the database the team entry that matches the teamId passed as parameter.
```
jespino, line 562: Maybe we can explain here that this is not a "recuperable" action, and maybe say something like "to soft-delete the team you can Update it with the DeleteAt field set to the current millisecond using `model.GetMillis()`"

PR #14795, opened 2020-06-10
MM-24914 Fix updateCategory endpoint returning the wrong type

hmhealey:
The endpoint was returning an array of categories instead of just the one that was updated. I also fixed a couple cases where `model.Client4` constructed URLs wrong and a spot where I broke updating sorting method in my last PR.


hmhealey:
@fmunshi Yep, we've been ignoring those. We'll add the docs when this is merged into master.

PR #14796, opened 2020-06-10
Prevent spurious error

rvillablanca:
#### Summary

Just a minor fix to prevent logging of spurios error when the response to an outgoing webhook is empty.

#### Ticket Link

https://mattermost.atlassian.net/browse/MM-25964


app/webhook.go
```diff
@@ -107,7 +107,7 @@ func (a *App) TriggerWebhook(payload *model.OutgoingWebhookPayload, hook *model.
 
 		a.Srv().Go(func() {
 			webhookResp, err := a.doOutgoingWebhookRequest(url, body, contentType)
-			if err != nil {
+			if err != nil && err != io.EOF {
```
lieut-data, line 110: I'm slightly opposed to catching `io.EOF` at this layer, because it implies this caller has to know something about how the underlying object is consumed. Could we instead have `doOutgoingWebhookRequest` check for `io.EOF` and just return `nil, nil` in that case, clearly signalling there is no response to handle?

rvillablanca: Yep you are right, it is better to make the caller agnostic of the error type, I'll make the changes and I will add the unit tests ;)

rvillablanca:
@lieut-data friendly reminder this is ready to be merged :)

lieut-data:
Thank you for the reminder, @rvillablanca! I've triggered the enterprise tests, and have assigned this my way so it doesn't get lost again :)

PR #14799, opened 2020-06-11
[MM-25406] Include missing methods in the search layer

ethervoid:
#### Summary

This PR adds:

- Missing methods in the channel, user and post layers like `PermanenteDeleteByUser`, `PermanenteDeleteByChannel`, etc
- Two new methods in the post layer: `deleteUserPostsIndex` and `deleteChannelPostsIndex`
- Include those new methods in all the search engines: bleve and ES

#### Ticket Link

https://mattermost.atlassian.net/browse/MM-25406


services/searchengine/bleveengine/search.go
```diff
@@ -209,6 +212,72 @@ func (b *BleveEngine) SearchPosts(channels *model.ChannelList, searchParams []*m
 	return postIds, matches, nil
 }
 
+func (b *BleveEngine) DeletePosts(searchRequest *bleve.SearchRequest, batchSize int) (int64, error) {
+	resultsCount := int64(0)
+
+	for {
+		// From is not a offset but from where to start picking results, so
+		// keeping it always to 0 makes the work
```
mgdelacroix, line 220: I find the wording slightly confusing as `From` is technically an offset, is just that we're not using it. A different proposal:

```suggestion
		// As we are deleting the posts after fetching them, we need to keep
		// From fixed always to 0
```

services/searchengine/bleveengine/search.go
```diff
@@ -209,6 +212,72 @@ func (b *BleveEngine) SearchPosts(channels *model.ChannelList, searchParams []*m
 	return postIds, matches, nil
 }
 
+func (b *BleveEngine) DeletePosts(searchRequest *bleve.SearchRequest, batchSize int) (int64, error) {
```
mgdelacroix, line 215: nit: does this method need to be public?

ethervoid: Not necessary but in case we want to start using it instead of creating specialized methods could be useful

mgdelacroix: As the method receives a `bleve.SearchRequest` object, I don't think is very likely that we use it  from outisde of the engine, but I'm really 1/5 on this, so if you feel strongly about it, I'm happy with the method being public :)

ethervoid: Hum, I see your point. Ok, I'll make it private :)

PR #14806, opened 2020-06-12
MM-24914 Make some changes to UpdateSidebarCategories

hmhealey:
This PR is divided into two things:
1. I've changed `UpdateSidebarCategories` to silently ignore any invalid changes passed to make the API easier to use. While doing that, I fixed some small issues for `UpdateSidebarCategories` and added tests to them.
2. I fixed the subquery in `completePopulatingCategoryChannels` (which is now named `doesNotHaveSidebarChannel`) because, when it was called on a DM channel, it wouldn't return DMs that were part of a custom category on another team.


hmhealey:
I just found a bug in the reordered `UpdateChannelCategories` code where it sometimes fails to move channels from one category to another.

The problem happens when you move a channel from one category to another in a specific order. For example, suppose you have CategoryA and CategoryB and you want to move SomeChannel from CategoryA to CategoryB. If you call `UpdateChannelCategories` with the changes to CategoryB first, it fails to move the channel because:
1. First, the new SidebarChannel entry in CategoryB is created ([on this line](https://github.com/mattermost/mattermost-server/blob/dd43e60c54a214d5a94c55194d8060b02d967f4b/store/sqlstore/channel_store.go#L3725))
2. Then, we delete the SidebarChannels for anything that was in CategoryA, including the channel created above ([on this line](https://github.com/mattermost/mattermost-server/blob/dd43e60c54a214d5a94c55194d8060b02d967f4b/store/sqlstore/channel_store.go#L3707).

Now that I think of it, since the unique constraints had to be removed for this, I think this also means that you can add a channel to multiple categories by calling `UpdateChannelCategories` repeatedly which will also cause more problems elsewhere.

store/sqlstore/channel_store.go
```diff
@@ -3659,73 +3670,83 @@ func (s SqlChannelStore) UpdateSidebarCategories(userId, teamId string, categori
...
 
-		// clean previous SidebarChannels
-		sql, args, _ := s.getQueryBuilder().Delete("SidebarChannels").Where(sq.Eq{"ChannelId": categoryToUpdate.Channels}).ToSql()
+		if category.Type != model.SidebarCategoryDirectMessages {
+			updatedCategory.Channels = make([]string, len(category.Channels))
+			copy(updatedCategory.Channels, category.Channels)
+		}
 
-		if _, err = transaction.Exec(sql, args...); err != nil {
+		if _, err = transaction.UpdateColumns(func(col *gorp.ColumnMap) bool {
```
jespino, line 3702: You said in the yesterday meeting the `UpdateColumns` is error prone and have some drawbacks, why no use an sql approach, like using Squirrel to build a query that clearly express the intention. "Update SidebarCategories SET DisplayName = X, Sorting = Y WHERE Id = Z"?

store/sqlstore/channel_store.go
```diff
@@ -3627,7 +3638,7 @@ func (s SqlChannelStore) updateSidebarCategoryOrderT(transaction *gorp.Transacti
 		runningOrder += model.MinimalSidebarSortDistance
 	}
 
-	if _, err := s.GetMaster().UpdateColumns(func(col *gorp.ColumnMap) bool {
+	if _, err := transaction.UpdateColumns(func(col *gorp.ColumnMap) bool {
```
jespino, line 3641: Same here, why not build your SQL query by hand?

store/sqlstore/channel_store.go
hmhealey: Just in case anyone comes across this later, the followup PR #14843 addresses this

store/sqlstore/channel_store.go
hmhealey: In the followup, this one is staying as-is. `UpdateColumns` loops through each item passed in end executes an UPDATE query for each one, and I'm thinking it might have something to make that looping more efficient than writing it by hand

PR #14807, opened 2020-06-12
Added SearchPostsInTeamForUser for plugin API

marianunez:
#### Summary
https://github.com/mattermost/mattermost-server/pull/10106 introduced `SearchPostsInTeam` to the plugin API to search posts with no user restrictions. However, if a plugin needed the same results as a user would see when searching in the webApp, the above endpoint would not be consistent (this was evident with a hashtag search using elastic search). 

This PR introduces the `SearchPostsInTeamForUser` that would do the team post search but also takes into account the user that executed the search. This is equivalent to the server REST API endpoint `/teams/{team_id}/posts/search`. 

#### Related PR
Agenda plugin: https://github.com/mattermost/mattermost-plugin-agenda/pull/44

#### Ticket Link
This was required to fix Agenda Plugin issue https://github.com/mattermost/mattermost-plugin-agenda/issues/32


app/plugin_api.go
```diff
@@ -415,6 +415,49 @@ func (api *PluginAPI) SearchPostsInTeam(teamId string, paramsList []*model.Searc
...
+	}
+
+	includeDeletedChannels := false
+	if searchParams.IncludeDeletedChannels != nil {
+		includeDeletedChannels = *searchParams.IncludeDeletedChannels
+	}
+
+	postSearchResult, err := api.app.SearchPostsInTeamForUser(terms, userId, teamId, isOrSearch, includeDeletedChannels, timeZoneOffset, page, perPage)
+
+	if err != nil {
```
lieut-data, line 455: nit: should we just return on the line above?

marianunez: Yes, much better!

app/plugin_api.go
```diff
@@ -415,6 +415,44 @@ func (api *PluginAPI) SearchPostsInTeam(teamId string, paramsList []*model.Searc
 	return postList.ToSlice(), nil
 }
 
+func (api *PluginAPI) SearchPostsInTeamForUser(teamId string, userId string, searchParams *model.SearchParameter) (*model.PostSearchResults, *model.AppError) {
+
```
hanzei, line 419: Nit:
```suggestion
func (api *PluginAPI) SearchPostsInTeamForUser(teamId string, userId string, searchParams *model.SearchParameter) (*model.PostSearchResults, *model.AppError) {
```

app/plugin_api.go
```diff
@@ -415,6 +415,44 @@ func (api *PluginAPI) SearchPostsInTeam(teamId string, paramsList []*model.Searc
...
+	if searchParams == nil {
```
hanzei, line 420: If `nil` is an invalid value for `searchParams`, what value does a pointer provide?

app/plugin_api.go
marianunez: What is the suggested change? I see them the same 🤔

app/plugin_api.go
marianunez: Not sure I follow the question, `searchParams` must have a pointer value to be able to perform the search.

hanzei, line 420: Now the suggestion can be read better:
```suggestion
func (api *PluginAPI) SearchPostsInTeamForUser(teamId string, userId string, searchParams *model.SearchParameter) (*model.PostSearchResults, *model.AppError) {
	if searchParams == nil {
```

app/plugin_api.go
hanzei: See https://github.com/mattermost/mattermost-server/pull/14807#discussion_r440650228

app/plugin_api.go
hanzei: If `searchParams` must not be `nil`, what is the value of using a pointer vs. using a struct directly?

marianunez: Not sure, probably no extra value. 

I set it as a pointer to be consistent with how the api currently handles structs as pointer parameters in the majority of the cases.

hanzei: More recent methods like `KVSetWithOptions` did more away from using pointer all the time. I see here an opportunity to continue to break with a pattern that has not proven valuable.

marianunez: Done!

prapti:
@marianunez Is there a Jira ticket for this PR?

marianunez:
> @marianunez Is there a Jira ticket for this PR?

There is not. This was required to complete the agenda ticket https://github.com/mattermost/mattermost-plugin-agenda/issues/32

marianunez:
@prapti I've updated the agenda ticket to include QA steps.

prapti:
In test queue.

PR #14808, opened 2020-06-13
Migrate all methods from ComplianceStore to return plain errors

### LINKED ISSUE #14639: Migrate AppError to error from store/sqlstore/compliance_store.go
Mattermost is migrating all usages of the `AppError` type to the more idiomatic `error` type. `AppError` is a loaded error type containing fields like HTTP Status code which is not really necessary for errors coming from the DB layer. We should migrate to using normal types satisfying the error interface.

This ticket is about migrating all usages of AppError from `store/sqlstore/compliance_store.go` to return a normal error. The migration should use the error types defined in `store/errors.go`. If the need should arise to add another specific error type, feel free to add it. For any generic error that is not something like an invalid input, or something not found, we can just wrap the error using `errors.Wrap(f)` and return it.

And then in the app layer, we would need to convert the `error` type to the usual `AppError`. Essentially moving the `AppError` creation from the store layer to the app layer. This means we need a switch case to check for the type of error and appropriately convert to the correct AppError. It can be something like:

```go
if _, err := a.Srv().Store.Bot().Update(bot); err != nil {
	var nfErr *store.ErrNotFound
	var appErr *model.AppError
	switch {
	case errors.As(err, &nfErr):
		return model.MakeBotNotFoundError(nfErr.Id)
	case errors.As(err, &appErr): // in case we haven't converted to plain error.
		return appErr
	default: // last fallback in case it doesn't map to an existing app error.
		return model.NewAppError("DeleteBotIconImage", "app.bot.patchbot.internal_error", nil, err.Error(), http.StatusInternalServerError)
	}
}
```

Also note that the error codes need to be changed from being a store level code (`store.<>.edit.error`) to something like `app.<>.edit.internal_error`.

Example Pull Request: #14339

If you have questions about the ticket or you need any help, feel free to contact `agniva.de.sarker` in https://community.mattermost.com/

----------------------------------

If you're interested please comment here and come [join our "Contributors" community channel](https://community.mattermost.com/core/channels/tickets) on our daily build server, where you can discuss questions with community members and the Mattermost core team. For technical advice or questions, please [join our "Developers" community channel](https://community.mattermost.com/core/channels/developers).

JIRA: https://mattermost.atlassian.net/browse/MM-25478

farah:
Summary
Migrate all ComplianceStore methods to return plain errors.

Fixes #14639


store/storetest/mocks/ComplianceStore.go
```diff
@@ -127,19 +127,18 @@ func (_m *ComplianceStore) Save(compliance *model.Compliance) (*model.Compliance
...
-		}
+		r1 = ret.Error(1)
 	}
 
 	return r0, r1
 }
 
 // Update provides a mock function with given fields: compliance
+// Edit
 func (_m *ComplianceStore) Update(compliance *model.Compliance) (*model.Compliance, *model.AppError) {
```
farah, line 142: this should be updated to return plain error i think?

but when I run make store-mocks I get

```
mattermost-server$ make store-mocks 
go get -modfile=go.tools.mod github.com/vektra/mockery/...
/Users/go/src/github.com/farah/mattermost-server/bin/mockery -dir store -all -output store/storetest/mocks -note 'Regenerate this file using `make store-mocks`.'
Error parsing file:  /Users/go/src/github.com/farah/mattermost-server/store/storetest/store.go:60:77: cannot use &s.ComplianceStore (value of type *mocks.ComplianceStore) as store.ComplianceStore value in return statement: wrong type for method Update
Error parsing file:  /Users/go/src/github.com/farah/mattermost-server/store/storetest/store.go:60:77: cannot use &s.ComplianceStore (value of type *mocks.ComplianceStore) as store.ComplianceStore value in return statement: wrong type for method Update
```

store/storetest/compliance_store.go
```diff
@@ -165,40 +165,40 @@ func testComplianceExport(t *testing.T, ss store.Store) {
 	time.Sleep(100 * time.Millisecond)
 
 	cr1 := &model.Compliance{Desc: "test" + model.NewId(), StartAt: o1.CreateAt - 1, EndAt: o2a.CreateAt + 1}
-	cposts, err := ss.Compliance().ComplianceExport(cr1)
-	require.Nil(t, err)
+	cposts, complianceErr := ss.Compliance().ComplianceExport(cr1)
```
farah, line 168: Had to rename this as it was conflicting with the error returned by by `t1, err := ss.Team().Save(t1)` [above](https://github.com/mattermost/mattermost-server/pull/14808/files#diff-9ea510a524d48d6e88c95997e3c8047dL106).

farah:
How can I remove the WIP label?

store/storetest/compliance_store.go
agnivade: We are using `nErr` as the convention here for other PRs. Let's do that here.

agnivade:
@farah - I guess this is ready for review now? Before I dive into review, there is a small comment here: https://github.com/mattermost/mattermost-server/pull/14808#discussion_r440164020 which you might want to take a look at. Thanks :)

jespino:
@farah once the pr is submitted and we star to review it, we prefer to avoid push -f because that way we can't properly track the changes to review, at the end it will end up being squashed and merged into master as one single commit, so there is no point in push or reading.

farah:
> at the end it will end up being squashed and merged into master as one single commit, so there is no point in push or reading.

Ah ok, that makes a lot of sense now.

app/compliance.go
```diff
@@ -16,19 +16,25 @@ func (a *App) GetComplianceReports(page, perPage int) (model.Compliances, *model
 		return nil, model.NewAppError("GetComplianceReports", "ent.compliance.licence_disable.app_error", nil, "", http.StatusNotImplemented)
 	}
 
-	return a.Srv().Store.Compliance().GetAll(page*perPage, perPage)
+	compliance, err := a.Srv().Store.Compliance().GetAll(page*perPage, perPage)
+
```
jespino, line 20: this empty lines looks weird to me, normally that is not the style that we use here. between the execution and the error handling we don't put blank lines.

store/sqlstore/compliance_store.go
```diff
@@ -36,19 +37,19 @@ func newSqlComplianceStore(sqlStore SqlStore) store.ComplianceStore {
...
+		return nil, errors.Wrap(err, "error saving compliance")
 	}
 	return compliance, nil
 }
 
-func (s SqlComplianceStore) Update(compliance *model.Compliance) (*model.Compliance, *model.AppError) {
+func (s SqlComplianceStore) Update(compliance *model.Compliance) (*model.Compliance, error) {
 	if err := compliance.IsValid(); err != nil {
 		return nil, err
 	}
```
jespino, line 55: You are returning a couple of line later a `model.AppError` should be a regular error wrapped with `errors.Wrap`

app/compliance.go
```diff
@@ -16,19 +16,25 @@ func (a *App) GetComplianceReports(page, perPage int) (model.Compliances, *model
...
 	if license := a.Srv().License(); !*a.Config().ComplianceSettings.Enable || license == nil || !*license.Features.Compliance || a.Compliance() == nil {
-		return nil, model.NewAppError("saveComplianceReport", "ent.compliance.licence_disable.app_error", nil, "", http.StatusNotImplemented)
+		return nil, model.NewAppError("saveComplianceReport", "app.compliance.licence_disable.app_error", nil, "", http.StatusNotImplemented)
 	}
 
 	job.Type = model.COMPLIANCE_TYPE_ADHOC
 
 	job, err := a.Srv().Store.Compliance().Save(job)
+
 	if err != nil {
```
jespino, line 36: Save can return different kind of errors, actually it can return a `model.AppError` from the `IsValid` check of the compliance, or a regular error, you should handle the regular errors like you have done here, but if you receive a `model.AppError` you should return it directly (you can see an example of that in the example PR).

store/sqlstore/compliance_store.go
```diff
@@ -36,19 +37,19 @@ func newSqlComplianceStore(sqlStore SqlStore) store.ComplianceStore {
 func (s SqlComplianceStore) createIndexesIfNotExists() {
 }
 
-func (s SqlComplianceStore) Save(compliance *model.Compliance) (*model.Compliance, *model.AppError) {
+func (s SqlComplianceStore) Save(compliance *model.Compliance) (*model.Compliance, error) {
 	compliance.PreSave()
 	if err := compliance.IsValid(); err != nil {
-		return nil, err
+		return nil, errors.Wrap(err, "error validating compliance")
```
jespino, line 43: You shouldn't be wrapping this error, because we need the real underneath data to handle properly this case (this will change in the future, but for now, just `return nil, err`)

agnivade:
Mostly what Jesus said. I will do a fresh pass once these comments are taken care of.

Most important is the error return from `IsValid`. Since that returns an AppError, we need to return that as is, and pass that through the `error` interface. At the call site, we do the same check to cast an nErr to AppError, and if it matches, we return the underlying `AppError`. Please see the other PRs that have went in master.

Thanks.

agnivade:
Hi @farah - just checking if you are still interested in working on this?

agnivade:
Let's go with https://github.com/mattermost/mattermost-server/pull/14969 since this appears to be inactive.

hanzei:
https://github.com/mattermost/mattermost-server/pull/15074 did already fix the ticket. Still, thanks for you contribution @farah.

PR #14822, opened 2020-06-15
[MM-16473] Make plugins' ServerHTTP http.ResponseWriter hijackable

streamer45:
#### Summary

PR implements https://github.com/mattermost/mattermost-server/issues/11288.

#### Example (WebSocket Upgrade)

```golang
package main

import (
	"net/http"

	"github.com/gorilla/websocket"
)

type Plugin struct {
	plugin.MattermostPlugin
}

func (p *Plugin) ServeHTTP(c *plugin.Context, w http.ResponseWriter, r *http.Request) {
	upgrader := websocket.Upgrader{}
	ws, err := upgrader.Upgrade(w, r, nil)
	if err != nil {
		return
	}
	defer ws.Close()

	for {
		mt, msg, err := ws.ReadMessage()
		if err != nil {
			break
		}
		err = ws.WriteMessage(mt, msg)
		if err != nil {
			break
		}
	}
}
```

#### Benchmarks

```
» go test -v -bench=. -benchtime 5s 
goos: linux
goarch: amd64
pkg: github.com/mattermost/mattermost-plugin-starter-template
BenchmarkPluginWSConn512
BenchmarkPluginWSConn512-8    	  200758	    122026 ns/op
BenchmarkPluginWSConn1024
BenchmarkPluginWSConn1024-8   	   67030	    129356 ns/op
BenchmarkPluginWSConn4096
BenchmarkPluginWSConn4096-8   	   67404	    127929 ns/op
BenchmarkPluginWSConn8192
BenchmarkPluginWSConn8192-8   	   62041	    135238 ns/op
PASS
ok  	github.com/mattermost/mattermost-plugin-starter-template	57.806s
```
```golang
package main

import (
	"bytes"
	"crypto/rand"
	"net/url"
	"testing"

	"github.com/gorilla/websocket"
)

func BenchmarkPluginWSConn512(b *testing.B) {
	u := url.URL{Scheme: "ws", Host: "localhost:8065", Path: "/plugins/com.mattermost.plugin-starter-template"}
	c, _, err := websocket.DefaultDialer.Dial(u.String(), nil)
	if err != nil {
		panic(err.Error())
	}
	defer c.Close()

	done := make(chan struct{})

	data := make([]byte, 512)
	if _, err := rand.Read(data); err != nil {
		panic(err.Error())
		return
	}

	received := 0
	go func() {
		defer close(done)
		for {
			_, msg, err := c.ReadMessage()
			if err != nil {
				panic(err.Error())
				return
			}
			if bytes.Compare(data, msg) != 0 {
				panic("bad message received")
			}
			received++
			if received == b.N {
				break
			}
		}
	}()

	b.ResetTimer()
	for i := 0; i < b.N; i++ {
		err := c.WriteMessage(websocket.BinaryMessage, data)
		if err != nil {
			panic(err.Error())
			return
		}
	}
	
	<-done
        b.StopTimer()

	c.WriteMessage(websocket.CloseMessage, websocket.FormatCloseMessage(websocket.CloseNormalClosure, ""))
}
```

#### Ticket

https://mattermost.atlassian.net/browse/MM-16473


plugin/hijack.go
```diff
@@ -0,0 +1,205 @@
...
+	if err := w.client.Call("Plugin.HjConnRWWrite", b, &n); err != nil {
+		return 0, err
+	}
+	return n, nil
+}
+
+func (w *hijackedConn) Read(b []byte) (int, error) {
+	var data []byte
+	if err := w.client.Call("Plugin.HjConnRead", len(b), &data); err != nil {
+		return 0, nil
```
lieut-data, line 154: Intentional to return `nil` for the `error` here?

app/plugin_api_tests/manual.test_http_hijack_plugin/main.go
```diff
@@ -0,0 +1,35 @@
...
+}
+
+func (p *Plugin) ServeHTTP(c *plugin.Context, w http.ResponseWriter, r *http.Request) {
+	hj, ok := w.(http.Hijacker)
+	if !ok {
+		w.WriteHeader(http.StatusInternalServerError)
+		return
+	}
+
+	conn, brw, err := hj.Hijack()
```
lieut-data, line 23: nit: `brw` vs. the example of `bufrw` in the https://golang.org/pkg/net/http/#Hijacker example? No objection to the abbreviating (I was expecting it given the context), but wondering if it would boost readability to align with the docs.

plugin/hijack.go
streamer45: Definitely not. Good catch! :+1:

app/plugin_api_tests/manual.test_http_hijack_plugin/main.go
streamer45: Yes, since it's quite an internal detail I didn't spend too time much on naming. 
We can surely align with the docs though :+1:

plugin/hijack.go
```diff
@@ -0,0 +1,205 @@
...
+		return err
+	}
+
+	w.hjr = &hijackedResponse{
+		conn:    conn,
+		bufrw:   bufrw,
+		readBuf: make([]byte, hijackedConnReadBufSize),
+	}
+	return nil
+}
```
hanzei, line 124: Why is this method needed? It doesn't seam to be called anywhere in the code nor is it needed to fulfill an interface.

streamer45: It's not called directly but through RPC:

https://github.com/mattermost/mattermost-server/blob/cae976643196afd812e106aa477025141fae49ba/plugin/hijack.go#L200-L202

hanzei: Got it, thanks for the explanation :+1:

PR #14823, opened 2020-06-15
[MM-25673] Upgrades and migrations to start using master gorp

ethervoid:
#### Summary

Due to [this change](https://github.com/mattermost/gorp/pull/5) in our Gorp fork, we are now changing the default column type in both MySQL and Postgres. 

One change of this PR is that if a column doesn't have a defined size in MySQL we're creating a `longtext` column which is a BLOG non-binary type and that causes problems if we want to create an index because for those fields is mandatory to specify a size.

The solution is to include all the missing ColMaps with a defined size so we can create indexes without any problem.

This PR also includes:

- Migration code that checks first if the columns are alterable and if not abort the migration to avoid leave the DB in an inconsistent state
- Upgrade the gorp dependency that [includes some fixes too](https://github.com/mattermost/gorp/pull/9).

#### Related to

https://github.com/mattermost/gorp/pull/5

#### Ticket Link

https://mattermost.atlassian.net/browse/MM-25673

#### Linked

https://github.com/mattermost/gorp/pull/9


go.mod
```diff
@@ -60,7 +60,7 @@ require (
 	github.com/magiconair/properties v1.8.1 // indirect
 	github.com/mailru/easyjson v0.7.1 // indirect
 	github.com/mattermost/go-i18n v1.11.0
-	github.com/mattermost/gorp v2.0.1-0.20200527092429-d62b7b9cadfc+incompatible
+	github.com/mattermost/gorp v1.6.2-0.20200526162043-8cc0a796020f
```
ethervoid, line 63: Don't know why master is 1.6.2 but the current branch, which is below master, is 2.0.1 :\

ethervoid:
@lieut-data @jespino @agnivade given that this implies migrations I want to be really sure it's done correctly. I'm going to test different migration scenarios but I would love a first review to check if I missed something importante

scripts/mattermost-postgresql-5.0.sql
```diff
@@ -550,13 +550,13 @@ CREATE TABLE public.teams (
...
     email character varying(128),
-    type text,
+    type character varying(255),
     companyname character varying(64),
     alloweddomains character varying(500),
     inviteid character varying(32),
     allowopeninvite boolean,
     lastteamiconupdate bigint,
-    schemeid text
+    schemeid varying(26)
```
ethervoid, line 559: Given that our kind of Id, shouldn't be a problem. In any case I've put it first in the migration so it fails first

scripts/mattermost-mysql-5.0.sql
```diff
@@ -369,7 +369,7 @@ CREATE TABLE `IncomingWebhooks` (
   `DisplayName` varchar(64) DEFAULT NULL,
   `Description` varchar(128) DEFAULT NULL,
   `Username` varchar(255) DEFAULT NULL,
-  `IconURL` varchar(255) DEFAULT NULL,
+  `IconURL` text,
```
lieut-data, line 372: Changing this file is probably incorrect: it represents the schema of the v5.0 database at the time /that/ code existed, and it's used to verify that the migration /from/ this schema to the current schema is correct. See `scripts/mysql-migration-test.sh` (and same for Postgres).

store/sqlstore/supplier.go
```diff
@@ -911,7 +911,7 @@ func (ss *SqlSupplier) createIndexIfNotExists(indexName string, tableName string
 
 		_, err = ss.GetMaster().ExecNoTimeout("CREATE  " + uniqueStr + fullTextIndex + " INDEX " + indexName + " ON " + tableName + " (" + strings.Join(columnNames, ", ") + ")")
 		if err != nil {
-			mlog.Critical("Failed to create index", mlog.Err(err))
+			mlog.Critical("Failed to create index", mlog.String("table", tableName), mlog.Err(err))
```
lieut-data, line 914: Should we include the index name while we're touching this?

go.mod
lieut-data: The pseudo-version should guarantee the right behaviour here, but agree that the tag prefix is often confusing. It's /usually/ because the tag contains a commit that's newer than the version in question, but I can't find evidence of that here.

Probably good to do a file-by-file comparison with the expected changes in the `vendor` package to be 100% certain you're getting what you want.

scripts/mattermost-mysql-5.0.sql
ethervoid: Oh I see. Reverting...

store/sqlstore/supplier.go
ethervoid: I was going to include it but didn't if it was interesting. Given that looks interesting to include it I'll do :)

go.mod
ethervoid: I see. Ok, I'll review the vendor package to verify is consistent

ethervoid:
One thing I don't like is that with our migration system we can perform partial upgrades and given that MySQL doesn't allow transactional DDL...

go.mod
ethervoid: Checked file by file and everything seems ok with the vendor files

agnivade: As discussed offline, should we just make a v2 of gorp to align the tag version and go.mod version?

store/sqlstore/upgrade.go
```diff
@@ -803,3 +808,81 @@ func upgradeDatabaseToVersion524(sqlStore SqlStore) {
...
+	sqlStore.AlterColumnTypeIfExists("Teams", "SchemeId", "VARCHAR(26)", "VARCHAR(26)")
+	sqlStore.AlterColumnTypeIfExists("IncomingWebhooks", "Username", "varchar(255)", "varchar(255)")
+	sqlStore.AlterColumnTypeIfExists("IncomingWebhooks", "IconURL", "text", "varchar(1024)")
+
+	saveSchemaVersion(sqlStore, VERSION_5_26_0)
+	//}
+}
+
+func precheckMigrationToVersion526(sqlStore SqlStore) error {
+	teamsQuery, _, err := sqlStore.getQueryBuilder().Select(`CASE
```
ethervoid, line 836: This could be done by the `AlterColumnTypeIfExists` but that would make some changes in the function that we should not tackle here.

store/sqlstore/upgrade.go
```diff
@@ -803,3 +808,81 @@ func upgradeDatabaseToVersion524(sqlStore SqlStore) {
...
+
+func upgradeDatabaseToVersion525(sqlStore SqlStore) {
+	//if shouldPerformUpgrade(sqlStore, VERSION_5_24_0, VERSION_5_25_0) {
+	saveSchemaVersion(sqlStore, VERSION_5_25_0)
+	//}
+}
+
+func upgradeDatabaseToVersion526(sqlStore SqlStore) {
+	//if shouldPerformUpgrade(sqlStore, VERSION_5_25_0, VERSION_5_26_0) {
+	err := precheckMigrationToVersion526(sqlStore)
```
agnivade, line 820: Isn't this specific to the changes below? I am wondering whether it will still be applicable if more changes are added here?

We are considering this entire function to be atomic, but will that be always the case? Otherwise it might make sense to only skip adding the 4 column changes rather than returning.

go.sum
```diff
@@ -392,6 +392,7 @@ github.com/lann/ps v0.0.0-20150810152359-62de8c46ede0 h1:P6pPBnrTSX3DEVR4fDembhR
 github.com/lann/ps v0.0.0-20150810152359-62de8c46ede0/go.mod h1:vmVJ0l/dxyfGW6FmdpVm2joNMFikkuWg0EoCKLGUMNw=
 github.com/lib/pq v1.0.0 h1:X5PMW56eZitiTeO7tKzZxFCSpbFZJtkMMooicw2us9A=
 github.com/lib/pq v1.0.0/go.mod h1:5WUZQaWbwv1U+lTReE5YruASi9Al49XbQIvNi/34Woo=
+github.com/lib/pq v1.3.0/go.mod h1:5WUZQaWbwv1U+lTReE5YruASi9Al49XbQIvNi/34Woo=
```
agnivade, line 395: Run a go mod tidy please. All these changes will go away.

store/sqlstore/upgrade.go
ethervoid: Yes that function is performing specific checks for the changes below. If more changes are included And those should be checked before executing then we should add more checks in that function.

We usually don’t need to do this part but I’m the AlterColumn case if one of the fails we leave the database screens inconsistent.

In my opinion the migration should be all or nothing so if one of the checks fails e we should abort the entire migration

agnivade: Sounds good to me.

store/sqlstore/upgrade.go
```diff
@@ -803,3 +808,81 @@ func upgradeDatabaseToVersion524(sqlStore SqlStore) {
...
+	if err != nil {
+		mlog.Error("Error upgrading DB schema to 5.26.0", mlog.Err(err))
+		os.Exit(EXIT_GENERIC_FAILURE)
+	}
```
agnivade, line 824: I am channeling my inner @streamer45 here. We can shorten this to the `if err := fn(); err != nil {}` format. :)

go.mod
ethervoid: We could talk about this after the merge but is possible that we need to do it for our own fork

store/sqlstore/upgrade.go
```diff
@@ -818,8 +819,73 @@ func upgradeDatabaseToVersion525(sqlStore SqlStore) {
...
+				ELSE TRUE
+			END as iconurlok`).
+		From("IncomingWebhooks").ToSql()
+	if err != nil {
+		return err
+	}
+
+	var schemeIdOk bool
+	var typeOk bool
+	row := sqlStore.GetMaster().Db.QueryRow(teamsQuery)
```
lieut-data, line 866: Does this only check the first row of the result? Do we need to check every row?

agnivade:
There are no plans to migrate off gorp as of yet. They need to be discussed and approved first.

store/sqlstore/upgrade.go
ethervoid: Yes you’re absolutely right! I wanted to aggregate but I made the mistake of think about the select clause as an aggregation one 🤦. I’ll change it to an aggregate one, thank you!

ethervoid:
@lieut-data Yes I was wondering the same when I started working on this and that's why asked about the state of the gorp removal feature but is not even defined and given that we have master unstable and also we need to fix [another bug](https://mattermost.atlassian.net/browse/MM-26220) pending to fix I have two options:

- Revert the changes
- Stabilize gorp and do the migration

I've chosen to stabilize and execute the migration because is a low-risk migration and improves our current database schema which is going to be a long-lasting improvement and not just for gorp. In case of problems I've thought about how to fix those possible problems:

- Teams and IncomingWebHooks should have a small number of rows
- `Teams:SchemeId` should be a 26 character field, because is one of our UUID, and if not we need to fix that. In this case, the answer will be to recreate the scheme and that should create a 26 character Id instead of the buggy one.
- `Teams:Type` should be a 1 character and in case any user has another data here shouldn't be larger than 255 characters, and that part only affected Postgresql which has a text field. In that case, if possible, we need to tackle that individually.
- `IncomingWebhooks:Username` here we have a 255 limit, we didn't have it in Postgres, in case a Postgres user has a username larger than 255 in an incoming webhook we should ask them to recreate the webhook or find a better solution
- `IncomingWebhooks:IconURL` here we introduced a 1024 limit for Postgres users. In case we get a problem here the easy solution is to add a new URL that complies with the limit.

In the worst-case scenario, we're able to recreate the Scheme or the Webhook to fix the problem but I don't think that is going to happen although I'm ready to get some surprises.

jespino:
I'm a bit worry about the migration failing and how the user have to deal with that. I can imagine the scenario of a customer finding a upgrade window, shutting down the service, getting the new version, deploying it, trying to upgrade, and suddenly they have a problem with the migration that they don't have any knowledge about how to solve it. My concern is about how often it is going to happen.

ethervoid:
@jespino 

> My concern is about how often it is going to happen

Sincerely we don't know. In this particular case, it shouldn't happen because we've included long enough limits to avoid any kind of problem. The only one that should be restrictive is the Id one because in that case there is an integrity problem.

> They have a problem with the migration that they don't have any knowledge about how to solve it

I agree, but that's better than having an inconsistent scenario. Also, if they're customers they have support for it.

I see your point but we have to deal with this kind of migrations given that almost all the deployments are external and we don't control them.

I'll happily revert the gorp changes and abort the migration part if enough people are concerned about this but we're going to face this kind of migrations, at least the `SchemeId` one in the future.

store/sqlstore/upgrade.go
```diff
@@ -818,8 +819,73 @@ func upgradeDatabaseToVersion525(sqlStore SqlStore) {
 func upgradeDatabaseToVersion526(sqlStore SqlStore) {
 	// TODO: uncomment when the time arrive to upgrade the DB for 5.26
 	//if shouldPerformUpgrade(sqlStore, VERSION_5_25_0, VERSION_5_26_0) {
+	if err := precheckMigrationToVersion526(sqlStore); err != nil {
```
lieut-data, line 822: Thinking about this some more, I'm assuming the precheck is to prevent MySQL's default (!) behaviour of silently truncating a field if it's too long.

From reading the docs on [ALTER TABLE](https://dev.mysql.com/doc/refman/5.7/en/alter-table.html), it looks like we could get this checking "for free" with strict mode:
> This conversion may result in alteration of data. For example, if you shorten a string column, values may be truncated. To prevent the operation from succeeding if conversions to the new data type would result in loss of data, enable strict SQL mode before using ALTER TABLE (see Section 5.1.10, “Server SQL Modes”). 

Would it be worth pivoting to that instead of trying to do our own check? I think this also gets us a transactional guarantee, as compared to checking ourselves and risking a (incredibly unlikely) race from a concurrent server writing.

(I'm also assuming Postgres "does the right thing" here.)

ethervoid: The problem with that we could end with the DB  partially upgraded which is not something we want. 

Eg We execute the migration and  while the first two alters work the third could fail, you will get an error in Strict mode but at that time you’ve altered two columns, because MySQL does auto commit and doesn’t support DDL transaction.

The idea of this precheck is to test all the affected columns and not begin the migration if some of them is going to fail.

lieut-data: @ethervoid, agree that we can't do the schema change transactionally across multiple columns. Admittedly, this has been a problem for every migration we've ever done. Is there something special about this case that warrants more scrutiny? My hope is that this is still a backwards compatible change in that `VERSION-1` can still read/write to these altered column types.

ethervoid: > Is there something special about this case that warrants more scrutiny?

This is a multiple alter situation and we should try to do it atomically. Why? as you said if we left this inconsistent for some users shouldn't be a problem because it should be compatible with version-1. But think about what happens if we revert the change, we don't have downgrade script, and in the future, this inconsistency is breaking something.

From my PoV, all the MM users must have the same database schema to avoid problems doing migrations in Vanilla MM.

>  I think this also gets us a transactional guarantee, as compared to checking ourselves and risking an (incredibly unlikely) race from a concurrent server writing

I thought about it and I think a good midterm solution is to include precheck and also use the strict SQL mode

What do you think?

lieut-data: > From my PoV, all the MM users must have the same database schema to avoid problems doing migrations in Vanilla MM.

Within minor versions, we actually explicitly support backwards compatibility: it's the reason why `community-daily` / `community-release` and `community` all run against the same database without issue. The only real requirement is that the migration is idempotent, so that if it does fail midway, it can be restarted from the beginning without issue.

> I thought about it and I think a good midterm solution is to include precheck and also use the strict SQL mode.

This seems reasonable! :)

ethervoid: Included the new check. I've only add it for this migration but if we don't get any problems with it we can apply to all the migrations

store/sqlstore/upgrade.go
```diff
@@ -889,3 +895,31 @@ func precheckMigrationToVersion526(sqlStore SqlStore) error {
...
+			newSqlMode := currentSqlMode + "," + MYSQL_STRICT_SQL_MODE
+			_, err = sqlStore.GetMaster().Exec("SET @@SESSION.sql_mode = :Mode",
+				map[string]interface{}{"Mode": newSqlMode})
+			if err != nil {
+				mlog.Error("Error setting sql_mode", mlog.Err(err))
+			}
+		} else if err != nil {
+			mlog.Error("Error getting sql_mode", mlog.Err(err))
+		}
+		defer func() {
```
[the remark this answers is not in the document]
ethervoid: Yes, you're right :\

store/sqlstore/upgrade.go
```diff
@@ -889,3 +895,31 @@ func precheckMigrationToVersion526(sqlStore SqlStore) error {
...
+		// If we can retrieve the sql_mode variable we continue
+		if err == nil && !strings.Contains(currentSqlMode, MYSQL_STRICT_SQL_MODE) {
+			newSqlMode := currentSqlMode + "," + MYSQL_STRICT_SQL_MODE
+			_, err = sqlStore.GetMaster().Exec("SET @@SESSION.sql_mode = :Mode",
+				map[string]interface{}{"Mode": newSqlMode})
+			if err != nil {
+				mlog.Error("Error setting sql_mode", mlog.Err(err))
+			}
+		} else if err != nil {
+			mlog.Error("Error getting sql_mode", mlog.Err(err))
```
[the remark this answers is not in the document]
ethervoid: Yes that's the idea to not block if we're not able to change the mod

store/sqlstore/upgrade.go
```diff
@@ -889,3 +895,31 @@ func precheckMigrationToVersion526(sqlStore SqlStore) error {
...
+		}
+		defer func() {
+			// If fails it will be restored when session ends because
```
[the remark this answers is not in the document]
ethervoid: Yes, I thought about the session and the connection pool, I got some problems and though they're caused by that. That could be fixed using the `@@GLOBAL` but that needs admin permissions for the DB user.

My intention was to provide a rough example of the code so we can discuss it and I have the same feeling about this. Don't like to mess with internal modes and I'll not use this change.

And yes, let's convince Cloud to use Postgres hahaha

ethervoid:
@lieut-data I want to thank you for the `sql_mode` suggestion, that way I've learned about it and also I came up with a suggestion about it and I want to know your opinion. 

What about if we do a check on server start checking for the strict SQL mode and show a warning for those that don't have it activated, versions 5.7.4 and below. That way we could help our users and ourselves to have a more reliable database for upgrades.

ethervoid:
@lieut-data :+1: , the idea was to include the warning in other PR. I'll share this in the server meeting. Thanks!

ethervoid:
Adding @thefactremains and @grundleborg because this migration/change has the potential to make the upgrade not possible to the customers. Shouldn't happen, and there are some mitigation measures if that happens, but we should be cautious.

For PM review we should take into account that we're reducing the size of some fields for both databases:

### Teams table

- **Type** column changes from text to `character varying (255)` in Postgres. MySQL keeps the same size
- **SchemeId** changes to 26 characters from `text` in Postgres and `varchar(255)` in MySQL

### IncomingWebhooks

- **Username** changes from `text` to `character vayring(255)` in Postgres and remains equally for MySQL
- **IconURL** changes from `text` to `character varying(1024)` in Postgres and keeps the same for MySQL

srkgupta:
The entire Cypress test suite was triggered on this PR to verify if there were any regressions and there were no regressions found. Here is the [link to the report](https://mm-cypress-report.s3.amazonaws.com/535-server-pr-14823-d1331c6/mochawesome.html)

ethervoid:
Thank you very much @srkgupta

thefactremains:
> Adding @thefactremains and @grundleborg because this migration/change has the potential to make the upgrade not possible to the customers. Shouldn't happen, and there are some mitigation measures if that happens, but we should be cautious.

Can you please help me understand the risk level here? It seems like we should know for sure if this change will prevent upgrades before we decide to do it. Maybe I'm missing something?

ethervoid:
> Can you please help me understand the risk level here?

The main risk in this change would be that upgrade could fail and the user is not able to upgrade to a new version until they fix the problem. There are two possible fixes:

- Problems with the Teams Scheme Id that would be fixed recreating the permissions schema. If that happens is because at some point we allowed to created UUIDs larger than 26 characters but that seems very improbable
- Problems with the IncomingWebhooks Username and Icon URLs. In this case, the solution is to make the username or icon url fit into the new size of the columns (255 and 1024 characters respectively)

> It seems like we should know for sure if this change will prevent upgrades before we decide to do it.

We're pretty sure that almost all the users are going to be able to upgrade to the new version but we're not able to check all the databases and we could find users that have a username larger than 255 characters, etc. So there is a risk to upgrade, as always.

Another issue to take into account is that we're changing the size of those fields and that should be reviewed and be taken into account by a PM.

ethervoid:
@srkgupta Thank you very much for that amazing work!

store/sqlstore/team_store.go
```diff
@@ -209,10 +209,12 @@ func newSqlTeamStore(sqlStore SqlStore) store.TeamStore {
 		table.ColMap("DisplayName").SetMaxSize(64)
 		table.ColMap("Name").SetMaxSize(64).SetUnique(true)
 		table.ColMap("Description").SetMaxSize(255)
+		table.ColMap("Type").SetMaxSize(255)
```
grundleborg, line 212: Why 255?

store/sqlstore/team_store.go
```diff
@@ -209,10 +209,12 @@ func newSqlTeamStore(sqlStore SqlStore) store.TeamStore {
...
 		table.ColMap("Email").SetMaxSize(128)
 		table.ColMap("CompanyName").SetMaxSize(64)
 		table.ColMap("AllowedDomains").SetMaxSize(1000)
 		table.ColMap("InviteId").SetMaxSize(32)
+		table.ColMap("SchemeId").SetMaxSize(26)
```
grundleborg, line 217: This makes sense. Our IDs are always 26 chars anyway.

store/sqlstore/webhook_store.go
```diff
@@ -35,6 +35,8 @@ func newSqlWebhookStore(sqlStore SqlStore, metrics einterfaces.MetricsInterface)
 		table.ColMap("TeamId").SetMaxSize(26)
 		table.ColMap("DisplayName").SetMaxSize(64)
 		table.ColMap("Description").SetMaxSize(500)
+		table.ColMap("Username").SetMaxSize(255)
```
grundleborg, line 38: Why 255?

store/sqlstore/webhook_store.go
```diff
@@ -35,6 +35,8 @@ func newSqlWebhookStore(sqlStore SqlStore, metrics einterfaces.MetricsInterface)
...
+		table.ColMap("IconURL").SetMaxSize(1024)
```
grundleborg, line 39: Why 1024? What is the length of this field currently? Will 1024 be enough?

store/sqlstore/team_store.go
ethervoid: 255 is the limit [we have in gorp](https://github.com/mattermost/gorp/pull/5/files#diff-da7e85db8d7b43216497bfee21f74cd2R89) for MySQL before converting the field into text which is not recommended to use as an index

store/sqlstore/webhook_store.go
ethervoid: Same as before

store/sqlstore/webhook_store.go
ethervoid: Well right now that field in MySQL is 255 characters and in Postgres is a text field. The problematic situation could be in Postgres. We can remove the `MaxiSize` and that would convert those fields in text ones.

amyblais:
Today is code complete and it sounds that this is only on master so removing cherry-pick label. Let me know if I'm wrong,

ethervoid:
@amyblais given that it needs another review approval, I'd move it to 5.28 because this is not critical

agnivade:
Gentle ping @grundleborg :) It would be good to merge this before more merge conflicts happen.

store/sqlstore/webhook_store.go
grundleborg: I think we should do that. My worry is this breaks the upgrade for customers if they have a long one already in the database with Postgres.

store/sqlstore/team_store.go
grundleborg: :+1:

store/sqlstore/webhook_store.go
grundleborg: :+1:

store/sqlstore/webhook_store.go
ethervoid: @grundleborg 

I've found that we're enforcing 1024 for IconURL in our webhook models, both in the [incoming](https://github.com/mattermost/mattermost-server/blob/master/model/incoming_webhook.go#L101-L107) and the [outgoing](https://github.com/mattermost/mattermost-server/blob/master/model/outgoing_webhook.go#L189-L195), so no one should have an IconURL larger than 1024 characters. It also checks that the Username is smaller than 65 characters.

So I think is not needed to include a text field. What do you think?

grundleborg: Yeah in that case I'm fine with it like this.

ethervoid:
@agnivade  one last thing I want to check before I merge this PR. Is the go.mod,sum and vendor/module.txt looks fine?

agnivade:
Yeah, they look good.

amyblais:
@ethervoid Is this a breaking change for v5.28?

ethervoid:
@amyblais Nope should migrate the database without a problem. The only thing that could happen is the user gets an error if they have any corner case and we don't allow to migrate but is practically impossible

PR #14825, opened 2020-06-16
Migration of AuditStore to return plain errors

rvillablanca:
#### Summary

Continuing with the migration to plain errors in store layer campaign, this time is for `AuditStore` by migrating `Save`, `Get` and `PermanentDeleteByUser`.

#### Ticket Link

Fixes https://github.com/mattermost/mattermost-server/issues/14819

Jira ticket: https://mattermost.atlassian.net/browse/MM-26032


store/sqlstore/audit_store.go
```diff
@@ -36,19 +35,20 @@ func (s SqlAuditStore) createIndexesIfNotExists() {
...
+		return errors.Wrapf(err, "failed to save Audit with userId=%s and action=%s", audit.UserId, audit.Action)
 	}
 	return nil
 }
 
-func (s SqlAuditStore) Get(user_id string, offset int, limit int) (model.Audits, *model.AppError) {
+func (s SqlAuditStore) Get(userId string, offset int, limit int) (model.Audits, error) {
 	if limit > 1000 {
-		return nil, model.NewAppError("SqlAuditStore.Get", "store.sql_audit.get.limit.app_error", nil, "user_id="+user_id, http.StatusBadRequest)
+		// TODO: is it appropiate to set field to limit ?
```
agnivade, line 50: Perhaps a new error is in order?

app/audit.go
```diff
@@ -27,11 +30,31 @@ var (
 )
 
 func (a *App) GetAudits(userId string, limit int) (model.Audits, *model.AppError) {
-	return a.Srv().Store.Audit().Get(userId, 0, limit)
+	audits, err := a.Srv().Store.Audit().Get(userId, 0, limit)
+	if err != nil {
+		var iErr *store.ErrInvalidInput
```
agnivade, line 35: Weren't these written as `invErr`?

store/sqlstore/audit_store.go
rvillablanca: Yep, maybe a new error type would be appropiate for those cases

app/audit.go
rvillablanca: Not sure,let me see

rvillablanca: You are right :), I'll make the changes

store/sqlstore/audit_store.go
agnivade: @jespino - what do you think?

rvillablanca: @agnivade @jespino Friendly ping reminder

jespino: I agree that a different error can be really valuable here. The plain `ErrBadRequest` we can call it `ErrInvalidParams` but follow the `Detail` field approach if we want. But I think I prefer the approach with `extra data` for example, the `input field` and the `input value`, and maybe, a `Detail` too. Provide to the upper level a sensible error that give all the information structured is the best way to relay in the app layer to make the correct decisions (for example, the userId passed, probably is available in the context where the error must be handled). So, I think the invalid param is the most important one (I'm not sure if we need the value at all). the `details` can be built from the ParamName if we want.

Ok, my proposal is, lets go for `ErrInvalidParameter` with `ParamName` field, and the `Error` function is going to build something like `Invalid param "%s" passed to the function`.

What do you think @agnivade and @rvillablanca?

rvillablanca: Yep, I agree it is valuable to have error with fields giving more context about the problem but there are cases like this when the only one error available is `ErrInvalidInput` but it is coupled to the any entity model (I think), so we can create another one as suggested with other fields and call it `ErrInvalidParams` and that is not the problem. What I think would be valuable is to have a more generic error for cases where `ErrInvalidInput` and `ErrInvalidParams` are not suitable, maybe because the invalid input is not a entity field nor a generic parameter but still being an invalid input. 

This can be extrapolated to any other error, it is more a generic concern and not related only with an invalid input. We need a generic type like `error` but slighly different so we can assert with it and decide if the response should be a bad request, an internal rrror or a not found resource.

Anyway, it is just a suggestion. We can have many kinds of specific errors with different field on them, or have just only a few specific errors plus other ones with `string` detail being different from plain `error` interface.

agnivade: Yes, `ErrInvalidInput` is tied to an invalid value of an entity model. So that definitely does not fit here.

However, before adding a more generic error to capture any invalid input not tied to an entity model, I would really like to see some more practical examples of it in our codebase to make a proper decision.

@rvillablanca - you have converted a lot of these errors and probably have the most experience. Would this be the first time you are seeing an invalid input which is not tied to a domain model? The doubts raised in https://github.com/mattermost/mattermost-server/pull/14836 would again fall into `ErrInvalidInput` category because that's related to the role and scheme respectively.

Do you have any more examples like these?

store/sqlstore/audit_store.go
```diff
@@ -58,26 +58,26 @@ func (s SqlAuditStore) Get(user_id string, offset int, limit int) (model.Audits,
...
-	if len(user_id) != 0 {
-		query = query.Where(sq.Eq{"UserId": user_id})
+	if len(userId) != 0 {
+		query = query.Where(sq.Eq{"UserId": userId})
 	}
 
 	queryString, args, err := query.ToSql()
 	if err != nil {
-		return nil, model.NewAppError("SqlAuditStore.Get", "store.sql_audit.app_error", nil, err.Error(), http.StatusInternalServerError)
+		return nil, errors.Wrapf(err, "failed to create query to get Audit with userId=%s", userId)
```
agnivade, line 67: I did a check for errors from `ToSql` and we are not returning the actual value with the error. I would like to be consistent here too.

Here are the errors for ToSql:
```
channel_store.go-		return nil, errors.Wrap(err, "failed to create query")
channel_store.go-		return 0, errors.Wrap(err, "failed to create query")
channel_store.go-		return nil, errors.Wrap(err, "channel_members_tosql")
```

Maybe use `audits_tosql`? It's fine to keep the others like that as long as we later make a full sweep to make all of these consistent.

store/sqlstore/audit_store.go
rvillablanca: No, I've not seen more other examples for these cases until now. Another case could be `ErrNotFound` where it assumes the search was by id, so when I have a case when some entity was not found but it was searched by multiple criteria then I have to use `fmt.Sprintf`. 

Of course it is not a problem but the point is the more specific the error is, the more difficult is to fill all fields. But I understand your point, maybe we must wait for more cases like these and take the right decision and until now those are the only cases that I've found.

agnivade: Ok, sounds good then. Let's go with `ErrOutOfBounds` with the field name and value for now. If we see more cases, then we can decide on what the right action will be.

Re: `ErrNotFound` great point on that. I think what we can do is make the `id` field to be variadic args of key-value pairs so that multiple fields can be passed. Feel free to send a PR fixing all those `fmt.Sprintf` cases.

store/errors.go
```diff
@@ -88,3 +90,21 @@ func NewErrNotFound(resource, id string) *ErrNotFound {
...
+
+func (e *ErrOutOfBounds) Error() string {
+	msg := "invalid parameter"
+	if len(e.Args) > 1 {
+		msg = "invalid parameters"
+	}
+	return fmt.Sprintf("%s resource %s: %v", msg, e.Entity, e.Args)
+}
+
+func NewErrOutOfBound(entity string, args model.StringInterface) *ErrOutOfBounds {
```
agnivade, line 108: `NewErrOutOfBounds`?

store/errors.go
```diff
@@ -88,3 +90,21 @@ func NewErrNotFound(resource, id string) *ErrNotFound {
 func (e *ErrNotFound) Error() string {
 	return "resource: " + e.resource + " id: " + e.Id
 }
+
+// ErrOutOfBounds indicates that an invalid parameter, usually not related with the entity, was provided
+type ErrOutOfBounds struct {
+	Entity string
+	Args   model.StringInterface
```
agnivade, line 97: Is there a reason for these to be exported? Doesn't look like we are accessing these fields from outside.

store/errors.go
```diff
@@ -88,3 +90,21 @@ func NewErrNotFound(resource, id string) *ErrNotFound {
...
+}
+
+func (e *ErrOutOfBounds) Error() string {
+	msg := "invalid parameter"
+	if len(e.Args) > 1 {
+		msg = "invalid parameters"
```
agnivade, line 103: We can just do `msg += "s"`. Or personally I would just write "invalid parameter(s)". But upto you on whichever you prefer.

store/errors.go
```diff
@@ -88,3 +90,21 @@ func NewErrNotFound(resource, id string) *ErrNotFound {
 func (e *ErrNotFound) Error() string {
 	return "resource: " + e.resource + " id: " + e.Id
 }
+
+// ErrOutOfBounds indicates that an invalid parameter, usually not related with the entity, was provided
```
agnivade, line 94: I would specifically say what it exactly is: an error when the total number of rows requested was beyond limits. It's not just any parameter.

Currently the description equates it to a generic invalid error not attached to an entity, which it's not.

rvillablanca: So I did not understand well, I thought it will be a generic error for invalid parameters

store/errors.go
rvillablanca: Yep, but other errors here also have exported fields so I wanted to be consistent

store/errors.go
rvillablanca: Ok

store/errors.go
rvillablanca: Will fix it

store/errors.go
```diff
@@ -88,3 +88,16 @@ func NewErrNotFound(resource, id string) *ErrNotFound {
...
+// ErrOutOfBounds indicates that the requested total numbers of rows was greater than allowed
+type ErrOutOfBounds struct {
+	value int
+}
+
+func (e *ErrOutOfBounds) Error() string {
+	return fmt.Sprintf("invalid limit parameter: %d", e.value)
+}
+
+func NewErrOutOfBound(value int) *ErrOutOfBounds {
```
agnivade, line 101: This still does not appear to be fixed.

store/errors.go
```diff
@@ -88,3 +88,16 @@ func NewErrNotFound(resource, id string) *ErrNotFound {
 func (e *ErrNotFound) Error() string {
 	return "resource: " + e.resource + " id: " + e.Id
 }
+
+// ErrOutOfBounds indicates that the requested total numbers of rows was greater than allowed
```
agnivade, line 92: "... was greater than the allowed limit".

rvillablanca: yep, my english

store/errors.go
rvillablanca: Will fix

agnivade:
@lindy65 - This requires testing loading of audits.

agnivade:
We do have negative unit tests for some store methods. But not in audit store. I agree that it could use some negative unit tests. But that is not really in the scope of this campaign. This campaign is just about converting `AppError` to `error`.

PR #14836, opened 2020-06-17
Migration of SchemeStore to return plain errors

rvillablanca:
#### Summary

Continuing with the campaing to migrate store layer to return plain errors, this PR migrates completely `SchemeStore`.

#### Ticket Link

Fixes https://github.com/mattermost/mattermost-server/issues/14815


store/sqlstore/role_store.go
```diff
@@ -128,10 +130,11 @@ func (s *SqlRoleStore) Save(role *model.Role) (*model.Role, *model.AppError) {
 	return dbRole.ToModel(), nil
 }
 
-func (s *SqlRoleStore) createRole(role *model.Role, transaction *gorp.Transaction) (*model.Role, *model.AppError) {
+func (s *SqlRoleStore) createRole(role *model.Role, transaction *gorp.Transaction) (*model.Role, error) {
 	// Check the role is valid before proceeding.
 	if !role.IsValidWithoutId() {
-		return nil, model.NewAppError("SqlRoleStore.Save", "store.sql_role.save.invalid_role.app_error", nil, "", http.StatusBadRequest)
+		// TODO: Here another type of error would be better, waiting for suggestions in review.
```
jespino, line 136: I think this is the proper error to show here, but @agnivade is the expert in this.

store/sqlstore/scheme_store.go
```diff
@@ -46,48 +46,49 @@ func (s SqlSchemeStore) createIndexesIfNotExists() {
...
 		if err := transaction.Commit(); err != nil {
-			return nil, model.NewAppError("SqlSchemeStore.SchemeSave", "store.sql_scheme.save_scheme.commit_transaction.app_error", nil, err.Error(), http.StatusInternalServerError)
+			return nil, errors.Wrap(err, "commit_transaction")
 		}
 		return newScheme, nil
 	}
 
 	if !scheme.IsValid() {
-		return nil, model.NewAppError("SqlSchemeStore.Save", "store.sql_scheme.save.invalid_scheme.app_error", nil, "schemeId="+scheme.Id, http.StatusBadRequest)
+		// TODO: Here another type of error would be better, waiting for suggestions in review.
```
jespino, line 68: Same here, I think is the right error, but maybe @agnivade have a different opinion.

store/sqlstore/role_store.go
```diff
@@ -128,10 +130,11 @@ func (s *SqlRoleStore) Save(role *model.Role) (*model.Role, *model.AppError) {
 	return dbRole.ToModel(), nil
 }
 
-func (s *SqlRoleStore) createRole(role *model.Role, transaction *gorp.Transaction) (*model.Role, *model.AppError) {
+func (s *SqlRoleStore) createRole(role *model.Role, transaction *gorp.Transaction) (*model.Role, error) {
```
agnivade, line 133: It is not clear to me why we are migrating methods from role_store. I thought this PR was about scheme_store?

store/sqlstore/scheme_store.go
```diff
@@ -108,7 +109,7 @@ func (s *SqlSchemeStore) createScheme(scheme *model.Scheme, transaction *gorp.Tr
 	}
 
 	if len(defaultRoles) != 6 {
-		return nil, model.NewAppError("SqlSchemeStore.SaveScheme", "store.sql_scheme.save.retrieve_default_scheme_roles.app_error", nil, "", http.StatusInternalServerError)
+		return nil, errors.New("creatScheme: unable to retrieve default scheme roles")
```
agnivade, line 112: createScheme

store/sqlstore/scheme_store.go
agnivade: Yes, this is fine. The entire scheme is invalid.

store/sqlstore/scheme_store.go
```diff
@@ -241,51 +242,52 @@ func filterModerated(permissions []string) []string {
...
 		}
-		return nil, model.NewAppError("SqlSchemeStore.GetByName", "store.sql_scheme.get.app_error", nil, err.Error(), http.StatusInternalServerError)
+		return nil, errors.Wrapf(err, "failed to get Scheme with schemeName=%s", schemeName)
 	}
 
 	return &scheme, nil
 }
 
-func (s *SqlSchemeStore) Delete(schemeId string) (*model.Scheme, *model.AppError) {
+// Question: this method touches a lot of tables and it is not transactional
```
agnivade, line 270: Yes, probably needs to be a transaction. A PR would be most welcome. :)

store/sqlstore/role_store.go
rvillablanca: Yep, but `SchemeStore` uses `RoleStore` when creating a new schema is required to create multiples roles, but the good part is that migrated method from `RoleStore` is internal and used mostly by `SchemeStore` but just once by itself, that is why I decided to migrate this method. You can see in the changes that their impact is only internal and the API of `RoleStore` was not changed :)

agnivade: Got it. Thanks for the clarification!

agnivade:
@srkgupta - this needs all scheme/permission related tests. Thanks.

PR #14838, opened 2020-06-17
MM-26206: Add GroupMentions permissions in default channel admin

agnivade:
This permission was skipped from the default permissions function.
Which led to a bug where permissions reset would not bring back
this permission to the different roles.

#### Ticket link

https://mattermost.atlassian.net/browse/MM-26206


agnivade:
I'll reuse the test in https://github.com/mattermost/mattermost-server/pull/14768, so waiting for that. But the code can be reviewed.

@mkraft / @fmunshi - Just wanted to confirm that the group_mention is good to be at a channel_admin level? Or would you prefer it somewhere else? From the system console, I saw that it was grayed out from channel admin level, and assumed this is the right place.

Thanks.

mkraft:
> I'll reuse the test in #14768, so waiting for that. But the code can be reviewed.
> 
> @mkraft / @fmunshi - Just wanted to confirm that the group_mention is good to be at a channel_admin level? Or would you prefer it somewhere else? From the system console, I saw that it was grayed out from channel admin level, and assumed this is the right place.
> 
> Thanks.

No objection from me, but I would defer to @fmunshi since I wasn't directly involved in this feature.

agnivade:
I have no issues. @amyblais - can this be backported to 5.25?

PR #14840, opened 2020-06-17
MM-26203: Fix TestChannelStore//SearchAllChannels flake

agnivade:
Using random strings for channel names creates chances of collisions
when searching for specific strings.

We increase the search space increasing the size of the search string
to be searched. This reduces collisions and makes the tests more reliable.

For the "off-" family of tests, this probably needs more length. But
it would also need the tests to be changed slightly as the "off-" part
acts as a prefix. I accidentally stumbled upon it while running the test
50 times. Adding a "-" is probably good enough for now. We can revisit
if it fails again in CI.

#### Ticket link

https://mattermost.atlassian.net/browse/MM-26203

### DIFF
--- a/store/storetest/channel_store.go
+++ b/store/storetest/channel_store.go
@@ -5242,7 +5242,7 @@ func testChannelStoreSearchAllChannels(t *testing.T, ss store.Store) {
 
 	o14 := model.Channel{
 		TeamId:      t2.Id,
-		DisplayName: "FOOBAR",
+		DisplayName: "FOOBARDISPLAYNAME",
 		Name:        "whatever",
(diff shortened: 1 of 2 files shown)


agnivade:
@lindy65 - No testing needed here as it just fixes a flaky test. Thanks.

agnivade:
Sorry, I actually didn't see the label. @lindy65 - any reason this is slated for 5.25? This just fixes a flaky test and not related to the code in any way.

amyblais:
Per Agniva's comment, this seems fine for v5.26.

lindy65:
Apologies @agnivade! An error on my part!

PR #14841, opened 2020-06-17
Include response log trace

ethervoid:
#### Summary

Added response traces to our logs

#### Ticket Link

https://mattermost.atlassian.net/browse/MM-25296


ethervoid:
Changed because go vet is complaining.

I've spotted a weird situation:

```
DEBU[2020-06-17T19:49:13.9958901+02:00] HTTP response [0]                             caller="web/handlers.go:281" method=GET request_id=kyduqperkfbk5rt9n1ywp1akpc status_code=0 url=/api/v4/websocket
```

web/handlers.go
```diff
@@ -268,15 +268,22 @@ func (h Handler) ServeHTTP(w http.ResponseWriter, r *http.Request) {
...
 		c.App.Metrics().IncrementHttpRequest()
 
 		if r.URL.Path != model.API_URL_SUFFIX+"/websocket" {
 			elapsed := float64(time.Since(now)) / float64(time.Second)
-			statusCode := strconv.Itoa(w.(*responseWriterWrapper).StatusCode())
 			c.App.Metrics().ObserveApiEndpointDuration(h.HandlerName, r.Method, statusCode, elapsed)
 		}
 	}
+
+	mlog.Debug("HTTP response ["+statusCode+"]",
```
agnivade, line 281: Is there a reason behind adding status code in 2 places?

ethervoid: I was receiving go vet error about unstructured logging so I decided to include it as part of the parameters as well. In any case If we’re going to reduce it to just one I’ll leave the first one

agnivade: Yes, let's use it only as a parameter so that it can be consumed by logstash and other tools.

And avoid logging that if the status code is 0 to keep things clean.

agnivade: Ah no, I think you misunderstood. You did the opposite of what I was suggesting :P

I was suggesting 2 things:
- To have the status_code as a field in structured logging so that it can be consumed by other tools. Otherwise it will be impossible to filter by a status code, which is partially our objective.
- Append the status code to the slice of fields only if it is not 0. For example:
```
fields := []mlog.Field{.., .., ,,}
if statusCode != "0" {fields = append(fields, mlog.String(..))}
mlog.Debug(.., fields...)
```

Sorry for any confusion.

ethervoid: Oh I see. Such a small PR and three times doing it wrong :facepalm:

ethervoid: Done and right this time :sweat:

ethervoid:
Final form:

```
{"level":"debug","ts":1592490406.0884986,"caller":"web/handlers.go:290","msg":"HTTP response","method":"GET","url":"/api/v4/channels/4qoy1k44yff97j4d6qzs56xj7h/timezones","request_id":"ta188a3whpyexxxx3is9frqgpo","status_code":"200"}
```

For responses with status code:
```
{"level":"debug","ts":1592490409.8319173,"caller":"app/web_conn.go:417","msg":"websocket.read: client side closed socket","user_id":"enixogszc3b45qi5kb6rduu8fe"}
{"level":"debug","ts":1592490409.831987,"caller":"web/handlers.go:290","msg":"HTTP response","method":"GET","url":"/api/v4/websocket","request_id":"eexbz6tmuby65ywghir7rpxn3o"}
```

agnivade:
Ok, I have to apologize again. But I found a very _small_ problem. It's nothing big, but will require another commit. :sob: 

The issue is that we are logging the response at the end, which means there are 2 log lines. And more importantly, I see that if the rate limit exceeds, we are returning early. So in that case, the response will never be logged. What I was envisioning is that we will have the same log line, but with the response code attached. The timing should not be an issue since everything is captured in Prometheus anyways. And even in case of rate limits, it gets logged.

Here is the diff:

```diff
--- a/web/handlers.go
+++ b/web/handlers.go
@@ -86,7 +86,19 @@ func (h Handler) ServeHTTP(w http.ResponseWriter, r *http.Request) {
        now := time.Now()
 
        requestID := model.NewId()
-       mlog.Debug("Received HTTP request", mlog.String("method", r.Method), mlog.String("url", r.URL.Path), mlog.String("request_id", requestID))
+       var statusCode string
+       defer func() {
+               responseLogFields := []mlog.Field{
+                       mlog.String("method", r.Method),
+                       mlog.String("url", r.URL.Path),
+                       mlog.String("request_id", requestID),
+               }
+               // Websockets are returning status code 0 to requests after closing the socket
+               if statusCode != "0" {
+                       responseLogFields = append(responseLogFields, mlog.String("status_code", statusCode))
+               }
+               mlog.Debug("Received HTTP request", responseLogFields...)
+       }()
 
        c := &Context{}
        c.App = app.New(
@@ -268,7 +280,7 @@ func (h Handler) ServeHTTP(w http.ResponseWriter, r *http.Request) {
                }
        }
 
-       statusCode := strconv.Itoa(w.(*responseWriterWrapper).StatusCode())
+       statusCode = strconv.Itoa(w.(*responseWriterWrapper).StatusCode())
        if c.App.Metrics() != nil {
                c.App.Metrics().IncrementHttpRequest()
 
@@ -277,17 +289,6 @@ func (h Handler) ServeHTTP(w http.ResponseWriter, r *http.Request) {
                        c.App.Metrics().ObserveApiEndpointDuration(h.HandlerName, r.Method, statusCode, elapsed)
                }
        }
-
-       responseLogFields := []mlog.Field{
-               mlog.String("method", r.Method),
-               mlog.String("url", r.URL.Path),
-               mlog.String("request_id", requestID),
-       }
-       // Websockets are returning status code 0 to requests after closing the socket
-       if statusCode != "0" {
-               responseLogFields = append(responseLogFields, mlog.String("status_code", statusCode))
-       }
-       mlog.Debug("HTTP response", responseLogFields...)
 }
```

My apologies again for bringing this up at the last moment.

ethervoid:
I see. Well, I liked the idea to have those two log lines but I don't mind unifying them into one to be written at the end of the execution. 

I saw the rate limiter part but given that we don't have status code I didn't mind to not having that logline but this way is better :)

PR #14846, opened 2020-06-18
[MM-21703] Send telemetry for all Marketplace plugins

hanzei:
#### Summary
Instead of relying on a list of known plugin to figure out for which plugins the version number and if it's enabled should be send, fetch the Marketplace for the list of plugins and use that list.

If the Marketplace can't be reached, fall back to the list of known plugins.

As requested by @aaronrothschild the Marketplace is also fetched if `Remote Marketplace` is disabled or if a custom Marketplase is used. This means we will get telemetry data from custom Marketplaces.

All plugins had a custom mapping from plugin id to the segement id that is used to identify the data of each plugin. I did add a workaround to ensure the data stay consistency. Future plugins will use there plugin id as identification. This has to be kept in mind when adding new  plugins to `knownPlugins`.

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-21703


app/diagnostics.go
```diff
@@ -1110,3 +1061,128 @@ func (s *Server) trackChannelModeration() {
...
+		"enable_marketplace":            *cfg.PluginSettings.EnableMarketplace,
+		"require_pluginSignature":       *cfg.PluginSettings.RequirePluginSignature,
+		"enable_remote_marketplace":     *cfg.PluginSettings.EnableRemoteMarketplace,
+		"automatic_prepackaged_plugins": *cfg.PluginSettings.AutomaticPrepackagedPlugins,
+		"is_default_marketplace_url":    isDefault(*cfg.PluginSettings.MarketplaceUrl, model.PLUGIN_SETTINGS_DEFAULT_MARKETPLACE_URL),
+		"signature_public_key_files":    len(cfg.PluginSettings.SignaturePublicKeyFiles),
+	}
+
+	// knownPlugins maps plugin IDs to telemetry suffixes
+	var knownPlugins = map[string]string{
```
lieut-data, line 1081: Since we fully control the dataplane now, should we just sync with the team there and have them maintain the historical mapping?

app/diagnostics.go
```diff
@@ -1110,3 +1061,128 @@ func (s *Server) trackChannelModeration() {
...
+		"skype4business":                              "skype4business",
+		"com.mattermost.plugin-todo":                  "todo",
+		"com.mattermost.webex":                        "webex",
+		"com.mattermost.welcomebot":                   "welcome_bot",
+		"zoom":                                        "zoom",
+	}
+
+	marketplacePlugins, err := s.getAllMarketplaceplugins(cfg)
+	if err != nil {
+		log.Println("keks 1")
```
lieut-data, line 1103: Debugging logs?

app/diagnostics.go
```diff
@@ -1110,3 +1061,128 @@ func (s *Server) trackChannelModeration() {
...
+			}
+		}
+	}
+
+	s.SendDiagnostic(TRACK_CONFIG_PLUGIN, pluginConfigData)
+}
+
+func (s *Server) getAllMarketplaceplugins(cfg *model.Config) ([]*model.BaseMarketplacePlugin, error) {
+	marketplaceClient, err := marketplace.NewClient(
+		*cfg.PluginSettings.MarketplaceUrl,
```
lieut-data, line 1160: I was expecting we would unconditionally query from /our/ marketplace here, not the configured one. Otherwise we're using the customer's list of private plugins as an authority for the whitelist, which I'm assuming is not desired.

app/diagnostics.go
```diff
@@ -1110,3 +1061,128 @@ func (s *Server) trackChannelModeration() {
...
+	license := s.License()
+	if license != nil && *license.Features.EnterprisePlugins {
+		filter.EnterprisePlugins = true
+	}
+
+	if model.BuildEnterpriseReady == "true" {
+		filter.BuildEnterpriseReady = true
+	}
+
+	plugins, err := marketplaceClient.GetPlugins(filter)
```
lieut-data, line 1182: Can we just return directly here?

app/diagnostics.go
aaronrothschild: @Adovenmuehle do you have any thoughts on this?

app/diagnostics.go
aaronrothschild: Yes, I think that's correct (that we do NOT want to implicitly track the customer's list of private plugins).  @hanzei Perhaps I misunderstood something.

I think we should track the number of plugins installed/enabled on a server as a count.  But only have data about version, etc. From plugins that are within the MM marketplace.

app/diagnostics.go
Adovenmuehle: Would this mean that, for instance, instead of us getting an `enable_gitlab` flag we would instead get `enable_com.github.manland.mattermost-plugin-gitlab`? I'd prefer to keep the former if possible. I've also talked to @larkox and we were discussing having a plugin "short name" defined for each plugin so that we could just have `gitlab` instead of the `com.github...` one for use in telemetry.

lieut-data: @Adovenmuehle, can you expand on the rationale for having "short" names sent by the application instead of the canonical ids? Is there a performance concern?

app/diagnostics.go
hanzei: Removed

app/diagnostics.go
hanzei: Yes, changed that.

app/diagnostics.go
hanzei: Got it. Thanks for the clarification. Changed the code to always fetch the list from https://api.integrations.mattermost.com.

> I think we should track the number of plugins installed/enabled on a server as a count. 

We already track that data in `enabled_plugins`.

app/diagnostics.go
```diff
@@ -1110,3 +1061,128 @@ func (s *Server) trackChannelModeration() {
...
+		"zoom":                                        "zoom",
+	}
+
+	marketplacePlugins, err := s.getAllMarketplaceplugins(cfg)
+	if err != nil {
+		log.Println("keks 1")
+		mlog.Info("Failed to fetch marketplace plugins for telemetry. Using predefined list.", mlog.Err(err))
+
+		for k, v := range knownPlugins {
+			pluginConfigData["enable_"+v] = pluginActivated(cfg.PluginSettings.PluginStates, k)
```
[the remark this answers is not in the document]
hanzei: What is the advantage of using `Sprintf` over string concatenation?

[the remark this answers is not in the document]
marianunez: No performance advantage. I think it makes the code more cleaner but it's a matter of personal preference. 0/5 either way

app/diagnostics.go
Adovenmuehle: Not performance, just usability on the Snowflake side. essentially each one of those names will become a column name in the database, so `com.mattermost.confluence` will end up as `ENABLE_COM_MATTERMOST_CONFLUENCE` which isn't necessarily horrible but it's just easier for it to be `ENABLE_CONFLUENCE`. Also, if we start adding more telemetry in the plugins themselves, and they standardize on the short name, it'll be a lot nicer to have track events prefixed by the short plugin name (e.g. `CONFLUENCE_SOME_EVENT` vs `COM_MATTERMOST_CONFLUENCE_SOME_EVENT`).

hanzei: There might be a future where multiple plugins of the same name exist in the Marketplace, e.g. an official JIRA plugin and a community written one. Using the plugin name as a id will cause problems in that case.

@Adovenmuehle Is it responsable to "prettify" the data  upstream e.g. in Snowflake?

Adovenmuehle: Yeah having collisions on short names would make it less than ideal.

I'm most invested in having the plugin specific telemetry (e.g. events that are sent directly from the plugin) to follow a nicer pattern than I am for these `enabled` flags.

For context, RudderStack is responsible for creating the column names in Snowflake (and we don't really have control over it).

hanzei: > I'm most invested in having the plugin specific telemetry (e.g. events that are sent directly from the plugin) to follow a nicer pattern than I am for these `enabled` flags.

@Adovenmuehle Plugin specific events are already using the plugin ID (https://github.com/mattermost/mattermost-plugin-api/blob/5b8d13c39e262600255d59e82d864db28ab4c1f1/experimental/telemetry/tracker.go#L80) Hence, I think it makes sense to also use the plugin ID for the `enabled` flags.

Adovenmuehle: I talked to @enelson720 and we're good to go for using the plugin IDs for the enabled flags and version numbers here. So you no longer need the `knownPlugins` map.

hanzei:
I've update the code and dropped the historical mapping.

But I kept the list of known plugins. In the rare case that a lot of requests fail to the Marketplace losing all of the telemetry data would be bad. I expect that this list of only used in a few cases and hence won't have a high impact on which data is send. 1/5 that keeping this list is usefull.

aaronrothschild:
> As requested by @aaronrothschild the Marketplace is also fetched if Remote Marketplace is disabled or if a custom Marketplase is used. This means we will get telemetry data from custom Marketplaces.

If I recall this was to track the Incident Response plugins too.  

@hanzei Are we adding the community plugin's PLuginID to the whitelist for telemetry when we add community plugins to the marketplace?  Is that repeatable (ie: we need to update the server whitelist before the plugin ships ideally)?

hanzei:
> > As requested by @aaronrothschild the Marketplace is also fetched if Remote Marketplace is disabled or if a custom Marketplase is used. This means we will get telemetry data from custom Marketplaces.
> 
> If I recall this was to track the Incident Response plugins too.
> 

Good point! @itao is this still a request? Once the Incident Response plugin gets added to the "normal" Marketplace the number of installations will be tracked by default.

> @hanzei Are we adding the community plugin's PLuginID to the whitelist for telemetry when we add community plugins to the marketplace? Is that repeatable (ie: we need to update the server whitelist before the plugin ships ideally)?

@aaronrothschild Yes, it's desirable to add the plugins ID to the whitelist first, but not requires as fetching the list of Marketplace plugin should work almost every time.

amyblais:
@aaronrothschild Is this for v5.26? Today is code complete so wanted to check on expected fix version,

aaronrothschild:
@amyblais I'm sorry, this slipped through the cracks.

Adovenmuehle:
@lindy65 tested this locally and it's good to go.

lindy65:
Thanks so much @Adovenmuehle! Appreciated 🤗

PR #14852, opened 2020-06-18
MM-26231: Reverts plugin API breaking change.

mkraft:
#### Summary

Reverts plugin API breaking change. 

The only field from `model.GroupSearchOpts` that is being used is `FilterAllowReference`, which defaults to the original behaviour when `false`. By invoking `(*App).GetGroupByName` with `model.GroupSearchOpts{}` there should be no changes to the response received by plugins. 

In the future, consideration can be made to adding that field to the plugin API in a SemVer-compatible way, and I don't see that requiring the whole `model.GroupSearchOpts` struct.

#### Ticket Link

https://mattermost.atlassian.net/browse/MM-26231

### DIFF
--- a/app/plugin_api.go
+++ b/app/plugin_api.go
@@ -471,8 +471,8 @@ func (api *PluginAPI) GetGroup(groupId string) (*model.Group, *model.AppError) {
 	return api.app.GetGroup(groupId)
 }
 
-func (api *PluginAPI) GetGroupByName(name string, opts model.GroupSearchOpts) (*model.Group, *model.AppError) {
-	return api.app.GetGroupByName(name, opts)
+func (api *PluginAPI) GetGroupByName(name string) (*model.Group, *model.AppError) {
+	return api.app.GetGroupByName(name, model.GroupSearchOpts{})
 }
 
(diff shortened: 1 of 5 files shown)


catalintomai:
Thanks for helping @mkraft , sorry for the mishap. @lieut-data , maybe we can consider adding a link/small text to the api.go file that describes/warns about the dos/donts when modifying the file to avoid these mistakes in the future.

DHaussermann:
@mkraft can you please provide testing guidance or steps for this? I'm not familiar with this initial change or the solution to it.

mkraft:
> @mkraft can you please provide testing guidance or steps for this? I'm not familiar with this initial change or the solution to it.

The idea for this would be to confirm that that plugin API's `GetGroupByName` method continues to behave as it did previously, so I think it's just a matter of re-running existing tests steps.

amyblais:
@mkraft This also needs to be cherry picked to v5.25.

mkraft:
> @mkraft This also needs to be cherry picked to v5.25.

Done. Thanks Amy.

PR #14857, opened 2020-06-19
MM-19548: Add a deadlock retry function for SaveChannel

agnivade:
A deadlock has been seen to occur in the upsertPublicChannelT method
during bulk import.

Here is a brief excerpt:

```
*** (1) TRANSACTION:
TRANSACTION 3141, ACTIVE 1 sec inserting
INSERT INTO
                            PublicChannels(Id, DeleteAt, TeamId, DisplayName, Name, Header, Purpose)
                        VALUES
                            (?, ?, ?, ?, ?, ?, ?)
                        ON DUPLICATE KEY UPDATE
                            DeleteAt = ?,
                            TeamId = ?,
                            DisplayName = ?,
                            Name = ?,
                            Header = ?,
                            Purpose = ?
*** (1) WAITING FOR THIS LOCK TO BE GRANTED:
RECORD LOCKS space id 76 page no 4 n bits 104 index Name of table `mydb`.`PublicChannels` trx id 3141 lock_mode X locks gap before rec insert intention waiting
** (2) TRANSACTION:
TRANSACTION 3140, ACTIVE 1 sec inserting
mysql tables in use 1, locked 1
5 lock struct(s), heap size 1136, 3 row lock(s), undo log entries 2
MySQL thread id 50, OS thread handle 140641523848960, query id 3226 172.17.0.1 mmuser update
INSERT INTO
                            PublicChannels(Id, DeleteAt, TeamId, DisplayName, Name, Header, Purpose)
                        VALUES
                            (?, ?, ?, ?, ?, ?, ?)
                        ON DUPLICATE KEY UPDATE
                            DeleteAt = ?,
                            TeamId = ?,
                            DisplayName = ?,
                            Name = ?,
                            Header = ?,
                            Purpose = ?
*** (2) HOLDS THE LOCK(S):
RECORD LOCKS space id 76 page no 4 n bits 104 index Name of table `mydb`.`PublicChannels` trx id 3140 lock_mode X locks gap before rec

*** (2) WAITING FOR THIS LOCK TO BE GRANTED:
RECORD LOCKS space id 76 page no 4 n bits 104 index Name of table `mydb`.`PublicChannels` trx id 3140 lock_mode X locks gap before rec insert intention waiting

*** WE ROLL BACK TRANSACTION (1)
```

Following is my analysis:

From the deadlock output, it can be seen that it's due to a gap lock.
And that's clear because the index is `Name` which is a multi-column index using `Name` and `TeamId`.
But interestingly, both transactions seem to be inserting the same data, which is what is puzzling me.
The multi-column index on `Name` and `TeamId` will guarantee that they are always unique. And from looking at the code,
it does not seem possible to me that it will try to insert the same data from 2 different transactions.
But even if they do, why does tx 2 try to acquire the same lock again when it already has that ?

Here is what I think the order of events happening

- Tx 2 gets a gap lock.
- Tx 1 tries to get the same gap lock.
- Tx 2 tries to again get the same gap lock ?

The last step is what is puzzling me. Why does an UPSERT statement acquire 2 gap locks ? From my reading of https://dev.mysql.com/doc/refman/8.0/en/innodb-locks-set.html:

> INSERT ... ON DUPLICATE KEY UPDATE differs from a simple INSERT in that an exclusive lock rather than a shared lock is placed on the row to be updated when a duplicate-key error occurs. An exclusive index-record lock is taken for a duplicate primary key value. An exclusive next-key lock is taken for a duplicate unique key value.

From what I understand, the expectation is that there will be one X lock and one gap lock is taken.
But that's not what the deadlock output seems to say.

The general advice on the internet seems to be that deadlocks will happen and not all of them can be understood.

For now, we add a generic deadlock retry function at the store package which can be reused by other queries too.

P.S.: This is a verbatim copy of my investigation posted at https://dba.stackexchange.com/questions/268652/mysql-deadlock-upsert-query-acquiring-gap-lock-twice

Testing:

This is ofcourse hard to test because it is impossible to reproduce this. I have tested this by manually returning an error
and confirming that it indeed retries.

```
WARN[2020-06-19T11:18:24.9585676+05:30] A deadlock happened. Retrying.                caller="sqlstore/channel_store.go:568" error="Error 1213: mydeadlock"
WARN[2020-06-19T11:18:24.959158+05:30] A deadlock happened. Retrying.                caller="sqlstore/channel_store.go:568" error="Error 1213: mydeadlock"
WARN[2020-06-19T11:18:24.9595072+05:30] A deadlock happened. Retrying.                caller="sqlstore/channel_store.go:568" error="Error 1213: mydeadlock"
WARN[2020-06-19T11:18:24.9595451+05:30] Deadlock happened 3 times. Giving up          caller="sqlstore/channel_store.go:579"
ERRO[2020-06-19T11:18:24.9596426+05:30] Unable to save channel.                       caller="mlog/log.go:175" err_details="Error 1213: mydeadlock" err_where=CreateChannel http_code=500 ip_addr="::1" method=POST path=/api/v4/channels request_id=745bsj13b7f6mnmsbn3t97grbw user_id=xcof1ipipbrfxpfjf6x4p6kx9e
```


store/sqlstore/channel_store.go
```diff
@@ -518,27 +518,32 @@ func (s SqlChannelStore) Save(channel *model.Channel, maxChannelsPerTeam int64)
...
-
-	newChannel, err := s.saveChannelT(transaction, channel, maxChannelsPerTeam)
-	if err != nil {
-		return newChannel, err
-	}
+	var newChannel *model.Channel
+	err := store.WithDeadlockRetry(func() error {
+		transaction, err := s.GetMaster().Begin()
+		if err != nil {
+			return errors.Wrapf(err, "begin_transaction: ")
```
lieut-data, line 525: Should we eliminate the trailing `: ` while in here?

store/sqlstore/channel_store.go
```diff
@@ -518,27 +518,32 @@ func (s SqlChannelStore) Save(channel *model.Channel, maxChannelsPerTeam int64)
...
+			return errors.Wrapf(err, "upsert_public_channel: ")
+		}
 
-	return newChannel, nil
+		if err := transaction.Commit(); err != nil {
+			return errors.Wrapf(err, "commit_transaction: ")
+		}
+		return nil
+	})
+	// There are cases when in case of conflict, the original channel value is returned.
```
lieut-data, line 544: I note that while we're now consistently returning `newChannel`, where there were cases before when we /didn't/. Is it worth fixing the caller in these cases to not rely on this behaviour?

store/store.go
```diff
@@ -770,3 +774,33 @@ type IntegrityCheckResult struct {
...
+			// But deadlocks are very rarely seen in postgres.
+			var mysqlErr *mysql.MySQLError
+			if errors.As(err, &mysqlErr) && mysqlErr.Number == mySQLDeadlockCode {
+				mlog.Warn("A deadlock happened. Retrying.", mlog.Err(err))
+				// This is a deadlock, retry.
+				continue
+			}
+			// Some other error, return as-is.
+			return err
+		} else {
```
lieut-data, line 799: nit: invert check and early return to avoid indentation of main body?

store/store.go
```diff
@@ -770,3 +774,33 @@ type IntegrityCheckResult struct {
...
+			// No error, return nil.
+			return nil
+		}
+	}
+	mlog.Error("Deadlock happened 3 times. Giving up")
```
lieut-data, line 804: This log may be emitted without context, especially if there are multiple things going on in the logs. Should we just wrap the error with a "giving up after 3 deadlocks" or some such? Also, does this deserve to be an `Error`, in that the sysadmin need take some corrective action?

store/store.go
```diff
@@ -770,3 +774,33 @@ type IntegrityCheckResult struct {
...
+// After exceeding a given threshold, it breaks and propagates the error upwards.
+// TODO: This can be a separate retry layer in itself where transaction retries
+// are automatically applied.
+func WithDeadlockRetry(f func() error) error {
+	var err error
+	for i := 0; i < 3; i++ {
+		err = f()
+		if err != nil {
+			// XXX: Possibly add check for postgres deadlocks later.
+			// But deadlocks are very rarely seen in postgres.
```
lieut-data, line 790: 🥇

store/store.go
```diff
@@ -770,3 +774,33 @@ type IntegrityCheckResult struct {
...
 	Err  error
 }
+
+const mySQLDeadlockCode = uint16(1213)
+
+// WithDeadlockRetry retries a given f if it throws a deadlock error.
+// After exceeding a given threshold, it breaks and propagates the error upwards.
+// TODO: This can be a separate retry layer in itself where transaction retries
+// are automatically applied.
+func WithDeadlockRetry(f func() error) error {
```
lieut-data, line 784: Is there an opportunity to unit-test this function?

store/store.go
agnivade: > Should we just wrap the error with a "giving up after 3 deadlocks" or some such?

Makes sense.

>  Also, does this deserve to be an Error, in that the sysadmin need take some corrective action?

Yes, I think so. If a deadlock happens consecutively 3 times, there's something seriously wrong somewhere.

store/sqlstore/channel_store.go
agnivade: I'll do a sweep and check for any such cases.

store/store.go
```diff
@@ -770,3 +774,31 @@ type IntegrityCheckResult struct {
 	Data interface{}
 	Err  error
 }
+
+const mySQLDeadlockCode = uint16(1213)
+
+// WithDeadlockRetry retries a given f if it throws a deadlock error.
+// After exceeding a given threshold, it breaks and propagates the error upwards.
```
streamer45, line 781: A couple of things:

- Is the threshold given? I would expect the caller to decide how much to try before eventually failing instead of relying on a hard-coded value.
- Are there cases where it makes more sense to pass a time value (e.g. a deadline) and rely on that rather than on a number of retries?

agnivade: - IMO, I am not seeing much value in making the threshold as a parameter. Letting the user pass the parameter just adds new API surface without much benefit. Do you have any specific use-cases in mind? The problem is that there is no ideal number here that can be left to the user. Increasing from 3 to 5 just moves the goal post further.

If the query deadlocks even after retrying 3 times, there is something seriously broken somewhere that needs to be fixed IMO. 

Of course, there maybe cases of high load where deadlock-able queries get executed repeatedly. But in that case, even 5 or 10 might not be good enough. And it might be better just to stop the query altogether to prevent additional load on the DB and look towards fixing the deadlock some other way.

- IMO, I don't think so. MySQL would rollback one transaction while letting the other pass through. So that automatically lets the retried transaction to go forward.

Thoughts?

streamer45: As far as my first point I think it's fine to just update the documentation for that function since the threshold is not really passed as an argument but defined in there.

The second point was about removing the need to re-try a fixed amount of times but instead try as many times as possible within the given deadline. This might give more control to the caller but as you said, if the point is just trying to fix the edge case, the fixed retries should be enough.

agnivade: Ah I misunderstood your 1st point. Updated the godocs. PTAL.

agnivade:
@lieut-data - Ohh totally my bad. I added the unit test but didn't add the file to the commit. And now I deleted the file thinking I had added it already. Now I have to write the entire test all over again.

agnivade:
@ogi-m - It is not really possible to test this in any way. I have added a unit test for my new function. And verified manually by returning an error from the code that if deadlocks happen, the function is retried.

Let me know your thoughts. Thanks.

PR #14859, opened 2020-06-19
MM-25890: Fix deadlock on deleting emoji reactions

agnivade:
A deadlock happens because `UPDATE_POST_HAS_REACTIONS_ON_DELETE_QUERY` is being called from 2 separate places.
1. From `DeleteAllWithEmojiName` where it's called as an independent query.
2. From `deleteReactionAndUpdatePost` where it's called as part of a transaction along with another DELETE query.

The deadlock occurs in such a scenario:

- tx #2 acquires an X lock from the DELETE query.
- tx #1 tries to acquire a S lock with the select query, but it's waiting for the X lock to be released from tx #2
- tx #2 now tries to acquire an S lock, but it can't because it is locked on tx #1.

Deadlock.

I have tested this and it does indeed deadlock. The root of the problem is that the Primary key is a multi-column index,
which means that a next-key lock has to be acquired to get a lock on the gap before the index.
Both the queries try to delete some reactions, and then select the new number of reactions. But they select different rows
due to which this happens.

This might just be an unavoidable deadlock due to the way the indexes are setup and next-key locks.
Unless we change the primary key to be a single-column index, it will be very hard to avoid this.

Therefore we just go with a simple retry.

Jira ticket: https://mattermost.atlassian.net/browse/MM-25890


store/storetest/reaction_store.go
```diff
@@ -422,3 +424,65 @@ func testReactionBulkGetForPosts(t *testing.T, ss store.Store) {
 	require.False(t, post4IdFound, "Wrong reaction returned")
 
 }
+
+// testReactionDeadlock is a best-case attempt to recreate the deadlock scenario.
+// It atleast deadlocks 2 times out of 5.
```
lieut-data, line 429: Epic :D

Nit:
```suggestion
 // It at least deadlocks 2 times out of 5.
```

store/storetest/reaction_store.go
```diff
@@ -422,3 +424,65 @@ func testReactionBulkGetForPosts(t *testing.T, ss store.Store) {
...
+	reaction4 := &model.Reaction{
+		UserId:    reaction1.UserId,
+		PostId:    reaction1.PostId,
+		EmojiName: model.NewId(),
+	}
+	_, err = ss.Reaction().Save(reaction4)
+	require.Nil(t, err)
+
+	var wg sync.WaitGroup
+	wg.Add(2)
```
lieut-data, line 473: nit: `wg.Add(1)` on two separate lines instead?

agnivade: I prefer the earlier pattern actually. :) If we know before-hand how many goroutines are going to be spun, pre-declaring them makes it appear simple to me.

store/sqlstore/supplier_reactions.go
```diff
@@ -56,19 +58,25 @@ func (s *SqlReactionStore) Save(reaction *model.Reaction) (*model.Reaction, *mod
...
-		return nil, model.NewAppError("SqlPreferenceStore.Delete", "store.sql_reaction.delete.app_error", nil, appErr.Error(), http.StatusInternalServerError)
-	}
+		err = deleteReactionAndUpdatePost(transaction, reaction)
+		if err != nil {
+			return errors.Wrap(err, "deleteReactionAndUpdatePost")
+		}
 
-	if err := transaction.Commit(); err != nil {
-		return nil, model.NewAppError("SqlPreferenceStore.Delete", "store.sql_reaction.delete.commit.app_error", nil, err.Error(), http.StatusInternalServerError)
+		if err := transaction.Commit(); err != nil {
```
streamer45, line 73: I think we can handle the two errors the same way :stuck_out_tongue:

agnivade:
@lindy65 - This is covered by unit tests, and it is very hard to reproduce otherwise. So I think we are good here.

PR #14869, opened 2020-06-19
Configurable dev environment

jespino:
Add the posibility to configure your dev enironment.

This works fine in linux, but I don't know if this works properly on mac, so, someone testing it would be great.

I have modified the docker-compose.yml file to generate the `start_dependencies` on the fly. I don't know if the `docker-compose.yml` file from the root directory have any other usage than provide that to the Makefile, if that is the case, I can figure out how to solve it.


.gitignore
```diff
@@ -110,3 +110,4 @@ client
 __debug_bin
 report.xml
 go.*.orig
+config.mk
```
lieut-data, line 113: nit: `config.mk` "sounds" like it should be part of the repository. Should we call this `override.mk` or `custom.mk`?

Makefile
```diff
@@ -122,6 +120,8 @@ MMCTL_REL_TO_DOWNLOAD = $(shell scripts/get_latest_release.sh 'mattermost/mmctl'
 
 all: run ## Alias for 'run'.
 
+-include config.mk
+include default-config.mk
```
lieut-data, line 124: It's not obvious to me why only four environment variables end up in `default-config.mk`, but not `BUILD_ENTERPRISE` or `GO` or `MMCTL_REL_TO_DOWNLOAD`.

Instead of a `default-config.mk`, could we just keep these definitions in this file (and above where they were originally located), and simply allow `config.mk` to override them? That gives `config.mk` the added advantage of being able to refer to the default values to make some decision.

Makefile
```diff
@@ -132,8 +132,10 @@ else ifeq ($(MM_NO_DOCKER),true)
 else
 	@echo Starting docker containers
 
-	docker-compose run --rm start_dependencies
-	cat tests/${LDAP_DATA}-data.ldif | docker-compose exec -T openldap bash -c 'ldapadd -x -D "cn=admin,dc=mm,dc=test,dc=com" -w mostest || true';
+	$(GO) run ./build/docker-compose-generator/main.go $(ENABLED_DOCKER_SERVICES) | docker-compose -f docker-compose.makefile.yml -f /dev/stdin run --rm start_dependencies
```
lieut-data, line 135: The `-f /dev/stdin` forces me to run through the `Makefile` in all cases, and I can't (easily) use `docker-compose` directly. 

I note from https://docs.docker.com/compose/extends/#multiple-compose-files that 
> By default, Compose reads two files, a `docker-compose.yml` and an optional `docker-compose.override.yml` file. By convention, the `docker-compose.yml` contains your base configuration. The override file, as its name implies, can contain configuration overrides for existing services or entirely new services.

Could we instead have add `docker-compose.override.yml` to `.gitignore` and unconditionally generate it on each `Makefile` invocation? Then it's around and ready for someone to use with `docker-compose` directly.

Makefile
```diff
@@ -367,8 +369,13 @@ run-server: prepackaged-binaries validate-go-version start-docker ## Starts the
 	@echo Running mattermost for development
 
 	mkdir -p $(BUILD_WEBAPP_DIR)/dist/files
+ifeq ($(RUN_SERVER_IN_BACKGROUND),true)
```
lieut-data, line 372: nit: Can we avoid the duplication of the `run` invocation by conditionally populating a variable with `&` based on this variable instead?

Makefile
```diff
@@ -545,3 +552,6 @@ endif
 ## Help documentatin à la https://marmelab.com/blog/2016/02/29/auto-documented-makefile.html
 help:
 	@grep -E '^[0-9a-zA-Z_-]+:.*?## .*$$' ./Makefile | sort | awk 'BEGIN {FS = ":.*?## "}; {printf "\033[36m%-30s\033[0m %s\n", $$1, $$2}'
+	@echo
```
lieut-data, line 555: nit: This is a "one-time" learning event. Could we just document this inline within the `Makefile` (and on mattermost.dev?), instead of permanently polluting `make help`'s output?

default-config.mk
```diff
@@ -0,0 +1,24 @@
...
+
+# Enable services to be run in docker.
+#
+# Possible options: mysql, postgres, minio, inbucket, openldap, dejavu,
+# keycloak and elasticsearch
+#
+# Must be space spearated names.
+#
+# Example: mysql postgres elasticsearch
+ENABLED_DOCKER_SERVICES ?= mysql
```
lieut-data, line 13: Our `sqlstore` unit tests require both `mysql` and `postgres` to run. Should we enable both by default to allow for passing unit tests?

Makefile
```diff
@@ -367,8 +369,13 @@ run-server: prepackaged-binaries validate-go-version start-docker ## Starts the
 	@echo Running mattermost for development
 
 	mkdir -p $(BUILD_WEBAPP_DIR)/dist/files
+ifeq ($(RUN_SERVER_IN_BACKGROUND),true)
 	$(GO) run $(GOFLAGS) -ldflags '$(LDFLAGS)' $(PLATFORM_FILES) --disableconfigwatch 2>&1 | \
 	    $(GO) run $(GOFLAGS) -ldflags '$(LDFLAGS)' $(PLATFORM_FILES) logs --logrus &
+else
+	$(GO) run $(GOFLAGS) -ldflags '$(LDFLAGS)' $(PLATFORM_FILES) --disableconfigwatch 2>&1 | \
```
lieut-data, line 376: This effectively makes `make run-server` a blocking call, which in turn means `make run` will never get to the `run-client` step.

default-config.mk
```diff
@@ -0,0 +1,24 @@
...
+
+# Disable entirely the use of docker
+MM_NO_DOCKER ?= false
+
+# Run the server in the background
+RUN_SERVER_IN_BACKGROUND ?= true
```
lieut-data, line 19: Could we plan for a more general variable name that might apply to a similar `webapp` change? Maybe `MM_DEVELOP_FOREGROUND`?

.gitignore
jespino: I'm ok changing that into something else, but `override.mk` or `custom.mk` i think is not so obvious. maybe have `config.mk` instead of `default-config.mk` and use `config.override.mk` to override the default config.

Makefile
jespino: I would limit this to the settings that the user really wants to configure. For example, `BUILD_ENTERPRISE`, by default is inferred from the existence of the `BUILD_ENTERPRISE_DIR`. Is it a common use case to modify that variables? I can move almost all the variables there, but I don't want to make it a place where is hard to understand what to touch and what not.

Makefile
jespino: Actually, we have 2 docker-compose files, one is the `docker-compose.makefile.yml` that is the one used from the makefile, and it is exactly the same as `docker-compose.yaml` but without the autostart block, the autostart block is added dynamically by the `./build/docker-compose-generator/main.go`.

The `docker-compose.yaml` is reserved for the regular docker-compose usage, without nothing to do with our own `docker-compose.makefile.yml` that you can use if you want to start the containers individually. I have tried to be conservative and not break anybodies workflow here.

Makefile
jespino: We could, but I don't buy the "one-time" learning event, we could say exactly the same for any other Makefile task (you only needs to learn what `run` does one time, why is showing it to me every single time?)

default-config.mk
jespino: I doubt about that, I can re-enabled everything that was enabled before as default to avoid anybodies workflow to be broken.

About unit tests... to be honest, almost nobody is running mysql/postgres tests locally except while they are working on the store, so... that is very focused to a small subset of our developers.

Said that, I still think is interesting to keep the thing working as was working before, and only add the "configurable" part. On the other hand, that means, the new comers are going to suffer the huge unneeded overhead anyway.

jespino: other option is to reenable everything that was enable before, except for enterprise (that means, remove elasticsearch and openldap by default), because we have more control on who is developing that with elastic and ldap, and we can talk with them directly.

Makefile
jespino: Actually, you are right, but we can change the order between run-client and run-server and should work.

default-config.mk
jespino: Mmm... interesting. we can assume that you may want to run things in background or in foreground, and if it is in foreground for example, I can fail if you try to execute `make run` saying that when you are using foreground mode you have to execute both comands independently. What do you think @jwilander? that would fit your workflow, or is not what you expect?

Makefile
jespino: let me try

ethervoid:
@jespino what about using MySQL 5.6 by default given that is our lower supported version?

jespino:
> @jespino what about using MySQL 5.6 by default given that is our lower supported version?

Totally agree

Makefile
reflog: I'd go with Jesus here. I think it's useful to have this note.

build/docker-compose.common.yml
```diff
@@ -1,7 +1,7 @@
 version: '2.4'
 services:
   mysql:
-    image: "mysql:5.7"
+    image: "mysql:5.6"
```
reflog, line 4: why the version change?

config.mk
```diff
@@ -0,0 +1,24 @@
...
+# Disable entirely the use of docker
+MM_NO_DOCKER ?= false
+
+# Run the server in the background
+RUN_SERVER_IN_BACKGROUND ?= true
+
+# Data loaded by default in openldap when container starts.
+#
+# Posible options: test or qa
+LDAP_DATA ?= test
```
reflog, line 24: could these be overridden with an env variable?

store/storetest/channel_store.go
```diff
@@ -24,6 +24,7 @@ import (
 
 type SqlSupplier interface {
 	GetMaster() *gorp.DbMap
+	DriverName() string
```
reflog, line 27: why is this change in this PR?  (same for the changes below)

jespino: I changed the version of mysql to 5.6, and I've being debugging this failing tests during hours and hours, and only fails on certain circustances on MySQL 5.6, I haven't find the way to reproduce that and I don't want to invest more time on this before this get merged. We can iterate later, but I've given up with this specific test case.

build/docker-compose.common.yml
jespino: We want to ensure that we are using in our dev environment (And in the CI) the oldest supported version. We have had problems with features that didn't work with MySQL 5.6 because we were using 5.7 in development and some features were not available.

config.mk
jespino: Yes it can, because we are using ?= it only sets the value if is not already set.

jespino:
This has been here for more than 2 weeks already, so I got 2 reviews from my team, and I'll merge it tomorrow morning unless somebody have any not-addressed concern:  @lieut-data @grundleborg @jwilander @cpanato @sbishel 

By the way, this is the first implementation, the idea is to iterate over this including new ideas, for example, we can add replicated databases, redis (for future redis cache implementation), prometheus/grafana, elk... whetever we think is interesting for optional "plugins" for development.

Makefile
```diff
@@ -122,8 +119,15 @@ MMCTL_REL_TO_DOWNLOAD = $(shell scripts/get_latest_release.sh 'mattermost/mmctl'
 
 all: run ## Alias for 'run'.
 
+-include config.override.mk
+include config.mk
 include build/*.mk
 
+RUN_IN_BACKGROUND ?=
```
lieut-data, line 126: `config.mk` has already defined `RUN_IN_BACKGROUND` as `true` -- do we still need to re-declare it here?

Makefile
```diff
@@ -122,8 +119,15 @@ MMCTL_REL_TO_DOWNLOAD = $(shell scripts/get_latest_release.sh 'mattermost/mmctl'
 
 all: run ## Alias for 'run'.
 
+-include config.override.mk
```
lieut-data, line 122: I'm still confused by `config.override.mk` and `config.mk`:
* We import the /override/ before we import the canonical `config.mk`, relying on the conditional assignment (`?=`) to avoid undoing any overrides. This is brittle to a future change to `config.mk` that doesn't realize the nuances here. 
* As noted before, I can't make decisions from my override script using the canonical definitions of these variables since they aren't defined yet.
* I don't understand the subset of things that were moved to `config.mk`, nor can I explain to someone under what conditions a future variable should reside in the `Makefile` or in `custom.mk`.

I remain inclined to think we should eliminate `config.mk` (moving the variables back here), import `build/*.mk` and finally import a `override.mk` (or more canonically, a `build/custom.mk` like our plugin repositories).

Thoughts?

Makefile
lieut-data: Ah, excellent! I'm worried about the duplication here -- these will invariably drift out of sync when only testing one workflow. Can we generate the `docker-compose.makefile.yml` on the go instead?

Makefile
jespino: Actually, is subtle, but `RUN_IN_BACKGROND` is defining `&` not `true` or `false`, I use the `RUN_SERVER_IN_BACKGROUND` to initialize that, to avoid some code repetition later.

Makefile
jespino: We can generate the whole thing on the fly, I would add that as another PR, I don't think is strictly neede for this PR.

Makefile
jespino: moving that into `build/custom.mk` make it hard to find for any new developer. Having a `config.mk` shows explicitly that you have certian things that we consider configuration of your development environment, and then `config.override.mk` make it pretty clear what is your configuration.

About the order, is on purpose, I want you to be able override settings (forcing them or not) and at the same time allowing you to override them with environment variables.

So, If I load the config.mk first, I'm force to define my settings as "mandatory" if I want to override the `config.mk`, and that remove the posibility of having environment variables overwriting that. My idea is, you can "enforce settings" or you can "softly define settings", the solftly defined can be override by environment variables, the enforced ones no.

Makefile
lieut-data: Ah, yes, that makes sense. Thank you for clarifying :)

Makefile
lieut-data: > So, If I load the `config.mk` first, I'm force to define my settings as "mandatory" if I want to override the `config.mk`, and that remove the posibility of having environment variables overwriting that. My idea is, you can "enforce settings" or you can "softly define settings", the solftly defined can be override by environment variables, the enforced ones no.

Got it -- but I admit this is /really/ subtle. Regardless of where we end up, can you update the `README.md` or developer docs to clearly document these decisions?

Are we sure we need to define two semantically different ways of overriding the base configuration? What if we only just supported environment variable overrides, with everything in `Makefile` being a `?=`? Less discoverable, but maybe less surprising.

Are there open source repositories using this strategy that might help clarify my understanding of the vision here? I'm worried this feels too "cobbled together" to be unambiguously useful.

jespino: The environment variables is hard to discover, really hard, and is not so easy to manage, having a config file that you have in your filesystem is way easier and comfortable. I don't see why it should be surprising? And having this files, instead of a set of variables in the Makefile is more clear an self-documented than "use environment variables to override things there in the Makefile and see what happens".

Makefile
lieut-data: Any reason not to "get this right" with the first pass? I'm pushing back slightly because this affects every single developer using Mattermost and immediately hits a wide audience.

Makefile
lieut-data: I remain surprised by the proposed behaviour, but I'm happy to withdraw my requested changes. There's lots of great stuff here that I hope to use soon :)

lieut-data:
I've commented above, but have also withdrawn my review -- happy to defer to the server team's feedback. Thanks @jespino!

PR #14870, opened 2020-06-19
MM-26343 Make CreateInitialSidebarCategories idempotent

hmhealey:
Check whether or not the default sidebar categories exist before trying to create them to avoid potentially creating duplicates in weird cases.

This unfortunately can't be prevented with a unique constraint like we originally wanted to do because the a user can have any number of `custom` categories. It sounds like Postgres has a way to do this with a partial unique constraint, but nothing like that exists for MySQL.

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-26343


hmhealey:
The build for this is just failing because https://github.com/mattermost/mattermost-api-reference/pull/532 has not been merged yet.

PR #14880, opened 2020-06-22
MM-26055 Remove session requirement from command args.

crspeller:
#### Summary
Removes session requirement from command args. This paves the way for slash commands to be executed by plugins as plugins can not provide session context. 

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-26055


model/command_args.go
```diff
@@ -20,7 +20,6 @@ type CommandArgs struct {
 	Command         string               `json:"command"`
 	SiteURL         string               `json:"-"`
 	T               goi18n.TranslateFunc `json:"-"`
-	Session         Session              `json:"-"`
```
ali-farooq0, line 23: Are we OK introducing a breaking change in this model for plugins? This model is being exposed through the `ExecuteCommand` [Hook API](https://github.com/mattermost/mattermost-server/blob/bf1d9b0ae9f2e8bcaab251e69eb6bb040584812c/plugin/hooks.go#L90). Should we just take away the functionality for this release and put in a comment saying this is deprecated? and eventually remove it 3-4 releases down the road?

crspeller: Right, custom command won't get this because of the `json:"-"` but plugins will. That's actually kinda a bug, I don't think that was intentional. Certainly not for the translation function.

Unfortunately nobody is actually going to see a comment that says this is deprecated.

We could re-institute as deprecated it no problem. It's just that anyone that happens to use it won't work with plugins that eventually use the execute slash command API.

I will re-estate it for this unlikely case. Filed an issue against 6.0: https://mattermost.atlassian.net/browse/MM-26398

model/command_args.go
```diff
@@ -11,18 +11,21 @@ import (
...
+	TeamId    string               `json:"team_id"`
+	RootId    string               `json:"root_id"`
+	ParentId  string               `json:"parent_id"`
+	TriggerId string               `json:"trigger_id,omitempty"`
+	Command   string               `json:"command"`
+	SiteURL   string               `json:"-"`
+	T         goi18n.TranslateFunc `json:"-"`
+
+	// DO NOT USE Session field is depricated. MM-26398
+	Session Session `json:"-"`
```
ali-farooq0, line 25: 👍 

Should we move it to the very bottom -- separate for all the valid fields?

model/command_args.go
```diff
@@ -11,18 +11,21 @@ import (
...
+	ChannelId string               `json:"channel_id"`
+	TeamId    string               `json:"team_id"`
+	RootId    string               `json:"root_id"`
+	ParentId  string               `json:"parent_id"`
+	TriggerId string               `json:"trigger_id,omitempty"`
+	Command   string               `json:"command"`
+	SiteURL   string               `json:"-"`
+	T         goi18n.TranslateFunc `json:"-"`
+
+	// DO NOT USE Session field is depricated. MM-26398
```
streamer45, line 24:
```suggestion
	// DO NOT USE Session field is deprecated. MM-26398
```

app/command_loadtest.go
```diff
@@ -339,7 +347,7 @@ func (me *LoadTestProvider) ThreadedPostCommand(a *App, args *model.CommandArgs,
 	}
 
 	client := model.NewAPIv4Client(args.SiteURL)
-	client.MockSession(args.Session.Token)
+	//client.MockSession(args.Session.Token)
```
streamer45, line 350: Is this commented line needed?

crspeller: Woopsie. Yes, a replacement for that is needed.

PR #14884, opened 2020-06-23
[MM-24526] Filter * characters from the search terms in DB

amarjeetanandsingh:
#### Ticket Link
Fixes https://github.com/mattermost/mattermost-server/issues/14438


amarjeetanandsingh:
Hi @ethervoid 
I have refactored the code.

ethervoid:
Thank you! I’ll take a look at it today :)

store/sqlstore/utils.go
```diff
@@ -52,3 +53,26 @@ func finalizeTransaction(transaction *gorp.Transaction) {
 		mlog.Error("Failed to rollback transaction", mlog.Err(err))
 	}
 }
+
+// removeNonAlphaNumericTerms removes all words that only contains non-alphanumeric chars from given line
+func removeNonAlphaNumericTerms(line string) string {
```
ethervoid, line 58: I'd include the character separator, for the split, as an argument so this function is more generic

store/sqlstore/utils.go
```diff
@@ -52,3 +53,26 @@ func finalizeTransaction(transaction *gorp.Transaction) {
...
+		if containsAlphaNumericChar(w) {
+			filteredResult = append(filteredResult, strings.TrimSpace(w))
+		}
+	}
+	return strings.Join(filteredResult, " ")
+}
+
+// isAlphaNumericWord returns true in case any letter or digit is present, false otherwise
+func containsAlphaNumericChar(s string) bool {
+	for _, r := range s {
```
ethervoid, line 72: Probably you need to work with runes. We support languages like Chinese or Japanese

store/sqlstore/utils_test.go
```diff
@@ -76,3 +76,26 @@ func TestSanitizeSearchTerm(t *testing.T) {
 	result = sanitizeSearchTerm(term, "*")
 	require.Equal(t, result, expected)
 }
+
+func TestFilterNonAlphaNumericWords(t *testing.T) {
```
ethervoid, line 80: Test with other languages or unicode characters

store/sqlstore/post_store.go
```diff
@@ -1745,11 +1744,9 @@ func (s *SqlPostStore) SearchPostsInTeamForUser(paramsList []*model.SearchParams
 	pchan := make(chan store.StoreResult, len(paramsList))
 
 	for _, params := range paramsList {
-		// Don't allow users to search for everything.
-		if params.Terms == "*" {
-			continue
-		}
-
+		// remove any term that contains only non-alphanumeric chars
+		// ex: "abcd ** ef*g &^!" >> "abcd ef*g"
```
ethervoid, line 1748: @thefactremains Our objective here is to remove non-alphanumeric strings of characters for the post search to avoid problems in the DB search. Do you see any problem or reduction of functionality doing this?

store/sqlstore/utils.go
amarjeetanandsingh: Nice suggestion @ethervoid. I have implemented that in [this commit](https://github.com/amarjeetanandsingh/mattermost-server/commit/47556c0719ea24c604cf590ad538e677a1240175).

store/sqlstore/utils.go
amarjeetanandsingh: > Range loops
> ...
> We've seen what happens with a regular for loop. A for range loop, by contrast, decodes one UTF-8-encoded rune on each iteration. Each time around the loop, the index of the loop is the starting position of the current rune, measured in bytes, and the code point is its value. Here's an example using yet another handy Printf format, %#U, which shows the code point's Unicode value and its printed representation:

As [doc](https://blog.golang.org/strings) says, `for` loop with `range over string` handles the rune

store/sqlstore/utils_test.go
amarjeetanandsingh: Added tests in Japanese & Chinese in [this commit](https://github.com/amarjeetanandsingh/mattermost-server/commit/47556c0719ea24c604cf590ad538e677a1240175)

store/sqlstore/utils.go
ethervoid: Cool. Just wanted to be sure. I didn't work with runes too much :D

ethervoid:
@amarjeetanandsingh You probably should revert the `go.mod` changes. Those are related to our enterprise code and are making the test fail :D

amarjeetanandsingh:
oh.. Sorry @ethervoid 
That's side effect of `git mod tidy` that cleans up the dependencies. Have reverted that.

ethervoid:
@amarjeetanandsingh Yes, no problem. Is a known issue with dynamic links :D

store/sqlstore/post_store.go
thefactremains: @ethervoid I want to make sure I understand. Search strings that contain ONLY non-alphanumeric characters will be removed?

ethervoid: Yes. For example having a search string “test 123 **^%” should end as “test 123”

store/sqlstore/post_store.go
```diff
@@ -6,18 +6,17 @@ package sqlstore
...
 	"github.com/mattermost/mattermost-server/v5/einterfaces"
 	"github.com/mattermost/mattermost-server/v5/mlog"
 	"github.com/mattermost/mattermost-server/v5/model"
 	"github.com/mattermost/mattermost-server/v5/store"
 	"github.com/mattermost/mattermost-server/v5/utils"
+	"net/http"
+	"regexp"
+	"strconv"
+	"strings"
+	"sync"
```
mgdelacroix, line 19: These imports should go in the first block, with the rest of the standard library imports

store/sqlstore/post_store.go
thefactremains: I cannot think of a use-case where a user needs to search by non-alphanumeric strings, but that doesn't mean one doesn't exist. I would prefer to leave the capability to search for non-alphanumeric strings (if the quotes are required to do so, that would be acceptable).

ethervoid: Ok. @amarjeetanandsingh then if the non-alphanumeric terms are enclosed with quotes, we should avoid the removal. Thank you @thefactremains

amarjeetanandsingh: > enclosed with quotes

@ethervoid It can be single quote `'` or double quote `"` both?

ethervoid: Double quote is what we're using to treat a term as literal

amarjeetanandsingh: Ok.
Thanks for clarification. Will make the changes tonight.

ethervoid: Great. Thank you!

store/sqlstore/post_store.go
amarjeetanandsingh: Thanks @mgdelacroix.
I have reverted the import statement changes [here](https://github.com/mattermost/mattermost-server/pull/14884/commits/11c082b2bb48a58773d588d94215f024588d8188)

store/sqlstore/post_store.go
amarjeetanandsingh: @ethervoid 
I have completed the [changes](https://github.com/amarjeetanandsingh/mattermost-server/commit/81ac7f614c48dbb7254966cd298181e89a92c631).

store/sqlstore/utils_test.go
```diff
@@ -76,3 +76,61 @@ func TestSanitizeSearchTerm(t *testing.T) {
...
 }
+
+func TestRemoveNonAlphaNumericUnquotedTerms(t *testing.T) {
+	const (
+		sep           = " "
+		chineseHello  = "你好"
+		japaneseHello = "こんにちは"
+	)
+
+	terms := ""
```
ethervoid, line 87: Given that we have multiple test cases in the same Test I'll go with structured testing cases like [this example](https://github.com/mattermost/mattermost-server/blob/ad68af10dfcf9d973349f89d025cb5729771a0bf/model/utils_test.go#L51-L72) so we can improve readability and addition of new tests if needed

amarjeetanandsingh: Thanks @ethervoid. 
Table driven test is actually better option here.
I have done the [changes](https://github.com/mattermost/mattermost-server/pull/14884/commits/88ca4510286fe296557d1810d9cf67f39065636b).

ethervoid:
Is not stale. Waiting for @ogi-m review

jasonblais:
@ogi-m friendly reminder to help with reviews, thanks!

PR #14885, opened 2020-06-23
[MM-26420] Adds wildcard support to Bleve

mgdelacroix:
#### Ticket Link
https://mattermost.atlassian.net/browse/MM-26420


store/searchtest/post_layer.go
```diff
@@ -111,7 +111,7 @@ var searchPostStoreTests = []searchTest{
 	{
 		Name: "Should support search with wildcards",
 		Fn:   testSupportWildcards,
-		Tags: []string{ENGINE_POSTGRES, ENGINE_MYSQL, ENGINE_ELASTICSEARCH},
+		Tags: []string{ENGINE_ALL},
```
jespino, line 114: :heart:

services/searchengine/bleveengine/search.go
```diff
@@ -156,10 +156,23 @@ func (b *BleveEngine) SearchPosts(channels *model.ChannelList, searchParams []*m
 			}
 		} else {
 			if len(params.Terms) > 0 {
-				messageQ := bleve.NewMatchQuery(params.Terms)
-				messageQ.SetField("Message")
-				messageQ.SetOperator(termOperator)
-				termQueries = append(termQueries, messageQ)
+				terms := []string{}
+				for _, term := range strings.Split(params.Terms, " ") {
```
ethervoid, line 160: Why do aggregate the queries in this for loop instead of doing here the wildcard query addition and after that the regular terms one?

store/searchtest/post_layer.go
ethervoid: :heart:

services/searchengine/bleveengine/search.go
mgdelacroix: `params.Terms` can contain a value such as `sear* post` (in the test example), so first to distinguish if we have to run a wildcard query, we need to isolate each term and check if the wildcard is the suffix, and then, if we find a term with a wildcard, we have to exclude it from the normal `MatchQuery` in case the `termOperator` is and, because if it's treated as a literal, it will never be found

mgdelacroix:
@ogi-m gentle reminder to take a look at this :)

ogi-m (approved):
LGTM 👍

mgdelacroix:
/cherry-pick-5.28

PR #14886, opened 2020-06-23
Better handling of errors on trial license request

jespino:
#### Summary
Now we check that your SiteURL is set before we request the trial license, and
we verify that the response from the customers portal is the right one.

#### Ticket Link
[MM-26413](https://mattermost.atlassian.net/browse/MM-26413)


i18n/en.json
```diff
@@ -1508,6 +1508,10 @@
     "id": "api.license.request_trial_license.app_error",
     "translation": "Unable to get a trial license, please try again or contact with support@mattermost.com."
   },
+  {
+    "id": "api.license.request_trial_license.no-site-url.app_error",
+    "translation": "Unable to request a trial license, please configure your SiteURL first in the web server section of the mattermost configuration."
```
saturninoabril, line 1513: Can we get approval of this error message from PM or technical writer?

jespino: I guess so

jespino:
@saturninoabril I'll fix the misalignment in this pr

i18n/en.json
jespino: @justinegeffen can you review this text?

jespino:
@saturninoabril sorry, yes, this need a PR in frontend code, probably I going to reuse other related to change some texts.

i18n/en.json
justinegeffen, line 1513:
```suggestion
    "translation": "Unable to request a trial license. Please configure a Site URL in the web server section of the Mattermost System Console."
```

justinegeffen: Thanks, @jespino. Just a few changes. I'm 3/5 on "your Site URL" and "a Site URL". Which do you think works better?

jespino: I think both works pretty well. by the way, this message is only going to be seen as part of the API response, because some design changes, the error is not shown in the user interface so... I going to pick the `a Site URL` option.

PR #14895, opened 2020-06-23
[MM-25568] LDAP Sync skip users and log when they dont match the restricted email of a team or channel

fmunshi:
#### Summary
- Skips adding a user to a team or channel if it errors due to a restricted domain, previously the entire sync job would fail but this makes it so that users that _can_ be synced will be synced and users that _cannot_ will be skipped and a message will be posted to the server logs

#### Ticket Link
- https://mattermost.atlassian.net/browse/MM-25568

Screen Shot 2020-06-23 at 1 06 21 PM


app/syncables.go
```diff
@@ -36,6 +36,15 @@ func (a *App) createDefaultChannelMemberships(since int64, channelID *string) er
 		if tmem == nil {
 			_, err = a.AddTeamMember(channel.TeamId, userChannel.UserID)
 			if err != nil {
+				if err.Id == "api.team.join_user_to_team.allowed_domains.app_error" {
+					a.Log().Info("Not adding user to channel because the domain associated with the account is not permitted in the given team",
```
fmunshi, line 40: @thefactremains What do you think of this message? Any changes you want to it?

app/syncables.go
```diff
@@ -36,6 +36,15 @@ func (a *App) createDefaultChannelMemberships(since int64, channelID *string) er
...
+						mlog.String("user_id", userChannel.UserID),
+						mlog.String("channel_id", userChannel.ChannelID),
+						mlog.String("team_id", channel.TeamId),
+					)
+
```
sbishel, line 45: Nit - unnecessary empty line

app/syncables.go
thefactremains: I prefer this:
**User not added to channel - the domain associated with the user is not in the list of allowed team domains**

fmunshi: Updated @thefactremains

fmunshi:
> A couple of questions:
> 
> 1. Have you tested this against both the LDAP sync and the _first login_ flow? The first login is where a user without an account logs in for the first time.

- It just skips any teams that have restricted domains and puts the following message in the system console for every channel/team skipped 

Screen Shot 2020-06-25 at 11 29 46 AM


> 2. What does this display in the LDAP sync table?

- Nothing the only way to know a user was skipped for this reason is to check the logs... we can add a "warning" state to the ldap sync table as part of our eventual refactor

mkraft:
> > A couple of questions:
> > 
> > 1. Have you tested this against both the LDAP sync and the _first login_ flow? The first login is where a user without an account logs in for the first time.
> 
> * It just skips any teams that have restricted domains and puts the following message in the system console for every channel/team skipped
> 
> Screen Shot 2020-06-25 at 11 29 46 AM
> 
> > 1. What does this display in the LDAP sync table?
> 
> * Nothing out of the ordinary
>   Screen Shot 2020-06-25 at 11 31 52 AM
> * the only way to know a user was skipped for this reason is to check the logs... we can add a "warning" state to the ldap sync table as part of our eventual refactor

re #1) sounds good

re #2) I have created https://mattermost.atlassian.net/browse/MM-26508

PR #14366, opened 2020-04-24
Pass query string params to interactive dialog request url

Ashniu123:
Closes #14347 on Github
Closes #24024 on Jira

#### Summary
Query string parameters are now passed onto the interactive dialog request url

#### Ticket Link
Github issue: https://github.com/mattermost/mattermost-server/issues/14347
Jira: https://mattermost.atlassian.net/browse/MM-24024

PR #14897, opened 2020-06-23
MM-26441 - fix for doPluginRequest

cpoile:
#### Summary
- Fix for a bug accidentally introduced by #14366 (because #14557 was merged while it was under review): the provided values were being overwritten by the rawQuery params. 
- Now we're merging them both.

#### Ticket Link
- https://mattermost.atlassian.net/browse/MM-26441

### DIFF
--- a/api4/command.go
+++ b/api4/command.go
@@ -332,8 +332,8 @@ func executeCommand(c *Context, w http.ResponseWriter, r *http.Request) {
 
 	commandArgs.UserId = c.App.Session().UserId
 	commandArgs.T = c.App.T
-	commandArgs.Session = *c.App.Session()
 	commandArgs.SiteURL = c.GetSiteURLHeader()
+	commandArgs.Session = *c.App.Session()
 
 	auditRec.AddMeta("commandargs", commandArgs) // overwrite in case teamid changed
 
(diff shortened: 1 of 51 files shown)


cpoile:
@lieut-data Done!  (Incidentally, unless I'm mistaken, I think we're missing tests for dynamic autocomplete (which would have caught this bug, too--because the `user_input` and `parsed` params were being lost). But these tests will catch such a regression in the future.

cpoile:
@DHaussermann Considering this is a fix for a previous PR (which couldn't be tested), but now has unit tests, can we skip manual tests again?

cpoile:
@amyblais Can this be cherry picked for 5.25?

amyblais:
Yes, sure.

DHaussermann:
@iomodo  or @cpoile Can you pleae provide guidance or steps for testing on this?  Did the related issue introduce a problem with slash command auto-complete that I can see and validate with functional testing?

cpoile:
@DHaussermann It would be very difficult to test without a bunch of coding (or impossible), since it only pops up in the middle of an dynamic autocomplete call. That's why I think the unit tests are as good as we're going go get for this.

cpoile:
@metanerd Thank you so much!

PR #14898, opened 2020-06-23
MM-26056 Add plugin ability to execute slash command.

crspeller:
#### Summary
Adds the ability for a plugin to execute a slash command. 

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-26056


app/plugin_api.go
```diff
@@ -70,6 +71,19 @@ func (api *PluginAPI) UnregisterCommand(teamId, trigger string) error {
 	return nil
 }
 
+func (api *PluginAPI) ExecuteSlashCommand(commandArgs *model.CommandArgs) (*model.CommandResponse, error) {
```
hanzei, line 74: I'm a bit worried that this naming collides  with the Hook  `ExecuteCommand`. Maybe something like `TriggerSlashCommand`?

hanzei, line 74: We are using `*model.AppError` everywhere in the plugin API. I don't think it's worth to introduce an inconsistency now. (I still dream of a world without  `*model.AppError`)

app/plugin_api.go
```diff
@@ -70,6 +71,19 @@ func (api *PluginAPI) UnregisterCommand(teamId, trigger string) error {
...
+	user, err := api.app.GetUser(commandArgs.UserId)
+	if err != nil {
+		return nil, err
+	}
+	commandArgs.T = utils.GetUserTranslations(user.Locale)
```
hanzei, line 79: Should the server take care of setting 'commandArgs.SiteURL`?

app/plugin_api.go
cpoile: I can see both arguments. At least `ExecuteCommand` is a hook, so that might be enough to remove ambiguity?

crspeller: Ya, not totally happy ether. I originally had it called `ExecuteCommand` then had the obvious conflict. I like ` ExecuteSlashCommand` over `TriggerSlashCommand` because it matches what we call that server side. Since it doesn't directly conflict with the hook I think it's fine.

app/plugin_api_test.go
```diff
@@ -1577,3 +1577,24 @@ func TestPluginAPIGetPostsForChannel(t *testing.T) {
...
+	t.Run("run invite command", func(t *testing.T) {
+		_, err := api.ExecuteSlashCommand(&model.CommandArgs{
+			Command:   "/invite @" + newUser.Username,
+			TeamId:    th.BasicTeam.Id,
+			UserId:    th.BasicUser.Id,
+			ChannelId: th.BasicChannel.Id,
+		})
+		require.NoError(t, err)
+		_, err2 := th.App.GetChannelMember(th.BasicChannel.Id, newUser.Id)
+		require.Nil(t, err2)
```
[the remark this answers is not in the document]
crspeller: `NoError` for when it's just a plain error. `Nil` for when `*model.AppError` is wreaking your day.

app/plugin_api.go
crspeller: Ya. I think so. Will fix.

app/plugin_api.go
crspeller: I don't really buy the consistency argument here. There are other APIs already that return just error and it really doesn't give us anything in this case.

hanzei: The only API endpoint that returns a plain error is `GetBundlePath()`. All the other ones return `*model.AppError`. Given that a mix in error types is more confusing the just using the same type one all over the place, I'm 3/5 on using `*model.AppError` here.

hanzei: What I like about `TriggerSlashCommand` is that it makes clear that the caller only _triggers_ a slash command. Other then with `ExecuteCommand`, where the plugin has to provide the implementation.

For a new developers trying out the Plugin toolkit it might not be obvious what the difference between a hook and an API call is here.

crspeller: No it's not. `RegisterCommand` and `UnregisterCommand` the other command APIs return `error`. I am being consistent with those APIs.

crspeller: As far as I am concerned `Trigger` and `Execute` in this context are synonyms. I agree that this might be confusing, but I don't see how to handle that other than docs. I would rather be internally consistent.

crspeller:
@hanzei Kind review reminder.

app/plugin_api.go
hanzei: To me `Execute` means "Do something" while `Trigger` means "Make someone do something".

What is the reasoning for valuing internal consistency vs external consistency? Given that this is an API endpoint I would rate a clear and easy to use API higher then internal consistency.

Can we rename our internal code to align with the external consistency?

hanzei: Good point, I forgot about `RegisterCommand' and `UnregisterCommand`. Still, these are only two API methods among many.

Actually, returning `*model.AppError` does provide value here. The message shown in the webapp is the one under `Id` while the logged error is `Error()`. A plugin might want to know which message was shown in the webapp and hence needs `Id`.

lieut-data: Given we resolve all inconsistencies with the plugin-api wrapper, I'm not sure the return signature matters too much. I'd personally lean towards returning `error` always, since there's no reason to propagate an earlier mistake and we're already inconsistent.

But one comment below on the return value.

app/plugin_api.go
```diff
@@ -70,6 +71,19 @@ func (api *PluginAPI) UnregisterCommand(teamId, trigger string) error {
...
+	response, appErr := api.app.ExecuteCommand(commandArgs)
+	if appErr != nil {
+		return response, fmt.Errorf(appErr.Error())
```
lieut-data, line 83: Why don't we just return `appErr` directly here? Also, I think `err` is already a `*model.AppError`, so we can even reuse the variable name.

app/plugin_api.go
lieut-data: As @cpoile noted, I think hooks and APIs are sufficiently distinct so as to not worry about a conflict over the name `ExecuteCommand`, especially since the author must call one and implement a receiver function for the other.

I also note the REST API explicitly [calls this endpoint](https://api.mattermost.com/#tag/commands/paths/~1commands~1execute/post) `Execute Command`. I don't see anywhere where `trigger` is used as a verb: it's exclusively referring to the `trigger word` in a slash command, so I'd be hesitant to adopt new terminology for this particular change.

Can we just keep this as `ExecuteCommand`, and potentially iterate on improved semantics in the plugin-api wrapper in the future?

app/plugin_api.go
crspeller: Good point. I think I was just trying to handle appErr appropriately but the null checks work fine. Changed them all to `appErr`

PR #14906, opened 2020-06-24
POC for mobile plugins

larkox:
#### Summary
This is a POC to allow plugins to work on mobile.

The main idea is to register the plugin for a "location". Depending on the location (channel header, post menu, etc...) the plugin will have a set of extras. In the implemented case (channel header) we will only pass the text to show (we could also add an icon in future implementations). We also pass a "trigger".

The trigger is just what to do when an action is performed (for channel header, a tap on the menu element). This will call the url `https://mattermostURL/plugins/pluginID/mobile/trigger`. Here the response will have what action to do among a subset of actions. The current implemented is RESET_TO_CHANNEL, which goes back to the channel, but we may have others like "GO_TO" to go to a different part of the app, or "OPEN_URL" to open an URL.

The plugin will register these triggers the same way slash commands are registered.

The idea is to add little by little more locations in the mobile app and extend what can be done with the triggers. Eventually, I would like to populate complete views thanks to this (i.e. RHS equivalent on mobile).

Disclaimer 1: I have not been very consistent with the naming or comments. When we get to a consensus on naming I will make sure it is applied.

Disclaimer 2: I haven't used an `interface{}` in `MobileTrigger` because I was having problems with the RPC. Any help on that front would be much appreciated. 

This is partly based on Maria's idea of Plugin Actions (https://docs.google.com/document/d/138uxYyUdjFFecB-tD3JWV7dTVxBBQjIo-cX7pGblI3E/edit). Here the main differences:

- Plugin Actions is designed to work also for WebApp. I preferred to focus on mobile, since that is where the problem resides. Nevertheless, it would be simple to make it work also on WebApp.
- My take tries to be more flexible, in order to add more locations as we go. For example, not all locations may need an icon, or a text.
- Type and Location would be the same concept.
- RequestURL and Trigger have a similar concept
- I did not consider permissions for my implementation, but it is something to keep in mind.
- Result actions are present in both ideas
- When fetching from the server, we fetch all plugins locations. That way we have everything cached for when it needs to be used.
- The content of what we fetch will be dependent on what is the location

#### Ticket Link
None

#### Related PRs
Webapp: https://github.com/mattermost/mattermost-webapp/pull/5887
Redux: https://github.com/mattermost/mattermost-redux/pull/1189
Zoom plugin: https://github.com/mattermost/mattermost-plugin-zoom/pull/139
Mobile App: https://github.com/mattermost/mattermost-mobile/pull/4480


model/mobile_trigger.go
```diff
@@ -0,0 +1,43 @@
...
+package model
+
+import (
+	"encoding/json"
+	"io"
+)
+
+type MobileTrigger struct {
+	PluginId string `json:"id"`
+	Location string `json:"location"`
```
hanzei, line 13: My opinion might be caused by understand for the English language: To me `Location` is a fixed place. But the trigger might be shown in different locations based on the used client. `Type` captures more that it's about "what" kind of Trigger.

model/mobile_trigger.go
```diff
@@ -0,0 +1,43 @@
...
+	Trigger  string `json:"trigger"`
+	//	Extra    interface{} `json:"extra"`
+	Extra string `json:"extra"` // Should use interface{} to be able to create different extras for different locations
+}
```
hanzei, line 17: I would like to see a `Request URL` field here. It would allow plugins (and potential other integrations) to specify there API themself and don't rely on a pre-defined pattern. It also follows what we have been doing with e.g. post actions.

model/mobile_trigger.go
```diff
@@ -0,0 +1,43 @@
...
+// See LICENSE.txt for license information.
+
+package model
+
+import (
+	"encoding/json"
+	"io"
+)
+
+type MobileTrigger struct {
```
hanzei, line 11: Maybe a general `Text` and `Icon` field make sense. For post menu actions and all dropdown that seams useful.

model/mobile_trigger.go
hanzei, line 17: I like the `Scope` idea from Maria, but that could also be done in a different PR. 1/5 that we should rather add it now, as it's easy to implment.

model/mobile_trigger.go
larkox: Probably the problem is calling Trigger to this object. I changed it to MobileIntegration. Each integration has a location.

api4/api.go
```diff
@@ -59,6 +59,8 @@ type Routes struct {
 	Plugins *mux.Router // 'api/v4/plugins'
 	Plugin  *mux.Router // 'api/v4/plugins/{plugin_id:[A-Za-z0-9\\_\\-\\.]+}'
 
+	MobilePlugins *mux.Router // 'api/v4/mobilePlugins'
```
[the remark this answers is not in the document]
larkox: I changed that to pluginIntegrations.

model/mobile_trigger.go
larkox: I changed it again to PluginIntegration, to allow scopes.

model/mobile_trigger.go
larkox: Yes, it is useful for all dropdowns. Once the functionality is more mature and I get to deal with the RPC problems, we can define properly each location.

model/mobile_trigger.go
```diff
@@ -0,0 +1,49 @@
...
+	PluginID   string   `json:"id"`
+	Location   string   `json:"location"`
+	RequestURL string   `json:"request_url"`
+	Scope      []string `json:"scope"`
+	//	Extra    interface{} `json:"extra"`
+	Extra string `json:"extra"` // Should use interface{} to be able to create different extras for different locations
+}
+
+type MobileIntegrationChannelHeader struct {
+	DefaultMessage string `json:"default_message"`
```
marianunez, line 26: How are these specific location structs intended to be used? Is this to be set in the `Extra` field?

I believe it would be simpler to initially add the `text` (and `icon`) directly to the plugin action struct that would probably be commonly used for most  action and once we have more usage we can think of abstracting into this or a concept of custom metadata.
It would also simplify at the client level of not having to special case handling to process the different type of data that the plugin action definition might have for something like text that most of them will use.

api4/api.go
```diff
@@ -59,6 +59,8 @@ type Routes struct {
 	Plugins *mux.Router // 'api/v4/plugins'
 	Plugin  *mux.Router // 'api/v4/plugins/{plugin_id:[A-Za-z0-9\\_\\-\\.]+}'
 
+	PluginIntegrations *mux.Router // 'api/v4/pluginIntegrations'
```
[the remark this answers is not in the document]
larkox: I prefer to avoid clashing urls. since below `api/v4/plugins/` anything can be a plugin id, there might be a plugin with id `integrations`.
Regarding naming, I am 100% flexible. I just want to keep in mind that the idea of this is not just have actions, but potentially full views. That is the reason I thought about "integrations".

app/plugin_mobile_triggers.go
```diff
@@ -0,0 +1,71 @@
...
+	PluginID    string
+}
+
+func (a *App) RegisterPluginIntegration(pluginID string, integration *model.PluginIntegration) error {
+	a.Srv().pluginIntegrationsLock.Lock()
+	defer a.Srv().pluginIntegrationsLock.Unlock()
+
+	integration.PluginID = pluginID
+	for _, pt := range a.Srv().pluginIntegrations {
+		if pt.Integration.RequestURL == integration.RequestURL && pt.Integration.Location == integration.Location {
```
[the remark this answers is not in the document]
larkox: My reason to not have a random generated ID is that I preferred the triggers to be unique (i.e. if I register twice a trigger for the same location and the same trigger it is the same trigger). Nevertheless, I will think about it again, and probably it makes sense to just leave that kind of checks to the developer and just deal with the ID.

model/mobile_trigger.go
larkox: Yes, I am having some problems with RPC, and that is why it is not implemented as intended.
The main idea is to have for each "integration" (I should have changed the name of this one) a struct to pass to extra. For channel header, it could be something like:
- Icon
- Text
- HasToggle
- CountURL

Meanwhile, for post menu would be just
- Icon
- Text

And on one of the most ambitious cases, a "RHS view" would be:
- HeaderTitle
- RootComponent

I know many will have text and icon, but not all of them will have, and I want to keep the open mind.

api4/api.go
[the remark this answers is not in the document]
hanzei: I still like to come up with a name that doesn't include an existing concept like `plugins`. Plugins are tied to the rpc layer, which this whole idea is not. It's HTTP only.

Because of that, I'm strongly think a new route like `api/v4/triggers` or `api/v4/actions` should be used.

model/mobile_trigger.go
```diff
@@ -0,0 +1,49 @@
...
+	"encoding/json"
+	"io"
+)
+
+const (
+	ScopeMobile = "mobile"
+	ScopeWebApp = "webapp"
+)
+
+type PluginIntegration struct {
```
hanzei, line 16: The struct name should be consistent with the route. Blocking here until https://github.com/mattermost/mattermost-server/pull/14906/files#r449028540 is resolved.

hanzei:
Could you elaborate a bit what you mean with views e.g. `RHS view`? Do you want to allow the plugin to serve an iframe to the app? Did you test that approach with the mobile app?

model/mobile_trigger.go
hanzei: I like @marianunez's idea of starting with simple `text` and `icon` fields inside of the main `struct` and iterate on it once we now the other use cases. `text` and `icon` will cover most cases for now. We can still add an `Extra` field later, if we need it. And it fixes the RPC problems.

larkox: As discussed offline, I want to keep the doors open to any possible location. There are several locations considered on the webapp, and I would like the possibility to reflect that with this. Here is the list if I am not mistaken:
- Root Component
- Popover User Attributes
- Popover User Actions
- Left Sidebar Header
- Bottom Team Sidebar
- Post Message Attachment
- Link Tooltip
- Channel Header Button
- Post Type
- Post Card Type
- Main Menu Action
- Channel Header Menu Action
- Post Dropdown Menu Action (and submenus)
- File Upload Method
- Admin Console Custom Setting
- RHS Component
- File preview

And also there are many "actions" that can be considered a "plugin integration".
- File will upload
- Websockets events
- Reconnect handler
- Message will be posted
- Slash command will be posted
- Message will format

Some of them are more easy to implement on this context, some are more complicated. We can decide on just a few to do the first version (the most used, the easiest, or anything like that), but as you see, not all would need a text or an icon.
I don't want to over-scope this functionality, but I want to keep the door open to extend it if needed to cover all the cases.

larkox:
Added the PRs for webapp and Redux to allow this to be used on Webapp. Check the PR description.

larkox:
Next steps: Waiting for UX feedback on what we would like to be able to do on version 1.0 of this feature, in order to better engineer the solution and deal with the multiple possible concerns.

lieut-data:
@larkox, I'd be happy to give a review when you think it's time -- going to remove my name from the list for now.

larkox:
Apart of naming, I consider this PR ready to "review and merge".

plugin/api.go
```diff
@@ -37,6 +37,16 @@ type API interface {
...
 
+	// RegisterPluginIntegration registers a custom integration from the plugin.
+	//
+	// Minimum server version: 5.28
+	RegisterPluginIntegration(trigger *model.PluginIntegration) error
+
+	// UnregisterPluginIntegration unregisters a custom integration previously registered via RegisterPluginIntegration.
+	//
+	// Minimum server version: 5.28
+	UnregisterPluginIntegration(location, trigger string) error
```
lieut-data, line 48: Is this supposed to be `requestURL` as a parameter? (The word `trigger` is unexpected in a few places in the proposed changes.)

model/plugin_integration.go
```diff
@@ -0,0 +1,59 @@
...
+)
+
+const (
+	ScopeMobile = "mobile"
+	ScopeWebApp = "webapp"
+)
+
+type PluginIntegration struct {
+	PluginID   string      `json:"id"`
+	Location   string      `json:"location"`
```
lieut-data, line 18: At a glance, it's not super clear how to interpret `Location` and `RequestURL` -- could we add GoDoc here?

plugin/api.go
larkox: Yes, `requestURL` should be the right parameter name at the moment. Naming is a bit chaotic at the moment, and I still haven't put enough effort on solving that waiting for a concensus on the naming (they may not be even called "pluginIntegrations").

model/plugin_integration.go
larkox: Sounds nice. Nevertheless, if you have any proposal for better naming, I am all in. I am not specially good at naming, and the naming of this whole infrastructure is still open to discussion.

lieut-data: Final naming aside (and given that I'm only looking at this review), I currently can't tell you what these two fields do. Maybe even just write a doc using the current name, `Location define where the plugin should ...` and `RequestURL is used by the plugin to ...`.

api4/api.go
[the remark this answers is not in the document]
marianunez: We should converge on the name here before we merge this given that it would be a breaking change to the API if we want to update to something else down the line.

plugin/api.go
larkox: Addressed on 3f8023b

model/plugin_integration.go
larkox: Addressed on 3f8023b

model/plugin_integration.go
```diff
@@ -0,0 +1,62 @@
...
+
+type PluginIntegration struct {
+	PluginID string `json:"id"`
+	// Location defines where the client should present this integration
+	Location string `json:"location"`
+	// RequestURL defines the URL the client will reach to perform the integration action
+	RequestURL string `json:"request_url"`
+	// Scope defines which clients should show this integration
+	Scope []string    `json:"scope"`
+	Extra interface{} `json:"extra"`
```
hanzei, line 24: Given that every implementation here and all the examples contain `Icon` and `Text`, I'm wondering if we should remove `Extra` and move `Icon` and `Text` into `PluginIntegration`.

hanzei (approved):
From the discussion at https://community-daily.mattermost.com/core/pl/rayya4fak3bujpedpmqekjsf4r `action` seams like the way to go. But naming can be addressed later.

model/mobile_trigger.go
marianunez: Is `PluginIntegration` the name that was decided?

larkox:
Closing as this effort is being now driven by the "Apps" effort.

PR #14907, opened 2020-06-24
changing license renewal link to point to new renewal webpage

esadur:
#### Summary

Changing the license renewal link to point to the new renewal webpage. It currently points to an old wufoo form.

#### Ticket Link

https://mattermost.atlassian.net/browse/MM-26486


jwilander:
@esadur this link gets appended with `?id=<license-id>` when it's sent in the email. Shouldn't cause any problems but just wanted to make sure you're aware

esadur:
@jwilander is the license ID appended for some purpose? Doesn't seem like it (or any part of the query string) gets passed through the link redirect to the wufoo form.

esadur:
Does the link need to be updated in mmctl too? https://github.com/mattermost/mmctl/blob/master/vendor/github.com/mattermost/mattermost-server/v5/model/license.go

jwilander:
> Does the link need to be updated in mmctl too? https://github.com/mattermost/mmctl/blob/master/vendor/github.com/mattermost/mattermost-server/v5/model/license.go

No, that imports this

jwilander:
> @jwilander is the license ID appended for some purpose? Doesn't seem like it (or any part of the query string) gets passed through the link redirect to the wufoo form.

I believe the idea was to use it but we never ended up doing it

esadur:
@jwilander you merged https://github.com/mattermost/mattermost-webapp/pull/5796, now I think this one has to merge as well?

PR #14910, opened 2020-06-25
Fix webapp checkout part 2

metanerd:
#### Summary

The pipeline errors if the commit for the webapp does not yet exist in the bucket for uploading the webapp PRs. 
Which means the if else did not work as expected. Instead now we use `when: on_fail`, which basically uses `trap` underneath so it still errors, but initiates the npm commands (so part of the webapp pipeline).

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -28,12 +28,9 @@ jobs:
           export WEBAPP_GIT_COMMIT=$(git rev-parse HEAD)
           echo "$WEBAPP_GIT_COMMIT"
 
+          trap 'npm ci && make build' ERR
           curl -f -o ./dist.tar.gz https://pr-builds.mattermost.com/mattermost-webapp/commit/${WEBAPP_GIT_COMMIT}/mattermost-webapp.tar.gz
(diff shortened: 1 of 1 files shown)


.circleci/config.yml
```diff
@@ -29,12 +29,11 @@ jobs:
           echo "$WEBAPP_GIT_COMMIT"
 
           curl -f -o ./dist.tar.gz https://pr-builds.mattermost.com/mattermost-webapp/commit/${WEBAPP_GIT_COMMIT}/mattermost-webapp.tar.gz
-          if [[ $? == 0 ]]
-          then
-            mkdir ./dist && tar -xvf ./dist.tar.gz -C ./dist --strip-components=1
-          else
-            npm ci && make build
-          fi
+          mkdir ./dist && tar -xvf ./dist.tar.gz -C ./dist --strip-components=1
```
agnivade, line 32: This is fine, but it is not an ideal fix. The problem with this approach is that if any of the commands fail, it will try to execute the on_fail step.

metanerd: OK, so I created the trap before the command so it only triggers once it reached the command.

metanerd:
Yes, you are correct. I should probably remove the trap afterwards.

PR #14911, opened 2020-06-25
Fix the error message from get oldest entity in ES indexer

ethervoid:
#### Summary

This is a fix for this PR that should've included this message


ethervoid:
To test this fix should be enough to do some smoke testing around the ES reindex feature

PR #14913, opened 2020-06-25
[MM-26507] Send license type for install plugin request

hanzei:
#### Summary
Also send `BuildEnterpriseReady` and `EnterprisePlugins` in install plugin requests.

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-26507


app/plugin.go
```diff
@@ -655,6 +646,23 @@ func (a *App) mergeLocalPlugins(remoteMarketplacePlugins map[string]*model.Marke
 	return nil
 }
 
+func (a *App) getBaseMarketplaceFilter() *model.MarketplacePluginFilter {
+	filter := &model.MarketplacePluginFilter{
+		ServerVersion: model.CurrentVersion,
```
ali-farooq0, line 651: Not sure if this is cleaner or what you have below, but just a suggestion 🤷‍♂️.
```suggestion
		ServerVersion: model.CurrentVersion,
		BuildEnterpriseReady: model.BuildEnterpriseReady == "true",
```

api4/plugin_test.go
```diff
@@ -1371,6 +1371,93 @@ func TestInstallMarketplacePlugin(t *testing.T) {
...
+		})
+
+		l := model.NewTestLicense()
+		// model.NewTestLicense generates a E20 license
+		*l.Features.EnterprisePlugins = false
+		th.App.Srv().SetLicense(l)
+
+		// The content of the request is irrelevant. This test only cares about enterprise_plugins.
+		pRequest := &model.InstallMarketplacePluginRequest{}
+		manifest, resp := th.SystemAdminClient.InstallMarketplacePlugin(pRequest)
```
ali-farooq0, line 1428: One thing these tests are missing is that: we're not asserting if the remote server was actually called. It is entirely possible that the remote request were never made and these tests would still pass because the asserts/requires are inside the `HandlerFunc`.

One way to assert that would be to set a flag inside the `HandlerFunc` and assert that. Or make sure a plugin was successfully installed from the remote server. Either one.

app/plugin.go
hanzei: I'm personally like the way it is right now. But that is very subjective. Objections on leaving it as it is?

api4/plugin_test.go
hanzei: Good point :+1: Added a check to ensure the server is hit.

api4/plugin_test.go
```diff
@@ -1372,6 +1372,8 @@ func TestInstallMarketplacePlugin(t *testing.T) {
 	})
 
 	t.Run("verify EnterprisePlugins is false for TE", func(t *testing.T) {
+		hitServer := false
```
ali-farooq0, line 1375:
```suggestion
		requestHandled := false
```

api4/plugin_test.go
```diff
@@ -1396,9 +1401,12 @@ func TestInstallMarketplacePlugin(t *testing.T) {
 		manifest, resp := th.SystemAdminClient.InstallMarketplacePlugin(pRequest)
 		CheckInternalErrorStatus(t, resp)
 		require.Nil(t, manifest)
+		assert.True(t, hitServer)
 	})
 
 	t.Run("verify EnterprisePlugins is false for E10", func(t *testing.T) {
+		hitServer := false
```
ali-farooq0, line 1408:
```suggestion
		requestHandled := false
```

api4/plugin_test.go
```diff
@@ -1428,9 +1439,12 @@ func TestInstallMarketplacePlugin(t *testing.T) {
 		manifest, resp := th.SystemAdminClient.InstallMarketplacePlugin(pRequest)
 		CheckInternalErrorStatus(t, resp)
 		require.Nil(t, manifest)
+		assert.True(t, hitServer)
 	})
 
 	t.Run("verify EnterprisePlugins is true for E20", func(t *testing.T) {
+		hitServer := false
```
ali-farooq0, line 1446:
```suggestion
		requestHandled := false
```

DHaussermann:
Tested and passed E20 Only licence links are now functional 
- Deployed market place PR https://github.com/mattermost/mattermost-marketplace/pull/79 to my local server
- Configured the server with this PR to point to my local marketplace
- Confirmed target plugin (MS Calendar) was only visible with E20
- Successfully installed using the Install button. 
LGTM!

PR #14916, opened 2020-06-25
MM-26057 - Add CreateCommand plugin API

cpoile:
#### Summary
- Added all the slash command functions to the plugin api. 
- Tested to make sure the commands are being manipulated as expected. Please let me know if you think I should be testing more thoroughly (I'm just testing what we know is different about the calls; I'm leaving all the other functionality to be tested by the existing command tests)

#### Important Notes:
- I added a `PluginId` field to the plugin struct -- this can be the flag to filter plugin commands in the future. And altered the validation. I added the new field to the `newSqlCommandStore` and the upgrade for 5.26.  Please let me know if there is something I missed (this is the first time I've added a field to the db).
- I'm treating the `PluginId` like the `CreatorId` field (you can't update it once the command is created)
- I'm making the assumption that plugins have permissions to do everything. This could be dangerous--a plugin can update or delete any command it wants, even if it is not the creator. Is that okay?
- Couldn't think of a reasonable way to test ListSlashCommands -- first: I can't seem to find a way to make the `app.T` work in a test, and second: `ListCommands` is not tested in the server either. Open to suggestions for this.
- I think testing `GetSlashCommand` or `DeleteSlashCommand` is not worth the test infrastructure :)  But I am using `GetSlashCommand` in the other tests, at least.

#### Ticket Link
- https://mattermost.atlassian.net/browse/MM-26057


model/command.go
```diff
@@ -80,7 +81,13 @@ func (o *Command) IsValid() *AppError {
 		return NewAppError("Command.IsValid", "model.command.is_valid.update_at.app_error", nil, "", http.StatusBadRequest)
 	}
 
-	if !IsValidId(o.CreatorId) {
+	// If the CreatorId is blank, this should be a command created by a plugin.
+	if o.CreatorId == "" && !IsValidPluginId(o.PluginId) {
```
crspeller, line 85: Having the CreatorId blank might break the frontend when it displays which user the slash command is supposed to be created by. If it's a quick fix, can we just fix that? Or we can make a follow up ticket.

plugin/api.go
```diff
@@ -966,6 +966,48 @@ type API interface {
...
+	//
+	// @tag SlashCommand
+	// Minimum server version: 5.26
+	UpdateSlashCommand(commandID string, cmd *model.Command) (*model.Command, *model.AppError)
+
+	// MoveSlashCommand moves a slash command (commandID) to a team (newTeamID).
+	//
+	// @tag SlashCommand
+	// Minimum server version: 5.26
+	MoveSlashCommand(commandID string, newTeamID string) *model.AppError
```
crspeller, line 1004: Is move necessary given the update above? As a useful helper we could add it to https://github.com/mattermost/mattermost-plugin-api

model/command.go
cpoile: Good point. I assume you're talking about the `integrations/commands/installed` list of slash commands? I edited my db, removed the creatorId, and it's fine -- it sends in an empty object if it can't find the creator's user profile: https://github.com/mattermost/mattermost-webapp/blob/1d6ad140c501f21ad5d2c9046845beefdd7fecd3/components/integrations/installed_commands/installed_commands.jsx#L96

If there's any other spots you can think of, I'll test those too.

plugin/api.go
cpoile: Looks like we can't -- `UpdateCommand` doesn't let you move the command: https://github.com/cpoile/mattermost-server/blob/6777a6b7e94f5b99a508d3b36d5191b3d1de3a43/app/command.go#L648 . I guess Joram wanted you to use the move command? :)

cpoile:
@lieut-data gentle ping! :)

app/plugin_api.go
```diff
@@ -914,3 +914,59 @@ func (api *PluginAPI) PluginHTTP(request *http.Request) *http.Response {
...
+}
+
+func (api *PluginAPI) ListSlashCommands(teamID string, customOnly bool) ([]*model.Command, *model.AppError) {
+	var commands []*model.Command
+	var appErr *model.AppError
+
+	if customOnly {
+		commands, appErr = api.app.ListTeamCommands(teamID)
+	} else {
+		commands, appErr = api.app.ListAllCommands(teamID, api.app.T)
```
lieut-data, line 934: Off topic for this review, but the presence of `api.app.T` at this layer surprises me. Do we support translating the built-in slash commands to the language of the requesting user? How should that translate into a plugin request? Do we "force english"?

app/plugin_api.go
```diff
@@ -914,3 +914,59 @@ func (api *PluginAPI) PluginHTTP(request *http.Request) *http.Response {
 	api.app.ServeInterPluginRequest(responseTransfer, request, api.id, destinationPluginId)
 	return responseTransfer.GenerateResponse()
 }
+
+func (api *PluginAPI) CreateSlashCommand(cmd *model.Command) (*model.Command, *model.AppError) {
```
lieut-data, line 918: Nit: We haven't traditionally used the term `SlashCommand` in the code, but just `Command`. Is now the time to diverge?

app/plugin_api.go
```diff
@@ -914,3 +914,59 @@ func (api *PluginAPI) PluginHTTP(request *http.Request) *http.Response {
...
+	}
+
+	if cmd.TeamId != oldCmd.TeamId {
+		return nil, model.NewAppError("UpdateSlashCommand", "api.command.team_mismatch.app_error", nil, "plugin_id="+api.id, http.StatusBadRequest)
+	}
+
+	return api.app.UpdateCommand(oldCmd, cmd)
+}
+
+func (api *PluginAPI) MoveSlashCommand(commandID string, newTeamID string) *model.AppError {
```
lieut-data, line 956: I know this was discussed elsewhere, but should we take the opportunity to simplify the API and just support moving directly in `UpdateSlashCommand`?

model/command.go
```diff
@@ -35,6 +35,7 @@ type Command struct {
 	DisplayName      string            `json:"display_name"`
 	Description      string            `json:"description"`
 	URL              string            `json:"url"`
+	PluginId         string            `json:"plugin_id"`
```
lieut-data, line 38: nit: Can we add godoc to at least this new field while we're in here?

store/sqlstore/upgrade.go
```diff
@@ -818,6 +818,7 @@ func upgradeDatabaseToVersion526(sqlStore SqlStore) {
 	// TODO: uncomment when the time arrive to upgrade the DB for 5.26
 	//if shouldPerformUpgrade(sqlStore, VERSION_5_25_0, VERSION_5_26_0) {
 	sqlStore.CreateColumnIfNotExists("Sessions", "ExpiredNotify", "boolean", "boolean", "0")
+	sqlStore.CreateColumnIfNotExistsNoDefault("Commands", "PluginId", "VARCHAR(190)", "VARCHAR(190)")
```
lieut-data, line 821: We seem to use `CreateColumnIfNotExistsNoDefault` relatively sparingly -- can you expand on the consequences for not defining a default in this case? Does the field default to `NULL` or `""`? Does the migration on an existing database cause any weirdness between MySQL or Postgres?

app/plugin_api.go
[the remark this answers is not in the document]
cpoile: As we chatted about offline: I'm great with returning error, but these are wrappers around `app/command.go` functions, which return appErrors, unlike the pluginAPI's register/unregister Command functions, which return error. So I'll just leave it as appError, and let the plugin-api handle the conversion.

app/plugin_api.go
```diff
@@ -914,3 +914,59 @@ func (api *PluginAPI) PluginHTTP(request *http.Request) *http.Response {
...
+func (api *PluginAPI) CreateSlashCommand(cmd *model.Command) (*model.Command, *model.AppError) {
+	cmd.CreatorId = ""
+	cmd.PluginId = api.id
+
+	cmd, appErr := api.app.CreateCommand(cmd)
+
+	return cmd, appErr
+}
+
+func (api *PluginAPI) ListSlashCommands(teamID string, customOnly bool) ([]*model.Command, *model.AppError) {
```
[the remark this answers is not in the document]
cpoile: For the customOnly flag, Christopher would usually call it with `customOnly` as `true`, but he wants to keep it as general as possible. You could argue that returning everything keeps it as general as possible, but then you'd have to add even more code in the plugin-api to re-narrow down the results. To me it makes more sense to narrow it with the customOnly flag, which the api already supports. Then we could simplify it in the plugin-api, as you suggest.

About returning custom commands from every team -- we could add that in the plugin-api, imo. Considering how many teams some orgs have, it would be a non-trivial time cost for that call, unless you wrote a new SQL query. Did you want to do that now? (I would vote for leaving that for later, if/when we need it, and we find that calling this per team was costing too much.)

app/plugin_api.go
```diff
@@ -914,3 +914,59 @@ func (api *PluginAPI) PluginHTTP(request *http.Request) *http.Response {
...
+	var commands []*model.Command
+	var appErr *model.AppError
+
+	if customOnly {
+		commands, appErr = api.app.ListTeamCommands(teamID)
```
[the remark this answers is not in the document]
cpoile: Yes, good point! There's a bit of a problem then: if `customOnly` is true we will return all custom commands even if `EnableCommands` is false, but if `customOnly` is false, we will not include custom commands if `EnableCommands` is false. So I'll have to remember that for the plugin-api PR and call this twice and merge the results.

app/plugin_api.go
cpoile: Looks like you're right: it's for built-in commands for translating the `AutoCompleteDesc` and `DisplayName`, eg: https://github.com/mattermost/mattermost-server/blob/a63684fcb5e3ba7b7522b35c29a4cb27779ba823/app/command_dnd.go#L26-L33
I don't know what language the plugin would take... the system default?

app/plugin_api.go
cpoile: Yep, changing it back.

app/plugin_api.go
cpoile: Talked to Joram, his message:

Two concerns I can think of:

1. Changing this will change the behaviour of the API (both REST and plugin) and could potentially break consumers relying on that behaviour (e.g. someone is using the API to update their commands but is not providing a team ID)
1. The current REST API for update command doesn't do any permission checks on the new team ID so we'd need to do that (not a concern for plugins)

store/sqlstore/upgrade.go
cpoile: Originally I had it as `CreateColumnIfNotExistsNoDefault`, and then changed it to `CreateColumnIfNotExists`, but then the circleCI tests started failing. I'll try again, and hope it doesn't.
If it does, I think a NULL would still turn into a Go zero value in the struct, wouldn't it?

cpoile: @lieut-data It looks like it's failing because of this. Is it okay if we go with no default? (Unless you can think of a fix for the schema test?)

cpoile: Ok, we definitely /cannot/ leave it with NoDefault. It will break because somewhere along the way the sql (driver?) is returning it cannot convert NULL to string. So I need to figure out how to make the default work, but I'm just not sure why the shema test is failing.

app/plugin_api.go
lieut-data: Thanks, for the follow-up, @cpoile. I think we've already long since diverged from the REST API proper, so I definitely wouldn't propose touching the REST API there. My chief focus is how we can make the plugin API "better" and drop the legacy behaviour that was needed to support the REST API.

Along those lines, should we require a `teamID` when calling the plugin's `UpdateCommand`, and transparently support moving it between teams when it changes?

app/plugin_api.go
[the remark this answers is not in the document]
lieut-data: @crspeller, I'm still hesitant to include `customOnly` as a filter parameter. As it stands right now, I feel like this API is trying to be too much like its REST equivalent, which is clearly geared towards the need for autocomplete. If we were standing up this API from scratch, would we support a `customOnly`? Why can't I query `pluginOnly` or `builtInOnly`? My inclination is still just to simplify this to return all, and support whatever filtering we need in the plugin-api wrapper instead.

I'm fine with keeping the `teamId`, since that's a fundamental aspect of how commands work and this is probably not the right layer to change that. We can definitely kick that out from our plans altogether right now.

app/plugin_api.go
cpoile: Sorry, I totally misunderstood what you meant. 🤦  That makes total sense.
Now I'm transparently moving between teams if `updatedCommand` has a new `teamId`, if it has no `teamId` then we'll use the `oldCmd`'s teamId. The comments are updated to reflect this too.

app/plugin_api.go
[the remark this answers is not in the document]
cpoile: Quick question -- is there any way to filter (from a list of commands) plugin commands vs built-in commands? Custom commands have URLs, so that's possible.

If there isn't, then I think we should add essentially a bitmask to the list commands:
```
type ListCommandsOptions struct {
  builtIn bool
  plugin bool
  custom bool
}
```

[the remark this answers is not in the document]
lieut-data: That's a good question -- do plugin commands have an `Id`, given that we don't actually persist these to the database?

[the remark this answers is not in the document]
cpoile: @lieut-data looks like there's no difference between built-in and plugin commands, they both look the same when they're in a list of commands. I think we're going to have to use something like an options. Any other ideas, or should I just do that?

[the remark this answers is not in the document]
lieut-data: I'm fine with keeping things simple -- I guess I'm actually looking to make things /simpler/ re: even supporting `customOnly`. Given this API comes as an ask from @crspeller, I'm fine deferring to his requirements.

store/sqlstore/upgrade.go
cpoile: Discussed offline, will be going with `CreateColumnIfNotExistsNoDefault` because it matches what Gorp gives us.

app/plugin_api.go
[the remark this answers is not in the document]
cpoile: With the simpler ListXCommands design, this isn't a problem anymore.

app/plugin_api.go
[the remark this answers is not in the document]
cpoile: The new ListXCommands api fixes this in a way that makes a bit more sense, I think.

cpoile:
@lieut-data gentle ping on this. Thanks!

app/plugin_api.go
```diff
@@ -929,3 +929,125 @@ func (api *PluginAPI) PluginHTTP(request *http.Request) *http.Response {
...
+				seen[cpy.Trigger] = true
+				commands = append(commands, &cpy)
+			}
+		}
+	}
+
+	return commands, nil
+}
+
+func (api *PluginAPI) GetCommand(commandID string) (*model.Command, error) {
```
lieut-data, line 1003: I know I've waffled on the terminology here, but I'm worried about the potential for confusion between `GetCommand` and `ListCustomCommands`. It's not clear from the naming that they are peers. One idea might be to keep it as `ListCommands` (and `ListPluginCommands`, etc.)

model/command.go
```diff
@@ -80,7 +83,13 @@ func (o *Command) IsValid() *AppError {
 		return NewAppError("Command.IsValid", "model.command.is_valid.update_at.app_error", nil, "", http.StatusBadRequest)
 	}
 
-	if !IsValidId(o.CreatorId) {
+	// If the CreatorId is blank, this should be a command created by a plugin.
+	if o.CreatorId == "" && !IsValidPluginId(o.PluginId) {
+		return NewAppError("Command.IsValid", "model.command.is_valid.plugin_id.app_error", nil, "", http.StatusBadRequest)
+	}
+
+	// If the PluginId is blank, this should be a command associated with a userId.
```
lieut-data, line 91: If I specify a non-epty `CreatorId` and a non-empty `PluginId`, I think neither of these checks run and the command appears to be valid. Should we assert that only one of these may be set?

app/plugin_api.go
```diff
@@ -929,3 +929,125 @@ func (api *PluginAPI) PluginHTTP(request *http.Request) *http.Response {
...
+	}
+
+	updatedCmd.Trigger = strings.ToLower(updatedCmd.Trigger)
+	updatedCmd.Id = oldCmd.Id
+	updatedCmd.Token = oldCmd.Token
+	updatedCmd.CreateAt = oldCmd.CreateAt
+	updatedCmd.UpdateAt = model.GetMillis()
+	updatedCmd.DeleteAt = oldCmd.DeleteAt
+	updatedCmd.CreatorId = oldCmd.CreatorId
+	updatedCmd.PluginId = oldCmd.PluginId
```
lieut-data, line 1021: Should we force `api.id` here? Should we allow both `CreatorId` and `PluginId` to be set?

app/plugin_api.go
```diff
@@ -929,3 +929,125 @@ func (api *PluginAPI) PluginHTTP(request *http.Request) *http.Response {
...
+
+	cmd, appErr := api.app.CreateCommand(cmd)
+
+	return cmd, normalizeAppErr(appErr)
+}
+
+func (api *PluginAPI) ListAllCommands(teamID string) ([]*model.Command, error) {
+	ret := make([]*model.Command, 0)
+
+	cmds, err := api.ListCustomCommands(teamID)
```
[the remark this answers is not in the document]
cpoile: Great point, done!

app/plugin_api.go
cpoile: Yep, that makes sense. Done!

app/plugin_api.go
cpoile: I was thinking about this, and had written a DM to you, along the lines of "are we sure we want to allow a plugin to basically take ownership over any command, and actually /force/ them to take ownership if they update an existing command?", but then I realized that we basically allow them to do that anyway. They can delete a command and create a new one over it anyway. 

So I'm letting them set CreatorId, and forcing PluginId to be api.id. 

By the way, I updated the `DeleteCommand` function--it would fail if `EnableCommands` was false, so I'm bypassing that check. Same for `GetCommand` and `CreateCommand`, and then I removed NormalizeAppError (didn't need it anymore).

There's a slight problem I noticed though--some of the community work to convert to error means `appErrors` are being cast into `error`, eg: https://github.com/mattermost/mattermost-server/blob/ab2ae46c2afae5e2dda7c1e5d28de564f9020dba/store/sqlstore/command_store.go#L179

Should we do something about this? Could cause havok down the line.

app/plugin_api.go
```diff
@@ -933,28 +934,45 @@ func (api *PluginAPI) PluginHTTP(request *http.Request) *http.Response {
 func (api *PluginAPI) CreateCommand(cmd *model.Command) (*model.Command, error) {
 	cmd.CreatorId = ""
 	cmd.PluginId = api.id
+	cmd.Trigger = strings.ToLower(cmd.Trigger)
 
-	cmd, appErr := api.app.CreateCommand(cmd)
+	teamCmds, err := api.app.Srv().Store.Command().GetByTeam(cmd.TeamId)
```
lieut-data, line 939: I'm worried about the code duplication with `App.CreateCommand` here. Should we factor out the common code instead?

cpoile: Yep, that's better now. Thanks!

justinegeffen:
@amyblais, the docs for this PR get updated automatically, so I'm marking as docs not needed.

PR #14922, opened 2020-06-26
Remove duplicated signal

rvillablanca:
#### Summary

Just a minor fix to remove a duplicated reference to `syscall.SIGINT` when listening for interruptions

#### Ticket Link

There is no ticket

### DIFF
--- a/cmd/mattermost/commands/server.go
+++ b/cmd/mattermost/commands/server.go
@@ -92,7 +92,7 @@ func runServer(configStore config.Store, disableConfigWatch bool, usedPlatform b
 
 	// wait for kill signal before attempting to gracefully shutdown
 	// the running service
-	signal.Notify(interruptChan, os.Interrupt, syscall.SIGINT, syscall.SIGTERM)
+	signal.Notify(interruptChan, syscall.SIGINT, syscall.SIGTERM)
 	<-interruptChan
 
 	return nil


cmd/mattermost/commands/server.go
```diff
@@ -92,7 +92,7 @@ func runServer(configStore config.Store, disableConfigWatch bool, usedPlatform b
 
 	// wait for kill signal before attempting to gracefully shutdown
 	// the running service
-	signal.Notify(interruptChan, os.Interrupt, syscall.SIGINT, syscall.SIGTERM)
+	signal.Notify(interruptChan, syscall.SIGINT, syscall.SIGTERM)
```
ethervoid, line 95: I think is not duplicated but in case it is, we should keep the `os.Interrupt` one because the os package is cross-platform.

Take into account that MM could run in windows too and in Windows, syscall is not implemented AFAIK so we should use `os.Interrupt` to catch that signal.

Probably @agnivade could share more light into this

rvillablanca: I was looking in the source code of go and here https://golang.org/src/os/exec_posix.go `os.Interrupt = syscall.SIGINT` and it applies to windows and linux so I think there should be no problems, though `os.Interrupt` is more portable as you said.

ethervoid: Probably we should also change `syscall.SIGTERM` to `os.Kill` to avoid problems with windows too

rvillablanca: But `os.Kill` can't be catched/ignored by none process and `syscall.SIGTERM` is supported on windows, so even leaving them as before is good.

rvillablanca: You can find the definitions of `syscall.SIGTERM` for windows here https://golang.org/src/os/types_windows.go

ethervoid: Yep:

```
Additionally, if Notify is called, and Windows sends CTRL_CLOSE_EVENT, CTRL_LOGOFF_EVENT or CTRL_SHUTDOWN_EVENT to the process, Notify will return syscall.SIGTERM. Unlike Control-C and Control-Break, Notify does not change process behavior when either CTRL_CLOSE_EVENT, CTRL_LOGOFF_EVENT or CTRL_SHUTDOWN_EVENT is received - the process will still get terminated unless it exits. But receiving syscall.SIGTERM will give the process an opportunity to clean up before termination. 
```

https://golang.org/pkg/os/signal/#hdr-Windows

Problem solved :D

agnivade: Yeah, you are right Rodrigo. I missed this. I should change it in mattermod as well.

ethervoid:
@saturninoabril sorry I’ve merged thinking that this was the mattermod repo. We need to check for Linux, macOS and Linux

agnivade:
@ethervoid - Do we want to revert this then?

ethervoid:
Don’t think it’s needed. It’s merged in master, is the command and should be tested shortly so I’d stay with it merged

PR #14935, opened 2020-06-29
MM-26563: Skip TestCreatePostCheckOnlineStatus if timeout exceeds

agnivade:
We just skip the test instead of failing because waiting for more than 5 seconds
to get a response does not make sense, and it will unncessarily slow down
the tests further in an already congested CI environment.

#### Ticket Link

https://mattermost.atlassian.net/browse/MM-26563


agnivade:
@lindy65 - Another flaky test improvement.

PR #14941, opened 2020-06-29
[MM-26541] Parse channel member counts using LIKE instead of using JSON

fmunshi:
#### Summary
- The previous PR https://github.com/mattermost/mattermost-server/pull/14673 updated the query to look at distinct timezones more accurately however I mistakenly assumed that our minimum supported MySQL version had support for JSON operators... this was not true our minimum supported version is 5.6 which does not have any of the JSON parsing functionality.

- I have instead added logic to parse the timezone json blob using LIKE and string locate / position 

#### Ticket Link
- https://mattermost.atlassian.net/browse/MM-26541

### DIFF
--- a/store/sqlstore/channel_store.go
+++ b/store/sqlstore/channel_store.go
@@ -18,6 +18,7 @@ import (
 	"github.com/mattermost/mattermost-server/v5/model"
 	"github.com/mattermost/mattermost-server/v5/services/cache2"
 	"github.com/mattermost/mattermost-server/v5/store"
+	"github.com/mattermost/mattermost-server/v5/utils"
 
 	sq "github.com/Masterminds/squirrel"
 	"github.com/pkg/errors"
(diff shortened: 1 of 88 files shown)


store/sqlstore/channel_store.go
```diff
@@ -1816,27 +1816,41 @@ func (s SqlChannelStore) GetMemberCountsByGroup(channelID string, includeTimezon
...
+						FROM ` + endOfFirstKey + `
+						FOR ` + lengthOfFirstValue + `
+					)
+					WHEN Timezone like '%"useAutomaticTimezone":"false"}' AND LENGTH(Timezone) > 74
+					THEN
+						SUBSTRING(
+						Timezone
+						FROM ` + endOfSecondKey + `
+						FOR ` + lengthOfSecondValue + `
+					)
```
sbishel, line 1849: These numbers should be made into constants and possibly an explanation, so someone later could understand the magic numbers. 
```
const LENGTH_OF_DEFAULT_TIMEZONE = 74 
// len("{"automaticTimezone":"","manualTimezone":"","useAutomaticTimezone":"true"}")
```

fmunshi:
> Is this approach resilient to a change in the JSON structure? If not, what's the history of the payload? Has it always been the same since the feature was added?

This approach assumes that the hash contains the following keys `automaticTimezone`, `manualTimezone`, `useAutomaticTimezone` in that order. If a key is added or removed this code will need to be modified to handle the changes...

Since I cant guarantee the length of the values of any of the keys I have to use the difference in position of the next key and the current key to find the values for specific keys. If any breaking change was made to this column a bunch of tests would fail so i think its safe to tie it to the structure of the column

fmunshi:
For future reference this approach works consistently since golangs json marshal methods handles maps by sorting the keys alphabetically https://golang.org/pkg/encoding/json/#Marshal.

Until someone modifies the code and adds / removes keys to the Timezone string map this will work.

PR #14943, opened 2020-06-30
MM-26064 - ENV var license override

wiggin77:
#### Summary

This PR adds support for a new environment variable `MM_LICENSE` which can contain the contents of a license file.  When set, this license takes priority over all other license sources.  

- Database and file licenses are not overwritten but simply ignored while `MM_LICENSE` is set.

- If MM_LICENSE is set to a non-empty string, but the license is not valid, the server will be started without a license.

#### Ticket Link

https://mattermost.atlassian.net/browse/MM-26064

### DIFF
--- a/app/license.go
+++ b/app/license.go
@@ -17,12 +17,12 @@ import (
 
 const (
 	requestTrialURL = "https://customers.mattermost.com/api/v1/trials"
-	LicenseENV      = "MM_LICENSE"
+	LicenseEnv      = "MM_LICENSE"
 )
 
 func (s *Server) LoadLicense() {
 	// ENV var overrides all other sources of license.
-	licenseStr := os.Getenv(LicenseENV)
+	licenseStr := os.Getenv(LicenseEnv)
 	if licenseStr != "" {
 		if s.ValidateAndSetLicenseBytes([]byte(licenseStr)) {
 			mlog.Info("License key from ENV is valid, unlocking enterprise features.")
--- a/app/license.go
+++ b/app/license.go
@@ -25,7 +25,7 @@ func (s *Server) LoadLicense() {
 	licenseStr := os.Getenv(LicenseENV)
 	if licenseStr != "" {
 		if s.ValidateAndSetLicenseBytes([]byte(licenseStr)) {
-			mlog.Info("License key from ENV valid unlocking enterprise features.")
+			mlog.Info("License key from ENV is valid, unlocking enterprise features.")
 		}
 		return
 	}
--- a/app/license.go
+++ b/app/license.go
@@ -7,6 +7,7 @@ import (
 	"bytes"
 	"errors"
 	"net/http"
+	"os"
 	"strings"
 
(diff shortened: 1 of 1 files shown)


app/license.go
```diff
@@ -7,16 +7,29 @@ import (
...
+	requestTrialURL = "https://customers.mattermost.com/api/v1/trials"
+	LicenseENV      = "MM_LICENSE"
+)
 
 func (s *Server) LoadLicense() {
+	// ENV var overrides all other sources of license.
+	licenseStr := os.Getenv(LicenseENV)
+	if licenseStr != "" {
+		if s.ValidateAndSetLicenseBytes([]byte(licenseStr)) {
+			mlog.Info("License key from ENV valid unlocking enterprise features.")
```
reflog, line 28: small nit: 
```suggestion
			mlog.Info("License key from ENV is valid, unlocking enterprise features.")
```

app/license.go
```diff
@@ -7,16 +7,29 @@ import (
...
 
 	"github.com/mattermost/mattermost-server/v5/mlog"
 	"github.com/mattermost/mattermost-server/v5/model"
 	"github.com/mattermost/mattermost-server/v5/utils"
 )
 
-const requestTrialURL = "https://customers.mattermost.com/api/v1/trials"
+const (
+	requestTrialURL = "https://customers.mattermost.com/api/v1/trials"
+	LicenseENV      = "MM_LICENSE"
```
jwilander, line 20: nit: should this be `LicenseEnv`? 0/5

jwilander:
cc @cpanato

app/license.go
cpanato: What happens with the current settings we have? for example, if the user sets the `ServiceSettings.LicenseFileLocation` and this new var?

side note: in the operator side, we will need to manage two ways to inject the license for versions `<= 5.25`

wiggin77: The ENV var takes priority over LicenseFileLocation and any DB saved license. In other words, those are ignored as long as MM_LICENSE is defined with a non-empty string.

wiggin77:
@lindy, is this something that needs QA?

amyblais:
@wiggin77 I noticed this wasn't cherry-picked. Is this too risky to include in v5.25 and would be better to push to v5.26?

wiggin77:
> @wiggin77 I noticed this wasn't cherry-picked. Is this too risky to include in v5.25 and would be better to push to v5.26?

This is a new feature, therefore I didn't think was a candidate for a stability release.

amyblais:
@wiersgallak I've pushed this to v5.26 since this is a new feature and we're past code complete and RC testing. Let me know if you'd like to re-discuss adding this to v5.25.

wiersgallak:
@amyblais thanks for the heads up.  5.26 is fine.

cwarnermm:
Documentation updated via: https://github.com/mattermost/docs/pull/4204

PR #14944, opened 2020-06-30
[MM-25647]: ability to permanently delete users

ashishbhate:
#### Summary

- Config option to allow permanent user deletion `ServiceSettings.EnableAPIUserDeletion`
- Expose permanent user deletion through API
- Local mode for delete user for use in mmctl

#### Ticket Link

- Server part of https://mattermost.atlassian.net/browse/MM-25647


ashishbhate:
@crspeller @mgdelacroix gentle bump for review

model/client4.go
```diff
@@ -1222,6 +1222,17 @@ func (c *Client4) DeleteUser(userId string) (bool, *Response) {
 	return CheckStatusOK(r), BuildResponse(r)
 }
 
+// PermanentDeleteUser deletes a user in the system based on the provided user id string.
+// If the EnableAPIUserDeletion config option is not set this falls back to user deactivation.
```
crspeller, line 1226: Since this is explicitly permanent deletion, should it fall back to user deactivation? As an API user I might expect to get an error. Having the users not be actually deleted might be unexpected.

ashishbhate: @crspeller The code here follows the precedent set by the `PermanentDeleteTeam`. I too was surprised by the implementation there so I added a comment here clarifying the behaviour.

Maybe a PM would like to weigh in on this @thefactremains @wiersgallak.

crspeller: Ya. Let's get a PM opinion on this behavior 👍

thefactremains: sorry guys, I might not understand this well enough yet (and I'm not clear on the reason we are doing this). Perhaps @wiersgallak should weigh in.

I do have a question though: Are we treating this the same as the `PermanentDeleteAllUsers`?

ashishbhate: >I do have a question though: Are we treating this the same as the PermanentDeleteAllUsers?

Functionally yes - `PermanentDeleteAllUsers` deletes all users. `PermanentDeleteUser`  deletes the specified user (with one caveat explained below). `PermanentDeleteAllUsers` is only available in local mode. 

>sorry guys, I might not understand this well enough yet (and I'm not clear on the reason we are doing this).


`PermanentDeleteUser` will delete all user data (i.e. actually remove the data from the DB) if the `EnableAPIUserDeletion` server option is set. Otherwise it just marks a user as inactive. This is the caveat I mentioned above.

I implemented it this way in this PR is to be consistent with existing code. The existing `PermanentDeleteTeam` has set a precedent - if the `EnableAPITeamDeletion` server option is set then the team is actually deleted from the DB, otherwise it does a "soft delete" (the team is just marked as deleted).

As Christopher commented above, this might be unexpected. Since this is explicitly permanent deletion, a dev or admin might expect an error instead of user deactivation.

In summary: Should we follow the existing convention in the code and behaviour (`PermanentDeleteTeam`), or should we update the behaviour to return an error instead of a soft delete.

thefactremains: That helps a lot. Thank you.

I definitely believe we should give them an error if the command does not actually do what it is intended/documented to do. 

That said, I also believe we should update `PermanentDeleteTeam` as well.

ashishbhate: I looked through the git history to figure why `PermanentDeleteTeam` was implemented that way - I found this jira ticket https://mattermost.atlassian.net/browse/MM-9916

ashishbhate: @thefactremains  and I discussed this offline. I will update this PR to return an error instead of doing a soft delete if the option is not set.
I've created https://mattermost.atlassian.net/browse/MM-26891 to update PermanentDeleteTeam as well.

model/client4.go
```diff
@@ -1226,6 +1226,17 @@ func (c *Client4) DeleteUser(userId string) (bool, *Response) {
...
```
mgdelacroix, line 1230: nit: the comment needs to be updated with the new behaviour

ashishbhate: Good catch! fixed.

ashishbhate:
@crspeller This is ready for review.

amyblais:
Pushing from v5.26 as it missed feature complete but let me know if this is very important for v5.26,

amyblais:
@ashishbhate Does the new config setting need to be added to telemetry?

ashishbhate:
@amyblais I've created a PR to add them to telemetry https://github.com/mattermost/mattermost-server/pull/15506

PR #14945, opened 2020-06-30
MM-24972: make GetHubForUserId lock-less and zero-alloc

agnivade:
The hubs aren't meant to be modified after the server starts up.
But the key problem was the SetHubs method which would zero out
the underlying hubs slice.

This wasn't ideally necessary because an hub would only be stopped
once during server shutdown. However, this was required by a test
which would shutdown the hub twice. And since the hub shutdown isn't
idempotent, zeroing the slice would just skip over shutting down the hub again.

To improve the overall situation, we apply several optimizations.

- We use the new hash/maphash package which exposes Go runtime's
internal hash algorithms to be used as a package. This is much faster
than hash/fnv.
- We move around the initialization of the hub to happen before
the metrics server starts. This allows us to initialize the hub
before any of the hub elements are being accessed.
- To make the test run successfully, we do not call th.TearDown.
This is fine for a test, because anyways the test process would eventually stop
and relinquish the resources to the OS.

This allows us to completely remove any mutexes and thereby
we can remove all the methods and any edge-case checks related to
index being out of bounds. As a result, the fast path becomes
very straightforward and zero-alloc.

```
GetHubForUserId-8     116ns ± 1%      38ns ± 7%   -67.22%  (p=0.000 n=10+10)

GetHubForUserId-8     36.0B ± 0%      0.0B       -100.00%  (p=0.000 n=10+10)

GetHubForUserId-8      2.00 ± 0%      0.00       -100.00%  (p=0.000 n=10+10)
```
Manually tested with some load testing and running Hub tests in -race mode.

#### Ticket link

https://mattermost.atlassian.net/browse/MM-24792


app/web_hub.go
```diff
@@ -129,19 +125,20 @@ func (a *App) HubStop() {
 
 // GetHubForUserId returns the hub for a given user id.
 func (s *Server) GetHubForUserId(userId string) *Hub {
-	if len(s.GetHubs()) == 0 {
+	// TODO: This check is probably redundant and can be removed.
```
crspeller, line 128: Then can we remove it?

app/web_hub.go
```diff
@@ -129,19 +125,20 @@ func (a *App) HubStop() {
...
+	if len(s.hubs) == 0 {
 		return nil
 	}
 
-	hash := fnv.New32a()
+	// TODO: check if caching the userId -> hub mapping
```
crspeller, line 133: Ticket rather than a TODO?

app/server.go
```diff
@@ -210,6 +213,11 @@ func NewServer(options ...Option) (*Server, error) {
 	// Use this app logger as the global logger (eventually remove all instances of global logging)
 	mlog.InitGlobalLogger(s.Log)
 
+	// It is important to initialize the hub only after the global logger is set
+	// to avoid race conditions while logging from inside the hub.
+	fakeApp := New(ServerConnector(s))
```
crspeller, line 218: I know @jespino is working on eliminating fakeApp. It would be nice if you where to move the hub code to avoid its use here.

agnivade: Yes, Jesus and I are aware of this. It's not possible to move the hub right away from the app layer because it's a bit tightly coupled. But it's on our plan to work on it in small chunks.

app/web_hub.go
agnivade: Yeah I think so. I just need to ensure that this method is _never_ called before the hub is initialized. I wanted to perhaps delay that investigation to a later point, but no harm in spending some more time on it.

app/web_hub_test.go
```diff
@@ -185,6 +183,34 @@ func TestHubConnIndex(t *testing.T) {
...
+	wc3 := registerDummyWebConn(t, th.App, s.Listener.Addr(), th.BasicUser.Id)
+	defer wc1.Close()
+	defer wc2.Close()
+	defer wc3.Close()
+
+	session1 := wc1.session.Load().(*model.Session)
+
+	assert.True(t, th.App.SessionIsRegistered(*session1))
+	// assert.True(t, th.App.SessionIsRegistered(*wc2.session.Load().(*model.Session)))
+	// assert.True(t, th.App.SessionIsRegistered(*wc3.session.Load().(*model.Session)))
```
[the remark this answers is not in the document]
agnivade: Oh woops, my bad. Yes they are needed. I had commented them out during testing.

app/web_hub.go
agnivade: https://mattermost.atlassian.net/browse/MM-26629

app/web_hub.go
```diff
@@ -82,18 +82,16 @@ func (a *App) HubStart() {
...
-	for i := 0; i < len(a.Srv().GetHubs()); i++ {
-		newHub := a.NewWebHub()
-		newHub.connectionIndex = i
-		err := a.Srv().SetHub(i, newHub)
-		if err != nil {
-			mlog.Warn("Error starting hub", mlog.Err(err), mlog.Int("index", i))
-			continue
-		}
-		newHub.Start()
+	for i := 0; i < numberOfHubs; i++ {
```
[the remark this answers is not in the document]
agnivade: Aha, this creates a scopelint issue :P

agnivade:
@lindy65 - This needs testing around all websocket related functionality on an HA setup.

PR #14951, opened 2020-06-30
Mm 16292 support command overrides

crspeller:
#### Summary
Support command overrides. Built in commands can now be overridden by plugins and custom commands.
Also fixes a bug in slash command autocomplete where duplicate checking was done incorrectly. 

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-26201

### DIFF
--- a/app/command.go
+++ b/app/command.go
@@ -77,21 +77,22 @@ func (a *App) CreateCommandPost(post *model.Post, teamId string, response *model
 func (a *App) ListAutocompleteCommands(teamId string, T goi18n.TranslateFunc) ([]*model.Command, *model.AppError) {
 	commands := make([]*model.Command, 0, 32)
 	seen := make(map[string]bool)
-	for _, value := range commandProviders {
-		if cmd := value.GetCommand(a, T); cmd != nil {
-			cpy := *cmd
-			if cpy.AutoComplete && !seen[cpy.Id] {
-				cpy.Sanitize()
-				seen[cpy.Trigger] = true
-				commands = append(commands, &cpy)
-			}
-		}
-	}
 
 	for _, cmd := range a.PluginCommandsForTeam(teamId) {
 		if cmd.AutoComplete && !seen[cmd.Trigger] {
 			seen[cmd.Trigger] = true
 			commands = append(commands, cmd)
+		}
+	}
+
+	for _, value := range commandProviders {
+		if cmd := value.GetCommand(a, T); cmd != nil {
+			cpy := *cmd
+			if cpy.AutoComplete && !seen[cpy.Trigger] {
+				cpy.Sanitize()
+				seen[cpy.Trigger] = true
+				commands = append(commands, &cpy)
+			}
 		}
 	}
 
@@ -102,7 +103,7 @@ func (a *App) ListAutocompleteCommands(teamId string, T goi18n.TranslateFunc) ([
 		}
 
 		for _, cmd := range teamCmds {
-			if cmd.AutoComplete && !seen[cmd.Id] {
+			if cmd.AutoComplete && !seen[cmd.Trigger] {
 				cmd.Sanitize()
 				seen[cmd.Trigger] = true
 				commands = append(commands, cmd)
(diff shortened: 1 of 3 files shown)


app/command.go
```diff
@@ -77,32 +77,33 @@ func (a *App) CreateCommandPost(post *model.Post, teamId string, response *model
...
+			seen[cmd.Trigger] = true
+			commands = append(commands, cmd)
+		}
+	}
+
 	for _, value := range commandProviders {
 		if cmd := value.GetCommand(a, T); cmd != nil {
 			cpy := *cmd
-			if cpy.AutoComplete && !seen[cpy.Id] {
+			if cpy.AutoComplete && !seen[cpy.Trigger] {
```
marianunez, line 91: We should log here that a built-in command has been overriden.

app/command.go
```diff
@@ -77,32 +77,33 @@ func (a *App) CreateCommandPost(post *model.Post, teamId string, response *model
...
-
 	if *a.Config().ServiceSettings.EnableCommands {
 		teamCmds, err := a.Srv().Store.Command().GetByTeam(teamId)
 		if err != nil {
 			return nil, err
 		}
 
 		for _, cmd := range teamCmds {
-			if cmd.AutoComplete && !seen[cmd.Id] {
+			if cmd.AutoComplete && !seen[cmd.Trigger] {
```
marianunez, line 106: From your PR description, I understood that you also want to allow custom commands to be able to override as well. 
Should we iterate on these before or remove these conditions then ?

app/plugin_commands_test.go
```diff
@@ -217,4 +217,89 @@ func TestPluginCommand(t *testing.T) {
...
+			}
+
+			type MyPlugin struct {
+				plugin.MattermostPlugin
+
+				configuration configuration
+			}
+
+			func (p *MyPlugin) OnConfigurationChange() error {
+				p.API.LogError("hello")
```
marianunez, line 247: Debugging log?

app/plugin_commands_test.go
```diff
@@ -217,4 +217,89 @@ func TestPluginCommand(t *testing.T) {
...
+				if err := p.API.LoadPluginConfiguration(&p.configuration); err != nil {
+					return err
+				}
+
+				return nil
+			}
+
+			func (p *MyPlugin) OnActivate() error {
+				p.API.LogError("team", "team", p.configuration.TeamId)
```
marianunez, line 256: Debugging log?

app/plugin_commands_test.go
```diff
@@ -217,4 +217,89 @@ func TestPluginCommand(t *testing.T) {
...
+		require.Equal(t, "text", resp.Text)
+
+		err2 := th.App.DisablePlugin(pluginIds[0])
+		require.Nil(t, err2)
+
+		commands, err3 := th.App.ListAutocompleteCommands(args.TeamId, utils.T)
+		require.Nil(t, err3)
+
+		for _, commands := range commands {
+			require.NotEqual(t, "plugin", commands.Trigger)
```
marianunez, line 300: I don't understand what this is asserting here given that the command registered was `/code`

app/command.go
crspeller: Why? It would spam the logs for every autocomplete.

app/command.go
crspeller: Yes. Fixed.

app/plugin_commands_test.go
crspeller: Ya. Just copy paste. Fixed.

app/plugin_commands_test.go
crspeller: Ya. Just copy paste. Fixed.

app/plugin_commands_test.go
crspeller: Ya. Just copy paste. Fixed.

app/command.go
marianunez: Oh, I see. I was under the impression this was just called at the time of registering the command.

PR #14954, opened 2020-07-01
Adding initial retry layer version

jespino:
#### Summary
Adding initial retry layer version. This can be simplified in the future
because the the store is not going handle any `AppError`.


store/layer_generators/retry_layer.go.tmpl
```diff
@@ -0,0 +1,98 @@
...
+        }
+        var pqErr *pq.Error
+        var mysqlErr *mysql.MySQLError
+        switch {
+            case errors.As(errors.Cause({{$element.Results | errorVar}}), &pqErr):
+                if pqErr.Code != "40001" && pqErr.Code != "40P01" {
+                    break
+                }
+            case errors.As(errors.Cause({{$element.Results | errorVar}}), &mysqlErr):
+                if mysqlErr.Number != mySQLDeadlockCode {
```
jespino, line 67: Mysql only have to deal with Deadlock code because the transactions doesn't fail in any other repeatable way (I think).

store/layer_generators/retry_layer.go.tmpl
```diff
@@ -0,0 +1,98 @@
...
+        {{else}}
+        if tries > 3 {
+            break
+            {{$element.Results | errorVar}} = errors.Wrap({{$element.Results | errorVar}}, "giving up after 3 consecutive repeatable transaction failures")
+        }
+        var pqErr *pq.Error
+        var mysqlErr *mysql.MySQLError
+        switch {
+            case errors.As(errors.Cause({{$element.Results | errorVar}}), &pqErr):
+                if pqErr.Code != "40001" && pqErr.Code != "40P01" {
```
jespino, line 63: Postgres repeat the transaction on 40001 (transaction not serializable) and "40P01" deadlock.

store/layer_generators/retry_layer.go.tmpl
```diff
@@ -0,0 +1,98 @@
...
+func (s *{{$.Name}}{{$substoreName}}Store) {{$index}}({{$element.Params | joinParamsWithType}}) {{$element.Results | joinResultsForSignature}} {
+    {{if $element.Results | len | eq 0}}
+    s.{{$substoreName}}Store.{{$index}}({{$element.Params | joinParams}})
+    return
+    {{else}}
+    {{$element.Results | genResultsVars}} := s.{{$substoreName}}Store.{{$index}}({{$element.Params | joinParams}})
+    {{if $element.Results | errorPresent}}
+    tries := 0
+    for {
+        tries++
```
jespino, line 48: The counter is always used to simplify the code generation.

jespino:
@agnivade I can give it a try.

jespino:
@agnivade extracted part of this into a function, I'm not sure if moving out the retry logic is going to increase or decrease the readability of the code because it can return `error`, `(value, error)`, `(value, value, error)` depending on the store call.

store/layer_generators/retry_layer.go.tmpl
```diff
@@ -0,0 +1,101 @@
...
+                for {
+                    {{$element.Results | genResultsVars}} := s.{{$substoreName}}Store.{{$index}}({{$element.Params | joinParams}})
+                    if {{$element.Results | errorVar}} == nil {
+                        return {{$element.Results | genResultsVars}}
+                    }
+                    if !isRepeatableError({{$element.Results | errorVar}}) {
+                        return {{$element.Results | genResultsVars}}
+                    }
+                    tries++
+                    if tries > 3 {
```
streamer45, line 73: If this is to mitigate deadlocks should we consider adding a small time (sleep?) before attempting again? This would help whatever is holding the lock finish doing its work.

jespino:
@streamer45 This is going to generate errors of serialization that needs you to retry the whole transaction (a rollback is going to happend, and you need to do the whole thing again). The whole thing can mean retry to execute the transactions, or some other logic that are not even visible to the SQL database. For example (I select data 1, select data 2, sum them from the outside, and store the result), if you blindly repeat that transaction, you going to execute 2 selects, and then try to write the "previous-old-calculated-result" into the database. Basically, the SQL server and the Gorp layer doesn't have enough information to decide how to repeat a transaction there.

store/retry_layer.go
```diff
@@ -0,0 +1,4315 @@
...
+}
+
+func (s *RetryLayerAuditStore) Get(user_id string, offset int, limit int) (model.Audits, error) {
+
+	tries := 0
+	for {
+		resultVar0, resultVar1 := s.AuditStore.Get(user_id, offset, limit)
+		if resultVar1 == nil {
+			return resultVar0, resultVar1
+		}
```
streamer45, line 355: I understand this being generated code but would I wonder if it's worth making the naming a bit more human-friendly.
Can we assume the second returned value will always be an `error` and if there's only one it's also an `error` or does this logic break in some cases?

store/layer_generators/retry_layer.go.tmpl
jespino: I'm ok introducing a tiny sleep, but normally, when you kill a deadlock transaction, you kill one of the offending transactions, so the other one is going to get the free lock and is going to finish it, and the retry one is going to succeed. (That is not necesarily alwasy true, so adding a random small amount of time there can help). What do you think @agnivade ?

store/retry_layer.go
jespino: Yes, that logic is not true, we can detect what is the error variable in the generator and name it accordingly, I can try to make the error variable always `err` instead of `resultVarN`. But we can't assume the position give any meaning, because is not true in all cases.

store/layer_generators/retry_layer.go.tmpl
agnivade: Not sure about the sleep part. My original retry function also doesn't have the sleep. I would expect this to work too. There is an extremely low probability that a valid transaction fails 3 times consecutively, but will pass if there's a sleep in between. This sounds like an edge-case of an already edge-case. Might not be worth adding it.

agnivade:
Some further thoughts:
- Let's also split this into a sub-package following the new trend?
- I also discovered that our app does not work with postgres in read-committed isolation level, _specifically_ because we don't have this retry layer. I guess "40001" is the code postgres returns when such an error occurs?

jespino:
@agnivade 
I removed the `store.WithDeadlockRetry` function and all the usages related to that. I removed entirely the `testReactionDeadlock` because I prefer to add specific tests to the retry layer mocking entirely the store, and testing the retry behavior. We can add the `testReactionDeadlock` again at a higher level (for example in the app layer).

jespino:
@agnivade I moved it into his own package, and yes, we handle all the `40001` (serialization error) and `40P01` (deadlock) errors. So this layer should fix the problem with Postgres higher isolation levels (in conjunction with the AppError removal from the store).

store/sqlstore/channel_store.go
```diff
@@ -735,28 +735,25 @@ func (s SqlChannelStore) Save(channel *model.Channel, maxChannelsPerTeam int64)
...
+	}
+	defer finalizeTransaction(transaction)
 
-		newChannel, err = s.saveChannelT(transaction, channel, maxChannelsPerTeam)
-		if err != nil {
-			return err
-		}
+	newChannel, err = s.saveChannelT(transaction, channel, maxChannelsPerTeam)
+	if err != nil {
+		return nil, err
```
agnivade, line 746: In the original code, it was `return newChannel, err`.

store/retrylayer/retrylayer_test.go
```diff
@@ -0,0 +1,114 @@
...
+		mock := genStore()
+		mockBotStore := mock.Bot().(*mocks.BotStore)
+		mysqlErr := mysql.MySQLError{Number: uint16(1213), Message: "Deadlock"}
+		mockBotStore.On("Get", "test", false).Return(nil, errors.Wrap(&mysqlErr, "test-error")).Times(3)
+		mock.On("Bot").Return(&mockBotStore)
+		layer := New(mock)
+		layer.Bot().Get("test", false)
+		mockBotStore.AssertExpectations(t)
+	})
+	t.Run("on mysql not repeatable error should retry", func(t *testing.T) {
```
agnivade, line 78: Should it not be "should not retry"?

store/retrylayer/retrylayer_test.go
```diff
@@ -0,0 +1,114 @@
...
+				mockBotStore.On("Get", "test", false).Return(nil, errors.Wrap(&pqErr, "test-error")).Times(3)
+				mock.On("Bot").Return(&mockBotStore)
+				layer := New(mock)
+				layer.Bot().Get("test", false)
+				mockBotStore.AssertExpectations(t)
+			})
+		}
+	})
+
+	t.Run("on postgres not repeatable error should retry", func(t *testing.T) {
```
agnivade, line 104: Same as above.

jespino:
@agnivade I'm not even able to reproduce the deadlock, maybe is because now most of it is in the same transaction?

agnivade:
The deadlock is pretty easily reproducible on mysql. If you run that test, a race happens atleast 1 out of 5 times according to my tests.

>  is because now most of it is in the same transaction?

I don't think any code has changed after I added that test. Both are two separate queries in 2 separate transactions.

jespino:
@agnivade I added the retry layer to the reaction store deadlock test, then I added some temporary traces to verify that sometimes it retries, and that was the case. Now the test passes even if it has a deadlock and has to retry.

PR #14955, opened 2020-07-02
MM-18006: Fix flaky test CreateOrRestoreGroupMember

agnivade:
Updating a Group or a GroupMember only changed the UpdateAt or CreateAt times
respectively. And it threw an error if number of rows changed was not 1.

However, it can happen that 2 calls happen so fast that 1 milisecond does not pass,
or even 2 concurrent calls at the same time might happen so that model.GetMillis
return the same timestamp. In those cases, the number of rows updated can be 0.

The error should just check if the number is greater than 1, instead of not equal to 1.
This makes it more robust and correct.

EDIT: This behavior only happens in MySQL (as expected). Postgres will always return a non-zero number
for an Update.

#### Ticket link

https://mattermost.atlassian.net/browse/MM-18006


agnivade:
@ogi-m - Some smoke testing related to user groups should be sufficient. Thanks.

agnivade:
Gentle ping on this @ogi-m

PR #14961, opened 2020-07-02
MM-26015: ensure admin has permission after removal from team

ashishbhate:
#### Summary

- Ensure admin retains permissions to team after being removed from a team

#### Ticket Link

- https://mattermost.atlassian.net/browse/MM-26015

### DIFF
--- a/app/authorization.go
+++ b/app/authorization.go
@@ -138,23 +138,18 @@ func (a *App) HasPermissionToTeam(askingUserId string, teamId string, permission
 	if teamId == "" || askingUserId == "" {
 		return false
 	}
+	teamMember, _ := a.GetTeamMember(teamId, askingUserId)
+	if teamMember != nil {
+		// If the team member has been deleted, they don't have permission.
+		// We still check if they have permissions from being an admin.
+		if teamMember.DeleteAt != 0 {
+			// Deleted TeamMember's roles are zeroed so we check permissions of user id
+			return a.HasPermissionTo(askingUserId, permission)
+		}
 
-	teamMember, err := a.GetTeamMember(teamId, askingUserId)
-	if err != nil {
-		return false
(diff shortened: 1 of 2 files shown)


app/authorization.go
```diff
@@ -138,23 +138,18 @@ func (a *App) HasPermissionToTeam(askingUserId string, teamId string, permission
 	if teamId == "" || askingUserId == "" {
 		return false
 	}
+	teamMember, _ := a.GetTeamMember(teamId, askingUserId)
+	if teamMember != nil {
```
lieut-data, line 142: For clarity, could we instead test `teamMember != nil && teamMember.DeleteAt == 0` for this block, and then just fall back to the existing `a.HasPermissionTo(askingUserId, permission)` at the bottom of the method instead of duplicating that check a second time?

ashishbhate:
@lieut-data thanks for the review, updated!

lindy65:
Hi @ashishbhate - do you know whether this is for v5.25?  Or 5.26?

FYI, Juho will be helping to test the PR...

ashishbhate:
>Hi @ashishbhate - do you know whether this is for v5.25? Or 5.26?

This is a bug fix, so it would be great if we can get this merged for v5.25. The code freeze for v.5.25 is in a couple days. 

I wasn't sure if this could be fixed in time for v5.25 so I hadn't added that milestone yet.

amyblais:
Is this a recent regression? If not this should be for v5.26 (unless this is a high severity issue which it doesn't look like).

ashishbhate:
>Is this a recent regression? If not this should be for v5.26 (unless this is a high severity issue which it doesn't look like).

Its not a recent regression, or a high severity issue. I'll move to v5.26.

Apologies for the noise, I misunderstood the release process and deadlines.

PR #14969, opened 2020-07-05
ComplianceStore migration

rvillablanca:
#### Summary

Migration of `ComplianceStore` to return plain errors.


rvillablanca:
Unable to reproduce these failed tests, random failed tests again :( ?

agnivade:
The test is fixed in master. Should be good once you get the latest.

rvillablanca:
Strange because it is already based on the latest master :thinking:, even the others PR are updated as well

jespino:
@agnivade: I think this needs changes in enterprise too.

agnivade:
@jespino - yep. Would you be open to fixing this one? :D I can take the next one.

rvillablanca:
> @jespino - yep. Would you be open to fixing this one? :D I can take the next one.

Give me access and I will fix both and more for you :sweat_smile:

agnivade:
> Give me access and I will fix both and more for you 

We are hiring :wink:

agnivade:
So there's already a PR for this https://github.com/mattermost/mattermost-server/pull/14808. I am not sure if we should go ahead with this without giving the other contributor a chance to fix the suggestions. I will ping on the other PR and wait for some time. If we don't get a reply, then we can go ahead with this.

rvillablanca:
> > Give me access and I will fix both and more for you
> 
> We are hiring

Not in my country :chile: unfortunately, also I applied in the past without success :sweat:

rvillablanca:
> So there's already a PR for this #14808. I am not sure if we should go ahead with this without giving the other contributor a chance to fix the suggestions. I will ping on the other PR and wait for some time. If we don't get a reply, then we can go ahead with this.

Sorry about that I didn't see it, giving priority to him is ok :+1:

agnivade:
@rvillablanca - can you sync with master and fix the conflicts? This is now ready to go.

rvillablanca:
@agnivade PR updated

rvillablanca:
Conflicts resolved

agnivade:
This needs EE changes again.

PR #14970, opened 2020-07-05
ClusterDiscoveryStore migration

rvillablanca:
#### Summary

Migration of `ClusterDiscoveryStore` to return plain errors.

Fixes https://github.com/mattermost/mattermost-server/issues/14596


agnivade:
@saturninoabril - this needs some cluster related testing.

Have a 2 app cluster. Bring one instance down, verify that the other one is correctly promoted to leader. Then bring this one up, verify it correctly joins the cluster. Do the same for the other instance and verify the same thing again.

PR #14971, opened 2020-07-06
Pluginstore migration

rvillablanca:
#### Summary

Migration of `PluginStore` to return plain errors.

Fixes https://github.com/mattermost/mattermost-server/issues/14986


store/sqlstore/plugin_store.go
```diff
@@ -56,23 +57,23 @@ func (ps SqlPluginStore) SaveOrUpdate(kv *model.PluginKeyValue) (*model.PluginKe
 		// Unfortunately PostgreSQL pre-9.5 does not have an atomic upsert, so we use
 		// separate update and insert queries to accomplish our upsert
 		if rowsAffected, err := ps.GetMaster().Update(kv); err != nil {
-			return nil, model.NewAppError("SqlPluginStore.SaveOrUpdate", "store.sql_plugin_store.save.app_error", nil, err.Error(), http.StatusInternalServerError)
+			return nil, errors.Wrap(err, "failed to update PluginKeyValue")
 		} else if rowsAffected == 0 {
 			// No rows were affected by the update, so let's try an insert
 			if err := ps.GetMaster().Insert(kv); err != nil {
-				return nil, model.NewAppError("SqlPluginStore.SaveOrUpdate", "store.sql_plugin_store.save.app_error", nil, err.Error(), http.StatusBadRequest)
+				return nil, store.NewErrInvalidInput("PluginKeyValue", "<PluginId, Key>", fmt.Sprintf("<%s, %s>", kv.PluginId, kv.Key))
```
jespino, line 64: I'm not sure if this is the type of error that is happening here. In theory the kv is valid, and there wasn't a valid update (so the key wasn't store) looks like this error only happens when the upsert (In this case a external implementation of upsert) fails. So for me, the error should ben the same that we see in the `mysql` case.

rvillablanca: yep I understand and it makes sense but I think this PR is not intended to fix this, but if you both agree with that suggestion, I will make the correction in another PR later :)

agnivade: Agree with Jesus that this needs to be an internal error. 0/5 on whether we want to do it now or later. Up to you.

app/plugin_key_value_store.go
```diff
@@ -81,17 +98,17 @@ func (a *App) CompareAndDeletePluginKey(pluginId string, key string, oldValue []
 func (a *App) GetPluginKey(pluginId string, key string) ([]byte, *model.AppError) {
 	if kv, err := a.Srv().Store.Plugin().Get(pluginId, key); err == nil {
 		return kv.Value, nil
-	} else if err.StatusCode != http.StatusNotFound {
+	} else if _, ok := err.(*store.ErrNotFound); !ok {
```
agnivade, line 101: Let's use `errors.As` here

app/plugin_key_value_store.go
```diff
@@ -81,17 +98,17 @@ func (a *App) CompareAndDeletePluginKey(pluginId string, key string, oldValue []
...
 		mlog.Error("Failed to query plugin key value", mlog.String("plugin_id", pluginId), mlog.String("key", key), mlog.Err(err))
-		return nil, err
+		return nil, model.NewAppError("GetPluginKey", "app.plugin_store.get.app_error", nil, err.Error(), http.StatusInternalServerError)
```
agnivade, line 103: This should return a 404, not 500.

app/plugin_key_value_store.go
```diff
@@ -81,17 +98,17 @@ func (a *App) CompareAndDeletePluginKey(pluginId string, key string, oldValue []
...
 	}
 
 	// Lookup using the hashed version of the key for keys written prior to v5.6.
 	if kv, err := a.Srv().Store.Plugin().Get(pluginId, getKeyHash(key)); err == nil {
 		return kv.Value, nil
-	} else if err.StatusCode != http.StatusNotFound {
+	} else if _, ok := err.(*store.ErrNotFound); !ok {
```
agnivade, line 109: same as above.

app/plugin_key_value_store.go
```diff
@@ -81,17 +98,17 @@ func (a *App) CompareAndDeletePluginKey(pluginId string, key string, oldValue []
...
 		mlog.Error("Failed to query plugin key value using hashed key", mlog.String("plugin_id", pluginId), mlog.String("key", key), mlog.Err(err))
-		return nil, err
+		return nil, model.NewAppError("GetPluginKey", "app.plugin_store.get.app_error", nil, err.Error(), http.StatusInternalServerError)
```
agnivade, line 111: same

store/sqlstore/plugin_store.go
rvillablanca: we are ok with this now :)

app/plugin_key_value_store.go
rvillablanca: I knew you could ask for that :) but what I don't like about that approach here is it forces us to create a pointer just to assert the error is of that type, but we won't use that pointer for anything else. Different is when we use this approach in the `switch` sentence because there we do want to use the pointed error. WDYT?

app/plugin_key_value_store.go
rvillablanca: I don't agree, in fact here we are checking for an error distinct of`ErrNotFound`

app/plugin_key_value_store.go
agnivade: The API is designed in a way that forces us to create that pointer. But not using `errors.As` makes the code fragile and prone to breakages. If you don't want to create a pointer as a separate variable, you can do

```
if nfErr := new(store.ErrNotFound); errors.As(err, &nfErr) {
}
```

app/plugin_key_value_store.go
agnivade: Sorry, could you clarify? Yes we are checking for ErrNotFound, and that's why I would like to return 404. The original code also does the same.

app/plugin_key_value_store.go
rvillablanca: Ok I will make that change, but I think you meant:

```go
if nfErr := new(store.ErrNotFound); !errors.As(err, &nfErr) {
}
```

Because we are checking that the error is **not** `store.ErrNotFound`. BTW that is why I'm returning 500 insted of 404 :)

Let me check the old code again because I'm hesitating now

agnivade: Ohhh that was the reason. My apologies, I couldn't understand what you meant by "distinct of ErrNotFound". Yes you are right of course. I failed to see it. No need to change to 404. Only the `errors.As` change is needed.

rvillablanca: :)

rvillablanca:
PR is ready to be reviewed

agnivade:
@ogi-m - This needs testing around all plugin related functionality. Thanks.

rvillablanca:
Conflicts resolved

jasonblais:
@ogi-m friendly reminder to help with QA review

agnivade:
@lindalumitchell - In light of some recent PRs that were approved for post-merge, I would like to request this one and https://github.com/mattermost/mattermost-server/pull/14978, https://github.com/mattermost/mattermost-server/pull/15028, https://github.com/mattermost/mattermost-server/pull/15027 as well for that.

These PRs just change internal errors and there is no visible change to the user. And given the latest decision we took not to translate server errors, there is even a lesser need to test this. They have been sitting for months and cause unnecessary churn for contributors who have to keep fixing merge conflicts that keep appearing from time to time.

Approving them would greatly improve the contributor experience and will take us one more step further our app error migration. Thanks.

PR #14972, opened 2020-07-06
MM-26346: check for nil Message in DisableMentionHighlights

agnivade:
#### Summary

Sometimes a PostPatch may not contain the message field. For example,
while pinning a post, only the IsPinned field is set, and everything
else is nil. In that case, DisableMentionHighlights would throw a panic.
    
We handle that case by checking for nil first.

#### Ticket Link

https://mattermost.atlassian.net/browse/MM-26346

### DIFF
--- a/model/post.go
+++ b/model/post.go
@@ -521,6 +521,9 @@ func (o *Post) DisableMentionHighlights() string {
 
 // DisableMentionHighlights disables mention highlighting for a post patch if required.
 func (o *PostPatch) DisableMentionHighlights() {
+	if o.Message == nil {
+		return
+	}
 	if _, hasMentions := findAtChannelMention(*o.Message); hasMentions {
 		if o.Props == nil {
 			o.Props = &StringInterface{}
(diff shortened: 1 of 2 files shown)


agnivade:
@ogi-m - This is already covered with unit tests. I don't see a need to test anything.

sadohert:
@amyblais  - FYI a customer has seen the issue addressed by this PR in their testing.  They had the `channel_mention` permission disabled.  Although they currently intend to have the permission ENABLED, and so I don't believe this is technically impacting them, if they were to change their configuration to disable this would impact them.

Would we consider this critical enough to be back-ported to the ESR?

amyblais:
There was a discussion previously on the ticket that this is not urgent but I don't have further details on it. @agnivade Can you help me understand the level of urgency on this?

agnivade:
The issue is that if a member does not have the `channel_mention` permission on a channel, and goes to pin a post, it won't happen.

amyblais:
My first thought is that we could consider backporting this to v5.25 if another issue comes up that needs to be backported to the ESR. It seems this is currently not an issue for the customer and I'm unsure if other customers have seen this. Happy to discuss further if needed.

PR #14974, opened 2020-07-06
[MM-26450] - Add ability to team member to manage channel members

nevyangelova:
#### Summary
Per new specification 'team_user' should be able to add members to channel. 

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-26450

Fix v5.26


hmhealey:
We should double check the spec on this. The new sidebar shouldn't require permissions changes, so the user might've just been missing the correct permissions when this was filed.

nevyangelova:
Closing in favour of a webapp change instead.

PR #14976, opened 2020-07-06
[MM-19640] Retrieve Compliance files from the System Console

hahmadia:
#### Summary

Add query parameter for downloading compliance report

This is currently following the patten of how we download files when uploaded on chat facing side (`getFile` in `api4/file.go`)

#### Ticket Link

https://mattermost.atlassian.net/browse/MM-19640

Depends on:
https://github.com/mattermost/enterprise/pull/695


api4/job.go
```diff
@@ -35,6 +37,37 @@ func getJob(c *Context, w http.ResponseWriter, r *http.Request) {
...
+	if forceDownload {
+		fileName := "actiance_export.xml"
+		fileMime := "application/xml"
+		if job.Data["export_type"] == "csv" {
+			fileName = "csv_export.zip"
+			fileMime = "applicaiton/zip"
+		}
+
+		filePath := "export/" + job.Id + "/" + fileName
+		data, readFileErr := c.App.ReadFile(filePath)
```
sbishel, line 50: Appears this is only being read to get the size of the file. Does it make more to write a function to just retrieve the filesize? I'm not sure how large these files get, but I would guess they are at least a few megabytes. And if it's coming from AWS, it would be retrieved twice.

api4/job.go
```diff
@@ -35,6 +37,37 @@ func getJob(c *Context, w http.ResponseWriter, r *http.Request) {
...
+		if readFileErr != nil {
+			c.Err = readFileErr
+			return
+		}
+
+		fileReader, err := c.App.FileReader(filePath)
+		if err != nil {
+			c.Err = err
+			c.Err.StatusCode = http.StatusNotFound
```
sbishel, line 59: I'm not sure it would ever get here. As wouldn't line 50 fail first.

api4/job.go
```diff
@@ -35,6 +37,37 @@ func getJob(c *Context, w http.ResponseWriter, r *http.Request) {
...
+	forceDownload, _ := strconv.ParseBool(r.URL.Query().Get("download"))
+	if forceDownload {
+		fileName := "actiance_export.xml"
+		fileMime := "application/xml"
+		if job.Data["export_type"] == "csv" {
+			fileName = "csv_export.zip"
+			fileMime = "applicaiton/zip"
+		}
+
+		filePath := "export/" + job.Id + "/" + fileName
```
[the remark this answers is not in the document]
hahmadia: Hey! So when I tried to define `const FILE_PATH = fmt.Sprintf("export/%s/%s", job.Id, model.FILE_NAME[job.Data["export_type"]])` , I would get the following error
`fmt.Sprintf("export/%s/%s", job.Id, model.FILE_NAME[job.Data["export_type"]]) (value of type string) is not constant`

However, I could do `const FILE_PATH = "export/" + job.Id + "/" + model.FILE_NAME[job.Data["export_type"]]` without an error. Thoughts? Should I do a non constant with fmt.sprintf or a constant with the concat of strings?

api4/job.go
```diff
@@ -35,6 +37,37 @@ func getJob(c *Context, w http.ResponseWriter, r *http.Request) {
 		return
 	}
 
+	forceDownload, _ := strconv.ParseBool(r.URL.Query().Get("download"))
+	if forceDownload {
+		fileName := "actiance_export.xml"
+		fileMime := "application/xml"
```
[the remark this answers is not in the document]
hahmadia: Done!

api4/job.go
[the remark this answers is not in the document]
hahmadia: In the meantime, I decided to go with the second option which doesn't use `fmt.sprintf` but still keeps the file path as a constant.

api4/job.go
```diff
@@ -35,6 +37,37 @@ func getJob(c *Context, w http.ResponseWriter, r *http.Request) {
...
+		if job.Data["export_type"] == "csv" {
+			fileName = "csv_export.zip"
+			fileMime = "applicaiton/zip"
+		}
+
+		filePath := "export/" + job.Id + "/" + fileName
+		data, readFileErr := c.App.ReadFile(filePath)
+		if readFileErr != nil {
+			c.Err = readFileErr
```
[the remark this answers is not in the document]
hahmadia: Good point. I just did a call on a valid job ID which I deleted the file for and called the endpoint, I got the following returned. It is returning the file path and it includes file name. Why are you suggesting that we don't want to return the file name?
```
{
    "id": "api.file.read_file.reading_local.app_error",
    "message": "Encountered an error reading from local server file storage.",
    "detailed_error": "open data/export/ztged6659i85fjkzqzb7o7fyuh/actiance_export.xml: no such file or directory",
    "request_id": "b7rog6dqrir95japc65kth4z7a",
    "status_code": 500
}
```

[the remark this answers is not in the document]
hahmadia: Check up with security on returning the full file path or not (currently happening for file uploads on chat facing side) 
We ideally shouldn't be returning the full path.

api4/job.go
hahmadia: **SELF NOTE** Make this another endpoint `/download`

api4/job.go
hahmadia: Done!

api4/job.go
hahmadia: Done!

api4/job.go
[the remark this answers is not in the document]
hahmadia: Done!

api4/job.go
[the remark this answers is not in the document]
hahmadia: I did notice that for the file uploads, the path was exposed if I manually deleted the file from the directory or the directory itself. We don't have an endpoint to delete the file directly. I was testing in community-daily where I uploaded the file and deleted the post and then tried doing a get call on that file ID. It returned an error saying that the file ID was not found in the database (indicating deletion of the file information in the database). 

However, regardless, I have sanitized the return error to blank out the detailed_mesage just in case as the message seems to be enough information to indicate the problem.

api4/job.go
```diff
@@ -4,16 +4,21 @@
...
 import (
+	"fmt"
 	"net/http"
+	"time"
 
 	"github.com/mattermost/mattermost-server/v5/audit"
 	"github.com/mattermost/mattermost-server/v5/model"
 )
 
+const FILE_PATH = "export/%s/%s"
```
agnivade, line 15: I don't see a need for this to be exported at a package level. Do we expect other methods to use this? If not, I would just make it a const inside the function.

model/job.go
```diff
@@ -30,6 +30,17 @@ const (
 	JOB_STATUS_WARNING          = "warning"
 )
 
+var FILE_INFO = map[string]map[string]string{
```
agnivade, line 33: `var fileInfo = ..`. There is no need to export this IMO.

api4/job.go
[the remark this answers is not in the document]
jupenur: I don't see an issue here. The only thing revealed is the path to the data directory (which is just `data`) buy default, and the rest is either random or static.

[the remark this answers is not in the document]
jupenur: Information disclosures like this would be a bigger issue if we were closed-source, but as it stands everything can be read from the sources anyway.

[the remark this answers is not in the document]
hahmadia: @jupenur Thank you Juhno for the input!

api4/job.go
hahmadia: Nope! Moved inside function.

model/job.go
hahmadia: Moved it inside the function.

api4/job.go
```diff
@@ -38,6 +42,56 @@ func getJob(c *Context, w http.ResponseWriter, r *http.Request) {
...
+	if err != nil {
+		mlog.Error(err.Error())
+		c.Err = err
+		c.Err.StatusCode = http.StatusNotFound
+		return
+	}
+	defer fileReader.Close()
+
+	// We are able to pass 0 for content size due to the fact that Golang's serveContent (https://golang.org/src/net/http/fs.go)
+	// already sets that for us (Line 291)
```
hahmadia, line 86: Is this comment needed?

sbishel: I would keep the comment, but remove line number as depending on version, that could change.

hahmadia:
@sbishel 
I've updated to check for the `is_downloadable` flag and updated the test to make sure we error one reports being downloaded without this flag set.

thefactremains:
note for documentation: 
We need to mention the config.json setting `DownloadExportResults` that must be set to true for this feature to work. And that all reports generated prior to this update will not be downloadable (only reports generated from this update onwards will be downloadable)

amyblais:
@hahmadia Does the new config setting need to be added to telemetry?

hahmadia:
> @hahmadia Does the new config setting need to be added to telemetry?

That is something I'm not sure about. @thefactremains @sbishel Do you guys know?

PR #14981, opened 2020-07-07
MM-26753 Create initial sidebar categories on demand if migration hasn't ran

hmhealey:
This works like the old `FileInfo` migration where `SidebarCategory` entries are created on demand when the client requests them.

A few things to note:
1. Assuming I understand how transactions work, this should prevent duplicate categories from being created even if this is called during migration because of https://github.com/mattermost/mattermost-server/pull/14870
2. This assumes that `GetSidebarCategories` will always be called before any other action involving the sidebar. I think it's safe to assume this because you can't fetch individual categories before they exist and unless the user is hitting the API directly, we won't allow them to create categories until the initial ones are loaded.
3. I am mildly concerned about a potential replication delay between databases when creating the categories on-demand since `CreateInitialSidebarCategories` uses the master DB and `GetSidebarCategories` uses a replica.

Also, I think on-demand migration in addition to the migration job is actually something we really needed. The migration job could be slow, and I think this also helps solve some problems in mixed-version clusters like the community cluster (like if a user joins a team on an older cluster and it doesn't know to create the categories for it).

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-26753

### DIFF
--- a/app/channel.go
+++ b/app/channel.go
@@ -2623,10 +2623,36 @@ func (a *App) GetSidebarCategories(userId, teamId string) (*model.OrderedSidebar
 			return nil, nErr
 		}
 
-		categories, err = a.Srv().Store.Channel().GetSidebarCategories(userId, teamId)
+		categories, err = a.waitForSidebarCategories(userId, teamId)
 	}
 
 	return categories, err
+}
+
+// waitForSidebarCategories is used to get a user's sidebar categories after they've been created since there may be
+// replication lag if any database replicas exist. It will wait until results are available to return them.
+func (a *App) waitForSidebarCategories(userId, teamId string) (*model.OrderedSidebarCategories, *model.AppError) {
+	if len(a.Config().SqlSettings.DataSourceReplicas) == 0 {
+		// The categories should be available immediately on a single database
+		return a.Srv().Store.Channel().GetSidebarCategories(userId, teamId)
+	}
+
+	now := model.GetMillis()
+
+	for model.GetMillis()-now < 12000 {
+		time.Sleep(100 * time.Millisecond)
+
+		categories, err := a.Srv().Store.Channel().GetSidebarCategories(userId, teamId)
+
+		if err != nil || len(categories.Categories) > 0 {
+			// We've found something, so return
+			return categories, err
+		}
+	}
+
+	mlog.Error("waitForSidebarCategories giving up", mlog.String("user_id", userId), mlog.String("team_id", teamId))
+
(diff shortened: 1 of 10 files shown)


app/channel.go
```diff
@@ -2614,7 +2614,19 @@ func (a *App) createInitialSidebarCategories(user *model.User, team *model.Team)
...
+	categories, err := a.Srv().Store.Channel().GetSidebarCategories(userId, teamId)
+
+	if len(categories.Categories) == 0 && err == nil {
+		// A user must always have categories, so migration must not have happened yet, and we should run it ourselves
+		nErr := a.createInitialSidebarCategories(userId, teamId)
+		if nErr != nil {
+			return nil, nErr
+		}
+
+		categories, err = a.Srv().Store.Channel().GetSidebarCategories(userId, teamId)
```
grundleborg, line 2626: Is replica lag going to be a problem here?

grundleborg: Ah I see you commented on this. Can we do some kind of wait-loop here? I think we've done that elsewhere with joining users to teams in the past? I.e. check if the result is still zero, and if it is, wait a bit and try again a couple of times, only then giving up and returning no categories.

hmhealey:
> Review was a little but cursory. One concern I have is: can the on-demand migration of individual users cause a transaction to fail in the background migration worker? If so, what will happen? Will it recover and continue, or will it permanently break the migration?

It'll still run fine. Previously, it would've created two identical categories for that one user, but I fixed that duplication by checking if those categories exist before creating them. Now, it can run the migration any number of times on an individual user, and it'll never create extra categories. I had also tried using a partial uniqueness index to prevent any potential duplication, but those only exist in Postgres, so they weren't an option.

app/channel.go
hmhealey: Yep, I'll take a look at that. If you happen to have any references on this, could you DM me a link?

hmhealey: Never mind. Found one: https://github.com/mattermost/mattermost-server/blob/master/app/channel.go#L393

agnivade: Ahh here we are :)

agnivade: So this maybe a stupid question, but it seems to me that we are doing read-after-write within a single request. Why can't 

```
nErr := a.createInitialSidebarCategories(userId, teamId)
		if nErr != nil {
			return nil, nErr
		}
```

return the categories itself?

It seems like we are creating a resource, but not returning that in that response itself, but rather again reading from the DB using the same identifier.

PR #14983, opened 2020-07-07
Fix jobserver run

jespino:
#### Summary
Fix jobserver run


jespino:
> I'm not familiar with how the server starts up. It's okay to call this after using `a.Srv()` to load the license?

Yes it is. Actually the `InitServer` now from the `app` perspective is for a late server initialization related to App stuff. Anything that needs the app to work is going to be instantiated  there under a `sync.Once` object, so, the first time somebody use the app, is going to configure certain things, that happens naturally in other context, but in this context wasn't happening fast enough.

cpanato:
I will merge this and wait for the job build master and redeploy daily and update the jobserver

PR #14984, opened 2020-07-07
MM-26750: Remove instances of InitBasic when it's not required

agnivade:
`th.InitBasic` initializes users, channels, and teams. But a lot of
tests don't use it at all, causing unncessary delays. A single `InitBasic`
call takes around 400ms. This PR removes 71 of them, approximately shaving
off 28 seconds. Arguably, this would be more because some tests run in loops.

#### Ticket link

https://mattermost.atlassian.net/browse/MM-26750

### DIFF
--- a/app/auto_responder_test.go
+++ b/app/auto_responder_test.go
@@ -12,7 +12,7 @@ import (
 )
 
 func TestSetAutoResponderStatus(t *testing.T) {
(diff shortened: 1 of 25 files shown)


agnivade:
Gentle ping on this @jespino

jespino:
This is only changing tests files, so QA is not needed.

PR #14988, opened 2020-07-08
MM-26793 Fix accidentally removing other users' favorites

hmhealey:
The missing where case made it so that, when one user favourites a channel, others have it removed from their Favorites. Gotta fight for those favourites!

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-26793


store/sqlstore/channel_store.go
```diff
@@ -3846,6 +3846,7 @@ func (s SqlChannelStore) UpdateSidebarCategories(userId, teamId string, categori
 			// Remove any old favorites
 			sql, args, _ := s.getQueryBuilder().Delete("Preferences").Where(
 				sq.Eq{
+					"UserId":   userId,
```
jespino, line 3849: Ouch!

PR #14989, opened 2020-07-08
[MM-24560] Always clear channel cache when scheme is deleted

fmunshi:
#### Summary
- Since a team scheme always has channel level role names also cached both channel and team caches have to be cleared... my previous PR cleared the Team cache but the channel cache was still only cleared when the scheme was a channel scheme, this fixes that.

- For 5.26 I will add a unit test to ensure that this behaviour always works as expected but for now this needs to go for 5.25 release

#### Ticket Link
- https://mattermost.atlassian.net/browse/MM-24560


amyblais:
@fmunshi Is there another dev who would be able to review this or would it be faster to revert the original PR?

fmunshi:
@amyblais I dont think it is worth it to revert the original PR. It has no negative effect it simply didnt solve the original issue. 
This bug has a workaround where the system admin can go into the system console and purge all caches immediately after deleting a team scheme to prevent this issue from arising. If this cant go for 5.25 then is there somewhere we can just document the stop-gap solution?

fmunshi:
Correct @agnivade ... thanks for the review!

PR #14990, opened 2020-07-09
PreferenceStore migration

rvillablanca:
#### Summary

Migration of `PreferenceStore` to return plain errors.

Fixes https://github.com/mattermost/mattermost-server/issues/15010

### DIFF
--- a/app/oauth.go
+++ b/app/oauth.go
@@ -160,7 +160,7 @@ func (a *App) AllowOAuthAppAccessToUser(userId string, authRequest *model.Author
 	}
 
 	if nErr := a.Srv().Store.Preference().Save(&model.Preferences{authorizedApp}); nErr != nil {
-		mlog.Error("error saving store prefrence", mlog.Err(nErr))
+		mlog.Error("error saving store preference", mlog.Err(nErr))
 		return authRequest.RedirectUri + "?error=server_error&state=" + authRequest.State, nil
 	}
 
--- a/app/oauth.go
+++ b/app/oauth.go
@@ -158,8 +158,8 @@ func (a *App) AllowOAuthAppAccessToUser(userId string, authRequest *model.Author
 		Value:    authRequest.Scope,
 	}
 
-	if err = a.Srv().Store.Preference().Save(&model.Preferences{authorizedApp}); err != nil {
-		mlog.Error("error saving store prefrence", mlog.Err(err))
+	if nErr := a.Srv().Store.Preference().Save(&model.Preferences{authorizedApp}); nErr != nil {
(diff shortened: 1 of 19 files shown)


app/preference.go
```diff
@@ -12,20 +13,18 @@ import (
 func (a *App) GetPreferencesForUser(userId string) (model.Preferences, *model.AppError) {
 	preferences, err := a.Srv().Store.Preference().GetAll(userId)
 	if err != nil {
-		err.StatusCode = http.StatusBadRequest
-		return nil, err
+		return nil, model.NewAppError("GetAlGetPreferencesForUser", "app.preference.get_all.app_error", nil, err.Error(), http.StatusBadRequest)
```
agnivade, line 16: Shouldn't this be "GetPreferencesForUser"?

app/oauth.go
```diff
@@ -159,8 +159,8 @@ func (a *App) AllowOAuthAppAccessToUser(userId string, authRequest *model.Author
 		Value:    authRequest.Scope,
 	}
 
-	if err = a.Srv().Store.Preference().Save(&model.Preferences{authorizedApp}); err != nil {
-		mlog.Error("error saving store prefrence", mlog.Err(err))
+	if nErr := a.Srv().Store.Preference().Save(&model.Preferences{authorizedApp}); nErr != nil {
+		mlog.Error("error saving store prefrence", mlog.Err(nErr))
```
agnivade, line 163: Let's fix the typo while we are here :)

app/preference.go
rvillablanca: :ok:

app/oauth.go
rvillablanca: :ok:

app/preference.go
```diff
@@ -49,8 +47,14 @@ func (a *App) UpdatePreferences(userId string, preferences model.Preferences) *m
 	}
 
 	if err := a.Srv().Store.Preference().Save(&preferences); err != nil {
-		err.StatusCode = http.StatusBadRequest
-		return err
+		var appErr *model.AppError
+		switch {
+		case errors.As(err, &appErr):
+			appErr.StatusCode = http.StatusBadRequest
+			return appErr
```
jespino, line 54: nit: There's only one case where it returns an `AppError` and is when is calling `IsValid` and the `IsValid` for `Preference` always return a `http.StatusBadRequest` so I think the overwrite of the status here is not needed.

rvillablanca: Will remove that overwrite then, thanks!

agnivade:
Needs EE changes again. I am going to close this and send a new PR with EE changes.

rvillablanca:
Sure!

PR #14991, opened 2020-07-09
MM-26575: Move app initialization inside Websocket router out of handler

agnivade:
The websocket router struct is shared amongst multiple handlers. Therefore,
there is a race condition while setting the app field and reading from it.

We move the initialization of the app field when the router struct is initialized.

### Ticket link

https://mattermost.atlassian.net/browse/MM-26575


app/app.go
```diff
@@ -111,10 +111,6 @@ func (a *App) InitServer() {
 		}
 		a.srv.RunJobs()
 	})
-	a.accountMigration = a.srv.AccountMigration
-	a.ldap = a.srv.Ldap
-	a.notification = a.srv.Notification
-	a.saml = a.srv.Saml
```
agnivade, line 117: @jespino - this looks to be an artifact of the fakeapp migration. If you see L68-71, they are already being set there. Setting them again to the same value does not seem to be intentional. Let me know.

agnivade:
And with this PR, we finally reach 0 races while running the app with a load test. :)

agnivade:
Please hold off on QA review until I fix the tests.

app/app.go
agnivade: Ok it seems like removing these lines broke the EE tests. Looking into it.

agnivade: Some fields were missing in `ServerConnector`. Hopefully it's fixed now.

app/options.go
```diff
@@ -107,6 +107,9 @@ func ServerConnector(s *Server) AppOption {
 		a.searchEngine = s.SearchEngine
 		a.messageExport = s.MessageExport
 		a.metrics = s.Metrics
+		a.ldap = s.Ldap
```
jespino, line 110: Actually, that is not going to work, the `Server.Ldap`, `Server.Saml`, etcetera are late initialized during the first `App.InitServer` execution. the lines before were needed because you need to set up the `App` layer pointers to that variables. Probably the solution is to not use any `App.{ldap,saml,notification,accountMigration}` call, and use the `App.Srv().{Ldap,Saml,Notification,AccountMigration}`.

Anyway, I would revert this part of the changes, and keep the others.

agnivade: Oh I see the problem now. The late initialization isn't the issue. The issue is the the `sync.Once` is done at a server level than at app level. So if you initialize a new app from an existing server, the fields don't get set. Yeah, I think we should remove these app fields and directly use the server fields.

I will send another PR to fix those. Because without fixing the race, this PR isn't much useful.

agnivade: @jespino - This is resolved now after my other PR is now merged to master, and rebased with this. PTAL.

agnivade:
Gentle ping on this @jespino

PR #14992, opened 2020-07-09
Fix i18n

metanerd:
#### Summary
https://github.com/mattermost/mattermost-mobile/pull/4552 . PR is empty, but the commits diverged. This should fix the merge problems with weblate.


metanerd:
Turns out if I use `Merge` instead of `Rebase`, I do not have to create this PR. Merge > Rebase!

PR #15004, opened 2020-07-10
Moving store layers into independent packages

jespino:
#### Summary
Moving store layers into independent packages

#### Ticket Link
[MM-26835](https://mattermost.atlassian.net/browse/MM-26835)


jespino:
To give some context, it was taking around 5 seconds to compile the `store` package, from where almost 3 seconds was for the open tracing layer, and 1.5 from the timer layer. now the store layer compiles way faster, and each layer compilation is cached independently.

PR #15009, opened 2020-07-10
[MM-25990] Add filters to search all channels

fmunshi:
#### Summary
- Add the ability to filter channels by `TeamIds`, `Public`, `Private`, `GroupConstrained`, `ExcludeGroupConstrained`

#### Ticket Link
- https://mattermost.atlassian.net/browse/MM-25990

#### Related PR:
https://github.com/mattermost/mattermost-api-reference/pull/547


model/channel_search.go
```diff
@@ -11,11 +11,18 @@ import (
...
+	Term                    string   `json:"term"`
+	ExcludeDefaultChannels  bool     `json:"exclude_default_channels"`
+	NotAssociatedToGroup    string   `json:"not_associated_to_group"`
+	TeamIds                 []string `json:"team_ids"`
+	GroupConstrained        bool     `json:"group_constrained"`
+	ExcludeGroupConstrained bool     `json:"exclude_group_constrained"`
+	Public                  bool     `json:"public"`
+	Private                 bool     `json:"private"`
+	IncludeDeleted          bool     `json:"include_deleted"`
+	Deleted                 bool     `json:"deleted"`
```
agnivade, line 23: I have some concerns about these being value fields. This means that they will fall back to their default values if in any case, the UI forgets to set them in the HTTP request. I see that in the code you are only checking for non-default values, but I am a bit concerned if in the future, somebody adds a condition which checks for the negative value for these fields.

Are all code paths from the app explicitly set these fields to either true or false?

model/channel.go
```diff
@@ -120,12 +120,18 @@ type ChannelModeratedRolesPatch struct {
...
-	NotAssociatedToGroup   string
-	ExcludeDefaultChannels bool
-	IncludeDeleted         bool
-	ExcludeChannelNames    []string
-	Page                   *int
-	PerPage                *int
+	NotAssociatedToGroup    string
+	ExcludeDefaultChannels  bool
+	IncludeDeleted          bool
+	Deleted                 bool
```
agnivade, line 126: What is the difference between these? Is it possible to use just one?

fmunshi: One of them is to include deleted results in the response and one of them is to return exclusively deleted results.
I dont think there is a way to do this without using 2 different options

model/channel_search.go
fmunshi: The default code path sets false for all of the above options.

model/channel.go
agnivade: Ah I see. Perhaps change `Deleted` to `JustDeleted/OnlyDeleted`? 0/5. IMO Deleted and IncludeDeleted is a bit confusing. Deleted does not clarify that it exclusively returns deleted results.

fmunshi: I think it makes more sense to leave it as `Deleted` instead of `OnlyDeleted` since `Private` and `Public` work the same way i.e. passing just `Private` returns "Only" Private channels

amyblais:
@jespino Kind note for reviewers that feature complete for v5.26 is tomorrow,

api4/channel_test.go
```diff
@@ -1225,109 +1225,166 @@ func TestSearchAllChannels(t *testing.T) {
...
-	search.Term = th.BasicPrivateChannel.Name
-	channels, resp = th.SystemAdminClient.SearchAllChannels(search)
-	CheckNoError(t, resp)
+	th.App.UpdateConfig(func(cfg *model.Config) {
+		*cfg.TeamSettings.ExperimentalViewArchivedChannels = true
+	})
 
-	assert.Len(t, *channels, 1)
-	assert.Equal(t, th.BasicPrivateChannel.Id, (*channels)[0].Id)
+	testCases := []struct {
```
jespino, line 1258: :heart:

store/sqlstore/channel_store.go
```diff
@@ -2854,6 +2855,30 @@ func (s SqlChannelStore) channelSearchQuery(term string, opts store.ChannelSearc
...
+	if len(opts.TeamIds) > 0 {
+		query = query.Where(sq.Eq{"c.TeamId": opts.TeamIds})
+	}
+
+	if opts.GroupConstrained {
+		query = query.Where(sq.Eq{"c.GroupConstrained": true})
+	} else if opts.ExcludeGroupConstrained {
+		query = query.Where(sq.Or{
+			sq.NotEq{"c.GroupConstrained": true},
+			sq.Eq{"c.GroupConstrained": nil},
```
jespino, line 2867: nit: this is not redundant? I mean, if the `GroupConstrained` is `true`, it is not `nil`, right?

store/sqlstore/channel_store.go
```diff
@@ -2854,6 +2855,30 @@ func (s SqlChannelStore) channelSearchQuery(term string, opts store.ChannelSearc
...
+		})
+	}
+
+	if opts.Public && opts.Private {
+		query = query.Where(sq.Or{
```
jespino, line 2872: nit: this is already done in the base query, no harm in it but a bit of repetition and maybe a future source of problems changing one and not the other. Probably I would remove that from the base query, and keep all the logic related to this here together, probably something like:

```go
if opts.Private && !opts.Public {
   ...filter by Private
} else if opts.Public && !opts.Private {
   ...filter by Public
} else {
   ...filter by Public or Private
}
```

store/sqlstore/channel_store.go
```diff
@@ -2825,8 +2825,9 @@ func (s SqlChannelStore) channelSearchQuery(term string, opts store.ChannelSearc
 			OrderBy("c.DisplayName, t.DisplayName").
 			Limit(uint64(limit))
 	}
-
-	if !opts.IncludeDeleted {
+	if opts.Deleted {
+		query = query.Where(sq.NotEq{"c.DeleteAt": int(0)})
+	} else if !opts.IncludeDeleted {
```
jespino, line 2830: nit: I'm not sure if this is the expected behavior. I would expect, If I ask to the api for channels, and I don't want to include the deleted ones in the list, and then I filter only by the deleted ones, I would expect and empty list, not to ignore one of my requests. I understand that makes no sense from the users perpective, but if somebody from the client is relaying on the idea of "I'm not including deleted channels, but I don't care about setting or not the Deleted filter" can end up with some weird behaviors.

store/sqlstore/channel_store.go
fmunshi: Unfortunately it doesnt work that way... if a column has a `NULL` value then it needs to be compared with `NULL`, so a column that has a value of `NULL` doesnt get returned if you do `Column <> true `

store/sqlstore/channel_store.go
fmunshi: Done

srkgupta:
Webapp PR is also expected to come along with this server PR. This PR will be tested together.

PR #15016, opened 2020-07-13
MM-26826 Populate orphaned channels in a transaction

hmhealey:
There's currently logic where orphaned channels (channels without a category) are automatically added to the DMs and Channels categories to prevent them from being inaccessible. This is done by calling `completePopulatingCategoryChannels` on those categories before returning them to the client. Any issues with that function are more prominent in DMs since we don't create SidebarChannels entries for the DMs category because it doesn't support manual ordering.

The problem we have currently is that any changes made are ran on the database master and `completePopulatingCategoryChannels` is ran on a replica, so if there's any lag between the two servers, the call to `completePopulatingCategoryChannels` will include stale data. This causes duplicated DMs when moving a DM into another category (since the DM that was moved still appears in DMs) and disappearing DMs when moving it back (since DMs doesn't include the now-orphaned DM yet).

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-26826

### DIFF
--- a/store/sqlstore/channel_store.go
+++ b/store/sqlstore/channel_store.go
@@ -3562,6 +3562,25 @@ func (s SqlChannelStore) CreateSidebarCategory(userId, teamId string, newCategor
 }
 
 func (s SqlChannelStore) completePopulatingCategoryChannels(category *model.SidebarCategoryWithChannels) (*model.SidebarCategoryWithChannels, *model.AppError) {
+	transaction, err := s.GetMaster().Begin()
+	if err != nil {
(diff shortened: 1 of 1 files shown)


hmhealey:
@agnivade Not that I'm aware of. I posted in the Server channel to ask for that, but I don't believe anyone responded.

agnivade:
Right, thanks.

jgilliam17:
Thank you @hmhealey 
DMs are no longer doubling 
Found one issue with the doubling of GMs.
1) Create a GM and move to custom category
2) Click on DM + and select the same GM participants to open the conversation again
Observed: GM is created and placed in DM category, instead of opening an existing conversation 
Screen Shot 2020-07-15 at 8 54 56 AM

PR #15019, opened 2020-07-14
MM-26836: Remove some app fields which cause race conditions

agnivade:
`(a *App) InitServer` was being called from multiple goroutines
and were modifying the fields of the App struct concurrently.
However, the fields were just reflecting the Server fields,
and we already have public getters for those fields.

Therefore, we remove those fields and just change the code to return
the server field directly.


agnivade:
Well, this was easier than I thought. It's good that the entire codebase uses those getters.

agnivade:
@ogi-m - Some general smoke testing would be sufficient here. Thanks.

app/options.go
```diff
@@ -100,7 +100,6 @@ func ServerConnector(s *Server) AppOption {
 		a.log = s.Log
 		a.notificationsLog = s.NotificationsLog
 
-		a.accountMigration = s.AccountMigration
 		a.cluster = s.Cluster
```
jespino, line 103: Not for this PR, but would be great to get rid of all this when is possible.

agnivade: https://mattermost.atlassian.net/browse/MM-27169

agnivade:
@ogi-m - Wondering if you had a chance to look at this? :)

PR #15021, opened 2020-07-14
MM-26871 Add test for sidebar when moving channels between teams

hmhealey:
I haven't been able to reproduce the bug we saw on community with the private dev channel, but here's a unit test at least to try to catch that case in the future.

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-26871


hmhealey:
Skipping QA review since this is just adding a unit test.

app/channel_test.go
```diff
@@ -103,78 +103,121 @@ func TestRemoveAllDeactivatedMembersFromChannel(t *testing.T) {
...
 
-	_, err = th.App.AddUserToTeam(targetTeam.Id, th.BasicUser.Id, "")
-	require.Nil(t, err)
+		_, err = th.App.AddUserToChannel(th.BasicUser2, channel1)
+		require.Nil(t, err)
 
-	_, err = th.App.AddUserToChannel(th.BasicUser, channel1)
-	require.Nil(t, err)
+		err = th.App.MoveChannel(targetTeam, channel1, th.BasicUser)
+		require.NotNil(t, err, "Should have failed due to mismatched members.")
```
ethervoid, line 136: I'd use `require.Error`

app/channel_test.go
```diff
@@ -103,78 +103,121 @@ func TestRemoveAllDeactivatedMembersFromChannel(t *testing.T) {
...
-	require.Nil(t, err)
-	_, err = th.App.AddUserToChannel(th.BasicUser, channel2)
-	require.Nil(t, err)
+		_, err = th.App.UpdateActive(deacivatedUser, false)
+		require.Nil(t, err)
 
-	_, err = th.App.AddUserToChannel(deacivatedUser, channel2)
-	require.Nil(t, err)
+		err = th.App.MoveChannel(targetTeam, channel2, th.BasicUser)
+		require.NotNil(t, err, "Should have failed due to mismatched deacivated member.")
```
ethervoid, line 162: Same here

app/channel_test.go
```diff
@@ -103,78 +103,121 @@ func TestRemoveAllDeactivatedMembersFromChannel(t *testing.T) {
 }
 
 func TestMoveChannel(t *testing.T) {
-	th := Setup(t).InitBasic()
-	defer th.TearDown()
-	var err *model.AppError
+	t.Run("should move channels between teams", func(t *testing.T) {
```
ethervoid, line 106: Please make a little refactor and create a different t.Run scenarios where:

- Move a channel without problem
- Move a channel fails because of member mismatch
- Move a channel with deactivated users fails
- Move a channel with deactivated users with the option works
- Move a channel with no members

This way is cleaner and readable when you're checking the tests

hmhealey: It would be nicer, but I don't have the time to do that now. This PR is just focused on adding the new test to confirm that this works with the sidebar correctly, and moving the rest of this into a subtest was just a side effect of that.

ethervoid: Ok, but is not just nicer is something really desirable. For example, when I'm working on something new, I tend to read the tests to know how it works so treating them as any other production code is a good practice. My suggestion was just cut that code and create differen test cases

hmhealey:
I manually cherry-picked the commit.

PR #15026, opened 2020-07-14
Rudder key and dataplane in Config

Adovenmuehle:
#### Summary
To allow for ease of testing telemetry changes, we should make it so that the rudder key and dataplane URL can be customized through the config or environment. This would allow developers to set the key in their local environment to test telemetry and also allow us to inject a QA key into instances spun up with matterwick.


Adovenmuehle:
@jespino @grundleborg this is still a WIP but I wanted to shoot it over to you earlier than later. I haven't written much Go so please tell me everything I've done wrong.

My idea is either people can set a dev rudder key in their config.json or override it with an env var.

cc @jaydeland

grundleborg:
@Adovenmuehle I see what you are trying to do there, but I'm a bit uncomfortable with adding fields to the config file that are really only there for developers and silently ignored in deployment.

Perhaps instead we could have a "secret" environment variable that we use for testing with, that isn't part of the regular config system but is documented in the developer docs instead?

Adovenmuehle:
@grundleborg yeah I can agree with that. 

Would you essentially be suggesting using `os.Getenv`?

Adovenmuehle:
The upside to having the rudder config as a real config instead of just reading from the environment is that we could then read it from the client instead of having to separately inject it there. I'm trying to figure out how to have those keys injected in the webapp at build time but not totally sure how to do it.

Adovenmuehle:
@grundleborg can you give me some guidance on what I need to do here to get it merged?

grundleborg:
@amyblais as on the webapp ticket, I see you've set the milestone for this to 5.27 but not added the Cherry-Pick-Approved label. Is this intended to be cherry-picked to 5.27, or merged just to master for 5.28?

Adovenmuehle:
I went through the telemetry tickets and I think we should be ok just merging this to master for 5.28.

hanzei:
@jespino These changes doesn't seem to be in https://github.com/mattermost/mattermost-server/blob/master/services/telemetry/telemetry.go any longer. They seem to got removed by https://github.com/mattermost/mattermost-server/commit/44079785eb126b13f593562f7450ad49abf1bd8a#diff-6d075395bdd93d9b78959b73d7a2f3d9. Was this intended?

jespino:
@hanzei I think it can be lost during the merges from master, probably we should try to reapply it.

hanzei:
@jespino Should I file a ticket?

Adovenmuehle:
Yes please. Thanks @hanzei for catching this.

hanzei:
@Adovenmuehle @jespino Filled https://mattermost.atlassian.net/browse/MM-28952

PR #15030, opened 2020-07-15
MM-26891: return error if option not set instead of soft delete for permanent team deletion

ashishbhate:
#### Summary

- If the `EnableAPITeamDeletion` option is not set, then PermanentDeleteTeam returns an error instead of falling back to soft delete. This change was made to bring the behaviour into alignment with admin expectations and docs - https://docs.mattermost.com/administration/config-settings.html#enable-api-team-deletion

#### Ticket Link

- https://mattermost.atlassian.net/browse/MM-26891

### DIFF
--- a/api4/team.go
+++ b/api4/team.go
@@ -367,8 +367,12 @@ func deleteTeam(c *Context, w http.ResponseWriter, r *http.Request) {
 	}
 
 	var err *model.AppError
-	if c.Params.Permanent && *c.App.Config().ServiceSettings.EnableAPITeamDeletion {
-		err = c.App.PermanentDeleteTeamId(c.Params.TeamId)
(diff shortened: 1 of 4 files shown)


mgdelacroix:
@ashishbhate should we explicitly state in both the docs for the API and the configuration properties that the configuration value doesn't apply to local mode?

ashishbhate:
> should we explicitly state in both the docs for the API and the configuration properties that the configuration value doesn't apply to local mode?

I was going to update the configuration docs at https://docs.mattermost.com/administration/config-settings.html#settings-configurable-only-in-config-json after the user delete PR was merged.

I hadn't thought of the API documentation. I'll send a PR there as well.

PR #15031, opened 2020-07-15
MM-25563: Add endpoint to fetch only archived channels

agnivade:
The `getChannelsForTeamForUser` can also return archived channels now.
This can be problematic for situations where a user is member of a huge
number of archived channels. And since this endpoint is being polled,
it can cost a significant amount of bandwidth.

Therefore, we add an additional query param `last_delete_at`, which will additionally filter the archived channels
greater than a timestamp. The logic for the client should be to use the `getChannelsForTeamForUser` with `include_deleted=true` and `last_delete_at=0` and store the latest delete_at timestamp. If there are no archived channels,
then keep using `last_delete_at=0`. But, when archived channels start to appear, note the the latest `delete_at` timestamp
for that team somewhere. And then use that time in the `last_delete_at` field in subsequent calls.

This way the `last_delete_at` value keeps increasing as time goes on. And we only query for a subset of the data
at a given point in time.

#### Ticket link

https://mattermost.atlassian.net/browse/MM-25563


enahum:
@agnivade what would be the flow then? and what do  we need to be mindful of when connecting to a server that does not have this endpoint available ?

agnivade:
I will clarify the flow in detail when I actually create tickets for app and mobile. But mostly, it's written in the 2nd para of the PR body. Let me know if you foresee any problems with it.

The API is going to be documented in api.mattermost.com. I don't believe we document the way clients should consume the API. But I am happy to do so if you point me to the right place. Thanks.

agnivade:
So after discussing with @enahum, I am going to change it to include the functionality in a single endpoint with the query string.

agnivade:
@jespino @ethervoid - ready for review now. I updated it to use a single endpoint. And also changed the sql query to use squirrel :)

Added store tests, api4 tests, and also tested manually.

model/client4.go
```diff
@@ -2442,6 +2442,18 @@ func (c *Client4) GetChannelsForTeamForUser(teamId, userId string, includeDelete
 	return ChannelSliceFromJson(r.Body), BuildResponse(r)
 }
 
+// GetChannelsForTeamForUserWithLastDeleteAt returns a list channels of a team for a user, additionally filtered with lastDeleteAt. This does not have any effect if includeDeleted is set to false.
+func (c *Client4) GetChannelsForTeamForUserWithLastDeleteAt(teamId, userId string, includeDeleted bool, lastDeleteAt int, etag string) ([]*Channel, *Response) {
```
ethervoid, line 2446: What about `GetChannelsForTeamAndUserWithLastDeleteAt`?

store/sqlstore/channel_store.go
```diff
@@ -1133,14 +1133,43 @@ func (s SqlChannelStore) PermanentDeleteMembersByChannel(channelId string) *mode
 	return nil
 }
 
-func (s SqlChannelStore) GetChannels(teamId string, userId string, includeDeleted bool) (*model.ChannelList, error) {
-	query := "SELECT Channels.* FROM Channels, ChannelMembers WHERE Id = ChannelId AND UserId = :UserId AND DeleteAt = 0 AND (TeamId = :TeamId OR TeamId = '') ORDER BY DisplayName"
+func (s SqlChannelStore) GetChannels(teamId string, userId string, includeDeleted bool, lastDeleteAt int) (*model.ChannelList, error) {
+	query := s.getQueryBuilder().
```
ethervoid, line 1137: :heart:

api4/channel_test.go
```diff
@@ -947,6 +947,11 @@ func TestGetChannelsForTeamForUser(t *testing.T) {
 		channels, resp = Client.GetChannelsForTeamForUser(th.BasicTeam.Id, th.BasicUser.Id, false, "")
 		CheckNoError(t, resp)
 		assert.Equal(t, 5, len(channels))
+
```
ethervoid, line 950: Please add some tests for `GetChannelsForTeamForUserWithLastDeleteAt` here

agnivade: You caught me :)

agnivade: Oh I remembered now why I didn't add that test. It's because to test the feature properly, I needed to set a time.Sleep, which I didn't like. I can filter with an old time, but then that becomes the same as getting all archived channels. I can do that as a sanity check if you prefer.

agnivade:
ping @jespino

agnivade:
@saturninoabril - This needs some API related testing, which is why assigning this to you. The steps are outlined in the ticket.

justinegeffen:
@agnivade, I've reviewed this change and it looks like the update would be here: https://developers.mattermost.com/extend/plugins/server/reference/#API.GetChannelsForTeamForUser. Could you confirm that this is correct? Thank you!

agnivade:
That is a different API for plugins. This PR does not touch that. No update is needed there.

This PR changes the REST API - which was done here: https://github.com/mattermost/mattermost-api-reference/pull/551/files.

I wasn't even aware that this documentation existed. This is just a straight copy of plugin/api.go, which is anyways available in godocs. We should probably just link the user to that page instead of manually duplicating everything again in a separate page.

justinegeffen:
Thanks, @agnivade! The API documentation is updated in a different workflow and isn't part of the documentation at docs.mattermost.com. I *think* @jwilander is the right person to talk to about the [API docs](https://api.mattermost.com/) and updating them.

@amyblais, I've marked this as ``Docs/Not needed`` as no changes to admin/user docs are required.

agnivade:
API docs are already updated with this PR:

> This PR changes the REST API - which was done here: https://github.com/mattermost/mattermost-api-reference/pull/551/files.

The site gets updated after every push to master.

PR #15038, opened 2020-07-15
[MM-26877] Fix flaky TestPluginAPIUpdateUserPreferences

streamer45:
#### Summary

PR *attempts* to fix flaky `TestPluginAPIUpdateUserPreferences`.
I wasn't able to reproduce it locally but I suspect the problem was that the order of the returned preferences cannot be relied upon.
Feel free to give it a spin and check if this actually fixes it.

#### Ticket Link

https://mattermost.atlassian.net/browse/MM-26877


jespino:
Removed the QA review because is only changing test files.

PR #15039, opened 2020-07-15
Updated the deleted_at timestamp for deactivated direct message channel

Vars-07:
#### Summary

When the user account is deactivated, direct message channels are not updated with deleted time

#### Ticket Link

https://mattermost.atlassian.net/browse/MM-26941

### DIFF
--- a/app/user.go
+++ b/app/user.go
@@ -979,8 +979,18 @@ func (a *App) invalidateUserChannelMembersCaches(userId string) *model.AppError 
 	return nil
 }
 
+func (a *App) updateDirectChannelForUser(channel *model.Channel, updatedTime int64) error {
+	err := a.Srv().Store.Channel().Delete(channel.Id, updatedTime)
+	if err != nil {
+		return err
+	}
+
+	return nil
+}
+
 func (a *App) UpdateActive(user *model.User, active bool) (*model.User, *model.AppError) {
-	user.UpdateAt = model.GetMillis()
+	updateTime := model.GetMillis()
+	user.UpdateAt = updateTime
 	if active {
 		user.DeleteAt = 0
 	} else {
@@ -997,6 +1007,18 @@ func (a *App) UpdateActive(user *model.User, active bool) (*model.User, *model.A
 		if err := a.userDeactivated(ruser.Id); err != nil {
 			return nil, err
 		}
+	}
+
+	directChannels, e := a.Srv().Store.Channel().GetDirectChannelsForUser(user.Id)
+	if e != nil {
+		return nil, e
+	}
+
+	for _, channel := range directChannels {
+		if active {
+			updateTime = 0
+		}
+		a.updateDirectChannelForUser(channel, updateTime)
 	}
 
 	a.invalidateUserChannelMembersCaches(user.Id)
--- a/store/sqlstore/channel_store.go
+++ b/store/sqlstore/channel_store.go
@@ -3442,6 +3442,39 @@ func (s SqlChannelStore) UserBelongsToChannels(userId string, channelIds []strin
 	return c > 0, nil
 }
 
+func (s SqlChannelStore) GetDirectChannelsForUser(userId string) ([]*model.Channel, *model.AppError) {
+	queryFormat := `
+			SELECT
+				C.*,
+				OtherUsers.Username as DisplayName
+			FROM
+				Channels AS C
+			JOIN
+				ChannelMembers AS CM ON CM.ChannelId = C.Id
+			INNER JOIN (
+				SELECT
+					ICM.ChannelId AS ChannelId, IU.Username AS Username
+				FROM
+					Users as IU
+				JOIN
+					ChannelMembers AS ICM ON ICM.UserId = IU.Id
+				WHERE
+					IU.Id != :UserId
+					%v
+				) AS OtherUsers ON OtherUsers.ChannelId = C.Id
+			WHERE
+			    C.Type = 'D'
+				AND CM.UserId = :UserId`
+
+	var channels model.ChannelList
+
+	if _, err := s.GetReplica().Select(&channels, fmt.Sprintf(queryFormat, ""), map[string]interface{}{"UserId": userId}); err != nil {
+		return nil, model.NewAppError("SqlChannelStore.GetDirectChannelsForUser", "store.sql_channel.direct_channels_for_user.app_error", nil, err.Error(), http.StatusInternalServerError)
+	}
+
+	return channels, nil
+}
+
 func (s SqlChannelStore) UpdateMembersRole(channelID string, userIDs []string) *model.AppError {
 	sql := fmt.Sprintf(`
 		UPDATE
--- a/store/store.go
+++ b/store/store.go
@@ -233,6 +233,7 @@ type ChannelStore interface {
 	RemoveAllDeactivatedMembers(channelId string) *model.AppError
 	GetChannelsBatchForIndexing(startTime, endTime int64, limit int) ([]*model.Channel, *model.AppError)
 	UserBelongsToChannels(userId string, channelIds []string) (bool, *model.AppError)
+	GetDirectChannelsForUser(userId string) ([]*model.Channel, *model.AppError)
 
 	// UpdateMembersRole sets all of the given team members to admins and all of the other members of the team to
 	// non-admin members.
--- a/store/storetest/mocks/ChannelStore.go
+++ b/store/storetest/mocks/ChannelStore.go
@@ -2109,3 +2109,28 @@ func (_m *ChannelStore) UserBelongsToChannels(userId string, channelIds []string
 
 	return r0, r1
 }
+
+// GetDirectChannelsForUser provides a mock function with given fields: userId
+func (_m *ChannelStore) GetDirectChannelsForUser(userId string) ([]*model.Channel, *model.AppError) {
+	ret := _m.Called(userId)
+
+	var r0 []*model.Channel
+	if rf, ok := ret.Get(0).(func(string) []*model.Channel); ok {
+		r0 = rf(userId)
+	} else {
+		if ret.Get(0) != nil {
+			r0 = ret.Get(0).([]*model.Channel)
+		}
+	}
+
+	var r1 *model.AppError
+	if rf, ok := ret.Get(1).(func(string) *model.AppError); ok {
+		r1 = rf(userId)
+	} else {
+		if ret.Get(1) != nil {
+			r1 = ret.Get(1).(*model.AppError)
+		}
+	}
+
+	return r0, r1
+}


hanzei:
Hi @Vars-07,

Thanks for the PR :+1: Could you please run `make store-layers` and commit the changes to fix the CI error?

Vars-07:
@hanzei Updated. Have a look :)

app/user.go
```diff
@@ -999,6 +1009,18 @@ func (a *App) UpdateActive(user *model.User, active bool) (*model.User, *model.A
 		}
 	}
 
+	directChannels, e := a.Srv().Store.Channel().GetDirectChannelsForUser(user.Id)
```
agnivade, line 1012: s/e/err

app/user.go
```diff
@@ -999,6 +1009,18 @@ func (a *App) UpdateActive(user *model.User, active bool) (*model.User, *model.A
...
+	if e != nil {
+		return nil, e
+	}
+
+	for _, channel := range directChannels {
+		if active {
+			updateTime = 0
+		}
```
agnivade, line 1020: Any reason this is inside the loop? The `active` variable doesn't seem to change.

app/user.go
```diff
@@ -999,6 +1009,18 @@ func (a *App) UpdateActive(user *model.User, active bool) (*model.User, *model.A
...
+		a.updateDirectChannelForUser(channel, updateTime)
```
agnivade, line 1021: Wondering if it even makes sense to refactor this to a separate method? It's just a single line call.

store/sqlstore/channel_store.go
```diff
@@ -3442,6 +3442,39 @@ func (s SqlChannelStore) UserBelongsToChannels(userId string, channelIds []strin
 	return c > 0, nil
 }
 
+func (s SqlChannelStore) GetDirectChannelsForUser(userId string) ([]*model.Channel, *model.AppError) {
```
agnivade, line 3445: Let's return `error` instead of `*model.AppError`. Take a look at some other methods of the store.

store/sqlstore/channel_store.go
```diff
@@ -3442,6 +3442,39 @@ func (s SqlChannelStore) UserBelongsToChannels(userId string, channelIds []strin
...
+			INNER JOIN (
+				SELECT
+					ICM.ChannelId AS ChannelId, IU.Username AS Username
+				FROM
+					Users as IU
+				JOIN
+					ChannelMembers AS ICM ON ICM.UserId = IU.Id
+				WHERE
+					IU.Id != :UserId
+					%v
```
agnivade, line 3463: I am unable to understand what is the `%v` doing here.

app/user.go
```diff
@@ -979,8 +979,18 @@ func (a *App) invalidateUserChannelMembersCaches(userId string) *model.AppError
...
+func (a *App) updateDirectChannelForUser(channel *model.Channel, updatedTime int64) error {
+	err := a.Srv().Store.Channel().Delete(channel.Id, updatedTime)
+	if err != nil {
+		return err
+	}
+
+	return nil
+}
+
 func (a *App) UpdateActive(user *model.User, active bool) (*model.User, *model.AppError) {
```
agnivade, line 991: Can we add a test for this?

store/sqlstore/channel_store.go
```diff
@@ -3442,6 +3442,39 @@ func (s SqlChannelStore) UserBelongsToChannels(userId string, channelIds []strin
 	return c > 0, nil
 }
 
+func (s SqlChannelStore) GetDirectChannelsForUser(userId string) ([]*model.Channel, *model.AppError) {
+	queryFormat := `
+			SELECT
```
agnivade, line 3447: A store layer test for this would be great.

app/user.go
Vars-07: Didn't get, can you please elaborate?

app/user.go
Vars-07: Done

app/user.go
Vars-07: Done

store/sqlstore/channel_store.go
Vars-07: Done, was trying something forget to remove :p

store/sqlstore/channel_store.go
Vars-07: done

app/user.go
agnivade: It is a standard code review syntax, which essentially means replace "e" by "err".

Vars-07: okay, got it but err is already been initialized & it references *model.AppError

agnivade: Right, in that case, we use `nErr` which stands for normalized error. We have an ongoing campaign which will convert all store level errors to `error`. So all of this will become `err` at a near point in future.

store/storetest/channel_store.go
```diff
@@ -8202,3 +8203,76 @@ func testDeleteSidebarCategory(t *testing.T, ss store.Store, s SqlSupplier) {
...
+		ss.Channel().SaveDirectChannel(&o1, &m1, &m2)
+		ss.Channel().SaveDirectChannel(&o2, &m3, &m4)
+
+		directChannelsUser1, nErr := ss.Channel().GetDirectChannelsForUser(u1.Id)
+		require.Nil(t, nErr)
+		require.Len(t, directChannelsUser1, 2)
+
+		directChannelsUser2, nErr := ss.Channel().GetDirectChannelsForUser(u2.Id)
+		require.Nil(t, nErr)
+		require.Len(t, directChannelsUser2, 1)
```
ashishbhate, line 8276: We should add some tests to confirm that the returned channel list has the correct display name (i.e. the other user's username), since the `GetDirectChannelsForUser` specifically does the extra inner join to add this information.

app/user.go
```diff
@@ -980,7 +980,8 @@ func (a *App) invalidateUserChannelMembersCaches(userId string) *model.AppError
 }
 
 func (a *App) UpdateActive(user *model.User, active bool) (*model.User, *model.AppError) {
-	user.UpdateAt = model.GetMillis()
+	updateTime := model.GetMillis()
+	user.UpdateAt = updateTime
```
ashishbhate, line 984: After the previous changes, it looks declaring a separate `updateTime` is no longer required.

store/storetest/channel_store.go
```diff
@@ -8256,3 +8257,79 @@ func testDeleteSidebarCategory(t *testing.T, ss store.Store, s SqlSupplier) {
...
+		m4.ChannelId = o2.Id
+		m4.UserId = u3.Id
+		m4.NotifyProps = model.GetDefaultChannelNotifyProps()
+
+		ss.Channel().SaveDirectChannel(&o1, &m1, &m2)
+		ss.Channel().SaveDirectChannel(&o2, &m3, &m4)
+
+		directChannelsUser1, nErr := ss.Channel().GetDirectChannelsForUser(u1.Id)
+		assert.Equal(t, directChannelsUser1[0].DisplayName, u3.Username)
+		assert.Equal(t, directChannelsUser1[1].DisplayName, u2.Username)
```
ashishbhate, line 8326: The query doesn't do any ordering, so this could randomly fail since we are assuming ordering. I suggest using `require.Contains`

store/storetest/channel_store.go
```diff
@@ -8256,3 +8257,79 @@ func testDeleteSidebarCategory(t *testing.T, ss store.Store, s SqlSupplier) {
...
+		require.Nil(t, nErr)
+		require.Len(t, directChannelsUser1, 2)
```
ashishbhate, line 8328: Lets move the require statements before the asserts so that we won't have a panic if the tests fail.

store/storetest/channel_store.go
```diff
@@ -8256,3 +8257,79 @@ func testDeleteSidebarCategory(t *testing.T, ss store.Store, s SqlSupplier) {
...
+
+		directChannelsUser2, nErr := ss.Channel().GetDirectChannelsForUser(u2.Id)
+		assert.Equal(t, directChannelsUser1[0].DisplayName, u3.Username)
+		require.Nil(t, nErr)
+		require.Len(t, directChannelsUser2, 1)
```
ashishbhate, line 8333: Same as above.

store/storetest/channel_store.go
```diff
@@ -8256,3 +8257,89 @@ func testDeleteSidebarCategory(t *testing.T, ss store.Store, s SqlSupplier) {
...
+		ss.Channel().SaveDirectChannel(&o2, &m3, &m4)
+
+		directChannelsUser1, nErr := ss.Channel().GetDirectChannelsForUser(u1.Id)
+		require.Nil(t, nErr)
+		require.Len(t, directChannelsUser1, 2)
+
+		var user1DireactChannelNames []string
+		for _, item := range directChannelsUser1 {
+			user1DireactChannelNames = append(user1DireactChannelNames, item.DisplayName)
+		}
```
ashishbhate, line 8331: Nit:

Maybe do the require.Contains in the loop:

```
require.Contains(t, []string{u3.Username, u2.Username}, item)
```

I think its more concise this way.

store/storetest/channel_store.go
```diff
@@ -8256,3 +8257,89 @@ func testDeleteSidebarCategory(t *testing.T, ss store.Store, s SqlSupplier) {
...
+		require.Contains(t, user1DireactChannelNames, u2.Username)
+
+		directChannelsUser2, nErr := ss.Channel().GetDirectChannelsForUser(u2.Id)
+		require.Nil(t, nErr)
+		require.Len(t, directChannelsUser2, 1)
+
+		var user2DireactChannelNames []string
+		for _, item := range directChannelsUser2 {
+			user2DireactChannelNames = append(user2DireactChannelNames, item.DisplayName)
+		}
```
ashishbhate, line 8342: Since we check the len is 1, we dont need a loop

store/storetest/channel_store.go
```diff
@@ -8256,3 +8257,89 @@ func testDeleteSidebarCategory(t *testing.T, ss store.Store, s SqlSupplier) {
...
+	o1.TeamId = teamID
+	o1.DisplayName = "Different Name" + model.NewId()
+	o1.Name = "zz" + model.NewId() + "b"
+	o1.Type = model.CHANNEL_DIRECT
+
+	o2 := model.Channel{}
+	o2.TeamId = teamID
+	o2.DisplayName = "Different Name 1" + model.NewId()
+	o2.Name = "zz1" + model.NewId() + "c"
+	o2.Type = model.CHANNEL_DIRECT
```
ashishbhate, line 8274: non-blocking nit:

How about `c1` and `c2` instead of `o1` and `o2`.

store/sqlstore/channel_store.go
```diff
@@ -3496,6 +3496,38 @@ func (s SqlChannelStore) UserBelongsToChannels(userId string, channelIds []strin
...
+					IU.Id != :UserId
+				) AS OtherUsers ON OtherUsers.ChannelId = C.Id
+			WHERE
+			    C.Type = 'D'
+				AND CM.UserId = :UserId`
+
+	var channels model.ChannelList
+
+	if _, err := s.GetReplica().Select(&channels, query, map[string]interface{}{"UserId": userId}); err != nil {
+		return nil, model.NewAppError("SqlChannelStore.GetDirectChannelsForUser", "store.sql_channel.direct_channels_for_user.app_error", nil, err.Error(), http.StatusInternalServerError)
```
ashishbhate, line 3525: Since this adds a new error id, you'll need to run `make i18n-extract` and add a translation in `i18n/en.json`

ashishbhate: Looks like we're moving away from AppError, instead directly using the `error` type.

I'll defer to @agnivade  on this.

Vars-07: yes, @agnivade in his previous comments mentioned switching to `error`

agnivade, line 3525: This should just return a normal error. See the other store methods for examples.

store/storetest/channel_store.go
```diff
@@ -8415,3 +8416,82 @@ func testDeleteSidebarCategory(t *testing.T, ss store.Store, s SqlSupplier) {
...
+	c2.Name = "zz1" + model.NewId() + "c"
+	c2.Type = model.CHANNEL_DIRECT
+
+	t.Run("direct messages with created user & others", func(t *testing.T) {
+		u1 := &model.User{}
+		u1.Email = MakeEmail()
+		u1.Nickname = model.NewId()
+		_, err := ss.User().Save(u1)
+		require.Nil(t, err)
+		_, err = ss.Team().SaveMember(&model.TeamMember{TeamId: model.NewId(), UserId: u1.Id}, -1)
```
agnivade, line 8441: Saving team member is not really necessary. We can remove this.

store/storetest/channel_store.go
```diff
@@ -8415,3 +8416,82 @@ func testDeleteSidebarCategory(t *testing.T, ss store.Store, s SqlSupplier) {
...
+
+		directChannelsUser1, nErr := ss.Channel().GetDirectChannelsForUser(u1.Id)
+		require.Nil(t, nErr)
+		require.Len(t, directChannelsUser1, 2)
+
+		for _, item := range directChannelsUser1 {
+			require.Contains(t, []string{u3.Username, u2.Username}, item.DisplayName)
+		}
+
+		directChannelsUser2, nErr := ss.Channel().GetDirectChannelsForUser(u2.Id)
```
agnivade, line 8491: Let's also add the same case for `u3.Id`.

store/storetest/channel_store.go
```diff
@@ -8415,3 +8416,82 @@ func testDeleteSidebarCategory(t *testing.T, ss store.Store, s SqlSupplier) {
...
+	c1.Name = "zz" + model.NewId() + "b"
+	c1.Type = model.CHANNEL_DIRECT
+
+	c2 := model.Channel{}
+	c2.TeamId = teamID
+	c2.DisplayName = "Different Name 1" + model.NewId()
+	c2.Name = "zz1" + model.NewId() + "c"
+	c2.Type = model.CHANNEL_DIRECT
+
+	t.Run("direct messages with created user & others", func(t *testing.T) {
```
agnivade, line 8435: Not sure if a single sub-test makes sense here. If you want, we can split the cases into u1 and then u2 and u3 together. Otherwise, we can just have a single test.

app/user_test.go
```diff
@@ -234,6 +234,7 @@ func TestUpdateActiveBotsSideEffect(t *testing.T) {
 	})
 
 	th.App.UpdateActive(th.BasicUser, false)
+	require.NotZero(t, th.BasicUser.DeleteAt)
```
agnivade, line 237: This just tests the user's deleteAt field, but does not test that the channel is deleted too. I would prefer a separate test where we create a user, and then create 2 DM channels with new users and then delete the user and verify that those DM channels have been marked as deleted.

app/user_test.go
```diff
@@ -261,6 +263,23 @@ func TestUpdateActiveBotsSideEffect(t *testing.T) {
 	th.App.UpdateActive(th.BasicUser, true)
 }
 
+func TestUserDeleteUpdateDirectChannelDeleted(t *testing.T) {
+	th := Setup(t).InitBasic()
+	defer th.TearDown()
+
+	th.CreateDmChannel(th.BasicUser2)
+	basicUser3 := th.CreateUser()
```
agnivade, line 271: We can just call it "user"

app/user_test.go
```diff
@@ -261,6 +263,23 @@ func TestUpdateActiveBotsSideEffect(t *testing.T) {
...
+	th.CreateDmChannel(basicUser3)
+
+	th.App.UpdateActive(th.BasicUser, false)
+	require.NotZero(t, th.BasicUser.DeleteAt)
+
+	directChannels, _ := th.App.Srv().Store.Channel().GetDirectChannelsForUser(th.BasicUser.Id)
```
agnivade, line 277: We should not ignore the error.

store/sqlstore/channel_store.go
```diff
@@ -3496,6 +3496,38 @@ func (s SqlChannelStore) UserBelongsToChannels(userId string, channelIds []strin
...
+					IU.Id != :UserId
+				) AS OtherUsers ON OtherUsers.ChannelId = C.Id
+			WHERE
+			    C.Type = 'D'
+				AND CM.UserId = :UserId`
+
+	var channels model.ChannelList
+
+	if _, err := s.GetReplica().Select(&channels, query, map[string]interface{}{"UserId": userId}); err != nil {
+		return nil, errors.Wrap(err, "GetDirectChannelsForUser: Failed to retrieve direct channels for the given user")
```
agnivade, line 3525: s/GetDirectChannelsForUser: F/f

Vars-07: Sorry didn't get, Failed --> failed?

agnivade: Yes. Error strings should not start with a capital letter. https://github.com/golang/go/wiki/CodeReviewComments#error-strings

agnivade:
@jgilliam17 - Assigning to you since you filed the original issue and have a general context. The overall behavior is tested with unit tests.

jgilliam17:
Thank you @Vars-07 

I am seeing a deactivated DM removed from the LHS as another channel/DM is opened, without having to close deactivated DM. Deactivated DMs should remain on the LHS even when not open on the main channel view.
Let me know if this change is related to this fix. If not, I can file separately.
If you look at the images, on the old sidebar deactivated DM is present when open, on exp. sidebar DM is not shown on the LHS. When switching away to a different channel, on both old and exp sidebar, deactivated DM should still be present on the LHS.
| Old sidebar   | Exp. Sidebar |
| ------------- | ------------- |
| Screen Shot 2020-08-05 at 10 39 15 AM|Screen Shot 2020-08-05 at 10 39 36 AM|

Vars-07:
Hi, @jgilliam17, yes that change of vanishing away archived/deactivated channel is not related to this fix. IMO that should be UI issue. you can file a separate Jira for that. Thanks

hmhealey:
@Vars-07 Can you explain again what problem you were having with DMs not being deleted when one of their members is deactivated? I'm not sure if marking the DM as deleted is the correct way to fix that since deleting DMs might have unintended side effects such as the one @jgilliam17 reported. In the past, we've assumed that DMs and GMs are only hidden, never archived, so I'd be worried about changing that now

jasonblais:
Hi @Vars-07, let us know if you had thoughts or questions on Harrison's latest message. Would be happy to clarify :) 

From @hmhealey:
> @Vars-07 Can you explain again what problem you were having with DMs not being deleted when one of their members is deactivated? I'm not sure if marking the DM as deleted is the correct way to fix that since deleting DMs might have unintended side effects such as the one @jgilliam17 reported. In the past, we've assumed that DMs and GMs are only hidden, never archived, so I'd be worried about changing that now

stylianosrigas:
Removing the test server for maintenance purposes. Please add the label again if you need it. Sorry for any inconvenience.

hanzei:
@Vars-07 Do you have any questions [about the feedback](https://github.com/mattermost/mattermost-server/pull/15039#issuecomment-679315545)? 
Please note that we will close this PR after another 10 days of inactivity. If you don't object, other contributors might complete this PR for you.

Vars-07:
Hey,
Sorry for the delay, was little occupied will pick it up asap.

Thanks

PR #15040, opened 2020-07-15
MM-26883/26884: properly identify interface name for test services

agnivade:
We were just hardcoding to dockerhost which would only work inside
a CI environment. This PR generalizes the logic to use fallbacks
if dockerhost is not available.

### Ticket link
https://mattermost.atlassian.net/browse/MM-26883
https://mattermost.atlassian.net/browse/MM-26884

### DIFF
--- a/utils/testutils/testutils.go
+++ b/utils/testutils/testutils.go
@@ -12,6 +12,7 @@ import (
 	"path/filepath"
 	"runtime"
 	"strconv"
+	"time"
 
 	"github.com/mattermost/mattermost-server/v5/utils/fileutils"
 )
@@ -36,7 +37,7 @@ func ReadTestFile(name string) ([]byte, error) {
 // This is helpful when a test is being run in a CI environment under docker.
 func GetInterface(port int) string {
 	dial := func(iface string, port int) bool {
-		c, err := net.Dial("tcp", iface+":"+strconv.Itoa(port))
+		c, err := net.DialTimeout("tcp", iface+":"+strconv.Itoa(port), time.Second)
 		if err != nil {
 			return false
 		}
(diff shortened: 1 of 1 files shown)


utils/testutils/testutils.go
```diff
@@ -27,3 +31,37 @@ func ReadTestFile(name string) ([]byte, error) {
...
+	}
+	// If not, we check localhost
+	iface = "localhost"
+	if ok := dial(iface, port); ok {
+		return iface
+	}
+	// If nothing works, we just attempt to use a hack and get the interface IP.
+	// https://stackoverflow.com/a/37212665/4962526.
+	if runtime.GOOS != "windows" {
+		cmd := exec.Command("bash", "-c", `ifconfig | grep -E "([0-9]{1,3}\.){3}[0-9]{1,3}" | grep -v 127.0.0.1 | awk '{ print $2 }' | cut -f2 -d: | head -n1`)
```
ethervoid, line 59: `ifconfig` could not be available. I tend to use `ip address` at least in Linux (Ubuntu)

utils/testutils/testutils.go
```diff
@@ -27,3 +31,37 @@ func ReadTestFile(name string) ([]byte, error) {
...
+
+// GetInterface returns the best match of an interface that might be listening on a given port.
+// This is helpful when a test is being run in a CI environment under docker.
+func GetInterface(port int) string {
+	dial := func(iface string, port int) bool {
+		c, err := net.Dial("tcp", iface+":"+strconv.Itoa(port))
+		if err != nil {
+			return false
+		}
+		c.Close()
```
ethervoid, line 43: One question, what about using `defer c.Close()`?

utils/testutils/testutils.go
```diff
@@ -27,3 +31,37 @@ func ReadTestFile(name string) ([]byte, error) {
...
+		return true
+	}
+	// First, we check dockerhost
+	iface := "dockerhost"
```
ethervoid, line 47: Constant?

utils/testutils/testutils.go
```diff
@@ -27,3 +31,37 @@ func ReadTestFile(name string) ([]byte, error) {
...
+	if ok := dial(iface, port); ok {
+		return iface
+	}
+	// If not, we check localhost
+	iface = "localhost"
```
ethervoid, line 52: Constant?

utils/testutils/testutils.go
```diff
@@ -27,3 +31,37 @@ func ReadTestFile(name string) ([]byte, error) {
 		return data.Bytes(), nil
 	}
 }
+
+// GetInterface returns the best match of an interface that might be listening on a given port.
+// This is helpful when a test is being run in a CI environment under docker.
+func GetInterface(port int) string {
+	dial := func(iface string, port int) bool {
+		c, err := net.Dial("tcp", iface+":"+strconv.Itoa(port))
```
ethervoid, line 39: What about using `net.DialTimeout` to avoid locking for closed/unreachable destinations

utils/testutils/testutils.go
agnivade: See the SO link. `ip` is not there on OSX. Happy to change it to something else though.

utils/testutils/testutils.go
agnivade: Because there's nothing after this. So `defer` doesn't really make sense IMO.

utils/testutils/testutils.go
agnivade: I am actually changing the variable later. But 0/5 on it frankly. Let me know.

utils/testutils/testutils.go
ethervoid: Take into account that `ifconfig` is deprecated. So maybe is good to make one for Linux and another for darwin

utils/testutils/testutils.go
ethervoid: Oh didn't see it was the same variable. Nevermind then

utils/testutils/testutils.go
ethervoid: nvm

agnivade:
@lindy65 - Nothing to test here as this just adds a function that gets called for some tests to pass. There are no external changes.

agnivade:
@amyblais - This was part of an EE PR https://github.com/mattermost/enterprise/pull/699 too. Should that also be backported?

amyblais:
@agnivade It can be cherry-picked too.

PR #15042, opened 2020-07-16
WebhookStore migration

rvillablanca:
#### Summary

Migration of `WebhookStore` to return plain errors.

Fixes https://github.com/mattermost/mattermost-server/issues/15043


rvillablanca:
I added a new http response code (404) when an outgoing webhook was not found and modified those test cases to check for 404 status code. Could this have any impact ? Before this it was returning only 500 so I would expect any client should be waiting for any 20x for a success case and any other http status should be handled as an error. 

Any concern about this just tell me

agnivade:
I don't mind. But we should document this in the release docs to be explicit.

app/webhook.go
```diff
@@ -514,18 +574,28 @@ func (a *App) GetOutgoingWebhooksPage(page, perPage int) ([]*model.OutgoingWebho
 
 func (a *App) GetOutgoingWebhooksPageByUser(userId string, page, perPage int) ([]*model.OutgoingWebhook, *model.AppError) {
 	if !*a.Config().ServiceSettings.EnableOutgoingWebhooks {
-		return nil, model.NewAppError("GetOutgoingWebhooksPage", "api.outgoing_webhook.disabled.app_error", nil, "", http.StatusNotImplemented)
+		return nil, model.NewAppError("GetOutgoingWebhooksPageByUser", "api.outgoing_webhook.disabled.app_error", nil, "", http.StatusNotImplemented)
```
agnivade, line 577: Nice

PR #15060, opened 2020-07-18
MM-27041: Do not run tests against both DBs in CI

agnivade:
We already run the entire test suite for both mysql and postgres in parallel in CI.
So we just run the tests for the current database set.

#### Ticket Link

https://mattermost.atlassian.net/browse/MM-27041

### DIFF
--- a/store/sqlstore/store_test.go
+++ b/store/sqlstore/store_test.go
@@ -4,6 +4,7 @@
 package sqlstore
 
 import (
+	"os"
 	"sync"
 	"testing"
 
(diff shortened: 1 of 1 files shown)


store/sqlstore/store_test.go
```diff
@@ -80,14 +81,31 @@ func initStores() {
...
-		SqlSettings: storetest.MakeSqlSettings(model.DATABASE_DRIVER_MYSQL),
-	})
-	storeTypes = append(storeTypes, &storeType{
-		Name:        "PostgreSQL",
-		SqlSettings: storetest.MakeSqlSettings(model.DATABASE_DRIVER_POSTGRES),
-	})
+	// In CI, we already run the entire test suite for both mysql and postgres in parallel.
+	// So we just run the tests for the current database set.
+	if os.Getenv("IS_CI") == "true" {
+		switch os.Getenv("IS_CI") {
```
jespino, line 87: I think this line is wrong, if `IS_CI` is `true` it will never be `mysql` or  `postgres` because is `true`. I guess you put the wrong env var there by mistake.

agnivade: :facepalm:

store/sqlstore/store_test.go
```diff
@@ -80,14 +81,31 @@ func initStores() {
...
-		SqlSettings: storetest.MakeSqlSettings(model.DATABASE_DRIVER_MYSQL),
-	})
-	storeTypes = append(storeTypes, &storeType{
-		Name:        "PostgreSQL",
-		SqlSettings: storetest.MakeSqlSettings(model.DATABASE_DRIVER_POSTGRES),
-	})
+	// In CI, we already run the entire test suite for both mysql and postgres in parallel.
+	// So we just run the tests for the current database set.
+	if os.Getenv("IS_CI") == "true" {
+		switch os.Getenv("MM_SQLSETTINGS_DRIVERNAME") {
```
jespino, line 87: This way we still are testing both, because we are going to pass through MySQL and Postgres for the whole applicatoin, right?

agnivade: Correct.

agnivade:
@lindy65 - No need to test this one as it just involves tests.

PR #15061, opened 2020-07-18
[MM-26023] - Return an array of category Ids when the categories are updated

nevyangelova:
#### Summary
Return an array of category Ids when the categories are updated. We need the category Ids as a part of the new favouriting/category logic in the new sidebar. I haven't touched golang before so if there are syntax errors or bad practices feel free to advice me on this. 

#### Ticket Link

https://mattermost.atlassian.net/browse/MM-26023


app/channel.go
```diff
@@ -2700,7 +2700,14 @@ func (a *App) UpdateSidebarCategories(userId, teamId string, categories []*model
 		return nil, err
 	}
 	message := model.NewWebSocketEvent(model.WEBSOCKET_EVENT_SIDEBAR_CATEGORY_UPDATED, teamId, "", userId, nil)
+
+	var categoryIds = []string
+	for _, category := range categories {
+		categoryIds = append(categoryIds, category.id)
+	}
```
fmunshi, line 2707: Since you are copying an existing slice its better to define the length of the new slice when you make it instead of making a new slice every time you append an element to it

something like 
```suggestion
	categoryIds := make([]string, len(categories))
	for i, category := range categories {
		categoryIds[i] = category.Id
	}
```

hmhealey:
Closing this since unfortunately the provided categoryIds don't fully reflect every category that might have changed since it's possible for others to get changed as a side effect of this (such as channels being removed from another category because our request moves them into one). Because of those cases, we need to re-request all categories on the team anyway, so this is unnecessary.

PR #15063, opened 2020-07-19
MM-27149: optimize initBasic

agnivade:
Mostly, all tests just needed the user initialization part and not
the channel and group creation. So we move the user initialization inside
the Setup call. This avoids unnecessary DB calls which take around 250-300ms
on average.

And we make the login requests concurrently to shave off a few more ms.
According to my tests, the 2 login calls take 140 ms on average, which
shaves off 70ms.

So approximately, we shave off 350ms per test.  And there are 114 occurences of these.
Thereby, approximately a total of 39 seconds.

#### Ticket Link

https://mattermost.atlassian.net/browse/MM-27149


agnivade:
@lindy65 - This just improves the test duration. So no need to test this one.

PR #15065, opened 2020-07-19
[MM-26969] Add team filters to search teams

fmunshi:
#### Summary
- Adds filter options to the search teams endpoint 

#### Ticket Link
- https://mattermost.atlassian.net/browse/MM-26969

#### Related PRs:
https://github.com/mattermost/mattermost-api-reference/pull/555
https://github.com/mattermost/mattermost-redux/pull/1200


store/sqlstore/team_store.go
```diff
@@ -372,45 +372,124 @@ func (s SqlTeamStore) GetByNames(names []string) ([]*model.Team, *model.AppError
...
+			groupConstrainedFilter = sq.Or{
+				sq.NotEq{"GroupConstrained": true},
+				sq.Eq{"GroupConstrained": nil},
+			}
+		}
+
+		if teamFilters == nil {
+			teamFilters = groupConstrainedFilter
+		} else {
+			teamFilters = sq.And{teamFilters, groupConstrainedFilter}
```
hahmadia, line 437: Should this be a `OR`?

store/sqlstore/team_store.go
```diff
@@ -372,45 +372,124 @@ func (s SqlTeamStore) GetByNames(names []string) ([]*model.Team, *model.AppError
...
 
-	term = sanitizeSearchTerm(term, "\\")
-	term = wildcardSearchTerm(term)
-
-	operatorKeyword := "ILIKE"
-	if s.DriverName() == model.DATABASE_DRIVER_MYSQL {
-		operatorKeyword = "LIKE"
+	queryString, args, err := s.teamSearchQuery(term, opts, false).ToSql()
+	fmt.Println(queryString)
+	fmt.Println(args)
```
hahmadia, line 461: Do we need these?

store/sqlstore/team_store.go
```diff
@@ -372,45 +372,124 @@ func (s SqlTeamStore) GetByNames(names []string) ([]*model.Team, *model.AppError
...
-	offset := page * perPage
 
-	term = sanitizeSearchTerm(term, "\\")
-	term = wildcardSearchTerm(term)
-	operatorKeyword := "ILIKE"
-	if s.DriverName() == model.DATABASE_DRIVER_MYSQL {
-		operatorKeyword = "LIKE"
+	queryString, args, err := s.teamSearchQuery(term, opts, false).ToSql()
+	fmt.Println(queryString)
+	fmt.Println(args)
```
hahmadia, line 480: Do we need these?

fmunshi: nope

store/sqlstore/team_store.go
fmunshi: No this should be `and` - despite the fact that the front end use case for this does an `OR` (by inverting the filters submitted) since in general if a user requests an API asking for teams filtered by `AllowOpenInvite` AND `GroupConstrained` they would expect the API to return only results matching both of those filters right? If they want to get `GroupConstrained` OR `AllowOpenInvite` then they would have to send `IncludeGroupConstrained` AND `AllowOpenInvite` parameters...

I struggled with what the best way to do this would be and eventually settled on this but now im reconsidering...

model/team_search.go
```diff
@@ -9,9 +9,12 @@ import (
...
 type TeamSearch struct {
-	Term    string `json:"term"`
-	Page    *int   `json:"page,omitempty"`
-	PerPage *int   `json:"per_page,omitempty"`
+	Term                    string `json:"term"`
+	Page                    *int   `json:"page,omitempty"`
+	PerPage                 *int   `json:"per_page,omitempty"`
+	AllowOpenInvite         *bool  `json:"allow_open_invite,omitempty"`
+	GroupConstrained        *bool  `json:"group_constrained,omitempty"`
+	IncludeGroupConstrained *bool  `json:"include_group_constrained,omitempty"`
```
mkraft, line 17: Will the fact that `IncludeGroupConstrained` has a default value of `false` cause calls to `(*TeamStore).SearchAll` and `(*TeamStore).SearchAllPaged` to exclude groups that used to be included?

fmunshi: No the way i've set this up is that it `IncludeGroupConstrained` is only applied as a true value and when it is not set (nil) the query remains unchanged. If someone wants to exclude group constrained groups all they have to do is `GroupConstrained: false`

fmunshi: Actually now that I re-think it it probably makes more sense to make this `ExcludeGroupConstrained`... which by default returns groups and if you set it to true it excludes group constrained results... that way the true and false value do exactly whats expected

store/sqlstore/team_store.go
```diff
@@ -372,45 +372,118 @@ func (s SqlTeamStore) GetByNames(names []string) ([]*model.Team, *model.AppError
...
+				sq.Eq{"GroupConstrained": nil},
+			}
+		}
+
+		if teamFilters == nil {
+			teamFilters = groupConstrainedFilter
+		} else {
+			teamFilters = sq.Or{teamFilters, groupConstrainedFilter}
+		}
+	}
```
fmunshi, line 445: @mkraft @hahmadia I was able to resolve this in a much cleaner way now by excluding group constrained teams when `AllowOpenInvite = false` essentially treating `Private` teams as a completely different entity than `GroupConstrained` teams. If the user wants to get all `Private` teams including `GroupConstrained` then they have to simply set `AllowOpenInvite = false` and `GroupConstrained = true`

srkgupta:
Sorry approved this PR by mistake. Reopening the PR request again

PR #15075, opened 2020-07-20
MM-26757 Run server tests on non-default port

wiggin77:
#### Summary
This PR modifies server run unit tests to launch the server on a non-default port. This avoids tests failing if a server instance is already bound to the default port.

#### Ticket Link

https://mattermost.atlassian.net/browse/MM-26889


cmd/mattermost/commands/server_test.go
```diff
@@ -15,6 +15,10 @@ import (
 	"github.com/stretchr/testify/require"
 )
 
+const (
+	UnitTestListeningPort = ":8066"
```
lieut-data, line 19: Would we consider a random available port instead? I have bee known to use `:8066` to run a parallel server instance :)

wiggin77: Fair point.  Detecting available port would be a bit more involved and probably not worth it, so I switched to a port very unlikely to be used.  Thanks.

agnivade: My apologies for an unsolicited comment. But if we set port to 0, kernel will assign a random port by itself, and then we can get query the network address to get the assigned port number if we want it in future.

wiggin77:
@agnivade did you have a comment?

lieut-data:
I think it got buried in a resolved thread: https://github.com/mattermost/mattermost-server/pull/15075#discussion_r457606713.

agnivade:
@wiggin77 https://github.com/mattermost/mattermost-server/pull/15075#discussion_r457606713

cmd/mattermost/commands/server_test.go
wiggin77: Done

PR #15076, opened 2020-07-21
MM-22784 Advanced logging config for audit

wiggin77:
#### Summary
This PR adds the advanced logging config for audit.  Existing support for auditing to a single file remains for E0 and E10 licenses instances, and a new config item `ExperimentalAuditSettings.AdvancedLoggingConfig` is added that behaves like `LogSettings.AdvancedLoggingConfig`.

Supported destinations:

- file
- syslog (with out without TLS)
- raw TCP socket (with out without TLS)

`ExperimentalAuditSettings.AdvancedLoggingConfig` can contain a filespec to a config file, a database DSN, or JSON.

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-22784


web/context.go
```diff
@@ -26,19 +26,19 @@ type Context struct {
 
 // LogAuditRec logs an audit record using default RestLevel.
```
streamer45, line 27:
```suggestion
// LogAuditRec logs an audit record using default LevelAPI.
```

app/audit.go
```diff
@@ -59,11 +61,11 @@ func (a *App) GetAuditsPage(userId string, page int, perPage int) (model.Audits,
 
 // LogAuditRec logs an audit record using default CLILevel.
```
streamer45, line 62:
```suggestion
// LogAuditRec logs an audit record using default LvlAuditCLI.
```

app/audit.go
```diff
@@ -156,21 +125,50 @@ func (s *Server) configureAudit(adt *audit.Audit) {
...
 			mlog.Debug("File audit target created successfully", mlog.String("filename", opts.Filename))
 			adt.AddTarget(target)
 		}
 	}
+
+	// Advanced logging for audit requires license.
+	if !bAllowAdvancedLogging || *s.Config().ExperimentalAuditSettings.AdvancedLoggingConfig == "" {
+		return errs
+	}
+	dsn := *s.Config().ExperimentalAuditSettings.AdvancedLoggingConfig
```
streamer45, line 144: Nit: If we move this up we can read the value just once.

mlog/logr.go
```diff
@@ -39,30 +39,32 @@ type LogTargetCfg map[string]*LogTarget
...
 	lgr.OnExit = func(int) {}
 	lgr.OnPanic = func(interface{}) {}
-
 	lgr.OnLoggerError = onLoggerError
 	lgr.OnQueueFull = onQueueFull
 	lgr.OnTargetQueueFull = onTargetQueueFull
 
+	err := logrAddTargets(lgr, targets)
+	logger := lgr.NewLogger()
+	return &logger, err
```
streamer45, line 51: Are we returning a logger also in case of error cause we accept the fact that adding some targets might fail?

web/context.go
```diff
@@ -26,19 +26,19 @@ type Context struct {
...
 func (c *Context) LogAuditRec(rec *audit.Record) {
-	c.LogAuditRecWithLevel(rec, app.RestLevel)
+	c.LogAuditRecWithLevel(rec, app.LevelAPI)
 }
 
 // LogAuditRec logs an audit record using specified Level.
```
streamer45, line 32:
```suggestion
// LogAuditRecWithLevel logs an audit record using specified Level.
```

Should we also mention that the level gets overridden in case of a permission error?

mlog/logr.go
```diff
@@ -39,30 +39,32 @@ type LogTargetCfg map[string]*LogTarget
...
 		}
 		lgr.AddTarget(target)
 	}
-	logger := lgr.NewLogger()
-	return &logger, errs
+	return errs
 }
 
-func newLogrTarget(name string, t *LogTarget) (logr.Target, error) {
+func NewLogrTarget(name string, t *LogTarget) (logr.Target, error) {
```
streamer45, line 67: Since we export this I'd probably add a godoc line explaining what it does.

mlog/logr.go
wiggin77: Right.  An individual target failing is not fatal.  We want to log that failure on the targets that were successful.  For example, if we cannot connect to the syslog server we want to log that error to file.

app/audit.go
wiggin77: Fixed via https://github.com/mattermost/mattermost-server/pull/15076/commits/3339d015f5a56ff38398bac6188d4d396243ebb9

mlog/logr.go
wiggin77: Fixed via https://github.com/mattermost/mattermost-server/pull/15076/commits/04e503a93b678005c9c71ebd5ad897bede654ce1

web/context.go
wiggin77: Fixed via https://github.com/mattermost/mattermost-server/pull/15076/commits/04e503a93b678005c9c71ebd5ad897bede654ce1

mlog/logr.go
wiggin77: BTW, the error returned may be a multi-error, one for each target that failed to load.

amyblais:
@wiggin77 Does this add a new config setting and does it need to be added to telemetry?

wiggin77:
> @wiggin77 Does this add a new config setting and does it need to be added to telemetry?

@amyblais it is already added to telemetry. So far it just reports whether advanced audit is enabled or not.

PR #15077, opened 2020-07-21
MM-27150: Remove TE tests as part of tbs workflow

agnivade:
There's no need to run TE all over again in tbs, when it's already run
in the normal workflow.


agnivade:
Amazing, now it only takes 9 minutes instead of 39 minutes for tbs !

agnivade:
@lindy65 - This is just a change in the CI pipeline. No testing needed.

PR #15078, opened 2020-07-21
Add psd preview support

jespino:
#### Summary
Add psd/psb preview support

#### Ticket Link
[MM-27153](https://mattermost.atlassian.net/browse/MM-27153)


jespino:
@esethna I added this as a personal initiative, would be great to confirm if it is ok to include this in the product from the PM POV.

esethna:
From a PM perspective I'm fine adding this. I'll add @deanwhillier to comment from a dev perspective regarding maintainability as that would be my concern with this

deanwhillier:
Thanks @esethna! This looks like an entirely server-side feature so likely needs @grundleborg to answer that question. I presume that at the end of the day, this just delivers a jpg/png thumbnail/image representing the PSD file to the client so shouldn't have much baring? Correct me if I'm wrong @jespino!

jespino:
Yes, exactly @esethna and @deanwhillier, this is entirely handled by a third party library in the server, so the maintenance cost should be low.

jespino:
@grundleborg it is a dependency of the psd library, I haven't investigated it.

ashishbhate:
The psd library includes a javascript build target that requires the gopherjs library.

srkgupta:
Hi @jespino 
I tested for the PSD file previews and they are working absolutely fine. However the ticket also says the support for PSB previews. However I tried with a sample PSB file format and it fails to display the preview for the same. 

Downloaded the PSB file from here:
http://www.fileformatcommons.com/psb-file-format/

Preview Error:
Screenshot 2020-07-28 at 6 25 39 PM

If the PSB file format is not supported yet, can you please update the JIRA ticket.

jespino:
In theory the library supports PSB, but we can remove the PSB support from the list if is not working fine.

srkgupta:
Thanks @jespino. I have updated the JIRA ticket to indicate support for PSD files. However with this PR, since the PSB files are not supported, it results in error and does not even provides option to download the file. Can we fix this and handle the traditional way to download a file if preview generation fails? 

This is how it shows with this PR:
Screenshot 2020-07-28 at 6 25 39 PM

This is how it was working prior to this PR:
Screenshot 2020-07-28 at 6 36 44 PM

jespino:
@srkgupta Ok, then I going to change a bit the code to only try to handle the PSDs.

cpanato:
test server removed due some underlying issues in the cluster that made this one unstable.

jespino:
@srkgupta removed the psb support, can you test it again?

justinegeffen:
@esethna, from what I can see there are no doc changes required for this. If you could confirm that'd be awesome. :)

PR #15079, opened 2020-07-21
Upgrading to the last mattermost-utilities/mmgotool

jespino:
#### Summary
Upgrading to the last mattermost-utilities/mmgotool


go.tools.mod
```diff
@@ -4,7 +4,8 @@ go 1.14
 
 require (
 	github.com/jstemmer/go-junit-report v0.9.1 // indirect
-	github.com/mattermost/mattermost-utilities v0.0.0-20200521071950-1be063e1b2fe // indirect
+	github.com/mattermost/mattermost-utilities v0.0.0-20200720080147-779d19605c57 // indirect
```
agnivade, line 7: This line is not needed.

jespino:
@agnivade thanks, I think I would expect "no params to work properly" using the mattermost dir and enterprise dir as expected, and on the other hand, I would expect to not need to pass the `--portal-dir=""` I going to fix that in the original repo and make it work here.

agnivade:
Right, that's also fine. The issue is that `--portal-dir` takes a value by default. So not supplying anything confuses it. We need to check for the case where nothing is specified, then ignore portal-dir.

jespino:
@agnivade I going to keep the `--portal-dir=""` approach and thing how to properly solve this problem later. And now I going to close this PR and open it again in a branch in upstream, because if not i can't modify the Makefile.

PR #15082, opened 2020-07-21
Replacing blacklist with blocklist

jespino:
#### Summary
Replacing blacklist with blocklist


lindalumitchell:
Side question: Are we also intending to search through other repos to make this change everywhere? I see some here for example, but it's unclear to me how making changes might affect functionality: https://github.com/mattermost/mattermost-webapp/search?q=blacklist&unscoped_q=blacklist @jespino @jasonblais

jasonblais:
Thanks @lindalumitchell! There is in fact a PR open for the webapp repo https://github.com/mattermost/mattermost-webapp/pull/5952

lindalumitchell:
Of course, thanks @jasonblais!

PR #15087, opened 2020-07-21
MM-27007 Remove automatic sidebar migration

hmhealey:
Unfortunately we found rather late that doing an automatic migration of everyone's sidebars will cause problems on the community cluster because old versions of the server don't know to update the new sidebar structure for the Favorites category. Basically, once someone has been migrated, any channels they favourite on community and community-release will be lost once they're moved to any version supporting the new sidebar.

Since we're far too late to implement the [Backwards Compatible Database Changes](https://docs.google.com/presentation/d/1TAh2pPIp8mu6InAn_eIBZ9z6SfNjr4DZxNhbSSPzD-w/edit#slide=id.g398d353b3c_0_154) idea, we've decided to mitigate against this by only migrating a user once they've actually used the sidebar (aka the on-demand migration). The issue will still exist, but it will only affect users who use the new sidebar on community-daily and then go back to community/community-release. It doesn't affect people who turn off the new sidebar and remain on community-daily, only those who go back to community/community-release.

This also adds proper migration of favourites to the on-demand migration (CreateInitialSidebarCategory) since that was previously missing and would have previously been done by the migration job.

Sorry, @reflog, for removing a bunch of your code :/

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-27007


store/sqlstore/channel_store.go
```diff
@@ -521,9 +481,16 @@ func (s SqlChannelStore) createInitialSidebarCategoriesT(transaction *gorp.Trans
 	}
 
 	if !hasCategoryOfType(model.SidebarCategoryFavorites) {
+		favoritesCategoryId := model.NewId()
+
+		// Create the SidebarChannels first since there's more opportunity for something to fail heree
```
lieut-data, line 486:
```suggestion
 		// Create the SidebarChannels first since there's more opportunity for something to fail here
```

store/sqlstore/channel_store.go
```diff
@@ -521,9 +481,16 @@ func (s SqlChannelStore) createInitialSidebarCategoriesT(transaction *gorp.Trans
 	}
 
 	if !hasCategoryOfType(model.SidebarCategoryFavorites) {
```
lieut-data, line 483: nit: a `hasCategoryOfType` that is simply a `map[model.SidebarCategoryType]bool` would be more idiomatic than looping every time.

store/sqlstore/channel_store.go
```diff
@@ -595,6 +562,88 @@ func (s SqlChannelStore) migrateMembershipToSidebar(transaction *gorp.Transactio
 	return memberships, nil
 }
 
+func (s SqlChannelStore) migrateFavoritesToSidebarT(transaction *gorp.Transaction, userId, teamId, favoritesCategoryId string) error {
+	if s.DriverName() == model.DATABASE_DRIVER_POSTGRES {
```
lieut-data, line 566: Since we have to do it for MySQL anyway, and the code there is already cross-database compatible, do we need the additional complexity (and testing consideration) here for Postgres?

store/sqlstore/channel_store.go
```diff
@@ -521,9 +481,16 @@ func (s SqlChannelStore) createInitialSidebarCategoriesT(transaction *gorp.Trans
...
 	if !hasCategoryOfType(model.SidebarCategoryFavorites) {
+		favoritesCategoryId := model.NewId()
+
+		// Create the SidebarChannels first since there's more opportunity for something to fail heree
+		if err := s.migrateFavoritesToSidebarT(transaction, userId, teamId, favoritesCategoryId); err != nil {
+			return errors.Wrap(err, "createInitialSidebarCategoriesT: failed to migrate favorites to sidebar")
+		}
+
 		if err := transaction.Insert(&model.SidebarCategory{
 			DisplayName: "Favorites", // This will be retranslated by the client into the user's locale
```
lieut-data, line 492: Do we have a database-level constraint on `DisplayName`? Worried that a racey double write for the first customer will result in duplicate `Favorites` sidebar categories.

store/sqlstore/channel_store.go
hmhealey: Ugh, I need to clean my keyboard or something. I make typos like that constantly.

store/sqlstore/channel_store.go
hmhealey: We tried adding a unique constraint on the `Type` field, but we weren't able to do it because users can have multiple "custom" categories. I looked into a partial unique constraint, but it's unfortunately not supported on MySQL. `hasCategoryOfType` and running this in a transaction should hopefully mitigate against it at least.

store/sqlstore/channel_store.go
hmhealey: Yeah, that's true. I was thinking that only making 1 database call vs N+1 database calls (where N is the number of favourited channels) would be better, but that's probably an unnecessary optimization.

store/sqlstore/channel_store.go
lieut-data: I'm fine with accepting the risk, but just want to note that without a unique constraint, a transaction won't guard against this at all.

hmhealey:
@ogi-m Any chance you could look at this as soon as possible? We need to get this onto community-daily to undo some of the previous automatic migration since it will cause issues when people using community start to use the new sidebar

hmhealey:
Cherry-picked manually onto release-5.26

PR #15088, opened 2020-07-21
MM-25744 Remove (*App).Shutdown and update all references.

wiggin77:
#### Summary
This PR removes `(*App).Shutdown()`, which forwarded the call to (*Server).Shutdown().  All references to it now call Shutdown on server directly.

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-25744

### DIFF
--- a/app/app.go
+++ b/app/app.go
@@ -117,14 +117,6 @@ func (a *App) InitServer() {
 	a.saml = a.srv.Saml
 }
 
-// DO NOT CALL THIS.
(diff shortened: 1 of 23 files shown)


wiggin77:
@lindy65 this PR changes all instances of `a.Shutdown()` to `a.Srv().Shutdown()` across the server CLI.  Little risk in that `a.Shutdown` simply called Srv().Shutdown anyway.  I suspect no specific QA is needed and the unit tests are enough.

PR #15094, opened 2020-07-23
MM-27187: Use the correct page offset for cache clear methods

agnivade:
We were just using page and incrementing by 1. This would fetch pages
one by one like (1-100, 2-102, 3-103) rather than (1-100,100-200,200-300).

We fix that to update the correct page offset.

#### Ticket Link

https://mattermost.atlassian.net/browse/MM-27187


agnivade:
@amyblais - This was a bug from 5.22, which needs to be backported to all the release branches.

PR #15100, opened 2020-07-23
MM-27178: Use FileSettings.Directory in export tool

agarciamontoro:
#### Summary

The export tool had the default path for data hardcoded, so when `FileSettings.Directory` was customized, the export failed to find emoji files. This PR fixes this.

While debugging the bug, I found that when modifying this value from the UI, switching from a custom value to an empty string, the setting was set to the root path `/` (as the value was set to an actual empty string and the code made sure that there is a trailing slash). I fixed this by setting the default value in SetDefaults not only when there is a nil pointer but also when we have an empty string.

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-27178


agarciamontoro:
We discussed offline whether it makes sense to check that this setting is non-empty in `FileSettings.isValid`. We can do that, and it will probably not hurt, but `IsValid` is called after setting the defaults when updating the config, so it won't be used in this specific case
https://github.com/mattermost/mattermost-server/blob/master/api4/config.go#L74

Another concern I have is that the `fixConfig` function adds a trailing slash to the setting, so if the value is empty (it won't be in this case, but I'm not aware of all the code paths this can follow), it will set this to the root path of the system. Which seems a bit dangerous I would not add that to the `isValid` function because someone may want to set the directory to root path (who knows).

Thoughts?

gabrieljackson:
Hmm, is that failed test just from something flakey?

agarciamontoro:
> Hmm, is that failed test just from something flakey?

@gabrieljackson I had forgotten to run `make i18n-extract`

gabrieljackson:
> so if the value is empty (it won't be in this case, but I'm not aware of all the code paths this can follow), it will set this to the root path of the system.

So this part is a bit scary for sure. I am not sure of all the code paths this can take so maybe we want to avoid that? What do you think @jespino?

agarciamontoro:
@gabrieljackson that is ineed fixed in the `fixConfig` function:

https://github.com/mattermost/mattermost-server/pull/15100/files#diff-ea169bc74a5d0a4e02747bf67c4757c6R70

What I meant before is that I did not add that check to `isValid` because someone may want to actually do it. So the value won't be automatically changed to be the root path, but if the user sets it to `/` explicitly, we assume it's valid.

agarciamontoro:
@amyblais what fix version should we add here?

amyblais:
I've added it to v5.27, unless it's something that should soak in master for a longer period of time then it can go to v5.28.

agarciamontoro:
Thank you, Amy! I think 5.27 is ok. Merging, then :)

PR #15104, opened 2020-07-23
MM-27197 - fix reference to non-instantiated error var

catalintomai:
Fix invalid reference to non-instantiated var.

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-27197


amyblais:
@catalintomai Is this only on master?

catalintomai:
> @catalintomai Is this only on master?
@amyblais I included it into 5.26 as well but not as a direct cherry-pick. The webapp version of the original (parent) PR was not cherry-picked at the time so I just made the change directly on the parent before manually cherry-picking it into 5.26.

PR #15108, opened 2020-07-24
MM-27204 Remove channels from old category when creating new one

hmhealey:
When creating a new category with channels in it, we previously would add the channels to the new category, but we wouldn't remove them from the old one. This fixes that an adds some tests.

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-27204


jgilliam17:
@hmhealey I am seeing an issue when creating new channels after a channel is archived. 
1. Archive a channel 
2. Create new channel
Observed: New channel is not added on LHS, it is added to Channels category only after refresh

PR #15109, opened 2020-07-24
MM-27258 Cannot update categories with deleted channels

hmhealey:
We have a check to make sure that users aren't trying to add channels they aren't a member of to their sidebar, but it was failing to check for archived channels, so that was causing it to reject any updates to categories containing an archived channel.

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-27258


prapti:
@hmhealey Is this PR only allowing the channels in categories containing archived channel can be reordered properly? Wondering if anything else needs to be tested here.

hmhealey:
@prapti It also affects renaming them, but they both go through the same code path

hmhealey:
Manually cherry-picked onto release-5.26

PR #15110, opened 2020-07-24
[MM-26837] attempt at fixing flaky test on mysql-5.6

ashishbhate:
#### Summary

- Attempt to fix the AutocompleteInTeam/empty_string test which is flaky on MySQL-5.6. This is likely an index_merge_intersection optimization issue. See https://community.mattermost.com/core/pl/uzgygwa9e3d4pns14ayfs7mobw
for more discussion. The fix here simply modifies the where clause to change the query plan so index_merge_intersection is not used, and while also not introducing a temporary table or a filesort. 

#### Ticket Link

- https://mattermost.atlassian.net/browse/MM-26837


ashishbhate:
> Since we are not just touching a test, it would be interesting to know if using Channels instead of PublicChannels in the comparison has any performance implication.

Do you have any ideas on how we could compare performance? I ran some local loadtests but that probably wont generate enough load, and it doesn't really output an easily comparable metric.

streamer45:
> > Since we are not just touching a test, it would be interesting to know if using Channels instead of PublicChannels in the comparison has any performance implication.
> 
> Do you have any ideas on how we could compare performance? I ran some local loadtests but that probably wont generate enough load, and it doesn't really output an easily comparable metric.

I think a low level benchmark might be more suited here. I am just not sure if it will be much meaningful unless we run it against a reasonably big dataset (in the order of thousands of channels).

This doesn't have to be done in the PR of course but could be some extra work done on the side just to make sure we are not introducing a performance regression.  

I can offer some help. I have a decently sized db dump from a recent load-test which could be probably used for such a purpose.

lindalumitchell:
Removed QA Review label; no manual QA testing needed for this one.

ashishbhate:
Benchmarking code is in draft PR #15240 

This change results in a slight performance improvement:

Without fix
```
pkg: github.com/mattermost/mattermost-server/v5/store/sqlstore
BenchmarkChannelStoreAutoComplete
BenchmarkChannelStoreAutoComplete/MySQL
BenchmarkChannelStoreAutoComplete/MySQL/AutoCompleteInTeam
BenchmarkChannelStoreAutoComplete/MySQL/AutoCompleteInTeam                   608           1909407 ns/op
BenchmarkChannelStoreAutoComplete/MySQL/AutoCompleteInTeam-2                 675           1822926 ns/op
BenchmarkChannelStoreAutoComplete/MySQL/AutoCompleteInTeam-4                 664           1837866 ns/op
BenchmarkChannelStoreAutoComplete/MySQL/AutoCompleteInTeam-8                 666           1785153 ns/op
```
	

With fix:
```
pkg: github.com/mattermost/mattermost-server/v5/store/sqlstore
BenchmarkChannelStoreAutoComplete
BenchmarkChannelStoreAutoComplete/MySQL
BenchmarkChannelStoreAutoComplete/MySQL/AutoCompleteInTeam
BenchmarkChannelStoreAutoComplete/MySQL/AutoCompleteInTeam                   747           1607595 ns/op
BenchmarkChannelStoreAutoComplete/MySQL/AutoCompleteInTeam-2                 807           1587679 ns/op
BenchmarkChannelStoreAutoComplete/MySQL/AutoCompleteInTeam-4                 778           1520310 ns/op
BenchmarkChannelStoreAutoComplete/MySQL/AutoCompleteInTeam-8                 756           1585894 ns/op
```

PR #15113, opened 2020-07-27
MM-26882: Ensure to start docker containers during test

agnivade:
This starts up docker containers which can help avoid failures
during a test run if a developers forgets to manually start them.

#### Ticket link

https://mattermost.atlassian.net/browse/MM-26882

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -317,7 +317,7 @@ gomodtidy:
 	fi;
 	@rm go.*.orig;
 
-test-server: check-prereqs-enterprise go-junit-report do-cover-file ## Runs tests.
+test-server: check-prereqs-enterprise start-docker go-junit-report do-cover-file ## Runs tests.
 ifeq ($(BUILD_ENTERPRISE_READY),true)
 	@echo Running all tests
 else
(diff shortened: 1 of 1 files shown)


agnivade:
Ah my bad, gotcha. Sorry for the to and fro.

Makefile
```diff
@@ -128,6 +128,12 @@ ifeq ($(RUN_SERVER_IN_BACKGROUND),true)
 	RUN_IN_BACKGROUND := &
 endif
 
+ifeq ($(BUILD_ENTERPRISE_READY),true)
```
jespino, line 131: My concern with this is, you are ignoring the user settings when you have enterprise ready builds. I'm ok to fix the current situation,. but I would prefer to iterate over this and find a better way to only run "openldap" container if you are running the tests and you are in a enterprise ready build.

agnivade: Yeah definitely, there's some Makefile magic needed which I don't know enough about. I initially moved the entire condition inside the test-server target, but it wasn't working. Then fell back to the global variable thing.

Some Makefile expertise is most welcome :D

agnivade:
@ogi-m - Not much to test here since it's just a Makefile change.

PR #15115, opened 2020-07-27
MM-27275: Log the correct error when GetUser fails in MFA Authentication

agnivade:
We were not logging the error returned from GetUser which makes debugging things
very hard in case of logout issues like https://mattermost.atlassian.net/browse/MM-27270.

We fix this by appending the error inside model.AppError and setting a proper error string
which says exactly what has happened.

Fixes https://github.com/mattermost/mattermost-server/issues/15041

#### Ticket link

https://mattermost.atlassian.net/browse/MM-27275


web/context.go
```diff
@@ -154,35 +154,36 @@ func (c *Context) MfaRequired() {
...
-			c.Err = model.NewAppError("", "api.context.mfa_required.app_error", nil, "MfaRequired", http.StatusForbidden)
-			return
-		}
+	// Bots are exempt
+	if user.IsBot {
+		return
+	}
+
+	if !user.MfaActive {
+		c.Err = model.NewAppError("", "api.context.mfa_required.app_error", nil, "MfaRequired", http.StatusForbidden)
```
wiggin77, line 185: Should this get a "where" value of "MfaRequired" like above?

wiggin77, line 185:
```suggestion
		c.Err = model.NewAppError("MfaRequired", "api.context.mfa_required.app_error", nil, "", http.StatusForbidden)
```

agnivade: Nice one. I fixed the original error, but missed this one.

agnivade:
@lindy65 - No need to test this as it's already covered by a unit test and there is no functional change outside to be observed. Just a logging enhancement.

PR #15116, opened 2020-07-27
MM-27199 - Add metrics to MM server for total enabled user count

reflog:
#### Summary
To track the user count usage of a Mattermost Cloud installation we need to export that metric. To do that:

1. Add a new metric for total enabled users (note: does not include deactivated users)
2. Read an env var containing the installation ID that can be assumed will always be set for cloud installations and include that as a tag on all metrics
3. Read the same env var for installation ID and add it to the telemetry we send in diagnostics.go

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-27199

#### Related enterprise PR
https://github.com/mattermost/enterprise/pull/713


reflog:
@wiggin77 absolutely. It would be a waste to have all this code for just one metric 😳

PR #15123, opened 2020-07-28
First part of PostStore migration

rvillablanca:
#### Summary

Continuing with the campaign to migrate to plain errors in the store layer, this is the first part of the PostStore migration.

Fixes : https://github.com/mattermost/mattermost-server/issues/15132


app/post.go
```diff
@@ -859,7 +917,11 @@ func (a *App) DeletePost(postId, deleteByID string) (*model.Post, *model.AppErro
 	}
 
 	if err := a.Srv().Store.Post().Delete(postId, model.GetMillis(), deleteByID); err != nil {
-		return nil, err
+		var nfErr *store.ErrNotFound
```
jespino, line 920: We don't need all this here,  it would be just `return nil, model.NewAppError(...`, right?

jespino: Actually, I think this is changing the behavior. This was failing always previously, and now only fails if the error is a "not-found" error. Maybe we need the switch, but we need the default case too.

store/sqlstore/post_store.go
```diff
@@ -254,18 +256,24 @@ func (s *SqlPostStore) Update(newPost *model.Post, oldPost *model.Post) (*model.
 	}
 
 	if _, err := s.GetMaster().Update(newPost); err != nil {
-		return nil, model.NewAppError("SqlPostStore.Update", "store.sql_post.update.app_error", nil, "id="+newPost.Id+", "+err.Error(), http.StatusInternalServerError)
+		return nil, errors.Wrapf(err, "failed to update Post with id=%s", newPost.Id)
 	}
 
 	time := model.GetMillis()
-	s.GetMaster().Exec("UPDATE Channels SET LastPostAt = :LastPostAt  WHERE Id = :ChannelId AND LastPostAt < :LastPostAt", map[string]interface{}{"LastPostAt": time, "ChannelId": newPost.ChannelId})
+	if _, nErr := s.GetMaster().Exec("UPDATE Channels SET LastPostAt = :LastPostAt  WHERE Id = :ChannelId AND LastPostAt < :LastPostAt", map[string]interface{}{"LastPostAt": time, "ChannelId": newPost.ChannelId}); nErr != nil {
```
jespino, line 263: You are changing the behavior here, this is intentionally fire and forget, because if this fails, we don't care, will be eventually consistent in the future with the next post, is something that is not going to happen often, and is something that we prefer to not break the new post creation because of this. Maybe we can add logs here instead of returning the error.

store/sqlstore/post_store.go
```diff
@@ -254,18 +256,24 @@ func (s *SqlPostStore) Update(newPost *model.Post, oldPost *model.Post) (*model.
...
 	if len(newPost.RootId) > 0 {
-		s.GetMaster().Exec("UPDATE Posts SET UpdateAt = :UpdateAt WHERE Id = :RootId AND UpdateAt < :UpdateAt", map[string]interface{}{"UpdateAt": time, "RootId": newPost.RootId})
+		if _, nErr := s.GetMaster().Exec("UPDATE Posts SET UpdateAt = :UpdateAt WHERE Id = :RootId AND UpdateAt < :UpdateAt", map[string]interface{}{"UpdateAt": time, "RootId": newPost.RootId}); nErr != nil {
+			return nil, errors.Wrap(nErr, "failed to update Posts")
+		}
 	}
 
 	// mark the old post as deleted
-	s.GetMaster().Insert(oldPost)
+	if nErr := s.GetMaster().Insert(oldPost); nErr != nil {
```
jespino, line 274: I'm not 100% sure about this generating a failure. what do you think @agnivade I think here should fail.

store/sqlstore/post_store.go
```diff
@@ -254,18 +256,24 @@ func (s *SqlPostStore) Update(newPost *model.Post, oldPost *model.Post) (*model.
...
 
 	time := model.GetMillis()
-	s.GetMaster().Exec("UPDATE Channels SET LastPostAt = :LastPostAt  WHERE Id = :ChannelId AND LastPostAt < :LastPostAt", map[string]interface{}{"LastPostAt": time, "ChannelId": newPost.ChannelId})
+	if _, nErr := s.GetMaster().Exec("UPDATE Channels SET LastPostAt = :LastPostAt  WHERE Id = :ChannelId AND LastPostAt < :LastPostAt", map[string]interface{}{"LastPostAt": time, "ChannelId": newPost.ChannelId}); nErr != nil {
+		return nil, errors.Wrap(nErr, "failed to update Channels")
+	}
 
 	if len(newPost.RootId) > 0 {
-		s.GetMaster().Exec("UPDATE Posts SET UpdateAt = :UpdateAt WHERE Id = :RootId AND UpdateAt < :UpdateAt", map[string]interface{}{"UpdateAt": time, "RootId": newPost.RootId})
+		if _, nErr := s.GetMaster().Exec("UPDATE Posts SET UpdateAt = :UpdateAt WHERE Id = :RootId AND UpdateAt < :UpdateAt", map[string]interface{}{"UpdateAt": time, "RootId": newPost.RootId}); nErr != nil {
```
jespino, line 268: Same here, we don't want to fail the whole function because this is not updated.

app/post.go
rvillablanca: yep you are right Jesús, it is a bug and I forgot to add the default case, thanks!

store/sqlstore/post_store.go
rvillablanca: Yes I know I'm changing the behaviour here and I understand your point. Agree with you that maybe it is not very important to update the Channel for this case but any error received while doing an sql operation should be handled no matter it is important or no for our business rules. I would understand that this for cases when for example the error could be `sql.ErrNoRows` because it is expected in some cases not finding some values for bad requests.

For this case I think we should handle the error not because it is important to update the channel itself, but simply an error here should not happen and if that is ocurring then ignore or not handle meas we are ignoring a possible future error, perhaps something like logging it is less drastic.

store/sqlstore/post_store.go
rvillablanca: Same as above, return the error or logging it at least if the return is too much

app/import_functions.go
```diff
@@ -1229,7 +1238,8 @@ func (a *App) importMultiplePostLines(lines []LineImportWorkerData, dryRun bool)
 		user := users[*line.Post.User]
 
 		// Check if this post already exists.
-		posts, err := a.Srv().Store.Post().GetPostsCreatedAt(channel.Id, *line.Post.CreateAt)
+		var posts []*model.Post
```
agnivade, line 1241: Not sure why we need to declare this. Are we running into shadowing issues?

app/import_functions.go
```diff
@@ -1256,7 +1266,8 @@ func (a *App) importMultiplePostLines(lines []LineImportWorkerData, dryRun bool)
 			post.Props = *line.Post.Props
 		}
 
-		fileIds, err := a.uploadAttachments(line.Post.Attachments, post, team.Id, dryRun)
+		var fileIds map[string]bool
```
agnivade, line 1269: Same here too

app/import_functions.go
```diff
@@ -1281,14 +1292,26 @@ func (a *App) importMultiplePostLines(lines []LineImportWorkerData, dryRun bool)
...
+			var appErr *model.AppError
+			var invErr *store.ErrInvalidInput
+			var retErr *model.AppError
+			switch {
+			case errors.As(err, &appErr):
+				retErr = appErr
+			case errors.As(err, &invErr):
+				retErr = model.NewAppError("importMultiplePostLines", "app.post.save.existing.app_error", nil, invErr.Error(), http.StatusBadRequest)
+			default:
+				retErr = model.NewAppError("importMultiplePostLines", "app.post.save.app_error", nil, err.Error(), http.StatusInternalServerError)
```
agnivade, line 1305: It should be nErr inside the model.NewAppError

app/integration_action.go
```diff
@@ -90,16 +90,22 @@ func (a *App) DoPostActionWithCookie(postId, actionId, userId, selectedOption st
...
+			switch {
+			case errors.As(result.NErr, &nfErr):
+				return "", model.NewAppError("DoPostActionWithCookie", "app.post.get.app_error", nil, nfErr.Error(), http.StatusNotFound)
+			default:
+				return "", model.NewAppError("DoPostActionWithCookie", "app.post.get.app_error", nil, result.NErr.Error(), http.StatusInternalServerError)
+			}
 		}
 		if cookie.Integration == nil {
-			return "", model.NewAppError("DoPostAction", "api.post.do_action.action_integration.app_error", nil, "no Integration in action cookie", http.StatusBadRequest)
+			return "", model.NewAppError("DoPostActionWithCookie", "api.post.do_action.action_integration.app_error", nil, "no Integration in action cookie", http.StatusBadRequest)
```
agnivade, line 104: :+1:

app/post.go
```diff
@@ -290,9 +290,18 @@ func (a *App) CreatePost(post *model.Post, channel *model.Channel, triggerWebhoo
...
+	if nErr != nil {
+		var appErr *model.AppError
+		var invErr *store.ErrInvalidInput
+		switch {
+		case errors.As(err, &appErr):
+			return nil, appErr
+		case errors.As(err, &invErr):
+			return nil, model.NewAppError("CreatePost", "app.post.save.existing.app_error", nil, invErr.Error(), http.StatusBadRequest)
+		default:
+			return nil, model.NewAppError("CreatePost", "app.post.save.app_error", nil, err.Error(), http.StatusInternalServerError)
```
agnivade, line 303: nErr

app/import_functions.go
```diff
@@ -1281,14 +1292,26 @@ func (a *App) importMultiplePostLines(lines []LineImportWorkerData, dryRun bool)
 	}
 
 	if len(postsForCreateList) > 0 {
-		if _, idx, err := a.Srv().Store.Post().SaveMultiple(postsForCreateList); err != nil {
+		if _, idx, nErr := a.Srv().Store.Post().SaveMultiple(postsForCreateList); nErr != nil {
+			var appErr *model.AppError
+			var invErr *store.ErrInvalidInput
+			var retErr *model.AppError
+			switch {
+			case errors.As(err, &appErr):
```
agnivade, line 1300: `errors.As(nErr, <>)`. This is there in multiple places.

app/post.go
```diff
@@ -290,9 +290,18 @@ func (a *App) CreatePost(post *model.Post, channel *model.Channel, triggerWebhoo
...
 
-	rpost, err := a.Srv().Store.Post().Save(post)
-	if err != nil {
-		return nil, err
+	rpost, nErr := a.Srv().Store.Post().Save(post)
+	if nErr != nil {
+		var appErr *model.AppError
+		var invErr *store.ErrInvalidInput
+		switch {
+		case errors.As(err, &appErr):
```
agnivade, line 298: Here too

app/post.go
agnivade: Thanks Rodrigo. I am going to unresolve this so that it remains visible until a new commit resolves it.

store/sqlstore/post_store.go
agnivade: Yes, I think the updating of the newPost and insertion of oldPost should happen in a transaction and we should fail if any one fails. But this PR is already pretty big, I think we can leave that part of changing this to a transaction in a separate PR.

Regarding the other 2, we are not much bothered to check the errors, because like @jespino said, the next post update should fix this. Yes, ideally we should log them. But using a logger at such a low level like store is not really a good practice. The leaf functions should just do computation and return values. Only the top-level functions should be responsible for logging the success/error.

That said, I completely agree that having an unhandled error function in the code is not a solution too. Let me think some more on it. For now, let's keep the old behavior of ignoring errors.

app/import_functions.go
rvillablanca: Fixed

app/import_functions.go
rvillablanca: Fixed

app/import_functions.go
rvillablanca: Fixed

app/import_functions.go
rvillablanca: Sorry I didn't understand this suggestions. Did you mean we have code duplication ?

app/post.go
rvillablanca: Fixed

app/post.go
rvillablanca: Fixed

app/post.go
rvillablanca: It is strange because I thought I already added it :thinking:

store/sqlstore/post_store.go
rvillablanca: Yep, I missed the fact of adding a logger in this layer is not appropiate. I will revert this part of the code because it lacks of transaction so we can check it again in another PR with more calm and time

store/sqlstore/post_store.go
rvillablanca: Finally I'm going to leave the code as is. We can get into this in another PR. Thanks @jespino

store/sqlstore/post_store.go
rvillablanca: Ok

app/post.go
rvillablanca: It is already solved

app/import_functions.go
rvillablanca: This seems fixed with changes I made so I'm goint to mark it as revolved :)

PR #15140, opened 2020-07-29
[MM-27281] Fix possible panic on server shutdown

streamer45:
#### Summary

PR fixes a panic happening when the server was getting shutdown before `InitPlugins()` was able to complete.

#### Ticket

https://mattermost.atlassian.net/browse/MM-27281

### DIFF
--- a/app/plugin.go
+++ b/app/plugin.go
@@ -184,6 +184,10 @@ func (a *App) InitPlugins(pluginDir, webappPluginDir string) {
 
 	plugins := a.processPrepackagedPlugins(prepackagedPluginsDir)
 	pluginsEnvironment = a.GetPluginsEnvironment()
+	if pluginsEnvironment == nil {
+		mlog.Info("Plugins environment not found, server is likely shutting down")
+		return
+	}
 	pluginsEnvironment.SetPrepackagedPlugins(plugins)
 
 	// Sync plugin active state when config changes. Also notify plugins.


app/plugin.go
```diff
@@ -184,6 +184,10 @@ func (a *App) InitPlugins(pluginDir, webappPluginDir string) {
 
 	plugins := a.processPrepackagedPlugins(prepackagedPluginsDir)
 	pluginsEnvironment = a.GetPluginsEnvironment()
+	if pluginsEnvironment == nil {
+		mlog.Info("Plugins environment not found, server is likely shutting down")
```
streamer45, line 188: Not sure if this should be an error.

app/plugin.go
```diff
@@ -184,6 +184,10 @@ func (a *App) InitPlugins(pluginDir, webappPluginDir string) {
 
 	plugins := a.processPrepackagedPlugins(prepackagedPluginsDir)
 	pluginsEnvironment = a.GetPluginsEnvironment()
+	if pluginsEnvironment == nil {
```
lieut-data, line 187: Would it make sense to just use the `env` created a few lines above and sidestep the unexpected shutdown?

streamer45: That was another option that I considered but I opted for returning early and avoid the extra time since getting a `nil` environment at that point can only mean that either something is very wrong or we are simply shutting down.
I wasn't sure it made sense to continue at that point but I guess both ways work.

lieut-data: :+1:

streamer45:
@ogi-m 
this is quite tricky to reproduce and I was able to do it only by messing with the code.
I don't think there's much to do as far as QA. The PR only adds a `nil` check.

amyblais:
@streamer45 Kind note to cherry-pick this,

PR #15145, opened 2020-07-30
Removed empty translation

enahum:
#### Summary
For some reason there is a translation with an empty key and value

### DIFF
--- a/i18n/en.json
+++ b/i18n/en.json
@@ -1,9 +1,5 @@
 [
   {
-    "id": "",
-    "translation": ""
-  },
-  {
     "id": "April",
     "translation": "April"
   },


metanerd:
I think someone tried to remove this one already, but the translation check script makes it fail. Can we just merge it even if it fails?

enahum:
@jespino would you know why the check fails?

metanerd:
```
/go/src/github.com/mattermost/mattermost-server/bin/mmgotool i18n extract --portal-dir=""
2a3,6
>     "id": "",
>     "translation": ""
>   },
>   {
```

enahum:
There must be something in the codebase that has an empty string

jespino:
@metanerd yes, the problem was an error that was using an empty string, now is fixed here: #15166

Should be fixed in master.

PR #15149, opened 2020-07-30
[MM-27407] Return archived channels even if view archived config is false

fmunshi:
#### Summary
- These two functions are only used in endpoints that are gated by `manage_system` permission so we're fine to just remove this view archived config setting since this setting is only applicable for the chat facing UI and not for the system console 

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-27407


amyblais:
@mkraft @ashishbhate Kind note to prioritize reviews for v5.26,

api4/channel_test.go
```diff
@@ -969,15 +969,6 @@ func TestGetAllChannels(t *testing.T) {
...
 
-	var originalConfigVal bool
-	th.App.UpdateConfig(func(cfg *model.Config) {
-		originalConfigVal = *cfg.TeamSettings.ExperimentalViewArchivedChannels
-		*cfg.TeamSettings.ExperimentalViewArchivedChannels = true
-	})
-	defer th.App.UpdateConfig(func(cfg *model.Config) {
-		*cfg.TeamSettings.ExperimentalViewArchivedChannels = originalConfigVal
-	})
-
```
ashishbhate, line 980: Maybe explicitly set to `false` to test that it doesn't affect the behaviour of these methods. The test will also make the intent and usage clearer.

PR #15156, opened 2020-07-31
MM-26584: Bump up Go patch version

agnivade:
This contains a patch fix for https://github.com/golang/go/issues/38023
which have been encountered in our load tests and some users.

I have verified on such a stuck instance and the thread dump seem to match
with what is seen on the issue. All threads stuck on runtime.futex, except one,
which is at runtime.osyield.

And of all the times load tests were run which _did not_ show this issue,
it was run with a version greater than 1.14.1, which further points to the
theory that this bug is the culprit.

Fixes https://github.com/mattermost/mattermost-server/issues/15092

#### Ticket Link

https://mattermost.atlassian.net/browse/MM-26584


agnivade:
@amyblais - This is a corner case bug which got introduced with 5.24. Since this happens rarely, IMO this does not deserve to be a backport candidate. But I would propose this to be cherry-picked to 5.26 so that customers don't have to wait for another release to get the fix.

The fix is very small and only contains patch upgrades targeted for specific bugs.

build/README.md
```diff
@@ -18,3 +18,4 @@ In our Docker Hub Repository we have the following images:
 - `mattermost/mattermost-build-server:feb-28-2019` which is based on Go 1.12 you can use for MM versions >= `5.9.0` <= `5.15.0`
 - `mattermost/mattermost-build-server:sep-17-2019` which is based on Go 1.12.9 you can use for MM versions >= `5.16.0`
 - `mattermost/mattermost-build-server:20200322_golang-1.14.1` which is based on Go 1.14.1 you can use for MM versions >= `5.24.0`
+- `mattermost/mattermost-build-server:20200731_golang-1.14.6` which is based on Go 1.14.6 you can use for MM versions >= `5.27.0`
```
metanerd, line 21: Someone please check if this is true.

agnivade: Should be >= 5.26. Since we are going to cherry-pick this.

metanerd, line 21:
```suggestion
- `mattermost/mattermost-build-server:20200731_golang-1.14.6` which is based on Go 1.14.6 you can use for MM versions >= `5.26.0`
```

agnivade:
No QA required for this one as this has been hard to repro. And there is no code change in Mattermost.

PR #15165, opened 2020-07-31
MM-27184 deprecate model.SetExpireInDays to fix mobile logout bug

wiggin77:
#### Summary
Mobile users were having their sessions unexpectedly expired, despite having `ServiceSettings.ExtendSessionLengthWithActivity` enabled.  Bug was found here:

/model/session.go:118
```go
func (me *Session) SetExpireInDays(days int) {
    if me.CreateAt == 0 {
        me.ExpiresAt = GetMillis() + (1000 * 60 * 60 * 24 * int64(days))
    } else {
        me.ExpiresAt = me.CreateAt + (1000 * 60 * 60 * 24 * int64(days))
    }
}
```
This gets called for mobile users every time the app is opened via `/api/v4/sessions/device` which calls `attachDeviceId` which calls the code above.  The code above assumes the expiry should be relative to CreateAt which is incorrect when `ExtendSessionLengthWithActivity` is enabled.  Therefore, every time the mobile app was opened, the maximum expiry was set in memory to CreateAt + session_length, even if the session was extended.

`(*Session)SetExpireInDays` is now deprecated and replaced with `(*App)SetSessionExpireInDays` which takes into account the `ExtendSessionLengthWithActivity` setting.

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-27184


lindy65:
Thanks @wiggin77 - currently testing on iOS, Android, Desktop app and browser (there were reports of the bug affecting all platforms - not only RN apps) 👍

wiggin77:
> Thanks @wiggin77 - currently testing on iOS, Android, Desktop app and browser (there were reports of the bug affecting all platforms - not only RN apps)

Thanks Lindy.  Testing all platforms is great, however I've not seen any reports of this specific issue happening other than mobile.   If you are referencing this https://mattermost.atlassian.net/browse/MM-27270 ticket, it is not related, despite it being referenced in the ticket I addressed.

lindy65:
Ah, ok - good to know @wiggin77 👍 So far, so good - I haven't been logged out on any platform...  Will approve the PR tomorrow morning after checking one final time :)

amyblais:
@wiggin77 Kind note to cherry-pick this,

PR #15171, opened 2020-08-02
MM-27456: Use reflect-free serialization for hot structs

agnivade:
We use the tinylib/msgp serializer which generates methods for fast
serialization and de-serialization. Because of this, we can get rid of the
reflect overhead which comes with generic serializer.

However, the library is a cumbersome to work with. Because it does not support package
level parsing, we need to move some code temporarily for the generation to work
and then move them back.

We choose model.User and model.Session as our candidates for this migration.

model.User is by far our hottest cache struct because of getUserStatusesByIds calls.
Of all the cache calls, model.User is nearly 90% of them. Following the 80-20 rule,
this should be the only struct good enough to migrate. However, we add one more.

model.Session is actually the 4th hottest struct. But it is called very frequently
during user logins, as found from CPU profiles during load testing.

```
name            old time/op    new time/op    delta
LRU/User=new-8    19.2µs ± 1%     8.6µs ± 5%  -55.28%  (p=0.000 n=10+10)

name            old alloc/op   new alloc/op   delta
LRU/User=new-8    9.93kB ± 0%    4.78kB ± 0%  -51.87%  (p=0.000 n=9+10)

name            old allocs/op  new allocs/op  delta
LRU/User=new-8       102 ± 0%        91 ± 0%  -10.78%  (p=0.000 n=10+10)

LRU/Session=new-8    7.14µs ± 3%    3.21µs ± 7%  -55.01%  (p=0.000 n=9+10)

LRU/Session=new-8    2.92kB ± 0%    1.71kB ± 0%  -41.59%  (p=0.002 n=8+10)

LRU/Session=new-8      36.0 ± 0%      28.0 ± 0%  -22.22%  (p=0.000 n=10+10)
```

Comparison of pooled with un-pooled with only sessions.

```
name \ time/op     master.txt   new-unpooled.txt  new-pooled.txt
LRU/Session=new-8  7.35µs ± 7%       3.59µs ± 9%     3.47µs ± 6%

name \ alloc/op    master.txt   new-unpooled.txt  new-pooled.txt
LRU/Session=new-8  2.92kB ± 0%       2.15kB ± 0%     1.71kB ± 0%

name \ allocs/op   master.txt   new-unpooled.txt  new-pooled.txt
LRU/Session=new-8    36.0 ± 0%         29.0 ± 0%       28.0 ± 0%
```

This gives a clear reduction in the avg API times, although at some extra cost of a stack allocation which is can be considered as practically free. This is because we have to allocate an object on stack for msgp to deserialize.

There are still lots of places to improve, especially the allocation. For every serialization, we have to pay for the cost of a byte slice which leads to a lot of mark and sweep time being spent by the CPU. Now that msgp allows alloc-free marshaling, we need to start using a sync.Pool to help reduce allocations and take back some CPU time.

We also need to move vmihailenco/msgpack API to be alloc-free. MM-27457 tracks that effort.

There are still more tricks like using striped-mutexes and COW pattern to reduce lock contention. More investigation tickets will be created for those.

p99-api-times
avg-api-times
stack-in-use
heap-in-use
cpu-utilization-app

#### Ticket Link

https://mattermost.atlassian.net/browse/MM-27456


jespino:
@agnivade I'll be on holidays the next week, I'll try to take a look, but no promises, if you need a quick review, reassign it, if not, I'm happy to review it when I can or asap after I back from my holidays.

agnivade:
@jespino - Oh there's absolutely no rush. Enjoy your holidays and come back and review :)

agnivade:
I found some discrepancies in the code due to which the fast path wasn't triggering. No wonder I wasn't seeing any benefits in the graphs.

agnivade:
Ok to review. But CI won't pass until https://github.com/mattermost/mattermost-govet/pull/8 is merged.

services/cache/lru.go
```diff
@@ -182,6 +196,34 @@ func (l *LRU) get(key string, value interface{}) error {
 
 		l.evictList.MoveToFront(ent)
 
+		// We use a fast path for hot structs.
+		if msgpVal, ok := value.(msgp.Unmarshaler); ok {
+			_, err := msgpVal.UnmarshalMsg(e.value)
+			return err
+		}
```
streamer45, line 203: Is this actually needed?

agnivade: Fair concern. I did think about it, and at the end decided to keep it mainly for correctness. If somewhere in the codebase, we ever need to pass pointers to values for hot structs, this should still work.

0/5 if you prefer it to be removed, but we need to mention somewhere that passing pointers to values won't work for hot structs. I thought a couple of extra lines to not have any surprising behavior for a developer is a fair tradeoff.

streamer45: Sure, I guess it's fine to leave it also because it appears to be the proper way of dealing with cached objects.

services/cache/lru.go
```diff
@@ -182,6 +196,34 @@ func (l *LRU) get(key string, value interface{}) error {
...
+			var u model.User
+			_, err := u.UnmarshalMsg(e.value)
+			*v = &u
+			return err
+		case **model.Session:
+			var s model.Session
+			_, err := s.UnmarshalMsg(e.value)
+			*v = &s
+			return err
+		}
```
streamer45, line 224: A couple of question:

1. What was the reason to pass a pointer to a pointer to the `Get` method?
2. (Unrelated to the PR) Do we still need to do deep copies now that we serialize?

agnivade: 1. It's OK to just pass a pointer. But the problem is that it only works with normal Marshal/Unmarshal functions. When you have a method of a struct doing the marshal/unmarshalling, then it the problem arises.

This is what we do for example:

```go
var session *model.Session
sessionCache.Get(token, &session)
```

So session is just a `nil` pointer, and then we are populating that pointer variable during unmarshalling. Now since this is essentially a `**`, that means the method set of the type does not satisfy `msp.Unmarshaler`. That's the problem.

2. Nope, I have a PR planned to fix that. Good spot. :)

streamer45: > 1. It's OK to just pass a pointer. But the problem is that it only works with normal Marshal/Unmarshal functions. When you have a method of a struct doing the marshal/unmarshalling, then it the problem arises.
> 
> This is what we do for example:
> 
> ```go
> var session *model.Session
> sessionCache.Get(token, &session)
> ```
> 
> So session is just a `nil` pointer, and then we are populating that pointer variable during unmarshalling. Now since this is essentially a `**`, that means the method set of the type does not satisfy `msp.Unmarshaler`. That's the problem.

I see. I guess the type switch for the fast path could be generated as well at some point if it becomes a burden to maintain.

agnivade: Agree. I don't think we need to add any more structs in the hot path for now. These 2 are fine.

A considerable CPU time is just spent in GC generated from the marshal/unmarshal. So the next steps are to remove the deep copy and try to use a sync.Pool for msgp. And in future modify the other msgpack library to have alloc-free APIs too.

services/cache/lru.go
```diff
@@ -182,6 +196,34 @@ func (l *LRU) get(key string, value interface{}) error {
...
+		// We use a fast path for hot structs.
+		if msgpVal, ok := value.(msgp.Unmarshaler); ok {
+			_, err := msgpVal.UnmarshalMsg(e.value)
+			return err
+		}
+
+		// This is ugly and makes the cache package aware of the model package.
+		// But this is due to 2 things.
+		// 1. The msgp package works on methods on structs rather than functions.
+		// 2. Our cache interface passes pointers to empty pointers, and not pointers
```
jespino, line 208: Pointers to empty pointers? did you mean pointeres to empty interfaces?

model/serialized_gen.go
```diff
@@ -0,0 +1,1622 @@
+// Copyright (c) 2015-present Mattermost, Inc. All Rights Reserved.
```
jespino, line 1: A lot of generated code here in the same model package. I know we are adding methods to the same struct, but my concern is generating too much code in the same package and make the compilation process of the `model` package slower because of that (no compile cache).

agnivade: That's a fair concern. I don't see a way around that unfortunately. The methods need to be of the same struct. And also arguably the `model` layer is modified much less often than app or api4.

services/cache/lru.go
agnivade: I meant the first. `var session *Session` and then `&session` is a pointer to a nil pointer. The fact that it eventually gets passed through `interface{}` is different, but not strictly related to the problem.

agnivade:
Assining to @lindy65 to balance Rohitesh's load.

PR #15173, opened 2020-08-02
Document sql store 'SqlTeamStore' methods

trebabcock:
#### Summary

Added documentation for team_store.go
#### Ticket Link

Fixes https://github.com/mattermost/mattermost-server/issues/14004

### DIFF
--- a/store/sqlstore/team_store.go
+++ b/store/sqlstore/team_store.go
@@ -951,6 +951,7 @@ func (s SqlTeamStore) UpdateMember(member *model.TeamMember) (*model.TeamMember,
 	return members[0], nil
 }
 
+// GetMember returns a member based on the member's UserId
 func (s SqlTeamStore) GetMember(teamId string, userId string) (*model.TeamMember, *model.AppError) {
 	query := s.getTeamMembersWithSchemeSelectQuery().
 		Where(sq.Eq{"TeamMembers.TeamId": teamId}).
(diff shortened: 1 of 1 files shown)


store/sqlstore/team_store.go
```diff
@@ -951,6 +951,7 @@ func (s SqlTeamStore) UpdateMember(member *model.TeamMember) (*model.TeamMember,
 	return members[0], nil
 }
 
+// GetMember returns a member based on the member's UserId
```
jespino, line 954: I think you are assuming because you are in the TeamStore the `teamId` is obvious enough, but I think is better to be explicit, @jgilliam17 what do you think?

jespino: :point_up: @justinegeffen

justinegeffen: Thanks, @jespino. Apologies for the delay, I missed this ping.

```suggestion
//GetMember returns a member based on the their teamId and userId
```
Not sure if "based on" is correct or we should say "using" but I see "based on" is used for GetTeamsForUser as well so 0/5 on that.

store/sqlstore/team_store.go
```diff
@@ -1082,6 +1087,7 @@ func (s SqlTeamStore) GetMembersByIds(teamId string, userIds []string, restricti
 	return dbMembers.ToModel(), nil
 }
 
+// GetTeamsForUser returns all teams that a user is a member of, based on the user's UserId
```
justinegeffen, line 1090:
```suggestion
// GetTeamsForUser returns all teams that a user is a member of, based on the user's userId
```

jasonblais:
@trebabcock thanks for the contribution! Would you be open to reviewing the two suggestions from @justinegeffen?

hanzei:
This PR has become orphaned and will be closed. Still, thanks for your contribution @trebabcock.

PR #15177, opened 2020-08-03
[MM-27328] Upgrade minio-go to v7

streamer45:
#### Summary

PR upgrades `minio-go` dependency from `v6` to `v7`.

#### Ticket

https://mattermost.atlassian.net/browse/MM-27328


services/filesstore/s3store.go
```diff
@@ -149,14 +155,17 @@ func (b *S3FileBackend) CopyFile(oldPath, newPath string) *model.AppError {
...
 	newPath = filepath.Join(b.pathPrefix, newPath)
-
-	source := s3.NewSourceInfo(b.bucket, oldPath, nil)
-	destination, err := s3.NewDestinationInfo(b.bucket, newPath, encrypt.NewSSE(), nil)
-	if err != nil {
-		return model.NewAppError("copyFile", "api.file.write_file.s3.app_error", nil, err.Error(), http.StatusInternalServerError)
+	srcOpts := s3.CopySrcOptions{
+		Bucket:     b.bucket,
+		Object:     oldPath,
+		Encryption: encrypt.NewSSE(),
```
streamer45, line 161: This is not related to the upgrade but I've noticed that for some reason we encrypt by default on `CopyFile` and `MoveFile` operations while on `WriteFile` we do that based on the value of `S3FileBackend.encrypt`.

I am wondering if anyone has more context on this.

grundleborg: I have no idea... does `git blame` point to anyone in particular who we could ask?

streamer45: I checked and it looks like it's been that way for years. 
The actual code was introduced during an upgrade in https://github.com/mattermost/mattermost-server/commit/baba8fa92f47cad604f6d2a2714d09f89178fbff but before that there was this `s3CopyMetadata()` function which would default to encryption anyway.

grundleborg: Hmm OK. I'm not sure if this is something it makes sense to "fix" or not.

PR #15197, opened 2020-08-06
[MM-27535] User invite limits for MM Cloud

nickmisasi:
#### Summary
In MM cloud we'll have user invite limits. This PR adds a configuration for these limits, as well as some logic for inviting users via email. If there are n users allowed in a server, and x users on the server already, n - x users will be allowed to be invited. Anyone else will receive a graceful error.

Note: This restriction only exists for Administrators. Some more front-end work is coming to prevent these emails from being sent. This logic is more of a fallback.

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-27534


api4/team.go
```diff
@@ -1193,7 +1206,45 @@ func inviteUsersToTeam(c *Context, w http.ResponseWriter, r *http.Request) {
...
+		}
+		var invitesWithError []*model.EmailInviteWithError
+		var err *model.AppError
+		if emailList != nil {
+			invitesWithError, err = c.App.InviteNewUsersToTeamGracefully(emailList, c.Params.TeamId, c.App.Session().UserId)
+		}
+		// Add the EmailInviteWithError's we have from user's over the limit
+		if len(invitesNotSent) > 0 {
+			invitesWithError = append(invitesWithError, invitesNotSent...)
+		}
```
jwilander, line 1247: Generally we try to keep the logic inside the api handlers to a minimum and we avoid any calls to the database in them. Instead, we use the `app` package to handle all that. With that in mind, can you factor this out as a function into the `app` package?

nickmisasi: I've shuffled around the meatier part of the logic into the app package.

api4/team_test.go
```diff
@@ -2721,6 +2721,59 @@ func TestImportTeam(t *testing.T) {
...
+	defer th.TearDown()
+	email1 := th.GenerateTestEmail()
+	email2 := th.GenerateTestEmail()
+	email3 := th.GenerateTestEmail()
+	th.App.UpdateConfig(func(cfg *model.Config) {
+		*cfg.ServiceSettings.EnableEmailInvitations = true
+		*cfg.ExperimentalSettings.CloudUserLimit = 2
+	})
+
+	// System admin, invite when at limit should fail
```
mgdelacroix, line 2735: nit: these cases can be moved to `t.Run` blocks

app/team.go
```diff
@@ -1127,6 +1127,45 @@ func (a *App) prepareInviteNewUsersToTeam(teamId, senderId string) (*model.User,
 	return user, team, nil
 }
 
+func genEmailInviteWithErrorList(emailList []string) []*model.EmailInviteWithError {
+	var invitesNotSent []*model.EmailInviteWithError
```
mgdelacroix, line 1131: nit: here we can optimize with `invitesNotSent := make([]*model.EmailInviteWithError, len(emailList))`

app/team.go
```diff
@@ -1127,6 +1127,45 @@ func (a *App) prepareInviteNewUsersToTeam(teamId, senderId string) (*model.User,
...
+	for i := range emailList {
+		invite := &model.EmailInviteWithError{
+			Email: emailList[i],
+			Error: model.NewAppError("inviteUsersToTeam", "api.team.invite_members.limit_reached.app_error", map[string]interface{}{"Addresses": emailList[i]}, "", http.StatusBadRequest),
+		}
+		invitesNotSent = append(invitesNotSent, invite)
```
mgdelacroix, line 1137: nit: if the previous optimization is applied, this could be `invitesNotSent[i] = invite`

jwilander:
@mgdelacroix do you know what's up with the translations in the EE test?

PR #15199, opened 2020-08-06
MM-27574: Add Learn More text to Admin Advisor bot message

catalintomai:
#### Summary
Add "Learn More" link and text to Admin. Advisor "Contact us" bot message.

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-27574

#### Related Pull Requests
webapp: https://github.com/mattermost/mattermost-webapp/pull/6088


i18n/en.json
```diff
@@ -2000,23 +2000,23 @@
...
   },
   {
     "id": "api.server.warn_metric.number_of_active_users_400.notification_title",
     "translation": "Integrate SAML 2.0"
   },
   {
     "id": "api.server.warn_metric.number_of_active_users_500.notification_body",
-    "translation": "Mattermost strongly recommends that deployments of over 500 users upgrade to Mattermost Enterprise E20, which offers features such as user management, server clustering, and performance monitoring."
+    "translation": "Mattermost strongly recommends that deployments of over 500 users upgrade to Mattermost Enterprise E20, which offers features such as user management, server clustering, and performance monitoring.\r\n\r\nContacting support sends your contact information to Mattermost, Inc. [Learn more](https://mattermost.com/pl/default-admin-advisory)"
```
mkraft, line 2019: I think this could be phrase in a more compelling way. On the one hand it says we "strongly recommend" and upgrade, but that's not substantiated with any concrete reasons, just a list of features. The risk is we sound self-serving rather than a trusted advisor.

Specifically _why_ do we strongly recommend an upgrade, from the reader's perspective and needs?

catalintomai: @mkraft, there has been a long (and fairly heated) discussion on it, @jasonblais can give more details, but the gist of it for me is that the end-goal for this effort is to drive upgrade.

PR #15207, opened 2020-08-07
MM-27575: Use index hints to prevent index_merge_intersection

agnivade:
Even after specifying all the 3 columns to coerce MySQL into choosing
the right multi-column index, it sometimes uses 2 separate indices
and does an index_merge of them.

This creates problems because the DeleteAt=0 search is essentially
the entire Posts table, and causes a disastrously bad performance
than even choosing the wrong index (idx_create_at).

The problem with this approach is that we hardcode the decision
to a specific index when MySQL was free to choose the right index
depending on table statistics. However, there does not appear to be
a case where this index can cause regressions than using some other index.

Another option here was to set optimizer_switch="index_merge_intersection=off"
at a session level for a transaction and then switch it back on after the
query is done.

However, this can cause some unintentional consequences because this setting
is only available at a session level and not at a query level.
There is no need to set something at the session level when an index hint suffices.

https://mattermost.atlassian.net/browse/MM-27575

### DIFF
--- a/store/sqlstore/post_store.go
+++ b/store/sqlstore/post_store.go
@@ -708,10 +708,18 @@ func (s *SqlPostStore) getPostsAround(before bool, options model.GetPostsOptions
 		direction = ">"
 		sort = "ASC"
 	}
+	table := "Posts p"
+	// We force MySQL to use the right index to prevent it from accidentally
+	// using the index_merge_intersection optimization.
+	// See MM-27575.
+	if s.DriverName() == model.DATABASE_DRIVER_MYSQL {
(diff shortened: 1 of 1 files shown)


agnivade:
@lindy65 - Just a general smoke testing of channel loading related features with special attention towards slowness should be sufficient.

agnivade:
Load tested on a setup starting with ~4M posts and confirmed that the regression is fixed. With 5.25, it is able to barely reach 150 users before the p99 response times flats out at 10 seconds. With the fix, it is able to go back to the normal ~2K users with P99 of 150-200ms.
p99-api-times
avg-api-times
websocket-connections

ethervoid:
Really nice!!

PR #15218, opened 2020-08-10
MM-27508 - Fix slash commands

cpoile:
#### Summary
- slash commands will fail if `PluginId` is null in the database. So set all `PluginId`s to a blank string.
- move the new `PluginId` field to the 5.28 migration (which is when it will appear)

#### Ticket Link
- https://mattermost.atlassian.net/browse/MM-27508


hanzei:
Is there a way to automatically catch these kind of bugs in the future? Should it be caught by the `test-schema` job in CirleCI?

cpoile:
@hanzei Not really, it's not a schema problem, it was a casting error b/c of preexisting rows. My fault for not thinking it through enough.

PR #15221, opened 2020-08-10
[MM-27623] Add new session prop for oauth

fmunshi:
#### Summary
- Looks like https://github.com/mattermost/mattermost-server/pull/14874 introduced a breaking change which conflated the `Session.isOAuth` to mean a session created for an OAuth App _and_ a session created by a user logging through an OAuth provider.

- This PR adds a new prop to sessions `USER_AUTH_SERVICE_IS_OAUTH  = "isOAuthUser"` which is set when a _user_ logs in through an oauth provider, while keeping `Session.IsOauth = false`. 

#### Ticket Link
- https://mattermost.atlassian.net/browse/MM-27623

### DIFF
--- a/app/app_iface.go
+++ b/app/app_iface.go
@@ -461,7 +461,7 @@ type AppIface interface {
 	DoEmojisPermissionsMigration()
 	DoGuestRolesCreationMigration()
 	DoLocalRequest(rawURL string, body []byte) (*http.Response, *model.AppError)
-	DoLogin(w http.ResponseWriter, r *http.Request, user *model.User, deviceId string, isMobile, isOAuth, isSaml bool) *model.AppError
+	DoLogin(w http.ResponseWriter, r *http.Request, user *model.User, deviceId string, isMobile, isOAuthUser, isSaml bool) *model.AppError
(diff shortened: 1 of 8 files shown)


amyblais:
@mkraft Please help prioritize reviews on this,

fmunshi:
> LGTM 👍 One nit: (me *Session) IsOAuthUser() bool could probably use a unit test.

Done, thanks for the review @wiggin77

srkgupta:
Hi @fmunshi 
I am trying to test it on the Test server mentioned above and currently I am seeing the following error when user logs in via Gitlab. 

`Bad response from token request.`

and the server log shows this error:
```
{"level":"error","ts":1597124222.688222,"caller":"web/oauth.go:282","msg":"AuthorizeOAuthUser: Bad response from token request., response_body={\"error\":\"invalid_grant\",\"error_description\":\"The provided authorization grant is invalid, expired, revoked, does not match the redirection URI used in the authorization request, or was issued to another client.```

I verified that the Gitlab OAuth credentials were properly configured. Can you please check if this PR causes this to break?

PR #15228, opened 2020-08-11
MM-27525: Obscure Global Relay's SMTP password.

mkraft:
#### Summary

Updates the Global Relay SMTP password field to be displayed as **************.

#### Ticket Link

https://mattermost.atlassian.net/browse/MM-27525


mkraft:
> Hi @mkraft
> 
> I am seeing the following issue. Can you please fix the same.
> 
> If a user is trying to configure the Global Relay for the first time, the application sets the Global Relay SMTP password as `********************************` and user is required to reset this. The expected behaviour is that the value should be shown as `********************************` only if the config was previously set. If the previous config was empty/null/not set, then this field should also be empty.
> 
> I confirmed the behaviour on other screens, i.e. say OAuth -> Gitlab and the expected behaviour is working fine on that screen, i.e. if the config is not set then it defaults to empty. If the config is set, then it resets the field and shows it as `********************************`.

You're right. I have pushed a fix. Thank you.

srkgupta (approved):
Thanks @mkraft. Tested on the updated PR and the issue is fixed now and is working fine as expected. Approving the PR.

PR #15230, opened 2020-08-12
MM-27507: Propagate rate limit errors to client

agnivade:
We return an error from SendInviteEmails instead of just logging it
to let the client know that a rate limit error has happened.

The status code is chosen as 413 (entity too large) instead of 429 (too many requests)
because it's not the request which is rate limited, but the payload inside it which is.

Ideally, the email sending should have been implemented by a queue which would just return
an error to the client when full. That is also why we are not returning an X-Retry-After
and X-Reset-After in the headers because that would mix with the actual rate limiting.

A separate header X-Email-Invite-Reset-After might do the job, but it comes at an extra cost
of additional API surface and a clunky API. Instead, that information is contained in the error
response. The web client needs to just surface the error. An API client will have to do
a bit more work to parse the error if it needs to automatically know when to retry. Given that
an email sending client is not a very common use case, we decide to keep the API clean.

This decision can be revisited if it becomes problematic in the future.

https://mattermost.atlassian.net/browse/MM-27507


agnivade:
This still needs a client side change. Apparently the webapp simply ignores any error from this endpoint.

agnivade:
Needs to be tested along with https://github.com/mattermost/mattermost-webapp/pull/6123

PR #15238, opened 2020-08-13
MM-27735: Do not build test and other tools in CI

agnivade:
Our codebase uses main_test.go files to control tests at a package level.
However, the problem with that is when we use ./... to build the codebase,
those binaries also get built.

This is unnecessary and creates further delays in the pipeline. All we need to build
are only mattermost and platform binaries in CI. The rest are dependant on the developer.

In my dev environment, this reduces the build times from 13 seconds to 4 seconds.

https://mattermost.atlassian.net/browse/MM-27735


agnivade:
Interesting. From 4m48s to 4m27s. Well, not much, but I'll take it.

agnivade:
This is just a CI change so no QA needed.

PR #15239, opened 2020-08-13
Fix command url validation

iomodo:
#### Summary
This PR will fix command URL validation.
I'm not sure if `scheme` and `host` check should be incorporated in `IsValidHttpUrl` method

#### Ticket Link
  Fixes https://mattermost.atlassian.net/browse/MM-27636


model/command.go
```diff
@@ -109,7 +110,8 @@ func (o *Command) IsValid() *AppError {
 		return NewAppError("Command.IsValid", "model.command.is_valid.url.app_error", nil, "", http.StatusBadRequest)
 	}
 
-	if !IsValidHttpUrl(o.URL) {
+	u, err := url.ParseRequestURI(o.URL)
+	if err != nil || !IsValidHttpUrl(o.URL) || u.Scheme == "" || u.Host == "" {
```
crspeller, line 114: Maybe all these checks should be combined into the `IsValidHttpUrl` function?

model/utils.go
```diff
@@ -496,7 +496,7 @@ func IsValidHttpUrl(rawUrl string) bool {
 		return false
 	}
 
-	if _, err := url.ParseRequestURI(rawUrl); err != nil {
+	if u, err := url.ParseRequestURI(rawUrl); err != nil || u.Scheme == "" || u.Host == "" {
```
crspeller, line 499: A table test for this function would be nice. Not blocking.

prapti (approved):
LGTM!
Creating a new slash command, as well as editing an existing one checks for the correct number of slashes in the url (tested with no slashes and >=< 2 slashes).
Approving.

hanzei:
@iomodo For which release is this change?

iomodo:
It's not tied to a specific release, is not a regression, so merging just on master will work fine I think

PR #15241, opened 2020-08-13
Drain body to reuse tcp connection

yashgandhi-32:
As per docs

    // The HTTP Client and Transport guarantee that Body is always
    // non-nil, even on responses without a body or responses with
    // a zero-length body. It is the caller's responsibility to
    // close Body. The default HTTP client's Transport may not
    // reuse HTTP/1.x "keep-alive" TCP connections if the Body is
    // not read to completion and closed.

`Response.Body` is not read all the way to the EOF as` json.Decoder` is used,
which stops at the end of the object, which is probably followed by a \n.

It's necessary to discard bytes if left so the connection can be reused


services/mailservice/inbucket.go
```diff
@@ -99,7 +100,13 @@ func GetMessageFromMailbox(email, id string) (JSONMessageInbucket, error) {
 	if err != nil {
 		return record, err
 	}
-	defer emailResponse.Body.Close()
+	defer func() {
+		// To reuse same tcp connection
+		// Response body is read and closed.
+
+		io.Copy(ioutil.Discard, emailResponse.Body)
```
agnivade, line 107: Well, we are creating a new HTTP client every time, so actually this code doesn't give us much benefit. To actually reuse the connection, we need to change the `get` function to use `http.Get` rather than construct the client from scratch.

yashgandhi-32: ok I will update the get function, just to be sure you are asking about this `req, err := http.NewRequest("GET", url, nil)` to `http.Get(url)` @agnivade

agnivade: That's correct.

yashgandhi-32: can you check now?

services/mailservice/inbucket.go
```diff
@@ -57,18 +58,19 @@ func GetMailBox(email string) (results JSONMessageHeaderInbucket, err error) {
...
-	req, err := http.NewRequest("GET", url, nil)
+	resp, err := http.Get(url)
+
 	if err != nil {
 		return nil, err
 	}
 
-	client := &http.Client{}
+	defer func() {
+		// To reuse same tcp connection
```
agnivade, line 68: Let's remove the comments. The draining of the body is done in a lot of places.

yashgandhi-32: done

services/mailservice/inbucket.go
```diff
@@ -134,17 +142,10 @@ func downloadAttachment(url string) ([]byte, error) {
 }
 
 func get(url string) (*http.Response, error) {
```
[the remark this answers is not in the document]
yashgandhi-32: done

yashgandhi-32:
@agnivade For CLA I have added in the list `@yashgandhi-32 ` rather than `yashgandhi-32` can you fix it that why cla check is failing

agnivade:
@hanzei should be able to help.

hanzei:
@yashgandhi-32 Could you just  please re-sign the CLA with your correct usernane? Let me know if you have any questions.

PR #15243, opened 2020-08-13
[MM-27274] Dont allow convert to private channel without create private channel permission

fmunshi:
#### Summary
- Prevent team admins without create private channel permission from converting an existing public channel to a private channel

#### Ticket Link
- https://mattermost.atlassian.net/browse/MM-27274


jespino:
@thefactremains I have some doubts here. Are we "creating a private channel"? then... are we "deleting a public channel"?, or maybe are we doing a "Team wide" operation that needs you the "Manage Team" permission?

The initial though was exactly that, we weren't creating or deleting any channel, we were flagging a channel as private. I understand that there is a use-case behind this, and maybe we want to add another permission for this specific use-case, but I don't think "Create Private Channel" is exactly what you are doing.

srkgupta (approved):
Tested on the updated PR and the issue is fixed now and is working fine. The option to convert a public channel to private is not provided if the user does not has permission. Even the API returns a permission denied error.

fmunshi:
Will open a new PR as the new solution is a different story

PR #15246, opened 2020-08-14
Migration of TeamStore Part 1

### LINKED ISSUE #15128: Migrate AppError to error from store/sqlstore/team_store.go part 1
Mattermost is migrating all usages of the `AppError` type to the more idiomatic `error` type. `AppError` is a loaded error type containing fields like HTTP Status code which is not really necessary for errors coming from the DB layer. We should migrate to using normal types satisfying the error interface.

This ticket is about migrating all usages of AppError from `store/sqlstore/team_store.go part 1` to return a normal error. The migration should use the error types defined in `store/errors.go`. If the need should arise to add another specific error type, feel free to add it. For any generic error that is not something like an invalid input, or something not found, we can just wrap the error using `errors.Wrap(f)` and return it.

And then in the app layer, we would need to convert the `error` type to the usual `AppError`. Essentially moving the `AppError` creation from the store layer to the app layer. This means we need a switch case to check for the type of error and appropriately convert to the correct AppError. It can be something like:

```go
if _, err := a.Srv().Store.Bot().Update(bot); err != nil {
	var nfErr *store.ErrNotFound
	var appErr *model.AppError
	switch {
	case errors.As(err, &nfErr):
		return model.MakeBotNotFoundError(nfErr.Id)
	case errors.As(err, &appErr): // in case we haven't converted to plain error.
		return appErr
	default: // last fallback in case it doesn't map to an existing app error.
		return model.NewAppError("DeleteBotIconImage", "app.bot.patchbot.internal_error", nil, err.Error(), http.StatusInternalServerError)
	}
}
```

Also note that the error codes need to be changed from being a store level code (`store.<>.edit.error`) to something like `app.<>.edit.internal_error`.

Example Pull Request: #14339

If you have questions about the ticket or you need any help, feel free to contact `agniva.de.sarker` in https://community.mattermost.com/

----------------------------------

If you're interested please comment here and come [join our "Contributors" community channel](https://community.mattermost.com/core/channels/tickets) on our daily build server, where you can discuss questions with community members and the Mattermost core team. For technical advice or questions, please [join our "Developers" community channel](https://community.mattermost.com/core/channels/developers).

JIRA: https://mattermost.atlassian.net/browse/MM-27321

rvillablanca:
#### Summary

First set of methods in `TeamStore` that return error instead of `*model.AppError`

#### Ticket Link

Fixes #15128


app/team.go
```diff
@@ -700,96 +776,182 @@ func (a *App) JoinUserToTeam(team *model.Team, user *model.User, userRequestorId
...
 func (a *App) GetTeamByInviteId(inviteId string) (*model.Team, *model.AppError) {
-	return a.Srv().Store.Team().GetByInviteId(inviteId)
+	team, err := a.Srv().Store.Team().GetByInviteId(inviteId)
+	if err != nil {
+		var nfErr *store.ErrNotFound
+		switch {
+		case errors.As(err, &nfErr):
+			return nil, model.NewAppError("GetTeamByInviteId", "app.team.get_by_invite_id.finding.app_error", nil, nfErr.Error(), http.StatusNotFound)
+		default:
+			return nil, model.NewAppError("SqlTeamStore.GetByInviteId", "app.team.get_by_invite_id.finding.app_error", nil, err.Error(), http.StatusInternalServerError)
```
agnivade, line 816: `SqlTeamStore.GetByInviteId` seems wrong here.

app/user.go
```diff
@@ -48,7 +48,8 @@ const (
 )
 
 func (a *App) CreateUserWithToken(user *model.User, token *model.Token) (*model.User, *model.AppError) {
-	if err := a.IsUserSignUpAllowed(); err != nil {
+	err := a.IsUserSignUpAllowed()
```
agnivade, line 51: Any reason to break this up? The lines below already use `nErr`.

store/retrylayer/retrylayer.go
```diff
@@ -4776,21 +4776,63 @@ func (s *RetryLayerTeamStore) AnalyticsGetTeamCountForScheme(schemeId string) (i
 
 }
 
-func (s *RetryLayerTeamStore) AnalyticsPrivateTeamCount() (int64, *model.AppError) {
+func (s *RetryLayerTeamStore) AnalyticsPrivateTeamCount() (int64, error) {
 
-	return s.TeamStore.AnalyticsPrivateTeamCount()
+	tries := 0
```
agnivade, line 4781: @jespino - Why does changing the error type generate this code? Shouldn't it be already there?

app/team.go
rvillablanca: Thanks!

app/user.go
rvillablanca: Don't remember but I fixed it now :)

agnivade:
Looks good to me. I am just waiting for @jespino's response to my question.

store/retrylayer/retrylayer.go
jespino: We are only able to retry if this is a plain error, we can't retry an AppError because the AppError doesn't provide the underneath pq or mysql error to verify if it is a repeatable error or not. So, if you don't return a plain error, I just not even try to retry.

agnivade: Ah yes, of course. Thanks.

agnivade (approved):

jespino (approved):
Looks good to me. Great work @rvillablanca, thanks! :tada:

PR #15251, opened 2020-08-14
MM-27512: Use an authenticated user to bump up request rate limit

agnivade:
An unauthenticated user can only make 60 requests per hour which means 1
request every minute. This can lead to frequent rate limit errors while
getting the latest release.

We change that to use an authenticated user which is already available
in the CI. This moves us to make 5000 requests per hour.

We also add additional logging in the Makefile targets in case
the command fails again so that it's clear what has happened, and not return
cryptic 404 errors again.

Ideally, we should be able to inspect the output of the curl command, but since
the output value of the entire bash script is fed into the variable, it is a bit
difficult to print debug output.

If this still gives error, then we need to either use a cached artifact somehow
or add additional logging and add a retry logic on top of it.

https://mattermost.atlassian.net/browse/MM-27512

### DIFF
--- a/scripts/get_latest_release.sh
+++ b/scripts/get_latest_release.sh
@@ -29,7 +29,7 @@ THIS_BRANCH=$(git rev-parse --abbrev-ref HEAD)
 if [[ $(echo "$THIS_BRANCH" | grep -c ^"$BRANCH_TO_USE") == 1 ]];
 then
     VERSION_REL=${THIS_BRANCH//$BRANCH_TO_USE/v}
-    REL_TO_USE=$(curl --silent --user $BASIC_AUTH "https://api.github.com/repos/$REPO_TO_USE/releases" | grep '"tag_name":' | sed -E 's/.*"([^"]+)".*/\1/' | sed -n "/$VERSION_REL/p" | sort -rV | head -n 1)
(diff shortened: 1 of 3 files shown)


scripts/get_latest_release.sh
```diff
@@ -7,17 +7,31 @@
...
 THIS_BRANCH=$(git rev-parse --abbrev-ref HEAD)
 
 if [[ $(echo "$THIS_BRANCH" | grep -c ^"$BRANCH_TO_USE") == 1 ]];
 then
     VERSION_REL=${THIS_BRANCH//$BRANCH_TO_USE/v}
-    REL_TO_USE=$(curl --silent "https://api.github.com/repos/$REPO_TO_USE/releases" | grep '"tag_name":' | sed -E 's/.*"([^"]+)".*/\1/' | sed -n "/$VERSION_REL/p" | sort -rV | head -n 1)
+    REL_TO_USE=$(curl --silent --user $BASIC_AUTH "https://api.github.com/repos/$REPO_TO_USE/releases" | grep '"tag_name":' | sed -E 's/.*"([^"]+)".*/\1/' | sed -n "/$VERSION_REL/p" | sort -rV | head -n 1)
 else
-    REL_TO_USE=$(echo "$THIS_BRANCH" | sed "s/\(.*\)/$LATEST_REL/")
+    REL_TO_USE=$LATEST_REL
```
agnivade, line 34: The sed command seemed to be redundantly used. I simplified it. Let me know if it was intentional somehow.

jaydeland: You are correct - thank you for optimizing this.

PR #15262, opened 2020-08-14
[MM-27836] Fix possible panic when patching config

streamer45:
#### Summary

PR fixes two instances where sending an `nil` `SiteURL` could cause a panic when patching the config.

#### Ticket

https://mattermost.atlassian.net/browse/MM-27836

### DIFF
--- a/api4/config.go
+++ b/api4/config.go
@@ -79,7 +79,7 @@ func updateConfig(c *Context, w http.ResponseWriter, r *http.Request) {
 	}
 
 	appCfg := c.App.Config()
-	if *appCfg.ServiceSettings.SiteURL != "" && *cfg.ServiceSettings.SiteURL == "" {
+	if *appCfg.ServiceSettings.SiteURL != "" && (cfg.ServiceSettings.SiteURL == nil || *cfg.ServiceSettings.SiteURL == "") {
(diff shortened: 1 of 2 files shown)


api4/config.go
```diff
@@ -79,7 +79,7 @@ func updateConfig(c *Context, w http.ResponseWriter, r *http.Request) {
 	}
 
 	appCfg := c.App.Config()
-	if *appCfg.ServiceSettings.SiteURL != "" && *cfg.ServiceSettings.SiteURL == "" {
+	if *appCfg.ServiceSettings.SiteURL != "" && (cfg.ServiceSettings.SiteURL == nil || *cfg.ServiceSettings.SiteURL == "") {
```
ashishbhate, line 82: How is it that `cfg.ServiceSettings.SiteURL` can be `nil` here? On line 74 we call `cfg.SetDefaults` which calls `ServiceSettings.SetDefaults` which ensures `SiteURL` wont be nil.

streamer45: I forgot to update the method and saw the test fail but you are right, only the patch has the issue. Let me update.

amyblais:
@streamer45 Kind note to cherry-pick this,

PR #15271, opened 2020-08-16
MM-27883 - remove deep copies from the cache layer

agnivade:
When we are reading from or putting to the cache, there is no need to deep copy
objects now because the cache is already serialized.

We also delete some lines from tests because the mock store directly returns the pointers
whereas actually they would be returned from the database where a serialization occurs again.

So we would be testing for the wrong thing by unnecessarily keeping the deep copies for DB reads.

https://mattermost.atlassian.net/browse/MM-27883


store/localcachelayer/user_layer_test.go
```diff
@@ -91,11 +91,6 @@ func TestUserStoreCache(t *testing.T) {
 
 		for i := 0; i < len(storedUsers); i++ {
 			assert.Equal(t, storedUsers[i].Id, cachedUsers[i].Id)
-			if storedUsers[i] == cachedUsers[i] {
```
jespino, line 94: This was testing that the object returned wasn't a pointer to the same struct. I think that still makes sense. I would expect it to work, but the cache always should return a different instance, and I think is what this is checking.

agnivade: Yes, that was the intention, but the test was wrong. I have explained it in the PR commit message. The issue is that the first call to the cache would miss the cache and would cause the item to be returned from the DB. In a real world, that would return a different struct, but here we are using the mocked store, which would return the same pointer.

The code was working so far because we were doing a copy before returning inspite of it being returned from the DB. But that's not really needed. We could still keep the test, but we need to do the deep copy ourselves. So then IMO the test loses its meaning if we are changing the logic ourselves. To do it properly, we would need to use a real DB, but that is going a bit out of scope.

I could still be wrong though somewhere. Please let me know.

jespino (approved):

srkgupta:
Hi @jespino  Looks like this a technical improvement which needs a regression testing. Please confirm. 

-

jespino:
@srkgupta yes, this is super hard to test explicitly, we have pass some of this kind of deep changes without QA (after @lindalumitchell approval) to test it as part of the community-daily regular use and release testing process.

PR #15274, opened 2020-08-17
MM-27887: set traceback to crash from code

agnivade:
Unless we set traceback level to crash, goroutines from all threads aren't printed.
See: https://github.com/golang/go/issues/13161.

This is important because SIGQUIT traces will miss on goroutines.

https://mattermost.atlassian.net/browse/MM-27887

### DIFF
--- a/cmd/mattermost/commands/server.go
+++ b/cmd/mattermost/commands/server.go
@@ -7,6 +7,7 @@ import (
 	"net"
 	"os"
 	"os/signal"
+	"runtime/debug"
 	"syscall"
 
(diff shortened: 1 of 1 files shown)


wiggin77:
https://github.com/golang/go/issues/13161 mentions that after outputting the stacks it aborts when GOTRACEBACK=crash.  I couldn't find any mention of that being changed.  Do you know if this is still the behaviour?

agnivade:
This change will cause the service to crash with a coredump instead of just dumping the stack trace to console.

PR #15275, opened 2020-08-17
Restrict the sampledata generation to not run without enough users for group channels

jespino:
#### Summary
Restrict the sampledata generation to not run without enough users for group channels

#### Ticket Link
[MM-27886](https://mattermost.atlassian.net/browse/MM-27886)


jespino:
@grundleborg added test cases and improved a bit the tests.

PR #15290, opened 2020-08-17
Add AWS Metering service support

catalintomai:
#### Summary
Add AWS Metering service support. Every hour the number of active users for the last hour metric is being sent to AWS Metering service. 

Note: 
1) The code is intended to be run on EC2 instances only (with the MM_SERVICESETTINGS_ENABLEAWSMETERING and AWS_REGION env variables set).
1) This is not fully-tested code (lacking AWS-provided product code for full end-to-end testing).


app/app.go
```diff
@@ -103,6 +103,10 @@ func (a *App) InitServer() {
 		if a.Srv().runjobs {
 			a.Srv().Go(func() {
 				runLicenseExpirationCheckJob(a)
+
+				if *a.Config().ServiceSettings.EnableAWSMetering {
```
jespino, line 107: Should this be exposed as a config settings, shouldn't be enough an environment variable? If this is a config, what happens if the user disable it?

What happens if I change that setting through the API in runtime (Should we monitor config updates?)

app/aws_meter.go
```diff
@@ -0,0 +1,147 @@
+// Copyright (c) 2015-present Mattermost, Inc. All Rights Reserved.
```
jespino, line 1: Is there any reason to add this to the app layer? Would it makes sense to move it into a service and pass the pointer to the store during the `newAWSMeterService` call? The rest of the code looks like not related to the app at all, and I don't see any dependency.

store/sqlstore/user_store.go
```diff
@@ -1255,7 +1255,31 @@ func (us SqlUserStore) AnalyticsActiveCount(timePeriod int64, options model.User
 
 	v, err := us.GetReplica().SelectInt(queryStr, args...)
 	if err != nil {
-		return 0, model.NewAppError("SqlUserStore.AnalyticsDailyActiveUsers", "store.sql_user.analytics_daily_active_users.app_error", nil, err.Error(), http.StatusInternalServerError)
+		return 0, model.NewAppError("SqlUserStore.AnalyticsActiveCount", "store.sql_user.analytics_daily_active_users.app_error", nil, err.Error(), http.StatusInternalServerError)
+	}
+	return v, nil
+}
+
+func (us SqlUserStore) AnalyticsActiveCountForPeriod(startTime int64, endTime int64, options model.UserCountOptions) (int64, *model.AppError) {
```
jespino, line 1263: I miss some tests around the new method.

jespino: We are trying to return plain errors from the store now, do you mind to return `error` instead of `*model.AppError`, and replace the `model.NewAppError` to regular errors? Because all of them are considered `InternalServerErrors` you only have to return plain standar errors, can be just `errors.Wrap(err, "error description")` (probably you can see examples of this in other stores.

app/aws_meter.go
jespino: I would move it to `services/awsmeter` and rename the `newAWSMeterService` to just `New` method receiving the `store`.

app/aws_meter.go
```diff
@@ -0,0 +1,147 @@
...
+// See LICENSE.txt for license information.
+
```
jespino, line 3: Is there any way to add tests to this service without a ton of effort mocking things? maybe for the `getUserCategoryUsage`?

app/app.go
jespino: I don't feel super strong about it, my concern is to expose something to the user that is not needed to be exposed to the user, that settings is going to be shown in the `config/config.json` file. Not a big deal anyway.

We are not going to expose that in the admin console anyway, so I think we can keep as is now and change it later if we see any drawback.

app/aws_meter.go
jespino: Thanks

store/sqlstore/user_store.go
jespino: Thanks

model/system.go
```diff
@@ -23,6 +24,8 @@ const (
 	SYSTEM_WARN_METRIC_NUMBER_OF_ACTIVE_USERS_200 = "warn_metric_number_of_active_users_200"
 	SYSTEM_WARN_METRIC_NUMBER_OF_ACTIVE_USERS_400 = "warn_metric_number_of_active_users_400"
 	SYSTEM_WARN_METRIC_NUMBER_OF_ACTIVE_USERS_500 = "warn_metric_number_of_active_users_500"
+	AWS_METERING_REPORT_INTERVAL                  = 3
```
jespino, line 27: This constants looks like is not used anywhere, I think it should be used to avoid magic numbers (I think now we are using `1*time.Hour` and I guess it should be `model.AWS_METERING_REPORT_INTERVAL*time.Hour`

jespino: other option is to remove this constant.

app/server.go
```diff
@@ -1094,6 +1095,24 @@ func runLicenseExpirationCheckJob(a *App) {
 	}, time.Hour*24)
 }
 
+func runReportToAWSMeterJob(a *App) {
```
jespino, line 1098: This can be done receiving only the server, and instanciating it as part of the `app/server.go` `Server.RunJobs` method. This way you are coupling less this feature with the `App` layer, and only coupling it to the server layer.

app/server.go
```diff
@@ -1094,6 +1095,24 @@ func runLicenseExpirationCheckJob(a *App) {
...
+	awsMeter := awsmeter.New(a.srv.Store, a.Config())
+	if awsMeter == nil {
+		mlog.Error("Cannot obtain instance of AWS Metering Service.")
+		return
+	}
+
+	dimensions := []string{model.AWS_METERING_DIMENSION_USAGE_HRS}
+	reports := awsMeter.GetUserCategoryUsage(dimensions, time.Now().UTC(), time.Now().Add(-1*time.Hour).UTC())
+	awsMeter.ReportUserCategoryUsage(reports)
+}
```
jespino, line 1114:
```suggestion
func runReportToAWSMeterJob(s *Server) {
	model.CreateRecurringTask("Collect and send usage report to AWS Metering Service", func() {
		doReportUsageToAWSMeteringService(s)
	}, time.Hour*1)
}

func doReportUsageToAWSMeteringService(s *Server) {
	awsMeter := awsmeter.New(s.Store, s.Config())
	if awsMeter == nil {
		mlog.Error("Cannot obtain instance of AWS Metering Service.")
		return
	}

	dimensions := []string{model.AWS_METERING_DIMENSION_USAGE_HRS}
	reports := awsMeter.GetUserCategoryUsage(dimensions, time.Now().UTC(), time.Now().Add(-1*time.Hour).UTC())
	awsMeter.ReportUserCategoryUsage(reports)
}
```

jespino: This also need to move the `runReportToAWSMeterJob` call from `app/app.go` file to the `app/server.go` file.

model/system.go
```diff
@@ -155,3 +158,14 @@ func SendWarnMetricAckFromJson(r io.Reader) *SendWarnMetricAck {
 	json.NewDecoder(r).Decode(&swma)
 	return swma
 }
+
+type AWSMeterReport struct {
```
jespino, line 162: Looks like this model is not relevant to any other part of the application, should we move it into the service itself? And looks like is not serialized anywhere, so maybe we can remove the json annotations and the `ToJSON`.

services/awsmeter/awsmeter.go
```diff
@@ -0,0 +1,146 @@
...
+	return &AwsMeter{
+		store:   store,
+		service: svc,
+		config:  config,
+	}
+}
+
+func newAWSMarketplaceMeteringService() (*marketplacemetering.MarketplaceMetering, error) {
+	region := os.Getenv("AWS_REGION")
+	s := session.Must(session.NewSession(&aws.Config{Region: &region}))
```
jespino, line 59: This `session.Must` can panic, let's use `s, err := session.NewSession` instead.

services/awsmeter/awsmeter.go
```diff
@@ -0,0 +1,146 @@
...
+	reports := make([]*model.AWSMeterReport, 0)
+
+	for _, dimension := range dimensions {
+		var userCount int64
+		var err error
+
+		switch dimension {
+		case model.AWS_METERING_DIMENSION_USAGE_HRS:
+			userCount, err = awsm.store.User().AnalyticsActiveCountForPeriod(model.GetMillisForTime(startTime), model.GetMillisForTime(endTime), model.UserCountOptions{})
+			mlog.Info("GetUserCategoryUsage", mlog.Int64("usercount", userCount), mlog.Err(err), mlog.Bool("bool", err != nil))
```
jespino, line 90: This is a debug log leaved there? if this is a regular log entry, that is there to inform the user that a dimension was reported, probably we can do that as an `else` of the `if err != nil` and provide a better message, excluding the `mlog.Err` and the `mlog.Bool`. If was for debugging purposes, just get rid of it.

services/awsmeter/awsmeter.go
```diff
@@ -0,0 +1,146 @@
...
+			if err != nil {
+				mlog.Error("Failed to obtain usage data", mlog.String("dimension", dimension), mlog.String("start", startTime.String()), mlog.Int64("count", userCount), mlog.Err(err))
```
jespino, line 92: This is adding this log entry here, and something almost duplicated in the `if err != nil` after the switch, maybe we can just `return` or `continue` after that.

services/awsmeter/awsmeter.go
```diff
@@ -0,0 +1,146 @@
...
+			}
+		default:
+			mlog.Error("Dimension does not exist!", mlog.String("dimension", dimension))
```
jespino, line 95: Same here, this message and the one after the switch is pretty similar data, probably is easier to just log and `continue` or `return`.

services/awsmeter/awsmeter.go
```diff
@@ -0,0 +1,146 @@
...
+			err = errors.New("Dimension does not exist")
+		}
+
+		if err != nil {
+			mlog.Error("Failed to obtain usage.", mlog.String("dimension", dimension), mlog.Err(err))
+			return reports
```
jespino, line 101: This return here is going to cut the execute of the rest of the dimensions (for now works perfectly fine, because we only have a dimension, but in the future a failure in one dimension stop the report for every dimension after that, but sends the report for any dimension before that.

The current behavior is one option, but I think is the less consistent. I prefer or "all or nothing" approach, or the "all that I can report" approach, that means, If it fails for 1 dimension, lets send the data for the rest of the dimensions that works.

services/awsmeter/awsmeter.go
```diff
@@ -0,0 +1,146 @@
...
+			Timestamp: startTime,
+		}
+
+		reports = append(reports, report)
+	}
+
+	return reports
+}
+
+func (awsm *AwsMeter) ReportUserCategoryUsage(reports []*model.AWSMeterReport) *model.AppError {
```
jespino, line 116: Because this is a service, maybe makes way more sense to use plain errors instead of `AppError`, actually, in this case the error itself is not exposed to the user in any way, so can just use regular errors and everything should work out of the box.

jespino: This applies to all the `AppError`s used in the whole service.

services/awsmeter/awsmeter.go
```diff
@@ -0,0 +1,146 @@
...
+
+	resp, err := ams.AwsMeteringSvc.MeterUsage(params)
+	if err != nil {
+		return model.NewAppError("sendReportToMeteringService", "app.system.aws_metering_service.error", nil, err.Error(), http.StatusNotFound)
+	}
+	if resp.MeteringRecordId == nil {
+		return model.NewAppError("sendReportToMeteringService", "app.system.aws_metering_service.error", nil, "", http.StatusNotFound)
+	}
+
+	mlog.Debug("Sent record to AWS metering service", mlog.String("dimension", report.Dimension), mlog.Int64("value", report.Value), mlog.String("timestamp", report.Timestamp.String()))
```
jespino, line 143: I guess we want to keep this. Sounds sensible to have this debug message during development and the Debug log level is ok :)

jespino (changes requested):
The code in general is great, I added some small suggestions for improvement.

model/system.go
catalintomai: Yes, it is was intended to be used in the 1*time.Hour, got lost in some refactoring.

app/server.go
catalintomai: Sure - though at a certain point, as I was mentioning in a previous comment, might make sense to spend a bit of time and capture these guidelines so that you or somebody else from the server team don't have to repeat yourselves when people don't follow the recommended patterns.

app/server.go
catalintomai: sure, same comment as above.

model/system.go
catalintomai: I prefer to keep the ToJSON, I find it useful in debugging and I don't have to add and then remove it.

services/awsmeter/awsmeter.go
catalintomai: That looks to be the prevalent pattern used in the aws-sdk but I'm 0/5 on this.

services/awsmeter/awsmeter.go
catalintomai: agree, to be removed.

services/awsmeter/awsmeter.go
catalintomai: Yes, the whole method actually had initially another interior loop I got rid off (maybe for different time intervals?), hence the inconsistencies observed below.  My bad.

services/awsmeter/awsmeter.go
catalintomai: same comment as above.

services/awsmeter/awsmeter.go
catalintomai: agree, same comment as above.

services/awsmeter/awsmeter.go
catalintomai: sure, see comment above wrt to guidelines/etc. for this area of the codebase.

app/server.go
jespino: I going to raise that at the next server meeting, probably we should be capturing this kind of feedback from the pull requests and adding to some place (not only here, but almost as part of any PR). Thanks.

model/system.go
jespino: Sounds good to me

services/awsmeter/awsmeter.go
jespino: Yes, looks like that are in the code examples in the documentation, but looks to me like a bad practice, as part of our policy we never panic, and that code leads to possible panics. Anyway, you already changed it :)

go.sum
```diff
@@ -845,6 +849,7 @@ golang.org/x/net v0.0.0-20190613194153-d28f0bde5980/go.mod h1:z5CRVTTTmAJ677TzLL
 golang.org/x/net v0.0.0-20190620200207-3b0461eec859/go.mod h1:z5CRVTTTmAJ677TzLLGU+0bjPO0LkuOLi4/5GtJWs/s=
 golang.org/x/net v0.0.0-20190827160401-ba9fcec4b297/go.mod h1:z5CRVTTTmAJ677TzLLGU+0bjPO0LkuOLi4/5GtJWs/s=
 golang.org/x/net v0.0.0-20190923162816-aa69164e4478/go.mod h1:z5CRVTTTmAJ677TzLLGU+0bjPO0LkuOLi4/5GtJWs/s=
+>>>>>>> master
```
jespino, line 852: Bad merge here

services/awsmeter/awsmeter.go
```diff
@@ -0,0 +1,155 @@
...
+
+func (o *AWSMeterReport) ToJSON() string {
+	b, _ := json.Marshal(o)
+	return string(b)
+}
+
+func New(store store.Store, config *model.Config) *AwsMeter {
+	svc := &AWSMeterService{
+		AwsDryRun:      false,
+		AwsProductCode: "12345", //TODO
```
jespino, line 51: I guess this TODO is here because we are waiting to have the valid code for this in the future, right?

services/awsmeter/awsmeter.go
```diff
@@ -0,0 +1,155 @@
...
+	}))), nil
+}
+
+// a report entry is for all metrics
+func (awsm *AwsMeter) GetUserCategoryUsage(dimensions []string, startTime time.Time, endTime time.Time) []*AWSMeterReport {
+	reports := make([]*AWSMeterReport, 0)
+
+	for _, dimension := range dimensions {
+		var userCount int64
+		var err error
```
jespino, line 99: nit: this variable is no longer needed at this scope, you can move it in the scope of `case model.AWS_METERING_DIMENSION_USAGE_HRS`

app/server.go
```diff
@@ -1142,6 +1147,24 @@ func runLicenseExpirationCheckJob(a *App) {
...
+
+func doReportUsageToAWSMeteringService(s *Server) {
+	awsMeter := awsmeter.New(s.Store, s.Config())
+	if awsMeter == nil {
+		mlog.Error("Cannot obtain instance of AWS Metering Service.")
+		return
+	}
+
+	dimensions := []string{model.AWS_METERING_DIMENSION_USAGE_HRS}
+	reports := awsMeter.GetUserCategoryUsage(dimensions, time.Now().UTC(), time.Now().Add(-1*time.Hour).UTC())
```
jespino, line 1164: should `-1*time.Hour` be `-time.Hour*model.AWS_METERING_REPORT_INTERVAL`?

jespino (changes requested):
One small bad merge problem, and some tiny details and doubts.

go.sum
catalintomai: should be fixed now

services/awsmeter/awsmeter.go
catalintomai: yes, this is pretty much not fully-baked code, might suffer further changes once the full e2e testing is performed.

services/awsmeter/awsmeter.go
catalintomai: moved it down.

app/server.go
catalintomai: good point, to be consistent.

jespino (approved):
Now looks good to me. Thanks!

srkgupta (approved):
As discussed with Catalin, this is not testable yet. Hence approving the PR as is. Regression will run with the daily cypress tests

PR #15292, opened 2020-08-18
MM-27916: Improve logging when session is not found

agnivade:
The error handling in this code is pretty bad and the same error message happens
for multiple conditions, making it difficult to diagnose the real issue.

Most of the times, we get a log like:

```
"Invalid session        {"error": "GetSession: Invalid session token=jodb6sau47rnugaqj1fy7khmpr, err=<no value>, "}"
```

And it could have happened from multiple places. So this log turns out to be not that useful.

We improve this by populating the Error field to fix the "<no value>" issue and also add a separate detailed error
field for each log line to uniquely identify each error.


mgdelacroix (approved):

agnivade:
@lindalumitchell - Only some internal logging change. No externally visible change.

PR #15294, opened 2020-08-18
MM-27613: Update dependencies

agnivade:
https://mattermost.atlassian.net/browse/MM-27613


grundleborg:
QA Requirements: general basic smoke test of the system. @agnivade any particular areas that you suspect might have problems due to these updates, or does nothing look concerning?

agnivade:
Nothing stands out in particular. Mostly transitive dependencies. Of importance, we should test Sentry and OpenTracing.

PR #15309, opened 2020-08-19
MM-27948: Use --depth=1 to clone git repos faster

agnivade:
https://mattermost.atlassian.net/browse/MM-27948

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -22,7 +22,7 @@ jobs:
       - checkout
       - run: |
           cd ../
-          git clone --depth=1 https://github.com/mattermost/mattermost-webapp.git
+          git clone --depth=1 --no-single-branch https://github.com/mattermost/mattermost-webapp.git
           cd mattermost-webapp
           git checkout $CIRCLE_BRANCH || git checkout master
           export WEBAPP_GIT_COMMIT=$(git rev-parse HEAD)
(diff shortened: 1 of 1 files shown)


.circleci/config.yml
```diff
@@ -22,7 +22,7 @@ jobs:
       - checkout
       - run: |
           cd ../
-          git clone https://github.com/mattermost/mattermost-webapp.git
+          git clone --depth=1 --no-single-branch https://github.com/mattermost/mattermost-webapp.git
```
khos2ow, line 25: Why did you add `--no-single-branch` which in reality fetches the near tip of all branches? Is it needed to have the extra information available?

.circleci/config.yml
```diff
@@ -134,7 +134,7 @@ jobs:
           at: ~/mattermost/
       - run:
           command: |
-            git clone https://github.com/mattermost/mattermost-api-reference.git
+            git clone --depth=1 --no-single-branch https://github.com/mattermost/mattermost-api-reference.git
```
khos2ow, line 137: The same here.

.circleci/config.yml
agnivade: It does not fetch the tip of all branches by default. My initial PR did not have "--no-single-branch" and that failed to fetch the other branch in TE and EE repos respectively.

Here's what the man page says:

>  --depth <depth>
           Create a shallow clone with a history truncated to the specified number of commits. **Implies --single-branch unless --no-single-branch is given to fetch the histories near the tips of all branches**.

Emphasis mine.

khos2ow: Yes, I think my question was confusing. What I'd like to understand is that do you actually need the extra histories near the tip of other branches on the CI?

agnivade: Because we are checking out other branches than master. `git checkout $CIRCLE_BRANCH || git checkout master`. I have not done this in other places where we don't need to check out other branches.

khos2ow: :+1:

PR #15313, opened 2020-08-20
MM-22786 enterprise metrics for logging

wiggin77:
#### Summary
This PR adds metrics for logging engine which are collected by Prometheus.

- current queue level(s)
- rate of logging records emitted
- rate of logging errors

Companion PR in Enterprise:  https://github.com/mattermost/enterprise/pull/745

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-22786


app/server.go
```diff
@@ -483,6 +483,14 @@ func NewServer(options ...Option) (*Server, error) {
 		}
 	}
 
+	if s.Metrics != nil && allowAdvancedLogging {
```
ethervoid, line 486: I have one question about `allowAdvancedLogging`. If the user starts with a community/e0 edition and adds a license, are we reevaluating that flag? I mean we have license listeners [here](https://github.com/mattermost/mattermost-server/blob/master/app/server.go#L279) that checks that I think

ethervoid (commented):
Just one comment

app/server.go
wiggin77: Not currently.  It appears the metrics service gets started after a config change here:  https://github.com/mattermost/mattermost-server/blob/c424be882c18225466f03da2aee6ff8ab968ebf5/app/config.go#L408  I could enable logger metrics there as well (I think).  However, I don't see how this works since metrics service will only be started if `Server.Metrics` is non-nil, and that only gets set in `initEnterprise` which does not seem to be called when the license changes.

ethervoid: Maybe I'm totally lost but I'm seeing [here](https://github.com/mattermost/mattermost-server/blob/master/app/server.go#L479) that we're shutting down the AdvancedLogging if that flag is false and that flag depends on the features of the license right? So if the license change that should be re-evaluated?

wiggin77: That is true... whether advanced logging is enabled could be re-evaluated on license change.  But whether metrics is enabled for logging is dependent also on whether metrics is enabled and that doesn't seem to happen on license change, unless I missed where that happens.

ethervoid: In [this](https://github.com/mattermost/mattermost-server/blob/master/app/server.go#L290) point it gets initialized and at [this point](https://github.com/mattermost/mattermost-server/blob/master/app/server.go#L495) is started but if the configuration changes it's reevaluated to stop or start if needed. Is that what you're asking?

wiggin77: Thanks @ethervoid  I think my confusion may be around whether Server.Metric is non-nil for an unlicensed server.  I assumed it was, but I guess I must be wrong, otherwise metrics would require a server restart when the license is added.  I also thought initEnterprise did nothing without a license, but I guess it always executes if you have an enterprise capabile build.

ethervoid: For what I've read it's always get's initialized but not always is started. It just gets started if it has the license on start or if the configuration changes.

> otherwise metrics would require a server restart when the license is added

No need, I think, having the license listener if something change we can start or stop the metrics server

wiggin77: I have created a ticket to have advanced logging enabled on license upgrade.  That work is out of scope for this PR.  Thanks for pointing this out Mario.   https://mattermost.atlassian.net/browse/MM-28031

wiggin77: I also made it so that logging metrics get enabled on license upgrade.

ethervoid: Great thank you!!!!

app/server.go
```diff
@@ -617,6 +623,18 @@ func (s *Server) initLogging() error {
 	return nil
 }
 
+func (s *Server) enableLoggingMetrics() {
+	if s.Metrics == nil {
+		return
```
mkraft, line 628: This is an expected path for TE, right? That's why we don't need to log it?

wiggin77: Correct

wiggin77:
@lindalumitchell since this is for tracking metrics via Prometheus/Grafana, it may be that QA would pass on testing this.  Would like to get this approved either way relatively soon as there are multiple PR's (Enterprise and Server) that need to ber merged at the same time.

PR #15316, opened 2020-08-20
MM-27918 In-Product notices support

reflog:
#### Summary
This PR introduces support for In-Product Notices as described in https://docs.google.com/document/d/1LiEuv95fws__sjULbDkLvzj83U3yH_s2zk5FnMkd9tk/edit

It contains the following:

1. A new table storing 'viewed' state for each userid/noticeid pair
2. New job-server task that once an hour does the following:
 a) fetches new notices from an external server (waiting for DevOps to bring it up, in the meantime - uses public github repo. I'm leaving Do-Not-Merge label until the actual host is up and running)
 b) cleans the 'viewed' state for notices that are no longer existing
 c) calculates user and post count and stores them, since they are used for filtering notices
3. New API endpoints to 
 a) fetch notices appropriate for user
 b) allow user to mark notice as viewed
4. Two new settings AdminNoticesEnabled and UserNoticesEnabled , set to true by default

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-27918

### Related PRs
enterprise -https://github.com/mattermost/enterprise/pull/763
webapp - https://github.com/mattermost/mattermost-webapp/pull/6426


vendor/modules.txt
```diff
@@ -397,6 +400,9 @@ github.com/prometheus/common/model
 github.com/prometheus/procfs
 github.com/prometheus/procfs/internal/fs
 github.com/prometheus/procfs/internal/util
+# github.com/reflog/dateconstraints v0.2.1
+## explicit
+github.com/reflog/dateconstraints
```
jespino, line 405: You need to add the changes to include in the NOTICE.txt file the licenses that applies to this new packages included. This one and the `Masterminds/semver/v3` one.

app/product_notices.go
```diff
@@ -0,0 +1,260 @@
...
+var noticesCache = utils.RequestCache{}
+
+// cached counts that are used during notice condition validation
+var cachedPostCount int64
+var cachedUserCount int64
+
+// previously fetched notices
+var cachedNotices model.ProductNotices
+
+func noticeMatchesConditions(a *App, client model.NoticeClientType, clientVersion, locale string, postCount, userCount int64, isSystemAdmin, isTeamAdmin bool, isCloud bool, sku string, notice *model.ProductNotice) (bool, error) {
```
ashishbhate, line 36: `locale` is passed to this function but doesnt appear to be used.

model/product_notices.go
```diff
@@ -0,0 +1,200 @@
...
+	err := json.Unmarshal(data, &r)
+	return r, err
+}
+
+// List of product notices. Order is important and is used to resolve priorities.
+// Each notice will only be show if conditions are met.
+type ProductNotice struct {
+	Conditions        Conditions                       `json:"conditions"`
+	ID                string                           `json:"id"`                   // Unique identifier for this notice. Can be a running number. Used for storing 'viewed'; state on the server.
+	LocalizedMessages map[string]NoticeMessageInternal `json:"localizedMessages"`    // Notice message data, organized by locale.; Example:; "localizedMessages": {; "en-US": { "title": "English", description: "English description"},; "fr-FR": { "title": "Frances", description: "French description"}; }
```
ashishbhate, line 29: Here we've documented that the keys are in the format `en-US` but in the code we've used `enUS`

app/product_notices.go
```diff
@@ -0,0 +1,260 @@
...
+			&notice)
+		if err != nil {
+			return nil, model.NewAppError("GetProductNotices", "api.system.update_notices.validating_failed", nil, err.Error(), http.StatusBadRequest)
+		}
+		if result {
+			selectedLocale := "enUS"
+			filteredNotices = append(filteredNotices, model.NoticeMessage{
+				NoticeMessageInternal: notice.LocalizedMessages[selectedLocale],
+				ID:                    notice.ID,
+			})
```
ashishbhate, line 221: Doesn't this mean the returned filtered notices will all be in `enUS` instead of the requested locale?

app/product_notices.go
reflog: @ashishbhate  correct. in this phase - locale is ignored

app/product_notices.go
reflog: @ashishbhate correct. in this phase - locale is ignored

ashishbhate (commented):
Looks good, some questions about locale inline.

model/product_notices.go
reflog: i'll update the comment

ashishbhate (approved):
LGTM after comment update.

app/product_notices.go
jespino, line 36: Looks like you are receiving an `*App` parameter but you are not using it.

app/product_notices.go
```diff
@@ -0,0 +1,260 @@
...
+	}
+	if vt.Kind() == reflect.Ptr {
+		vt = vt.Elem()
+	}
+	val := vt.Interface()
+	return val == expectedValue
+}
+
+func (a *App) GetProductNotices(lastViewed int64, userId, teamId string, client model.NoticeClientType, clientVersion string, locale string) (model.NoticeMessages, *model.AppError) {
+	isSystemAdmin := a.SessionHasPermissionTo(*a.Session(), model.PERMISSION_MANAGE_SYSTEM)
```
jespino, line 154: I don't think this is strictly wrong, but I don't know if this is strictly right either. This is saying that any user that has permissions to manage the system is a system admin, and any user with permissions to manage a team is a team admin. And that is kinda right. But maybe is bit more "right" to say that any user with the "system_admin" role is a system admin, and any user with the "team_admin" role is a team admin. From the notifications point of view, maybe makes more sense one or the other. probably is a question for PMs. The current approach is perfectly fine, but I think the other approach too.

app/product_notices.go
```diff
@@ -0,0 +1,260 @@
...
+			if v.NoticeId == nid {
+				return &v
+			}
+		}
+		return nil
+	}
+
+	for _, notice := range cachedNotices {
+		// check if the notice has been viewed already
+		view := getViewState(notice.ID)
```
jespino, line 188: this inline function looks a bit more complex to understad for me than a regular inline block, somethig like:

```suggestion
		var view *model.ProductNoticeViewState
		for _, v := range views {
			if v.NoticeId == nid {
				view = &v
				break
			}
		}
```

But maybe is matter of taste.

app/product_notices.go
```diff
@@ -0,0 +1,260 @@
...
+				continue
+			}
+			if repeatable && view.Viewed > MAX_REPEAT_VIEWINGS {
+				continue
+			}
+			if view.Timestamp < lastViewed {
+				continue
+			}
+		}
+		notice := notice // pin
```
jespino, line 201: This feels confusing, why not to use another variable?

config/utils.go
```diff
@@ -180,3 +181,52 @@ func JSONToLogTargetCfg(data []byte) (mlog.LogTargetCfg, error) {
 	}
 	return cfg, nil
 }
+
+func GetConfigValueByPath(config *model.Config, path []string) (interface{}, bool) {
```
jespino, line 185: Why this wrapper function here?

jespino (changes requested):
Some tiny details to discuss, but in general looks good to me.

app/product_notices.go
reflog: No idea which is the righter one :) Pinging @esethna

config/utils.go
reflog: @jespino - we didn't have a function for getting config values by dotted path, so I've copied one from mmctl

jespino: Yes, still, I don't think the wrapper is providing any value.

app/product_notices.go
reflog: @jespino *App is being used by validateConfigEntry function

jespino: o yes, I didn't see that, but probably you only need to pass the current config, right? not the whole App

reflog: yup. i'll fix that for clarity

app/product_notices.go
esethna: For the purposes of notices we want to look at the user role. ie any user with the "system_admin" role is a system admin, and any user with the "team_admin" role is a team admin @reflog

esethna:
Thanks for the PR @reflog, whats our plan for testing so we can get it merged?

reflog:
@jespino - anything else blocking the PR?

jespino:
@reflog nothing blocking, sorry the delay

jespino (approved):

esethna:
@reflog are we good to merge this?

reflog:
@esethna - not yet. The notices hosting is still unavailable.

app/product_notices.go
```diff
@@ -33,7 +34,7 @@ var cachedUserCount int64
 // previously fetched notices
 var cachedNotices model.ProductNotices
 
-func noticeMatchesConditions(conf *model.Config, client model.NoticeClientType, clientVersion, locale string, postCount, userCount int64, isSystemAdmin, isTeamAdmin bool, isCloud bool, sku string, notice *model.ProductNotice) (bool, error) {
+func noticeMatchesConditions(app AppIface, userId string, client model.NoticeClientType, clientVersion, locale string, postCount, userCount int64, isSystemAdmin, isTeamAdmin bool, isCloud bool, sku string, notice *model.ProductNotice) (bool, error) {
```
jespino, line 37: I would prefer to pass the conf here, and pass an `store.Store` parameter too. Passing the app you are passing too much logic, and binding this to the app layer without explicit need.

jespino (approved):
An improvement suggestion, up to you to cover that, but I think looks better with less coupling with the App layer.

ogi-m (approved):
**QA approving for merge.**

We verified in-product notices for: 
- All User Roles: Sysadmin, TeamAdmin, Member and All
- Client Type (desktop/web)
- Server Versions (various values)
- Desktop Version (various values)
- Repeatable
- Priority of notices (ordered by json placement)
- SKU

^ and some combinations of the conditions  

Existing Issues:

- Teamadmin notice preference setting (filed new issue for followup)
- Date Bounds - we only tested for today and received a white screen and console error, @reflog is looking into it
- We added few observations/questions to webapp PR from @sudheerDev concerning hand pointer on circle indicators and <previous button

cc @jgilliam17 @esethna

jgilliam17 (approved):

esethna (approved):

PR #15320, opened 2020-08-20
[MM-27535] Change string to be more in-line with designs and the front end verbiage

nickmisasi:
#### Summary
Matt requested this string be changed. 
#### Ticket Link
https://mattermost.atlassian.net/browse/MM-27535


go.tools.mod
```diff
@@ -4,7 +4,7 @@ go 1.14
 
 require (
 	github.com/jstemmer/go-junit-report v0.9.1 // indirect
-	github.com/mattermost/mattermost-utilities/mmgotool v0.0.0-20200814100246-5fc14ed48292 // indirect
```
devinbinnie, line 7: Not sure if this is the same case as `package-lock.json` where it doesn't need to be committed?
Or is this an explicit upgrade?

jwilander: It's fine to commit 👍

PR #15321, opened 2020-08-21
MM-23063: Fix flaky test TestTermsOfServiceStore

agnivade:
The call to testGetLatestTermsOfService would happen after testSaveTermsOfService
which would persist the data between calls. Therefore, if they were to happen under
a milisecond, there would be 2 rows with the same CreateAt timestamp and the DB
would randomly return any row.

If this were to happen, then the wrong row would not match the user id and would fail.

To fix this, we just clear the table data before proceeding with the test.

https://mattermost.atlassian.net/browse/MM-23063

### DIFF
--- a/store/storetest/terms_of_service_store.go
+++ b/store/storetest/terms_of_service_store.go
@@ -13,25 +13,26 @@ import (
 )
 
 func TestTermsOfServiceStore(t *testing.T, ss store.Store) {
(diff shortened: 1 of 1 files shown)


mgdelacroix (approved):
Awesome! Thanks for making the changes! :)

PR #15324, opened 2020-08-21
MM-27722 Make sidebar category validation silently reject bad channel IDs

hmhealey:
I think we've cleaned up all the ways that a user might get into a state where they have bad channel IDs already existing in a sidebar category, except for the brief period on the community cluster where community didn't know to update the sidebar tables.

Regardless, this makes the validation a bit less annoying so that it should stop causing people from being able to move sidebar channels if it does get into a weird state again. Instead of failing with an error and causing a moved channel to revert back to its original position, this'll just remove any bad channel IDs silently without bothering the player. I did leave in a log message if that case occurs for future debugging, but I don't expect it to happen often.

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-27722


reflog (approved):
LGTM

amyblais:
@hmhealey Claudio is off this week. Perhaps another dev can review this?

hmhealey:
@amyblais Thanks for the heads up! Re-assigned.

mgdelacroix (approved):
Amazing! Thanks for applying the change to all the tests! :rocket:

PR #15327, opened 2020-08-22
MM-27744 Unit tests using too many file handles

wiggin77:
#### Summary
Zap has no concept of shutdown or close.  Zap is only shutdown when the app exits.  Not a problem for console logging, but when creating new Zap logger outputting to files on every unit test, that leaves no easy way to clean up until process exit.

To work around this all files being opened are tracked and a shutdown method to mlog.Logger was added.  A call to `shutdown` was added to all the places that a new logger was created. 

With this PR I was not able to reproduce the "too many files" error after multiple runs.  Before this PR I was able to reproduce it with VMware and VSCode open, which use many handles.

This PR is based on https://github.com/mattermost/mattermost-server/pull/15313 to avoid merge conflict.

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-27744


wiggin77:
Closing in favour of https://github.com/mattermost/mattermost-server/pull/15398

PR #15331, opened 2020-08-22
MM-28043:NoRows error now being returned, handle in callers

sbishel:
#### Summary
The store was changed and now returns a NoRows error rather than eating it. The calling functions need updated in order to handle that error when returned.

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-28043

#### Related PR
https://github.com/mattermost/enterprise/pull/750


sbishel:
I had commented on that in the enterprise repo -
> There is also a call in worker.go. I originally made the same change there. However, after testing, realized that both paths work the same way in worker.go, so I reverted those changes.

PR #15332, opened 2020-08-23
[MM-30562] Indicate deleteBy in post_deleted websocket event

dalcde:
This lets the client know who deleted the post. This information is stored in the database, but cannot be retrieved from an API endpoint because the API does not return information about deleted post.

The motivation for the change is to improve bridges to other chat platforms, which want to know who performed the deletion, but I can see this being useful in general.
```release-note
In post_deleted websocket event, notify system administrators of the user initiating the deletion.
```

#### Ticket

https://mattermost.atlassian.net/browse/MM-30562

### DIFF
--- a/api4/post_test.go
+++ b/api4/post_test.go
@@ -1993,6 +1993,56 @@ func TestDeletePost(t *testing.T) {
 	CheckNoError(t, resp)
 }
 
+func TestDeletePostMessage(t *testing.T) {
+	th := Setup(t).InitBasic()
+	th.LinkUserToTeam(th.SystemAdminUser, th.BasicTeam)
+	th.App.AddUserToChannel(th.SystemAdminUser, th.BasicChannel)
+
+	defer th.TearDown()
+
+	testCases := []struct {
+		description string
+		client      *model.Client4
+		delete_by   interface{}
+	}{
+		{"Do not send delete_by to regular user", th.Client, nil},
+		{"Send delete_by to system admin user", th.SystemAdminClient, th.SystemAdminUser.Id},
+	}
+
+	for _, tc := range testCases {
+		t.Run(tc.description, func(t *testing.T) {
+			wsClient, err := th.CreateWebSocketClientWithClient(tc.client)
+			require.Nil(t, err)
+			defer wsClient.Close()
+
+			wsClient.Listen()
+
+			post := th.CreatePost()
+
+			status, resp := th.SystemAdminClient.DeletePost(post.Id)
+			require.True(t, status, "post should return status OK")
+			CheckNoError(t, resp)
+
+			timeout := time.After(5 * time.Second)
+
+			for {
+				select {
+				case ev := <-wsClient.EventChannel:
+					if ev.EventType() == model.WEBSOCKET_EVENT_POST_DELETED {
+						assert.Equal(t, tc.delete_by, ev.GetData()["delete_by"])
+						return
+					}
+				case <-timeout:
+					// We just skip the test instead of failing because waiting for more than 5 seconds
+					// to get a response does not make sense, and it will unncessarily slow down
+					// the tests further in an already congested CI environment.
+					t.Skip("timed out waiting for event")
+				}
+			}
+		})
+	}
+}
+
 func TestGetPostThread(t *testing.T) {
 	th := Setup(t).InitBasic()
 	defer th.TearDown()
--- a/app/post.go
+++ b/app/post.go
@@ -1041,9 +1041,18 @@ func (a *App) DeletePost(postId, deleteByID string) (*model.Post, *model.AppErro
 		}
 	}
 
-	message := model.NewWebSocketEvent(model.WEBSOCKET_EVENT_POST_DELETED, "", post.ChannelId, "", nil)
-	message.Add("post", a.PreparePostForClient(post, false, false).ToJson())
-	a.Publish(message)
+	postData := a.PreparePostForClient(post, false, false).ToJson()
+
+	userMessage := model.NewWebSocketEvent(model.WEBSOCKET_EVENT_POST_DELETED, "", post.ChannelId, "", nil)
+	userMessage.Add("post", postData)
+	userMessage.GetBroadcast().ContainsSanitizedData = true
+	a.Publish(userMessage)
+
+	adminMessage := model.NewWebSocketEvent(model.WEBSOCKET_EVENT_POST_DELETED, "", post.ChannelId, "", nil)
+	adminMessage.Add("post", postData)
+	adminMessage.Add("delete_by", deleteByID)
+	adminMessage.GetBroadcast().ContainsSensitiveData = true
+	a.Publish(adminMessage)
 
 	a.Srv().Go(func() {
 		a.DeletePostFiles(post)


streamer45:
Hi @dalcde , thank you for your contribution :+1: 

I am adding @thefactremains to the review as I'd like his thoughts on this. Mainly I'd like to understand if there is any concern with sending this type of information out by default or if it makes sense to enable this through a configuration setting instead.

On a more technical note, I'd really like to see a couple of tests that cover this case. Would you be willing to add those?

dalcde:
On Sun, Aug 30, 2020 at 11:23:03PM -0700, Claudio Costa wrote:
>On a more technical note, I'd really like to see a couple of tests that cover this case. Would you be willing to add those?

I would be happy to write some tests, but I would rather do it once the
change itself is approved. A bit of guidance would also be helpful, as
I'm not sure where the test would belong. Grepping the codebase seems to
indicate this event is not tested anywhere.

streamer45:
> I would be happy to write some tests, but I would rather do it once the change itself is approved. A bit of guidance would also be helpful, as I'm not sure where the test would belong. Grepping the codebase seems to indicate this event is not tested anywhere.

Sure. As far as the test, you can take a look at:

https://github.com/mattermost/mattermost-server/blob/c6b20ab133416762ab222e402d1dfb021f77cea2/api4/post_test.go#L589

Let me know if that helps.

thefactremains:
This seems useful for sure. Thank you for the contribution! 

If I'm understanding correctly, this would reveal to all members of the channel who deleted a post. If so, this needs to be opt-in and not enabled by default.

agnivade:
I am not sure if adding yet another config setting for such a thing is the right way forward. Perhaps we should just take a decision and go with it. I am 0/5 on either, but a bit wary on adding a config setting for this. Has there been any precedent in the past where we added a config setting for adding a prop to a websocket event?

dalcde:
An alternative solution would be to provide an option to query metadata of deleted posts, which includes the deleter in props. This also makes a bit more sense from the design point of view, because currently the only way you can possibly receive this information is if you happen to have a websocket connection when the deletion happens.

This could also be useful for auditing purposes

streamer45:
> I am not sure if adding yet another config setting for such a thing is the right way forward. Perhaps we should just take a decision and go with it. I am 0/5 on either, but a bit wary on adding a config setting for this. Has there been any precedent in the past where we added a config setting for adding a prop to a websocket event?

I agree, the config setting was probably not the best example. The main point here is that we'd like to avoid surprises for existing users which might not expect such information to be included by default. Maybe it makes more sense to add a permission for this?

> An alternative solution would be to provide an option to query metadata of deleted posts, which includes the deleter in props. This also makes a bit more sense from the design point of view, because currently the only way you can possibly receive this information is if you happen to have a websocket connection when the deletion happens.

Agree on this as well. Although, I think we might want both things to make it work as expected. That's how we deal with posts at the moment. We fetch from the API when loading a channel and after that we wait for websocket events. If we included the information only in the HTTP call then we might miss it on newly (deleted) posts.

jfrerich:
Hi all!  There seems to be some good design discussion. Was there a final decision so the author can continue?

agnivade:
Agree that we should make it work both ways (using API and websocket events). I am wondering if we could piggyback an existing role to achieve this? I am thinking that either the channel_admin (for the channel to which the posts belong), or the team_admin, or the sysadmin should be able to view the deleteBy field.

dalcde:
Making it available to system_admin seems like a good and least
controversial starting point to me; the information is in the database
anyway. We can make it available to other roles in the future.

On the API part, I think that would be a completely separate change
that lets certain users read about deleted posts; the deleter is in the
props.

agnivade:
Removing it from my queue until a decision is made.

streamer45:
@wiersgallak 

I am adding you to the discussion since Dennis is currently unavailable. 
We are looking for some PM guidance on this before moving forward with the implementation.

wiersgallak:
@dalcde @streamer45 I agree that we should limit this to the permission of the system_admin.

dalcde:
Is there a good way to broadcast a message to the administrators in the
channel only? (and similarly, broadcast to non-administrators only so
that administrators don't receive the event twice)

streamer45:
> Is there a good way to broadcast a message to the administrators in the channel only? (and similarly, broadcast to non-administrators only so that administrators don't receive the event twice)

I believe broadcast are sent at a team level then it's up to the client to deal with them.
As far as broadcasting an event only to admins/non admins it should be easily doable. You can take a look at [`shouldSendEvent`](https://github.com/mattermost/mattermost-server/blob/14fe307436b1fdb6e26c9c3fd8db5ca5bf24d432/app/web_conn.go#L327-L347) for an example.

dalcde:
I've updated the PR to only send the deleter to administrators.

agnivade:
We should add some tests for this.

And the rest API should also be updated. If you prefer to send a separate PR, that's also fine. But both of them should be checked in at the same time. It's good to always keep the websocket API and rest API in sync with each other.

dalcde:
The REST API would be a rather different story. Currently you cannot
retrieve *any* information about deleted posts whatsoever; if it were
possible, the delete_by field is already present.

agnivade:
I see. Thanks for that context. So all we need is a test for this. Please take a look at https://github.com/mattermost/mattermost-server/pull/15332#issuecomment-683629363 to get an understanding of how it's done.

dalcde:
Pushed.

api4/post_test.go
```diff
@@ -1993,6 +1993,54 @@ func TestDeletePost(t *testing.T) {
...
+	require.Nil(t, err)
+	defer wsClient.Close()
+
+	wsClient.Listen()
+
+	wsAdminClient, err := th.CreateWebSocketClientWithClient(th.SystemAdminClient)
+	require.Nil(t, err)
+	defer wsAdminClient.Close()
+
+	wsAdminClient.Listen()
```
agnivade, line 2013: Let's create two sub-tests for this.

api4/post_test.go
```diff
@@ -1993,6 +1993,54 @@ func TestDeletePost(t *testing.T) {
...
+
+	waitForEvent := func(client *model.WebSocketClient, delete_by interface{}) {
```
agnivade, line 2015: Was `delete_by` made to be `interface{}` because the map value is `interface{}`? Wondering if we could just pass `*string` and then check if value exists or not in the map if it's nil?

agnivade (commented):
Looking good overall. Just some minor comments.

api4/post_test.go
dalcde: I originally tried something along these lines but I decided on a more uniform version of the code. I also think it is better to use the "correct" types for the assertion

api4/post_test.go
dalcde: Done

api4/post_test.go
agnivade: Ok.

agnivade (approved):

api4/post_test.go
```diff
@@ -1993,6 +1993,56 @@ func TestDeletePost(t *testing.T) {
...
+			require.True(t, status, "post should return status OK")
+			CheckNoError(t, resp)
+
+			timeout := time.After(5 * time.Second)
+
+			for {
+				select {
+				case ev := <-wsClient.EventChannel:
+					if ev.EventType() == model.WEBSOCKET_EVENT_POST_DELETED {
+						assert.Equal(t, ev.GetData()["delete_by"], tc.delete_by)
```
streamer45, line 2032:
```suggestion
						assert.Equal(t, tc.delete_by, ev.GetData()["delete_by"])
```

The expected value (`tc.delete_by` in this case) should be passed to `assert.Equal()` before than the actual value. This will avoid getting unclear/confusing messages in case the check were to fail.

streamer45 (approved):
Nicely done @dalcde :+1: Happy to see this finally get in.
I just left one suggestion.

dalcde:
Done and rebased as well.

lindy65 (approved):
Thanks @dalcde @streamer45 - looks good to me 👍

PR #15335, opened 2020-08-24
cleanup error logging from user_store

lieut-data:
#### Summary
Tracing this back, it looks like I'm responsible for leaving some spurious logging in the user store.

#### Ticket Link
None.


lieut-data:
Skipping PM/QA review for what amounts to a logging change.

ethervoid (approved):
LGTM!

PR #15338, opened 2020-08-25
Fix tests for no linux systems

jespino:
#### Summary
Fix tests for no linux systems


reflog (approved):

hmhealey:
I assume we're only supporting this feature on Linux OSes for now then?

jespino:
@hmhealey yes, for now, but the thing is, we are only publishing the linux binary in our download page, so unless that changes in the future, this is going to be the unique way to do it.

PR #15341, opened 2020-08-25
[MM-27919] Allow i18n and resources to be loaded using an environment variable

mgdelacroix:
#### Summary
This modifies the test resource loading and the `i18n` files to be loaded from a dynamic location using an environment variable. This mechanism will be used by `mmctl` to start the server reusing the same test helpers that we have for the API to run end to end tests and to be able to use the app layer within them.

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-27919


jespino (approved):
LGTM

testlib/resources.go
```diff
@@ -36,17 +36,26 @@ type testResourceDetails struct {
...
+	paths := []string{
+		".",
+		"..",
+		"../..",
+		"../../..",
+		"../../../..",
+	}
+
+	// this enables the server to be used in tests from a different repository
+	if mmPath := os.Getenv("MM_SRC_PATH"); mmPath != "" {
```
jespino, line 50: Maybe "MM_SERVER_PATH"? or something less ambiguous?

mgdelacroix: makes sense, modified!

PR #15343, opened 2020-08-25
Add functionality to update password with password hash

gabrieljackson:
This can be used to update a user's password with a direct password
hash instead of providing the password in plaintext.

Notes:
 - Tested this with a very crude mmctl alteration
 - Can be used directly over the normal API or via `local` mode. This will be used for cloud user onboarding.

Fixes https://mattermost.atlassian.net/browse/MM-28106


api4/user_test.go
```diff
@@ -2806,6 +2806,26 @@ func TestUpdateUserPassword(t *testing.T) {
 	CheckNoError(t, resp)
 }
 
+func TestUpdateUserHashedPassword(t *testing.T) {
+	th := Setup(t).InitBasic()
+	defer th.TearDown()
+
+	password := "SuperSecurePass23!"
+	passwordHash := "$2a$10$CiS1iWVPUj7rQNdY6XW53.DmaPLsETIvmW2p0asp4Dqpofs10UL5W"
+	pass, resp := th.SystemAdminClient.UpdateUserHashedPassword(th.BasicUser.Id, passwordHash)
```
mgdelacroix, line 2815: this test can be wrapped using the `th.TestForSystemAdminAndLocal` helper so it checks local mode too

mgdelacroix (changes requested):
Requested one change to include local mode in testing. Otherwise looks good!

model/client4.go
```diff
@@ -1186,6 +1186,17 @@ func (c *Client4) UpdateUserPassword(userId, currentPassword, newPassword string
 	return CheckStatusOK(r), BuildResponse(r)
 }
 
+// UpdateUserHashedPassword updates a user's password with an already-hashed password. Must be a system administrator.
+func (c *Client4) UpdateUserHashedPassword(userId, newHashedPassword string) (bool, *Response) {
+	requestBody := map[string]string{"already_hashed": "true", "new_password": newHashedPassword}
```
mgdelacroix, line 1191: Here we need to include an empty `current_password` for the endpoint not to complain
```suggestion
	requestBody := map[string]string{"already_hashed": "true", "new_password": newHashedPassword, "current_password": ""}
```

mgdelacroix: Said nothing, I was making the wrong request... :man_facepalming:

api4/user_test.go
gabrieljackson: Awesome, thanks for the tip!

mgdelacroix (approved):
LGTM!! :+1:

PR #15356, opened 2020-08-27
[MM-28083] CWS one-time login logic

ethervoid:
#### Summary

This PR adds the needed logic to perform a transparent login when the installation process has finished and we want to redirect the user to their new installation without entering credentials

#### Reference link

https://github.com/mattermost/mattermost-api-reference/pull/571

#### Ticket Link

https://mattermost.atlassian.net/browse/MM-28083


api4/user.go
```diff
@@ -1720,6 +1730,35 @@ func login(c *Context, w http.ResponseWriter, r *http.Request) {
...
+		parameters["token"] = props["token"]
+		parameters["device_id"] = props["device_id"]
+		parameters["ldap_only"] = props["ldap_only"]
+	}
+
+	return parameters
+}
+
+func isCWSLogin(c *Context, cwsToken string) bool {
+	return *c.App.Config().ServiceSettings.EnableCWSTokenLogin && cwsToken != ""
```
jwilander, line 1759: We should also check that the license feature `Cloud` is `true` too. Actually we could just check the license flag and not use a config setting at all, what do you think?

ethervoid: If we're going to have this feature available for all the cloud installations would be better to use it yes

web/handlers.go
```diff
@@ -299,6 +299,12 @@ func (h *Handler) checkCSRFToken(c *Context, r *http.Request, token string, toke
 
 	if csrfCheckNeeded {
 		csrfHeader := r.Header.Get(model.HEADER_CSRF_TOKEN)
+		if csrfHeader == "" && *c.App.Config().ServiceSettings.EnableCWSTokenLogin {
```
ethervoid, line 302: This concerns me a bit @jupenur . I've added it because in case the user from another domain tries to enter again using the same CWS route (form and so on) gets a session error because is not able to fetch de CSRF token from the cookie. Do you see anything weird/wrong?

api4/user.go
```diff
@@ -1670,11 +1670,16 @@ func login(c *Context, w http.ResponseWriter, r *http.Request) {
...
 	c.LogAuditWithUserId(id, "attempt - login_id="+loginId)
 
-	user, err := c.App.AuthenticateUserForLogin(id, loginId, password, mfaToken, ldapOnly)
+	user, err := c.App.AuthenticateUserForLogin(id, loginId, password, mfaToken, cwsToken, ldapOnly)
 	if err != nil {
 		c.LogAuditWithUserId(id, "failure - login_id="+loginId)
-		c.Err = err
-		return
+		if isCWSLogin(c, cwsToken) {
+			http.Redirect(w, r, *c.App.Config().ServiceSettings.SiteURL, 302)
```
ethervoid, line 1677: This redirects to reload the login page if not the user gets a JSON API response with the error

api4/user.go
jwilander: All installations will need it so yeah let's do that then

jwilander: That would also prevent it being used on-prem

web/handlers.go
jupenur: Yes, this is effectively disables the anti-CSRF measures for instances that set `EnableCWSTokenLogin`. The solution is to split the autlogin into its own REST endpoint and use `api.ApiHandlerTrustRequester` for the handler.

ethervoid:
@jupenur and @jwilander changes addressed

app/login.go
```diff
@@ -55,6 +58,26 @@ func (a *App) AuthenticateUserForLogin(id, loginId, password, mfaToken string, l
...
+			return nil, errSessions
+		}
+		if len(sessions) > 0 {
+			return nil, model.NewAppError("AuthenticateUserForLogin",
+				"api.user.login_by_cws.invalid_token.app_error", nil, "", http.StatusBadRequest)
+		}
+		token, ok := os.LookupEnv(cwsTokenEnv)
+		if ok && cwsToken != "" && token == cwsToken {
+			return user, nil
+		}
```
jwilander, line 78: Might be a good idea to check if the token is set first before checking if they're a bot, pulling the sessions, etc. otherwise we'll be doing all that extra work every time someone logs in normally, right?

ethervoid: Yep you're right fixed checking the cws token in the `IsCWSLogin` method

app/login.go
```diff
@@ -55,6 +58,26 @@ func (a *App) AuthenticateUserForLogin(id, loginId, password, mfaToken string, l
...
+	// installation for the first time
+	if IsCWSLogin(a, cwsToken) {
+		if err = checkUserNotBot(user); err != nil {
+			return nil, err
+		}
+		sessions, errSessions := a.GetSessions(user.Id)
+		if errSessions != nil {
+			return nil, errSessions
+		}
+		if len(sessions) > 0 {
```
jupenur, line 71: What's the purpose of this check? To make sure that this is the first login?  If so, woudn't the CWS login get re-enabled if a user simply logs out of all of their sessions? That's probably not what we want.

app/login.go
```diff
@@ -55,6 +58,26 @@ func (a *App) AuthenticateUserForLogin(id, loginId, password, mfaToken string, l
...
+			return nil, model.NewAppError("AuthenticateUserForLogin",
+				"api.user.login_by_cws.invalid_token.app_error", nil, "", http.StatusBadRequest)
+		}
+		token, ok := os.LookupEnv(cwsTokenEnv)
+		if ok && token == cwsToken {
```
jupenur, line 76: I think we need to use [`subtle.ConstantTimeCompare`](https://golang.org/pkg/crypto/subtle/#ConstantTimeCompare) here to avoid a timing side-channel.

jupenur (changes requested):

app/login.go
ethervoid: Yeah that’s the idea. We have two ways of check that is the food login: sessions or status ‘LastActivityAt’ property. The status logic is optional using a flag so I ended up using sessions.

You’re right that if the user logs out you can enter again this way but that is a very small opportunity window. Yesterday talking with @gabrieljackson I was thinking on adding a another env bar with the expiration for the token, something like ‘CWS_CLOUD_TOKEN_EXPIRATION’ si after that time the token is not anymore. What do you think? // @jwilander

jupenur: The `loginId` (and thus `user`) comes from CWS, right? So an attacker who gains access to the CWS token could use it to login to any account that doesn't have currently active sessions? That's not good.

ethervoid: Yeah but that’s the same problem that you have if someone gains access to the personal token for example right?

jupenur: Yeah, all other tokens are tied to a specific account, whereas this one gives universal access.

ethervoid: Do think using an env var for the login_id should be enough? If not then we should move to use mmctl, modify it so we can pass a generated token, because we need to be have the token and mmctl doesn’t return the he enraged token,  limit  that to one account and then remove the token once is used

jupenur: I don't have a preference on the mechanism, be it env vars or mmctl, but I'd really like to see the token implementation be truly one-time instead of something close to it but not quite. And if the token is truly one-time then we don't need to care about the login ID either.

ethervoid: The only two ways I can think of are:

- generate a token and insurer using mmctl, then remove/expire it after the use
- have three env. vars (token, login_id, expiration) in the installation

Do you see any problem in any of them? If no then we’re going to decide which option to use

jupenur: I'd be happier with explicit removal rather than time-based expiration. With a small enough expiration window I won't say no to the time-based option either, but there's a risk there.

ethervoid: Ok then thank you very much for your help Juho. Then we should talk @jwilander, @gabrieljackson and me about this :)

jwilander: @ethervoid @jupenur could we do something like once the token is used here, store it in the MM database (the general purpose Tokens store would work) as "used" and don't allow login again with a "used" token? That should effectively make it one-time use

I'd prefer that instead of having mmctl generate it or making the CWS remove the env var since that introduces an extra hop which will be more complexity and more likely to fail

ethervoid: Yes, that should do the trick. Didn't suggest that given that is kind of a hidden token flow and different of what we are using with the other tokens but I was working on something similar so all that I've done work for that solution too.

I'll suggest to use 64 characters token to match the other type of tokens

ethervoid: Changes done and tested. I've added some tests too

jwilander:
@jupenur @mgdelacroix mind taking another look? We'd like to get this in ASAP

@ethervoid is there anything @stevemudie can test here or do we need to wait until we have it all hooked up with the CWS?

ethervoid:
Yes we should test for regression in the login: user/password, ldap, saml.... the cws login should be tested all together with cws

mgdelacroix (approved):
LGTM! :+1:

jupenur (approved):

PR #15370, opened 2020-08-31
[MM-27274] Add permission to convert public channel to private

fmunshi:
#### Summary
- Adds permission `convert_public_channel_to_private` and uses it to enforce converting a public channel to private
- Adds permission `convert_private_channel_to_public`
- These permissions have been default added to roles that contain `manage_team` which ensures the existing behaviour will not change since `team_admin` and `system_admin` roles had that permission before and now only those two roles will have this new permission by default.

#### Ticket Link
- https://mattermost.atlassian.net/browse/MM-27274

#### Related Pull Requests
- Has web changes: https://github.com/mattermost/mattermost-webapp/pull/6337
- Has redux changes: https://github.com/mattermost/mattermost-redux/pull/1226


srkgupta (changes requested):
Hi @fmunshi 

I am seeing the following issues while testing this PR:

1. "Convert channels" option for private channels has no significance, because from the chat facing side, the user is not allowed to change the private channel to public. We should remove this option from the Permissions page. Even in the comments on JIRA ticket, @michaelgamble had suggested only to add the Public to Private permissions.  

2. Can you please add a cypress test to verify this new permission.

fmunshi:
@srkgupta Re: MM-28274 - Convert channel permission should not be shown for private channels
- I have kept both permissions in the database but the UI only shows one (Convert public channels to private) toggling this on and off toggles both of the new permissions as per discussion with @thefactremains 

Re: MM-28275 - "Unable to get role" error is shown when permissions are reset & admin tries to save permission
- I have made comments on the ticket, but to reiterate this is a caching issue after running the cobra command the server needs to be restarted (not just purge cache in the UI) 

- I have added E2E tests on the webapp PR

srkgupta (approved):
Thanks @fmunshi. Tested the updated PR along with its dependent PR and the feature is working fine as expected. Approving the PR.

PR #15371, opened 2020-08-31
[MM-28196] Log message if ExperimentalSettings is on, but no interface

hahmadia:
**Needs to be cherry-picked into 5.25 / 5.26 / 5.27**

#### Summary

Log message if Experimental Settings is on, but interface not set and revert to using the old SAML library (xmlsec1)

#### Ticket Link

https://mattermost.atlassian.net/browse/MM-28196

**Requires The Following PR to work!!**
https://github.com/mattermost/enterprise/pull/753


srkgupta (approved):
Tested the changes and now the application ignores the experimental setting and works fine with the old xmlsec library. Approving the PR.

PR #15378, opened 2020-09-01
Update the get release script for mmctl to get the latest release eve…

jaydeland:
#### Summary
Update mmctl get logic for master builds

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-28205


gabrieljackson:
@jaydeland, looks like there may be an unrelated build issue here?

jwilander:
Should be fixed with https://github.com/mattermost/mattermost-server/pull/15380 @gabrieljackson

gabrieljackson:
If I understand our new mmctl pull-in process correctly, we should cut a pre-release in the mmctl with a `-rc` suffix before merging this?

jaydeland:
> If I understand our new mmctl pull-in process correctly, we should cut a pre-release in the mmctl with a `-rc` suffix before merging this?

Correct

PR #15380, opened 2020-09-02
Add redux build for when webapp needs to be built in server pipeline

metanerd:
#### Summary
When webapp is still building the latest master commit, the appropriate commit is not found as upload in the s3 bucket, when at the same time the server pipeline runs. That's when the trap gets triggered, but it missed the redux building steps and erred on those. This PR adds those steps.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -28,7 +28,7 @@ jobs:
           export WEBAPP_GIT_COMMIT=$(git rev-parse HEAD)
           echo "$WEBAPP_GIT_COMMIT"
 
-          trap 'npm ci && make build' ERR
+          trap 'npm ci && cd node_modules/mattermost-redux && npm i && npm run build && cd - && make build' ERR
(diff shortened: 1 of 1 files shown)


agnivade:
We are going back and forth multiple times on this one. I changed the `cd` code in https://github.com/mattermost/mattermost-shared-pipelines/pull/76 for exactly this. The `npm prepare` script in mattermost-redux already contains the building code. So is it still necessary to explicitly build mattermost-redux? Doing npm ci should already build mattermost-redux. Or perhaps there is a difference between npm ci and npm install with regard to building transitive dependencies? Not sure if that's the issue.

In any case, to be consistent, we should change https://github.com/mattermost/mattermost-shared-pipelines/pull/76 again to add the cd logic.

metanerd:
Npm ci does build mattermost-redux, but it seems only halfway, because there are some errors produced with missing mattermost-redux/actions. I wanted to add npm ci for building mattermost-redux, turns out npm ci needs package-lock.json, which it doesn't have with webapps call to npm ci . Then I talked with @hmhealey about packaging in npm in general and we came to the consensus that we need to look more deeply into it. You are right, I will create a PR to add it to jenkins. Thank you!

agnivade:
I see. Thanks!

PR #15396, opened 2020-09-06
TeamStore migration part 2

rvillablanca:
#### Summary

Migration of `TeamStore` to return plain errors.

#### Ticket Link

Fixes https://github.com/mattermost/mattermost-server/issues/15127

#### Jira Ticket

https://mattermost.atlassian.net/browse/MM-27320


app/team.go
```diff
@@ -353,13 +353,19 @@ func (a *App) GetSchemeRolesForTeam(teamId string) (string, string, string, *mod
...
+		case errors.As(nErr, &nfErr):
+			return nil, model.NewAppError("UpdateTeamMemberRoles", "app.team.get_member.missing.app_error", nil, nfErr.Error(), http.StatusNotFound)
+		default:
+			return nil, model.NewAppError("UpdateTeamMemberRoles", "app.team.get_member.app_error", nil, nErr.Error(), http.StatusInternalServerError)
+		}
 	}
 
 	if member == nil {
-		err = model.NewAppError("UpdateTeamMemberRoles", "api.team.update_member_roles.not_a_member", nil, "userId="+userId+" teamId="+teamId, http.StatusBadRequest)
+		err := model.NewAppError("UpdateTeamMemberRoles", "api.team.update_member_roles.not_a_member", nil, "userId="+userId+" teamId="+teamId, http.StatusBadRequest)
```
agnivade, line 368: Let's return the error directory.

agnivade (approved):
Just one small comment. Great work

app/team.go
rvillablanca: Solved!

store/sqlstore/team_store.go
```diff
@@ -1098,20 +1098,20 @@ func (s SqlTeamStore) GetActiveMemberCount(teamId string, restrictions *model.Vi
...
 	}
 
 	return count, nil
 }
 
-func (s SqlTeamStore) GetMembersByIds(teamId string, userIds []string, restrictions *model.ViewUsersRestrictions) ([]*model.TeamMember, *model.AppError) {
+func (s SqlTeamStore) GetMembersByIds(teamId string, userIds []string, restrictions *model.ViewUsersRestrictions) ([]*model.TeamMember, error) {
 	if len(userIds) == 0 {
-		return nil, model.NewAppError("SqlTeamStore.GetMembersByIds", "store.sql_team.get_members_by_ids.app_error", nil, "Invalid list of user ids", http.StatusInternalServerError)
+		return nil, model.NewAppError("SqlTeamStore.GetMembersByIds", "app.team.get_members_by_ids.app_error", nil, "Invalid list of user ids", http.StatusInternalServerError)
```
jespino, line 1114: Why is this still returning an AppError instead of a plain error, or a custom built error?

jespino (changes requested):
A tiny detail, but otherwise looks good to me. Thanks @rvillablanca! :tada:

store/sqlstore/team_store.go
rvillablanca: Good catch!

rvillablanca: I think I missed it because it is returning 500 when validating an input parameter, in those cases a 400 http status is mor appropriate but that would change the the implementation and could break something so, maybe I could keep it as 500 but using a plain error instead of `store.ErrInvalidInput`. what do you think @jespino ?

/cc @agnivade

agnivade: Yeah .. a simple `errors.New` sounds good to me.

jespino (approved):

PR #15398, opened 2020-09-07
MM-27744 disable Zap for unit tests.

wiggin77:
#### Summary
Zap has no concept of shutdown or close. Zap is only shutdown when the app exits. Not a problem for console logging, but when creating a new Zap logger that outputs to files on every unit test, that leaves no easy way to clean up until process exit.  Depending on what else is running this can exhaust all file handles and cause unit tests to fail.

This PR disables Zap for unit tests and uses Logr instead, regardless of config settings.  With this PR `make test-server` peak file handle usage drops from ~5K to less than 100.

Once Zap is completely removed this special case will no longer be needed.

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-27744


app/admin.go
```diff
@@ -59,6 +60,10 @@ func (s *Server) GetLogsSkipSend(page, perPage int) ([]string, *model.AppError)
 	var lines []string
 
 	if *s.Config().LogSettings.EnableFile {
+		timeoutCtx, timeoutCancel := context.WithTimeout(context.Background(), time.Second*5)
```
reflog, line 63: a constant instead of a hardcoded timeout would be better I think.

app/server.go
```diff
@@ -623,13 +619,30 @@ func (s *Server) initLogging() error {
...
+	}
+
+	timeoutCtx, cancelCtx := context.WithTimeout(context.Background(), time.Second*10)
+	defer cancelCtx()
+
+	mlog.RemoveTargets(timeoutCtx, func(ti mlog.TargetInfo) bool {
+		if ti.Type != "*target.Writer" && ti.Type != "*target.File" {
+			return true
+		}
+		return false
```
reflog, line 635:
```suggestion
		return ti.Type != "*target.Writer" && ti.Type != "*target.File" 
```

reflog (approved):
looks great, left a couple of small comments

api4/apitestlib.go
```diff
@@ -93,6 +93,8 @@ func setupTestHelper(dbStore store.Store, searchEngine *searchengine.Broker, ent
 	options = append(options, app.ConfigStore(memoryStore))
 	options = append(options, app.StoreOverride(dbStore))
 
+	mlog.DisableZap()
```
agnivade, line 96: Is it possible to do the disabling inside the TestMain functions? Since this is purely for tests, that might be a better place for it.

app/server.go
```diff
@@ -112,20 +112,20 @@ type Server struct {
...
+	sessionCache            cache.Cache
+	seenPendingPostIdsCache cache.Cache
+	statusCache             cache.Cache
+	configListenerId        string
+	licenseListenerId       string
+	logListenerId           string
+	clusterLeaderListenerId string
+	searchConfigListenerId  string
+	searchLicenseListenerId string
+	loggerLicenseListenerId string
```
agnivade, line 125: This seems unrelated?

app/server.go
```diff
@@ -623,13 +619,30 @@ func (s *Server) initLogging() error {
 	return nil
 }
 
+func (s *Server) removeUnlicensedLogTargets(license *model.License) {
```
agnivade, line 622: Just trying to understand this, is it strictly related to the PR? If so, can you expand on it a bit more?

mlog/log.go
```diff
@@ -26,6 +27,15 @@ const (
 	LevelError = "error"
 )
 
+var (
+	// disableZap is set when Zap should be disabled and Logr used instead.
+	// This is needed for unit testing as Zap has no shutdown capabilities
+	// and holds file handles until process exit. Currently unit test create
+	// many server instances, and thus many Zap log files.
+	// This flag will be removed when Zap is permanently replaced.
+	disableZap bool
```
agnivade, line 36: I understand this is temporary. But having an unguarded global boolean variable can be racy.

Also, is there a ticket for zap removal? We should add that here.

app/server.go
wiggin77: Previously I was just shutting down Logr completely when unlicensed.  Now I need to keep Logr running (i.e. when using Logr in place of Zap targets) so instead of shutting down Logr we remove the unlicensed targets and leave the others running.

app/server.go
wiggin77: Previously this listenerid was just for logger metrics, but now we have multiple logger related action on a license change to faciliate removing unlicensed targets (see below).  This var was renamed to indicated it is for all logger related license watching.  The rest is just gofmt adjusting the column based on shorter name.

mlog/log.go
wiggin77: It is racy.  Not likely to be an issue now, but I'll guard it in case it gets accessed differently later.

wiggin77: I've made the flag atomic.  Thanks.  https://github.com/mattermost/mattermost-server/pull/15398/commits/3fc53b99414463cd7d2a7b5e92c0fc3acf5ad75a

app/admin.go
wiggin77: Thanks.  Fixed https://github.com/mattermost/mattermost-server/pull/15398/commits/ebdd80bd4f8b76eaaeb3862fe021bbe1c30fecdc

mlog/log.go
wiggin77: Also added ticket for Zap removal:  https://mattermost.atlassian.net/browse/MM-28405

api4/apitestlib.go
agnivade: This seemed to modify a higher level logic than the usual app-level setup performed in `setupTestHelper`. That's why I wanted to move it inside `TestMain`. It's not exactly a lot of deduplication either; it is already duplicated multiple times in api4, app, web and other packages.

But it's temporary code at the end of day. So if you feel this is okay, I don't mind.

mlog/log.go
```diff
@@ -24,6 +27,19 @@ const (
 	LevelWarn = "warn"
 	// Errors are messages about things we know are problems
 	LevelError = "error"
+
+	// DefFlushTimeout is the default amount of time mlog.Flush will wait
+	// before timing out.
+	DefFlushTimeout = time.Second * 5
```
agnivade, line 33: Nit: `DefaultFlushTimeout` is more readable.

wiggin77: Fixed:  https://github.com/mattermost/mattermost-server/pull/15398/commits/62c5a4cde5dd7dc36a4e32128ac5fbedec21962f

api4/apitestlib.go
wiggin77: Ok, I moved all the DisableZap calls to TestMain.

agnivade (approved):
Thanks

PR #15407, opened 2020-09-08
Clean caches on depcheck failure

jupenur:
#### Summary
When the check-deps step on CircleCI fails due to a bad DB version, delete the cached DB and try again.


metanerd (approved):
LGTM

PR #15409, opened 2020-09-08
MM-27953: Fix flaky test TestTeamStore//SearchPrivate

agnivade:
We got hit by yet another random text search issue, but this was
on the more extreme end because this was a string of length 4 and still
it collided.

I am not sure if there is a fool proof solution to this than just
going for a larger string.

https://mattermost.atlassian.net/browse/MM-27953


wiggin77 (approved):
Collision should be so rare that it'll be fun to see when it happens.

PR #15411, opened 2020-09-08
MM-28444: Optimize GetPostsSince in postgres

agnivade:
GetPostsSince is used when loading posts for a channel.

An opportunity for optimization is that the primary SQL query
is repeated twice and then a UNION is constructed for the results.

```
SELECT
		       *
		FROM
		       Posts
		WHERE
		       UpdateAt > :Time AND ChannelId = :ChannelId
		       LIMIT 1000
```

But we can use a CTE for this which caches the results to be reused later.

This leads to the main query being executed once rather than twice. And from
Postgres 12 onwards, CTEs can be inlined which opens the door to further optimizations.

From the docs (https://www.postgresql.org/docs/10/queries-with.html)

> A useful property of WITH queries is that they are evaluated only once per
execution of the parent query, even if they are referred to more than once
by the parent query or sibling WITH queries. Thus, expensive calculations
that are needed in multiple places can be placed within a WITH query to avoid redundant work.
Another possible application is to prevent unwanted multiple evaluations of functions
with side-effects. However, the other side of this coin is that the optimizer is
less able to push restrictions from the parent query down into a WITH query than an ordinary subquery.

In our case, the caveat does not apply because we are only filtering columns and not rows,
so we can safely use it.

Following are the query plan comparisons:
Old: http://tatiyants.com/pev/#/plans/plan_1599394993105
New: http://tatiyants.com/pev/#/plans/plan_1599395970886

As we can see, in old bitmap index scan+heap scan happens twice, but in the new one,
it happens only once.

This has been load tested by running simul coordinator for 1.5 hrs starting with a base dataset of aroun 2M posts and confirmed to exhibit good improvements.

p99-getposts
avg-getposts

https://mattermost.atlassian.net/browse/MM-28444

### DIFF
--- a/store/sqlstore/post_store.go
+++ b/store/sqlstore/post_store.go
@@ -6,12 +6,13 @@ package sqlstore
 import (
 	"database/sql"
 	"fmt"
-	"github.com/mattermost/mattermost-server/v5/store/searchlayer"
 	"net/http"
 	"regexp"
 	"strconv"
 	"strings"
 	"sync"
+
+	"github.com/mattermost/mattermost-server/v5/store/searchlayer"
 
 	"github.com/pkg/errors"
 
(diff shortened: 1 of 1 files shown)


agnivade:
Looks like the explain visualizer does not allow one to share visualizations. Here is the raw data, which can be uploaded to the site and seen for yourself.

[q1.sql.txt](https://github.com/mattermost/mattermost-server/files/5188644/q1.sql.txt)
[out1.json.txt](https://github.com/mattermost/mattermost-server/files/5188645/out1.json.txt)
[q2.sql.txt](https://github.com/mattermost/mattermost-server/files/5188646/q2.sql.txt)
[out2.json.txt](https://github.com/mattermost/mattermost-server/files/5188647/out2.json.txt)

store/sqlstore/post_store.go
```diff
@@ -647,35 +648,19 @@ func (s *SqlPostStore) GetPostsSince(options model.GetPostsSinceOptions, allowFr
...
-                                               AND ChannelId = :ChannelId
-                               LIMIT 1000) temp_tab))
-               ORDER BY CreateAt DESC`
+		query = `WITH cte AS (SELECT
+		       *
+		FROM
+		       Posts
+		WHERE
+		       UpdateAt > :Time AND ChannelId = :ChannelId
+		       LIMIT 1000)
```
grundleborg, line 657: Does there need to be an ORDER BY in here to ensure deterministic results?

agnivade: The `ORDER BY` is done later after the UNION is done. If you recall https://community-daily.mattermost.com/core/channels/developers-server/s1ab7695z3rqumsgsdsj91gp1h, this API is known to be buggy and we just documented it as magic.

grundleborg: Ah yeah, of course. We can open that can of worms another day (hopefully when this is handled by a different endpoint).

wiggin77 (approved):
LGTM

streamer45 (approved):
Nice :+1:

PR #15421, opened 2020-09-09
[MM-28192] Obfuscate Google API Key

hahmadia:
#### Summary

Obfuscates the google API key on System Console > Site Configuration > Posts. 

#### Ticket Link

https://mattermost.atlassian.net/browse/MM-28192

image

### DIFF
--- a/config/utils.go
+++ b/config/utils.go
@@ -37,6 +37,10 @@ func desanitize(actual, target *model.Config) {
 		target.GoogleSettings.Secret = actual.GoogleSettings.Secret
 	}
 
+	if target.ServiceSettings.GoogleDeveloperKey != nil && *target.ServiceSettings.GoogleDeveloperKey == model.FAKE_SETTING {
+		target.ServiceSettings.GoogleDeveloperKey = actual.ServiceSettings.GoogleDeveloperKey
+	}
+
 	if target.Office365Settings.Secret != nil && *target.Office365Settings.Secret == model.FAKE_SETTING {
 		target.Office365Settings.Secret = actual.Office365Settings.Secret
 	}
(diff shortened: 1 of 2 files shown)


grundleborg:
0/5 but is there any point obfuscating a value that is sent in plain text in the client config to every logged in user (admin or not) anyway?

sbishel (commented):
Should update/add unit tests.

sbishel:
@grundleborg Do you know the history behind why these are passed in the client config? They are not used in the client are they? This PR will at least prevent them from being displayed in the System Console and be consistent with other setting in the System Console.

grundleborg:
@sbishel I looked at the webapp source and it appears this key is for embedding youtube videos, so it is used in the client. On that basis, I don't think it needs to be obfuscated in the system console since it has to be shared with all clients anyway.

sbishel:
@grundleborg Thanks. I missed it when I searched. I just saw it being set in config. Not used anywhere.

hahmadia:
@sbishel @grundleborg 
As per your guys comments, do it mean I should close this PR and close the ticket as won't do linking to the discussion here?

grundleborg:
@hahmadia I asked the product security team about this, and their response was basically that they aren't too worried about what we choose to do. Based on their comments, I wonder whether the best course of action would be to instead update the text beneath the input to note that this is not a secret key and is exposed to all logged-in clients as a result of what it is used for?

Which PM requested this? Perhaps you can circle back to them with the comments on this PR and see if they can compose a suitable text change?

hahmadia:
Thank you all. Closed the Jira ticket and closing this PR.

PR #15422, opened 2020-09-10
[MM-28383] Fix nil pointer dereference in addTeamMember

streamer45:
#### Summary

PR fixes a nil pointer dereference in addTeamMember in case of an invalid JSON request body.

#### Ticket

https://mattermost.atlassian.net/browse/MM-28383


model/team_member.go
```diff
@@ -69,10 +69,12 @@ func (o *TeamUnread) ToJson() string {
 	return string(b)
 }
 
-func TeamMemberFromJson(data io.Reader) *TeamMember {
+func TeamMemberFromJson(data io.Reader) (*TeamMember, error) {
```
streamer45, line 72: Do we consider this a *breaking change* that should only go in a major release?
I've tried to do the *right* thing here and avoid ignoring the error but I was left with this doubt.

streamer45: @jespino 

Do you think it makes sense to add a `TeamMemberFromJSON` (notice the uppercase at the end) version that handles the error and start the deprecation of `TeamMemberFromJson`? Or were you suggesting to just add an internal (unexported) method?

streamer45: Alright, keeping it simple then.

jespino (approved):
Lgtm

mgdelacroix (approved):

streamer45:
@lindalumitchell 

PR adds some error handling to fix a possible panic. I've added a unit test to cover this case.

lindalumitchell (approved):

PR #15428, opened 2020-09-11
MM-27147 - `/cloud` api endpoints

marianunez:
#### Summary

- Added initial `/cloud` endpoints needed for in-product cloud purchase handled by the cloud app layer in enterprise through the `einterfaces.CloudInterface`
	- GET `/cloud/products`
 	- POST `/cloud/payment/`
	- POST `/cloud/payment/confirm`

- Defined new `CloudSettings` for CWS url configuration and to include other cloud settings (i.e `UserLimit` that will be in a separate PR).
- Included `js.stripe.com/v3` as an allowed source in the CSP that is overwritten when the webapp is served based on cloud licenses.

#### Pending:

- [x] Unit tests
- [x] Adding new endpoints to OpenAPI

#### Ticket Link
[MM-27147](https://mattermost.atlassian.net/browse/MM-27147)

#### Related PRs
- Enterprise: https://github.com/mattermost/enterprise/pull/776
- CWS: https://github.com/mattermost/customer-web-server/pull/119
- WebApp - WIP
- Api Reference: https://github.com/mattermost/mattermost-api-reference/pull/577


utils/subpath.go
```diff
@@ -85,13 +85,13 @@ func UpdateAssetsSubpath(subpath string) error {
 
 	newRootHtml := string(oldRootHtml)
 
-	reCSP := regexp.MustCompile(`<meta http-equiv="Content-Security-Policy" content="script-src 'self' cdn.rudderlabs.com/([^"]*)">`)
+	reCSP := regexp.MustCompile(`<meta http-equiv="Content-Security-Policy" content="script-src 'self' cdn.rudderlabs.com/ js.stripe.com/v3([^"]*)">`)
```
marianunez, line 88: We may not need to add here given but only below in `web/handlers.go`, unless we see there is a possibility that in cloud the webapp may be hosted in a subpath.

marianunez:
@devinbinnie this is the boilerplate to get the server connected to CWS which you can branch off from. You would need the corresponding [enterprise](https://github.com/mattermost/enterprise/pull/761) and [CWS](https://github.com/mattermost/customer-web-server/pull/119) branches for all of this to work properly. 

[Here](https://github.com/mattermost/customer-web-server/blob/MM-27145/README.md#set-up-to-connect-mattermost-webapp-with-cws-locally) are setup instructions to develop on the webapp locally. We can sync offline on the details of the endpoints needed.

api4/cloud.go
```diff
@@ -0,0 +1,106 @@
...
+}
+
+func getCloudProducts(c *Context, w http.ResponseWriter, r *http.Request) {
+	if c.App.Srv().License() == nil || !*c.App.Srv().License().Features.Cloud {
+		c.Err = model.NewAppError("Api4.getCloudProducts", "api.cloud.license_error", nil, "", http.StatusNotImplemented)
+		return
+	}
+
+	if !c.App.SessionHasPermissionTo(*c.App.Session(), model.PERMISSION_MANAGE_SYSTEM) {
+		c.SetPermissionError(model.PERMISSION_MANAGE_SYSTEM)
```
jwilander, line 31: Good for now but we might want to consider adding a new permission specific to billing

utils/subpath.go
jwilander: Good to keep IMO 👍 I could see us using subpaths in the future

jwilander (approved):
LGTM

api4/cloud.go
marianunez: Agreed. The new concept of `Billing Admins` part of the billing epic would cover this.

api4/cloud.go
```diff
@@ -0,0 +1,106 @@
...
+		c.Err = model.NewAppError("Api4.createCustomerPayment", "api.cloud.license_error", nil, "", http.StatusNotImplemented)
+		return
+	}
+
+	if !c.App.SessionHasPermissionTo(*c.App.Session(), model.PERMISSION_MANAGE_SYSTEM) {
+		c.SetPermissionError(model.PERMISSION_MANAGE_SYSTEM)
+		return
+	}
+
+	intent, appErr := c.App.Cloud().CreateCustomerPayment()
```
ethervoid, line 61: Probably we want to include audit records like [this one](https://github.com/mattermost/mattermost-server/blob/master/api4/user.go#L107-L110)

model/cloud.go
```diff
@@ -0,0 +1,32 @@
...
+// See LICENSE.txt for license information.
+
+package model
+
+// Product model represents a product on the cloud system.
+type Product struct {
+	ID             string   `json:"id"`
+	Name           string   `json:"name"`
+	Description    string   `json:"description"`
+	DollarsPerSeat float64  `json:"dollars_per_seat"`
```
ethervoid, line 11: Shouldn't be better PricePerSeat to not couple it to a specific currency?

ethervoid (commented):
Some questions about these changes

api4/cloud.go
marianunez: Added! Thanks!

model/cloud.go
marianunez: This was inherited directly from CWS definition but I like your naming proposal better. Thanks!

api4/cloud.go
```diff
@@ -32,6 +33,9 @@ func getCloudProducts(c *Context, w http.ResponseWriter, r *http.Request) {
 		return
 	}
 
+	auditRec := c.MakeAuditRecord("getCloudProducts", audit.Fail)
```
ethervoid, line 36: You have to specifically mark the audit as success `auditRec.Success()` in the end, if not is going to be taken as failure

ethervoid (approved):
LGTM! Noice job!

agnivade:
This PR needs to be reverted. This was checked in without pushing the EE branch. Therefore master is broken right now.

jwilander:
Heads-up @marianunez ^

marianunez:
> Heads-up @marianunez ^

Yes, my bad. Will revert once https://github.com/mattermost/enterprise/pull/776 is merged. Thanks @agnivade!

marianunez:
@amyblais this was documented in the api reference here: mattermost/mattermost-api-reference#577 Any other place that needs to be updated?

amyblais:
@marianunez New config settings (in this case "CloudSettings") normally get documented in https://docs.mattermost.com/administration/config-settings.html. However, since this is for cloud, perhaps it doesn't need to be documented there.

marianunez:
> @marianunez New config settings (in this case "CloudSettings") normally get documented in [docs.mattermost.com/administration/config-settings.html](https://docs.mattermost.com/administration/config-settings.html). However, since this is for cloud, perhaps it doesn't need to be documented there.

Yes, given that the use of the `CloudSettings` is internal, I don't see any value in adding in the Settings documentation.

amyblais:
Thanks, sounds good.

PR #15431, opened 2020-09-11
MM-28115: Add env var for telemetry installation type.

grundleborg:
#### Summary
Add an env var to indicate the installation type for telemetry. Various installers will set this appropriately, and the default is an empty string (i.e. en var unset).

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-28115

### DIFF
--- a/services/telemetry/telemetry.go
+++ b/services/telemetry/telemetry.go
@@ -29,6 +29,8 @@ const (
 	RUDDER_KEY           = "placeholder_rudder_key"
 	RUDDER_DATAPLANE_URL = "placeholder_rudder_dataplane_url"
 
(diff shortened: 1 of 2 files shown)


jespino (approved):

mgdelacroix (approved):
LGTM! :+1:

grundleborg:
@lindalumitchell this change causes Mattermost to check the value of an environment variable and include it's value in a telemetry field. There is a unit test which tests both without and with the variable set and makes sure the appropriate telemetry data is sent. In terms of manual testing, I think it would probably be just the same as what the unit test does.

lindalumitchell (approved):

PR #15433, opened 2020-09-11
[MM-28033] Perform localcachelayer tests for one DB type only when in CI

streamer45:
#### Summary

PR adds a check in `localcachelayer.initStores()` so that when in CI we can run the tests on one DB type only.

#### Ticket

https://mattermost.atlassian.net/browse/MM-28033

### DIFF
--- a/store/localcachelayer/layer_test.go
+++ b/store/localcachelayer/layer_test.go
@@ -4,6 +4,7 @@
 package localcachelayer
 
 import (
+	"os"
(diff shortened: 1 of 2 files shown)


isacikgoz (approved):
Nicely done 👍

agnivade (approved):
Hmm, the `newStoreType` function isn't there in the sqlstore package. Wondering if we want to duplicate it there too?

PR #15437, opened 2020-09-11
[MM-28475] Add server telemetry for the new system roles

fmunshi:
#### Summary
- Adds telemetry to each of the new system roles (count of users and whether the permissions have been modified from the defaults) 

#### Ticket Link
- https://mattermost.atlassian.net/browse/MM-28475


model/role.go
```diff
@@ -97,6 +100,59 @@ func init() {
 			PERMISSION_EDIT_BRAND,
 		},
 	}
+
+	SystemUserManagerDefaultPermissions = []string{
```
Adovenmuehle, line 104: Just making sure I understand, this is basically just refactoring these permission lists so the code is more readable?

services/telemetry/telemetry.go
```diff
@@ -65,6 +65,7 @@ const (
 	TRACK_PERMISSIONS_GENERAL       = "permissions_general"
 	TRACK_PERMISSIONS_SYSTEM_SCHEME = "permissions_system_scheme"
 	TRACK_PERMISSIONS_TEAM_SCHEMES  = "permissions_team_schemes"
+	TRACK_PERMISSIONS_SYSTEM_ROLES  = "permissions_system_roles"
```
Adovenmuehle, line 68: FYI @enelson720 this'll be a new table that'll be generated for server telemetry.

Adovenmuehle (approved):
LGTM

model/role.go
Adovenmuehle: (And also makes it easier to add the common ancillary permissions)

fmunshi: Its mostly just to expose the defaults as something that can be used by the telemetry package, before this there was no way to grab initial server defaults for a role without duplicating the stuff here

services/telemetry/telemetry.go
fmunshi: @Adovenmuehle does it make sense to just make this part of the `permissions_system_scheme` table then? Maybe thats a question for @thefactremains

Adovenmuehle: Yeah that actually seems like it would simplify it. Good call.

fmunshi: and maybe for completeness I can add the actual permissions for each of these roles to that table too

Adovenmuehle: The permissions for each role are hard-coded and common across all servers (of the same version), correct? In that case I don't think there's any huge need.

fmunshi: If the admin modifies them then they would be different - I guess the question is _if_ permissions are modified do we care what the value of those permissions are? cc @thefactremains

Adovenmuehle: Disregard my last comment, yeah I think you should do that.

thefactremains: > If the admin modifies them then they would be different (hence the `_permissions_modified` metrics) - I guess the question is _if_ permissions are modified do we care what the value of those permissions are? cc @thefactremains

Yes. We want to know that.

thefactremains (approved):

sbishel (approved):

mkraft (approved):

srkgupta (approved):
Telemetry will be tested on the QA test instances post merge. Approving the PR as is.

PR #15500, opened 2020-09-15
[MM-28717] Refactor applyMultiRoleFilters to use sq builder

fmunshi:
#### Summary
- Allows the roles query to work with the new system roles
- Also add the roles check to the `GetAllProfiles` endpoint 

#### Ticket Link
- https://mattermost.atlassian.net/browse/MM-28717

#### Release Notes

```release-notes
Added a user interface for the existing new system roles functionality including a new permission to `read/write system_roles`. This feature is currently in Beta.
```


fmunshi:
In the process of adding some tests...

hahmadia (approved):
Very nice cleanup

store/sqlstore/user_store.go
```diff
@@ -456,119 +457,59 @@ func applyRoleFilter(query sq.SelectBuilder, role string, isPostgreSQL bool) sq.
 	return query.Where("u.Roles LIKE ? ESCAPE '*'", roleParam)
 }
 
-func applyMultiRoleFilters(query sq.SelectBuilder, roles []string, teamRoles []string, channelRoles []string) sq.SelectBuilder {
-	queryString := ""
-	if len(roles) > 0 && roles[0] != "" {
-		schemeGuest := false
-		schemeAdmin := false
-		schemeUser := false
+func applyMultiRoleFilters(query sq.SelectBuilder, roles []string, teamRoles []string, channelRoles []string, isPostgreSQL bool) sq.SelectBuilder {
```
mkraft, line 460: Are `roles` *system* roles in this context? If so I think `systemRoles` would be a clearer parameter name.

store/sqlstore/user_store.go
```diff
@@ -456,119 +457,59 @@ func applyRoleFilter(query sq.SelectBuilder, role string, isPostgreSQL bool) sq.
...
+	sqOr := sq.Or{}
 
+	if len(roles) > 0 && roles[0] != "" {
```
mkraft, line 463: Why are does `roles` sometimes have an empty string in the first position?

store/sqlstore/user_store.go
```diff
@@ -456,119 +457,59 @@ func applyRoleFilter(query sq.SelectBuilder, role string, isPostgreSQL bool) sq.
...
-			} else if schemeUser {
-				queryString += `(cm.SchemeUser = true AND cm.SchemeAdmin = false AND u.Roles = 'system_user')`
-			}
-
-			if schemeGuest {
-				if queryString != "" && queryString[len(queryString)-3:] != "OR " {
-					queryString += "OR "
-				}
-				queryString += `(cm.SchemeGuest = true AND u.Roles = 'system_guest')`
+				sqOr = append(sqOr, sq.Eq{"cm.SchemeGuest": true})
```
mkraft, line 490: Why do some of these cases not need to check for the `Roles` value?

store/sqlstore/user_store.go
fmunshi: it doesnt, _unless_ someone someone were to pass an empty string as the value for roles... I just decided to protect against that case, do you think its overkill?

store/sqlstore/user_store.go
fmunshi: The logic here is basically designed to only get a user if their "highest level" role matches the queried role so if the query is for a `team_admin` then we want to exclude users that are `system_admins` and only grab `system_users`

But for team / channel guest we dont need to do that since a guest is a guest in all scenarios

store/sqlstore/user_store.go
fmunshi: done

jasonblais:
@mkraft is this something you/team will take from here?

mkraft:
> @mkraft is this something you/team will take from here?

Yes. Thanks Jason.

mkraft (approved):

srkgupta:
@hahmadia can you please update the QA Test steps on what needs to be tested because of this change? Also some insights on what pages are affected so that it can be regression tested.

hahmadia:
@srkgupta 
As I'm not completely sure, I've DM'ed Martin to hopefully see if he has a better idea. 
Sorry about that.

hahmadia:
> @hahmadia can you please update the QA Test steps on what needs to be tested because of this change? Also some insights on what pages are affected so that it can be regression tested.

Hey @srkgupta 

Okay so I tried looking and it seems like there is just a ton of places this particular function is being called. However, some areas that you can test specifically to ensure it works as it should is in the searching modals for adding users and finding users throughout the system console. Adding users to a channel or team modal is one place that users is one place that eventually gets to the refactored function.

api4/role.go
```diff
@@ -125,14 +130,45 @@ func patchRole(c *Context, w http.ResponseWriter, r *http.Request) {
...
+
 	if c.App.Srv().License() != nil && (oldRole.Name == "system_guest" || oldRole.Name == "team_guest" || oldRole.Name == "channel_guest") && !*c.App.Srv().License().Features.GuestAccountsPermissions {
 		c.Err = model.NewAppError("Api4.PatchRoles", "api.roles.patch_roles.license.error", nil, "", http.StatusNotImplemented)
 		return
 	}
 
-	if !c.App.SessionHasPermissionTo(*c.App.Session(), model.PERMISSION_SYSCONSOLE_WRITE_USERMANAGEMENT_PERMISSIONS) {
-		c.SetPermissionError(model.PERMISSION_SYSCONSOLE_WRITE_USERMANAGEMENT_PERMISSIONS)
-		return
+	if oldRole.Name == "team_admin" || oldRole.Name == "channel_admin" || oldRole.Name == "system_user" || oldRole.Name == "team_user" || oldRole.Name == "channel_user" || oldRole.Name == "system_guest" || oldRole.Name == "team_guest" || oldRole.Name == "channel_guest" {
```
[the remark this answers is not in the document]
sbishel: Also a comment on what is happening here. Would it make more sense to check "other" roles.

api4/role.go
```diff
@@ -125,14 +130,45 @@ func patchRole(c *Context, w http.ResponseWriter, r *http.Request) {
...
 	}
 
+	if patch.Permissions != nil {
+		deltaPermissions := model.PermissionsChangedByPatch(oldRole, patch)
+
+		notAllowedPermissions := []string{
+			model.PERMISSION_SYSCONSOLE_WRITE_USERMANAGEMENT_SYSTEM_ROLES.Id,
+			model.PERMISSION_SYSCONSOLE_READ_USERMANAGEMENT_SYSTEM_ROLES.Id,
+			model.PERMISSION_MANAGE_ROLES.Id,
+		}
```
sbishel, line 140: Create as const, it doesn't need created each time through.

api4/role.go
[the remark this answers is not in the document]
hahmadia: OK I can use the constants. I was just following how it was being done above.
This is for the permission schemes in the permissions section of the system console. I don't think it makes sense to check any other role as these are the only roles that are changed in the permission section of the system console. 

Basically what it's saying is if the user has write access in the permission section of the system console, they can edit these roles and these roles only.

Otherwise, for other roles, they require the `WRITE_SYSTEM_ROLES` permission.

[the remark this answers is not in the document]
hahmadia: I've changed the other strings in that function to use the constants from `model` as well. Good change you've pointed out there.

api4/role.go
hahmadia: Do you mean drag it out to a global scope and use a `var` ?  I took them out to a global scope and instead of using `:=` used `=` making them somewhat of a pseudo-constant.

api4/role.go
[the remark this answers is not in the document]
sbishel: Here, i just thought if there were less "other" roles, it might be easier to turn the If statement around.

sbishel (approved):

mkraft (approved):

furqanmlk (approved):
Tested all test cases 
https://mattermost.atlassian.net/projects/MM?selectedItem=com.atlassian.plugins.atlassian-connect-plugin%3Acom.kanoah.test-manager__main-project-page#!/design?projectId=10302

Verified following issues
MM-29892 , MM-29895 , MM-29926 , MM-29927

cwarnermm:
Documentation updated via: https://github.com/mattermost/docs/pull/4175

PR #15502, opened 2020-09-16
Fix incorrect error handling

### LINKED ISSUE #15501: Fix error handling in LeaveChannel operation
#### Summary

While doing the migration of another store implementation I realized that an error was introduced in a previous `ChannelStore` migration.

It turns out when leaving a channel if there was an error getting it then that error will never be handled since we are using another reference value, that is because `store.StoreResult` now has 2 error values:

```go
type StoreResult struct {
	Data interface{}
	Err  *model.AppError

	NErr error
}
```
The error is set into `NErr` but it is being handled with `Err` which is always nil.

rvillablanca:
#### Summary

This PR solves an error introduced in another PR belonging to the error migration campaign. More details in the linked issue.

#### Ticket Link

Fixes #15501

### DIFF
--- a/app/channel.go
+++ b/app/channel.go
@@ -1974,18 +1974,18 @@ func (a *App) LeaveChannel(channelId string, userId string) *model.AppError {
 	}()
 
 	cresult := <-sc
-	if cresult.Err != nil {
-		return cresult.Err
-	}
-	uresult := <-uc
-	if uresult.NErr != nil {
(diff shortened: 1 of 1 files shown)


jespino (approved):
LGTM

agnivade (approved):
Nice catch!

lindalumitchell (approved):

PR #15507, opened 2020-09-16
MM-28370: Fix send on closed channel

agnivade:
We check if the push notification hub is shutting down before sending
anything to the channel.

This is accomplished by closing the stopChan instead of sending struct{}{} to it,
and doing a non-blocking read of it before sending the message to the hub.

The primary reason for this strange pattern is that the shutdown of HTTP server is
partly asynchronous, and even after the server is shutdown, there is no guarantee
that any stray goroutines are active.

Alternatively, we could choose to _not_ close the notificationsChan at all, but then
the request will just get stuck and the server has to do an unclean shutdown.

This approach doesn't look so bad.

Admittedly, this is still racy because there is no guarantee in the order of instructions
executed in the goroutine that's closing the channel and the goroutine that's sending to the channel.
But this should atleast take care of the most common case.

https://mattermost.atlassian.net/browse/MM-28370

### DIFF
--- a/app/notification_push.go
+++ b/app/notification_push.go
@@ -142,6 +142,13 @@ func (a *App) sendPushNotification(notification *PostNotification, user *model.U
 	channelName := notification.GetChannelName(nameFormat, user.Id)
 	senderName := notification.GetSenderName(nameFormat, *cfg.ServiceSettings.EnablePostUsernameOverride)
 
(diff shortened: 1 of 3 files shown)


agnivade:
I am calling it the "Ibrahim pattern", after we did a similar thing with mattermod. :P

PR #15509, opened 2020-09-16
[MM-28707] Fix plugin version telemetry for air gaped installations

hanzei:
#### Summary
If a server runs in an air gaped environment, the plugin versions field in the telemetry data would return if the plugin is enabled instead of the actual version. 

Please note that this case is quite rare. Currently 237 of 92k servers are effected by this bug.

I tried adding a test but it's quite complicated to install a plugin in a service package. We could, instead of using the actual implementation of a plugin environment, mock it away by replacing https://github.com/mattermost/mattermost-server/blob/86b841eac5960969e50e0ad4f7f422a44e53eaf0/services/telemetry/telemetry_test.go#L46-L54 with the mock. 0/5 if that refactoring is worth for this little bug fix.

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-28707


agarciamontoro (approved):
Thank you for tackling this, @hanzei! I don't think that the refactor is worth it if it's only useful for this bug fix. I'd say to merge this directly.

lieut-data (approved):

hanzei:
@lindy65 I'm not sure if this can be tested pre merge.

hanzei:
@prapti @lindy65 told me that you are familiar with testing telemetry changes. Could you please take a look?

amyblais:
@prapti Kind reminder to review for v5.28.

prapti:
Thanks @amyblais!
@hanzei I'm not very familiar, but looking into it (setting up my local with looker). I can definitely take a look, and if I can't get it figured out, will check with Dylan.
Adding this PR to my test queue.

prapti:
@hanzei Looks like this PR will need https://mattermost.atlassian.net/browse/MM-28952 to be resolved first? I am not able to see any local telemetry data on rudder/looker, so I wanted to make sure.

hanzei:
@prapti You can edit `RUDDER_KEY` and `RUDDER_DATAPLANE_URL` in `services/telemetry/telemetry.go`. It's a bit of a hacky workaround, but it works.

prapti:
Got it. Thanks @hanzei! I'll give that try soon.

prapti:
@hanzei Could you please help me understand what the "air gaped environment" is? Looks like a normal test should suffice to verify that the incident response plugin version is showing up correctly in Looker but is there a different scenario to keep in mind? Thank you!

lieut-data:
@prapti, for this case, a Mattermost server that can't reach the marketplace should suffice. I sometimes disconnect from the internet to simulate this case, as it's not sufficient to just change the marketplace URL -- we're always querying the canonical one here.

prapti:
Thanks @lieut-data!

@DHaussermann sorry to bug you again about telemetry, but would be able to help with this? I have a few plugins installed and working but it's been about an hour or so since I've been trying to see if any of the plugin data comes through on Looker, but there is none. Wondering if you can help verify?

prapti (approved):
Ah nevermind @DHaussermann, it just takes longer for the data to show up. I can actually see some Incident Response activities now that I had tried on my local server come through.
Approving.

PR #15513, opened 2020-09-16
Trimmed trailing slash in URL provided to create Client API

shieldsjared:
#### Summary
As described in [this issue](https://github.com/mattermost/mattermost-plugin-starter-template/issues/115) while attempting to utilize the MM Plugin Started Template, it was determined that including a trailing slash when calling `model.NewAPIv4Client` with a URL that includes a trailing slash, the resulting URL's become malformed (double-slash) and MM is unable to route the requests appropriately.  Trimming a trailing slash prevents users from having to be too particular when specifying the URL of their target server.

Note that this fix does not correct if a user attempts to use a URL with multiple trailing-slashes (however, this should be a bit more obvious of an issue to the end-user if they incorrectly set the MM url with multiple trailing-slashes). 

#### Ticket Link
Fixes: https://github.com/mattermost/mattermost-plugin-starter-template/issues/115

Also, please see [PR on that repository](https://github.com/mattermost/mattermost-plugin-starter-template/pull/120) for discussion and agreed-upon path forward for resolution (included in this PR)


model/client4_test.go
```diff
@@ -12,6 +12,12 @@ import (
 	"github.com/stretchr/testify/assert"
 )
 
+// https://github.com/mattermost/mattermost-plugin-starter-template/issues/115
+func TestClient4TrimTrailingSlash(t *testing.T) {
+	client := NewAPIv4Client("https://foo.com:1234/")
+	assert.False(t, strings.HasSuffix(client.Url, "/"))
```
lieut-data, line 18: Can we be more explicit here, and test the expected URL /and/ the expected API url? The latter, I think, is the most important part of this change:
```suggestion
 	assert.Equal(t, "https://foo.com:1234", client.Url)
 	assert.Equal(t, "https://foo.com:1234/api/v4", client.ApiUrl)
```

lieut-data (changes requested):
Thanks, @shieldsjared! One comment below.

model/client4.go
```diff
@@ -116,6 +116,7 @@ func (c *Client4) Must(result interface{}, resp *Response) interface{} {
 }
 
 func NewAPIv4Client(url string) *Client4 {
+	url = strings.TrimSuffix(url, "/")
```
lieut-data, line 119: Would we consider using `strings.TrimRight`, to get the automatic removal of multiple slashes? Easy to unit test as well.

shieldsjared: Great point. Changed to `TrimRight()`.  Didn't realize that method was hanging out there!

model/client4_test.go
shieldsjared: Certainly.  I also updated the case to stress the three likely inputs: 
- 0 slashes (which is expected), 
- 1 slash (which is probable), and 
- 5 slashes (unlikely but possible if you fell asleep at the wheel). An arbitrary number to demonstrate it's removing "all"

shieldsjared:
Changes implemented and ready for review again @lieut-data - Thanks!

hanzei (approved):
LGTM :+1:

lieut-data (approved):
Sweet!

amyblais:
@prapti Kind reminder to review for v5.28.

prapti:
Thanks @amyblais!
In test queue.

prapti (approved):
Looks good to me! Thank you @shieldsjared for this fix and thank you for noting that "this fix does not correct if a user attempts to use a URL with multiple trailing-slashes", that was helpful :)

PR #15522, opened 2020-09-16
[MM-28767] Fix failing TestS3FileBackendTestSuite/TestAppendFile

streamer45:
#### Summary

PR fixes failing `TestS3FileBackendTestSuite/TestAppendFile` due to a missing S3 bucket.
Other tests would work only because they would get executed after `TestConnection` which would setup the bucket.
Unfortunately `TestAppendFile` would get executed before that cause of the lexicographical order :facepalm: .

#### Ticket

https://mattermost.atlassian.net/browse/MM-28767


metanerd (approved):
Thank you @streamer45 !

PR #15525, opened 2020-09-17
[MM-28100]: fix getChannelByNameForTeamName permissions

ashishbhate:
#### Summary

- fix `getChannelByNameForTeamName` to allow team members to join public channels. I'm not sure why, but the authorization logic of getChannelByName and getChannelByNameForTeamName was different. After this change they behave the same. This fixes an issue with the mobile app.

#### Ticket Link

- https://mattermost.atlassian.net/browse/MM-28100
- Related PR: https://github.com/mattermost/mattermost-mobile/pull/4810


enahum (approved):
Lgtm, let’s wait for ppl with better gofu than me to review it.

reflog (approved):

api4/channel.go
```diff
@@ -1150,9 +1150,17 @@ func getChannelByNameForTeamName(c *Context, w http.ResponseWriter, r *http.Requ
 		return
 	}
 
-	if !c.App.SessionHasPermissionToChannel(*c.App.Session(), channel.Id, model.PERMISSION_READ_CHANNEL) {
-		c.Err = model.NewAppError("getChannelByNameForTeamName", "app.channel.get_by_name.missing.app_error", nil, "teamId="+channel.TeamId+", "+"name="+channel.Name+"", http.StatusNotFound)
-		return
+	if channel.Type == model.CHANNEL_OPEN {
+		if !c.App.SessionHasPermissionToTeam(*c.App.Session(), channel.TeamId, model.PERMISSION_READ_PUBLIC_CHANNEL) &&
+			!c.App.SessionHasPermissionToChannel(*c.App.Session(), channel.Id, model.PERMISSION_READ_CHANNEL) {
```
agnivade, line 1155: Nit: It might be cleaner to refactor the permission bools to separate variables and then uplift the checks in the if statement itself.

```go
channelOk := c.App.SessionHasPermissionToChannel(*c.App.Session(), channel.Id, model.PERMISSION_READ_CHANNEL)
teamOk := c.App.SessionHasPermissionToTeam(*c.App.Session(), channel.TeamId, model.PERMISSION_READ_PUBLIC_CHANNEL)


if channel.Type == model.CHANNEL_OPEN && !channelOk && !teamOk {
   // throw error
} else if !channelOk {
  // throw error
}
```

api4/channel.go
```diff
@@ -1150,9 +1150,17 @@ func getChannelByNameForTeamName(c *Context, w http.ResponseWriter, r *http.Requ
...
+			c.SetPermissionError(model.PERMISSION_READ_PUBLIC_CHANNEL)
```
agnivade, line 1156: Any reason the handling of errors in the 2 cases are different? Calling `SetPermissionError` does not capture the teamId and channelName.

agnivade (commented):
Got some small questions

api4/channel.go
ashishbhate: > Any reason the handling of errors in the 2 cases are different? Calling SetPermissionError does not capture the teamId and channelName.

Not really, my intent was to just to update `getChannelByNameForTeamName` to behave the same (re permissions) as `getChannelByName`. I'm not sure why permissions were handled differently. 

Same for the comment above. I'd prefer to keep the code, behaviour and logging the same, but I'm open to updating/cleaning up this method.

agnivade: Understood. The refactoring was just a suggestion anyways.

agnivade (approved):

api4/channel.go
ashishbhate: This suggested change improves readability a lot, so I've applied it. The if channelOk and teamOk check can't be uplifted since we want the `else` to trigger only if the channel type is not public, and not on either of the other failing checks.

josephbaylon (approved):
LGTM. User is able to tap on public permalink and join the channel in both when links that already exist on a channel or a new link that user posts on the a channel.

Tested in conjunction with https://github.com/mattermost/mattermost-mobile/pull/4810

PR #15534, opened 2020-09-18
[MM-28638] Improve image thumbnail generation logic

streamer45:
#### Summary

PR improves the logic for generating an image thumbnail.
It also adds two utility functions (`genPreview` and `genThumbnail`) to clean up the code a bit.

A couple of benchmarks quickly revealed that the main issue was not in the image library implementation but in how `imaging.Resize()` was used:

In case of very extreme aspect ratios (either width much greater than height or vice versa) we would end up upscaling the output image to an incredible amount. A small example:

Input image dimensions: 1x10000

To generate the thumbnail for this case we would call:

```go
imaging.Resize(rawImg, 120, 0, imaging.Lanczos)
```
We force the width to be 120 pixels and we pass 0 as the height since we want to keep the aspect ratio. Unfortunately this causes the resulting image to be:

120x(120 * 10000 / 1) = 120x1200000

Which is a 144MP image, weighting (in raw format) ~550MB of memory, clearly not what we expect from a thumbnail :)

#### Ticket

https://mattermost.atlassian.net/browse/MM-28638


app/image.go
```diff
@@ -0,0 +1,50 @@
...
+func genThumbnail(img image.Image) image.Image {
+	thumb := img
+	w := img.Bounds().Dx()
+	h := img.Bounds().Dy()
+
+	if h > ImageThumbnailHeight || w > ImageThumbnailWidth {
+		ratio := float64(h) / float64(w)
+		if ratio < ImageThumbnailRatio {
+			// we pre-calculate the thumbnail's width to make sure we are not upscaling.
+			tw := int(float64(ImageThumbnailHeight) * float64(w) / float64(h))
```
agnivade, line 21: Maybe `targetWidth` or `thumbWidth` is a bit clearer. Same for height too.

agnivade (approved):
Looks good. Just a minor concern: since we are not imposing a restriction in image height or width, does the resize operation perform well in cases of bad inputs like these where the overall number of pixels is within limits but the individual dimensions are extremely big?

streamer45:
> Looks good. Just a minor concern: since we are not imposing a restriction in image height or width, does the resize operation perform well in cases of bad inputs like these where the overall number of pixels is within limits but the individual dimensions are extremely big?

Yes. As one would imagine, to process the image there's a loop on both dimensions since every pixel needs to be scanned so it's the product of the twos that constitutes most of the complexity.

agnivade:
Gentle ping on this @DSchalla

DSchalla (approved):

streamer45:
@lindalumitchell 

This PR fixes an issue with resizing certain images. I've provided integration tests to cover the affected cases.
Also left some QA steps in the Jira issue for manual verification.

lindalumitchell (approved):
Looks good to me.

PR #15540, opened 2020-09-18
MM-28803 - Cloud settings defaults

marianunez:
#### Summary
Update settings defaults specifically for cloud **ONLY on `cloud_beta` branch**. Ticket for the server to support setting defaults based on license [MM-28883](https://mattermost.atlassian.net/browse/MM-28883). 

**Settings changed:**
- Max Users for team: `1000`
- EnableCustomEmoji: `true`
- ExperimentalSideBar: `default_on`
- EnableBotAccountCreation: `true`
- RequireEmailVerification: `true`
- RestrictSystemAdmin: `true`
- TeammateNameDisplay: `FullName`

- ExperimentalGossip:
	- UseExperimentalGossip: `true`
	- EnableExperimentalGossipEncryption: `true`
- SqlSettings:
	- MaxIdleConns:`10`
	- MaxOpenConns `100`
- Email:
	- FeedbackEmail: `feedback@mattermost.com`
	- ReplyToAddress: `noreply@mattermost.com`
	- ConnectionSecurity: `TLS`

#### Ticket Link
[MM-28803](https://mattermost.atlassian.net/browse/MM-28803)

### DIFF
--- a/model/config.go
+++ b/model/config.go
@@ -344,12 +344,7 @@ type ServiceSettings struct {
 
 func (s *ServiceSettings) SetDefaults(isUpdate bool) {
 	if s.EnableEmailInvitations == nil {
-		// If the site URL is also not present then assume this is a clean install
-		if s.SiteURL == nil {
-			s.EnableEmailInvitations = NewBool(false)
-		} else {
-			s.EnableEmailInvitations = NewBool(true)
-		}
+		s.EnableEmailInvitations = NewBool(true)
 	}
 
 	if s.SiteURL == nil {
--- a/model/config.go
+++ b/model/config.go
@@ -102,7 +102,7 @@ const (
 	SERVICE_SETTINGS_DEFAULT_GFYCAT_API_SECRET  = "3wLVZPiswc3DnaiaFoLkDvB4X0IV6CpMkj4tf2inJRsBY6-FnkT08zGmppWFgeof"
 
 	TEAM_SETTINGS_DEFAULT_SITE_NAME                = "Mattermost"
-	TEAM_SETTINGS_DEFAULT_MAX_USERS_PER_TEAM       = 50
+	TEAM_SETTINGS_DEFAULT_MAX_USERS_PER_TEAM       = 1000
 	TEAM_SETTINGS_DEFAULT_CUSTOM_BRAND_TEXT        = ""
 	TEAM_SETTINGS_DEFAULT_CUSTOM_DESCRIPTION_TEXT  = ""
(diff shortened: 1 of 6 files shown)


marianunez:
Some of the defaults here are already being overridden by environment variables in the provisioner.

However, some of the ones overridden are still shown in the System Console with a message that has been overridden and the admin can not change. We should either remove the env var or hide them from the System Console. Those settings are: 

- ExperimentalSideBar: `default_on`
- EnableEmailInvitations: `true`
- RequireEmailVerification: `true`
- SendEmailNotifications: `true` (overridden with the same default value and just disables it in the System Console). 
- ExperimentalViewArchivedChannels: `true` (overridden with the same default value and just disables it in the System Console). 

0/5 if we keep the SQL, email and gossip settings just in the env vars, and remove from this PR.

model/config.go
```diff
@@ -679,7 +674,7 @@ func (s *ServiceSettings) SetDefaults(isUpdate bool) {
 	}
 
 	if s.ExperimentalChannelSidebarOrganization == nil {
-		s.ExperimentalChannelSidebarOrganization = NewString("disabled")
+		s.ExperimentalChannelSidebarOrganization = NewString("default_on")
```
marianunez, line 677: See https://github.com/mattermost/mattermost-server/pull/15540#issuecomment-695106160

jwilander (approved):
Looks good 👍

ethervoid (approved):
I'm seeing that we're only changing this in the `cloud-beta` branch so it LGMT

jwilander:
My guess is tests are failing because they expect certain defaults

marianunez:
> My guess is tests are failing because they expect certain defaults

Correct, been resetting the defaults for some tests but more keep popping up. 😩

marianunez:
> Some of the defaults here are already being overridden by environment variables in the provisioner.
> 
> However, some of the ones overridden are still shown in the System Console with a message that has been overridden and the admin can not change. We should either remove the env var or hide them from the System Console. Those settings are:
> 
> * ExperimentalSideBar: `default_on`
> * EnableEmailInvitations: `true`
> * RequireEmailVerification: `true`
> * SendEmailNotifications: `true` (overridden with the same default value and just disables it in the System Console).
> * ExperimentalViewArchivedChannels: `true` (overridden with the same default value and just disables it in the System Console).
> 
> 0/5 if we keep the SQL, email and gossip settings just in the env vars, and remove from this PR.

@joram thoughts on the above? Remove from env vars or hide in System Console?

jwilander:
I'd say hide in system console. I don't think we want cloud admins to ever be able to turn those settings off.

PR #15544, opened 2020-09-21
MM28858 Basic framework for use of feature flags. Enviroment variable overrides.

crspeller:
#### Summary
Basic framework for the use of feature flags. 
Basically just a configuration struct that takes defaults from code that can be overridden by environment variables. 
No fancy management system yet. 

Curious about future plans? See the epic: https://mattermost.atlassian.net/browse/MM-28843

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-28858
Epic: https://mattermost.atlassian.net/browse/MM-28843


api4/system_test.go
```diff
@@ -64,6 +65,36 @@ func TestGetPing(t *testing.T) {
 		})
 
 	})
+
+	t.Run("ping feature flag test", func(t *testing.T) {
+		resp, err := th.Client.DoApiGet(th.Client.GetSystemRoute()+"/ping", "")
+		require.Nil(t, err)
+		require.Equal(t, http.StatusOK, resp.StatusCode)
+		respBytes, err2 := ioutil.ReadAll(resp.Body)
```
lieut-data, line 73: nit: `err2` vs. `appErr` earlier?

api4/system_test.go
```diff
@@ -64,6 +65,36 @@ func TestGetPing(t *testing.T) {
...
+		require.Nil(t, err2)
+		respString := string(respBytes)
+		require.NotContains(t, respString, "TestFeatureFlag")
+
+		// Run the enviroment variable override code to test
+		os.Setenv("MM_FEATUREFLAGS_TESTFEATURE", "testvalue")
+		defer os.Unsetenv("MM_FEATUREFLAGS_TESTFEATURE")
+		memoryStore, cerr := config.NewMemoryStore()
```
lieut-data, line 81: Worth consolidating `cerr`, `err3`, and `err4` too?

config/common.go
```diff
@@ -77,7 +77,7 @@ func (cs *commonStore) set(newCfg *model.Config, allowEnvironmentOverrides bool,
 		}
 	}
 
-	if err := persist(cs.RemoveEnvironmentOverrides(newCfg)); err != nil {
+	if err := persist(cs.RemoveNonPersistable(newCfg)); err != nil {
```
lieut-data, line 80: I seem to recall that we did want to persist the current configuration -- effectively using the config as a cache for the event that the feature flag system went down and so to avoid "flopping" if a pod restarts. Can you remind me what's actually going to happen?

config/common_test.go
```diff
@@ -39,6 +39,7 @@ func init() {
 			DefaultClientLocale: sToP("en"),
 		},
 	}
+	minimalConfig.SetDefaults()
```
lieut-data, line 42: It's not super obvious why this is required? In particular, I recall `minimalConfig` supposedly representing the minimum number of "bits" defined to "pass", but with the defaults set, we no longer achieve that minimum.

config/unmarshal.go
```diff
@@ -139,13 +49,18 @@ func unmarshalConfig(r io.Reader, allowEnvironmentOverrides bool) (*model.Config
 		return nil, nil, errors.Wrapf(err, "failed to read")
 	}
 
-	var rawConfig interface{}
+	var rawConfig model.Config
```
lieut-data, line 52: If we unmarshal directly to `model.Config`, won't we get a `UnmarshalTypeError` if the persisted config happened to contain fields we no longer care about (or contained spurious extra fields)?

vendor/github.com/mattermost/viper/go.mod
```diff
@@ -1,5 +1,7 @@
 module github.com/mattermost/viper
 
+go 1.15
```
lieut-data, line 3: Intentional update?

config/unmarshal.go
lieut-data: As an aside, when we deprecate feature flags, do we intend to leave them there permanently? (I imagine yes, given semver requirements.)

config/unmarshal.go
```diff
@@ -33,99 +33,9 @@ func newViper(allowEnvironmentOverrides bool) *viper.Viper {
 		v.AutomaticEnv()
 	}
 
-	// Set zeroed defaults for all the config settings so that Viper knows what environment variables
```
lieut-data, line 36: Thrilled to see this go... but it worries me slightly ;P

What happens if I try to define `MM_PLUGINSETTINGS_PLUGINS_*` or some such? Do we correctly not panic / corrupt the config?

lieut-data (changes requested):
Lovely to see viper magic code go away, but a few clarifying comments along the way.

config/common.go
crspeller: There will be some persistence in order to make sure the cluster is in sync. Not part of this PR, will be part of the management system PR. It might involve changing this code so that it is persisted in some cases (such as only on MM cloud where we have DB control)

config/common_test.go
crspeller: Basically it's because some tests check that `unmarshalConfig` returns an exact match of `minimalConfig`. Since `unmarshalConfig` now returns the config with defaults applied, I needed to add a set defaults here to match.

vendor/github.com/mattermost/viper/go.mod
crspeller: No. Not sure why it did that.

config/unmarshal.go
crspeller: No. `UnmarshalTypeError` only occurs when the type is wrong. As long as we don't reuse names, this won't bite us. 
I verified the behavior is the same on master, the only difference is that it's the `json` library that reports the error rather than viper.  

For deprecation we remove them from the code, but they live forever in the management system. That way previous MM version can get the correct values.

config/unmarshal.go
crspeller: It doesn't seem to work before or after my changes.

lieut-data (approved):
Thanks for the clarifications!

config/common_test.go
```diff
@@ -73,16 +74,6 @@ func init() {
 	}
 }
 
-func prepareExpectedConfig(t *testing.T, expectedCfg *model.Config) *model.Config {
```
catalintomai, line 76: @crspeller , I miss the context, are these change and the Viper one related/prereq to this feature work or cleanup in the area?

model/config_test.go
```diff
@@ -1326,3 +1326,21 @@ func TestConfigMarketplaceDefaults(t *testing.T) {
 		require.Equal(t, "https://marketplace.example.com", *c.PluginSettings.MarketplaceUrl)
 	})
 }
+
+func TestSetDefaultFeatureFlagBehavour(t *testing.T) {
```
catalintomai, line 1330: to fix name.

config/common_test.go
crspeller: It's related. Basically the way we where doing the defaults for viper was causing one of my environment variable override tests to fail, but it was working overall. Me and Jesse traced it back to `set` in the common config store getting called and the overrides being applied again after set defaults. This change makes that behavior default to simplify the code and make this use case work.

model/config_test.go
crspeller: Sorry I don't understand this comment.

catalintomai: My bad, wasn't aware that behavour is a word, looks like it is the current behavior is an alteration of that.

config/common_test.go
catalintomai: Thanks for the explanation.

catalintomai (approved):

model/config_test.go
lieut-data: Oh, I see the typo now:
```suggestion
 func TestSetDefaultFeatureFlagBehaviour(t *testing.T) { 
```

prapti (approved):
LGTM. Worked as expected: any value other than "off", including empty string was successfully set on the server, in both TE and EE editions.
Approving.

PR #15547, opened 2020-09-22
MM-28189 - Sentry CI improvements

reflog:
#### Summary

1. Version/hash should now be set correctly, linking source code to Sentry events
2. Event sanitation added, removing query, params and data from reported URLs
3. Avoid crashing in error reporter

#### Ticket Link

https://mattermost.atlassian.net/browse/MM-28189
https://mattermost.atlassian.net/browse/MM-28527
https://mattermost.atlassian.net/browse/MM-28826


.circleci/config.yml
```diff
@@ -170,7 +170,8 @@ jobs:
             make build-cmd BUILD_NUMBER='${CIRCLE_BRANCH}-${CIRCLE_BUILD_NUM}'
             make package BUILD_NUMBER='${CIRCLE_BRANCH}-${CIRCLE_BUILD_NUM}'
             curl -sL https://sentry.io/get-cli/ | bash
-            export VERSION=$(sentry-cli releases propose-version)
+            export VERSION=`git rev-parse HEAD`
+            echo Sentry will use version "$VERSION"
```
agnivade, line 174: To be safe, I wouldn't export the variable to prevent the env space of forked processes from being polluted. We could just directly use the result of the command in the echo line itself.

model/utils.go
```diff
@@ -9,6 +9,7 @@ import (
 	"encoding/base32"
 	"encoding/json"
 	"fmt"
+	"github.com/getsentry/sentry-go"
```
agnivade, line 12: Please move this to the 3rd party imports group.

model/utils.go
```diff
@@ -70,6 +71,12 @@ type AppError struct {
 }
 
 func (er *AppError) Error() string {
+	// below is a short-circuit to avoid crashing during error reporting
+	// if by some chance the error is nil, we don't want crash, but we do report the problem to Sentry
+	if er == nil {
+		sentry.CaptureMessage("Error() is called on a nil AppError instance")
```
agnivade, line 77: Let's dedup the error message.

Also, I am not clear how this can happen only from error reporting code. The crash can happen from anywhere in the codebase which does `var err *model.AppError; _ = err.Error()` right? Shouldn't we be crashing in that case?

reflog: @agnivade - it was decided that crashing during error reporting is not the best idea. So instead we just report it to Sentry and proceed. 
Since this happens in error handling anyways, code flow will not change, even if the crash is avoided.

agnivade: Oh my apologies, I misunderstood. This is not exactly "error in error reporting". Yes the crash is from inside the error reporting code, but the bug comes from the actual code. @grundleborg - is this what you understood on the call too? I'm not sure I fully agree with this. This suppresses real bugs where the code is supposed to crash.

And if we are doing this, why just limit ourselves to `Error` method, and not extend it to `SystemMessage` and `ToJson`?

Sorry to be asking such questions when we already decided this. Just clarifying that we are all on the same page.

grundleborg: Yeah I don't think I understood this right in the meeting. @reflog looks like you removed this block - what's the current thinking here?

reflog: @grundleborg " removed this block " <-- what do you mean?

model/utils.go
```diff
@@ -70,6 +71,13 @@ type AppError struct {
 }
 
 func (er *AppError) Error() string {
+	// below is a short-circuit to avoid crashing during error reporting
+	// if by some chance the error is nil, we don't want crash, but we do report the problem to Sentry
+	if er == nil {
+		msg := "Error() is called on a nil AppError instance"
+		sentry.CaptureMessage(msg)
```
grundleborg, line 78: Will we still get a stack trace indicating where in the code this was invoked by the nil error? Or does this just capture the text of the message above?

reflog: this captures the stacktrace as well

model/utils.go
grundleborg: I think I got confused by the "outdated" block.

grundleborg: @reflog @agnivade so to try and make sure I understand this right: my interpretation is that this will catch any crash where `err.Error()` is called on a `nil` instance of `*model.AppError`. Is that correct? So it then becomes "when would this happen"? Would it only happen when handling an error anyway? Seems like that may be a reasonable, although not definite, argument... in that case it might be acceptable to assume that. Would you ever call `err.Error()` anywhere else than in error reporting code? I assume that's the argument you are making Eli?

reflog: That is correct, that's argument I am making I guess :)

agnivade: > Would you ever call err.Error() anywhere else than in error reporting code?

Yeah .. but that doesn't necessarily mean something bad has happened and it's better to quit anyways. Case in point: https://github.com/mattermost/mattermost-server/commit/095e5d788b11f0f12077484c42ad3c6e17d09cbd. And this is just one example. There are other cases where no rows is returned as an error, and that is a completely valid result which needs to be inspected and handled.

I am not 5/5 against this. But I just want to clarify that there are some unknowns here and we cannot be 100% sure if _not_ crashing is the right thing.

reflog: @agnivade - can you elaborate on specific issue you see in https://github.com/mattermost/mattermost-server/commit/095e5d788b11f0f12077484c42ad3c6e17d09cbd ? I might be missing something obvious :)

agnivade: I'll have to dig deep into why it was happening. But what I was trying to say is that "joining a team when a default channel is archived" is a normal thing to do, and we cannot be 100% sure if returning a dummy error string instead of panicking is always the right thing to do.

Don't want to be _that_ guy blocking a simple thing. So if you and George are okay with this, no objections from my side :)

grundleborg: I'm on the fence - I'm happy to approve this in the absence of counter-examples, but if the possibility of creating unexpected behaviour as a result of this is high, or there are definite counter-examples, I'd prefer not to stop the panic here.

reflog: @grundleborg don't get me wrong, I am not 100% onboard either :) Let's discuss this more in Server meeting?

grundleborg: Good idea!

reflog:
@agnivade @grundleborg I've removed the error related hack as discussed. Please re-review

agnivade (approved):

grundleborg (approved):

lindalumitchell (approved):

PR #15553, opened 2020-09-22
MM-28548: Change default data source to postgres

agnivade:
We now move to start using postgres as our default database.

https://mattermost.atlassian.net/browse/MM-28548


model/config.go
```diff
@@ -107,7 +107,7 @@ const (
 	TEAM_SETTINGS_DEFAULT_CUSTOM_DESCRIPTION_TEXT  = ""
 	TEAM_SETTINGS_DEFAULT_USER_STATUS_AWAY_TIMEOUT = 300
 
-	SQL_SETTINGS_DEFAULT_DATA_SOURCE = "mmuser:mostest@tcp(localhost:3306)/mattermost_test?charset=utf8mb4,utf8&readTimeout=30s&writeTimeout=30s"
+	SQL_SETTINGS_DEFAULT_DATA_SOURCE = "postgres://mmuser:mostest@localhost/mattermost_test?sslmode=disable"
```
streamer45, line 110: Should we retain the 10s `connect_timeout` parameter in the dsn? 
I see we use that regularly in the documentation and throughout the code.

grundleborg:
@agnivade does any build/release pipeline stuff need an update for this? Also, if this changes the default for release packages, do we need to update docs?

agnivade:
@grundleborg 

> does any build/release pipeline stuff need an update for this? 

Nope, this just changes the default dsn for config.json, if you don't set anything. CI pipelines already set their connection string explicitly.

> Also, if this changes the default for release packages, do we need to update docs?

Hmm .. okay. Do you mean like a release note announcement or update our existing docs? I couldn't really find if there is a place where we document our default DB choice somewhere.

streamer45 (approved):
Looks good :+1:

grundleborg:
@agnivade I checked the docs and it seems fine - both end user and developer docs cover both databases already.

What I meant about CI is I could imagine a sed script in the release process that changes the default connection string, which might get broken by you changing it here.

agnivade:
Gotcha. I think the circleCI stuff is fine. @metanerd - Is there anything in the Jenkins pipeline that can be affected by this change?

metanerd:
https://github.com/mattermost/mattermost-shared-pipelines/search?q=tcp%2F

Not sure about other.

agnivade:
So it looks like the sed command is only changing config.json and not the code directly. That should be fine then. @grundleborg - this should be good to go then.

grundleborg:
@agnivade I might be missing something, but I think `config.json` is generated from these go defaults, so that sed script will fail once this PR is merged unless it is updated.

agnivade:
Clarified offline. @metanerd - If the sed command [here](https://github.com/mattermost/mattermost-shared-pipelines/blob/105581887be9d74c49bf20bfff65c9abdd51fc70/src/org/mattermost/PlatformStages.groovy#L334) is working on a default config.json then we need to change the line to `postgres://mmuser:mostest@localhost/mattermost_test?sslmode=disable&connect_timeout=10`.

But if it doesn't, then we don't need to do any changes. I do see that there are no readTimeout and writeTimeout specified, which is the default string in the config. So maybe it's some other config.json. I am not sure. I'll let you handle that.

grundleborg (approved):

agnivade:
Gentle bump on this @metanerd. Let me know if this is good to go or the groovy file needs to be updated.

metanerd:
0/5 if it needs modification honestly. Let's go ahead with it.

amyblais:
@agnivade Does this need a release note for v5.30?

agnivade:
This is only applicable to developers. It is unlikely customers will be affected by this.

PR #15558, opened 2020-09-22
[MM-28536] [MM-28535] sysconsole_write_authentication bug fixes 

fmunshi:
#### Summary
- These endpoints should not require manage_system instead they should just check `sysconsole_write_auth...` 

#### Ticket Link
- https://mattermost.atlassian.net/browse/MM-28536
- https://mattermost.atlassian.net/browse/MM-28535

### DIFF
--- a/api4/team_test.go
+++ b/api4/team_test.go
@@ -3352,19 +3352,22 @@ func TestInvalidateAllEmailInvites(t *testing.T) {
 	defer th.TearDown()
 
 	t.Run("Forbidden when request performed by system user", func(t *testing.T) {
-		_, res := th.Client.InvalidateAllEmailInvites()
+		ok, res := th.Client.InvalidateEmailInvites()
+		require.Equal(t, ok, false)
 		CheckForbiddenStatus(t, res)
 	})
 
 	t.Run("OK when request performed by system user with requisite system permission", func(t *testing.T) {
 		th.AddPermissionToRole(model.PERMISSION_SYSCONSOLE_WRITE_AUTHENTICATION.Id, model.SYSTEM_USER_ROLE_ID)
 		defer th.RemovePermissionFromRole(model.PERMISSION_SYSCONSOLE_WRITE_AUTHENTICATION.Id, model.SYSTEM_USER_ROLE_ID)
-		_, res := th.Client.InvalidateAllEmailInvites()
+		ok, res := th.Client.InvalidateEmailInvites()
+		require.Equal(t, ok, true)
 		CheckOKStatus(t, res)
 	})
 
 	t.Run("OK when request performed by system admin", func(t *testing.T) {
-		_, res := th.SystemAdminClient.InvalidateAllEmailInvites()
+		ok, res := th.SystemAdminClient.InvalidateEmailInvites()
+		require.Equal(t, ok, true)
 		CheckOKStatus(t, res)
 	})
 }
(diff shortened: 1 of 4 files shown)


api4/team_test.go
```diff
@@ -3346,3 +3346,25 @@ func TestTeamMembersMinusGroupMembers(t *testing.T) {
...
+func TestInvalidateAllEmailInvites(t *testing.T) {
+	th := Setup(t).InitBasic()
+	defer th.TearDown()
+
+	t.Run("Forbidden when request performed by system user", func(t *testing.T) {
+		_, res := th.Client.InvalidateAllEmailInvites()
+		CheckForbiddenStatus(t, res)
+	})
+
+	t.Run("OK when request performed by system admin with requisite system permission", func(t *testing.T) {
```
fmunshi, line 3359:
```suggestion
	t.Run("OK when request performed by system user with requisite system permission", func(t *testing.T) {
```

model/client4.go
```diff
@@ -5605,3 +5605,13 @@ func (c *Client4) UploadData(uploadId string, data io.Reader) (*FileInfo, *Respo
...
+
+// InvalidateAllEmailInvites invalidates all email invites.
+func (c *Client4) InvalidateAllEmailInvites() (bool, *Response) {
+	r, err := c.DoApiDelete(c.GetTeamsRoute() + "/invites/email")
+	if err != nil {
+		return false, BuildErrorResponse(r, err)
+	}
+	defer closeBody(r)
+	return CheckStatusOK(r), BuildResponse(r)
+}
```
streamer45, line 5617: I believe this existed already. Did you miss it or are we duplicating it on purpose?

https://github.com/mattermost/mattermost-server/blob/96d3ddd4f10662cca856c2c197629f58527d16f9/model/client4.go#L2215-L2223

api4/team_test.go
```diff
@@ -3346,3 +3346,25 @@ func TestTeamMembersMinusGroupMembers(t *testing.T) {
...
 	}
 }
+
+func TestInvalidateAllEmailInvites(t *testing.T) {
+	th := Setup(t)
+	defer th.TearDown()
+
+	t.Run("Forbidden when request performed by system user", func(t *testing.T) {
+		_, res := th.Client.InvalidateAllEmailInvites()
+		CheckForbiddenStatus(t, res)
```
streamer45, line 3356: I'd add a check to verify that in case of error the returned boolean is false.

model/client4.go
fmunshi: I missed it, thanks for catching this, `cmd+shift+f` failed me 😛

api4/team_test.go
```diff
@@ -3352,19 +3352,22 @@ func TestInvalidateAllEmailInvites(t *testing.T) {
 	defer th.TearDown()
 
 	t.Run("Forbidden when request performed by system user", func(t *testing.T) {
-		_, res := th.Client.InvalidateAllEmailInvites()
+		ok, res := th.Client.InvalidateEmailInvites()
+		require.Equal(t, ok, false)
```
streamer45, line 3356: Very minor detail but I've been hit by this in the past while debugging test failures:
these assertions assume the first value passed is the *expected value* while the second is the *actual value* we want to verify. This makes a difference in the final test output in case of a failure so I'd probably swap them.

streamer45 (approved):
Looks good! :+1:
Left a minor, non-blocking suggestion.

catalintomai (approved):

srkgupta (approved):
Tested and both the issues are fixed now and are working fine now. Approving the PR.

PR #15562, opened 2020-09-22
MM-28530 - Changes are not automatically propagated to system console when a permission is added/removed from a role

hahmadia:
#### Summary

When a role is patched, no websocket event is sent (so we need referesh on webapp).
This PR sends the websocket event and that means you don't need to refresh on front end to see the role take effect.

#### Ticket Link

https://mattermost.atlassian.net/browse/MM-28530

### DIFF
--- a/app/role.go
+++ b/app/role.go
@@ -171,7 +171,7 @@ func (a *App) UpdateRole(role *model.Role) (*model.Role, *model.AppError) {
 		model.CHANNEL_ADMIN_ROLE_ID,
 	}
 
-	builtInRolesMinusChannelRoles := utils.RemoveStringsFromSlice(model.BuiltInSchemeManagedRoleIDs, builtInChannelRoles...)
+	builtInRolesMinusChannelRoles := append(utils.RemoveStringsFromSlice(model.BuiltInSchemeManagedRoleIDs, builtInChannelRoles...), model.NewSystemRoleIDs...)
 
 	if utils.StringInSlice(savedRole.Name, builtInRolesMinusChannelRoles) {
 		return savedRole, nil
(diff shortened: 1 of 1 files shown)


app/role.go
```diff
@@ -125,6 +125,8 @@ func (a *App) PatchRole(role *model.Role, patch *model.RolePatch) (*model.Role,
 		return nil, err
 	}
 
+	a.sendUpdatedRoleEvent(role)
```
fmunshi, line 128: a.UpdateRole is also _sometimes_ sending a role updated event... can we have that function skip the current role on line 212? basically add a condition here
```
	for _, ir := range impactedRoles {
		a.sendUpdatedRoleEvent(ir)
	}
``` 

changed to 

```
	for _, ir := range impactedRoles {
            if ir.Name != role.Name {
		a.sendUpdatedRoleEvent(ir)
            }
	}
```

Also can we change line 172 to be 
`builtInRolesMinusChannelRoles := append(utils.RemoveStringsFromSlice(model.BuiltInSchemeManagedRoleIDs, builtInChannelRoles...), model.NewSystemRoleIDs...)`? to ignore all the channel moderation related logic if its a new system role

hahmadia: Done!

fmunshi (approved):

sbishel (approved):

srkgupta (changes requested):
Hi @hahmadia 
I am trying to test the PR and the issue is still not fixed and I still have to reload the page for role changes to be reflected. Can you please recheck if the fix is tested and is working fine?

hahmadia:
Got on a call with @srkgupta and it seemed a reset was all that was needed. It's now working fine for him and he will continue his testing.

srkgupta:
Sorry for the confusion @hahmadia. It's working fine now. For some reason, it wasn't working when I last tested.

srkgupta (approved):
Thanks @hahmadia. Tested the PR and the issue is fixed now and is working fine. Approving the PR.

PR #15571, opened 2020-09-23
MM-28247 Threads metadata table

reflog:


store/sqlstore/thread_store.go
```diff
@@ -0,0 +1,180 @@
...
+		}
+	}
+	_, err := s.GetMaster().Exec("DELETE FROM Threads WHERE PostId = :Id", map[string]interface{}{"Id": postId})
+	if err != nil {
+		return errors.Wrap(err, "failed to update Threads")
+	}
+	return nil
+}
+
+func (s *SqlThreadStore) UpdateFromPosts(transaction *gorp.Transaction, posts []*model.Post) error {
```
jespino, line 75: Is there any reason to expose the transaction outside the store?

reflog: Thread store is sort of an addon to post store. Not passing transaction will cause inconsistent behavior

reflog: @jespino - I didn't like this as well, so I reworked the code a bit to move post-store-related stuff inside Post store, so they are not part of the interface now.

store/sqlstore/post_store.go
```diff
@@ -159,22 +160,33 @@ func (s *SqlPostStore) SaveMultiple(posts []*model.Post) ([]*model.Post, int, er
...
+
+	defer finalizeTransaction(transaction)
+
+	if _, err = transaction.Exec(query, args...); err != nil {
 		return nil, -1, errors.Wrap(err, "failed to save Post")
 	}
 
 	for channelId, count := range channelNewPosts {
-		if _, err := s.GetMaster().Exec("UPDATE Channels SET LastPostAt = GREATEST(:LastPostAt, LastPostAt), TotalMsgCount = TotalMsgCount + :Count WHERE Id = :ChannelId", map[string]interface{}{"LastPostAt": maxDateNewPosts[channelId], "ChannelId": channelId, "Count": count}); err != nil {
+		if _, err = transaction.Exec("UPDATE Channels SET LastPostAt = GREATEST(:LastPostAt, LastPostAt), TotalMsgCount = TotalMsgCount + :Count WHERE Id = :ChannelId", map[string]interface{}{"LastPostAt": maxDateNewPosts[channelId], "ChannelId": channelId, "Count": count}); err != nil {
```
jespino, line 175: This was intentionally excluded from the transaction because this is "fire and forget", if something gets wrong there, will be eventually consistent with later changes.

store/sqlstore/post_store.go
```diff
@@ -159,22 +160,33 @@ func (s *SqlPostStore) SaveMultiple(posts []*model.Post) ([]*model.Post, int, er
...
 			mlog.Error("Error updating Channel LastPostAt.", mlog.Err(err))
 		}
 	}
 
 	for rootId := range rootIds {
-		if _, err := s.GetMaster().Exec("UPDATE Posts SET UpdateAt = :UpdateAt WHERE Id = :RootId", map[string]interface{}{"UpdateAt": maxDateRootIds[rootId], "RootId": rootId}); err != nil {
+		if _, err = transaction.Exec("UPDATE Posts SET UpdateAt = :UpdateAt WHERE Id = :RootId", map[string]interface{}{"UpdateAt": maxDateRootIds[rootId], "RootId": rootId}); err != nil {
```
jespino, line 181: Same here.

store/sqlstore/post_store.go
```diff
@@ -159,22 +160,33 @@ func (s *SqlPostStore) SaveMultiple(posts []*model.Post) ([]*model.Post, int, er
...
 			mlog.Error("Error updating Post UpdateAt.", mlog.Err(err))
 		}
 	}
 
+	if err = s.updateThreadsFromPosts(transaction, posts); err != nil {
```
jespino, line 186: I think here is where makes sense to use the transaction, because it is not idempotent, so a transaction failure doesn't get fixed eventually. I would move this up before the other requests, and if the transaction goes well, fire and forget the other 2 updates.

store/sqlstore/post_store.go
```diff
@@ -188,11 +200,16 @@ func (s *SqlPostStore) SaveMultiple(posts []*model.Post) ([]*model.Post, int, er
 	}
 
 	if len(unknownRepliesPosts) > 0 {
-		if err := s.populateReplyCount(unknownRepliesPosts); err != nil {
+		if err := s.populateReplyCount(transaction, unknownRepliesPosts); err != nil {
```
jespino, line 203: Again, this was intentionally fire and forget, so shoulden't be part of the same transaction.

store/sqlstore/post_store.go
```diff
@@ -188,11 +200,16 @@ func (s *SqlPostStore) SaveMultiple(posts []*model.Post) ([]*model.Post, int, er
...
 			mlog.Error("Unable to populate the reply count in some posts.", mlog.Err(err))
 		}
 	}
 
+	if err := transaction.Commit(); err != nil {
```
jespino, line 208: I would move it up too, before the "fire and forget" queries.

store/sqlstore/post_store.go
```diff
@@ -551,6 +602,18 @@ func (s *SqlPostStore) PermanentDeleteByUser(userId string) error {
 }
 
 func (s *SqlPostStore) PermanentDeleteByChannel(channelId string) error {
+	results := []postIds{}
+	_, err := s.GetMaster().Select(&results, "SELECT Id, RootId, UserId FROM Posts WHERE ChannelId = :ChannelId", map[string]interface{}{"ChannelId": channelId})
```
jespino, line 606: I don't love this selects, but I don't know if there is other way to solve it.

store/sqlstore/post_store.go
```diff
@@ -1858,3 +1921,79 @@ func (s *SqlPostStore) GetOldestEntityCreationTime() (int64, error) {
...
+		// skip if post is not a part of a thread
+		if len(post.RootId) == 0 {
+			continue
+		}
+		postsByRoot[post.RootId] = append(postsByRoot[post.RootId], post)
+	}
+	now := model.GetMillis()
+	for rootId, posts := range postsByRoot {
+		var thread model.Thread
+		if err := transaction.SelectOne(&thread, "SELECT * from Threads WHERE PostId=:PostId", map[string]interface{}{"PostId": rootId}); err != nil {
```
jespino, line 1960: I think is not super important (probably only going to affect bulk import), but we could just select everything in one single query instead of one query per root post. But I think that is not critical, because is not in the critical path.

store/sqlstore/thread_store.go
```diff
@@ -0,0 +1,103 @@
...
+}
+
+func newSqlThreadStore(sqlStore SqlStore) store.ThreadStore {
+	s := &SqlThreadStore{
+		SqlStore: sqlStore,
+	}
+
+	for _, db := range sqlStore.GetAllConns() {
+		table := db.AddTableWithName(model.Thread{}, "Threads").SetKeys(false, "PostId")
+		table.ColMap("PostId").SetMaxSize(26)
```
jespino, line 27: Can we explicitly define here the `Who` field? I don't want surprises about the default field size.

store/sqlstore/thread_store.go
```diff
@@ -0,0 +1,103 @@
...
+	if _, err := s.GetMaster().Update(thread); err != nil {
+		return nil, errors.Wrapf(err, "failed to update thread with id=%s", thread.PostId)
+	}
+
+	return thread, nil
+}
+
+func (s *SqlThreadStore) Get(id string) (*model.Thread, error) {
+	var thread model.Thread
+	err := s.GetReplica().SelectOne(&thread, "SELECT * from Threads WHERE PostId=:PostId", map[string]interface{}{"PostId": id})
```
jespino, line 86: This are clearly small enough queries to be written in plain SQL, but maybe we want to move it to squirrel.

store/sqlstore/upgrade.go
```diff
@@ -849,6 +849,8 @@ func upgradeDatabaseToVersion528(sqlStore SqlStore) {
 		sqlStore.AlterColumnTypeIfExists("IncomingWebhooks", "Username", "varchar(255)", "varchar(255)")
 		sqlStore.AlterColumnTypeIfExists("IncomingWebhooks", "IconURL", "text", "varchar(1024)")
 
+		sqlStore.AlterColumnTypeIfExists("Threads", "Who", "longtext", "text")
```
jespino, line 852: This should happen in a new migration, not the 27 to 28 one. And, probably we don't need it if we define it properly in the sqlstore.

store/storetest/thread_store.go
```diff
@@ -0,0 +1,141 @@
...
+
+import (
+	"github.com/mattermost/mattermost-server/v5/model"
+	"github.com/mattermost/mattermost-server/v5/store"
+	"github.com/stretchr/testify/require"
+	"testing"
+)
+
+func TestThreadStore(t *testing.T, ss store.Store, s SqlSupplier) {
+	t.Run("SaveMultiple", func(t *testing.T) { testThreadStoreSaveMultiple(t, ss) })
```
jespino, line 14: What about tests around deletion? What about deleting the root post?

reflog:
@jespino thanks for a thorough review. all great points. Hopefully I addressed them all

store/storetest/thread_store.go
```diff
@@ -0,0 +1,200 @@
...
```
ashishbhate, line 14: Test name and test function name should be renamed after addition of additional tests in 2b365a8884dd5f5e87783338c852fcd89d01f655

ashishbhate (approved):
Apologies for the delay in reviewing. :100: LGTM, some minor comments inline.

jespino (approved):
LGTM. Thanks @reflog! :tada:

amyblais:
@esethna Is this one of the back-end server work for threading that should be cherry-picked to v5.29?

esethna:
@amyblais yes please

store/sqlstore/thread_store.go
```diff
@@ -0,0 +1,108 @@
...
+		thread.PostId,
+		thread.LastReplyAt,
+		thread.ReplyCount,
+		thread.Participants,
+	}
+}
+
+func (s *SqlThreadStore) createIndexesIfNotExists() {
+	s.CreateIndexIfNotExists("idx_threads_last_reply_at", "Threads", "LastReplyAt")
+	s.CreateIndexIfNotExists("idx_threads_post_id", "Threads", "PostId")
```
mkraft, line 51: This index is automatically created by both MySQL and Postgres as part of `PostId` being the primary key for the table.  Creating this is unnecessary and will lead to increased storage, slower backups and restores, and slower writes.

reflog: @mkraft yikes! thanks for noticing this. I'll remove in the next PR related to this feature.

esethna:
@amyblais did this get cherry picked?

amyblais:
@esethna Yes https://github.com/mattermost/mattermost-server/pull/16183

store/sqlstore/post_store.go
```diff
@@ -495,19 +515,49 @@ func (s *SqlPostStore) Delete(postId string, time int64, deleteByID string) erro
 		return errors.Wrap(err, "failed to update Posts")
 	}
 
-	return nil
+	return s.cleanupThreads(post.Id, post.RootId, post.UserId)
```
mkraft, line 518: Since cleaning up child thread posts is not an invariant in the system (for example in the case of data retention where you delete the parent but not the children), coupled with the fact that this performs another separate SQL query, I think this should be performed in the App tier rather than here in the data tier.

mkraft: Also, was this a breaking change to the delete post API endpoint or did it always also delete child posts?

reflog: @mkraft - this was not a breaking change. Regarding data retention - we're aware that collapsed threads do not work with retention enabled at the moment, we have a ticket tracking it

PR #15577, opened 2020-09-23
[MM-28447] Avoid "ghost mentions" when "First Name" is not set, but "First Name trigger mention" setting is set

### LINKED ISSUE #15245: Emptying first name triggers notification flood
#### Summary
After emptying your first name, and if it was a word that triggers mentions, you will receive a notification for nearly all new messages.

#### Steps to reproduce
1. Set up a first name under _General Settings > Full Name_
2. Turn on the check box _Notifications > Words That Trigger Mentions > Your case sensitive first name_
3. Delete your _General Settings > Full Name_
4. The checkbox _Words That Trigger Mentions > Your case sensitive first name_ disappeared from the interface
5. Wait for a while until new messages come in.
Mattermost Version: 5.23.1
Database Schema Version: 5.23.0
Database: mysql

#### Expected behavior
The user should only receive notifications for appropriate @ mentions.

The user will receive notifications for nearly all new messages, including the ones that don't contain mentions or are in threads that he/she didn't participate in.

#### Possible fixes
I'm only guessing here, but it seems that the option _Words That Trigger Mentions > Your case sensitive first name_ remains activated after emptying the first name. Because the name is now an empty string, and basically any message contains an empty string, the user will get flooded with notifications for just about any message. The issue could possibly be fixed if _Words That Trigger Mentions > Your case sensitive first name_ was deactivated when the name is emptied.

larkox:
#### Summary
If we have the "First name trigger mention" setting set, but we don't have a "First Name" set, we start receiving "ghost mentions" (we get the mention, but when we get to the channel, nothing is highlighted as a mention).

The easiest way to reproduce this was using the dot (`.`) or the underscore (`_`), which apparently always trigger the mention. Not sure if there are other characters or compositions that trigger mentions.

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-28447
Fix #15245

### DIFF
--- a/app/notification.go
+++ b/app/notification.go
@@ -873,7 +873,7 @@ func addMentionKeywordsForUser(keywords map[string][]string, profile *model.User
 	}
 
 	// If turned on, add the user's case sensitive first name
-	if profile.NotifyProps[model.FIRST_NAME_NOTIFY_PROP] == "true" {
+	if profile.NotifyProps[model.FIRST_NAME_NOTIFY_PROP] == "true" && profile.FirstName != "" {
 		keywords[profile.FirstName] = append(keywords[profile.FirstName], profile.Id)
 	}
 
--- a/app/notification_test.go
+++ b/app/notification_test.go
@@ -1424,6 +1424,24 @@ func TestAddMentionKeywordsForUser(t *testing.T) {
 		assert.NotContains(t, keywords["Robert"], user.Id)
 	})
 
+	t.Run("should not add case sensitive first name if enabled but empty First Name", func(t *testing.T) {
+		user := &model.User{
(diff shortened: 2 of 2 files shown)


fmunshi (approved):

fmunshi:
Could add a simple unit test for this case too @larkox

fmunshi (approved):
perfect

hahmadia (approved):
LGTM! Thanks! 
THANKYOU

srkgupta (approved):
I was able to reproduce the original issue on the latest master. And I had retested this on top of this PR and the issue is fixed now and is working fine. Approving the PR. Thanks @larkox

PR #15585, opened 2020-09-24
MM-29011: Add feature flag for shared channels

agnivade:
https://mattermost.atlassian.net/browse/MM-29011


wiggin77 (approved):
LGTM.  Good call adding it to client config as well.  I'll add something similar to `License.Features` via another ticket.

agnivade:
> Good call adding it to client config as well.

Yeah, otherwise the system console can't check that and gate the UI display depending on the config.

> I'll add something similar to License.Features via another ticket.

Yeah, I was thinking about that. Sounds good!

streamer45 (approved):
Looks good :+1:

agnivade:
Gentle ping on this @lindalumitchell. Just adds a new config setting with no visible result on the outside.

lindalumitchell (approved):
So sorry for the delay on this one! To not keep it any longer, I'll QA-defer it and check it after merge on a test server there (the old-style test server doesn't seem to be building, and I think that's the way I would be able to SSH in, but let's not hold this up as it's such a small change).

PR #15587, opened 2020-09-24
MM-28890 Enable the gif picker, timezones, and viewing archived channels by default

hmhealey:
Similar to the last PR to enable custom emoji by default, these are all ones we want turned on to help improve NPS for end users since admins tend not to customize stuff like this.

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-28990

### DIFF
--- a/model/config.go
+++ b/model/config.go
@@ -626,7 +626,7 @@ func (s *ServiceSettings) SetDefaults(isUpdate bool) {
 	}
 
 	if s.EnableGifPicker == nil {
-		s.EnableGifPicker = NewBool(false)
+		s.EnableGifPicker = NewBool(true)
 	}
 
 	if s.GfycatApiKey == nil || *s.GfycatApiKey == "" {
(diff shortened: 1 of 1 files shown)


crspeller (approved):

jgilliam17:
@hmhealey 
Verified Gif and Timezones are set True by default and can be turned off by the sysadmin.
There's an issue with Archived setting. It is set to true through an env. variable, preventing sysadmins from being able to change it.
Screen Shot 2020-09-25 at 8 56 49 AM

hmhealey:
@jgilliam17 That's just something with how our test servers are set up. Environment variables are configured for the server outside of Mattermost. We can probably just skip testing that one then since the changes for it are identical to the ones for the other setting.

jgilliam17 (approved):
Thank you @hmhealey for clarifying 
Tested, looks good to merge.
- Verified gif picker, timezones and viewing archived channels is set to true by default for new installs.

hmhealey:
Manually cherry-picked

PR #15598, opened 2020-09-25
MM-27523: Remove ability to read/write admin filter settings.

mkraft:
#### Summary

Prevents non-system-admins from viewing or updating the LDAP admin filter settings.

#### Ticket Link

https://mattermost.atlassian.net/browse/MM-27523

### DIFF
--- a/model/config.go
+++ b/model/config.go
@@ -1931,8 +1931,8 @@ type LdapSettings struct {
 	UserFilter        *string `access:"authentication"`
 	GroupFilter       *string `access:"authentication"`
 	GuestFilter       *string `access:"authentication"`
-	EnableAdminFilter *bool   `access:"authentication"`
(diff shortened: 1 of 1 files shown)


sbishel (approved):

hahmadia (approved):

srkgupta (approved):
Tested and the user with sysconsole_write_authentication permission is not allowed to read/write admin filter settings from the server. However this still needs to be fixed in UI and hence I have raised the following issue which needs to be fixed in the future versions, i.e. 

[MM-29089](https://mattermost.atlassian.net/browse/MM-29089) - Admin Filter option should be disabled in AD/LDAP page for admin roles with permission sysconsole_write_authentication

Tested and the PR is working fine. Approving the PR.

PR #15601, opened 2020-09-25
MM-27909: Add manage_shared_channels permission

agnivade:
We add a new permission to manage shared channels.

It's a channel scoped permission and only the system admin
has that by default.

https://mattermost.atlassian.net/browse/MM-27909


wiggin77 (approved):
LGTM

app/permissions_migrations.go
```diff
@@ -69,6 +69,7 @@ const (
 	PERMISSION_READ_PUBLIC_CHANNEL_GROUPS        = "read_public_channel_groups"
 	PERMISSION_READ_PRIVATE_CHANNEL_GROUPS       = "read_private_channel_groups"
 	PERMISSION_EDIT_BRAND                        = "edit_brand"
+	PERMISSION_MANAGE_SHARED_CHANNELS            = "manage_shared_channels"
```
mkraft, line 72: Why do we always re-add permission names here rather than just using, say, `PERMISSION_MANAGE_SHARED_CHANNELS.Id` in the migration functions?

mkraft (commented):
Could you elaborate on how this permission is going to be used? I'd like to reconcile my understanding of the name "manage shared channels" (plural) with the fact that it's channel-scoped.

app/permissions_migrations.go
agnivade: Absolutely no reason other than to maintain consistency with the old style. We could do it of course, and I am 0/5 on it. Let me know what you prefer.

agnivade:
> Could you elaborate on how this permission is going to be used? I'd like to reconcile my understanding of the name "manage shared channels" (plural) with the fact that it's channel-scoped.

To quote @wiggin77,

> This permission allows someone to share a channel, invite another cluster to sync with the shared channel, unshare a channel.

https://github.com/mattermost/mattermost-webapp/pull/6555#discussion_r499652608. So everything is channel-scoped.

Other entities also seem to be mentioned in plural in the documentation:

```
"admin.permissions.group.edit_posts.description": "Edit own and others' posts.",
  "admin.permissions.group.edit_posts.name": "Edit Posts",
  "admin.permissions.group.guest_create_post.description": "Allow users to create posts.",
  "admin.permissions.group.guest_create_post.name": "Create Posts",
  "admin.permissions.group.guest_create_private_channel.description": "Create new private channels.",
  "admin.permissions.group.guest_create_private_channel.name": "Create Channels",
```

app/permissions_migrations.go
mkraft: I have also added to this pattern recently. I think the next person to touch this (likely someone from Enterprise) can take care of cleaning it up, no need to do it as part of this PR. I wasn't sure if I was not seeing a reason.

mkraft:
> > Could you elaborate on how this permission is going to be used? I'd like to reconcile my understanding of the name "manage shared channels" (plural) with the fact that it's channel-scoped.
> 
> To quote @wiggin77,
> 
> > This permission allows someone to share a channel, invite another cluster to sync with the shared channel, unshare a channel.
> 
> [mattermost/mattermost-webapp#6555 (comment)](https://github.com/mattermost/mattermost-webapp/pull/6555#discussion_r499652608). So everything is channel-scoped.
> 
> Other entities also seem to be mentioned in plural in the documentation:
> 
> ```
> "admin.permissions.group.edit_posts.description": "Edit own and others' posts.",
>   "admin.permissions.group.edit_posts.name": "Edit Posts",
>   "admin.permissions.group.guest_create_post.description": "Allow users to create posts.",
>   "admin.permissions.group.guest_create_post.name": "Create Posts",
>   "admin.permissions.group.guest_create_private_channel.description": "Create new private channels.",
>   "admin.permissions.group.guest_create_private_channel.name": "Create Channels",
> ```

So it _is_ meant to be a channel-scoped permission. In that case I think the plural on "channels" is confusing because the permission is only granting privileges to a single channel. "Edit posts" is plural because many posts can be affected by the permission (all the posts in the channel). 

I could see "channels" being plural in the name if the permission were team-scoped or system-scoped. Then the permission would be granting privileges to multiple channels (either all the channels in the team or all the channels in the system). 

The permission names are used in code, the CLI, and in mmctl. And since the `Scope` field isn't always obvious the name is often the only cue to it's scope, so I think it does matter. I think`manage_shared_channel` would be better.

wiggin77:
TBH, I'm not fully up to speed on permissions scoping, but I can say that this permission is not applied on a per channel basis.  It allows someone to create/delete/manage shared channels on any team within their system.  @mkraft does that suggest this should be system scoped?

mkraft:
> TBH, I'm not fully up to speed on permissions scoping, but I can say that this permission is not applied on a per channel basis. It allows someone to create/delete/manage shared channels on any team within their system. @mkraft does that suggest this should be system scoped?

Yes, then the scope is wrong. It should be team-scoped.

wiggin77:
Thanks for the guidance @mkraft .  Can you explain why team scope versus system scope?  The permission is given to system admins by default, and can be assigned by them to other users, but I don't think will be assigned on a per team basis.

agnivade:
My apologies. I did read https://docs.mattermost.com/deployment/permissions-backend.html and still managed to confused myself. Will update the scope.

mkraft (approved):

srkgupta (approved):
Tested the PR along with its dependent PRs and permission is showing up and working fine on System Console. Approving the PR.

PR #15605, opened 2020-09-25
Add the API for search files

jespino:
#### Summary
Add the API for search files

#### Release Note
```release-note
Added API for files search (Behind a feature flag)
```


esethna:
@ethervoid @mgdelacroix can you guys please help with reviews on this when possible to unblock Jesus on this feature?

app/migrations.go
```diff
@@ -283,6 +284,32 @@ func (a *App) DoSystemConsoleRolesCreationMigration() {
...
+func (a *App) doContentExtractionConfigMigration() {
+	// If the migration is already marked as completed, don't do it again.
+	if _, err := a.Srv().Store.System().GetByName(ContentExtractionConfigMigrationKey); err == nil {
+		return
+	}
+
+	if usersCount, err := a.Srv().Store.User().Count(model.UserCountOptions{}); err != nil {
+		mlog.Critical("Failed to get the users count for migrating the content extraction, using default value", mlog.Err(err))
+	} else {
+		if usersCount < 500 {
```
ethervoid, line 296: Is this a hard limit we're setting right? Can we extract this number as a constant?

ethervoid (commented):
Some comments

api4/file.go
```diff
@@ -725,3 +727,71 @@ func writeFileResponse(filename string, contentType string, contentSize int64, l
...
+	results, err := c.App.SearchFilesInTeamForUser(terms, c.App.Session().UserId, c.Params.TeamId, isOrSearch, includeDeletedChannels, timeZoneOffset, page, perPage)
+
+	elapsedTime := float64(time.Since(startTime)) / float64(time.Second)
+	metrics := c.App.Metrics()
+	if metrics != nil {
+		metrics.IncrementFilesSearchCounter()
+		metrics.ObserveFilesSearchDuration(elapsedTime)
+	}
+
+	if err != nil {
```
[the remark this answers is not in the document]
jespino: Strictly based on the pre-existing code located in `api4/post.go` if this is wrong, we should do the changes later and consistently across all the searches.

app/file.go
```diff
@@ -771,6 +772,26 @@ func (a *App) UploadFileX(channelId, name string, input io.Reader,
 		}
 	}
 
+	if *a.Config().FileSettings.ExtractContent {
+		infoCopy := *t.fileinfo
+		a.Srv().Go(func() {
+			file, aerr := a.FileReader(t.fileinfo.Path)
+			if aerr != nil {
+				mlog.Error("Failed to open file for extract file content", mlog.Err(aerr))
```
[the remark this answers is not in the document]
jespino: good point.

app/migrations.go
jespino: Good point.

app/file.go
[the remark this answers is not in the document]
jespino: done

app/migrations.go
jespino: done

app/upload.go
```diff
@@ -295,6 +296,21 @@ func (a *App) UploadData(us *model.UploadSession, rd io.Reader) (*model.FileInfo
...
 	}
 
+	if *a.Config().FileSettings.ExtractContent {
+		infoCopy := *info
+		a.Srv().Go(func() {
+			text, err := docextractor.Extract(infoCopy.Name, file, docextractor.ExtractSettings{
+				ArchiveRecursion: *a.Config().FileSettings.ArchiveRecursion,
+			})
+			if err != nil {
+				mlog.Error("Failed to extract file content", mlog.Err(err))
```
ethervoid, line 306: I missed to mention this one to include the return

ethervoid (approved):
Another return more and LGTM! :D

app/upload.go
jespino: Done

api4/file.go
```diff
@@ -725,3 +727,76 @@ func writeFileResponse(filename string, contentType string, contentSize int64, l
...
+		isOrSearch = *params.IsOrSearch
+	}
+
+	page := 0
+	if params.Page != nil {
+		page = *params.Page
+	}
+
+	perPage := 60
+	if params.PerPage != nil {
```
mgdelacroix, line 775: I think we should check for a max items per page here, something like:

```suggestion
	if params.PerPage != nil && *params.PerPage < 60 {
```

and if we make this change, I'd suggest to move that `60` to a `const`

mgdelacroix (approved):
This is going to be great!! Adding just one nonblocking nit

api4/file.go
jespino: This is strictly based on what is done with posts, If you feel strong about this file a ticket to fix this in all the searches.

jespino:
@ogi-m this shouldn't affect anything without the feature flag, and we can test the whole feature later. I would check uploading files, if that works, I think is pretty safe to say that this is not breaking things..

ogi-m (approved):
Tested with various file types, file upload working as expected 👍🏻

PR #15610, opened 2020-09-25
WIP: Fix Wrong validation message when Bot username is absent

### LINKED ISSUE #13908: Wrong validation message when Bot username is absent
### Description

I tried to create via REST a Bot without setting the username and I get an "Invalid email" error

Steps:

POST http://localhost:8065/api/v4/bots
`{"display_name":"Alfred","description":"CMS Automation"}`

And response is: 

`{"id":"model.user.is_valid.email.app_error","message":"Invalid email","detailed_error":"","request_id":"4by66j8tcjfpmjn65sjjbrwn1y","status_code":400}`

Seem similar to [MM-19310](https://mattermost.atlassian.net/browse/MM-19310)

yzdann:
#### Summary
Add pre-flight `IsValidUsername` check to prevent wrong validation error on email instead of username for bot creation api.

#### Ticket Link
Fixes #13908

### DIFF
--- a/app/bot.go
+++ b/app/bot.go
@@ -18,6 +18,10 @@ import (
 
 // CreateBot creates the given bot and corresponding user.
 func (a *App) CreateBot(bot *model.Bot) (*model.Bot, *model.AppError) {
+	if !model.IsValidUsername(bot.Username) {
+		return nil, model.NewAppError("Bot.IsValid", "model.bot.is_valid.username.app_error", nil, "", http.StatusBadRequest)
+	}
+
 	user, err := a.Srv().Store.User().Save(model.UserFromBot(bot))
 	if err != nil {
 		return nil, err
--- a/app/bot_test.go
+++ b/app/bot_test.go
@@ -99,6 +99,19 @@ func TestCreateBot(t *testing.T) {
 		require.NotNil(t, err)
 		require.Equal(t, "store.sql_user.save.username_exists.app_error", err.Id)
 	})
+
+	t.Run("create bot, without username", func(t *testing.T) {
+		th := Setup(t).InitBasic()
+		defer th.TearDown()
+
+		_, err := th.App.CreateBot(&model.Bot{
+			Username:    "",
+			Description: "a bot",
+			OwnerId:     th.BasicUser.Id,
+		})
+		require.NotNil(t, err)
+		require.Equal(t, "model.bot.is_valid.username.app_error", err.Id)
+	})
 }
 
 func TestPatchBot(t *testing.T) {


app/bot.go
```diff
@@ -18,6 +18,10 @@ import (
 
 // CreateBot creates the given bot and corresponding user.
 func (a *App) CreateBot(bot *model.Bot) (*model.Bot, *model.AppError) {
+	if !model.IsValidUsername(bot.Username) {
+		return nil, model.NewAppError("Bot.IsValid", "model.bot.is_valid.username.app_error", nil, "", http.StatusBadRequest)
+	}
```
hanzei, line 23: Bots have a bot specific `IsValid()` method. I'm leaning toward using that one given that it also does other checks.

hanzei (changes requested):
Thanks for submitting this PR :+1:

app/bot.go
yzdann: As far as I know, bot creation is based on the user model so username validation will bypass for bot because the user model will create a random username in lack of username.
So i taught with that check we will fix that bug.

hanzei: Surely, the check you added does fix the bug. `bot.IsValid()` would fix that bug _and_ prevent similar ones.

hanzei:
@yzdann Is it feasible to add a test for this?

yzdann:
Sure. I will work on tests and using `IsValid`.

app/bot.go
yzdann: If we want to create a bot first we need to create user to pass `id` of that user to bot `UserId` so we need to create user:
```
	user, err := a.Srv().Store.User().Save(model.UserFromBot(bot))
```
so if we put `bot.IsValid()` before that line code will return error on `if !IsValidId(b.UserId) { ...`  validation .
so should we put validation in [app](https://github.com/mattermost/mattermost-server/blob/50e37068b5f50d365c39bb2508a59aed10a77912/app/bot.go#L20)or in model  before [this](https://github.com/mattermost/mattermost-server/blob/50e37068b5f50d365c39bb2508a59aed10a77912/model/bot.go#L169).

yzdann: @hanzei

hanzei: Can you check what kind of error is returned before returning it in `CreateBot`? And if the error is caused by a missing UserID, just ignore it.

yzdann: For clarification to myself: `IsValid` check `id, username, ...` sequentially so if we have a problem with `id` it always returns `id error`.
To fix this issue these solutions come to my mind:

Create custom `IsValid` for checking fields that will send on create a bot (`username, description, display_name`) and use this inside of the main bot `IsValid` for reducing duplicate code.

Or 
we can set `fake id` or `reserved id`. With `fake id` won't work because `User.IsValid` for checking `id` will return an error to surely prevent.
I don't know if mattermost have sth such as `reserved id` for this kind of purpose.

yzdann: @hanzei

hanzei: I was wondering if the code could check f `IsValid` returned an error because `id` is empty. If that is the case, the code could just ignore it. Does this seem suitable to you?

yzdann: Sorry I answered late.
If we add `IsValid` before 
```
	user, err := a.Srv().Store.User().Save(model.UserFromBot(bot))
```
No matter what POST request contains `IsValid` will alwasy stuck on `id` check and if we ignore it.
It won't fix this issue.

hanzei: I see that `IsValid` will return an eror if `id` is not set. But I'm wondering if it's suitable to ignore that specific error from `IsValid`.

PR #15616, opened 2020-09-28
[MM-27277] Move away from storing uploaded files into memory

streamer45:
#### Summary

PR removes full file memory allocations for uploads and instead relies on streaming them through the use of readers and writers.

**Benchmarks**

```
name                                                          old alloc/op   new alloc/op   delta
UploadFile/random-5Mb-gif-raw_UploadFileX_Content-Length-8      15.8MB ± 0%     1.1MB ± 0%  -92.84%  (p=0.029 n=4+4)
UploadFile/random-5Mb-gif-raw_UploadFileX_chunked-8             15.8MB ± 0%     1.1MB ± 0%  -92.84%  (p=0.029 n=4+4)
UploadFile/random-5Mb-gif-image_UploadFileX_Content-Length-8    68.3MB ± 0%    20.3MB ± 0%  -70.24%  (p=0.029 n=4+4)
UploadFile/random-5Mb-gif-image_UploadFileX_chunked-8           68.3MB ± 0%    20.3MB ± 0%  -70.24%  (p=0.029 n=4+4)
UploadFile/random-2Mb-jpg-raw_UploadFileX_Content-Length-8      3.17MB ± 0%    1.13MB ± 0%  -64.36%  (p=0.029 n=4+4)
UploadFile/random-2Mb-jpg-raw_UploadFileX_chunked-8             3.17MB ± 0%    1.13MB ± 0%  -64.36%  (p=0.029 n=4+4)
UploadFile/random-2Mb-jpg-image_UploadFileX_Content-Length-8    59.3MB ± 0%    59.4MB ± 0%   +0.10%  (p=0.029 n=4+4)
UploadFile/random-2Mb-jpg-image_UploadFileX_chunked-8           59.3MB ± 0%    59.4MB ± 0%   +0.11%  (p=0.029 n=4+4)
UploadFile/zero-10Mb-raw_UploadFileX_Content-Length-8           32.6MB ± 0%     1.1MB ± 0%  -96.53%  (p=0.029 n=4+4)
UploadFile/zero-10Mb-raw_UploadFileX_chunked-8                  32.6MB ± 0%     1.1MB ± 0%  -96.53%  (p=0.029 n=4+4)
UploadFile/zero-10Mb-image_UploadFileX_Content-Length-8         32.6MB ± 0%     1.1MB ± 0%  -96.53%  (p=0.029 n=4+4)
UploadFile/zero-10Mb-image_UploadFileX_chunked-8                32.6MB ± 0%     1.1MB ± 0%  -96.53%  (p=0.029 n=4+4)

name                                                          old allocs/op  new allocs/op  delta
UploadFile/random-5Mb-gif-raw_UploadFileX_Content-Length-8         144 ± 0%       174 ± 0%  +20.83%  (p=0.029 n=4+4)
UploadFile/random-5Mb-gif-raw_UploadFileX_chunked-8                144 ± 1%       174 ± 0%  +21.04%  (p=0.029 n=4+4)
UploadFile/random-5Mb-gif-image_UploadFileX_Content-Length-8     3.70M ± 0%     0.00M ± 0%  -99.97%  (p=0.029 n=4+4)
UploadFile/random-5Mb-gif-image_UploadFileX_chunked-8            3.70M ± 0%     0.00M ± 0%  -99.97%  (p=0.029 n=4+4)
UploadFile/random-2Mb-jpg-raw_UploadFileX_Content-Length-8         140 ± 0%       175 ± 0%  +25.00%  (p=0.029 n=4+4)
UploadFile/random-2Mb-jpg-raw_UploadFileX_chunked-8                140 ± 0%       174 ± 0%  +24.64%  (p=0.029 n=4+4)
UploadFile/random-2Mb-jpg-image_UploadFileX_Content-Length-8     3.71M ± 0%     3.71M ± 0%   +0.00%  (p=0.029 n=4+4)
UploadFile/random-2Mb-jpg-image_UploadFileX_chunked-8            3.71M ± 0%     3.71M ± 0%   +0.00%  (p=0.029 n=4+4)
UploadFile/zero-10Mb-raw_UploadFileX_Content-Length-8              147 ± 0%       173 ± 0%  +17.69%  (p=0.029 n=4+4)
UploadFile/zero-10Mb-raw_UploadFileX_chunked-8                     147 ± 1%       173 ± 0%  +17.89%  (p=0.029 n=4+4)
UploadFile/zero-10Mb-image_UploadFileX_Content-Length-8            147 ± 1%       173 ± 0%  +17.89%  (p=0.029 n=4+4)
UploadFile/zero-10Mb-image_UploadFileX_chunked-8                   146 ± 0%       173 ± 0%  +18.09%  (p=0.029 n=4+4)

```

Comparison with a 100MB file:

uploadmem

#### Ticket

https://mattermost.atlassian.net/browse/MM-27277

#### Related PRs

https://github.com/mattermost/mattermost-webapp/pull/6568


app/upload.go
```diff
@@ -19,6 +19,65 @@ import (
 
 const minFirstPartSize = 5 * 1024 * 1024 // 5MB
 
+func (a *App) runPluginsHook(info *model.FileInfo, file io.Reader) *model.AppError {
+	// call plugins upload hook
+	if pluginsEnvironment := a.GetPluginsEnvironment(); pluginsEnvironment != nil {
```
isacikgoz, line 24: Nit: we can return early if `pluginsEnvironment == nil` to save a one level indentation.

isacikgoz (approved):
Awesome improvement!! 🚀  Have a suggestion regarding to code readability.

app/upload.go
streamer45: Yeah, nice catch. This was moved from inside another function so I missed the possible improvement. Will update :+1:

levb:
A while ago, I was under impression that we had a _requirement_ (!) from a major customer that uploads are not stored in the file store until they are scanned by the plugins, that's why the buffering before plugins was in-memory, This requirement never made sense to me, and combined with the FileWillBeUploaded.

If the file is in the file store, I'd also advocate to change the RPC to pass around the file moniker, and not a stream, and let the plugins open it directly.

streamer45:
> A while ago, I was under impression that we had a _requirement_ (!) from a major customer that uploads are not stored in the file store until they are scanned by the plugins, that's why the buffering before plugins was in-memory, This requirement never made sense to me, and combined with the FileWillBeUploaded.
> 

Thanks @levb for the additional context . That explains the original design choice. I'll see if I can find more info about this unusual requirement.

> If the file is in the file store, I'd also advocate to change the RPC to pass around the file moniker, and not a stream, and let the plugins open it directly.

That would work but it might require a breaking change to the current plugin API unless we provide a new hook. Maybe we can open a ticket to revisit this when we start working on the next major version.

levb:
@streamer - Yep (breaking change). That's exactly why I left it in memory back then, gave up on fixing the plugin API part of it. I ❤️  and 😍  this PR!

Another huge improvement assuming mist uploads are images, would be to get rid of go image processing (thumbnail, etc), replace it with a more efficient service. Go may be a good language, but the image libraries are inadequate.

streamer45:
> @streamer - Yep (breaking change). That's exactly why I left it in memory back then, gave up on fixing the plugin API part of it. I  and  this PR!

:tada: 

> Another huge improvement assuming mist uploads are images, would be to get rid of go image processing (thumbnail, etc), replace it with a more efficient service. Go may be a good language, but the image libraries are inadequate.

Totally agree. Image support is not even that bad if you compare it with videos, then there's a real lack of options.
Moving to a dedicated service for a/v stuff would indeed make things simpler on this side and it's something I'd definitely consider working on in the future.

app/file.go
```diff
@@ -623,32 +623,46 @@ func (a *App) UploadFileX(channelId, name string, input io.Reader,
...
-	aerr = t.readAll()
+	written, aerr := t.writeFile(io.MultiReader(t.buf, t.limitedInput), t.fileinfo.Path)
 	if aerr != nil {
-		return t.fileinfo, aerr
+		return nil, aerr
 	}
 
-	aerr = t.runPlugins()
+	if written > t.maxFileSize {
+		a.RemoveFile(t.fileinfo.Path)
```
agnivade, line 632: Do we want to log the error returned from here?

app/file.go
```diff
@@ -623,32 +623,46 @@ func (a *App) UploadFileX(channelId, name string, input io.Reader,
 		}
 	}
 
-	aerr = t.readAll()
+	written, aerr := t.writeFile(io.MultiReader(t.buf, t.limitedInput), t.fileinfo.Path)
```
agnivade, line 626: Couple of questions here:

What is the objective here by writing before reading? Is it to just check the file upload size? If the file size is greater than the initial buffer size, how are we ensuring the file size limit then?

Sorry about so many questions, I'm a bit confused :)

 A comment would be helpful here.

app/file.go
```diff
@@ -623,32 +623,46 @@ func (a *App) UploadFileX(channelId, name string, input io.Reader,
...
+	if aerr != nil {
+		return nil, aerr
 	}
 
 	// Concurrently upload and update DB, and post-process the image.
 	wg := sync.WaitGroup{}
 
-	if !t.Raw && t.fileinfo.IsImage() {
+	if !t.Raw && t.fileinfo.IsImage() && t.fileinfo.HasPreviewImage {
+		file, aerr = a.FileReader(t.fileinfo.Path)
```
agnivade, line 654: Can you clarify why we are getting a fileReader again here after L639?

app/upload.go
```diff
@@ -19,6 +19,67 @@ import (
...
+		defer close(errChan)
+		pluginContext := a.PluginContext()
+		pluginsEnvironment.RunMultiPluginHook(func(hooks plugin.Hooks) bool {
+			newInfo, rejStr := hooks.FileWillBeUploaded(pluginContext, info, file, w)
+			if rejStr != "" {
+				errChan <- model.NewAppError("runPluginsHook", "app.upload.run_plugins_hook.rejected", nil, rejStr, http.StatusBadRequest)
+				return false
+			}
+			if newInfo != nil {
+				info = newInfo
```
agnivade, line 42: Isn't there a race here? We are writing to info from a goroutine, but reading `info.Path` at L48.

app/upload.go
```diff
@@ -19,6 +19,67 @@ import (
...
+		}, plugin.FileWillBeUploadedId)
+	}()
+
+	tmpPath := info.Path + ".tmp"
+	written, err := a.WriteFile(r, tmpPath)
+	if err != nil {
+		if fileErr := a.RemoveFile(tmpPath); fileErr != nil {
+			mlog.Error("Failed to remove file", mlog.Err(fileErr))
+		}
+		return err
```
agnivade, line 54: There's a bit of a problem here. If we return from here, then we will never be reading from `errChan` so `RunMultiPluginHook` will have a goroutine leak.

app/upload.go
```diff
@@ -19,6 +19,67 @@ import (
...
 
+func (a *App) runPluginsHook(info *model.FileInfo, file io.Reader) *model.AppError {
+	pluginsEnvironment := a.GetPluginsEnvironment()
+	if pluginsEnvironment == nil {
+		return nil
+	}
+
+	// using a pipe to avoid loading the whole file content in memory.
+	r, w := io.Pipe()
+	errChan := make(chan *model.AppError, 1)
```
agnivade, line 30: It's not immediately apparent to me why we need a buffered channel. The write happens in a separate goroutine, so even if we haven't started to read yet, we will at some point anyways.

Sorry, I do realize that this is not new code and you just moved it to a new file. Nevertheless, since you are making a great overhaul here, maybe it's good to look into these as well.

go.tools.mod
```diff
@@ -4,7 +4,7 @@ go 1.14
 
 require (
 	github.com/jstemmer/go-junit-report v0.9.1 // indirect
-	github.com/mattermost/mattermost-utilities/mmgotool v0.0.0-20200915114419-f4421bc07461 // indirect
+	github.com/mattermost/mattermost-utilities/mmgotool v0.0.0-20200926180007-fd1b679200e5 // indirect
```
agnivade, line 7: Unrelated, but I am recently seeing this. `make i18n-extract` will automatically bump the version in go.tools.mod, even when there is no `-u`. Do you know why this would happen? I need to see.

agnivade (commented):
Great stuff! Demo this week? ;)

app/file.go
streamer45: Yes, that can't hurt.

app/file.go
streamer45: I think the code is a bit spread due to the `UploadFileTask` structure keeping most of the state. This can make it harder to follow. I have plans to improve this a bit in https://mattermost.atlassian.net/browse/MM-29081.

As far as your questions:

we write to the filestore to avoid loading the file in memory. The size limit should be guaranteed by the limited reader initialized in `t.init()`:

https://github.com/mattermost/mattermost-server/blob/155bd35e56a978a77f28f197878b7fdbcf1545da/app/file.go#L578-L581

app/file.go
streamer45: Because `a.runPluginsHook()` can potentially read and change the file.

app/upload.go
streamer45: It looks like it. We should make a copy of the path before launching the goroutine. Great catch :+1:

app/upload.go
streamer45: My understanding is that we don't need to be reading from it for the goroutine to exit since it's using a buffered channel.

app/upload.go
streamer45: > It's not immediately apparent to me why we need a buffered channel. The write happens in a separate goroutine, so even if we haven't started to read yet, we will at some point anyways.

I believe the main reason was to make the send non-blocking.

> Sorry, I do realize that this is not new code and you just moved it to a new file. Nevertheless, since you are making a great overhaul here, maybe it's good to look into these as well.

This code is fairly new actually. And it's all my doing so better to keep it sane :stuck_out_tongue:

go.tools.mod
streamer45: Not sure. I just ran `make i18n-extract` to update some missing translations and the version got bumped. I can revert this.

app/upload.go
agnivade: It's a buffered channel of size 1. But the `RunMultiPluginHook` will be called for each plugin. So it will block whenever we have more than 1 plugin. The buffer needs to be equal to the number of plugins.

app/upload.go
agnivade: > I believe the main reason was to make the send non-blocking.

Gotcha, see my comment below.

app/file.go
agnivade: Ah okay, so instead of storing the file in memory, we are reading, writing, reading and writing again right?

I am wondering how much of an effect this has for S3 file stores. Would it be possible to do a comparison of time taken for small, medium and large files against localstore and S3?

Asking because I have a branch which does a streaming file read while serving emojis to leverage the `sendfile` system call. But I saw performance drops when using it for S3.

app/upload.go
streamer45: That's definitely something I failed to consider. A case where more than one plugin has the `FileWillBeUploaded` hook registered.

app/file.go
streamer45: > Ah okay, so instead of storing the file in memory, we are reading, writing, reading and writing again right?

Unless the plugin hook modifies the file we should be doing one write only to the backing store.
If the file is an image it's a bit different since we need to read it once more to allow for post-processing (preview and thumbnail generation). We might be able to optimize a bit such case a bit but I'd have to look deeper at how the image functions operate as far as memory allocations.

> I am wondering how much of an effect this has for S3 file stores. Would it be possible to do a comparison of time taken for small, medium and large files against localstore and S3?
> 
> Asking because I have a branch which does a streaming file read while serving emojis to leverage the `sendfile` system call. But I saw performance drops when using it for S3.

I can take a look but I imagine there will be some latency trade-off. After all there's a bit of back and forth especially with images and the plugin hook.

agnivade: > I can take a look but I imagine there will be some latency trade-off. After all there's a bit of back and forth especially with images and the plugin hook.

Yes, just want to see how much of a tradeoff is that. We definitely want to have this for sure. The test is more to get some data and confirm that the latency is what we expect it to be.

streamer45: So I might have gone a bit further than anticipated with this but I really wanted to have a clear picture.

It appears the `WriteFile()` call in `runPluginsHook()` is actually adding a significant amount of latency when S3 is used, even if no plugin is present so I'll see if I can find a way to trigger it from within the `RunMultiPluginHook()` function parameter. That should at least improve the default case. The image upload case has naturally worsened since we have to read data again from the filestore but that might be harder to improve.

comparison

agnivade: This looks great. Yeah improving the default case would be very helpful here.

hanzei:
> A while ago, I was under impression that we had a requirement (!) from a major customer that uploads are not stored in the file store until they are scanned by the plugins, that's why the buffering before plugins was in-memory, This requirement never made sense to me, and combined with the FileWillBeUploaded.

I remember this being a requirement as well, back in the day from Corey.

streamer45:
> > A while ago, I was under impression that we had a requirement (!) from a major customer that uploads are not stored in the file store until they are scanned by the plugins, that's why the buffering before plugins was in-memory, This requirement never made sense to me, and combined with the FileWillBeUploaded.
> 
> I remember this being a requirement as well, back in the day from Corey.

Not allowing the data to go from memory to the backing store seems a bit strict as a requirement.
We are currently verifying if it's enough for the file not be made public. A common use case would be an antivirus plugin rejecting malicious files. Those files would get removed as soon as they are found to be problematic and never allowed to be accessed by a user.

agnivade:
I'm curious to understand the reasoning behind this requirement. Having this requirement puts a very strict limit on the minimum memory required to serve Mattermost, and sometimes even double that depending on plugins.

app/file.go
streamer45: Much better now in case the plugin hook does not run:

comparison_new

hanzei:
> I'm curious to understand the reasoning behind this requirement. Having this requirement puts a very strict limit on the minimum memory required to serve Mattermost, and sometimes even double that depending on plugins.


I think it's because of legal and compliant reasons. One could argue that a temporary copy of an illegal file is still illegal.

agnivade:
But the file is already in memory. What difference does it make from a legal perspective once a file is inside a system - be it stored in memory or a file system? A program can always dump the entire memory and read the file if it wants to.

Anyways, this PR is not meant for this discussion :P And we are anyways doing this. So it's okay.

levb:
BTW, @agnivade does this also fix https://mattermost.atlassian.net/browse/MM-12001? At least in the sense that we get the content-length without the need to buffer in-memory?

agnivade:
I think @streamer45 fixed this recently by setting the min part size for PUT requests. So I'll defer to him.

streamer45:
> BTW, @agnivade does this also fix https://mattermost.atlassian.net/browse/MM-12001? At least in the sense that we get the content-length without the need to buffer in-memory?

As @agnivade said I recently pushed a fix related to this (https://github.com/mattermost/mattermost-server/pull/15737). I was unaware of the internal minio behaviour and found out that passing an unknown size meant allocating a 500MB+ buffer. 

We temporarily fixed this by setting the part size to be the minimum of 5MB. Of course there are still improvements to be made: for example, content-length can be passed when known, also images could be passed as a buffer since we need to load them in memory anyway. Unfortunately uploads coming from the web client come as multipart and that doesn't make it possible to figure out the content-length. For very large files I think it makes sense to use the new resumable uploads API which requires the file size to be known in advance.

agnivade (approved):
Good stuff

streamer45:
@lindalumitchell 

This is an improvement in memory usage when uploading files. I've provided benchmarks and ran a couple of loadtests.
I think it makes sense to merge this so we can further verify it on daily. As far as QA is concerned, a general smoke-test around file uploads (both with local and S3 backing filestores) should be sufficient.

lindalumitchell (approved):
File upload smoke tests look good.

PR #15625, opened 2020-09-28
Thread autoresponse posts with the post they are autoresponding to

larkox:
#### Summary
Thread autoresponse posts with the post they are autoresponding to.
Right now autoresponses have the style of a thread, but they do not work as threads (they do not appear on the RHS, and they are not removed when deleting the root post).

There is still an edge case that can be considered "normal conversation flow" and be left as it is:
- Send a message
- Reply to that message, expecting an autoresponse
  - The autoresponse message will be added to the thread
- Remove the second message we sent (the one that generated the autoresponse)
  - The autoresponse message will stay in place even if the message that generated it is deleted, and may lead to think the autoresponse was made to the root post.

#### Ticket Link
Fix https://mattermost.atlassian.net/browse/MM-29057
Fix https://mattermost.atlassian.net/browse/MM-29059


app/auto_responder_test.go
```diff
@@ -185,14 +213,65 @@ func TestSendAutoResponseSuccess(t *testing.T) {
...
+		autoResponderPostFound := false
+		for _, post := range list.Posts {
+			if post.Type == model.POST_AUTO_RESPONDER {
+				autoResponderPostFound = true
+				assert.Equal(t, savedPost.Id, post.RootId)
+				assert.Equal(t, savedPost.Id, post.ParentId)
+			}
+		}
+		assert.True(t, autoResponderPostFound)
+	}
```
hanzei, line 240: Is the `if` needed? Can it be dropped? Can we also change this in the other test?

app/auto_responder.go
```diff
@@ -43,11 +43,21 @@ func (a *App) SendAutoResponse(channel *model.Channel, receiver *model.User) (bo
 		return false, nil
 	}
 
+	rootID := post.Id
```
hanzei, line 46: Is this really needed? Can we rely on `a.CreatePost` doing the "right thing" can just use `post.RootId` in any case?

app/auto_responder.go
```diff
@@ -43,11 +43,21 @@ func (a *App) SendAutoResponse(channel *model.Channel, receiver *model.User) (bo
...
+	if post.RootId != "" {
+		rootID = post.RootId
+	}
+
+	parentID := post.Id
+	if post.ParentId != "" {
+		parentID = post.ParentId
+	}
```
hanzei, line 54: According to https://community-daily.mattermost.com/core/pl/np785upq8pgujcag478cdyjs6o `ParentId` is deprecated. Should it still get used here?

hanzei (changes requested):
Glad this finally gets fixed :+1:

app/auto_responder.go
larkox: Not sure what do you mean. I guess you are saying that we should only define rootID and not define parentID since CreatePost will make them the same either way?

app/auto_responder.go
larkox: I know it is deprecated, but it is still there. For the sake of consistency, I preferred to check both. But if you prefer, I can remove this piece of code and use only RootID

app/auto_responder.go
hanzei: Can we just pass `post.RootId` as `rootID` and remove the `if` check?

app/auto_responder.go
hanzei: I'm 3/5 that new code shouldn't use a deprecated field any longer.

app/auto_responder.go
larkox: No, since RootId will be empty for non threaded messages.

hanzei (approved):
LGTM :+1:

mgdelacroix (approved):
LGTM! :+1:

lindy65:
@larkox - not sure if it's the failing `ee/circleci` that is preventing the spinwick from building?

larkox:
@lindy65 It seems that the test server has been created already :tada:

lindy65 (approved):
Thanks @larkox - tested according to steps in both Jira tickets as well as the case you listed above in the summary.  All scenarios are fixed 👍

PR #15648, opened 2020-09-29
MM-28815: Message export query optimization

mkraft:
#### Summary

About a 5x performance improvement on the `(*SqlComplianceStore).ComplianceExport` query. Does remove the `Posts.CreateAt > :StartTime` part of the `WHERE` condition since `UpdateAt` is set when posts are created.

#### Ticket Link

https://mattermost.atlassian.net/browse/MM-28815


store/sqlstore/compliance_store.go
```diff
@@ -220,7 +229,6 @@ func (s SqlComplianceStore) MessageExport(after int64, limit int) ([]*model.Mess
 			Posts.Props AS PostProps,
 			Posts.OriginalId AS PostOriginalId,
 			Posts.RootId AS PostRootId,
-			Posts.Props AS PostProps,
```
mkraft, line 223: This was just duplicated in the query.

PR #15657, opened 2020-09-29
[MM-29174] fix: enable automatic emoji resize feature

Leryan:
#### Summary

Enable emoji image resize on upload, before the file gets written on disk.

In fact, the image was indeed resized, but written twice: once in the resizing code, and a second time at the very end of the function, taking the original file as parameter.

#### Ticket

https://mattermost.atlassian.net/browse/MM-29174


Leryan:
ps: it’s late and didn’t search for existing tests to update. if that is required please point me to the nearest file/folder/repo where to implement such tests :)

streamer45:
@Leryan 

Thanks for the PR :+1: I've created a ticket for this.  
We will review it soon.

streamer45 (commented):
Nice catch @Leryan! Code changes look good :+1: 

> ps: it’s late and didn’t search for existing tests to update. since i saw it’s required please point me to the nearest file/folder/repo where to implement such tests :) also, i didn’t find any `emoji_test.go` file so…

I think we can add a case in `TestCreateEmoji()` under `api4/emoji_test.go`.
There are already a couple of cases there that will create an emoji that needs resizing but they are failing to verify that the output image is actually resized. Something like this should work:

```go
// check that emoji gets resized
path, _ := fileutils.FindDir("data")
file, fileErr := os.Open(filepath.Join(path, "/emoji/"+newEmoji.Id+"/image"))
require.NoError(t, fileErr)
config, imageType, err := image.DecodeConfig(file)
require.NoError(t, err)
require.Equal(t, "gif", imageType)
require.LessOrEqual(t, config.Width, app.MaxEmojiWidth)
require.LessOrEqual(t, config.Height, app.MaxEmojiHeight)
```

Leryan:
@streamer45 i implemented the tests and created a helper as closure inside `TestCreate` function. i don’t think this matches the style of mattermost helpers, let me know if i should modify it and i’ll do :)

streamer45:
> @streamer45 i implemented the tests and created a helper as closure inside `TestCreate` function. i don’t think this matches the style of mattermost helpers, let me know if i should modify it and i’ll do :)

It looks perfectly fine to me. I'll review shortly :)

api4/emoji_test.go
```diff
@@ -5,8 +5,11 @@ package api4
...
 import (
 	"bytes"
+	"github.com/mattermost/mattermost-server/v5/utils/fileutils"
 	"image"
 	_ "image/gif"
+	"os"
+	"path/filepath"
 	"testing"
 
 	"github.com/stretchr/testify/assert"
```
streamer45, line 15:
```suggestion
	"bytes"
	"image"
	_ "image/gif"
	"os"
	"path/filepath"
	"testing"
	
	"github.com/mattermost/mattermost-server/v5/utils/fileutils"

	"github.com/stretchr/testify/assert"
```

We tend to separate imports in three different groups. Standard library, mattermost packages and external dependencies.

api4/emoji_test.go
```diff
@@ -33,6 +36,20 @@ func TestCreateEmoji(t *testing.T) {
 		th.RestoreDefaultRolePermissions(defaultRolePermissions)
 	}()
 
+	emojiWidth := app.MaxEmojiWidth * 2
+	emojiHeight := app.MaxEmojiHeight * 2
+	// check that emoji gets resized and is of expected type
+	checkEmojiFile := func(id, expectedImageType string) {
+		path, _ := fileutils.FindDir("data")
+		file, fileErr := os.Open(filepath.Join(path, "/emoji/"+id+"/image"))
+		require.NoError(t, fileErr)
```
streamer45, line 45: We want to close the file when we are done with it. I'd suggest adding a `defer file.Close()` right after the error check.

streamer45 (changes requested):

api4/emoji_test.go
Leryan: do you use gofmt/goimports? because that kind of thing is usually handled by such tools

Leryan: done

api4/emoji_test.go
Leryan: done

api4/emoji_test.go
streamer45: > do you use gofmt/goimports? because that kind of thing is usually handled by such tools

We do use `gofmt` as part of [golangci-lint](https://github.com/golangci/golangci-lint). We have a `make check-style` command to verify that files are formatted correctly. That said, we do not enforce the grouping of imports.

streamer45 (approved):
Looks great @Leryan, thank you again for your contribution :+1:

isacikgoz (approved):
Awesome @Leryan!! 🎉

streamer45:
@lindalumitchell 

This change is covered by unit tests which should be enough. A general smoke-test around custom emojis can probably be done post merge.

lindalumitchell (approved):

PR #15659, opened 2020-09-30
MM-29033: Modify Test S3 connection to handle path prefix

agnivade:
When a IAM user is just allowed access to a specific path prefix,
the bucket exists test will fail because that is applicable to the entire bucket.

We modify that check to just list objects under that path prefix and return
on the first non-nil error.

Another consideration was to do a StatObject, but it is only applicable to actual
objects and not at a directory level. Path prefixes are entirely logical and depends
on the underlying implementation. Therefore, ListObjects is a better choice.

https://mattermost.atlassian.net/browse/MM-29033

### DIFF
--- a/i18n/en.json
+++ b/i18n/en.json
@@ -1409,7 +1409,7 @@
     "translation": "Don't have permissions to write to local path specified or other error."
   },
   {
-    "id": "api.file.test_connection.s3.bucked_create.app_error",
+    "id": "api.file.test_connection.s3.bucket_create.app_error",
     "translation": "Unable to create bucket."
   },
   {
@@ -1419,6 +1419,10 @@
   {
     "id": "api.file.test_connection.s3.connection.app_error",
     "translation": "Bad connection to S3 or minio."
+  },
+  {
+    "id": "api.file.test_connection.s3.list_objects.app_error",
+    "translation": "Error trying to list objects."
   },
   {
     "id": "api.file.upload_file.incorrect_channelId.app_error",
@@ -5583,18 +5587,6 @@
     "translation": " attached a file."
   },
   {
-    "id": "ent.cloud.authentication_failed",
-    "translation": "Unable to authenticate to CWS"
-  },
-  {
-    "id": "ent.cloud.json_encode.error",
-    "translation": "Internal error marshaling request to CWS"
-  },
-  {
-    "id": "ent.cloud.request_error",
-    "translation": "Error processing request to CWS"
-  },
-  {
     "id": "ent.cluster.404.app_error",
     "translation": "Cluster API endpoint not found."
   },
@@ -5801,6 +5793,10 @@
   {
     "id": "ent.elasticsearch.index_channel.error",
     "translation": "Failed to index the channel"
+  },
+  {
+    "id": "ent.elasticsearch.index_channels_batch.error",
+    "translation": ""
   },
   {
     "id": "ent.elasticsearch.index_post.error",
--- a/i18n/en.json
+++ b/i18n/en.json
@@ -5587,6 +5587,18 @@
     "translation": " attached a file."
   },
   {
+    "id": "ent.cloud.authentication_failed",
+    "translation": "Unable to authenticate to CWS"
+  },
+  {
+    "id": "ent.cloud.json_encode.error",
+    "translation": "Internal error marshaling request to CWS"
+  },
+  {
+    "id": "ent.cloud.request_error",
+    "translation": "Error processing request to CWS"
+  },
+  {
     "id": "ent.cluster.404.app_error",
     "translation": "Cluster API endpoint not found."
   },
@@ -5793,10 +5805,6 @@
   {
     "id": "ent.elasticsearch.index_channel.error",
     "translation": "Failed to index the channel"
-  },
-  {
-    "id": "ent.elasticsearch.index_channels_batch.error",
-    "translation": ""
   },
   {
     "id": "ent.elasticsearch.index_post.error",
--- a/services/filesstore/s3store.go
+++ b/services/filesstore/s3store.go
@@ -34,6 +34,11 @@ type S3FileBackend struct {
 	trace      bool
 }
 
+const (
+	// This is not exported by minio. See: https://github.com/minio/minio-go/issues/1339
+	bucketNotFound = "NoSuchBucket"
+)
+
 // Similar to s3.New() but allows initialization of signature v2 or signature v4 client.
 // If signV2 input is false, function always returns signature v4.
 //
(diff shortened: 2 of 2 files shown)


agnivade:
@gabrieljackson - I have tested this with this policy:

```json
{
    "Version": "2012-10-17",
    "Statement": [
        {
            "Sid": "Allowexactmatch",
            "Action": [
                "s3:ListBucket"
            ],
            "Effect": "Allow",
            "Resource": [
                "arn:aws:s3:::agnivaperm"
            ],
            "Condition": {
                "StringEquals": {
                    "s3:prefix": [
                        "mytest"
                    ],
                    "s3:delimiter": [
                        "/"
                    ]
                }
            }
        },
        {
            "Sid": "allowsubmatch",
            "Action": [
                "s3:ListBucket"
            ],
            "Effect": "Allow",
            "Resource": [
                "arn:aws:s3:::agnivaperm"
            ],
            "Condition": {
                "StringLike": {
                    "s3:prefix": [
                        "mytest/*"
                    ]
                }
            }
        },
        {
            "Sid": "AllowAllS3ActionsInUserFolder",
            "Effect": "Allow",
            "Action": [
                "s3:*"
            ],
            "Resource": [
                "arn:aws:s3:::agnivaperm/mytest/*"
            ]
        }
    ]
}
```

Essentially, the `ListBucket` needs to have both Resource and Condition. And then everything inside the pathPrefix can have `s3:*` with just that resource.

This is necessary for the ListObjects for that pathPrefix to work. Please check and ensure that this is how it's implemented.

Or let me know if I need to test with a different policy.

streamer45:
What would happen if there are no objects under the configured prefix?

agnivade:
Nothing. It will just pass normally.

agnivade:
Yep, all good. The first object is the directory itself. So  `exists = true` will still be set. I did test it, but I had that in my mind, and forgot again.

`{ mytest/ 0001-01-01 00:00:00 +0000 UTC 0  0001-01-01 00:00:00 +0000 UTC map[] map[] map[] 0 { } []  false false   0001-01-01 00:00:00 +0000 UTC  <nil>}`

services/filesstore/s3store.go
```diff
@@ -73,20 +78,45 @@ func (b *S3FileBackend) TestConnection() *model.AppError {
...
+	// If a path prefix is present, we attempt to test the bucket by listing objects under the path
+	// and breaking after the first non-nil error. This is because the BucketExists call is only at a bucket level
+	// and sometimes the user might only be allowed access to the specified path prefix.
+	if b.pathPrefix != "" {
+		// In the case where the bucket exists, but the prefix doesn't, and the user has permissions for the entire bucket,
+		// this call cannot differentiate between bucket doesn't exist and prefix doesn't exist.
+		// In that case, it will just redundantly try to create the bucket. It's possible to avoid this with yet another call,
+		// but it's the cost of another network call either ways.
+		objInfoChan := s3Clnt.ListObjects(context.Background(), b.bucket, s3.ListObjectsOptions{Prefix: b.pathPrefix})
+		for obj := range objInfoChan {
```
streamer45, line 91: Do we actually need to range? Since we care only about the first object returned we could do a simple receive.

i18n/en.json
```diff
@@ -5802,6 +5794,10 @@
     "id": "ent.elasticsearch.index_channel.error",
     "translation": "Failed to index the channel"
   },
+  {
+    "id": "ent.elasticsearch.index_channels_batch.error",
+    "translation": ""
+  },
```
streamer45, line 5800: Was this actually missing or is it an out of sync issue?

agnivade: Out of sync issue. Fixed now.

services/filesstore/s3store.go
agnivade: Yeah much better. I also improved the logic a bit and fixed the case of prefix doesn't exist. It was a misunderstanding of mine. The range loop was actually exiting without even going inside, so `exists` would remain false. Inverted the boolean value to improve it.

services/filesstore/s3store.go
```diff
@@ -73,20 +78,36 @@ func (b *S3FileBackend) TestConnection() *model.AppError {
...
+	// and just checking the first response. This is because the BucketExists call is only at a bucket level
+	// and sometimes the user might only be allowed access to the specified path prefix.
+	if b.pathPrefix != "" {
+		obj := <-s3Clnt.ListObjects(context.Background(), b.bucket, s3.ListObjectsOptions{Prefix: b.pathPrefix})
+		if obj.Err != nil {
+			typedErr := s3.ToErrorResponse(obj.Err)
+			if typedErr.Code != bucketNotFound {
+				return model.NewAppError("TestFileConnection", "api.file.test_connection.s3.list_objects.app_error", nil, obj.Err.Error(), http.StatusInternalServerError)
+			}
+			exists = false
```
streamer45, line 92: What happens if the bucket exists but the path prefix doesn't? Do we still get a `bucketNotFound` or else?
Should we attempt to recover from this by creating the path prefix (en empty directory I suppose)?

agnivade: That case was always handled. When we would loop, the loop would exit without running once. Now that we are seeking, so the default value of `obj` will come out of the channel. No err will be set. And everything will be hunky dory.

streamer45: So just to confirm I am getting this right: `ListObjects()` is not returning an error if we call it on an non-existent path under an existing bucket.

agnivade: Correct.

streamer45 (approved):
No further questions. Looks good :+1:

gabrieljackson (approved):
Looks good!

agnivade:
@prapti - Would you be able to QA review this one? Or should I just assign to @lindalumitchell ?

prapti:
@agnivade If the test steps are still the same as noted in https://mattermost.atlassian.net/browse/MM-29033, I ran definitely test it, otherwise, is there anything else that needs to be checked?

agnivade:
If that's how Cloud servers are provisioned right now. Then yes, that's all that's needed. Thank you!

prapti:
@agnivade Spun up a new test server at https://pr-15659.test.mattermost.cloud/. I still see the following error although it's not exactly the same as the one reported in the ticket but seems related:

`{"level":"error","ts":1601573718.3111703,"caller":"app/server.go:438","msg":"Problem with file storage settings","error":"TestFileConnection: Error trying to list objects., Access Denied"}` 

Cloud instance created by pointing the version flat to the last commit SHA of this PR: `/cloud create pr-15659 --version 175263c --test-data`

agnivade:
Yes, so this is what I mentioned in my comment [here](https://github.com/mattermost/mattermost-server/pull/15659#issuecomment-701214556). The Cloud team needs to check that the above is indeed the policy applied. Without that, the change doesn't work.

prapti (approved):
Thank you for clarifying that @agnivade!
In that case, it looks good to me!
Approving.

agnivade:
Cool, I am going to push this.

/cc @jwilander @gabrieljackson on the note above https://github.com/mattermost/mattermost-server/pull/15659#issuecomment-702300511.

PR #15665, opened 2020-09-30
Document extractor service

jespino:
#### Summary
Add document extractor service as part of the work around document indexing

#### Ticket Link
[MM-29156](https://mattermost.atlassian.net/browse/MM-29156)


jespino:
From the QA perspective I think is not needed because is not exposed anywhere or used anywhere yet, It should be tested as part of the final feature that is going to be exposed in the user interface.

services/docextractor/archive.go
```diff
@@ -0,0 +1,61 @@
...
+		return "", fmt.Errorf("error copying data into temporary file: %v", err)
+	}
+	_, err = io.Copy(f, r)
+	f.Close()
+	if err != nil {
+		return "", fmt.Errorf("error copying data into temporary file: %v", err)
+	}
+
+	text := ""
+	err = archiver.Walk(f.Name(), func(file archiver.File) error {
```
reflog, line 43: string.Builder instead of text += "" ?

services/docextractor/archive.go
```diff
@@ -0,0 +1,61 @@
...
+	"github.com/mholt/archiver/v3"
+)
+
+type archiveExtractor struct {
+	SubExtractor Extractor
+}
+
+func (ae *archiveExtractor) Match(filename string) bool {
+	_, err := archiver.ByExtension(filename)
+	if err == nil {
```
reflog, line 19: `return err == nil`

reflog (changes requested):
left some minor reviews, mostly requesting comments

services/docextractor/plain.go
```diff
@@ -0,0 +1,44 @@
...
+)
+
+type plainExtractor struct{}
+
+func (pe *plainExtractor) Match(filename string) bool {
+	return true
+}
+
+func (pe *plainExtractor) Extract(filename string, r io.Reader) (string, error) {
+	validRanges := []*unicode.RangeTable{
```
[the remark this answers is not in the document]
jespino: Sure I can, I just trying to detect that the text is readable text (printable characters plus spaces, carrier returns and new lines).

services/docextractor/mmpreview.go
```diff
@@ -0,0 +1,80 @@
...
+	"io"
+	"mime/multipart"
+	"net/http"
+	"path"
+	"strings"
+
+	"code.sajari.com/docconv"
+	"github.com/pkg/errors"
+)
+
```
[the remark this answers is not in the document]
jespino: Yes, good thing, actually I can remove entirely the MMPreview support here because MMPreview is not freeded yet. But I think I going to add the comment.

services/docextractor/archive.go
jespino: Good point.

services/docextractor/archive.go
jespino: good point :+1:

services/docextractor/pdf.go
```diff
@@ -0,0 +1,68 @@
...
+	for pageIndex := 1; pageIndex <= totalPage; pageIndex++ {
+		p := reader.Page(pageIndex)
+		if p.V.IsNull() {
+			continue
+		}
+
+		rows, _ := p.GetTextByRow()
+		for _, row := range rows {
+			for _, word := range row.Content {
+				text += " " + fmt.Sprintf(word.S)
```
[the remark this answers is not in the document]
jespino: yes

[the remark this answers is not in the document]
jespino: Done

services/docextractor/plain.go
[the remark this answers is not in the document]
jespino: I think now is more explicit, but anyway I added a comment on it.

services/docextractor/mmpreview.go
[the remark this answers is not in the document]
jespino: Done

services/docextractor/archive.go
jespino: Done

services/docextractor/archive.go
jespino: Done

reflog (approved):
LGTM!

services/docextractor/documents.go
```diff
@@ -0,0 +1,69 @@
...
+	"path"
+	"strings"
+
+	"code.sajari.com/docconv"
+)
+
+type documentExtractor struct{}
+
+func (de *documentExtractor) Match(filename string) bool {
+	supportedExtensions := map[string]bool{
```
mgdelacroix, line 21: nit: this could be a `const` rather than defining it for every `Match` call

services/docextractor/documents.go
```diff
@@ -0,0 +1,69 @@
...
+		"html":  true,
+		"pages": true,
+		"rtf":   true,
+	}
+	extension := strings.TrimPrefix(path.Ext(filename), ".")
+	return supportedExtensions[extension]
+}
+
+func (de *documentExtractor) Extract(filename string, r io.Reader) (string, error) {
+	convertersByExtensions := map[string]func(io.Reader) (string, map[string]string, error){
```
mgdelacroix, line 35: Same, and actually could this be combined on the same map? So you can check both existence and the related converter?

services/docextractor/mmpreview.go
```diff
@@ -0,0 +1,80 @@
...
+	secret       string
+	pdfExtractor pdfExtractor
+}
+
+func newMMPreviewExtractor(url string, secret string, pdfExtractor pdfExtractor) *mmPreviewExtractor {
+	return &mmPreviewExtractor{url: url, secret: secret, pdfExtractor: pdfExtractor}
+}
+
+func (mpe *mmPreviewExtractor) Match(filename string) bool {
+	supportedExtensions := map[string]bool{
```
mgdelacroix, line 32: Same `const` here

mgdelacroix (approved):
LGTM! :+1: left a couple of nonblocking nits

jespino:
@mgdelacroix applied your proposed changes.

mgdelacroix (approved):
Thanks for applying the changes! LGTM :+1:

PR #15689, opened 2020-10-01
[MM-28420] Fix possible panic during plugin installation

streamer45:
#### Summary

PR fixes a possible panic during plugin installation. The main issue is that `Environment.Activate()` can return both a `nil` manifest and no error in case of an already active plugin. During the installation process this case was not handled and would cause a `nil` dereference panic.

I've added a test to reproduce this: a plugin is installed and activated, then its installation folder is removed. A second attempt to install the same plugin would fail to remove it and de-activate it since the directory would not be found anymore (apparently we don't check for active plugins during installation, but only if the bundle exists). The installation would go ahead but the plugin's activation would fail, returning the `nil` manifest.

#### Ticket

https://mattermost.atlassian.net/browse/MM-28420


lieut-data (approved):

mgdelacroix (approved):
LGTM! :+1:

lindalumitchell (approved):

PR #15737, opened 2020-10-02
[MM-29229] Set part size for S3 uploads to avoid excessive allocation

streamer45:
#### Summary

PR addresses an issue introduced in https://github.com/mattermost/mattermost-server/pull/15252 which is causing an excessive allocation for file uploads when using an S3 compatible filestore.

We now limit the allocation by setting the part size for the upload to the minimum allowed value of 5MBs.

##### Comparison (1MB file):

```
name           old time/op    new time/op    delta
S3WriteFile-8    34.9ms ± 2%    13.9ms ±38%  -60.24%  (p=0.001 n=7+7)

name           old alloc/op   new alloc/op   delta
S3WriteFile-8     671MB ± 0%       5MB ± 0%  -99.20%  (p=0.000 n=7+8)

name           old allocs/op  new allocs/op  delta
S3WriteFile-8     1.03k ± 1%     1.00k ± 0%   -2.64%  (p=0.000 n=8+8)
```

#### Ticket

https://mattermost.atlassian.net/browse/MM-29229

### DIFF
--- a/services/filesstore/filesstore_test.go
+++ b/services/filesstore/filesstore_test.go
@@ -364,3 +364,40 @@ func (s *FileBackendTestSuite) TestAppendFile() {
 		s.EqualValues(append(append(b, b2...), b3...), read)
 	})
 }
+
+func BenchmarkS3WriteFile(b *testing.B) {
+	utils.TranslationsPreInit()
+
+	settings := &model.FileSettings{
+		DriverName:              model.NewString(model.IMAGE_DRIVER_S3),
+		AmazonS3AccessKeyId:     model.NewString(model.MINIO_ACCESS_KEY),
+		AmazonS3SecretAccessKey: model.NewString(model.MINIO_SECRET_KEY),
+		AmazonS3Bucket:          model.NewString(model.MINIO_BUCKET),
+		AmazonS3Region:          model.NewString(""),
+		AmazonS3Endpoint:        model.NewString("localhost:9000"),
+		AmazonS3PathPrefix:      model.NewString(""),
+		AmazonS3SSL:             model.NewBool(false),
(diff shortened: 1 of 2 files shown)


grundleborg (approved):

services/filesstore/filesstore_test.go
```diff
@@ -364,3 +364,40 @@ func (s *FileBackendTestSuite) TestAppendFile() {
...
+	b.ResetTimer()
+
+	for i := 0; i < b.N; i++ {
+		written, err := backend.WriteFile(bytes.NewReader(data), path)
+		defer backend.RemoveFile(path)
+		require.Nil(b, err)
+		require.Equal(b, len(data), int(written))
+	}
+
+	b.StopTimer()
```
agnivade, line 402: We don't need this.

agnivade (approved):

services/filesstore/filesstore_test.go
streamer45: I added it to avoid counting the deferred `RemoveFile()` call. Not that it matters much in such case, I was interested in memory more than anything.

streamer45:
@lindalumitchell 

This PR fixes a rather severe performance regression with file uploads. I've provided a benchmark which covers the affected case.

services/filesstore/filesstore_test.go
agnivade: Ah I see. Gotcha :+1:

lindalumitchell (approved):

amyblais:
@streamer45 You may need to cherry-pick manually due to build issues on automated cherry-pick PRs in the server repo. I think devops team hasn't had a chance to investigate yet.

PR #15749, opened 2020-10-02
Add the content field to FileInfo

jespino:
#### Summary
Add the content field to FileInfo

This is going to store the Content data of the file in the future, for now is not used at all, is going to be used to search indexed documents by contents (that is in combination with other PRs around document text extraction, and file info indexing).

#### Ticket Link
[MM-29256](https://mattermost.atlassian.net/browse/MM-29256)


jespino:
From the QA perspective this is not used anywhere yet, it should be tested when the rest of the feature is done. In subsequent PRs.

jespino:
@lindalumitchell I added the QA tag, but I don't think is necessary here, see my previous comment.

lindalumitchell (approved):

reflog (approved):

grundleborg:
@jespino my question is: why? If you want to add file search, wouldn't this be a better opportunity to make the the Elasticsearch feature more compelling, or if you want it in the Team Edition, to use Bleve for this? I feel like database search is "on the way out" as a feature, so adding a big new thing to it with possible major performance consequences seems like an odd thing to do.

jespino:
@grundleborg good point. I was trying to be consistent with the existing implementation of searching. What do you think @esethna?

jespino:
@grundleborg anyway, I think a lot of customers and team edition users heavily relay on the DB searches, so I'm not sure about making it only available for ES and Bleve.

esethna:
@jespino and I discussed and we're aligned that we should offer this with all search backends (database, ES, bleve). Reason being is that this is a core search feature, and there are a lot of default installs out there using database search where this feature could significantly improve user experience.

esethna (approved):

jespino:
@grundleborg gentle ping here ;)

grundleborg (approved):
Approving on the basis that we are going to set limits on it's usage to avoid terrible performance consequences. I'll also be expecting a load test at the maximum scale we are expecting for this to be carried out before the actual PR containing the feature implementation is merged.

PR #15752, opened 2020-10-03
[GH-15742] Remove direct Store calls from the Api layer for the file api4/system.go

ataboo:
#### Summary
Moved two `Srv().Store` calls used for DB health check from `api4/system.go` to `app.go`.

#### Ticket Link
Fixes https://github.com/mattermost/mattermost-server/issues/15742

### DIFF
--- a/api4/system.go
+++ b/api4/system.go
@@ -97,21 +97,14 @@ func getSystemPing(c *Context, w http.ResponseWriter, r *http.Request) {
 		dbStatusKey := "database_status"
 		s[dbStatusKey] = model.STATUS_OK
 
-		// Database Write Check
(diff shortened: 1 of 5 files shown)


reflog (approved):

jespino (approved):
Nice work! Thanks @ataboo! :tada:

jespino:
I going to do the QA myself, because is an API endpoint changes that is not exposed in the UI and is not easy to tests (make the database read only/stop the database...)

jespino:
QA pass done, and working. I tried it stopping the database and making the database read only, both cases works as expected :)

jespino:
Thanks @ataboo! everything works!

PR #15756, opened 2020-10-03
MM-29264  Fix invalid server version handling / date constraint in InProduct Notices

reflog:
#### Ticket Link

https://mattermost.atlassian.net/browse/MM-29264


ogi-m (approved):
QA will test post-merge

app/product_notices.go
```diff
@@ -77,17 +77,20 @@ func noticeMatchesConditions(config *model.Config, preferences store.PreferenceS
...
-				return false, errors.Wrapf(err, "Cannot parse version range %s", v)
-			}
-			if !c.Check(serverVersion) {
-				return false, nil
+	if cnd.ServerVersion != nil {
+		serverVersion, err := semver.NewVersion(model.BuildNumber)
+		if err != nil {
+			mlog.Warn("Build number is not in semver format", mlog.String("build_number", model.BuildNumber))
+			return false, nil
+		} else {
```
ashishbhate, line 85: non-blocker: With the added return the else is no longer required

app/product_notices.go
```diff
@@ -66,7 +66,7 @@ func noticeMatchesConditions(config *model.Config, preferences store.PreferenceS
 
 	// check if notice date range matches current
 	if cnd.DisplayDate != nil {
-		now := time.Now().UTC()
+		now := time.Now().UTC().Truncate(time.Hour * 24)
```
ashishbhate, line 69: Why do we truncate to hour now?

ashishbhate (approved):

streamer45 (approved):

app/product_notices.go
reflog: @ashishbhate that's to avoid problems like this:
1. PM set condition to "= 2020-03-04T00:00"
2. The user requests notices at 2020-03-04 15:00
3. This condition fails

So we are now truncating to the day to reduce time resolution

app/product_notices.go
```diff
@@ -77,10 +80,23 @@ func noticeMatchesConditions(config *model.Config, preferences store.PreferenceS
...
+		version := strings.Replace(model.BuildNumber, "release-", "", 1)
+		version = rcStripRegexp.ReplaceAllString(version, `$1$3`)
+		versionParts := strings.Split(version, ".")
+		var versionPartsOut []string
+		for _, part := range versionParts {
+			if _, err := strconv.ParseInt(part, 10, 16); err == nil {
+				versionPartsOut = append(versionPartsOut, part)
+			}
+		}
+		version = strings.Join(versionPartsOut, ".")
```
streamer45, line 94: This would probably look better as a standalone function.

reflog: @streamer45 - agreed. will do.

PR #15770, opened 2020-10-05
[MM-29273] Fix flaky test: TestUploadDataConcurrent

streamer45:
#### Summary

PR adds a check to prevent possible inconsistencies in case of parallel upload requests.
While the code was correctly enforcing one request per session at any given time, there was still a case where concurrent requests for the same session could enter the function with an outdated `UploadSession` struct, causing the call to go through without errors.

#### Ticket

https://mattermost.atlassian.net/browse/MM-29273

### DIFF
--- a/app/upload_test.go
+++ b/app/upload_test.go
@@ -278,6 +278,7 @@ func TestUploadDataConcurrent(t *testing.T) {
 
 	// Verify that only 1 request was able to perform the upload.
 	require.Equal(t, int32(n-1), nErrs)
+	nErrs = 0
 
 	wg.Add(n)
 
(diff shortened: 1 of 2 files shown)


app/upload_test.go
```diff
@@ -300,7 +301,7 @@ func TestUploadDataConcurrent(t *testing.T) {
 	wg.Wait()
 
 	// Verify that only 1 request was able to finish the upload.
-	require.Equal(t, int32(n*2-2), nErrs)
+	require.Equal(t, int32(n-1), nErrs)
```
agnivade, line 304: This is not very clear to me. We are still doing (n-1) + (n-1) right? How does it make a difference if we set the variable to 0 in between?

streamer45: The checks are the mathematically the same. Since we want only one call to go through and the rest to error out  (n-1 errors) I actually thought not resetting the counter was making things **less** clear as we had to account for the errors from a previous step, hence the change.

isacikgoz (approved):
LGTM 👍

agnivade (approved):

PR #15788, opened 2020-10-05
Include user data in the cloud endpoints

ethervoid:
#### Summary

This PR includes two new headers that will be sent to CWS: MM_CLOUD_USER_ID and MM_CLOUD_USER_EMAIL.

#### Ticket Link

https://mattermost.atlassian.net/browse/MM-28694

### DIFF
--- a/api4/cloud.go
+++ b/api4/cloud.go
@@ -39,6 +39,12 @@ func getSubscription(c *Context, w http.ResponseWriter, r *http.Request) {
 		return
 	}
 
+	err := setUserDataHeaders(c, w)
+	if err != nil {
+		c.Err = model.NewAppError("Api4.confirmCustomerPayment", "api.cloud.app_error", nil, err.Error(), http.StatusInternalServerError)
+		return
+	}
+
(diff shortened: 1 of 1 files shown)


api4/cloud.go
```diff
@@ -174,3 +204,15 @@ func confirmCustomerPayment(c *Context, w http.ResponseWriter, r *http.Request)
...
+	user, err := c.App.GetUser(c.App.Session().UserId)
+	if err != nil {
+		return err
+	}
+
+	w.Header().Add("MM_CLOUD_USER_ID", user.Id)
+	w.Header().Add("MM_CLOUD_USER_EMAIL", user.Email)
+
+	return nil
+}
```
marianunez, line 218: Shouldn't we need to set this in the request that is created when connecting to CWS at https://github.com/mattermost/enterprise/blob/f22d88ec7fcc3faea5258fba5282e600bba0cfc4/cloud/cloud.go#L229-L243

ethervoid: Yes, I misunderstood the change. Thanks for the heads up!

PR #15818, opened 2020-10-06
[MM-29242] Disable CGO to force cross-compilation

streamer45:
#### Summary

PR fixes an issue with `mattermost-server` local builds failing to run on systems with a different `GLIBC` version.
This happens when the host `GOOS` and `GOARCH` are the same as the target build.
By explicitly setting `CGO_ENABLED=0` we can force a cross compilation and avoid the error.

##### Before

```
github.com/mattermost/mattermost-server » ldd bin/mattermost 
	linux-vdso.so.1 (0x00007ffef899f000)
	libpthread.so.0 => /usr/lib/libpthread.so.0 (0x00007f313a523000)
	libc.so.6 => /usr/lib/libc.so.6 (0x00007f313a35a000)
	/lib64/ld-linux-x86-64.so.2 => /usr/lib64/ld-linux-x86-64.so.2 (0x00007f313a567000)
```

##### After

```
github.com/mattermost/mattermost-server » ldd bin/mattermost
	not a dynamic executable
```

#### Ticket

https://mattermost.atlassian.net/browse/MM-29242


agnivade:
I am 1/5 on not doing this in the server repo. Doing it in the load test repo was okay, but here it has more far reaching consequences.

I would be more comfortable with this being a well-documented behavior that in case of GLIBC version mismatches, we should set CGO_ENABLED to 0, rather than do it always.

Actually I don't think we can even do this. sqlite is a package which uses Cgo. So doing this would effectively prevent the binaries from being used with sqlite I believe. And any other packages which use Cgo.

streamer45:
> I am 1/5 on not doing this in the server repo. Doing it in the load test repo was okay, but here it has more far reaching consequences.

Do you think it could break other things?

> I would be more comfortable with this being a well-documented behavior that in case of GLIBC version mismatches, we should set CGO_ENABLED to 0, rather than do it always.

This can be done of course, just a bit annoying since it might become the default case unless you run exactly the same OS (ubuntu 18.04 I believe, which ships 2.27).

> Actually I don't think we can even do this. sqlite is a package which uses Cgo. So doing this would effectively prevent the binaries from being used with sqlite I believe. And any other packages which use Cgo.

As far as I know `sqlite` is only used when running tests so it shouldn't affect the build.

agnivade:
Yeah, I mean it is plausible that in a customer environment they maybe unknowingly relying on their libc. So this might break those scenarios. Just playing a bit safe here. We don't know all the different environments that are out there in the wild.

> As far as I know sqlite is only used when running tests so it shouldn't affect the build.

Right, usually it won't. But there are some non-test files too, which require different build tags. So normally it won't affect the build, but it's possible to use that build tag and compile mattermost. And now it won't be.

Overall my biggest concern is the breakage of some customer in some crazy linux environment they have, and suddenly Mattermost stops working.

streamer45:
Alright. I'll close this for now then and properly document it on the other side.
If it becomes a frequent issue for others we might revisit this and maybe provide a *static* build target.

agnivade:
>  and maybe provide a static build target.

Makes sense to me.

PR #15820, opened 2020-10-06
Move environment variables to env-file

metanerd:


build/dotenv/test.env
```diff
@@ -0,0 +1,15 @@
+TEST_DATABASE_MYSQL_DSN=mmuser:mostest@tcp(mysql:3306)/mattermost_test?charset=utf8mb4,utf8\u0026readTimeout=30s\u0026writeTimeout=30s
```
agnivade, line 1: While we are here, let's also unescape `\u0026` to just `&` to make things a little bit cleaner.

.circleci/config.yml
```diff
@@ -217,27 +217,15 @@ jobs:
           command: |
             ulimit -n 8096
             mkdir -p mattermost-server/client/plugins
+            cd mattermost-server/build
```
agnivade, line 220: Wondering why do we need to cd here? Is it not possible to do `--env-file=./build/dotenv/test.env`?

metanerd:
```
- attach_workspace:
          at: ~/mattermost
```
So we are in `~/mattermost`. I could also change it to go to mattermost-server/build before the `mkdir -p mattermost-server/client/plugins` part or I could go and load the env file from `--env-file mattermost-server/build/dotenv/test.env`.  But I decided to go with this, because the test-schema followup steps are similar to this. So we would keep it somewhat consistent ¯\_(ツ)_/¯ 
WDYT?

agnivade: Yes that's true. Sounds good to me.

agnivade (approved):

PR #15825, opened 2020-10-06
MM-29223-Retrieve Compliance

sbishel:
#### Summary
Updates for new file locations and name for retrieving message export zip files.

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-29223
https://mattermost.atlassian.net/browse/MM-19640

```release-note
added ability to retrieve compliance files from system console
```

#### Related PRs
https://github.com/mattermost/enterprise/pull/795


api4/job.go
```diff
@@ -45,17 +45,8 @@ func getJob(c *Context, w http.ResponseWriter, r *http.Request) {
...
-		"csv": {
-			"fileName": "csv_export.zip",
-			"fileMime": "application/zip",
-		},
-		"actiance": {
-			"fileName": "actiance_export.zip",
-			"fileMime": "application/zip",
-		},
-	}
+	const FILE_PATH = "export/%s"
```
mkraft, line 48: This path should probably be constructed with something like [`filepath.Join`](https://golang.org/pkg/path/filepath/#Join) so that it works with our Windows builds.

mkraft (changes requested):

hahmadia (approved):
Awesome! Nice way to add this downloading function without removing what's already in place! Great solution Scott 👍

mkraft (approved):

srkgupta (changes requested):
Hi @sbishel 

While testing this PR, I found that its working fine with Local Storage. However if the File Storage is configured to use AWS S3 Bucket, the Download option does not works fine. The downloaded zip file contains empty data. Can you please check and fix this issue. 

Steps to reproduce:
1. In server's config.json file, set DownloadExportResults config to true and restart the server.
2. Configure the File Storage in AWS S3 bucket.
3. Login as sysadmin and visit System Console > Compliance Export page.
4. Configure the Export settings in CSV format and save.
5. Post few messages with attachments.
6. Run the Compliance Export and check if an option to Download files is provided.
7. Click on the Download link and check the contents of the exported zipped file. It should contain contents of the messages posted and also include file attachments

srkgupta (approved):
Thanks for checking this @sbishel. I retested this and its working fine now on the latest master. I am not able to replicate the original issue again. I will go ahead and approve this PR and have a close watch and check if this issue is seen again. If so, it will be raised as a JIRA ticket.

PR #15878, opened 2020-10-08
MM-28249  Auto follow threads

reflog:
#### Ticket Link

https://mattermost.atlassian.net/browse/MM-28249

    ```release-note
	Added support for thread auto follow
    ```


store/sqlstore/thread_store.go
```diff
@@ -106,3 +114,69 @@ func (s *SqlThreadStore) Delete(threadId string) error {
...
+
+		if !errors.As(err, &nfErr) {
+			return errors.Wrap(err, "failed to get thread membership")
+		}
+		_, err = s.SaveMembership(&model.ThreadMembership{
+			PostId:      postId,
+			UserId:      userId,
+			Following:   true,
+			LastViewed:  0,
+			LastUpdated: 0,
```
jespino, line 176: The LastUpdate shouldn't be now?

store/sqlstore/thread_store.go
```diff
@@ -106,3 +114,69 @@ func (s *SqlThreadStore) Delete(threadId string) error {
...
+
+func (s *SqlThreadStore) DeleteMembershipForUser(userId string, postId string) error {
+	if _, err := s.GetMaster().Exec("DELETE FROM ThreadMemberships Where PostId = :PostId AND UserId = :UserId", map[string]interface{}{"PostId": postId, "UserId": userId}); err != nil {
+		return errors.Wrap(err, "failed to update thread membership")
+	}
+
+	return nil
+}
+
+func (s *SqlThreadStore) UpdateMembershipFromMention(userId, postId string) error {
```
jespino, line 163: This naming is leaking bussiness logic into the store layer, maybe something like `CreateMembershipIfNeeded`

store/sqlstore/thread_store.go
```diff
@@ -106,3 +114,69 @@ func (s *SqlThreadStore) Delete(threadId string) error {
...
+		_, err = s.SaveMembership(&model.ThreadMembership{
+			PostId:      postId,
+			UserId:      userId,
+			Following:   true,
+			LastViewed:  0,
+			LastUpdated: 0,
+		})
+		return err
+	}
+	_, err = s.UpdateMembership(membership)
```
jespino, line 180: If there is not error getting the membership we are storing again the same membership, I think makes no sense unless you process it, actually. I think if we are leveraging LastUpdated to have Sync api endpoint, probably we want to update the membership if the following is false, to true, and the LastUpdated if that happens.

store/sqlstore/thread_store.go
```diff
@@ -106,3 +116,73 @@ func (s *SqlThreadStore) Delete(threadId string) error {
...
+	}
+
+	return nil
+}
+
+func (s *SqlThreadStore) CreateMembershipIfNeeded(userId, postId string) error {
+	membership, err := s.GetMembershipForUser(userId, postId)
+	now := utils.MillisFromTime(time.Now())
+	if err == nil {
+		membership.Following = true
```
jespino, line 169: Should we check if the membership is already in `Following == true` before we update it? With the current approach we are trying to create the membership or updating it, and changing the update data, even when is not needed. So... the LastUpdated date is going to change, and is going to generate websocket events, extra load in the future "sync" method for mobile... and extra unneeded writes.

app/notification.go
```diff
@@ -165,14 +165,20 @@ func (a *App) SendNotifications(post *model.Post, team *model.Team, channel *mod
...
 			nErr := a.Srv().Store.Channel().IncrementMentionCount(post.ChannelId, userId)
 			if nErr != nil {
 				umc <- model.NewAppError("SendNotifications", "app.channel.increment_mention_count.app_error", nil, nErr.Error(), http.StatusInternalServerError)
-			} else {
-				umc <- nil
+				return
 			}
-
-			close(umc)
+			if *a.Config().ExperimentalSettings.ThreadAutoFollow && post.RootId != "" {
```
jespino, line 174: I don't think this code belongs here. Looks like you are hijacking the async task to increment the mention count, to do other things. This method is already complex and hard to understand. I would prefer to separate it in its own go routing, and actually encapsulate that in a method to not contribute more to the growth of this method.

app/notification.go
```diff
@@ -165,14 +165,20 @@ func (a *App) SendNotifications(post *model.Post, team *model.Team, channel *mod
...
+				nErr = a.Srv().Store.Thread().CreateMembershipIfNeeded(userId, post.RootId)
+				if nErr != nil {
+					umc <- model.NewAppError("SendNotifications", "app.channel.increment_mention_count.app_error", nil, nErr.Error(), http.StatusInternalServerError)
```
jespino, line 177: The error message here is reused, and should be its own error message.

jespino (changes requested):
Some small changes needed, but looking pretty good.

app/notification.go
ashishbhate: Do you mean a separate go routine in this loop? The code is creating a thread membership for all people mentioned in a thread. It seems appropriate to put it in this loop.

reflog: @jespino - I also think that the placement for this code is not that bad. This method does similar things, I'm just adding one more 'task'

app/notification.go
reflog: Oops. will fix

store/sqlstore/thread_store.go
reflog: Right. Will fix

app/notification.go
jespino: Yes, inside the loop, or later on looping through the `mentionedUsersList`. I don't know what is the best approach there. Probably makes no sense to fail the "notification" if the thread doesn't get followed anyway, so... maybe a different goroutine that does the looping and the execution.

jespino: I disagree with that, the "updateMentionChans" looks like for update mentions count, not for threads autofollow, at least I would have a different name.

At the same time we are binding the success/failure of the incrementnotificationcount to the success or failure of the CreateMembershipIfNeeded, I don't know how much do we want that.

If you feel strong about it, I'm not against it, but looks like there are some rough edges there.

ashishbhate (approved):

app/notification.go
```diff
@@ -160,19 +160,35 @@ func (a *App) SendNotifications(post *model.Post, team *model.Team, channel *mod
...
+			if *a.Config().ExperimentalSettings.ThreadAutoFollow && post.RootId != "" {
+				nErr := a.Srv().Store.Thread().CreateMembershipIfNeeded(userId, post.RootId)
+				if nErr != nil {
+					umc <- model.NewAppError("SendNotifications", "app.channel.autofollow.app_error", nil, nErr.Error(), http.StatusInternalServerError)
+					return
+				}
+			}
+			umc <- nil
+		}(id)
+		updateMentionChans = append(updateMentionChans, umc)
```
jespino, line 177: This is going to create log warnings that was unable to update the channel mention count, that is not related to this. I think we should have another list channels here, to make it independent and generate the logs in the right way.

jespino (changes requested):
Another tiny detail.

app/notification.go
reflog: @jespino done

app/notification.go
```diff
@@ -174,7 +175,7 @@ func (a *App) SendNotifications(post *model.Post, team *model.Team, channel *mod
 			}
 			umc <- nil
 		}(id)
-		updateMentionChans = append(updateMentionChans, umc)
+		mentionAutofollowChans = append(mentionAutofollowChans, umc)
```
jespino, line 178: Probably the channel variable shoud be `mac` here too, instead of reuse the `umc` name.

jespino (approved):
A cosmetic detail, but otherwise looks good to me.

model/config.go
```diff
@@ -876,6 +876,7 @@ type ExperimentalSettings struct {
 	CloudUserLimit                  *int64  `access:"experimental,write_restrictable,cloud_restrictable"`
 	CloudBilling                    *bool   `access:"experimental,write_restrictable,cloud_restrictable"`
 	EnableSharedChannels            *bool   `access:"experimental"`
+	ThreadAutoFollow                *bool   `access:"experimental"`
```
esethna, line 879: @reflog discussing with Jesus, can we please add this to the service settings section of the config?

esethna (changes requested):

model/config.go
reflog: @esethna  done

reflog:
@ogi-m - there's not much to review here, no user facing changes. A general smoke test around creating posts with replies

ogi-m (approved):
Smoke tested posts and post replies, LGTM 👍

esethna (approved):

amyblais:
@esethna Is this for v5.29?

esethna:
@amyblais yes

esethna:
@amyblais noting that this PR should have a note in the compatibility changelog section, config changelog section and important upgrade notes.

Reason it's needed in the compatibility changelog section is to warn admins that it will be turned on by-default and it may have an impact on performance for large scale organizations with a lot of users.

store/sqlstore/thread_store.go
```diff
@@ -49,6 +54,11 @@ func threadToSlice(thread *model.Thread) []interface{} {
 func (s *SqlThreadStore) createIndexesIfNotExists() {
 	s.CreateIndexIfNotExists("idx_threads_last_reply_at", "Threads", "LastReplyAt")
 	s.CreateIndexIfNotExists("idx_threads_post_id", "Threads", "PostId")
+
+	s.CreateIndexIfNotExists("idx_thread_memberships_last_update_at", "ThreadMemberships", "LastUpdated")
+	s.CreateIndexIfNotExists("idx_thread_memberships_last_view_at", "ThreadMemberships", "LastViewed")
+	s.CreateIndexIfNotExists("idx_thread_memberships_post_id", "ThreadMemberships", "PostId")
```
mkraft, line 60: The `idx_thread_memberships_post_id` index is redundant because the primary key automatically creates a multi-column index for which `PostId` is the first column. The DBs should use that index for `PostId` lookups, so creating this index uses unnecessary storage and slows backups, restores, and writes.

reflog: @mkraft what about UserId?

mkraft: I think that's still necessary because it's not the first position of the primary key index. But I would verify it.

mkraft: This shows that the query plan is almost identical with the difference just in the index used:

```
mattermost_test=# explain (format json) select * from ThreadMemberships where PostId = 'foo';
                        QUERY PLAN
-----------------------------------------------------------
 [                                                        +
   {                                                      +
     "Plan": {                                            +
       "Node Type": "Bitmap Heap Scan",                   +
       "Relation Name": "threadmemberships",              +
       "Alias": "threadmemberships",                      +
       "Startup Cost": 4.16,                              +
       "Total Cost": 9.50,                                +
       "Plan Rows": 2,                                    +
       "Plan Width": 157,                                 +
       "Recheck Cond": "((postid)::text = 'foo'::text)",  +
       "Plans": [                                         +
         {                                                +
           "Node Type": "Bitmap Index Scan",              +
           "Parent Relationship": "Outer",                +
           "Index Name": "idx_thread_memberships_post_id",+
           "Startup Cost": 0.00,                          +
           "Total Cost": 4.16,                            +
           "Plan Rows": 2,                                +
           "Plan Width": 0,                               +
           "Index Cond": "((postid)::text = 'foo'::text)" +
         }                                                +
       ]                                                  +
     }                                                    +
   }                                                      +
 ]
(1 row)
```
```
mattermost_test=# drop index idx_thread_memberships_post_id;
DROP INDEX
```
```
mattermost_test=# explain (format json) select * from ThreadMemberships where PostId = 'foo';
                        QUERY PLAN
----------------------------------------------------------
 [                                                       +
   {                                                     +
     "Plan": {                                           +
       "Node Type": "Bitmap Heap Scan",                  +
       "Relation Name": "threadmemberships",             +
       "Alias": "threadmemberships",                     +
       "Startup Cost": 4.16,                             +
       "Total Cost": 9.50,                               +
       "Plan Rows": 2,                                   +
       "Plan Width": 157,                                +
       "Recheck Cond": "((postid)::text = 'foo'::text)", +
       "Plans": [                                        +
         {                                               +
           "Node Type": "Bitmap Index Scan",             +
           "Parent Relationship": "Outer",               +
           "Index Name": "threadmemberships_pkey",       +
           "Startup Cost": 0.00,                         +
           "Total Cost": 4.16,                           +
           "Plan Rows": 2,                               +
           "Plan Width": 0,                              +
           "Index Cond": "((postid)::text = 'foo'::text)"+
         }                                               +
       ]                                                 +
     }                                                   +
   }                                                     +
 ]
(1 row)
```

mkraft: And here's the same analysis done in MySQL:
```
mysql> explain format=json select * from ThreadMemberships where PostId = 'foo';
EXPLAIN: {
  "query_block": {
    "select_id": 1,
    "table": {
      "table_name": "ThreadMemberships",
      "access_type": "ref",
      "possible_keys": [
        "PRIMARY",
        "idx_thread_memberships_post_id"
      ],
      "key": "PRIMARY",
      "used_key_parts": [
        "PostId"
      ],
      "key_length": "106",
      "ref": [
        "const"
      ],
      "rows": 1,
      "filtered": 100,
      "attached_condition": "(`mattermost_test`.`ThreadMemberships`.`PostId` = 'foo')"
    }
  }
}
1 row in set, 1 warning (0.00 sec)
```
```
mysql> drop index idx_thread_memberships_post_id on ThreadMemberships;
Query OK, 0 rows affected (0.01 sec)
```
```
mysql> explain format=json select * from ThreadMemberships where PostId = 'foo';
EXPLAIN: {
  "query_block": {
    "select_id": 1,
    "table": {
      "table_name": "ThreadMemberships",
      "access_type": "ref",
      "possible_keys": [
        "PRIMARY"
      ],
      "key": "PRIMARY",
      "used_key_parts": [
        "PostId"
      ],
      "key_length": "106",
      "ref": [
        "const"
      ],
      "rows": 1,
      "filtered": 100,
      "attached_condition": "(`mattermost_test`.`ThreadMemberships`.`PostId` = 'foo')"
    }
  }
}
1 row in set, 1 warning (0.00 sec)
```

reflog: thanks for the details, @mkraft . I'm going to remove the first one and leave `idx_thread_memberships_user_id` for now.

mkraft:
@reflog Sounds good. And you probably saw my comment that `idx_threads_post_id` can be removed from the `Threads` table too?

reflog:
@mkraft yup. already done in PR: https://github.com/mattermost/mattermost-server/pull/15994

amyblais:
> @amyblais noting that this PR should have a note in the compatibility changelog section, config changelog section and important upgrade notes.
> 
> Reason it's needed in the compatibility changelog section is to warn admins that it will be turned on by-default and it may have an impact on performance for large scale organizations with a lot of users.

@esethna Changelog and upgrade notes done. Does this need docs in https://docs.mattermost.com/administration/config-settings.html yet?

esethna:
@amyblais yes it should

amyblais:
@esethna Please help submit docs for this for https://docs.mattermost.com/administration/config-settings.html.

PR #15880, opened 2020-10-08
[MM-29295] Fix data race in app.CreatePost()

streamer45:
#### Summary

PR fixes a data race caused by sharing the same `*model.Post` between goroutines. We now make a copy
This case is covered by `TestHookMessageHasBeenPosted` which would fail when run with `-race`.

#### Ticket

https://mattermost.atlassian.net/browse/MM-29295


wiggin77 (approved):
LGTM :+1:

app/post.go
```diff
@@ -308,11 +308,13 @@ func (a *App) CreatePost(post *model.Post, channel *model.Channel, triggerWebhoo
 	// might be duplicating requests.
 	a.Srv().seenPendingPostIdsCache.SetWithExpiry(post.PendingPostId, rpost.Id, PENDING_POST_IDS_CACHE_TTL)
 
+	// We make a copy of the post for the plugin hook to avoid a race condition.
+	postCopy := rpost.Clone()
```
agnivade, line 312: 0/5, perhaps `rPostCopy` makes it clearer.

agnivade (approved):

streamer45:
@lindalumitchell 

This PR fixes an existing race condition triggered by the `MessageHasBeenPosted` plugin hook.
It does not alter functionality and the case is covered by unit tests. I've also left QA steps to verify this.

prapti (approved):
Tested on local server using demo plugin that registers the `MessageHasBeenPosted` hook. There is no race related error noted in the Jira ticket.
Approving.
P.s.: Thank you for the patience through all my questions @streamer45!

PR #15608, opened 2020-09-25
prepackage incident response v0.7

lieut-data:
#### Summary
Deliver Incident Response v0.7 to cloud beta customers.

PR #15900, opened 2020-10-08
MM-29354 - Remove pre-packaged incident response for v5.28

cpoile:
#### Summary
- (from ticket): We have been using pre-packaging as a means to distribute incident response automatically to cloud, but the version currently available is not final and should not be included in v5.28 proper. Let's remove it, as we intend to list in the marketplace first for on-premise customers.
- Revert of #15608 

#### Ticket Link
- https://mattermost.atlassian.net/browse/MM-29354

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -95,7 +95,6 @@ PLUGIN_PACKAGES += mattermost-plugin-antivirus-v0.1.2
 PLUGIN_PACKAGES += mattermost-plugin-jira-v2.3.2
 PLUGIN_PACKAGES += mattermost-plugin-gitlab-v1.1.0
 PLUGIN_PACKAGES += mattermost-plugin-jenkins-v1.0.0
-PLUGIN_PACKAGES += mattermost-plugin-incident-response-v0.6.0
 
 # Prepares the enterprise build if exists. The IGNORE stuff is a hack to get the Makefile to execute the commands outside a target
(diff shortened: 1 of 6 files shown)


crspeller (approved):

amyblais:
@lieut-data I believe this is for today's v5.28 code freeze deadline.

lieut-data (approved):

lieut-data:
@lindalumitchell, my gut says we don't need explicit coverage here, but wanted to run this by you.

lindalumitchell (approved):

agnivade:
I think this PR incorrectly added empty translation strings to en.json which were removed in this PR (https://github.com/mattermost/mattermost-server/pull/15718). The CI might have passed in the span of time when the PR will not have gone in and /update-branch has not been run yet again.


So i18n-extract in 5.28 is broken right now.

We need to remove those blank strings.

cpoile:
@agnivade Sorry for the trouble -- this PR was failing on CI, the only way I could make it pass was to update the translations (and that was after an update-branch). I wonder why this PR had that bad state? Anyway, thanks for fixing the problem.

agnivade:
No worries, these things happen from time to time. I think it was failing because Hossein's changes hadn't gone in yet. And then both this PR and Hossein's revert went in, breaking it again.

PR #15909, opened 2020-10-08
Update golangci-lint to v1.31.0

camgraff:
Fixes https://github.com/mattermost/mattermost-server/issues/15901


hanzei (approved):
That was easier then expected :+1:

agnivade (approved):

agnivade:
Note that this should not be merged until https://github.com/mattermost/enterprise/pull/799 has gone in. Both should be merged together.

PR #15932, opened 2020-10-12
MM-29548: remove ability to update SAML admin filter settings

sbishel:
#### Summary
Prevents non-system-admins from viewing or updating the SAML admin filter settings.

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-29548

### DIFF
--- a/model/config.go
+++ b/model/config.go
@@ -2212,8 +2212,8 @@ type SamlSettings struct {
 	// User Mapping
 	IdAttribute          *string `access:"authentication"`
 	GuestAttribute       *string `access:"authentication"`
-	EnableAdminAttribute *bool   `access:"authentication"`
-	AdminAttribute       *string `access:"authentication"`
+	EnableAdminAttribute *bool
+	AdminAttribute       *string
 	FirstNameAttribute   *string `access:"authentication"`
 	LastNameAttribute    *string `access:"authentication"`
 	EmailAttribute       *string `access:"authentication"`


model/config.go
```diff
@@ -2212,8 +2212,8 @@ type SamlSettings struct {
 	// User Mapping
 	IdAttribute          *string `access:"authentication"`
 	GuestAttribute       *string `access:"authentication"`
-	EnableAdminAttribute *bool   `access:"authentication"`
-	AdminAttribute       *string `access:"authentication"`
```
hahmadia, line 2216: Just want to ask since it's being highlighted in this PR but what does this access label do for these settings? Any documentation I could read more about since I was always unsure about it.

hahmadia (changes requested):
LGTM! Only leaving a `requested changes` so I can be re-requested review for the follow up answer on the question I've left. Nothing actually requires changing. Thanks

srkgupta (approved):
Tested and the issue is fixed now and is working fine. Also did some regression testing and no issues were found. Approving the PR. 

@sbishel please feel free to re-request my review if there are significant changes due to Dev Reviews. In order to speed up the process, I started the QA testing even before the Dev Review was complete.

hahmadia (approved):

model/config.go
mkraft: @hahmadia there's some Godoc related to it: https://github.com/mattermost/mattermost-server/blob/2bf6992397b50f2c5a7cf84480daed39a5b4b0fa/model/config.go#L2829-L2859

mkraft (approved):

PR #15948, opened 2020-10-13
MM-29573: Reuse s3 client amongst all requests

agnivade:
An http.Client is meant as a long-term object to be reused between
consecutive requests. This reduces burden on HTTP servers by reusing TCP connections
and also makes things faster by avoiding the TCP handshake for every single request.

https://mattermost.atlassian.net/browse/MM-29573

### DIFF
--- a/go.tools.mod
+++ b/go.tools.mod
@@ -4,7 +4,7 @@ go 1.14
 
 require (
 	github.com/jstemmer/go-junit-report v0.9.1 // indirect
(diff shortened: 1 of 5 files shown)


grundleborg (approved):
LGTM

isacikgoz (approved):
LGTM 👍

lindalumitchell (approved):

PR #15949, opened 2020-10-13
MM-29525: Strip path_prefix from ListDirectory

agnivade:
An AWS path prefix is meant to be an implementation detail
which the calling application should not be aware of. Hence, we should
strip the path prefix when returning objects in a directory
because they anyways get applied while using the other APIs

https://mattermost.atlassian.net/browse/MM-29525


mickmister (approved):
Awesome, LGTM :+1:

agnivade:
@lindalumitchell - This needs some testing around data retention which I have mentioned in the test steps.

srkgupta (approved):
Tested the Data Retention Job using S3 as File Storage and with & without Path Prefix and the files & posts were deleted successfully and no errors were shown. Approving the PR

PR #15953, opened 2020-10-14
[MM15037] Stop sending notifications to user whose Oauth bot posted

camgraff:
Fixes https://github.com/mattermost/mattermost-server/issues/15928

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-15037

#### Release Note
```release-note
Stop sending notifications to user whose Oauth bot posted
```


streamer45:
Thank you for your contribution @camgraff :+1: . I'll proceed to code review shortly.

@esethna 

The proposed changes will make it so that any post created from a webhook (whether it was created from an OAuth2 app/bot or not) will not trigger a mention notification for the user associated with the root post. 

I just want to make sure this is what is expected/requested and point out that some MM users might be relying on mentions to get sent in such cases and this would change that, possibly causing an unexpected behaviour.

I am also wondering if we should add an additional check to verify if the username was overridden, and only in such case prevent the notification from getting sent.

esethna:
@streamer45 totally agree. Only the case that the username is overridden should we be silencing those notifications ^

esethna:
@camgraff thanks for this PR!! Can't wait till these notifications are silenced 🥇

camgraff:
Made changes to require that username is overriden and post is from webhook in order to prevent notifications :slightly_smiling_face:

model/post.go
```diff
@@ -665,3 +665,7 @@ func RewriteImageURLs(message string, f func(string) string) string {
 
 	return string(result)
 }
+
+func (p *Post) IsFromOAuthBot() bool {
```
streamer45, line 669: This is causing the linter checks to fail cause of an inconsistency in receiver naming. Although `p` is the right choice we are currently using the letter `o` in all the other occurrences. We could either take this time to fix them all if you wish or you can just use `o` to make it pass.

model/post.go
```diff
@@ -665,3 +665,7 @@ func RewriteImageURLs(message string, f func(string) string) string {
...
+	return p.GetProp("from_webhook") == "true" && p.GetProp("override_username") != ""
```
streamer45, line 670: Since we are checking multiple props I'd suggest to use `p.GetProps()` once and check on the returned value.
This is mainly to avoid additional locking since post props are guarded by a mutex.

streamer45 (changes requested):
Thank you for this PR @camgraff :+1:
I've left a couple of suggestions and a hint on how to fix the build failure.

app/notification_test.go
```diff
@@ -68,6 +68,53 @@ func TestSendNotifications(t *testing.T) {
...
+				Message:   "a reply",
+			}
+			childPost, appErr = th.App.CreatePostMissingChannel(childPost, false)
+			require.Nil(t, appErr)
+
+			postList := model.PostList{
+				Order: []string{rootPost.Id, childPost.Id},
+				Posts: map[string]*model.Post{rootPost.Id: rootPost, childPost.Id: childPost},
+			}
+			mentions, err := th.App.SendNotifications(childPost, th.BasicTeam, th.BasicChannel, th.BasicUser2, &postList, true)
```
streamer45, line 99: This line is causing the linter to fail due to [shadowing](https://yourbasic.org/golang/gotcha-shadowing-variables/) the `err` variable.
What that means is that we are declaring it again when we already had one defined before, in this case at line 32.
If it helps you can run the `make check-style` locally to easily check for these errors.

camgraff: Thanks :slightly_smiling_face:

streamer45 (approved):
Thank you @camgraff , PR looks good! :+1: 

Can you help merging with master and push so we can get a green build?

gabrieljackson (approved):
Nice work and good tests!

prapti (approved):
The change works as expected -- tested in the PR test server: the @-mention is not triggered  when a reply is posted to the OAuth app bot post.
Thank you for providing the solution @camgraff! 🎉 

@aaronrothschild I did the QA review before yours. Would you mind giving this a quick look? If there is no change request from you I think it's ready for merge, otherwise I can review it again :)

jasonblais:
@aaronrothschild friendly reminder to help with reviews :)

jfrerich:
@aaronrothschild friendly reminder to help with reviews :)

aaronrothschild (approved):
Sorry!  Late review here, but LGTM!  Thanks @camgraff

PR #15956, opened 2020-10-14
MM-28490: Add channel object to websocket posted event

agnivade:
https://mattermost.atlassian.net/browse/MM-28490


agnivade:
@hmhealey / @nevyangelova - Is the channel object needed to fix the bug that we are seeing, or is more of a good to have? From the issue, it wasn't clear to me if this is actual fix or a potential fix. Is there a way to verify that?

I am a bit wary about adding more allocations to the createPost path without really understanding if it fixes the problem at hand.

agnivade:
Discussed offline. Apps team to take it forward.

PR #15958, opened 2020-10-14
[MM-29646] Add check for IsPaidTier, and don't send emails over limit for guest invites

nickmisasi:
#### Summary
When a user upgraded their cloud subscription, the backend would still restrict them from sending more invites than the user limit. This PR adds a check for IsPaidTier for members.

Until today, guest invites were not included in the checking for over user limit when sent. This PR adds the backend check for this, along with: https://github.com/mattermost/mattermost-webapp/pull/6783
#### Ticket Link
https://mattermost.atlassian.net/browse/MM-29646
https://mattermost.atlassian.net/browse/MM-29630

```release-note
NONE
```


api4/team.go
```diff
@@ -1200,8 +1200,11 @@ func inviteUsersToTeam(c *Context, w http.ResponseWriter, r *http.Request) {
 	if graceful {
 		cloudUserLimit := *c.App.Config().ExperimentalSettings.CloudUserLimit
 		var invitesOverLimit []*model.EmailInviteWithError
-		if cloudUserLimit > 0 && c.IsSystemAdmin() {
-			emailList, invitesOverLimit, _ = c.App.GetErrorListForEmailsOverLimit(emailList, cloudUserLimit)
+		if *c.App.Srv().License().Features.Cloud && cloudUserLimit > 0 && c.IsSystemAdmin() {
```
ethervoid, line 1203: You also have to check if the license is not nil because it can be

api4/team.go
```diff
@@ -1200,8 +1200,11 @@ func inviteUsersToTeam(c *Context, w http.ResponseWriter, r *http.Request) {
...
+			subscription, _ := c.App.Cloud().GetSubscription()
+			if subscription.IsPaidTier != "true" || subscription == nil {
```
ethervoid, line 1205: I'd check first if the subscription is nil because if it is we're going to get a panic
```suggestion
			if subscription == nil || subscription.IsPaidTier != "true" {
```

api4/team.go
```diff
@@ -1279,7 +1282,26 @@ func inviteGuestsToChannels(c *Context, w http.ResponseWriter, r *http.Request)
 	auditRec.AddMeta("channels", guestsInvite.Channels)
 
 	if graceful {
-		invitesWithError, err := c.App.InviteGuestsToChannelsGracefully(c.Params.TeamId, guestsInvite, c.App.Session().UserId)
+		cloudUserLimit := *c.App.Config().ExperimentalSettings.CloudUserLimit
+		var invitesOverLimit []*model.EmailInviteWithError
+		if *c.App.Srv().License().Features.Cloud && cloudUserLimit > 0 && c.IsSystemAdmin() {
```
ethervoid, line 1287: Same here

api4/team.go
```diff
@@ -1279,7 +1282,26 @@ func inviteGuestsToChannels(c *Context, w http.ResponseWriter, r *http.Request)
...
+			subscription, _ := c.App.Cloud().GetSubscription()
+			if subscription.IsPaidTier != "true" || subscription == nil {
```
ethervoid, line 1289:
```suggestion
			if subscription == nil || subscription.IsPaidTier != "true"  {
```

api4/team.go
```diff
@@ -1200,8 +1200,11 @@ func inviteUsersToTeam(c *Context, w http.ResponseWriter, r *http.Request) {
 	if graceful {
 		cloudUserLimit := *c.App.Config().ExperimentalSettings.CloudUserLimit
 		var invitesOverLimit []*model.EmailInviteWithError
-		if cloudUserLimit > 0 && c.IsSystemAdmin() {
-			emailList, invitesOverLimit, _ = c.App.GetErrorListForEmailsOverLimit(emailList, cloudUserLimit)
+		if *c.App.Srv().License().Features.Cloud && cloudUserLimit > 0 && c.IsSystemAdmin() {
+			subscription, _ := c.App.Cloud().GetSubscription()
```
ethervoid, line 1204: We shouldn't avoid check or act in case of errors. In this case, if we don't want to react to the error because is nil we should at least log it so we know what is happening

api4/team.go
```diff
@@ -1279,7 +1282,26 @@ func inviteGuestsToChannels(c *Context, w http.ResponseWriter, r *http.Request)
 	auditRec.AddMeta("channels", guestsInvite.Channels)
 
 	if graceful {
-		invitesWithError, err := c.App.InviteGuestsToChannelsGracefully(c.Params.TeamId, guestsInvite, c.App.Session().UserId)
+		cloudUserLimit := *c.App.Config().ExperimentalSettings.CloudUserLimit
+		var invitesOverLimit []*model.EmailInviteWithError
+		if *c.App.Srv().License().Features.Cloud && cloudUserLimit > 0 && c.IsSystemAdmin() {
+			subscription, _ := c.App.Cloud().GetSubscription()
```
ethervoid, line 1288: Same here

ethervoid (changes requested):
Some changes

api4/team.go
nickmisasi: Meant to add a check but forgot after testing. Added one to check for the error being nil

ethervoid (approved):
LGTM!

marianunez (approved):
LGTM. Let's create a follow up ticket to move those tests to enterprise.

nickmisasi:
@marianunez https://mattermost.atlassian.net/browse/MM-29683

PR #15959, opened 2020-10-14
DOPS-243: Add manifest.txt

metanerd:

### DIFF
--- a/build/release.mk
+++ b/build/release.mk
@@ -108,8 +108,8 @@ else
 endif
 	cp NOTICE.txt $(DIST_PATH)
 	cp README.md $(DIST_PATH)
-	if [ -f manifest.txt ]; then \
-		cp manifest.txt $(DIST_PATH); \
+	if [ -f ../manifest.txt ]; then \
+		cp ../manifest.txt $(DIST_PATH); \
 	fi; \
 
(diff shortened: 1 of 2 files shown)


khos2ow (approved):

build/release.mk
```diff
@@ -108,6 +108,9 @@ else
 endif
 	cp NOTICE.txt $(DIST_PATH)
 	cp README.md $(DIST_PATH)
+	if [ -f manifest.txt ]; then \
+		cp manifest.txt $(DIST_PATH); \
+	fi; \
```
khos2ow, line 113: `; \` at the end of the line should be removed.

khos2ow (changes requested):

build/release.mk
metanerd: Which line? The whole file has a semicolon at the end though. It would keep it consistent.

khos2ow: only line 113, since it's the last line of multi-line if block. `;` does nothing special here but `\` is going to try to append the next line to it.

metanerd, line 113:
```suggestion
	fi
```

khos2ow (approved):

jaydeland (approved):
LGTM

PR #15990, opened 2020-10-16
[MM-29725] Add websocket event for when user activation status changes

nickmisasi:
#### Summary
In cloud, admins will be able to activate and deactivate users to go over/under the user limit for the free tier. The number of users is important to the front end because it can trigger quite a few different warning banners, so we want the front-end to always be up to date. This PR adds a websocket event for when a users activation state changes, so that the front end can update its user count.

#### Ticket Link
**This doesn't apply directly to this ticket, but would be a logical next bug report I figured I'd fix now.
https://mattermost.atlassian.net/browse/MM-29725


marianunez (approved):
Thanks @nickmisasi!

ethervoid (approved):
LGTM

PR #15994, opened 2020-10-18
MM-29703  Mark threads as read when channels are marked

reflog:
#### Ticket Link

https://mattermost.atlassian.net/browse/MM-29703


store/sqlstore/thread_store.go
```diff
@@ -188,3 +188,47 @@ func (s *SqlThreadStore) CreateMembershipIfNeeded(userId, postId string) error {
...
+	var threadData []struct {
+		PostId      string
+		ChannelId   string
+		LastReplyAt int64
+	}
+	query, args, _ := s.getQueryBuilder().
+		Select("PostId, ChannelId, LastReplyAt").
+		From("Threads").
+		LeftJoin("Posts ON Threads.PostId=Posts.Id").
+		Where(sq.Eq{"Posts.ChannelId": channelIds}).
```
jespino, line 206: Can we filter here by the current LastUpdated to avoid unnecesary writes? I mean, for example, something like, if the `LastReplyAt < Min(LastReplyAt of channels)`

store/sqlstore/thread_store.go
```diff
@@ -188,3 +188,47 @@ func (s *SqlThreadStore) CreateMembershipIfNeeded(userId, postId string) error {
 	})
 	return err
 }
+
+func (s *SqlThreadStore) UpdateUnreadsByChannel(userId string, channelLastUnreads map[string]int64) error {
```
jespino, line 192: Because this is olny used internally, there is not reason to expose it publicly. I would convert it into a private method.

jespino (changes requested):
Looking pretty good, I think we can improve a bit the performance here.

store/sqlstore/thread_store.go
reflog: @jespino - 'internally' is subjective :) It's being used by App methods, so I do need to export the method...

store/sqlstore/thread_store.go
reflog: @jespino Sounds good, will do

store/sqlstore/thread_store.go
jespino: I don't see any call done from app layer in this pr, looks like this is call from other store methods, so it doesn't need to be directly exposed, or I'm missing something?

reflog: @jespino  :man_facepalming: you are correct. will fix

reflog: @jespino  Nope, you managed to confuse me. :)
You cannot make this method private, since it's being called from another Store class by it's interface.
And if the method is private, it cannot be used like:
`	go s.Thread().updateUnreadsByChannel(userId, map[string]int64{channelId: now}) `

jespino: We discuss this through videoconference and looks like the current way we define the SqlStore doesn't allow this kind of private methods, so we need to address that later in another ticket. For now this is fine.

store/sqlstore/thread_store.go
```diff
@@ -191,19 +192,27 @@ func (s *SqlThreadStore) CreateMembershipIfNeeded(userId, postId string) error {
...
 	}
+
 	query, args, _ := s.getQueryBuilder().
 		Select("PostId, ChannelId, LastReplyAt").
 		From("Threads").
 		LeftJoin("Posts ON Threads.PostId=Posts.Id").
-		Where(sq.Eq{"Posts.ChannelId": channelIds}).
+		Where(sq.And{
+			sq.Eq{"Posts.ChannelId": channelIds},
+			sq.Lt{"Threads.LastReplyAt": minLastUpdate},
```
jespino, line 214: You are getting all the Threads that the LastReplyAt is less than the minLastUpdate, so... you are getting the ones that doesn't need update, I think.

store/sqlstore/thread_store.go
```diff
@@ -188,3 +189,55 @@ func (s *SqlThreadStore) CreateMembershipIfNeeded(userId, postId string) error {
...
+		return errors.Wrap(err, "begin_transaction")
+	}
+	defer finalizeTransaction(transaction)
+
+	for _, data := range threadData {
+		if data.LastReplyAt < channelLastUnreads[data.ChannelId] {
+			updateQuery, updateArgs, _ := s.getQueryBuilder().
+				Update("ThreadMemberships").
+				Where(sq.Eq{"UserId": userId, "PostId": data.PostId}).
+				Set("LastUpdated", channelLastUnreads[data.ChannelId]).
```
ashishbhate, line 231: Why dont we update `LastViewed`?

reflog: Great point @ashishbhate , will update

store/sqlstore/channel_store.go
```diff
@@ -2104,6 +2108,19 @@ func (s SqlChannelStore) UpdateLastViewedAt(channelIds []string, userId string)
 		lastViewedQuery += fmt.Sprintf("WHEN :channelId%d THEN GREATEST(LastViewedAt, :lastViewed%d) ", index, index)
 
 		props["channelId"+strconv.Itoa(index)] = t.Id
+
+		if t.LastPostAt < minLastPostAt {
+			minLastPostAt = t.LastPostAt
```
jespino, line 2113: Why the min, and not the max? (The query say GREATEST)

store/sqlstore/channel_store.go
```diff
@@ -2121,6 +2138,9 @@ func (s SqlChannelStore) UpdateLastViewedAt(channelIds []string, userId string)
 		return nil, errors.Wrapf(err, "failed to update ChannelMembers with userId=%s and channelId in %v", userId, channelIds)
 	}
 
+	if updateThreads {
+		go s.Thread().UpdateUnreadsByChannel(userId, threadsToUpdate, utils.MillisFromTime(time.Now()))
```
jespino, line 2142: Why here is "now" when in the other place we are using the GREATEST?

store/sqlstore/channel_store.go
```diff
@@ -2168,7 +2188,16 @@ func (s SqlChannelStore) CountPostsAfter(channelId string, timestamp int64, user
...
+	var threadsToUpdate []string
+	if updateThreads {
+		var err error
+		threadsToUpdate, err = s.Thread().CollectThreadsWithNewerReplies(userID, []string{unreadPost.ChannelId}, unreadPost.CreateAt-1)
+		if err != nil {
+			return nil, err
+		}
+	}
+
 	unreadDate := unreadPost.CreateAt - 1
```
jespino, line 2201: Probably we should define this variable at the top, and use the same one in the CollectThreadsWithNewerReplies function.

store/sqlstore/channel_store.go
```diff
@@ -2225,10 +2254,27 @@ func (s SqlChannelStore) UpdateLastViewedAtPost(unreadPost *model.Post, userID s
 	if err = s.GetMaster().SelectOne(result, chanUnreadQuery, params); err != nil {
 		return nil, errors.Wrapf(err, "failed to get ChannelMember with channelId=%s", unreadPost.ChannelId)
 	}
+
+	go s.Thread().UpdateUnreadsByChannel(userID, threadsToUpdate, result.LastViewedAt)
```
jespino, line 2258: I miss here the `if updateThread`

store/sqlstore/channel_store.go
```diff
@@ -2225,10 +2254,27 @@ func (s SqlChannelStore) UpdateLastViewedAtPost(unreadPost *model.Post, userID s
...
 	return result, nil
 }
 
-func (s SqlChannelStore) IncrementMentionCount(channelId string, userId string) error {
+func (s SqlChannelStore) IncrementMentionCount(channelId string, userId string, updateThreads bool) error {
+	var threadsToUpdate []string
+	if updateThreads {
+		var err error
+		timestamp, err := s.GetReplica().SelectInt("SELECT LastUpdateAt FROM ChannelMembers WHERE UserId = :UserId AND ChannelId = :ChannelId", map[string]interface{}{"ChannelId": channelId, "UserId": userId})
```
jespino, line 2266: This timestmap is not needed, that is get internally in the database, the one that you need is the `now`, so you can define the `now` up in the function, and use that now here, in the update of the channelMembership and in the update of the threads.

reflog: @jespino - we are using 'now' after the update. Before the update - we want the actual timestamp I think

store/sqlstore/channel_store.go
reflog: @jespino - the query gets the greatest, but of all the selected ones, I am taking the earliest one so that I don't miss anything in the update later on.

store/sqlstore/channel_store.go
jespino: The actual timestamp is get as part of the `ChannelMembers.LastViewAt` in the query inside.

reflog: @jespino that's LastViewAt, here I am fetching LastUpdateAt and it's used for comparison:
```go
query, args, _ := s.getQueryBuilder().
		Select("Threads.PostId").
		From("Threads").
		LeftJoin("Posts ON Threads.PostId=Posts.Id").
		LeftJoin("ChannelMembers ON ChannelMembers.ChannelId=Posts.ChannelId").
		Where(sq.And{
			sq.Eq{"Posts.ChannelId": channelIds},
			sq.Eq{"ChannelMembers.UserId": userId},
			sq.Or{
				sq.Expr("Threads.LastReplyAt >= ChannelMembers.LastViewedAt"),
				sq.GtOrEq{"Threads.LastReplyAt": timestamp},
			},
		}).
		ToSql()
```

reflog:
@jespino I've addressed all things we've discussed. The only oddity is - one of the tests is failing in Postgres mode. I can't figure out why :scream: . Any help is very appreciated

store/sqlstore/channel_store.go
```diff
@@ -2094,6 +2094,7 @@ func (s SqlChannelStore) UpdateLastViewedAt(channelIds []string, userId string)
```
jespino, line 2094: Right before this (I can't comment not-in-context line) you have a early return on postgres, that is the reason why postgres is not doing the update at all.

esethna:
@ashishbhate can you help do a re-review and then approval so this is ready to merge after the above issue is fixed? ^

ashishbhate (approved):

store/sqlstore/channel_store.go
```diff
@@ -2084,6 +2084,21 @@ func (s SqlChannelStore) UpdateLastViewedAt(channelIds []string, userId string,
 		return nil, store.NewErrInvalidInput("Channel", "Id", fmt.Sprintf("%v", channelIds))
 	}
 
+	var threadsToUpdate []string
```
jespino, line 2087: The first half of this block should be before the Select query is executed, because for postgres this query is not only the select, is also the update, so you are not following the approach of using the "current existing LastViewAt" of the channel membership to detect the threads to mark as read.

jespino (approved):
Now looks good to me. Thanks @reflog!

esethna:
@jespino @reflog what does QA need to do here? Any regressions or things to test for?

esethna:
cc// @ogi-m ^

esethna (approved):

jespino:
@esethna @ogi-m I think the QA should do a bit of smoke testing around channel  unreads, verifying that everything is marked as read as expected (not the threads, because that is not exposed yet), and verify that the mark as unread works as expected too. I don't see anything else needed to be tested here.

reflog:
@esethna @jespino load testing done. seems ok, waiting for @streamer45 to comment
https://mattermost.atlassian.net/browse/MM-29961

jespino:
Yes, looks good, awesome! 🎊

esethna:
@ogi-m how is testing going on this?

ogi-m (approved):
Tested and passed - smoke tested around channels - channel unreads, marking posts as unread, etc and everything is working as expected 👍

store/sqlstore/channel_store.go
```diff
@@ -2089,11 +2099,15 @@ func (s SqlChannelStore) UpdateLastViewedAt(channelIds []string, userId string)
 		for _, t := range lastPostAtTimes {
 			times[t.Id] = t.LastPostAt
 		}
+		if updateThreads {
+			go s.Thread().UpdateUnreadsByChannel(userId, threadsToUpdate, now)
```
agnivade, line 2103: My apologies for an unsolicited comment, I just happened to notice this PR by accident.

I think this has possibility to cause more harm than benefit. I'd rather trade-off slower response times than higher throughput at the expense of increased DB load. We should be a bit careful when spinning up random goroutines without control. If this is a performance critical part that must be asynchronous, I would suggest creating a separate goroutine which receives such commands and batches them and converts them into a single query. That makes for a much controlled workload.

Otherwise, let's make it synchronous.

Same in L2139 too.

Again, sorry to suddenly barge in a review like this.

reflog: @agnivade - no worries, all reviews are welcome :+1: @jespino thoughts?

esethna: How does the proposal above ^ change user experience?

jespino: We need this merged in ASAP, can we try making it synchronous and execute a new performance test?

reflog: ok, i'll run load tests without goroutines and compare the results

agnivade: I think @streamer45 was already planning to run a 1M test. Claudio - would you be open to running another one by making these calls synchronous? Or let me know I can do that too.

The bigger question is:
- What is the SLA that we are agreeing upon for the delay in marking channels as viewed? From the preliminary tests, it looks like a 2x slowdown. And it will be slightly higher with more channels and more channel members.
- Let's say 2x is acceptable. Now how much more leeway is allowed for it to be slower?

In any case, we should wait before the load tests are done. Following the recent discussion in guild meeting, I would like all performance related concerns to be understood and accepted by everyone before we move on with this.

streamer45:
These are the results of comparative load-tests taken between this PR and its base commit (`f46661d7a`) on an instance cluster composed as such:

- 2 App servers
- 2 DBs cluster (1 reader + 1 writer)
- 1 load balancer in front (nginx)

The instance was pre-loaded with **2.5M** posts in 2 teams and 1000 channels per team.

### Unbounded load-test results

#### base

```
==================================================
load-test status:

State: done
Start time: Tue Oct 27 10:55:27 UTC 2020
Stop time: Tue Oct 27 11:39:35 UTC 2020
Duration: 44m9s
Active users: 0
Number of errors: 30042
Supported users: 3146
==================================================
```

#### MM-29703

```
==================================================
load-test status:

State: done
Start time: Tue Oct 27 13:12:20 UTC 2020
Stop time: Tue Oct 27 14:01:15 UTC 2020
Duration: 48m55s
Active users: 0
Number of errors: 19264
Supported users: 1183
==================================================
```

#### Bounded load-test results (2000 users)







#### Comments

- Both unbounded and bounded load-tests show critical performance degradation.
- Estimated supported users dropped by more than 60%. 
- Store times have increased overall as the DB, under same conditions, was not able to keep up with the added queries. This is of course accentuated by the bigger data set used, specifically the `Posts` table which is involved in the affected store calls.
- Several graphs show a sudden spike (goroutines related), to be further investigated, but possibly explained by the DB not able to cope with increased workload.

#### Conclusion 

I think the data is fairly self-explanatory, everything pointing to a severe worsening in performance affecting a core feature (messaging).
At this point, I wouldn't feel comfortable with merging this PR as it's clearly going to affect the overall performance of the application. I think we might want to rethink the approach a bit as I am not sure there is much margin to gain with possible tweaks to the queries. That said, I'll be available to run more tests/analysis when/if needed.

Full data (and graphs) can be accessed at:

http://54.83.109.237:3000/d/zX6_4UtMz/mm-29703-comparison-2000-users?orgId=1

agnivade:
The spike in goroutines matches with a spike in the number of master DB connections. I have a feeling that it might be related to the extra goroutines being spawned. If I have to guess, at some point, goroutine contention became pretty bad and then suddenly all of them were unleashed. It would be good to have a run with no goroutines to confirm this theory.

esethna:
FYI for those following this thread, it seems there may be either:

1) a bug in the query that's causing full table scans that are severely degrading perf, or
2) an issue in the load test related to very close timestamps of threads, resulting in full table scans that are severely degrading perf,
3) An issue with "LastViewedAt" not updating correctly.

Jesus, Eli and Claudio are investigating the above right now. If the approach still causes severe perf degradation then we'll need to cut this from the release and rethink the approach. cc// @amyblais

amyblais:
@esethna Let me know the decision on this as I'll start cutting RC-1 in a couple hours today.

streamer45:
Here are the results of new load-tests for this PR which were run with the latest changes:

### Unbounded load-test

#### PostgreSQL

```
================================================== base
load-test status:

State: done
Start time: Thu Oct 29 09:47:51 UTC 2020
Stop time: Thu Oct 29 10:29:10 UTC 2020
Duration: 41m18s
Active users: 0
Number of errors: 20734
Supported users: 2420
==================================================
```

```
================================================== MM-29703
load-test status:

State: done
Start time: Thu Oct 29 08:43:50 UTC 2020
Stop time: Thu Oct 29 09:26:04 UTC 2020
Duration: 42m13s
Active users: 0
Number of errors: 21991
Supported users: 2317
==================================================
```

#### MySQL

```
================================================== base
load-test status:

State: done
Start time: Thu Oct 29 14:05:29 UTC 2020
Stop time: Thu Oct 29 15:03:01 UTC 2020
Duration: 57m32s
Active users: 0
Number of errors: 36017
Supported users: 3135
==================================================
```

```
================================================== MM-29703
load-test status:

State: done
Start time: Thu Oct 29 12:31:28 UTC 2020
Stop time: Thu Oct 29 13:15:22 UTC 2020
Duration: 43m54s
Active users: 0
Number of errors: 12897
Supported users: 3076
==================================================
```

### Bounded load-test (2000 users, PostgreSQL) 

image
image
image
image

#### Comments

- Average API times are equal. 99-percentiles have slightly worsened (expected).
- Average duration of the affected store calls has increased as expected but does not show spikes.
- Memory consumption shows a noticeable decrease, possibly caused by the reduced amount of DB connections, not necessarily related to changes (needs further investigation).
- Unbounded load-tests results do not show performance worsening. Estimated user counts are very close on both DB engines.
- Overall, no significant performance degradation has been noticed.

#### Data:

PostgreSQL: 

http://54.83.109.237:3000/d/zX6_4UtMa/mm-29703-comparison-2000-users-new?orgId=1&from=1603975195606&to=1603978650000&var-server=app-0:8067&var-server=app-1:8067

MySQL: 

http://52.91.97.43:3000/d/000000011/mattermost-load-test-performance-monitoring?orgId=1&from=1603974504332&to=1603984150359&var-server=app-0:8067&var-server=app-1:8067

esethna:
@amyblais performance has been approved, we should move forward with this for v5.29. @reflog and @jespino are getting it ready for merge

amyblais:
@reflog Thanks, please remember to cherry pick https://github.com/mattermost/mattermost-server/pull/15878 and https://github.com/mattermost/mattermost-server/pull/15571 too

store/sqlstore/thread_store.go
```diff
@@ -188,3 +188,38 @@ func (s *SqlThreadStore) CreateMembershipIfNeeded(userId, postId string) error {
...
+			},
+		}).
+		ToSql()
+	if _, err := s.GetReplica().Select(&changedThreads, query, args...); err != nil {
+		return nil, errors.Wrap(err, "failed to fetch threads")
+	}
+	return changedThreads, nil
+}
+
+func (s *SqlThreadStore) UpdateUnreadsByChannel(userId string, changedThreads []string, timestamp int64) error {
```
jespino, line 213: We can improve a bit the performance here cheking if the changedThreads list is empty and not even trying to execute the query :)

store/sqlstore/upgrade.go
```diff
@@ -926,6 +926,11 @@ func upgradeDatabaseToVersion529(sqlStore SqlStore) {
 	sqlStore.AlterColumnTypeIfExists("SidebarChannels", "CategoryId", "VARCHAR(128)", "VARCHAR(128)")
 	sqlStore.AlterColumnDefaultIfExists("SidebarChannels", "CategoryId", model.NewString(""), nil)
 
+	sqlStore.CreateColumnIfNotExistsNoDefault("Threads", "ChannelId", "VARCHAR(26)", "VARCHAR(26)")
+	if _, err := sqlStore.GetMaster().Exec("UPDATE Threads SET ChannelId=PI.ChannelId FROM (SELECT ChannelId, Id from Posts) as PI WHERE PI.Id = PostId"); err != nil {
```
jespino, line 930: This migration query is executing a full scan of the Posts table, probably we can make it a bit better (that was my raw query), let me try to improve the performance there a bit.

jespino (approved):
Looks good to me

esethna:
@jespino @reflog is this ready to merge?

reflog:
@esethna no, waiting for last review from @jespino

jespino (approved):
LGTM

jespino:
@amyblais @esethna yes, this is ready now :)

reflog:
@amyblais @esethna cherry pick to 5.29 is blocked because https://github.com/mattermost/mattermost-server/tree/release-5.29 is broken (bad previous merge).
I've contacted @hmhealey to fix

PR #15995, opened 2020-10-18
Optimize images

JeffResc:
#### Summary

This pull request optimizes the images found in the repository using lossless compression. This means that the image quality will not be degraded. This will mean less space taken up by images along with faster loading times when viewing them. The overall file size has been reduced by around 8%.

#### Ticket Link

N/A

```release-note
Optimized images to reduce file size.
```


wiggin77 (changes requested):
@JeffResc, thanks for your interest in contributing to Mattermost.  Several of the images being shrunk are in the `vendor` directory, meaning they would need to be committed to their respective repositories.  Otherwise the changes you made will be undone the next time we update those libraries.  

The rest of the images are in the `test` directory and have no bearing on runtime performance other than during testing. That said I have no objection to approving these once the first issue is resolved.

JeffResc:
@wiggin77 I went ahead and restored the original vendor directory. Please let me know if you have any other changes.

wiggin77:
@JeffResc, it turns out these shrunken test files has caused a bunch of unit tests to fail.  The files are used to test file upload code and are byte compared with a remote copy of the files.  Since these are just for testing and it will be extra work to get the other set in sync, I don't think shrinking these is worthwhile.

jasonblais:
Thank you @JeffResc for your contributions here, along with https://github.com/mattermost/desktop/pull/1392 and https://github.com/mattermost/mattermost-server/pull/15995.

As noted in the docs PR, there are a couple of complexities we found after further review:
- The changes are undone every time the images are edited. A more efficient approach would be to include image optimization as part of our build process. Something we may look at long term
- We don't unfortunately have a good way to evaluate the changes, other than reviewing the images one by one

As a result, we'll close this contribution at this time. 

For pull requests greater than 20 lines of code, a Help Wanted ticket is recommended to increase the likelihood of it getting merged into the core product. More on the process can be found here: https://developers.mattermost.com/contribute/getting-started/contributions-without-ticket/

If you have any questions or suggestions about the contribution process, let us know! We are always looking to improve.

Jason

PR #16007, opened 2020-10-19
Skipping some telemetry tests on production builds

jespino:
#### Summary
Some telemetry tests were failing with the final builds because the RUDDER_KEY
and RUDDER_DATAPLANE_URL weren't the placeholders ones. This changes allow the
test to execute properly in that cases by skipping the tests here.


metanerd (approved):
Checked and works on my side of things also. Thank you so much!

PR #16010, opened 2020-10-19
MM-29812: Fix a panic in auth code if GetChannel returns an error.

grundleborg:
#### Summary
Fix a panic in authorization code if `GetChannel()` returns an error.

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-29812

#### Release Note

```release-note
NONE
```

####
QA Note: this is a specific failure that can't be triggered without the database being somehow broken, so the exact bug can't be manually triggered for testing. However, the unit test in this PR simulates that failure and covers testing it. Manual QA should be focused on just a smoke test of channel-level permissions to make sure there are no unexpected side-effects.


app/authorization.go
```diff
@@ -80,7 +80,7 @@ func (a *App) SessionHasPermissionToChannel(session model.Session, channelId str
 		return true
```
grundleborg, line 80: @mkraft now you see the context of my DM earlier :)

I left this where it is, but I was wondering about it because it feels like the flow of this function could be cleaned up a bit. However, I note moving this condition earlier in the function *would* have a behaviour change - meaning a "not found" on the channel would return `true` rather than `false` for unrestricted sessions - so I thought it would be better to leave any such change out of this specific fix.

jespino (approved):
Looks good to me.

mkraft (approved):

lindalumitchell (approved):

PR #16030, opened 2020-10-19
MM-29067 Add deterministic IDs for default sidebar categories

hmhealey:
Since we can't add a unique constraint on sidebar category type to prevent duplicate default categories from being created (possible if migration is triggered to run twice at the same time), we'll instead prevent duplicate default categories by giving them each a deterministic ID which will cause a collision if any race occurs.

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-29067

#### Release Note
```release-note
Fixed case where duplicate sidebar categories could be created on first use of experimental sidebar
```


model/channel_sidebar.go
```diff
@@ -109,3 +110,13 @@ func (o OrderedSidebarCategories) ToJson() []byte {
...
 }
+
+func IsValidCategoryId(s string) bool {
+	// Category IDs can either be regular IDs
+	if IsValidId(s) {
+		return true
+	}
+
+	// Or default categories can follow the pattern {type}_{userID}_{teamID}
+	return regexp.MustCompile("(favorites|channels|direct_messages)_[a-z0-9]{26}_[a-z0-9]{26}").MatchString(s)
```
hmhealey, line 121: I tried to avoid a regex here, but the extra underscore in direct_messages made it hard to split the string on an underscore and check each part individually.

store/sqlstore/upgrade.go
```diff
@@ -915,3 +917,13 @@ func precheckMigrationToVersion528(sqlStore SqlStore) error {
 
 	return nil
 }
+
+func upgradeDatabaseToVersion529(sqlStore SqlStore) {
+	// if shouldPerformUpgrade(sqlStore, VERSION_5_28_0, VERSION_5_29_0) {
+
+	sqlStore.AlterColumnTypeIfExists("SidebarCategories", "Id", "VARCHAR(255)", "VARCHAR(128)")
```
hmhealey, line 924: The statements may not be necessary since these actually match the types of the fields in my local DB, but I figure it's good to confirm their types in case some DB versions behave differently.

amyblais:
@ethervoid Kind reminder to review for v5.29. 
@hmhealey Agniva is off this week; not sure if the second dev reviewer could be re-assigned.

ethervoid:
@amyblais Given my lack of knowledge and context on this part of the server code and that we're fully engaged with the release of GA it's going to be hard to review it properly and given that this PR deserves to invest time on it to understand the changes properly I suggest to change me with another reviewer. I suggest a someone from core feature given that AFAIK they're in charge of this feature. Sorry :(

api4/channel.go
```diff
@@ -40,9 +40,9 @@ func (api *API) InitChannel() {
 	api.BaseRoutes.ChannelCategories.Handle("", api.ApiSessionRequired(updateCategoriesForTeamForUser)).Methods("PUT")
 	api.BaseRoutes.ChannelCategories.Handle("/order", api.ApiSessionRequired(getCategoryOrderForTeamForUser)).Methods("GET")
 	api.BaseRoutes.ChannelCategories.Handle("/order", api.ApiSessionRequired(updateCategoryOrderForTeamForUser)).Methods("PUT")
-	api.BaseRoutes.ChannelCategories.Handle("/{category_id:[A-Za-z0-9]+}", api.ApiSessionRequired(getCategoryForTeamForUser)).Methods("GET")
-	api.BaseRoutes.ChannelCategories.Handle("/{category_id:[A-Za-z0-9]+}", api.ApiSessionRequired(updateCategoryForTeamForUser)).Methods("PUT")
-	api.BaseRoutes.ChannelCategories.Handle("/{category_id:[A-Za-z0-9]+}", api.ApiSessionRequired(deleteCategoryForTeamForUser)).Methods("DELETE")
+	api.BaseRoutes.ChannelCategories.Handle("/{category_id:[A-Za-z0-9_-]+}", api.ApiSessionRequired(getCategoryForTeamForUser)).Methods("GET")
```
reflog, line 43: this change will break OpenAPI spec validator in `mattermost-govet`. See https://github.com/mattermost/mattermost-govet/blob/master/openApiSync/openApiSync.go#L61

reflog (approved):
looks good, just one comment related to mattermost-govet

api4/channel.go
hmhealey: Strange that it didn't break it already, but maybe something else handles it? The get user by username route does something similar since those can include more punctuation

amyblais:
@grundleborg Kind reminder to review for v5.29.

grundleborg:
I'll get to this on Monday (sorry got sucked into Customer issues the last 2 days so no time). feel free to reassign if it's needed urgently.

model/channel_sidebar.go
agnivade, line 121: We should avoid creating a new instance of `regexp.MustCompile` every time. They should be created once. Creating a new instance every time has a performance cost which should be avoided.

store/sqlstore/upgrade.go
```diff
@@ -915,3 +917,15 @@ func precheckMigrationToVersion528(sqlStore SqlStore) error {
 
 	return nil
 }
+
+func upgradeDatabaseToVersion529(sqlStore SqlStore) {
+	// if shouldPerformUpgrade(sqlStore, VERSION_5_28_0, VERSION_5_29_0) {
+
+	sqlStore.AlterColumnTypeIfExists("SidebarCategories", "Id", "VARCHAR(128)", "VARCHAR(128)")
+	sqlStore.AlterColumnDefaultIfExists("SidebarCategories", "Id", model.NewString(""), nil)
```
grundleborg, line 925: Why is the default different between Postgres and MySQL?

store/sqlstore/upgrade.go
```diff
@@ -915,3 +917,15 @@ func precheckMigrationToVersion528(sqlStore SqlStore) error {
...
+	sqlStore.AlterColumnTypeIfExists("SidebarChannels", "CategoryId", "VARCHAR(128)", "VARCHAR(128)")
+	sqlStore.AlterColumnDefaultIfExists("SidebarChannels", "CategoryId", model.NewString(""), nil)
```
grundleborg, line 927: Ditto

grundleborg (changes requested):
Just a couple of things, alongside Agniva's point about the regex. Otherwise looks fine to me.

store/sqlstore/upgrade.go
hmhealey: I'm honestly not sure why the schema is different with and without the upgrade. If I use a fresh database, I get
```
mysql> show columns in SidebarCategories;
+-------------+--------------+------+-----+---------+-------+
| Field       | Type         | Null | Key | Default | Extra |
+-------------+--------------+------+-----+---------+-------+
| Id          | varchar(128) | NO   | PRI | NULL    |       |
```
and if I run an upgrade without that line, I still get
```
mysql> show columns in SidebarCategories;
+-------------+--------------+------+-----+---------+-------+
| Field       | Type         | Null | Key | Default | Extra |
+-------------+--------------+------+-----+---------+-------+
| Id          | varchar(128) | NO   | PRI | NULL    |       |
```
but for some reason, without that line, the `test-schema` job failed for MySQL with 
```
Schema mismatch
547c547
<   `Id` varchar(128) NOT NULL DEFAULT '',
---
>   `Id` varchar(128) NOT NULL,
562c562
<   `CategoryId` varchar(128) NOT NULL DEFAULT '',
---
>   `CategoryId` varchar(128) NOT NULL,

Exited with code exit status 1
```

Since `AlterColumnDefaultIfExists` removes any default value when passed the empty string, it seems like altering the column type might be setting a default value on MySQL somehow

hmhealey: Actually, I was trying to test again locally, and it seems like `make test-db-migration` doesn't work since it's not even giving me an error if I remove the migration entirely.

hmhealey: I still haven't been able to come up with a better explanation about why the default seems to sometimes be set, but this at least ensures the column always correctly has no default even if this query might be redundant.

agnivade: Should we also make the postgres default to be empty string in that case?

hmhealey: Both DBs are actually going to have no default after this. Passing the empty string to `AlterColumnDefaultIfExists` makes it clear the default ([link](https://github.com/mattermost/mattermost-server/blob/master/store/sqlstore/supplier.go#L783)).

agnivade: Ugh sorry, I read it wrong :facepalm:

agnivade (approved):

store/sqlstore/upgrade.go
hmhealey: No worries. It's kind of confusing anyway since it prevents us from actually making a column have the empty string as default. Now that I think of it, that special case might have actually caused me problems in the past.

grundleborg (approved):
LGTM

hmhealey:
@jgilliam17 Reproducing the problem case for this PR is really difficult since it's a race condition that only happens for users that were created before the new sidebar was turned on when they first use the new sidebar. Perhaps to test this, you could create a couple users, enable the new sidebar, and then ensure the default categories are created successfully for them as before?

jgilliam17 (approved):
Thanks @hmhealey 
Tested, looks good to merge.
- Verified default new sidebar categories are created successfully for new users that joined the server before new sidebar was enabled by sysadmin.

amyblais:
@hmhealey Please cherry pick manually.

PR #16053, opened 2020-10-21
[MM-28363] User Limit Overage Warning Emails

nickmisasi:
#### Summary
In Mattermost Cloud we want to inform system administrators by email when they are at, and over the user limit on their installation. This PR includes a new template, as well as a function to send the email based on the subscription, license, user count, etc.

The button in the email directs the user to the Subscription section of the System Console. 

All system admins will be sent an email in their preferred locale for the following cases:
1. The installation is at 9 users, and a user joins (AT limit)
2. The installation is at 10 users, and a user joins (OVER limit)
3. The installation is at 9 active users, and an inactive user is activated (AT limit)
4. The installation is at 10 active users, and an inactive user is activated (OVER limit)

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-28363
Assets in the following:  https://github.com/mattermost/mattermost-webapp/pull/6874
#### Release Note

```release-note
All system admins will be sent an email in their preferred locale for the following cases (assuming an unpaid, cloud licensed installation):
1. The installation is at 9 users, and a user joins (AT limit)
2. The installation is at 10 users, and a user joins (OVER limit)
3. The installation is at 9 active users, and an inactive user is activated (AT limit)
4. The installation is at 10 active users, and an inactive user is activated (OVER limit)

```

### DIFF
--- a/app/cloud.go
+++ b/app/cloud.go
@@ -1,0 +1,59 @@
+// Copyright (c) 2015-present Mattermost, Inc. All Rights Reserved.
+// See LICENSE.txt for license information.
+
+package app
+
+import (
+	"github.com/mattermost/mattermost-server/v5/mlog"
+	"github.com/mattermost/mattermost-server/v5/model"
+)
+
+func (a *App) CheckAndSendUserLimitWarningEmails() *model.AppError {
+	if a.Srv().License() != nil && !*a.Srv().License().Features.Cloud {
+		// Not cloud instance, do nothing
+		mlog.Error("NOT CLOUD")
+		return nil
+	}
+
+	// subscription, subErr := a.Cloud().GetSubscription()
+	// if subErr != nil {
+	// 	return subErr
+	// }
+
+	// if subscription != nil && subscription.IsPaidTier == "true" {
+	// 	// Paid subscription, do nothing
+	// 	return nil
+	// }
+
+	cloudUserLimit := *a.Config().ExperimentalSettings.CloudUserLimit
+	systemUserCount, _ := a.Srv().Store.User().Count(model.UserCountOptions{})
+	remainingUsers := cloudUserLimit - systemUserCount
+
+	if remainingUsers > 0 {
+		return nil
+	}
+	userOptions := &model.UserGetOptions{
+		Page:     0,
+		PerPage:  100,
+		Role:     model.SYSTEM_ADMIN_ROLE_ID,
+		Inactive: false,
+	}
+	sysAdmins, err := a.GetUsers(userOptions)
+	if err != nil {
+		return err
+	}
+
+	if remainingUsers <= 0 {
+		// Over limit
+		for admin := range sysAdmins {
+			a.Srv().EmailService.SendOverUserLimitWarningEmail(sysAdmins[admin].Email, "en")
+		}
+
+	} else if remainingUsers == 0 {
+		// AT limit
+		for admin := range sysAdmins {
+			a.Srv().EmailService.SendAtUserLimitWarningEmail(sysAdmins[admin].Email, "en")
+		}
+	}
+	return nil
+}
--- a/app/cloud.go
+++ b/app/cloud.go
@@ -41,6 +41,7 @@ func (a *App) CheckAndSendUserLimitWarningEmails() *model.AppError {
 		return err
 	}
 
+	// -1 means they are 1 user over the limit - we only want to send the email for the 11th user
 	if remainingUsers == -1 {
 		// Over limit by 1 user
 		for admin := range sysAdmins {
--- a/app/cloud.go
+++ b/app/cloud.go
@@ -8,7 +8,7 @@ import (
 )
 
 func (a *App) CheckAndSendUserLimitWarningEmails() *model.AppError {
-	if a.Srv().License() != nil && !*a.Srv().License().Features.Cloud {
+	if a.Srv().License() == nil || (a.Srv().License() != nil && !*a.Srv().License().Features.Cloud) {
 		// Not cloud instance, do nothing
 		return nil
 	}
--- a/app/cloud.go
+++ b/app/cloud.go
@@ -13,15 +13,15 @@ func (a *App) CheckAndSendUserLimitWarningEmails() *model.AppError {
 		return nil
 	}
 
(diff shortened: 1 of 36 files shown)


nickmisasi:
Coordinating QA/design review for this in DM.

nickmisasi:
@mm-cloud-bot pls spin up my test server properly

app/cloud.go
```diff
@@ -0,0 +1,56 @@
...
+
+	// subscription, subErr := a.Cloud().GetSubscription()
+	// if subErr != nil {
+	// 	return subErr
+	// }
+
+	// if subscription != nil && subscription.IsPaidTier == "true" {
+	// 	// Paid subscription, do nothing
+	// 	return nil
+	// }
```
nickmisasi, line 24: This is commented out so we can test and get design review in a spinwick. Will uncomment before merging.

matthewbirtch (changes requested):
Hey @nickmisasi everything looks great. Just a few very minor things:
- Could we export the images at 2x size? They look a bit fuzzy on high-res screens. 
- Spacing at the bottom 'Questions' section doesn't quite look right
- Also looks like the font and size looks wrong for this callout. Should be using arial, but appears to be using roboto

**Designed:**
image

**Implemented:**
image

app/cloud.go
```diff
@@ -0,0 +1,56 @@
...
+
+	cloudUserLimit := *a.Config().ExperimentalSettings.CloudUserLimit
+	systemUserCount, _ := a.Srv().Store.User().Count(model.UserCountOptions{})
```
ethervoid, line 27: I think this is not going to filter verified from non-verified. If we're sending the email only when the verification is done then we should filter by that

app/cloud.go
```diff
@@ -0,0 +1,56 @@
...
+		PerPage:  100,
+		Role:     model.SYSTEM_ADMIN_ROLE_ID,
+		Inactive: false,
+	}
+	sysAdmins, err := a.GetUsers(userOptions)
+	if err != nil {
+		return err
+	}
+
+	if remainingUsers == -1 {
```
ethervoid, line 44: I'd put do
```suggestion
	if remainingUsers < 0 {
```
because in case we have some kind of race condition, which I don't know if can happen, and the number of users is -2 then we're not going to send the email

app/email.go
```diff
@@ -589,3 +589,47 @@ func (es *EmailService) CreateVerifyEmailToken(userId string, newEmail string) (
 
 	return token, nil
 }
+
+func (es *EmailService) SendAtUserLimitWarningEmail(email string, locale string, siteURL string) (bool, *model.AppError) {
```
ethervoid, line 593: Given that is the same template but with different data we can probably we can reuse most of the logic, create a `sendUserLimitEmail` and then call it from these two methods with the proper data. What do you think?

ethervoid (commented):
Some comments

stevemudie (approved):
Worked as expected. Looks good!

app/cloud.go
nickmisasi: This check for -1 is so that the email is only sent one time, and not for every single person added over the limit. I _could_ change this to < 0 but then I'd need to store something somewhere saying "Email has already been sent". I figured that might have been a little overkill for this

ethervoid: Ok, please comment the intent so other people knows the reason of it :)

app/email.go
nickmisasi: I've done it this way for a few reasons:
1. I think it's more readable when the code calls `SendAtUserLimitWarningEmail` instead of deriving which type of email it is from the arguments
2. It seems like this sort of pattern is followed elsewhere in the file, see `sendChangeUsernameEmail` and `sendEmailChangeEmail` which both use the same template
3. We'd need 6-7 function arguments, and I'm not a fan of huge definitions like that

I'm swaying to keeping it as is - if you really want it changed, let me know!

app/cloud.go
nickmisasi: Will do!

app/cloud.go
nickmisasi: I ran a test for this to be sure:
Put a log after this line to log the systemUserCount
User count was 10 when I started the test
Invited 5 people via email (and didn't verify their emails)
Invited a 6th person, and verified their email
The log for the verified email stated that the user count was 11 - so it seems to add up. 


There is an edge case here. User's created via an invite link (rather than an email) are created in the database _before_ their email is verified (whereas email invites are created _after_ their email is verified). The difference is, that an unverified email user created via link _does show in the Mattermost UI_ whereas the email invited one won't until its been both verified and created.
the third user in this image hasn't actually verified its email 
image
I think we'll need to discuss when we consider a user to actually be a user, because it will affect a sysadmins decision to deactivate/activate certain users.

nickmisasi: @marianunez What are your thoughts on this?

app/email.go
ethervoid: No hard opinion here 2/5 so if you think that's ok, works for me

app/cloud.go
nickmisasi: @ethervoid You can see discussion here - desired behaviour is in this branch! https://community.mattermost.com/private-core/pl/mm76dqizybrnujnboyj89tqpxh

ethervoid: Cool! Good to have it cleared

ethervoid (approved):
LGTM!

marianunez (approved):
Looks great! Thanks @nickmisasi!

matthewbirtch (approved):

app/cloud.go
```diff
@@ -0,0 +1,56 @@
...
+		// Not cloud instance, do nothing
+		return nil
+	}
+
+	// subscription, subErr := a.Cloud().GetSubscription()
+	// if subErr != nil {
+	// 	return subErr
+	// }
+
+	// if subscription != nil && subscription.IsPaidTier == "true" {
```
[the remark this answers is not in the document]
nickmisasi: Stripe returns the metadata as strings, I chose to not mess with the type of the return value throughout to prevent panics. Plus, the values for the configs we're checking in the front end are string true also - `license.Cloud === 'true'` etc. It could have gone either way, really.

PR #16055, opened 2020-10-22
Add golang migrate migrations

jespino:
#### Summary
Add golang migrate migrations proof of concept


store/sqlstore/migrate.go
```diff
@@ -0,0 +1,54 @@
...
+
+	"github.com/golang-migrate/migrate/v4"
+	"github.com/golang-migrate/migrate/v4/database"
+	"github.com/golang-migrate/migrate/v4/database/mysql"
+	"github.com/golang-migrate/migrate/v4/database/postgres"
+	_ "github.com/golang-migrate/migrate/v4/source/file"
+	bindata "github.com/golang-migrate/migrate/v4/source/go_bindata"
+	"github.com/mattermost/mattermost-server/v5/model"
+	mysqlmigrations "github.com/mattermost/mattermost-server/v5/store/sqlstore/migrations/mysql"
+	pgmigrations "github.com/mattermost/mattermost-server/v5/store/sqlstore/migrations/postgres"
```
agnivade, line 14: Make these 3 into a separate import group in the 2nd position.

store/sqlstore/migrate.go
```diff
@@ -0,0 +1,54 @@
...
+)
+
+func (ss *SqlSupplier) Migrate() error {
```
agnivade, line 17: A godoc here would be nice.

agnivade (changes requested):
This is more or less exactly what is done for mattermod which is okay. I'll also request a CLI command to run backward migrations as we did here: https://github.com/mattermost/mattermost-mattermod/blob/master/cmd/migrator/main.go. Without that it's not possible to leverage the down files.

But most importantly, without the actual sql files, this does not kick in yet. We need to discuss and decide on a minimum version before which we will not support upgrading to this version. Then convey that decision to PM and release management. And then finally write the sql migration code to actually run the migrations all the way from that version to latest.

That's the biggest thing.

jespino:
@agnivade I applied the proposed changes.

agnivade:
Nice. POC looks good.

grundleborg (commented):
@jespino this looks great to me. What are the next steps, an which of them are you willing to take vs. needing the Server Team to handle?

jespino:
Probably, the next step is to decide if this is what we want to move forward, see the missing pieces (commands that cloud team needs, like the db upgrade/db downgrade), and schedule a date to start using it instead of the previous existing approach. I would say, validation of the approach during the next server meeting. If nothing changes about the approach, scheduling the merge date and properly announce it (to not collide with anything done for the next version). Once is merge, the Cloud team should be able to add the commands that they need (they can start working before branching from migrations system PR).

jespino:
//cc @grundleborg @ethervoid :point_up:

store/sqlstore/upgrade.go
```diff
@@ -195,6 +195,13 @@ func upgradeDatabase(sqlStore SqlStore, currentModelVersionString string) error
 	upgradeDatabaseToVersion529(sqlStore)
 	upgradeDatabaseToVersion530(sqlStore)
 
+	err = sqlStore.Migrate()
```
ethervoid, line 198: Maybe we can check if the system is using the new versioning so that means we can skip the old migrations. What do you think?

ethervoid (approved):
PoC looks good to me

jespino:
Done by other PR :tada:

PR #16058, opened 2020-10-22
[MM-29937] Don't allow email and contact name to be updated from the MM app

devinbinnie:
#### Summary
When updating the Company Information, the stripe implementation expected that unchanged fields would be `nil`, whereas the MM app would send up blank strings, thus resetting those fields.

This PR stops the fields that aren't modified by the MM app from being added to the struct at all, so that they can't be wiped out.

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-29937

#### Release Note
```release-note
NONE
```

### DIFF
--- a/model/cloud.go
+++ b/model/cloud.go
@@ -45,9 +45,9 @@ type CloudCustomer struct {
 // CloudCustomerInfo represents editable info of a customer.
 type CloudCustomerInfo struct {
 	Name             string `json:"name"`
-	Email            string `json:"email"`
-	ContactFirstName string `json:"contact_first_name"`
-	ContactLastName  string `json:"contact_last_name"`
+	Email            string `json:"-"`
+	ContactFirstName string `json:"-"`
+	ContactLastName  string `json:"-"`
 	NumEmployees     int    `json:"num_employees"`
 }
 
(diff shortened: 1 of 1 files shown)


model/cloud.go
```diff
@@ -45,9 +45,9 @@ type CloudCustomer struct {
 // CloudCustomerInfo represents editable info of a customer.
 type CloudCustomerInfo struct {
 	Name             string `json:"name"`
-	Email            string `json:"email"`
-	ContactFirstName string `json:"contact_first_name"`
-	ContactLastName  string `json:"contact_last_name"`
+	Email            string `json:"-"`
```
marianunez, line 48: Would `omitempty` work here?

```suggestion
	Email            string `json:"email,omitempty"`
```

nickmisasi (approved):
I would say use `omitempty` like Maria mentioned - otherwise, LGTM.

marianunez (approved):
@devinbinnie we need to make sure to validate with test Stripe. We should include the case of updating without email/contact here https://github.com/mattermost/customer-web-server/blob/255efafac8fa2aa2f1e4cb04d1a92b0ee5f1fb69/payment/stripe_test.go#L33

devinbinnie:
> @devinbinnie we need to make sure to validate with test Stripe. We should include the case of updating without email/contact here https://github.com/mattermost/customer-web-server/blob/255efafac8fa2aa2f1e4cb04d1a92b0ee5f1fb69/payment/stripe_test.go#L33

https://github.com/mattermost/customer-web-server/pull/261

PR #16067, opened 2020-10-24
[MM-8497] Ability to set Do Not Disturb for a specified period of time

darkLord19:


darkLord19:
This is a draft PR. Created to drive discussion.

jfrerich:
Thnaks for the PR, @darkLord19.  Let me get those that are interested in your PR involved.

@esethna, looking at the ticket from which this originated, I found you had edited the ticket description.  Is this something you could help drive forward?  Thanks!

esethna:
Thanks @darkLord19! I'll add @jespino to help with an initial review and guidance

agnivade:
Please ping me when the PR is ready for review.

darkLord19:
@agnivade it's ready for initial review. I have tried to address most of the concerns we had about performance and implementation so can you please go through it and provide your comments.

api4/status.go
```diff
@@ -99,7 +99,11 @@ func updateUserStatus(c *Context, w http.ResponseWriter, r *http.Request) {
 	case "away":
 		c.App.SetStatusAwayIfNeeded(c.Params.UserId, true)
 	case "dnd":
-		c.App.SetStatusDoNotDisturb(c.Params.UserId)
+		if c.Params.DNDEndTime != "" {
+			c.App.SetStatusDoNotDisturbTimed(c.Params.UserId, c.Params.DNDEndTime)
+		} else {
+			c.App.SetStatusDoNotDisturb(c.Params.UserId)
+		}
```
agnivade, line 106: I don't think there is a need for a separate function for this. We can pass the `c.Params.DNDEndTime` in `c.App.SetStatusDoNotDisturb` itself and do the conditional operation there.

app/app.go
```diff
@@ -99,6 +99,9 @@ func (a *App) InitServer() {
 			})
 		}
 		a.srv.RunJobs()
+		if a.IsLeader() {
+			model.CreateRecurringTask("Unset DND Statuses", a.UpdateDNDStatusOfUsers, time.Duration(300)*time.Second)
```
agnivade, line 103: This uses the existing recurring task and not our modified task with an offset at a specific time. Did you plan to do it later?

app/status.go
```diff
@@ -281,6 +304,38 @@ func (a *App) SetStatusAwayIfNeeded(userId string, manual bool) {
...
+	if aErr != nil {
+		mlog.Error("Failed to fetch user's timezone", mlog.String("user_id", userId), mlog.String("err", aErr.Error()))
+	}
+
+	utc, cErr := convertToUTC(tz, endtime)
+
+	if cErr != nil {
+		mlog.Error("Failed to conver user timezone to utc", mlog.String("user_id", userId), mlog.String("err", cErr.Error()))
+		return
+	}
```
agnivade, line 332: I thought we agreed to send the time in UTC from the client itself? There is no need to do a conversion at the server side.

store/sqlstore/upgrade.go
```diff
@@ -865,6 +867,14 @@ func upgradeDatabaseToVersion5281(sqlStore SqlStore) {
 	}
 }
 
+func upgradeDatabaseToVersion5282(sqlStore SqlStore) {
```
agnivade, line 870: This should be in version 5.30. Not 5.28.2.

app/status.go
```diff
@@ -363,3 +418,32 @@ func (a *App) GetStatus(userId string) (*model.Status, *model.AppError) {
 func (a *App) IsUserAway(lastActivityAt int64) bool {
 	return model.GetMillis()-lastActivityAt >= *a.Config().TeamSettings.UserStatusAwayTimeout*1000
 }
+
+// UpdateDNDStatusOfUsers is a recurring task which is started when server starts
+// which unsets dnd status of users if needed and saves and broadcasts it
+func (a *App) UpdateDNDStatusOfUsers() {
+	statuses, err := a.Srv().Store.Status().GetExpiredDNDStatuses()
```
agnivade, line 425: We can do both the get and update operations in a single DB query. We can select the rows we want, and update the columns to the new values. status will be prevstatus, prevstatus will be dnd, manual to false, and endtimeunix to -1. And then return the affected rows, and broadcast status.

1 query instead of 2. More optimized :)

Plus this also has a transactional guarantee. Ideally, you need to run both queries inside a transaction.

store/sqlstore/status_store.go
```diff
@@ -37,6 +39,39 @@ func (s SqlStatusStore) createIndexesIfNotExists() {
...
+		status.Manual,
+		status.LastActivityAt,
+		status.ActiveChannel,
+		status.DNDEndTimeUnix,
+		status.PrevStatus,
+	}
+}
+
+func (s SqlStatusStore) SaveMultiple(statuses []*model.Status) ([]*model.Status, error) {
+	builder := s.getQueryBuilder().Insert("Status").Columns(statusSliceColumns()...)
```
agnivade, line 59: I think this wrong. We should update, not insert. Anyways, we don't need this. See my comment above.

agnivade (commented):
First pass. Good start. We need to improve some things.

api4/status.go
darkLord19: that will require changing signature of SetStatusDoNotDisturb which will break backward compatibility.

app/app.go
darkLord19: Yes, I am planning to do it once other implementation details are fleshed out.

app/status.go
darkLord19: I thought it'd be better at server side as it'll be easier to test and verify and should be faster than js.

store/sqlstore/upgrade.go
darkLord19: Yes, this will be updated to reflect the appropriate version needed at the time of merge

api4/status.go
agnivade: The app layer is our internal code. It should have been an internal package anyways. It's fine to break backward compatibility.

app/status.go
agnivade: At server side, it requires an extra DB call. And bear in mind that the server is a common thing for all users, whereas a user's browser is just one. I think it's cleaner to get the right data from the API itself. So the server code looks cleaner and simpler.

api4/status_test.go
```diff
@@ -42,6 +43,30 @@ func TestGetUserStatus(t *testing.T) {
...
+		task := model.CreateRecurringTask("Unset DND Statuses", th.App.UpdateDNDStatusOfUsers, time.Duration(30)*time.Second)
+		th.App.SetStatusOnline(th.BasicUser.Id, true)
+		userStatus, resp := Client.GetUserStatus(th.BasicUser.Id, "")
+		CheckNoError(t, resp)
+		assert.Equal(t, "online", userStatus.Status)
+		th.App.SetStatusDoNotDisturbTimed(th.BasicUser.Id, time.Now().Add(1*time.Minute).Format(time.RFC3339))
+		userStatus, resp = Client.GetUserStatus(th.BasicUser.Id, "")
+		CheckNoError(t, resp)
+		assert.Equal(t, "dnd", userStatus.Status)
+		time.Sleep(90 * time.Second)
```
jespino, line 63: This makes the test 90 seconds longer, what is a lot, we need something less time consuming here. I would try to do it in 1 second (2 seconds max) and check instantaneusly and then check again after the second passes.

app/plugin_api.go
```diff
@@ -311,6 +311,21 @@ func (api *PluginAPI) UpdateUserStatus(userId, status string) (*model.Status, *m
 	return api.app.GetStatus(userId)
 }
 
+func (api *PluginAPI) UpdateUserStatusV2(userId, status, endtime string) (*model.Status, *model.AppError) {
```
jespino, line 314: V2 means nothing, I would use something like `UpdateUserStatusWithDndTimeout` is a bit long, but at least describe why is there.

app/plugin_api.go
```diff
@@ -311,6 +311,21 @@ func (api *PluginAPI) UpdateUserStatus(userId, status string) (*model.Status, *m
...
+	case model.STATUS_OFFLINE:
+	case model.STATUS_AWAY:
+		api.UpdateUserStatus(userId, status)
+	case model.STATUS_DND:
+		api.app.SetStatusDoNotDisturbTimed(userId, endtime)
+	default:
+		return nil, model.NewAppError("UpdateUserStatusV2", "plugin.api.update_user_status.bad_status", nil, "unrecognized status", http.StatusBadRequest)
+	}
+
+	return api.app.GetStatus(userId)
```
jespino, line 326: We can simplify it a lot with something like:
```suggestion
	if status = model.STATUS_DND {
		api.app.SetStatusDoNotDisturbTimed(userId, endtime)
		return api.app.GetStatus(userId)
	}
	return api.UpdateUserStatus(userId, status)
```

app/status.go
```diff
@@ -281,6 +282,36 @@ func (a *App) SetStatusAwayIfNeeded(userId string, manual bool) {
...
+	status.Status = model.STATUS_DND
+	status.Manual = true
+
+	// take a string in RFC3339 format and return time
+	utc, cErr := time.Parse(time.RFC3339, endtime)
+
+	if cErr != nil {
+		mlog.Error("Failed to parse endtime", mlog.String("user_id", userId), mlog.String("err", cErr.Error()))
+		return
+	}
```
jespino, line 308: This validation sould happen even before of doing the `GetStatus` that way, any wrong request doesn't need to execute the DB query.

store/sqlstore/status_store.go
```diff
@@ -99,6 +101,38 @@ func (s SqlStatusStore) GetByIds(userIds []string) ([]*model.Status, error) {
...
+			Status
+		SET
+			DNDEndTimeUnix = %v, Manual = %v,
+			PrevStatus = '%v', Status = PrevStatus
+		WHERE
+			(Status = '%v' AND DNDEndTimeUnix <= %v AND DNDEndTimeUnix > %v)
+		RETURNING
+			UserId, Status`,
+		-1, false, model.STATUS_DND, model.STATUS_DND, time.Now().UTC().Unix(), 1,
+	)
```
jespino, line 116: Do no use string building here, it is dangeorus and can lead to SQL injection. We prefer to use squirrel library to build the sql queries.

store/sqlstore/status_store.go
```diff
@@ -99,6 +101,38 @@ func (s SqlStatusStore) GetByIds(userIds []string) ([]*model.Status, error) {
...
 
+func (s SqlStatusStore) UpdateExpiredDNDStatuses() ([]*model.Status, error) {
+	queryString := fmt.Sprintf(`
+		UPDATE
+			Status
+		SET
+			DNDEndTimeUnix = %v, Manual = %v,
+			PrevStatus = '%v', Status = PrevStatus
+		WHERE
+			(Status = '%v' AND DNDEndTimeUnix <= %v AND DNDEndTimeUnix > %v)
```
jespino, line 112: We can experiment a bit here when we start load-testing this feature with the indexes. Creating a new index around DNDEndTimeUnix can speedup drastically this query, but at the same time can add a bit more load on the status writes. We can test it.

store/sqlstore/upgrade.go
```diff
@@ -865,6 +867,14 @@ func upgradeDatabaseToVersion5281(sqlStore SqlStore) {
 	}
 }
 
+func upgradeDatabaseToVersion530(sqlStore SqlStore) {
+	if shouldPerformUpgrade(sqlStore, VERSION_5_29_0, VERSION_5_30_0) {
```
jespino, line 871: You shouldn't add the version here, the and the `shouldPerformUpgrade` check is comment until the release is create, so... we add all the migrations for the version in the same block.

store/sqlstore/upgrade.go
```diff
@@ -865,6 +867,14 @@ func upgradeDatabaseToVersion5281(sqlStore SqlStore) {
...
+		sqlStore.CreateColumnIfNotExistsNoDefault("Status", "DNDEndTimeUnix", "BIGINT", "BIGINT")
+		sqlStore.CreateColumnIfNotExistsNoDefault("Status", "PrevStatus", "VARCHAR(32)", "VARCHAR(32)")
+		saveSchemaVersion(sqlStore, VERSION_5_30_0)
+	}
+}
+
```
jespino, line 877:
```suggestion
func upgradeDatabaseToVersion529(sqlStore SqlStore) {
	// if shouldPerformUpgrade(sqlStore, VERSION_5_28_1, VERSION_5_29_0) {
	sqlStore.CreateColumnIfNotExistsNoDefault("Status", "DNDEndTimeUnix", "BIGINT", "BIGINT")
	sqlStore.CreateColumnIfNotExistsNoDefault("Status", "PrevStatus", "VARCHAR(32)", "VARCHAR(32)")
		// saveSchemaVersion(sqlStore, VERSION_5_29_0)
	//}
}

```

jespino, line 877:
```suggestion
func upgradeDatabaseToVersion530(sqlStore SqlStore) {
	// if shouldPerformUpgrade(sqlStore, VERSION_5_29_0, VERSION_5_30_0) {
	sqlStore.CreateColumnIfNotExistsNoDefault("Status", "DNDEndTimeUnix", "BIGINT", "BIGINT")
	sqlStore.CreateColumnIfNotExistsNoDefault("Status", "PrevStatus", "VARCHAR(32)", "VARCHAR(32)")
		// saveSchemaVersion(sqlStore, VERSION_5_30_0)
	// }
}

```

jespino: And it should goes after the upgradeDatabaseToVersion529 function

jespino (changes requested):
Some changes needed, but in general is looking good

api4/status_test.go
darkLord19: Done

store/sqlstore/status_store.go
darkLord19: I tried using it but there wasn't a way to update and get updated rows in a same query using squirrel library as there is no method to add "RETURNING" statement in query. If we want to use squirrel then need to make two db calls. one to get the statuses which meet our criteria and then update them. we need to perform both operations in a transaction.

esethna:
@darkLord19 how's this going? Let us know if you need any help or guidance

darkLord19:
@esethna It's going good eric. Functionality is working now, need to address some review comments and fix tests. then will again request review.

esethna:
@darkLord19 awesome, looking forward to it!

jespino:
I going to be some days out, so I've unassigned myself from this review.

esethna:
@asaadmahmood @jespino @darkLord19 what are next steps needed to move this forward? How can we help?

darkLord19:
@esethna only custom date time picker functionality is remaining in webapp. Was busy last week so will work on it this weekend.

esethna:
Thanks @darkLord19, let us know when you're looking for another review

esethna:
@darkLord19 just to verify, dnd clear time and status clear time are two separate values, correct? For example, a user could be in Do Not Disturb for 30 minutes and **also** have a custom status set that would clear at a different selected time.

darkLord19:
@esethna yes I realised it afterwards so that's why I haven't pushed server side changes for custom statu PR. I'll revert the changes.

agnivade:
@darkLord19 - Looks like the tests are failing due to the mocks not being set correctly. Once we fix those, we can move forward with the review.

darkLord19:
@agnivade when i am running tests locally there i am not seeing any failures so not sure why it is failing here. Do you have any ideas?

darkLord19:
In CircleCI, Plugin Tests are failing but when i try it locally then I am not getting any failures.
```
# uparmar @ uparmar-mbp in ~/Development/mattermost-server on git:MM-8497 o [17:27:08] C:130
$ go test github.com/mattermost/mattermost-server/v5/api4 -run ^TestPlugin$ -race -count=1
ok      github.com/mattermost/mattermost-server/v5/api4 134.832s
```

agnivade:
It may not be the plugin tests, and just CircleCI incorrectly detecting test failures. Try testing the entire api4 directory.

Here is the full stack trace btw, if it helps you understand it better:

```
panic: 
assert: mock: I don't know what to return because the method call was unexpected.
	Either do Mock.On("Status").Return(...) first, or remove the Status() call.
	This method was unexpected:
		Status()
		
	at: [Store.go:490 app.go:712 status.go:401 scheduled_task.go:51 asm_amd64.s:1373]

goroutine 55232 [running]:
github.com/stretchr/testify/mock.(*Mock).fail(0xc058e9f4a0, 0x29d8733, 0xc1, 0xc132c84c00, 0x4, 0x4)
	/mattermost/mattermost-server/vendor/github.com/stretchr/testify/mock/mock.go:254 +0x17c
github.com/stretchr/testify/mock.(*Mock).MethodCalled(0xc058e9f4a0, 0x40293c2, 0x6, 0x0, 0x0, 0x0, 0xc132c84a00, 0x4, 0x4)
	/mattermost/mattermost-server/vendor/github.com/stretchr/testify/mock/mock.go:391 +0xea8
github.com/stretchr/testify/mock.(*Mock).Called(0xc058e9f4a0, 0x0, 0x0, 0x0, 0x200, 0x203001, 0x20301f)
	/mattermost/mattermost-server/vendor/github.com/stretchr/testify/mock/mock.go:356 +0x1a7
github.com/mattermost/mattermost-server/v5/store/storetest/mocks.(*Store).Status(0xc058e9f4a0, 0x2, 0x2)
	/mattermost/mattermost-server/store/storetest/mocks/Store.go:490 +0x48
github.com/mattermost/mattermost-server/v5/app.(*App).UpdateExpiredDNDStatuses(0xc043714120, 0x18, 0x0, 0x0, 0x0, 0x0)
	/mattermost/mattermost-server/app/app.go:712 +0x4c
github.com/mattermost/mattermost-server/v5/app.(*App).UpdateDNDStatusOfUsers(0xc043714120)
	/mattermost/mattermost-server/app/status.go:401 +0x46
github.com/mattermost/mattermost-server/v5/model.createTask.func1(0xc07e45fb80, 0x45d964b800, 0xc07e4ab4b0)
	/mattermost/mattermost-server/model/scheduled_task.go:51 +0x14c
created by github.com/mattermost/mattermost-server/v5/model.createTask
	/mattermost/mattermost-server/model/scheduled_task.go:40 +0x101
FAIL	github.com/mattermost/mattermost-server/v5/api4	444.567s
```

Somewhere, a mock for the status store is missing.

darkLord19:
In app.go, I am calling
```go
return a.Srv().Store.Status().UpdateExpiredDNDStatuses()
```
so it is complaining about Status() method about not having mock but Status() call mock is already present in Store.go
```go
// Status provides a mock function with given fields:
func (_m *Store) Status() store.StatusStore {
	ret := _m.Called()  //This is creating error

	var r0 store.StatusStore
	if rf, ok := ret.Get(0).(func() store.StatusStore); ok {
		r0 = rf()
	} else {
		if ret.Get(0) != nil {
			r0 = ret.Get(0).(store.StatusStore)
		}
	}

	return r0
}
```
Not much familiar with testify so unclear on what it's expecting.

jespino:
@darkLord19 the status mock exists, but you have to declare in the test that you are going to use it, and give the result. Something like:
```go
statusMock := mocks.Status{}
statusMock.On("WhateverFunctionYouAreCalling".Return(ExpectedResult)
storemock.On("Status").Return(statusMock)
```

darkLord19:
@jespino I am not directly calling the status function but it's called from scheduled task so not sure where would be good place for it. I added the mock declaration(see last commit) before the places where I thought it could get called but still having the issue.

jespino:
Probably you can fix this adding that to the `GetMockStoreForSetupFunctions` in the `testlib/store.go` (But I'm not 100% sure, so... lets try and see what happens :))

darkLord19:
StatusStore is already being mocked in the function you mentioned though.

esethna:
cc// @jespino ^

esethna:
@jespino and @agnivade can you provide any more guidance on the test failures?

agnivade:
I'll take a look.

agnivade:
Hey @darkLord19 - There's an import path which is wrong

```diff
diff --git a/app/plugin_api_test.go b/app/plugin_api_test.go
index d36e6d999..2dcc51e64 100644
--- a/app/plugin_api_test.go
+++ b/app/plugin_api_test.go
@@ -24,7 +24,7 @@ import (
        "github.com/mattermost/mattermost-server/v5/einterfaces/mocks"
        "github.com/mattermost/mattermost-server/v5/model"
        "github.com/mattermost/mattermost-server/v5/plugin"
-       storemocks "github.com/mattermost/mattermost-server/v5/storetest/mocks"
+       storemocks "github.com/mattermost/mattermost-server/v5/store/storetest/mocks"
        "github.com/mattermost/mattermost-server/v5/utils"
        "github.com/mattermost/mattermost-server/v5/utils/fileutils"
``` 

It seems like I don't have push permissions to your fork. Can we try fixing this and merging with latest master and running CI again?

Also, it really helps us reviewers if you can add new commits instead of force pushing. Because we cannot say what changed in between reviews. Thanks!

agnivade:
@darkLord19 - Can we try what @jespino suggested and see if that works:

```diff
diff --git a/testlib/store.go b/testlib/store.go
index 88feb9237..3e3305abd 100644
--- a/testlib/store.go
+++ b/testlib/store.go
@@ -64,6 +64,7 @@ func GetMockStoreForSetupFunctions() *mocks.Store {
 
        statusStore := mocks.StatusStore{}
        statusStore.On("ResetAll").Return(nil)
+       statusStore.On("UpdateExpiredDNDStatuses").Return([]*model.Status{}, nil)
 
        channelStore := mocks.ChannelStore{}
        channelStore.On("ClearCaches").Return(nil)
```

darkLord19:
I had already tried that but that didn't work either. I force pushed on top of it to avoid commits clutter.

agnivade:
Thanks @darkLord19. So the previous fix should have done the job, but it appears that there is a bug in our test framework.

In all of our functions that setup a mock store, we ultimately reset the store to an empty store:
```
emptyMockStore := mocks.Store{}
emptyMockStore.On("Close").Return(nil)
th.App.Srv().Store = &emptyMockStore
```

Therefore, the caller has to do all the work of setting up mocks again, and any predefined mocks get wiped off. So any tests using the mock store which take more than 5 minutes to run will try to execute that code and fail.

Can you remove the above three lines from these places and run again:

```
api4/apitestlib.go:254: emptyMockStore := mocks.Store{}
api4/apitestlib.go:262: emptyMockStore := mocks.Store{}
api4/apitestlib.go:270: emptyMockStore := mocks.Store{}
app/helper_test.go:174: emptyMockStore := mocks.Store{}
app/helper_test.go:183: emptyMockStore := mocks.Store{}
web/web_test.go:56:     emptyMockStore := mocks.Store{}
```

api4/plugin_test.go
```diff
@@ -90,6 +91,11 @@ func TestPlugin(t *testing.T) {
 			}))
 			defer func() { slowTestServer.Close() }()
 
+			storeMock := mocks.Store{}
```
darkLord19, line 94: @agnivade  I am mocking store and status again in few places which are giving errors so is this not what you meant by mocking again from all callers?

agnivade: Did you mean that `TestPlugin` is already failing for you? It passes for me though.

Anyways, looks like my latest suggestion did not work. I will create a separate branch and look into it to avoid having to ask you every time for changes.

Thanks for your patience on this @darkLord19

darkLord19:
@agnivade you can now push edits to this. I've provided you access.

agnivade:
@darkLord19 - It should be good now. The issue was that the scheduled task wasn't cancelled on server shutdown. So if anything runs over 5 minutes, which our test suite does, it tries to run the task. But then the store was closed for that app instance, so it fails.

I have added a cancellation part during server shutdown. But this is making the tests you added fail. That's probably a timing issue I'll leave you to take a look at.

There were also some redundant lines added during this whole process which I commented out.

darkLord19:
@agnivade this is great. Thanks for your help. Please let me know about how did you figure out issue was due to timing when you have some time. May help me debug stuff in future by learning!

@jespino and you can have one review and let me know any concerns if any so I can address them. I'd like to have your opinion especially on how I am making sql query for updating statuses directly in db. PostgreSQL has inbuilt support to do such op but MySQL doesn't so I've had to use different methods based on db driver matching. So can you have a look and see if is the correct approach or we can do anything better.

app/app.go
```diff
@@ -100,6 +100,9 @@ func (a *App) InitServer() {
 			})
 		}
 		a.srv.RunJobs()
+		if a.IsLeader() {
```
jespino, line 103: The problem of this is, on leader change, it doesn't create the new task. For example, if the leader not gets restarted, and other node takes the leadership, is not going to unset the DND statuses.

store/sqlstore/status_store.go
```diff
@@ -99,6 +102,89 @@ func (s SqlStatusStore) GetByIds(userIds []string) ([]*model.Status, error) {
...
+		return nil, errors.Wrap(err, "getExpiredDNDStatusesT: failed to get expired dnd statuses")
+	}
+	updateQuery, args, _ := s.getQueryBuilder().
+		Update("Status").
+		Where(
+			sq.Eq{"Status": model.STATUS_DND},
+			sq.Gt{"DNDEndTimeUnix": 1},
+			sq.LtOrEq{"DNDEndTimeUnix": time.Now().UTC().Unix()},
+		).
+		Set("Status", "PrevStatus").
```
jespino, line 127: This is setting the status value to the string `"PrevStatus"`, I think you have to use something like `sq.Field` or something like that.

store/sqlstore/status_store.go
```diff
@@ -99,6 +102,89 @@ func (s SqlStatusStore) GetByIds(userIds []string) ([]*model.Status, error) {
...
 
+// MySql doesn't has support for RETURNING clause or any euivalent so use transaction
+func (s SqlStatusStore) updateExpiredStatuses(t *gorp.Transaction) ([]*model.Status, error) {
+	selectQuery, selectParams, _ := s.getQueryBuilder().
+		Select("UserId, Status").
+		From("Status").
+		Where(
+			sq.Eq{"Status": model.STATUS_DND},
+			sq.Gt{"DNDEndTimeUnix": -1},
+			sq.LtOrEq{"DNDEndTimeUnix": time.Now().UTC().Unix()},
```
jespino, line 113: I would get the `time.Now().UTC().Unix()` as a variable to make the both queries completly identical.

store/sqlstore/status_store.go
```diff
@@ -99,6 +102,89 @@ func (s SqlStatusStore) GetByIds(userIds []string) ([]*model.Status, error) {
...
 }
 
+// MySql doesn't has support for RETURNING clause or any euivalent so use transaction
+func (s SqlStatusStore) updateExpiredStatuses(t *gorp.Transaction) ([]*model.Status, error) {
+	selectQuery, selectParams, _ := s.getQueryBuilder().
+		Select("UserId, Status").
+		From("Status").
+		Where(
+			sq.Eq{"Status": model.STATUS_DND},
+			sq.Gt{"DNDEndTimeUnix": -1},
```
jespino, line 112: Why here is `-1` and later is `1`?

store/sqlstore/status_store.go
jespino: You can use the query builder normally and then use `.Suffix("RETURNING UserId, Status")`, for example.

store/sqlstore/status_store.go
jespino: By the way, we usually use `0` for "unset timestamp", so... using `-1` here is not consistent with the rest of the application.

app/app.go
darkLord19: should we start task on all servers or is there any way to know when leader change event happens?

jespino: There is a function in `app/cluster.go` called `AddClusterLeaderChangedListener` that allows you to register a cluster leader change, that should do the trick. You can see an example of the usage of this in the line 420 of the `app/server.go` file.

jespino: buy the way, this logic fits better in the `app/server.go:NewServer` function, than in the `app/app.go:InitServer` function.

darkLord19: but in new server we don't have App available so we can't access UpdateDNDStatusOfUsers method

app/app.go
```diff
@@ -103,6 +103,16 @@ func (a *App) InitServer() {
 		if a.IsLeader() {
 			a.srv.dndTask = model.CreateRecurringTask("Unset DND Statuses", a.UpdateDNDStatusOfUsers, 300*time.Second)
 		}
+		a.srv.AddClusterLeaderChangedListener(func() {
```
jespino, line 106: This should be done in the `app/server.go` I think.

app/app.go
```diff
@@ -103,6 +103,16 @@ func (a *App) InitServer() {
...
+			mlog.Info("Cluster leader changed. Determining if unset dnd status task should be running", mlog.Bool("isLeader", a.IsLeader()))
+			if a.IsLeader() {
+				a.srv.dndTask = model.CreateRecurringTask("Unset DND Statuses", a.UpdateDNDStatusOfUsers, 300*time.Second)
+			} else {
+				if a.srv.dndTask != nil {
+					a.srv.dndTask.Cancel()
```
jespino, line 112: This is a tricky one, but you are "Cancelling" the dndTask and not setting it to nil, so... for example if you have 3 nodes, and the leadership changes from the node 1 to the node 2, the node one cancel the task and the node 2 starts it, then, if the leadership changes from node 2 to node 3, the node one tries to cancel the task again (closing an already closed channel).

app/app.go
darkLord19: in app/server.go we don't have App receiver so it won't be possible to start a task for App.UpdateDNDStatusOfUsers

app/app.go
darkLord19: i think setting the task to nil in else if condition should take care of this
```go
else {
    if a.srv.dndTask != nil {
        a.srv.dndTask.Cancel()
        a.srv.dndTask = nil
    }
}
```
 what do you think?

esethna: @jespino see comments above ^

esethna:
@agnivade @jespino please re-review when you can

esethna (approved):

agnivade:
I am still recovering from my illness, and will be OOO from Fri to Mon. I can review this on Tuesday. If it is urgent, please feel free to ask a review from any other Server team member.

app/app.go
```diff
@@ -98,6 +98,20 @@ func (a *App) InitServer() {
 			})
 		}
 		a.srv.RunJobs()
+		if a.IsLeader() {
```
agnivade, line 101: This whole logic should be moved inside a separate function `runDNDStatusExpireJob(a)` and called from inside `a.Srv().runjobs` like above.

app/app.go
```diff
@@ -98,6 +98,20 @@ func (a *App) InitServer() {
...
+			a.srv.dndTask = model.CreateRecurringTaskFromNextIntervalTime("Unset DND Statuses", a.UpdateDNDStatusOfUsers, 5*time.Minute)
+		}
+		a.srv.AddClusterLeaderChangedListener(func() {
+			mlog.Info("Cluster leader changed. Determining if unset dnd status task should be running", mlog.Bool("isLeader", a.IsLeader()))
+			if a.IsLeader() {
```
agnivade, line 106: This whole event happens inside a goroutine. We need to guard access to `dndTask` with a mutex.

app/plugin_api.go
```diff
@@ -311,6 +311,14 @@ func (api *PluginAPI) UpdateUserStatus(userID, status string) (*model.Status, *m
 	return api.app.GetStatus(userID)
 }
 
+func (api *PluginAPI) UpdateUserStatusWithDNDTimeout(userId, status, endtime string) (*model.Status, *model.AppError) {
```
agnivade, line 314: userID, endTime

store/sqlstore/status_store.go
```diff
@@ -99,6 +102,98 @@ func (s SqlStatusStore) GetByIds(userIds []string) ([]*model.Status, error) {
...
+				sq.Eq{"Status": model.STATUS_DND},
+				sq.Gt{"DNDEndTimeUnix": 0},
+				sq.LtOrEq{"DNDEndTimeUnix": time.Now().UTC().Unix()},
+			},
+		).
+		Set("Status", sq.Expr("PrevStatus")).
+		Set("PrevStatus", model.STATUS_DND).
+		Set("DNDEndTimeUnix", 0).
+		Set("Manual", false).
+		Suffix("RETURNING userId, Status").
```
agnivade, line 174: It does not seem correct to be returning only userID and status here. We are setting the returned status in the cache, which is then used for other purposes. We should be returning all the fields.

store/sqlstore/status_store.go
```diff
@@ -99,6 +102,98 @@ func (s SqlStatusStore) GetByIds(userIds []string) ([]*model.Status, error) {
...
+		Set("Manual", false).
+		ToSql()
+
+	if _, err := t.Exec(updateQuery, args...); err != nil {
+		return nil, errors.Wrapf(err, "failed to update statuses")
+	}
+	return selectedUsers, nil
+}
+
+func (s SqlStatusStore) UpdateExpiredDNDStatuses() ([]*model.Status, error) {
```
agnivade, line 144: Please add exhaustive tests for this method.

store/sqlstore/status_store.go
```diff
@@ -99,6 +102,98 @@ func (s SqlStatusStore) GetByIds(userIds []string) ([]*model.Status, error) {
...
+		}
+		defer finalizeTransaction(transaction)
+		userIds, err := s.updateExpiredStatuses(transaction)
+		if err != nil {
+			return nil, errors.Wrap(err, "CreateInitialSidebarCategories: createInitialSidebarCategoriesT")
+		}
+		if err := transaction.Commit(); err != nil {
+			return nil, errors.Wrap(err, "CreateInitialSidebarCategories: commit_transaction")
+		}
+		return s.GetByIds(userIds)
```
agnivade, line 158: First of all, `GetByIds` query replica, so it won't work. Secondly, I don't think we need another query at all. We can just use the transaction to directly return the statuses.

There's also some amount of logic duplication which can be refactored.

store/sqlstore/status_store.go
```diff
@@ -99,6 +102,98 @@ func (s SqlStatusStore) GetByIds(userIds []string) ([]*model.Status, error) {
 	return statuses, nil
 }
 
+// MySql doesn't has support for RETURNING clause or any euivalent so use transaction
```
agnivade, line 105: Please use full sentences.

> MySQL doesn't have support for RETURNING clause, so we use a transaction to get the updated rows.

app/status.go
```diff
@@ -365,6 +396,19 @@ func (a *App) IsUserAway(lastActivityAt int64) bool {
 	return model.GetMillis()-lastActivityAt >= *a.Config().TeamSettings.UserStatusAwayTimeout*1000
 }
 
+// UpdateDNDStatusOfUsers is a recurring task which is started when server starts
+// which unsets dnd status of users if needed and saves and broadcasts it
+func (a *App) UpdateDNDStatusOfUsers() {
+	statuses, err := a.UpdateExpiredDNDStatuses()
+	if err != nil {
+		mlog.Error("Failed to fetch dnd statues from store", mlog.String("err", err.Error()))
```
agnivade, line 404: `mlog.Warn` and return.

agnivade (changes requested):
First round of comments. Looks good overall. My main question is why we need a `DNDEndTime` and `DNDEndTimeUnix`.

Also please send another PR to https://github.com/mattermost/mattermost-api-reference/ updating the API to include the `DNDEndTime` in the request body.

app/app.go
darkLord19: runJobs won't have app receiver available so can't call from inside it

app/app.go
```diff
@@ -98,6 +98,7 @@ func (a *App) InitServer() {
 			})
 		}
 		a.srv.RunJobs()
+		a.runDNDStatusExpireJob()
```
agnivade, line 101: That's not quite what I meant. See how it's done in L96 and 97.

model/status.go
```diff
@@ -25,6 +25,9 @@ type Status struct {
 	Manual         bool   `json:"manual"`
 	LastActivityAt int64  `json:"last_activity_at"`
 	ActiveChannel  string `json:"active_channel,omitempty" db:"-"`
+	DNDEndTime     string `json:"dnd_end_time" db:"-"`
+	DNDEndTimeUnix int64  `json:"-"`
```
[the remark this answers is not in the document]
darkLord19: removed string representation. frontend will send endtime in unix epoch only

plugin/api.go
```diff
@@ -229,6 +229,13 @@ type API interface {
 	// Minimum server version: 5.2
 	UpdateUserStatus(userId, status string) (*model.Status, *model.AppError)
 
+	// UpdateUserStatusWithDNDTimeout will set a user's status until the user, or another integration/plugin, sets it back to online.
+	// The status parameter can be: "online", "away", "dnd with end time", or "offline".
```
agnivade, line 233: Is it really "dnd with end time" or just "dnd"? From what I am seeing, the status is still `model.STATUS_DND`. What makes it timed is the `endTime` parameter.

I think we can remove the `status` parameter altogether and just keep this API endpoint for setting timed dnd statuses. Right now, for setting online or away status, there seems to be two APIs to do the same thing. It might confuse users as to which one to use. We can keep using `UpdateUserStatus` for the old stuff, and use the new API purely to set timed DND status. This way we can also get rid of the if condition which checks for DND and uses the new path.

store/sqlstore/status_store.go
```diff
@@ -99,6 +102,110 @@ func (s SqlStatusStore) GetByIds(userIds []string) ([]*model.Status, error) {
 	return statuses, nil
 }
 
+// MySQL doesn't have support for RETURNING clause, so we use a transaction to get the updated rows.
+func (s SqlStatusStore) updateExpiredStatuses(t *gorp.Transaction) ([]*model.Status, error) {
+	currUnixTime := time.Now().UTC().Unix()
+	selectQuery, selectParams, _ := s.getQueryBuilder().
```
agnivade, line 108: Handle error here and the next query

store/sqlstore/status_store.go
agnivade: Tests are yet to be added.

store/sqlstore/status_store.go
```diff
@@ -99,6 +102,110 @@ func (s SqlStatusStore) GetByIds(userIds []string) ([]*model.Status, error) {
...
+		return nil, errors.Wrap(err, "getExpiredDNDStatusesT: failed to get updated dnd statuses")
+	}
+	return updatedStatuses, nil
+}
+
+func (s SqlStatusStore) UpdateExpiredDNDStatuses() ([]*model.Status, error) {
+	if s.DriverName() == model.DATABASE_DRIVER_MYSQL {
+		transaction, err := s.GetMaster().Begin()
+		if err != nil {
+			return nil, errors.Wrap(err, "CreateInitialSidebarCategories: begin_transaction")
```
agnivade, line 159: Copy paste error here and other places too

store/sqlstore/status_store.go
```diff
@@ -99,6 +102,110 @@ func (s SqlStatusStore) GetByIds(userIds []string) ([]*model.Status, error) {
...
+		Set("DNDEndTimeUnix", 0).
+		Set("Manual", false).
+		ToSql()
+
+	if _, err := t.Exec(updateQuery, args...); err != nil {
+		return nil, errors.Wrapf(err, "failed to update statuses")
+	}
+
+	var updatedStatuses []*model.Status
+	selectQuery, selectParams, _ = s.getQueryBuilder().
```
agnivade, line 143: Why do we need to do this query? We know what are the things we are changing- status, prevStatus, dndendTimeUnix and manual. We can get all the rows from the first query, and update the records ourselves and return.

store/sqlstore/upgrade.go
```diff
@@ -937,6 +937,10 @@ func upgradeDatabaseToVersion530(sqlStore *SqlStore) {
 		sqlStore.CreateColumnIfNotExistsNoDefault("FileInfo", "Content", "longtext", "text")
 
 		sqlStore.CreateColumnIfNotExists("SidebarCategories", "Muted", "tinyint(1)", "boolean", "0")
+
+		sqlStore.CreateColumnIfNotExistsNoDefault("Status", "DNDEndTimeUnix", "BIGINT", "BIGINT")
+		sqlStore.CreateColumnIfNotExistsNoDefault("Status", "PrevStatus", "VARCHAR(32)", "VARCHAR(32)")
```
agnivade, line 942: These need to be moved to `upgradeDatabaseToVersion534` now. Otherwise they won't be applied.

app/server.go
```diff
@@ -868,6 +871,10 @@ func (s *Server) Shutdown() {
 		mlog.Warn("Error flushing logs", mlog.Err(err))
 	}
 
+	if s.dndTask != nil {
+		s.dndTask.Cancel()
+	}
```
agnivade, line 876: There is still a race here. Wrap it up with the mutex:
```suggestion
	s.dndnTaskMut.Lock()
	if s.dndTask != nil {
		s.dndTask.Cancel()
	}
	s.dndnTaskMut.Unlock()
```

agnivade (changes requested):
Looking really good now. Some more things to take care of. Thanks for the work on this @darkLord19 !

agnivade:
Also please don't forget to send another PR to https://github.com/mattermost/mattermost-api-reference/ updating the API to include the DNDEndTime in the request body.

esethna:
Hey @darkLord19 just checking on this, are next steps clear of can we help clarify?

store/sqlstore/status_store.go
```diff
@@ -99,6 +102,118 @@ func (s SqlStatusStore) GetByIds(userIds []string) ([]*model.Status, error) {
...
+	}
+
+	return statuses, nil
+}
+
+func (s SqlStatusStore) UpdateExpiredDNDStatuses() ([]*model.Status, error) {
+	if s.DriverName() == model.DATABASE_DRIVER_MYSQL {
+		transaction, err := s.GetMaster().Begin()
+		if err != nil {
+			return nil, errors.Wrap(err, "CreateInitialSidebarCategories: begin_transaction")
```
agnivade, line 163: This still does not seem to be resolved.

store/sqlstore/status_store.go
```diff
@@ -99,6 +102,118 @@ func (s SqlStatusStore) GetByIds(userIds []string) ([]*model.Status, error) {
...
+		}
+		defer finalizeTransaction(transaction)
+		statuses, err := s.updateExpiredStatuses(transaction)
+		if err != nil {
+			return nil, errors.Wrap(err, "CreateInitialSidebarCategories: createInitialSidebarCategoriesT")
```
agnivade, line 168: Same

store/sqlstore/status_store.go
```diff
@@ -99,6 +102,118 @@ func (s SqlStatusStore) GetByIds(userIds []string) ([]*model.Status, error) {
...
+		}
+		if err := transaction.Commit(); err != nil {
+			return nil, errors.Wrap(err, "CreateInitialSidebarCategories: commit_transaction")
```
agnivade, line 171: Same

store/storetest/status_store.go
```diff
@@ -61,3 +63,31 @@ type ByUserId []*model.Status
 func (s ByUserId) Len() int           { return len(s) }
 func (s ByUserId) Swap(i, j int)      { s[i], s[j] = s[j], s[i] }
 func (s ByUserId) Less(i, j int) bool { return s[i].UserId < s[j].UserId }
+
+func testUpdateExpiredDNDStatuses(t *testing.T, ss store.Store) {
+	userId := NewTestId()
```
agnivade, line 68: Nit: userID

store/storetest/status_store.go
```diff
@@ -61,3 +63,31 @@ type ByUserId []*model.Status
...
+
+	time.Sleep(3 * time.Second)
+
+	// after 3 more seconds test status should be updated
+	statuses, err = ss.Status().UpdateExpiredDNDStatuses()
+	require.Nil(t, err)
+	require.Len(t, statuses, 1)
+
+	updatedStatus := *statuses[0]
+	require.Equal(t, updatedStatus.Status, model.STATUS_ONLINE)
```
agnivade, line 89: Verify the userID also.

store/storetest/status_store.go
```diff
@@ -61,3 +63,31 @@ type ByUserId []*model.Status
...
+
+	status := &model.Status{UserId: userId, Status: model.STATUS_DND, Manual: true,
+		DNDEndTime: time.Now().Add(5 * time.Second).Unix(), PrevStatus: model.STATUS_ONLINE}
+	require.NoError(t, ss.Status().SaveOrUpdate(status))
+
+	time.Sleep(2 * time.Second)
+
+	// after 2 seconds no statuses should be expired
+	statuses, err := ss.Status().UpdateExpiredDNDStatuses()
+	require.Nil(t, err)
```
agnivade, line 78: require.NoError

store/storetest/status_store.go
```diff
@@ -61,3 +63,31 @@ type ByUserId []*model.Status
...
+	require.Len(t, statuses, 0)
+
+	time.Sleep(3 * time.Second)
+
+	// after 3 more seconds test status should be updated
+	statuses, err = ss.Status().UpdateExpiredDNDStatuses()
+	require.Nil(t, err)
```
agnivade, line 85: require.NoError

model/scheduled_task_test.go
```diff
@@ -73,3 +73,26 @@ func TestCancelTask(t *testing.T) {
...
 }
+
+func TestCreateRecurringTaskFromNextIntervalTime(t *testing.T) {
+	TASK_NAME := "Test Recurring Task starting from next interval time"
+	TASK_TIME := time.Second * 5
+
+	var executionTime time.Time
+	testFunc := func() {
+		executionTime = time.Now()
+	}
```
agnivade, line 84: This is racy because the ticker runs in a goroutine. Wrap this with a mutex, and always run your test with `-race` to verify that there are no races.

model/scheduled_task_test.go
```diff
@@ -73,3 +73,26 @@ func TestCancelTask(t *testing.T) {
...
+	assert.EqualValues(t, 0, executionTime.Second()%5)
+
+	time.Sleep(TASK_TIME)
+	assert.EqualValues(t, 0, executionTime.Second()%5)
+
+	assert.Equal(t, TASK_NAME, task.Name)
+	assert.Equal(t, TASK_TIME, task.Interval)
+	assert.True(t, task.Recurring)
+
+	task.Cancel()
```
agnivade, line 97: This is better suited in a `defer` statement.

model/scheduled_task_test.go
```diff
@@ -73,3 +73,26 @@ func TestCancelTask(t *testing.T) {
 	time.Sleep(TASK_TIME + time.Second)
 	assert.EqualValues(t, 0, atomic.LoadInt32(executionCount))
 }
+
+func TestCreateRecurringTaskFromNextIntervalTime(t *testing.T) {
+	TASK_NAME := "Test Recurring Task starting from next interval time"
+	TASK_TIME := time.Second * 5
```
agnivade, line 79: Reduce it to 2 seconds.

agnivade (changes requested):
Just a last round of changes needed to the tests. Otherwise looking great now. Thank you for your patience with this @darkLord19.

agnivade:
@streamer45 - This adds some migrations in the status table. Shouldn't need a perf review. Just an FYI.

agnivade (approved):
Great work @darkLord19 !

jespino (approved):
Looks good to me. Thanks @darkLord19!

esethna:
@jespino @agnivade is this PR still blocked/awaiting PR? How are we going to QA? Should we merge this and then QA on the webapp PR? 

cc// @ogi-m @darkLord19

agnivade:
`Do Not Merge/Awaiting PR` usually means there is a dependent PR. In this case, it is the webapp PR. Once both PRs are review complete, the process is to remove the label and merge both.

We can QA this by pulling down the branches locally and testing the feature.

esethna:
cc// @jgilliam17 ^ I've added you to the webapp PR as well. https://github.com/mattermost/mattermost-webapp/pull/7037

Dinoryong:
Right now, this issue only can set end time of dnd status mode. 
But in my opinion, users would strongly want to set their working hours specifically. 
do you have any plans for this idea?

hmhealey:
@jgilliam17 Nope, this one is also using the wrong branch of the web app. We'll have to set something up manually or test on master.

@darkLord19 Are you aware of anything that might break if we merge this PR into master without the web app one? It's been a while since I've looked at the code for any of it.

darkLord19:
@hmhealey  I am not aware of anything which can break so it should be fine to merge if needed. There is one api-reference pr https://github.com/mattermost/mattermost-api-reference/pull/629 which i'm not sure needs to be merge with server pr so you can check it out once

jgilliam17:
Crossposting here as I already posted a similar message in the related webapp PR.
I'll have the test server available some time tomorrow and will add test results there by eod Monday.

jgilliam17 (approved):
Thanks @darkLord19 
Tested on a test server Saturn provided that included changes from both PRs (webapp and server), looks good to merge.

esethna:
@agnivade @jespino can you guys help merge this?

PR #16072, opened 2020-10-25
MM-22950: Remove supplier test

agnivade:
This test is too flaky to be written as a test. Let's just remove it.

https://mattermost.atlassian.net/browse/MM-29950

```release-note
NONE
```


wiggin77 (approved):
Makes sense. Too timing sensitive.

streamer45 (approved):

PR #16087, opened 2020-10-26
Sync.pool session

arjunagl:
Uses the `sync.pool` for `Session` objects.

https://github.com/mattermost/mattermost-server/issues/15249

Please note: This is in draft version for @agnivade to have a quick review to make sure this is what is expected.  
I have added the `sync.pool` to the model layer, however if we add it to the app layer it might mean more changes to the references\imports.


services/cache/lru.go
```diff
@@ -217,9 +217,10 @@ func (l *LRU) get(key string, value interface{}) error {
 			*v = &u
 			return err
 		case **model.Session:
-			var s model.Session
+			var s = model.UserSessionPool.Get().(*model.Session)
```
agnivade, line 220: We don't need to get the object here. We can get it in the app layer itself and pass the session object in the get call. And then remove this case altogether because now it will get handled by the `value.(msgp.Unmarshaler);` case above.

agnivade (commented):
Yes, let's make the pool in the app layer please.

services/cache/lru.go
arjunagl: Sure, I will close the PR and reopen when done.

PR #16093, opened 2020-10-27
[MM-27746] Compare our current LRU cache with Ristretto 

### LINKED ISSUE #15247: Compare our current LRU cache with Ristretto and analyze the results
Our current cache implements a simple LRU algorithm.

This ticket is to replace that with ristretto which implements TinyLFU and is highly concurrent.

Essentially we want to compare the current LRU vs Ristretto. We need to run some load tests, compare cache hit rates, CPU and memory utilization, and look at the results.

Ristretto is still not production ready yet, but at a high level, this would be a great analysis of our cache algorithms and understand if LFU is a better fit for our workload.

And then, if the results are successful, when it is ready for production, we can make the switch.

----

JIRA: https://mattermost.atlassian.net/browse/MM-27746

darkLord19:
Fixes #15247


darkLord19:
@agnivade can you have a look at this and let me know if needed any changes. If not, then how can we run load tests?

agnivade:
Let's make sure that the tests pass first. We will need to get you an enterprise license for running the load tests. We can work on that once it's ready to be tested.

services/cache/lfu.go
```diff
@@ -0,0 +1,188 @@
...
+)
+
+// LFU cache implementation using the Ristretto library.
+type LFU struct {
+	name                   string
+	size                   int
+	cache                  *ristretto.Cache
+	defaultExpiry          time.Duration
+	invalidateClusterEvent string
+	len                    int
```
agnivade, line 23: Needs a mutex around len access.

services/cache/lfu.go
```diff
@@ -0,0 +1,188 @@
...
+	Size                   int
+	DefaultExpiry          time.Duration
+	InvalidateClusterEvent string
+}
+
+// NewLFU creates an LFU of the given size.
+func NewLFU(opts *LFUOptions) Cache {
+	lfu, err := ristretto.NewCache(&ristretto.Config{
+		NumCounters:        int64(10 * opts.Size),
+		MaxCost:            int64(opts.Size << 10),
```
agnivade, line 38: What is the reasoning behind "<< 10"?

agnivade (commented):
Basic thing looks good. Once the tests are passing, we can look into running load tests.

darkLord19:
@agnivade we may not be able to make all tests pass and even some functionality may not work either as currently there is no way to get all the keys stored in ristretto cache but we have Keys() method for current LRU cache and it is used in one or two places in code so those portions may not work as expected.

agnivade:
Yes, unfortunately some of the code is designed in a way which caches with a different key and then looks for a key in that value to match and delete. I think we can have a separate map of just the keys with value set to `struct{}`. And the `Keys()` can just return a copy of all the keys in that map. Just a hack so that we can perform the experiment. Depending on the results, we can then decide what to do.

darkLord19:
> Yes, unfortunately some of the code is designed in a way which caches with a different key and then looks for a key in that value to match and delete. I think we can have a separate map of just the keys with value set to `struct{}`. And the `Keys()` can just return a copy of all the keys in that map. Just a hack so that we can perform the experiment. Depending on the results, we can then decide what to do.

I was thinking about something similar using slice but the issue will be in ristretto cache some keys will be evicted due to policy and we'd be deleting some keys explicitly from code, we can handle explicit Del calls but our slice will contain evicted keys as well. I will test it out with having slice of keys and append when Set is called and delete when Del is called and return that slice when Keys is called.

agnivade:
There is an `OnEvict` hook in Config that you can use for that. https://pkg.go.dev/github.com/dgraph-io/ristretto#Config.

darkLord19:
> There is an `OnEvict` hook in Config that you can use for that. https://pkg.go.dev/github.com/dgraph-io/ristretto#Config.

I have already checked that. OnEvict gets ristretto.Item as argument which contains hashed key and not plain text key.

agnivade:
Ah my bad, I missed that. Yes, this is getting a bit complicated to implement.

Do you still want to go forward with your approach? It might work out, but I am wondering the benefits of investigating something which might not even have a proper way to implement it.

darkLord19:
I believe ristretto is missing many functionalities for which we'd anyway need to add our wrappers like for len and keys so that'd be obstacles to get benefits(if any) we get from ristretto cache itself. And since to implement len and keys we'd anyway need to have our own lock so, concurrent ops benchmark on LFU cache would be almost identical to that of current LRU. So, I think for the time being we should drop the idea of replacing current LRU with ristretto cache. I haven'd done much research but maybe we can look for other libraries in go which provide LFU caches and meet our requirements and then we can use that for comparison.

agnivade:
> So, I think for the time being we should drop the idea of replacing current LRU with ristretto cache. 

Sounds good to me. And thank you for your hard work!

> I haven'd done much research but maybe we can look for other libraries in go which provide LFU caches and meet our requirements and then we can use that for comparison.

Yes, I think that makes sense. We can just implement LFU using a library and use that. If you want to leave it at that, I can update the ticket accordingly with our findings and conclusions and leave it for somebody to pick else up. You are also most welcome if you want to keep working on it.

darkLord19:
> Yes, I think that makes sense. We can just implement LFU using a library and use that. If you want to leave it at that, I can update the ticket accordingly with our findings and conclusions and leave it for somebody to pick else up. You are also most welcome if you want to keep working on it.

I will keep working on it.

darkLord19:
@agnivade i have pushed new changes using github.com/bluele/gcache library and it is having all the things required and all the tests are passing as well. we can do comparison now I think.

services/cache/lfu.go
```diff
@@ -0,0 +1,175 @@
...
+type LFUOptions struct {
+	Name                   string
+	Size                   int
+	DefaultExpiry          time.Duration
+	InvalidateClusterEvent string
+}
+
+// NewLFU creates an LFU of the given size.
+func NewLFU(opts *LFUOptions) Cache {
+	lfu := gcache.New(opts.Size).LRU().Build()
```
agnivade, line 35: Shouldn't this be `.LFU()`?

darkLord19: ohh nice catch. I mis-typed.

agnivade:
Thanks. I am following up internally to get you a license.

agnivade:
Closing this as done.

PR #16096, opened 2020-10-27
MM-30005: Require manage_system to patch system_admin role.

mkraft:
#### Summary

Requires `manage_system` permission to patch `system_admin` role.

#### Ticket Link

https://mattermost.atlassian.net/browse/MM-30005

#### Release Note

```release-note
NONE
```


sbishel (approved):

jupenur (approved):

thefactremains (approved):

srkgupta (changes requested):
Hi @mkraft 
Although this is now handled for system_admin role, a user with sysconsole_write_user_management_permissions is still allowed to modify multiple other internal roles in the system, for example, try for system_manager/system_user_manager role. Since these roles are internal roles, the user should not be allowed to modify them. The user can escalate the system roles permissions using this. Can you please fix it across all the system roles as well.

srkgupta (approved):
Tested and the issue is fixed now and is working fine. Approving the PR.

PR #16105, opened 2020-10-28
MMCTL is not being included in the package

ethervoid:
#### Summary

It's was including it in the bin folder but when we package we need to include it in the dist/bin so right now the download_mmctl script accepts a second parameter to define the path where the binary is going to be stored

#### Release Note

```release-note
NONE
```

### DIFF
--- a/build/release.mk
+++ b/build/release.mk
@@ -1,4 +1,3 @@
-
 dist: | check-style test package
 
 build-linux:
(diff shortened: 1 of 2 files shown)


jwilander (approved):

jaydeland (approved):
Thank you - LGTM

PR #16115, opened 2020-10-29
MM-29979: make websocket writes zero-alloc

agnivade:
This reverts commit 31dac80e66c31a123e56b116e8173f64d1a88d2d.

### DIFF
--- a/app/web_conn.go
+++ b/app/web_conn.go
@@ -4,6 +4,8 @@
 package app
 
 import (
+	"bytes"
+	"encoding/json"
 	"fmt"
 	"sync"
 	"sync/atomic"
@@ -169,6 +171,11 @@ func (wc *WebConn) writePump() {
 		wc.WebSocket.Close()
 	}()
 
+	var buf bytes.Buffer
+	// 2k is seen to be a good heuristic under which 98.5% of message sizes remain.
+	buf.Grow(1024 * 2)
+	enc := json.NewEncoder(&buf)
+
 	for {
 		select {
 		case msg, ok := <-wc.send:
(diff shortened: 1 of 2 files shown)


agnivade:
Well, we are back to do a bit of allocations. So it's not exactly "zero-alloc" anymore. But the precomputeJSON already allocates small byte buffers. So unless we remove that, there's no escaping this. So might as well accept the compromise.

app/web_conn.go
```diff
@@ -201,20 +208,25 @@ func (wc *WebConn) writePump() {
...
 			}
 
-			var msgBytes []byte
+			buf.Reset()
+			var err error
 			if evtOk {
 				cpyEvt := evt.SetSequence(wc.Sequence)
-				msgBytes = []byte(cpyEvt.ToJson())
+				err = cpyEvt.Encode(enc)
 				wc.Sequence++
```
streamer45, line 216: Unrelated to the PR and probably a minor thing but should the sequence get bumped also in case of error?

streamer45 (approved):
Looks good :+1:

app/web_conn.go
agnivade: Hmm.. good point. I think it should not. But I need to verify a bit how the sequence number is used by the client. In any case, let's visit this in a separate PR.

isacikgoz (approved):
LGTM 👍

agnivade:
@lindalumitchell - This caused an issue earlier due to an oversight. Please do a quick smoke testing to ensure that all basic functionalities work, giving special focus to websocket events.

lindalumitchell (approved):
LGTM!

PR #16116, opened 2020-10-29
MM-29980: Optimize profilesInChannels cache to fast path

agnivade:
We add one more message type to the fast path- profiles in channels. There
are 2 primary reasons for this:

- This is not really a new model type, but just a map of users. And users already use
the fast path. So we can get some more gains without really investing much more code.
- A more important reason is that with the upcoming striped mutex changes, we will get
a higher throughput at the cost of a bit more CPU utilization. The reason being that
since less amount of time will be spent in lock-contention, the CPU is free to do more
stuff. So this change is to counter that increase.

As usual, this gives much better performance than the original decoder.

Micro-benchmark results
```
LRU/UserMap=new-8    16.6µs ± 3%     3.9µs ± 4%  -76.15%  (p=0.000 n=10+10)

LRU/UserMap=new-8    4.78kB ± 0%    2.74kB ± 0%  -42.65%  (p=0.000 n=10+10)

LRU/UserMap=new-8      38.0 ± 0%      30.0 ± 0%  -21.05%  (p=0.000 n=10+10)
```

https://mattermost.atlassian.net/browse/MM-29980

Here are some results from a load test. The comparison is done with a 2 node cluster; one running master
and one running with this patch so that it's easier to compare. The total users are 2000.

CPU
CPU

Avg Store times for GetAllProfilesInChannel
getallprofiles

Avg store times for InvalidateProfilesInChannelCacheByUser
invalidate

```release-note
NONE
```

### DIFF
--- a/services/cache/lru.go
+++ b/services/cache/lru.go
@@ -221,6 +221,11 @@ func (l *LRU) get(key string, value interface{}) error {
 			_, err := s.UnmarshalMsg(e.value)
 			*v = &s
 			return err
+		case *map[string]*model.User:
+			var u model.UserMap
+			_, err := u.UnmarshalMsg(e.value)
+			*v = u
+			return err
 		}
 
 		// Slow path for other structs.
--- a/model/serialized_gen.go
+++ b/model/serialized_gen.go
@@ -1620,3 +1620,160 @@ func (z *User) Msgsize() (s int) {
 	s += msgp.BoolSize + msgp.StringPrefixSize + len(z.MfaSecret) + msgp.Int64Size + msgp.BoolSize + msgp.StringPrefixSize + len(z.BotDescription) + msgp.Int64Size + msgp.StringPrefixSize + len(z.TermsOfServiceId) + msgp.Int64Size
 	return
 }
+
+// DecodeMsg implements msgp.Decodable
+func (z *UserMap) DecodeMsg(dc *msgp.Reader) (err error) {
+	var zb0003 uint32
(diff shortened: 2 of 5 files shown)


streamer45:
@agnivade FYI, there's a lint (gofmt) complain on `lru_test.go`.

agnivade:
Thanks, fixed.

services/cache/lru.go
```diff
@@ -221,6 +221,11 @@ func (l *LRU) get(key string, value interface{}) error {
 			_, err := s.UnmarshalMsg(e.value)
 			*v = &s
 			return err
+		case *map[string]*model.User:
+			var u model.UserMap
+			_, err := u.UnmarshalMsg(e.value)
```
streamer45, line 226: Unrelated question but since we are here I'd rather ask:
I understand a lot of locking pressure will be eased with the upcoming striped LRU implementation but did we ever consider doing the serialization/deserialization operations out of the critical section?

streamer45 (approved):
Looks like another great improvement, nicely done :+1:

services/cache/lru.go
agnivade: Nice one! Can you please file a ticket. I will take care of it.

streamer45: https://mattermost.atlassian.net/browse/MM-30140

agnivade:
@lindalumitchell - can be qa-deferred

lindalumitchell (approved):

PR #16117, opened 2020-10-29
MM-30041: Return correct error message for user save

agnivade:
We were collapsing all types of user conflict into a single
error message. Fixed it by inspecting the field of the invalidError type
and returning the correct message.

https://mattermost.atlassian.net/browse/MM-30041

```release-note
NONE
```

### DIFF
--- a/api4/apitestlib.go
+++ b/api4/apitestlib.go
@@ -875,7 +875,7 @@ func CheckErrorMessage(t *testing.T, resp *model.Response, errorId string) {
 	t.Helper()
 
 	require.NotNilf(t, resp.Error, "should have errored with message: %s", errorId)
-	require.Equalf(t, resp.Error.Id, errorId, "incorrect error message, actual: %s, expected: %s", resp.Error.Id, errorId)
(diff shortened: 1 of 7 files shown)


grundleborg (approved):

grundleborg:
@agnivade you've got some failing tests related to this change.

agnivade:
Yep, will fix.

rvillablanca:
Yep, sorry about that. We have been collapsing a lot of messages and in this case I missed the need for a detail when creating a user :disappointed:.

Thanks for the fix @agnivade. 

Another thing I noticed when looking into the messages of failed tests is that the expected and obtained value are in the wrong place. This causes a wrong message from stretchr/testify.

agnivade:
Ah yeah, that's a common mistake with require/assert. Thanks for that correction.

wiggin77 (approved):
LGTM :+1:

agnivade:
@srkgupta - do you want to test this? I have tested myself and added a unit test. But let me know if there is something else you want to check too. Thanks.

srkgupta (approved):
Thanks @agnivade. Approving it as

* it was tested by Agniva 
* we have daily cypress test runs which will auto-verify them.

PR #16170, opened 2020-10-30
MM-30026: Use DB master when getting team members from a session

agnivade:
A race condition happens when the read-replica isn't updated yet
by the time a session expiry message reaches another node in the cluster.

Here is the sequence of events that can cause it:
- Server1 gets any request which has to wipe session cache.

- The SQL query is written to DB master, and a cluster message is propagated
to clear the session cache for that user.

- Now before the read-replica is updated with the master’s update,
the cluster message reaches Server2. The session cache is wiped out for that user.

- _Any random_ request for that user hits Server2. Does NOT have to be
the update team name request. The request does not find the value
in session cache, because it’s wiped off, and picks it up from the DB.
Surprise surprise, it gets the stale value. Sticks it into the cache.

By now, the read-replica is updated. But guess what, we aren’t going to
ask the DB anymore, because we have it in the cache. And the cache has the stale value.

We use a temporary approach for now by introducing a context in the DB calls so that
the useMaster information can be easily passed. And this has the added advantage of
reusing the same context for future DB calls in case it happens. And we can also
add more context keys as needed.

A proper approach needs some architectural changes. See the issue for more details.

https://mattermost.atlassian.net/browse/MM-30026

```release-note
Fixed a bug where a session will hold on to a cached value
in an HA setup with read-replicas configured.
```


agnivade:
Gentle ping @mgdelacroix

store/sqlstore/context.go
```diff
@@ -0,0 +1,32 @@
...
+type contextValue string
+
+// Different possible values of contextValue.
+const (
+	useMaster contextValue = "useMaster"
+)
+
+// withMaster adds the context value that master DB should be selected for this request.
+func withMaster(ctx context.Context) context.Context {
+	return context.WithValue(ctx, storeContextKey(useMaster), "yes")
```
mkraft, line 21: `"yes"` vs `true`, what's the reason for choosing a string?

mkraft (commented):
I think `(*SqlTeamStore).GetMember` needs this to fix the issue for dancing-penguin.

store/sqlstore/context.go
agnivade: The exact value of yes/true doesn't really matter. We can choose any one. The reason to use string is that we can store more context values in the future, and keep the context key type same. A bool will not allow you to do that. Then you need different keys. It's cleaner to have a single key and different values.

mkraft (approved):

store/sqlstore/context.go
mkraft: Sure, I guess there could be some future values of "useMaster" other than "yes"/"no".

agnivade: Sorry for the confusion here. That's not quite what I mean. To expand further, a context.WithValue needs a "key" and a "value". And the key needs to be of a named type.

Right now, we have - storeContextKey("master"), "yes/no".

If we start to use a boolean, then the value part doesn't have any significance, and we need to have 2 separate keys.

```
useMaster = true
dontUse = false

storeContextKey(useMaster), ???
storeContextKey(dontUse), ???
```

It becomes something  like this: https://play.golang.org/p/a-s4l3Kuwe0

This is still okay, but now we have painted ourselves to a corner. Because `storeContextKey` is of type bool. So we cannot have any more values rather than true/false. Which means we need more context key types. Hence, the need to have string keys.

Does that help or was that even more confusing :P ?

mgdelacroix: I got a bit confused as well :sweat_smile: The ask is around using a boolean in the value rather than the key, right? Something like:

```go
context.WithValue(ctx, storeContextKey(useMaster), true)
```

With this the key type would remain the same and the value would express the "yes"/"no" as well without the need of the `dontUse`. In case we ever need to store another context value, we could reuse the key type but with a different value type. But I feel there is something I'm not seeing here

agnivade: Oops, you are right. My bad. I somehow thought we need to change the key type also. Let me correct that.

agnivade: @mgdelacroix - updated

mgdelacroix: Awesome, thanks!!

mgdelacroix (approved):
LGTM! :+1:

agnivade:
Ok, there's something strange going on. I can still reproduce the issue with this PR. But more importantly, a lot of other functionalities are not working too.

It has something to do with the cluster broadcast message. And it almost seems to skip alternate messages.

Here's what I am seeing:
1. Open two tabs. One with system admin, one with normal user.
2. As system admin, change the team name.
3. Observe the team name change in the user tab, and verify that the websocket message has come.
4. Change team name again.
5. NOTHING HAPPENS on the other tab. No ws message.
6. Change team name again.
7. The name changes now. Ws message also appears.

Keep repeating from steps 4-7.

Although the PR fixes a valid issue, I think there is a bigger underlying issue of missing cluster broadcast events somewhere.

/cc @jwilander

jwilander:
Hmm, that is concerning. I wonder if both your clients were connected to the same MM pod or if they were split across both. If some of the cluster events are getting passed to the other MM pods then it's not likely a networking issue. We could disable gossip on the test server to see if it's still reproducible after that.

@agnivade do you need anything to help reproduce or debug? Also do you know if we export any prometheus metrics that would let us see how often this is occurring?

agnivade:
From my browser, I am just connecting to the endpoint. I would assume it depends on the load balancer to redirect me to whichever pod.

> If some of the cluster events are getting passed to the other MM pods then it's not likely a networking issue. 

Well, I see alternate events being passed. i.e. one team name change event gets shown, the next does not, the next one again shows up. So I am not really sure what's happening.

I'll try disabling gossip and see. One thing I noticed is that the test HA servers are only available for 30 minutes. Is it possible to have a test HA server for a longer duration so that I can keep debugging ?

jwilander:
Hmm I didn't think we had anything that automatically turned off the test servers but you're right the bot removed the label

You could instead use `/cloud create <namehere> --size miniHA --version <dockertag>` on community to spin up one you have more control over

agnivade:
Ok I see that gossip has been enabled through an environment variable, so that won't work. Let me try with the /cloud command.

agnivade:
@jwilander has been kind enough to debug this with me and we figured out that the spinwicks created didn't have the EE license setup. But it did have 2 pods. So naturally, the lb would randomly route any request to any pod, but since the license wasn't there, the code wouldn't broadcast the message across the cluster. Hence the out of sync problem.

But I have setup a cloud instance with the /cloud command, and was able to reproduce the issue on master, and successfully confirmed that the issue does not exist with the PR!

@jupenur - I guess that takes off the responsibility off your shoulders. And given that I have 2 dev reviews, I'll pass this off to Lindy.

@lindalumitchell - I have already QA verified this and confirmed that the PR fixes the issue. For any other stuff, it can be qa-deferred.

lindalumitchell (approved):

PR #16188, opened 2020-11-01
MM-29981 - optimize markdown.Inspect

agnivade:
- Cache regexp.MustCompile
- Reuse slice in MergeInlineText
- Remove pointer to slice in closeBlocks
- Pre-allocate slice in ParseLines
- Some more small cleanups

```
name       old time/op    new time/op    delta
Inspect-8    10.5µs ± 3%     6.6µs ± 1%  -37.59%  (p=0.000 n=10+7)

name       old alloc/op   new alloc/op   delta
Inspect-8    6.66kB ± 0%    3.22kB ± 0%  -51.62%  (p=0.000 n=10+9)

name       old allocs/op  new allocs/op  delta
Inspect-8       117 ± 0%        76 ± 0%  -35.04%  (p=0.000 n=10+10)
```

https://mattermost.atlassian.net/browse/MM-29981


utils/markdown/lines.go
```diff
@@ -3,13 +3,16 @@
...
+
 type Line struct {
 	Range
 }
 
-func ParseLines(markdown string) (lines []Line) {
+func ParseLines(markdown string) []Line {
 	lineStartPosition := 0
 	isAfterCarriageReturn := false
+	lines := make([]Line, 0, strings.Count(markdown, "\n"))
```
agnivade, line 15: This is really an approximate capacity. And works for the most common case of multiple lines without carriage return.

utils/markdown/autolink.go
```diff
@@ -14,6 +14,7 @@ import (
 
 var (
 	DefaultUrlSchemes = []string{"http", "https", "ftp", "mailto", "tel"}
+	wwwAutoLinkRegex  = regexp.MustCompile(`^www\d{0,3}\.`)
```
streamer45, line 17: The need for this regexp escapes me. Why do we look for up to 3 digits after `www`?

utils/markdown/autolink.go
```diff
@@ -59,9 +60,8 @@ func isAllowedBeforeWWWLink(c byte) bool {
 	switch c {
 	case '*', '_', '~', ')':
 		return true
-	default:
-		return false
 	}
+	return false
```
streamer45, line 64: Interesting. Is returning outside the switch more idiomatic?

utils/markdown/autolink.go
agnivade: That's a good point. Maybe @hmhealey can shed some light.

utils/markdown/autolink.go
agnivade: I would say so.

```go
func IsSpace(r rune) bool {
  // This property isn't the same as Z; special-case it.
  if uint32(r) <= MaxLatin1 {
    switch r {
    case '\t', '\n', '\v', '\f', '\r', ' ', 0x85, 0xA0:
      return true
    }
    return false
  }
  return isExcludingLatin(White_Space, r)
}
```

The core idea is to reduce indentation as much as possible.

isacikgoz: I was thinking the opposite :) Somehow it looks more clear to me with the default case if the function/method only does a `switch`.

agnivade: Hmm .. I can remove it if you want. But in general, the convention has always been to try to reduce amount of code written with indentation. So as soon as the need for an indentation finishes, we can come back to our original indent. For the switch, I think of it as a compressed if block.

```
if x,y,z {
 return true
}
return false
```

isacikgoz: Yeah it makes total sense for `if` statements. And I see this occasion looks like a compressed `if` block. It may be a personal preference though but I tend to add a `default` case if I ever use switch :) And since for this specific case this is totally cosmetic, I leave it to you.

isacikgoz (approved):
Nice improvement 🚀

utils/markdown/autolink.go
hmhealey: I forget why specifically we added it, but I know I've seen websites hosted at things like `www1.example.com`.

hmhealey (approved):
It looks like pretty decent performance improvements relatively, but was this identified as somewhere that we'll need to be careful about performance in the future? I don't have any immediate plans to rewrite the server's markdown code, but it would be nice to know just in case 😛

agnivade:
This showed up in the flame graphs from load tests. This gets called in the createPost path, and within that, a lot of time was being spent in `markdown.Inspect`. The `regexp.MustCompile` was the main culprit. I just spent some more time optimizing more things I could find.

streamer45 (approved):
Looks good :+1: 
Maybe let's have QA do some smoke testing around Markdown just to be extra sure functionality is unchanged.

agnivade:
@lindalumitchell - Some testing around markdown in general would be helpful.

lindalumitchell:
@agnivade sorry for the delay on this one! I ran some Markdown tests, and all looks good except for syntax rendering/highlighting. There's a test file we load using `/test url test-syntax-highlighting`, and below are screenshots of what it looks like on this PR's test server (can view the post here: https://mattermost-server-pr-16188.test.mattermost.cloud/pr16188/pl/dacfp99hkif9zgsqxr5z3ttyqy), then what the expected looks like on master.

This PR:
unexpected-missing-code-syntax-highlighting

Expected on master:
master-expected-code-syntax-highlighting

agnivade:
@lindalumitchell - That's a bit strange, because syntax highlighting is done on the client side. There is nothing on the server that does that. I just tried that post in my local setup and it works fine. Not really sure why it's not showing up in the PR.

Re: latex, you'd have to enable that in the system console. I just did that and it works.

In short, this PR does not modify the post at all. As long as the raw message is the same in both cases, then it is unlikely that server is involved here.

@hmhealey - Any ideas as to why syntax highlighting won't work?

hmhealey:
The test server's web app is just out of date. That got fixed last week

agnivade:
@lindalumitchell - It works now.

lindalumitchell (approved):
Perfect, thanks for sorting out my false alarms! LGTM.

PR #16197, opened 2020-11-03
MM-26425: Make URL parsing for image proxy more robust

agnivade:
- We don't bypass protocol relative URLs.
- We don't bypass hostnames with a similar prefix.

https://mattermost.atlassian.net/browse/MM-26425

```release-note
NONE
```

### DIFF
--- a/services/imageproxy/atmos_camo.go
+++ b/services/imageproxy/atmos_camo.go
@@ -9,16 +9,25 @@ import (
 	"encoding/hex"
 	"io"
 	"net/http"
-	"strings"
+	"net/url"
 )
 
 type AtmosCamoBackend struct {
-	proxy *ImageProxy
+	proxy     *ImageProxy
+	siteURL   *url.URL
+	remoteURL *url.URL
 }
 
 func makeAtmosCamoBackend(proxy *ImageProxy) *AtmosCamoBackend {
+	// We deliberately ignore the error because it's from config.json.
+	// The function returns a nil pointer in case of error, and we handle it when it's used.
+	siteURL, _ := url.Parse(*proxy.ConfigService.Config().ServiceSettings.SiteURL)
+	remoteURL, _ := url.Parse(*proxy.ConfigService.Config().ImageProxySettings.RemoteImageProxyURL)
+
 	return &AtmosCamoBackend{
-		proxy: proxy,
+		proxy:     proxy,
+		siteURL:   siteURL,
+		remoteURL: remoteURL,
 	}
 }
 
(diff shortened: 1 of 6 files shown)


services/imageproxy/atmos_camo.go
```diff
@@ -45,22 +54,32 @@ func (backend *AtmosCamoBackend) GetImageDirect(imageURL string) (io.ReadCloser,
...
-	return getAtmosCamoImageURL(imageURL, siteURL, proxyURL, options)
-}
+	if imageURL == "" || backend.siteURL == nil {
+		return imageURL
+	}
+
+	// Parse url, bypass if error in parsing.
+	parsedURL, err := url.Parse(imageURL)
+	if err != nil {
+		return imageURL
```
jupenur, line 66: We should not return the original URL if it can't be parsed. Browsers are less strict and will likely still parse it, allowing more proxy bypasses.

services/imageproxy/atmos_camo.go
```diff
@@ -45,22 +54,32 @@ func (backend *AtmosCamoBackend) GetImageDirect(imageURL string) (io.ReadCloser,
...
+	}
 
-func getAtmosCamoImageURL(imageURL, siteURL, proxyURL, options string) string {
-	// Don't proxy blank images, relative URLs, absolute URLs on this server, or URLs that are already going through the proxy
-	if imageURL == "" || imageURL[0] == '/' || (siteURL != "" && strings.HasPrefix(imageURL, siteURL)) || strings.HasPrefix(imageURL, proxyURL) {
+	// If host is same as siteURL host/ remoteURL host, or it's a relative URL, return.
+	if parsedURL.Host == backend.siteURL.Host || parsedURL.Host == "" || parsedURL.Host == backend.remoteURL.Host {
 		return imageURL
```
jupenur, line 71: In fact, we should never return `imageURL`, only `parsedURL.String()` -- and even then we should always fill in the blanks first: if `parsedURL.Host` is empty we should set it to `backend.siteURL.Host`. Same for `Scheme`. That's because browsers will parse `https:///example.com` as pointing to the host `example.com`, whereas go sees an empty host name.

To quote Filippo on this: "any security property that relies on parser alignment is doomed" and "you don't parse and check, you rebuild and compare".

services/imageproxy/imageproxy.go
```diff
@@ -118,15 +124,24 @@ func (proxy *ImageProxy) GetImageDirect(imageURL string) (io.ReadCloser, string,
...
-
-func getProxiedImageURL(imageURL, siteURL string) string {
-	if imageURL == "" || imageURL[0] == '/' || strings.HasPrefix(imageURL, siteURL) {
+	if imageURL == "" || proxy.siteURL == nil {
 		return imageURL
 	}
+	// Parse url, bypass if error in parsing.
+	parsedURL, err := url.Parse(imageURL)
+	if err != nil {
+		return imageURL
```
jupenur, line 133: Same as above.

services/imageproxy/imageproxy.go
```diff
@@ -118,15 +124,24 @@ func (proxy *ImageProxy) GetImageDirect(imageURL string) (io.ReadCloser, string,
...
+	}
+	// If host is same as siteURL host, or it's a relative URL, return.
+	if parsedURL.Host == proxy.siteURL.Host || parsedURL.Host == "" {
+		return imageURL
```
jupenur, line 137: Same as above.

jupenur (changes requested):

services/imageproxy/atmos_camo.go
agnivade: >  if parsedURL.Host is empty we should set it to backend.siteURL.Host

There is a test against this which verifies that relative URLs should bypass the proxy. Are we sure we want to break this behavior? I am 0/5 on any of it, just confirming.

services/imageproxy/atmos_camo.go
agnivade: What do you want me to return in that place?

agnivade: What do you want me to return in its place?

services/imageproxy/atmos_camo.go
agnivade: Ah wait, you just want the host to be filled with the siteURL host right? It will still bypass, but the URL will be complete.

jupenur: That's it, yeah.

services/imageproxy/atmos_camo.go
jupenur: Not sure. Maybe just the siteURL or some other static dummy URL? 0/5 on whether that makes sense.

agnivade: Gotcha, will return the siteURL then.

isacikgoz (approved):
I think it is good to go after @jupenur's requests.

jupenur (changes requested):
Same comment as on Ibrahim's related PR: we also need to handle opaque URLs somehow, because with opaque URLs `String()` ignores the hostname.

In Ibrahim's patch the "somehow" was by returning an error; here we might need to return siteURL again instead.

Otherwise looks great now.

jupenur (approved):

agnivade:
@lindalumitchell - This is covered by unit tests. But it might be good to have some testing done around image proxy.

lindalumitchell (approved):

PR #16203, opened 2020-11-03
[MM-30126] User friendly invoice filenames in-app

nickmisasi:
#### Summary
A change was made to CWS to name invoices in a more user friendly way. Mattermost-server would receive that invoice and then rename it to the old way. This PR along with: https://github.com/mattermost/enterprise/pull/817 will pass along the filename returned by CWS instead of renaming it.

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-30126

#### Release Note
```release-note
None
```

### DIFF
--- a/api4/cloud.go
+++ b/api4/cloud.go
@@ -7,7 +7,6 @@ import (
 	"bytes"
 	"encoding/binary"
 	"encoding/json"
-	"fmt"
 	"io/ioutil"
 	"net/http"
 	"time"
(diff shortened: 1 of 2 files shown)


ethervoid (approved):
LGTM

marianunez (approved):

api4/cloud.go
```diff
@@ -303,14 +302,14 @@ func getSubscriptionInvoicePDF(c *Context, w http.ResponseWriter, r *http.Reques
...
-	pdfData, appErr := c.App.Cloud().GetInvoicePDF(c.Params.InvoiceId)
+	pdfData, filename, appErr := c.App.Cloud().GetInvoicePDF(c.Params.InvoiceId)
 	if appErr != nil {
 		c.Err = model.NewAppError("Api4.getSuscriptionInvoicePDF", "api.cloud.request_error", nil, appErr.Error(), http.StatusInternalServerError)
 		return
 	}
 
 	err := writeFileResponse(
-		fmt.Sprintf("%s.pdf", c.Params.InvoiceId),
+		filename,
```
ethervoid, line 312: Given that this could be empy, we should check for it and generate a filename if is not the case

nickmisasi: Added check in enterprise

PR #16212, opened 2020-11-04
MM-29709 Client side feature flags

crspeller:
#### Summary
Easy usage of feature flags from webapp. In limited client config because values are not confidential. 

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-29709

#### Release Note
```release-note
NONE
```


hmhealey (approved):
Using reflection so you can avoid having to add every single new feature flag to the client config, eh?

crspeller:
@hmhealey Exactly. Want to make them as painless as possible to use.

agarciamontoro (approved):
Looks good! :tada:

prapti:
Looks quite good looking at the unit test coverage. I am trying to test this locally with the split.io setup but looks like the old "split" is removed. @crspeller I probably need a new split key to test the real stuff, won't I?

prapti (approved):
Works on both local and cloud test server instances, despite having some trouble testing in cloud initially. The feature flag value gets set as expected.
Approving as the initial issue can't be reproduced anymore.

PR #16213, opened 2020-11-04
MM-30090 Add ManagedResourcePaths setting

hmhealey:
Due to some changes in 5.28 around the internal/external link handling in the web app, we broke the newly added [Desktop Managed Resources](https://docs.mattermost.com/install/desktop-managed-resources.html) feature which basically lets non-MM services host stuff at certain routes under the MM server.

Since most paths in the server could be a team URL, the new link handling is correct for anyone not using that feature, so instead of reverting the link handling changes, we're adding a way to the server to manually specify these paths that should not be handled by the web app.

Hopefully I'm setting the correct struct tags on this. I've set it as environment because it affects the web server, and it's not something that would be turned on without making changes to the proxy to intercept requests to certain paths.

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-30090

#### Related Pull Requests
https://github.com/mattermost/mattermost-redux/pull/1282
https://github.com/mattermost/mattermost-webapp/pull/7024
https://github.com/mattermost/docs/pull/4089

#### Release Note
```release-note
Added the ManagedResourcePaths setting for use with the Desktop Managed Resources feature
```


Willyfrog (approved):
LGTM

crspeller (approved):

PR #16220, opened 2020-11-05
[MM-29999] User Limit Email Notifications (Part 2)

nickmisasi:
#### Summary
At the 30, 90, and 91 day mark of being over the user limit without paying, we have emails to send to admin's to attempt to get them to upgrade (and then warn them of workspace suspension in the event they don't). This PR includes the 3 emails, as well as a background scheduled job that runs once per day to determine whether to send these emails or not. 

This is a WIP, but I don't expect major changes to this existing code, so looking for a pass of reviews. To do:

~~1. check that the status of the subscription is "in arrears", so if status == paid, don't run any of the above (this is going to be 3 different PR's but small ones)~~
~~2. "reset" the set date when we find that the subscription status == paid~~
~~3. A few things around putting exact dates into the emails, this is sort of challenging because you can't always rely on the latest_invoice. I think I'm going to need to add a GetInvoiceById method to server > enterprise > call to the CWS endpoint~~
~~4. Send an email to support@mattermost.com to open a ticket about the suspension of workspace~~
~~5. Figure out how we can test this, for both QA Review, and UX/PM Review.~~
~~6. Adjust forgiveness logic per Adam's recommendations (pending meeting)~~ - REMOVED, will wait to implement later
~~7. Include User Count in the support@mattermost.com email~~

#### Feature Flags
This PR includes a new feature flag:
`CloudDelinquentEmailJobsEnabled bool` - possible values `true`, `false`. 
- Setting to true will allow the scheduled job to run (so long as the ExperimentalFeatures.CloudUserLimit variable is non-zero)
- Setting to false will block the scheduled job from running.

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-29999
#### Release Note
```release-note
Mattermost cloud instances with delinquent subscriptions will be sent emails on a certain cadence, in certain cases:
A check happens twice per day to see if an installation is over the user limit
If that installation is over the user limit, they are marked for notification during the next billing cycle
If that installation is still over the limit 7, 14, 30, and 90 days into the next billing cycle, they will be sent emails with various wording to get them to add a payment method and upgrade their cloud billing 
If that installation is no longer over the limit 7, 14, and 30 days into the next billing cycle, we will "forgive" them, and not send any emails for that cycle. 
After the installation has gone over the limit, each time the job runs, it checks Stripe to check the state of the subscription. If it's flipped from "past_due" to active, it will cancel any emails for that cycle. Otherwise, it will continue on.

```
7 day email:
image
14 day email:
image
30 day email:
image
90 day email: 
image
91 day email: (not yet being sent)
image


jobs/cloud/worker.go
```diff
@@ -0,0 +1,284 @@
...
+
+	if systemValue == nil {
+		mlog.Info("Found nil end date")
+		if userDifference >= 0 {
+			// Under user limit, so no need to start tracking yet.
+			worker.LogAndSetJobSuccess(job)
+			return
+		}
+		// Our row doesn't exist, so create it and get it from the store
+		subCycleEndDate := time.Unix((subscription.EndAt / 1000), 0)
```
marianunez, line 191: Does the subscription `EndAt` value change for each monthly cycle or do we need to check the invoice?

jobs/cloud/worker.go
```diff
@@ -0,0 +1,284 @@
...
+	count, err := worker.app.Srv().Store.User().Count(model.UserCountOptions{IncludeDeleted: false})
+
+	if err != nil {
+		mlog.Error("Worker: Failed to get active user count", mlog.String("worker", worker.name), mlog.String("job_id", job.Id), mlog.String("error", err.Error()))
+		return
+	}
+
+	userDifference := cloudUserLimit - count
+
+	systemValue, nErr := worker.app.Srv().Store.System().GetByName(model.USER_LIMIT_OVERAGE_CYCLE_END_DATE)
```
marianunez, line 176: Given that this is much later on, can we rename to what is actually has for readability?
```suggestion
	overageEndDateSystemValue, nErr := worker.app.Srv().Store.System().GetByName(model.USER_LIMIT_OVERAGE_CYCLE_END_DATE)
```
My suggestion is a bit wordy but just to convey the idea

jobs/cloud/worker.go
```diff
@@ -0,0 +1,284 @@
...
+	// If we've reached this point, we know at some time the installation was over the limit
+
+	subCycleEndDate, err := time.Parse(dateLayout, systemValue.Value)
+	if err != nil {
+		mlog.Error("Unable to parse USER_LIMIT_OVERAGE_CYCLE_END_DATE", mlog.String("worker", worker.name), mlog.String("error", err.Error()))
+	}
+
+	now := time.Now()
+
+	daysDifference := int64(math.Floor(now.Sub(subCycleEndDate).Hours() / 24)) // Get the difference in days between now and cycle end date
```
marianunez, line 218: We need to send the email 7/14/30 days after the billing cycle right? So here wouldn't it be an addition?

So if the billing cycle end date is Oct. 31st., the first email to send would be Nov. 7th?

jobs/cloud/worker.go
```diff
@@ -0,0 +1,284 @@
...
+	sysAdmins, err := worker.app.GetUsers(userOptions)
+	switch daysDifference {
+	case 7:
+		forgiven := worker.ForgivenessCheck(int(userDifference))
+		if forgiven {
+			worker.LogAndSetJobSuccess(job)
+			return
+		}
+
+		// TODO cc support@mattermost.com for one of the emails
```
marianunez, line 236: We want to copy support only on the 30th day email

jobs/cloud/worker.go
```diff
@@ -0,0 +1,284 @@
...
+		for admin := range sysAdmins {
+			worker.app.Srv().EmailService.SendOverUserLimitThirtyDayWarningEmail(sysAdmins[admin].Email, sysAdmins[admin].Locale, *worker.app.Config().ServiceSettings.SiteURL)
+		}
+	case 90:
+		// TODO cc support@mattermost.com for one of the emails
+		overLimitDate, _ := worker.app.Srv().Store.System().GetByName(model.OVER_USER_LIMIT_DATE)
+		for admin := range sysAdmins {
+			worker.app.Srv().EmailService.SendOverUserLimitNinetyDayWarningEmail(sysAdmins[admin].Email, sysAdmins[admin].Locale, *worker.app.Config().ServiceSettings.SiteURL, overLimitDate.Value)
+		}
+	case 91:
```
marianunez, line 268: As we discussed, this would never happen (unless they shut it down after the job ran by chance). Could we create a follow up ticket for the 91 day case?

jobs/cloud/worker.go
```diff
@@ -0,0 +1,284 @@
...
+		subCycleEndDate := time.Unix((subscription.EndAt / 1000), 0)
+		sysVar := &model.System{Name: model.USER_LIMIT_OVERAGE_CYCLE_END_DATE, Value: subCycleEndDate.Format(dateLayout)}
+		err := worker.app.Srv().Store.System().SaveOrUpdate(sysVar)
+		if err != nil {
+			mlog.Error("Unable to save USER_LIMIT_OVERAGE_CYCLE_END_DATE count", mlog.String("worker", worker.name), mlog.String("error", nErr.Error()))
+		}
+
+		// Store today's date (first day over limit) so we can reference it later
+		t := time.Now()
+		firstDayOverVar := &model.System{Name: model.OVER_USER_LIMIT_DATE, Value: t.Format(dateLayout)}
```
marianunez, line 200: Given that is now based on the actual date overdue (the billing cycle date), we shouldn't need this any longer.

jobs/schedulers.go
```diff
@@ -73,6 +73,11 @@ func (srv *JobServer) InitSchedulers() *Schedulers {
 		schedulers.schedulers = append(schedulers.schedulers, productNoticesInterface.MakeScheduler())
 	}
 
+	if cloudInterface := srv.Cloud; cloudInterface != nil {
+		mlog.Debug("SETTING CLOUD SCHEDULE")
```
marianunez: We can probably remove this now.

marianunez (changes requested):
Great stuff @nickmisasi! My only other comment would be to possibly split the logic in the `DoWork` of the job because it's a bit log there to follow

jobs/cloud/scheduler.go
```diff
@@ -0,0 +1,50 @@
...
+	return JobName + "Scheduler"
+}
+
+func (scheduler *Scheduler) JobType() string {
+	return model.JOB_TYPE_CLOUD
+}
+
+func (scheduler *Scheduler) Enabled(cfg *model.Config) bool {
+	// If CloudUserLimit is non-zero we're in a cloud installation
+	return *cfg.ExperimentalSettings.CloudUserLimit > 0 && cfg.FeatureFlags.CloudEmailJobsEnabledMM29999
```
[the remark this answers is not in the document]
nickmisasi: AFAICT, there is no way to access the license from the Config struct, so I don't think so, but might be unaware of something.

model/feature_flags.go
```diff
@@ -7,8 +7,12 @@ type FeatureFlags struct {
 	// Exists only for unit and manual testing.
 	// When set to a value, will be returned by the ping endpoint.
 	TestFeature string
+
+	// Toggle on and off scheduled jobs for cloud user limit emails
+	CloudEmailJobsEnabledMM29999 bool
```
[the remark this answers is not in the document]
nickmisasi: In the past when I've done feature flags I found it useful to include the ticket number that caused the flag to be created so that people can go back to Jira and get a bigger picture of what that feature flag might control

What are your thoughts on that?

jobs/cloud/worker.go
nickmisasi: At this point using your example, the `subCycleEndDate` would be October 31st. When we calculate the difference between the time.Now(), and the subCycleEndDate, we get the difference in days between the subCycleEndDate and time.Now(). If time.Now() is any date after Oct 31, we will have a positive number that can be used in the switch statement (when that number is 7, 14, 30, etc, it triggers a different case). If it's negative, it won't trigger a case, because that means we're still in the current cycle.


I could have done this by saving the next date of the check but figured it would be a bit cleaner to just calculate the difference each time.

jobs/cloud/worker.go
nickmisasi: Removed TODO from other places

jobs/cloud/worker.go
```diff
@@ -0,0 +1,284 @@
...
+		return
+	} else if !claimed {
+		return
+	}
+
+	cloudUserLimit := *worker.app.Config().ExperimentalSettings.CloudUserLimit
+
+	var subscription *model.Subscription
+	var subErr *model.AppError
+	if worker.app.Srv().License() != nil && *worker.app.Srv().License().Features.Cloud {
```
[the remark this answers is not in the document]
nickmisasi: I thought I had added this check to the logic but I guess not. Added, so should be good.

jobs/cloud/worker.go
```diff
@@ -0,0 +1,284 @@
...
+			return
+		}
+
+		if subscription == nil {
+			mlog.Error("Worker: Failed to retrieve subscription", mlog.String("worker", worker.name), mlog.String("job_id", job.Id))
+			return
+		}
+	}
+
+	if subscription != nil && subscription.IsPaidTier == "true" {
```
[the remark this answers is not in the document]
nickmisasi: Added check at top of function to make sure we aren't in a non-licensed or non-cloud instance before continuing

jobs/cloud/worker.go
nickmisasi: Yes, the EndAt is actually the `current_period_end` data from stripe, as can be seen here: https://github.com/mattermost/customer-web-server/blob/cce87a1762885e6f65eeb0db610b66c67b0b0c00/payment/stripe.go#L580

So it changes according the current cycle

jobs/cloud/worker.go
nickmisasi: removed the code instead of commenting. Ticket here: https://mattermost.atlassian.net/browse/MM-30406

jobs/cloud/worker.go
```diff
@@ -0,0 +1,284 @@
...
+		forgivenessCount = 0
+	} else {
+		forgivenessCount, err = strconv.Atoi(systemValue.Value)
+		if err != nil {
+			forgivenessCount = 0
+		}
+	}
+
+	if forgivenessCount >= 3 {
+		// There's no forgiving someone who's broken our hearts 3 or more times
```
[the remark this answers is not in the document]
nickmisasi: :)

imports/placeholder.go
```diff
@@ -21,4 +21,7 @@ import (
 
 	// This is a placeholder so this package can be imported in Team Edition when it will be otherwise empty.
 	_ "github.com/mattermost/mattermost-server/v5/jobs/product_notices"
+
+	// This is a placeholder so this package can be imported in Team Edition when it will be otherwise empty.
+	_ "github.com/mattermost/mattermost-server/v5/jobs/cloud"
```
ethervoid, line 26: This should be moved to the enterprise part

jobs/cloud/scheduler.go
```diff
@@ -0,0 +1,50 @@
...
+
+import (
+	"time"
+
+	"github.com/mattermost/mattermost-server/v5/app"
+	"github.com/mattermost/mattermost-server/v5/model"
+)
+
+const (
+	SchedFreqMinutes = 2 // 1 day
```
ethervoid, line 13: Is the comment accurate or we can get rid of it

jobs/cloud/scheduler.go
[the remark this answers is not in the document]
ethervoid: Is better to include the server in the scheduler struct instead of the app and use it to check the license too. If you check the line 45, you're using the server instead of the app. For this kind of things is better to depend on the server, instead of the app because the app is "request time object" kindof

ethervoid (commented):
Some comments to start addressing

jobs/cloud/scheduler.go
nickmisasi: It will be accurate, it's like this for testing. Will leave this unresolved so I don't forget to move it back. Trying to figure out how to get this going in a spinwick for Steve and Matt.

jobs/cloud/scheduler.go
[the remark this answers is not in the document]
nickmisasi: I didn't think to check the Scheduler. I've dropped the userlimit check in favour of a non-nil, cloud license + the feature flag here!

app/enterprise.go
```diff
@@ -96,6 +96,12 @@ func RegisterJobsActiveUsersInterface(f func(*App) tjobs.ActiveUsersJobInterface
 	jobsActiveUsersInterface = f
 }
 
+var jobsCloudInterface func(*App) tjobs.CloudJobInterface
```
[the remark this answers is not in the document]
nickmisasi: https://github.com/mattermost/enterprise/pull/821

app/email.go
```diff
@@ -633,3 +634,134 @@ func (es *EmailService) SendOverUserLimitWarningEmail(email string, locale strin
...
+
+	return true, nil
+}
+
+func (es *EmailService) SendOverUserFourteenDayWarningEmail(email string, locale string, siteURL string) (bool, *model.AppError) {
+	T := utils.GetUserTranslations(locale)
+
+	subject := T("api.templates.over_limit_14_days_subject")
+
+	bodyPage := es.newEmailTemplate("over_user_limit_7_days_body", locale)
```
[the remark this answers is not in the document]
nickmisasi: Was able to re-use this template for this 14 day email.

matthewbirtch (approved):
Emails look good with the changes we discussed over MM (font fix).

model/feature_flags.go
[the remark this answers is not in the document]
marianunez: I personally feel is a bit verbose to have on the actual variable name and I haven't seen this pattern in the server code base. I do think the ticket number is valuable so I would be more inclined to put it in the actual variable definition as a comment.

jobs/cloud/worker.go
marianunez: Gotcha! Thanks for the explanation!

app/email.go
```diff
@@ -633,3 +634,135 @@ func (es *EmailService) SendOverUserLimitWarningEmail(email string, locale strin
...
+	// Localization not needed
+
+	subject := fmt.Sprintf("Cloud Installation %s Scheduled Suspension", installationID)
+	bodyPage := es.newEmailTemplate("over_user_limit_support_body", "en")
+	bodyPage.Props["CustomerID"] = customerID
+	bodyPage.Props["SiteURL"] = siteURL
+	bodyPage.Props["SubscriptionID"] = subscriptionID
+	bodyPage.Props["InstallationID"] = installationID
+	bodyPage.Props["SuspensionDate"] = time.Now().AddDate(0, 0, 61).Format("2006-01-02")
+	bodyPage.Props["UserCount"] = userCount
```
marianunez, line 761: 👌

marianunez (changes requested):
Would like to keep consistency on the feature flag name, besides that, LGTM

model/feature_flags.go
```diff
@@ -7,8 +7,12 @@ type FeatureFlags struct {
...
 }
 
 func (f *FeatureFlags) SetDefaults() {
 	f.TestFeature = "off"
+	f.CloudEmailJobsEnabledMM29999 = false
```
marianunez, line 17: In your current tests you are overriding this in the env var?

nickmisasi: Yes, you can set it via MM_FEATUREFLAGS_<feature flag name> just like other configs

nickmisasi:
@marianunez changed to CloudDelinquentEmailJobsEnabled and left a comment pointing to the ticket.

Makefile
```diff
@@ -475,7 +475,7 @@ restart-client: | stop-client run-client ## Restarts the webapp.
 
 run-job-server: ## Runs the background job server.
 	@echo Running job server for development
-	$(GO) run $(GOFLAGS) -ldflags '$(LDFLAGS)' $(PLATFORM_FILES) jobserver --disableconfigwatch &
+	$(GO) run $(GOFLAGS) -ldflags '$(LDFLAGS)' $(PLATFORM_FILES) jobserver --disableconfigwatch
```
ethervoid, line 478: Why is this needed? Can affect the server workflow in some way?

app/email.go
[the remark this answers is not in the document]
ethervoid: Then is kind of misleading. We can rename that to `over_user_limit_7_or_14_days_body`

jobs/server.go
```diff
@@ -28,6 +28,7 @@ type JobServer struct {
 	ExpiryNotify            tjobs.ExpiryNotifyJobInterface
 	ProductNotices          tjobs.ProductNoticesJobInterface
 	ActiveUsers             tjobs.ActiveUsersJobInterface
+	Cloud                   tjobs.CloudJobInterface
```
ethervoid, line 31:
```suggestion
	Cloud                   ejobs.CloudJobInterface
```

ethervoid (commented):
Some comments

Makefile
nickmisasi: This made testing a lot easier (ctrl + c to stop instead of trying to kill the process by pid) but I'll add it back for master.

ethervoid (approved):
LGTM :)

jobs/cloud/scheduler.go
nickmisasi: https://github.com/mattermost/enterprise/pull/821/commits/e1fe5f96fd1edd9d14dab6142e8cd846f3b331fa set to 1440, which is 1 day in minutes.

nickmisasi: Swapped to 720 minutes, so that it is guaranteed to happen once a day

stevemudie (approved):
Looks good and worked as expected. Thanks @nickmisasi !

nickmisasi:
Release notes added

marianunez (approved):
Great work! Thanks @nickmisasi!

model/system.go
```diff
@@ -33,6 +33,9 @@ const (
 	SYSTEM_WARN_METRIC_LAST_RUN_TIMESTAMP_KEY     = "LastWarnMetricRunTimestamp"
 	AWS_METERING_REPORT_INTERVAL                  = 1
 	AWS_METERING_DIMENSION_USAGE_HRS              = "UsageHrs"
+	USER_LIMIT_OVERAGE_CYCLE_END_DATE             = "UserLimitOverageCycleEndDate"
+	OVER_USER_LIMIT_DATE                          = "OverUserLimitDate"
```
marianunez, line 37: Do we use this? I think we removed this date stored

nickmisasi: You're right, we did. Removed.

PR #16225, opened 2020-11-06
MM-20897 Add category muting

hmhealey:
A new sidebar feature is muting entire categories at once. Muting a category mutes all channels in it and makes it so that channels moved into it get automatically muted.

Updating the ChannelMembers one-by-one to mute them after muting the category is a bit weird, but we don't have bulk websocket events for updating multiple ChannelMembers at once, and I was worried about mobile compatibility here. Unfortunately, we don't have any way to change the potential websocket events based on client versions, but that would be really nice for something like this.

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-20897

#### Related Pull Requests
https://github.com/mattermost/mattermost-redux/pull/1286
https://github.com/mattermost/mattermost-webapp/pull/7033

#### Release Note
```release-note
Added the ability to mute categories in the new sidebar
```

### DIFF
--- a/app/channel.go
+++ b/app/channel.go
@@ -1097,22 +1097,7 @@ func (a *App) UpdateChannelMemberRoles(channelId string, userId string, newRoles
 
 	member.ExplicitRoles = strings.Join(newExplicitRoles, " ")
 
-	member, nErr := a.Srv().Store.Channel().UpdateMember(member)
-	if nErr != nil {
-		var appErr *model.AppError
-		var nfErr *store.ErrNotFound
-		switch {
-		case errors.As(nErr, &appErr):
-			return nil, appErr
-		case errors.As(nErr, &nfErr):
-			return nil, model.NewAppError("UpdateChannelMemberRoles", MISSING_CHANNEL_MEMBER_ERROR, nil, nfErr.Error(), http.StatusNotFound)
-		default:
-			return nil, model.NewAppError("UpdateChannelMemberRoles", "app.channel.get_member.app_error", nil, nErr.Error(), http.StatusInternalServerError)
-		}
-	}
-
-	a.InvalidateCacheForUser(userId)
-	return member, nil
+	return a.updateChannelMember(member)
 }
 
 func (a *App) UpdateChannelMemberSchemeRoles(channelId string, userId string, isSchemeGuest bool, isSchemeUser bool, isSchemeAdmin bool) (*model.ChannelMember, *model.AppError) {
@@ -1134,27 +1119,7 @@ func (a *App) UpdateChannelMemberSchemeRoles(channelId string, userId string, is
 		member.ExplicitRoles = RemoveRoles([]string{model.CHANNEL_GUEST_ROLE_ID, model.CHANNEL_USER_ROLE_ID, model.CHANNEL_ADMIN_ROLE_ID}, member.ExplicitRoles)
 	}
 
-	member, nErr := a.Srv().Store.Channel().UpdateMember(member)
-	if nErr != nil {
-		var appErr *model.AppError
-		var nfErr *store.ErrNotFound
-		switch {
-		case errors.As(nErr, &appErr):
-			return nil, appErr
-		case errors.As(nErr, &nfErr):
-			return nil, model.NewAppError("UpdateChannelMemberSchemeRoles", MISSING_CHANNEL_MEMBER_ERROR, nil, nfErr.Error(), http.StatusNotFound)
-		default:
-			return nil, model.NewAppError("UpdateChannelMemberSchemeRoles", "app.channel.get_member.app_error", nil, nErr.Error(), http.StatusInternalServerError)
-		}
-	}
-
-	// Notify the clients that the member notify props changed
-	message := model.NewWebSocketEvent(model.WEBSOCKET_EVENT_CHANNEL_MEMBER_UPDATED, "", "", userId, nil)
-	message.Add("channelMember", member.ToJson())
-	a.Publish(message)
-
-	a.InvalidateCacheForUser(userId)
-	return member, nil
+	return a.updateChannelMember(member)
 }
 
 func (a *App) UpdateChannelMemberNotifyProps(data map[string]string, channelId string, userId string) (*model.ChannelMember, *model.AppError) {
@@ -1185,6 +1150,17 @@ func (a *App) UpdateChannelMemberNotifyProps(data map[string]string, channelId s
 		member.NotifyProps[model.IGNORE_CHANNEL_MENTIONS_NOTIFY_PROP] = ignoreChannelMentions
 	}
 
+	member, err = a.updateChannelMember(member)
+	if err != nil {
+		return nil, err
+	}
+
+	a.invalidateCacheForChannelMembersNotifyProps(member.ChannelId)
+
+	return member, nil
+}
+
+func (a *App) updateChannelMember(member *model.ChannelMember) (*model.ChannelMember, *model.AppError) {
 	member, nErr := a.Srv().Store.Channel().UpdateMember(member)
 	if nErr != nil {
 		var appErr *model.AppError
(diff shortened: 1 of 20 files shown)


app/channel_category.go
```diff
@@ -144,14 +144,125 @@ func (a *App) UpdateSidebarCategoryOrder(userId, teamId string, categoryOrder []
...
+			} else if !toCategory.Muted && fromCategory.Muted {
+				channelsToUnmute = append(channelsToUnmute, channelId)
+			}
+		}
+	}
+
+	for _, channelId := range channelsToMute {
+		_, err := a.setChannelMuted(channelId, userId, true)
+		if err != nil {
+			mlog.Error("Failed to mute channel to match category", mlog.String("channel_id", channelId), mlog.String("user_id", userId))
```
reflog, line 205: maybe include the original 'err' in the log?

app/channel_category.go
```diff
@@ -144,14 +144,125 @@ func (a *App) UpdateSidebarCategoryOrder(userId, teamId string, categoryOrder []
...
+		}
+	}
+
+	for _, channelId := range channelsToUnmute {
+		_, err := a.setChannelMuted(channelId, userId, false)
+		if err != nil {
+			mlog.Error("Failed to unmute channel to match category", mlog.String("channel_id", channelId), mlog.String("user_id", userId))
```
reflog, line 212: maybe include the original 'err' in the log?

reflog (approved):
looks great!

app/channel_category.go
hmhealey: Good idea. Will do.

hmhealey:
I'm pretty sure this build is failing due to some type issues on the corresponding webapp PR. I'll re-run it once those are resolved.

app/channel_category.go
```diff
@@ -144,14 +144,135 @@ func (a *App) UpdateSidebarCategoryOrder(userId, teamId string, categoryOrder []
...
+			if toCategory.Muted && !fromCategory.Muted {
+				channelsToMute = append(channelsToMute, channelId)
+			} else if !toCategory.Muted && fromCategory.Muted {
+				channelsToUnmute = append(channelsToUnmute, channelId)
+			}
+		}
+	}
+
+	for _, channelId := range channelsToMute {
+		_, err := a.setChannelMuted(channelId, userId, true)
```
agnivade, line 203: I realize there are no websocket bulk updates, but I think we can still do the DB updates in bulk. And split out the websocket updates to a separate step.

- Get all channel members from a set of channels - similar to what we have `ChannelStore.(GetMembersByIds)`, we can have `ChannelStore.(GetMembersByChannels)`.
- Toggle the mute state in a loop
- And we already have `ChannelStore.(UpdateMultipleMembers)`!
- And then finally broadcast the ws message in a loop and invalidate cache.

That reduces the number of DB queries to 2 for every set of muted and unmuted channels. So 4 in total, instead of n.

app/channel_category.go
```diff
@@ -144,14 +144,135 @@ func (a *App) UpdateSidebarCategoryOrder(userId, teamId string, categoryOrder []
...
+	return updatedCategories, nil
+}
+
+func (a *App) muteChannelsForUpdatedCategories(userId string, updatedCategories []*model.SidebarCategoryWithChannels, originalCategories []*model.SidebarCategoryWithChannels) {
+	var channelsToMute []string
+	var channelsToUnmute []string
+
+	// Mute or unmute all channels in categories that were muted or unmuted
+	for i, updatedCategory := range updatedCategories {
+		originalCategory := originalCategories[i]
```
agnivade, line 166: I'll prefer a length check somewhere that verifies length of both slices are same, rather than just directly accessing it.

app/channel_category.go
```diff
@@ -144,14 +144,135 @@ func (a *App) UpdateSidebarCategoryOrder(userId, teamId string, categoryOrder []
...
+
+		if originalCategoryId != updatedCategoryId && updatedCategoryId != "" {
+			channelsDiff[channelId] = &categoryChannelDiff{originalCategoryId, updatedCategoryId}
+		}
+	}
+
+	return channelsDiff
+}
+
+func getChannelIdsRemovedFromCategory(originalCategory *model.SidebarCategoryWithChannels, updatedCategory *model.SidebarCategoryWithChannels) []string {
```
agnivade, line 262: This function is not used anywhere. Any reason why this is kept?

app/channel_category.go
```diff
@@ -144,14 +144,135 @@ func (a *App) UpdateSidebarCategoryOrder(userId, teamId string, categoryOrder []
...
+			}
+		}
+
+		return result
+	}
+
+	updatedChannelIdsMap := mapChannelIdsToCategories(updatedCategories)
+	originalChannelIdsMap := mapChannelIdsToCategories(originalCategories)
+
+	// Check for any channels that have changed categories. Note that we don't worry about any channels that have moved
```
agnivade, line 248: > Note that we don't worry about any channels that have moved outside of these categories

Curious to know why is this. Is it because the webapp always correctly uses the API? Or is there another reason?

hmhealey: Yeah, when moving a channel from category A to B, the webapp will always call the API with both categories, but you could theoretically move the channel by just updating B with the new ID. That case will fail to mute the channel, but checking that would require extra database hits to find which category the channel was originally in before doing the updated, so keeping track of that got really complicated.

Since it should work with how we currently use the API, I wasn't too worried about it working incorrectly in an edge case at the moment. I left the note to explain that in case someone ever runs into it in the future.

app/channel_category.go
hmhealey: Yep, I can add that for safety. They should be the same length, but "should" can be dangerous

app/channel_category.go
hmhealey: Oh yeah, I'll remove that. I meant to use it, but then it got replaced by `diffChannelsBetweenCategories`

app/channel_category.go
hmhealey: Yeah, I can take a look at doing that. It probably negates some of the other cleanup I did elsewhere, but it'll at least be ready if we add bulk websocket events in the future.

app/channel.go
```diff
@@ -2788,20 +2765,65 @@ func (a *App) GetPinnedPosts(channelId string) (*model.PostList, *model.AppError
...
+	member.SetChannelMuted(muted)
+
+	member, err := a.updateChannelMember(member)
+	if err != nil {
+		return nil, err
+	}
+
+	a.invalidateCacheForChannelMembersNotifyProps(member.ChannelId)
+
+	return a.updateChannelMember(member)
```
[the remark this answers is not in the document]
hmhealey: Nope, that's a copy-paste error

app/channel_category.go
hmhealey: Updated. I didn't realize we had an UpdateMultipleMembers store method since that tends to involve one database hit per item anyway.

agnivade: > if we add bulk websocket events in the future.

Yep, this is something I plan to investigate. But this also needs support from the client side to really "understand" a bulk event.

agnivade (approved):
Terrific piece of work. Really appreciate the amount of tests written!

hmhealey:
@jgilliam17 I'm going to see if we can get a test server using both the webapp and server branches. If not, we'll have to merge this one and test on the webapp branch

hmhealey:
@jgilliam17 The PR should be testable here with both the server and web components. The web component is missing a few changes from master, but it should hopefully be correct for everything here

jgilliam17:
Thanks @hmhealey 
Just noticed if I mute `Channels` category and I'm added to a channel by another user - new channel is not immediately muted as it's added to the LHS Channels category. Other channels are muted at drop when added to already muted category. Should we treat this use case differently? 0/5 on this

esethna:
@jgilliam17 I think this behaviour is okay as is. I'd worry if we auto-muted channels I'm added to by other users we would miss messages.

jgilliam17 (approved):
Thank you @hmhealey 
Tested, looks good to merge.
- Verified category muting/unmuting. Verified channels and DMs are muted when dropped into muted category and unmuted when moved to an unmuted category.
- Verified when Channels category is muted, new channels user is added to are not muted.

hmhealey:
Good catch. I didn't even think of that case. Glad to know we accidentally ended up with the correct behaviour.

PR #16230, opened 2020-11-07
Soft fail on cloud user limit warnings when creating user

jwilander:
#### Summary
This is causing an issue on cloud where if this check fails then the CWS does not complete workspace setup since it thinks the user failed to create when in fact the user was created successfully.

#### Release Note

```release-note
None
```


angeloskyratzakos (approved):

gabrieljackson (approved):
👍

stylianosrigas (approved):
LGTM

nickmisasi (approved):

PR #16234, opened 2020-11-08
MM-30048 added thread related socket messages

reflog:
#### Summary

Added 3 new socket events for thread update/read/follow

#### Ticket Link

https://mattermost.atlassian.net/browse/MM-30048
#### Release Note

```release-note
New websocket events added: thread_updated,thread_follow_changed,thread_read_changed
```


app/notification.go
```diff
@@ -160,24 +160,31 @@ func (a *App) SendNotifications(post *model.Post, team *model.Team, channel *mod
 	mentionedUsersList := make([]string, 0, len(mentions.Mentions))
 	updateMentionChans := []chan *model.AppError{}
 	mentionAutofollowChans := []chan *model.AppError{}
+	threadParticipants := []string{post.UserId}
+	if *a.Config().ServiceSettings.ThreadAutoFollow && post.RootId != "" {
+		if parentPostList != nil {
+			threadParticipants = append(threadParticipants, parentPostList.Posts[parentPostList.Order[0]].UserId)
```
jespino, line 166: Are you fixing here the "autofollow" for the root post creator"?

reflog: @jespino  yup! somehow missed it in the previous PR

store/sqlstore/post_store.go
```diff
@@ -1976,7 +1976,7 @@ func (s *SqlPostStore) updateThreadsFromPosts(transaction *gorp.Transaction, pos
 		if thread, found := threadByRoot[rootId]; !found {
 			// calculate participants
 			var participants model.StringArray
-			if _, err := transaction.Select(&participants, "SELECT DISTINCT UserId FROM Posts WHERE RootId=:RootId", map[string]interface{}{"RootId": rootId}); err != nil {
+			if _, err := transaction.Select(&participants, "SELECT DISTINCT UserId FROM Posts WHERE RootId=:RootId OR Id=:RootId", map[string]interface{}{"RootId": rootId}); err != nil {
```
jespino, line 1979: Fixing something else here too, right? I think maybe this makes sense to be backported to 5.29

reflog: @jespino  yup. same issue. the root posts author wasn't included in the participants

reflog: once we are done with review, I'll create a backport PR for just these changes

api4/user_test.go
```diff
@@ -5401,6 +5401,70 @@ func TestGetThreadsForUser(t *testing.T) {
...
+					require.Contains(t, thread.Participants, th.BasicUser.Id)
+					require.Contains(t, thread.Participants, th.BasicUser2.Id)
+				}
+			case <-time.After(3 * time.Second):
+				return
+			}
+		}
+	}()
+
+	require.Truef(t, userHasReceivedFollow, "User should have received %s event", model.WEBSOCKET_EVENT_THREAD_FOLLOW_CHANGED)
```
jespino, line 5463: Can we make this more granular? something like "Call one thing, check the events sent by that things, reset the values, call another thing, check the events send by that second thing, etcetera"?

jespino (changes requested):
Some small suggestions and questions, but in general looks good to me.

store/sqlstore/post_store.go
jespino: Great, that sounds good to me :)

api4/user_test.go
reflog: @jespino - I guess, but... what for? :) It's a rather clear test the way it is right now, what would splitting into small subcases clarify?

jespino: Not necesarily clarifies, but avoids possible bugs generating events when are not expected, or not generating events on certain actions. It is not the same to say "This 3 actions emits this 3 events, than saying this action emit this event, this other action emits this other event and so on". This is not necessarily wrong, but is more error prone and can give more false negatives on the tests that a more specific approach.

reflog: Mkay. Noted :+1:

app/notification.go
```diff
@@ -403,6 +410,21 @@ func (a *App) SendNotifications(post *model.Post, team *model.Team, channel *mod
 	}
 
 	a.Publish(message)
+	// If this is a reply in a thread, notify participants
+	if *a.Config().ServiceSettings.ThreadAutoFollow && post.RootId != "" {
```
ashishbhate, line 414: Why only if `ThreadAutoFollow` is true? If this is false, and a user is following the thread, then will they not get a notification? 
And conversely, a user who participated in the thread but then unfollowed the thread will be notified?

reflog: You are absolutely right @ashishbhate . I forgot to introduce the new CollapsedThreads config setting.

ashishbhate (approved):

reflog:
@jespino - anything blocking this?

jespino:
@reflog not really, only finding the time to review it properly.

jespino (approved):
LGTM

reflog:
@ogi-m - some basic smoke testing around posts could be nice, just to check that I didn't break anything unexpected

ogi-m (approved):
Done smoke testing around posts, everything's working as expected. LGTM 👍

PR #16235, opened 2020-11-08
MM-27312 Allow uploading 10 files per post

DigasNikas:
**Summary**
Increase the number of files allowed per post from 5 to 10.
The assigned error message was also updated in several languages.

**Ticket Link**
Fixes https://github.com/mattermost/mattermost-server/issues/15645
JIRA ticket: https://mattermost.atlassian.net/browse/MM-27312

**Release Note**
```release-note
Changed number of files allowed per post, from 5 to 10.
```


hmhealey (changes requested):
I liked your proposal on the related issue, so would you mid having two constants, and checking them both when we're checking for the max number of FileIds allowed.

We're also going to need to increase the size of the database column in which Post.FileIds is stored. That needs to be done both when a new database is created (look at `newSqlPostStore` in `store/sqlstore/post_store.go`) and when doing an upgrade to 5.30 (`store/sqlstore/upgrade.go`).

jasonblais:
@DigasNikas thank you for your contribution! Let us know if you have any questions about the feedback shared above? We'd be happy to help :)

DigasNikas:
Addressed @hmhealey request and suggestion. Should be good now @jasonblais

jasonblais:
@DigasNikas Thank you! Looks like there are some test failures reported in the build checks, would you be open to help take a look?

DigasNikas:
@jasonblais could you retrigger the test-schema pipeline? It seems to have failed when pulling the CI docker image but I don't have permissions to re-run it.

jasonblais:
@DigasNikas Ah, thank you! Me neither -- @hmhealey and @streamer45 when reviewing the PR, please also help retrigger the test-schema pipeline? This should help resolve the test failures reported above

store/sqlstore/upgrade.go
```diff
@@ -341,7 +341,7 @@ func upgradeDatabaseToVersion35(sqlSupplier *SqlSupplier) {
 		sqlSupplier.GetMaster().Exec("UPDATE ChannelMembers SET Roles = 'channel_user channel_admin' WHERE Roles = 'admin'")
 
 		// The rest of the migration from Filenames -> FileIds is done lazily in api.GetFileInfosForPost
-		sqlSupplier.CreateColumnIfNotExists("Posts", "FileIds", "varchar(150)", "varchar(150)", "[]")
+		sqlSupplier.CreateColumnIfNotExists("Posts", "FileIds", "varchar(300)", "varchar(300)", "[]")
```
streamer45, line 344: This line should be kept as is since it's about a previous version. Our DB schema migration system works by defining a function for each new version. 

What we want to do is add a line performing the type change in the function for the next version, currently `upgradeDatabaseToVersion530`:

```go
sqlSupplier.AlterColumnTypeIfExists("Posts", "FileIds", "text", "varchar(300)")
```

There's a small caveat here which is we need to define a `text` type for MySQL instead of `varchar(300)`. This is because our underlying ORM `gorp` implicitly converts any size above 255 as a `text` on MySQL.

scripts/mattermost-postgresql-5.0.sql
```diff
@@ -398,7 +398,7 @@ CREATE TABLE public.posts (
     props character varying(8000),
     hashtags character varying(1000),
     filenames character varying(4000),
-    fileids character varying(150),
+    fileids character varying(300),
```
streamer45, line 401: It's not strictly needed but since we are here we could also update `scripts/mattermost-mysql-5.0.sql`.

streamer45 (changes requested):
Thank you @DigasNikas! It's looking good :+1: 
We just need to fix the migration code.

store/sqlstore/upgrade.go
DigasNikas: My bad :sweat_smile: should have scrolled a bit longer. 
Thanks for the introduction to Mattermost DB schema migration :) should be good now.

streamer45 (approved):
Thank you @DigasNikas, looks great :+1:

hmhealey (approved):
Excellent. Looks good to me!

streamer45:
@amyblais 

This PR has changes in the database schema that will trigger during migration to 5.30.
If we could get this in for the 5.30 release it would be ideal since PR is of low risk. Otherwise, to avoid possible issues, we would need to make further changes and add the needed migration code to the next version.

hmhealey:
At this point, we might want to update them to happen in 5.31 since the cloud version that'll become 5.30 is supposed to go out tomorrow, I believe.

amyblais:
@streamer45 We're planning a cloud release tomorrow and v5.30 release will be based on that https://community-release.mattermost.com/core/pl/qwq65andnibdtjwk5e1tjftiur. No new features are added to v5.30 anymore.

streamer45:
Fair enough. Adding the `Do not merge` label then.

DigasNikas:
Just to mention: Changes in places like [this](https://github.com/mattermost/mattermost-webapp/blob/master/utils/constants.jsx#L964) are still needed before the user can submit 10 files. Not sure if they fit the context of this ticket.

hmhealey:
@DigasNikas Good catch! If you have a webapp dev environment set up, would you mind submitting a PR to https://github.com/mattermost/mattermost-webapp for that? If the branch has the same name, I think we should be able to test both PRs together. I suppose the mobile app will also need the corresponding change as well, but we can file a separate ticket for that since it's more complicated to set up.

Also, could you also update this branch to resolve the merge conflict?

DigasNikas:
@hmhealey Conflicts resolved and Webapp PR created [here](https://github.com/mattermost/mattermost-webapp/pull/7160).

hmhealey:
Thanks for those. The PR is conflicting again, so I'm going to resolve the conflicts again by reverting the changes to that one translation file. The translation server will automatically pick up that it needs changes once this is in.

@lindalumitchell To test this, we'll need to either get the test server using both this and the webapp PR, or we'll need to wait until the webapp PR is merged as well. It's definitely possible to get the test server to have the changes from both PRs, but it hasn't been particularly reliable for me.

lindalumitchell (approved):
Thanks @hmhealey, I'm fine deferring QA on this until after merge. We'll have time to test before the next cloud release.

store/sqlstore/upgrade.go
```diff
@@ -19,7 +19,8 @@ import (
 )
 
 const (
-	CURRENT_SCHEMA_VERSION   = VERSION_5_30_0
+	CURRENT_SCHEMA_VERSION   = VERSION_5_31_0
+	VERSION_5_31_0           = "5.30.1"
```
streamer45, line 23:
```suggestion
	VERSION_5_31_0           = "5.31.0"
```

store/sqlstore/upgrade.go
```diff
@@ -949,3 +951,11 @@ func upgradeDatabaseToVersion530(sqlStore *SqlStore) {
 		saveSchemaVersion(sqlStore, VERSION_5_30_0)
 	}
 }
+
+func upgradeDatabaseToVersion531(sqlStore *SqlStore) {
+	if shouldPerformUpgrade(sqlStore, VERSION_5_30_0, VERSION_5_31_0) {
+		// allow 10 files per post
+		sqlStore.AlterColumnTypeIfExists("Posts", "FileIds", "text", "varchar(300)")
+		saveSchemaVersion(sqlStore, VERSION_5_31_0)
+	}
```
streamer45, line 960: We should comment these lines until we release next version:

```suggestion
	// if shouldPerformUpgrade(sqlStore, VERSION_5_30_0, VERSION_5_31_0) {
		// allow 10 files per post
		sqlStore.AlterColumnTypeIfExists("Posts", "FileIds", "text", "varchar(300)")
		// saveSchemaVersion(sqlStore, VERSION_5_31_0)
	// }
```

streamer45 (changes requested):
Great work @DigasNikas :+1: We just need to change a couple of things related to the store upgrade.

streamer45 (approved):

josephbaylon:
@amyblais This looks like this should be for v5.31. Should the milestone be the same?

amyblais:
This didn't make it to last week's cloud release so it will be in v5.32. @streamer45 @hmhealey Can you help update the version?

hmhealey:
I think just the mobile PR needs to be updated, but Joseph commented on that already, so I think we're good there.

streamer45:
Discussed with @hmhealey internally and we agreed we should bump the migration code to 5.32 to avoid possible issues during upgrade. I'll send a PR to fix that.

PR #16239, opened 2020-11-09
MM-29107: Fix race in LDAP login

agnivade:
We remove the goroutine to make things synchronous.

This removes the race and makes things more reliable.
It is already tested by TestLogin in ldap_test.go in -race mode.

https://mattermost.atlassian.net/browse/MM-29107

```release-note
NONE
```


app/login.go
```diff
@@ -217,9 +217,7 @@ func (a *App) DoLogin(w http.ResponseWriter, r *http.Request, user *model.User,
 	a.SetSession(session)
 
 	if a.Srv().License() != nil && *a.Srv().License().Features.LDAP && a.Ldap() != nil {
-		a.Srv().Go(func() {
-			a.Ldap().UpdateProfilePictureIfNecessary(*user, session)
-		})
+		a.Ldap().UpdateProfilePictureIfNecessary(*user, session)
```
streamer45, line 220: Was dereferencing the pointer (make a shallow copy of user) outside the closure (just before calling the goroutine) an option?

My only concern here, and possibly the reason this was added, is that we could significantly delay the login process. Not sure about the internals of `UpdateProfilePictureIfNecessary()` but it looks like it involves some re-try mechanism using sleep which could indeed make things sluggish.

agnivade: Yes, that's an option. But it updates the profile picture only if they have changed it. So in the happy path, not every login would be delayed. Only the ones where somebody has changed the picture would get delayed. That's why I thought it was a decent tradeoff.

streamer45: Alright. I'll then defer to @mkraft on that front since he surely has more context.

Regarding this change, do we still need to send a copy of user or just being extra safe?

agnivade: We technically don't. But that needs an API change. I don't mind either ways. If Martin is fine with the change, I can do it.

mkraft: These are the facts as I see them: *every* login will necessarily be slowed down for *all* LDAP-authenticated users on systems that have a value set for the `LdapSettings.PictureAttribute` configuration. These are the calls they'll be waiting for:

1. A call to LDAP from `(*LdapInterfaceImpl).getLdapPicture`
2. If there's a profile picture in LDAP then a call to the database from `(*App) GetProfileImage`
3. If the picture has changed then a call to the database from `(*App).SetProfileImageFromFile`

Whether that slowdown is acceptable depends on how much slower it is and what the dev. effort is to come up with an alternative. I would defer to @sbishel as he's the subject matter expert on LDAP profile photo syncing.

agnivade: Yes, the happy path has 2 extra calls. And 3 if the picture has changed.

I don't have any objections towards changing it to use the goroutine. But I mainly wanted to understand whether the decision to do this in a goroutine was done because a measurable lag was noticed due to 2 extra calls, or just the fact that this does not need to be synchronous.

If there is indeed a noticeable delay with the 2 extra calls - one alternative is to have a buffered channel with a single consumer and let `UpdateProfilePictureIfNecessary` just push to that channel (In a separate ticket of course). Spinning up goroutines in an unbounded manner is generally considered a bad practice.

mkraft (approved):

agnivade:
@sbishel - gentle ping

app/login.go
sbishel: Sorry, I was out last week and finally am catching up. @agnivade Thank you for finding and investigating this.

It was put into a goroutine because we are calling an LDAP server and retrieving an image. Although in my testing this has never really taken a significant amount of time, we were trying to protect the login from delay and/or failure. So, it wasn't because of a measurable lag, but the **potential** for a measurable lag. (Ie, large image file being downloaded or LDAP Server response lag). 

That said, the LDAP server must be up and working to have gotten this far, so probably the biggest risk of delay is the size of the image, which hopefully an LDAP server would throttle. I would prefer (2/5) it stays in a go routine to protect against unexpected delays, but killed my best arguments.

sbishel: I just ran some tests against an ADFS Server pulling down a small image. It added about 250 (no update) - 350 (update) ms to the login time. However, if this moves to synchronous, the code in the Enterprise branch that waits for the session to be registered needs to be removed. This ends up waiting and timing out since the session never gets registered. I'm probably moved to a 3/5 for keeping a go routine.

agnivade: Thanks for the context there @sbishel.

> So, it wasn't because of a measurable lag, but the potential for a measurable lag. (Ie, large image file being downloaded or LDAP Server response lag).

Understood. My point was to "prove" that there is a measurable lag, and get some data around how frequently do the users actually change their profile pictures. Is it every day? Once a week? Once a month? Based on that, the optimization decisions can matter a lot.

> I would prefer (2/5) it stays in a go routine to protect against unexpected delays, 

On the contrary, if there is a swarm of logins, this might even kill the LDAP server or even the MM server. That's the tradeoff with spinning unbounded goroutines.

> However, if this moves to synchronous, the code in the Enterprise branch that waits for the session to be registered needs to be removed. This ends up waiting and timing out since the session never gets registered. 

So it looks like we have moved the problem statement to a different one. This isn't about slowness anymore but that the session doesn't get registered until we make a ws connection after login.

This is a solid reason for asynchronous processing, not for "performance" improvements, but because the code needs it to be. I will move to a goroutine for now. But to really make the code robust against pathological scenarios (which almost always happens with large customers), we should have a long-running goroutine with a buffered channel.

sbishel (approved):

streamer45 (approved):
Looks good :+1:

agnivade:
Since this just fixes a race condition, no QA needed for this.

PR #16256, opened 2020-11-11
MM-29486: Fix racy test TestPatchChannelModeration

agnivade:
- We avoid appending to the slice in `(r *Role) IsValidWithoutId`
by just iterating the 2 slices separately.
- We pass deep copies of channels to prevent racy modification.

https://mattermost.atlassian.net/browse/MM-29486

```release-note
NONE
```

### DIFF
--- a/model/role.go
+++ b/model/role.go
@@ -484,12 +484,16 @@ func (r *Role) IsValidWithoutId() bool {
 
 	for _, permission := range r.Permissions {
 		permissionValidated := false
-		for _, p := range append(AllPermissions, DeprecatedPermissions...) {
-			if permission == p.Id {
-				permissionValidated = true
-				break
+		check := func(perms []*Permission) {
+			for _, p := range perms {
+				if permission == p.Id {
+					permissionValidated = true
+					break
+				}
 			}
 		}
+		check(AllPermissions)
+		check(DeprecatedPermissions)
 
 		if !permissionValidated {
 			return false
--- a/model/role.go
+++ b/model/role.go
@@ -482,21 +482,16 @@ func (r *Role) IsValidWithoutId() bool {
 		return false
 	}
 
-	for _, permission := range r.Permissions {
-		permissionValidated := false
-		check := func(perms []*Permission, permission string) bool {
-			for _, p := range perms {
-				if permission == p.Id {
-					return true
-				}
+	check := func(perms []*Permission, permission string) bool {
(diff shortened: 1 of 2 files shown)


model/role.go
```diff
@@ -484,12 +484,16 @@ func (r *Role) IsValidWithoutId() bool {
...
+		check := func(perms []*Permission, permission string) {
+			for _, p := range perms {
+				if permission == p.Id {
+					permissionValidated = true
+					break
+				}
 			}
 		}
+		check(AllPermissions, permission)
+		check(DeprecatedPermissions, permission)
```
streamer45, line 496: Maybe we could do the second check only if `permissionValidated` is `false`. If we make `check()` return a `bool` it might even simplify things further.

wiggin77 (approved):
I think streamer45's suggestion makes sense.  LGTM.

model/role.go
```diff
@@ -484,11 +484,17 @@ func (r *Role) IsValidWithoutId() bool {
 
 	for _, permission := range r.Permissions {
 		permissionValidated := false
-		for _, p := range append(AllPermissions, DeprecatedPermissions...) {
-			if permission == p.Id {
-				permissionValidated = true
-				break
+		check := func(perms []*Permission, permission string) bool {
```
streamer45, line 487: Wondering if having the check function defined just outside the loop would make things even more readable.

model/role.go
```diff
@@ -484,11 +484,17 @@ func (r *Role) IsValidWithoutId() bool {
...
+			for _, p := range perms {
+				if permission == p.Id {
+					return true
+				}
 			}
+			return false
+		}
+		permissionValidated = check(AllPermissions, permission)
+		if !permissionValidated {
+			permissionValidated = check(DeprecatedPermissions, permission)
```
streamer45, line 497: Would it be too much to exploit short circuiting and doing:

```go
permissionValidated := check(AllPermissions, permission) ||
   check(DeprecatedPermissions, permission)
```

That's why I originally suggested the `bool` return actually but I totally understand if it's just me that finds that preferable.

streamer45 (approved):
Looks good :+1: 
Left two very minor comments cause I just like to chat code sometimes.

model/role.go
agnivade: Nice. My brain cannot come up with anything more complicated than simple if conditions. :rofl:

PR #16257, opened 2020-11-11
MM-29488: Fix racy test TestScheduler

agnivade:
We iterate the nextRunTimes slice only after stopping the
scheduler.

And since we are checking the slice 3 times, we split
the test into 3 sub-tests, every time starting a new scheduler
and running the same steps again.

https://mattermost.atlassian.net/browse/MM-29488

```release-note
NONE
```

### DIFF
--- a/jobs/schedulers.go
+++ b/jobs/schedulers.go
@@ -197,7 +197,7 @@ func (schedulers *Schedulers) scheduleJob(cfg *model.Config, scheduler model.Sch
 	return scheduler.ScheduleJob(cfg, pendingJobs, lastSuccessfulJob)
 }
 
-func (schedulers *Schedulers) handleConfigChange(oldConfig *model.Config, newConfig *model.Config) {
+func (schedulers *Schedulers) handleConfigChange(_ *model.Config, newConfig *model.Config) {
 	mlog.Debug("Schedulers received config change.")
 	schedulers.configChanged <- newConfig
 }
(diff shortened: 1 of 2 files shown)


streamer45 (approved):
Looks good :+1:

jobs/schedulers.go
```diff
@@ -197,7 +197,7 @@ func (schedulers *Schedulers) scheduleJob(cfg *model.Config, scheduler model.Sch
 	return scheduler.ScheduleJob(cfg, pendingJobs, lastSuccessfulJob)
 }
 
-func (schedulers *Schedulers) handleConfigChange(oldConfig *model.Config, newConfig *model.Config) {
+func (schedulers *Schedulers) handleConfigChange(_ *model.Config, newConfig *model.Config) {
```
wiggin77, line 200: Have we decided to start doing this?  I feel like it reduces the documentation of the method just to save a pointer copy.  In this case a reader might guess the first arg was a pointer to the old config, but in other methods it is not clear and someone may have to lookup the interface just to know what that arg was.  Also I think this is something to be left for the compiler to optimize.

wiggin77 (approved):
LGTM :+1:

jobs/schedulers.go
agnivade: Oh it's not a performance optimization at all. This is just to clarify that the first argument is not used at all. So we don't pass what we don't use.

It's interesting that you see that it reduces the documentation. I can change it.

PR #16258, opened 2020-11-11
[MM-30136] - [GH-16253]  Issuing /dnd consecutively should not toggle the user status between …

OgmaJ:
#### Summary

/dnd will only do a do not disturb 

#### Ticket Link
Fixes https://github.com/mattermost/mattermost-server/issues/16253
JIRA https://mattermost.atlassian.net/browse/MM-30136

#### Release Note
```release-note
 /dnd command will ONLY set a do not disturb status whatever the  current user status is 
```


hmhealey (changes requested):
Thanks for the PR!

Since we're no longer going to use the `api.command_dnd.error` and `api.command_dnd.disabled` translation strings, could you remove those from `i18n/en.json`? The translation system will automatically remove them for the rest of the languages.

OgmaJ:
@hmhealey  : removing is done on i18n/en.json

hmhealey (approved):
Thanks! Hopefully that should get the build issues working since I think they were just temporary

ashishbhate (approved):

lindalumitchell (approved):
LGTM.

PR #16265, opened 2020-11-12
[MM-30436] Add support for configuration custom defaults

mgdelacroix:
#### Summary
Allows the configuration store to be created with a set of custom defaults that will be used on top of the server defaults. Makes the server command read the `MM_CUSTOMDEFAULTS` environment variable and loads the defaults from a file located there.

These custom defaults are applied on top of the configuration read from the store, taking precedence only if there is no previous value set, so they effectively work as a partial set of values that are applied when the server initializes the configuration the first time and before the "normal" set of defaults.

As the approach works as a "one-of", I doubt it makes sense to report its usage through telemetry as it is basically the same thing as modifying the server values the first time the server starts, but I'm 1/5 on this.

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-30436

#### Release Note
```release-note
Added ability to load a set of configuration custom defaults from a `MM_CUSTOM_DEFAULTS_PATH` environment variable
```


config/database_test.go
```diff
@@ -133,37 +133,79 @@ func TestDatabaseStoreNew(t *testing.T) {
...
 
+	t.Run("already minimally configured with custom defaults", func(t *testing.T) {
+		_, tearDown := setupConfigDatabase(t, minimalConfig, nil)
+		defer tearDown()
+
+		ds, err := newTestDatabaseStore(t, customDefaults)
+		require.NoError(t, err)
+		defer ds.Close()
+
+		// as the whole config has default values already, custom
```
ethervoid, line 202: I don't understand this test. Looks like the same than line 165

config/database_test.go
```diff
@@ -989,7 +1054,7 @@ func TestDatabaseStoreString(t *testing.T) {
 	_, tearDown := setupConfigDatabase(t, emptyConfig, nil)
 	defer tearDown()
 
-	ds, err := newTestDatabaseStore(t)
+	ds, err := newTestDatabaseStore(t, nil)
```
ethervoid, line 1057: What happens if the env. variable for the custome defaults is set but the file is not in place?

ethervoid (commented):
Love the approach :heart: . Some comments!

cmd/mattermost/commands/server.go
```diff
@@ -34,6 +36,25 @@ func init() {
...
+			return nil, errors.Wrapf(err, "unable to load config custom defaults")
+		}
+		defer file.Close()
+
+		err = json.NewDecoder(file).Decode(&customDefaults)
+		if err != nil {
+			return nil, errors.Wrapf(err, "unable to decode custom defaults configuration")
+		}
+	}
+	return customDefaults, nil
```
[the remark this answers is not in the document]
mgdelacroix: As the `customDefaults` variable is declared but not initialised, it has `nil` value if there are not custom defaults, so it would return `nil` in that case

config/database_test.go
mgdelacroix: The difference is in the `setupConfigDatabase` call. In the test at line 165 is initialising the configuration with `testConfig`, which only has the `SiteURL` set, so the defaults should overwrite `ExperimentalTimezone` but not `SiteURL`, as is loaded with the config.

In the test at 202, the `setupConfigDatabase` fn is called with `minimalConfig`, which instead of a partial config is a complete one, so it has all the values set.

config/database_test.go
mgdelacroix: Good question. This case is not contemplated at the `ConfigStore` creation level as we're passing the config struct to it, but managed [as part of the server command](https://github.com/mattermost/mattermost-server/pull/16265/files#diff-206ba9b9ee4ca76d959e08e934151f296c42002c0a03d50f44c64ecfdcdd6c85R46), and the command will fail if the file is not present or cannot be opened/correctly decoded.

I'm actually 0/5 on this, the server could just try to load the file and simply log an error if any and continue without the custom defaults, but my understanding is that we would prefer for the pod not to start if the custom defaults variable is present but cannot be loaded.

cmd/mattermost/commands/server.go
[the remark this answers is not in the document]
ethervoid: Yeah nvm I

config/database_test.go
ethervoid: Oh ok, I missed the Equal vs NotEqual

ethervoid (approved):
LGTM! Thank you very much @mgdelacroix :D

jwilander:
First of all, big thanks for this @mgdelacroix 👍 

Second, @mgdelacroix @ethervoid @marianunez what are your thoughts on including the cloud defaults file in the mattermost-server repo and then include it in distributable? Two reasons I want to do that:

1. The cloud config defaults are in git and version controlled
2. The cloud config defaults are somewhere anyone can see/edit so that anyone can make changes to it without having to do something somewhere else

Thoughts?

cmd/mattermost/commands/server.go
```diff
@@ -34,16 +36,42 @@ func init() {
...
 
 	if err := utils.TranslationsPreInit(); err != nil {
-		return errors.Wrapf(err, "unable to load Mattermost translation files")
+		return errors.Wrap(err, "unable to load Mattermost translation files")
 	}
-	configStore, err := config.NewStore(getConfigDSN(command, config.GetEnvironment()), !disableConfigWatch)
+
+	customDefaults, err := loadCustomDefaults()
+	if err != nil {
+		return errors.Wrap(err, "cannot load custom defaults")
```
marianunez, line 71: Should we have a soft fail here, we log the error but continue the server startup? Not sure if we should stop the server because of issues with custom defaults. Thoughts @jwilander @ethervoid ?

cmd/mattermost/commands/server.go
```diff
@@ -34,16 +36,42 @@ func init() {
 	RootCmd.RunE = serverCmdF
 }
 
+func loadCustomDefaults() (*model.Config, error) {
+	var customDefaults *model.Config
+	customDefaultsPath := os.Getenv("MM_CUSTOMDEFAULTS")
```
marianunez, line 41: Maybe we can name this env var `MM_CUSTOMDEFAULTSPATH` to make it clear is a file path?
```suggestion
	customDefaultsPath := os.Getenv("MM_CUSTOMDEFAULTSPATH")
```

cmd/mattermost/commands/server.go
```diff
@@ -34,16 +36,42 @@ func init() {
...
+
+	if customDefaultsPath != "" {
```
marianunez, line 43: Nit: we can return early here to reduce one level of indentation below

marianunez (commented):
Thanks @mgdelacroix! Looks great! Some minor comments and just double checking if we want a hard fail when having issues loading the custom defaults

cmd/mattermost/commands/server.go
mgdelacroix: Changed it to `MM_CUSTOM_DEFAULTS_PATH` and updated the Release Notes too, please let me know if that's fine or you prefer to have no separation between words

cmd/mattermost/commands/server.go
mgdelacroix: Addressed in 02e8146d2

cmd/mattermost/commands/server.go
mgdelacroix: We started a discussion [here as well](https://github.com/mattermost/mattermost-server/pull/16265#discussion_r522272366), I'm really 0/5 on this. If we're planning to use the feature mainly on cloud and we rely on this to work, my assumption is that we would prefer for the pods not to start on error, but I'm happy to change the approach if a soft failure is preferred

mgdelacroix:
Thanks all for the quick reviews!!

> Second, @mgdelacroix @ethervoid @marianunez what are your thoughts on including the cloud defaults file in the mattermost-server repo and then include it in the distributable? Two reasons I want to do that:

I'm kinda torn on this, as it feels like cloud configuration should be close to cloud deployment code, rather than in the general server and bundled with its distribution. We're going to bundle a file that exposes part of our internal configuration (tho I doubt it could be a security risk, future is uncertain) and is going to pose no value for the users that download the server. At the same time, having to create a PR to mattermost-server to modify a cloud default feels unnatural and ties the server repo to cloud infrastructure configuration details.

Having said that, and being aware that I lack of a lot of context on the cloud process, I do understand the positives:

> 1. The cloud config defaults are in git and version controlled

Completely understand this, tho I don't know how cloud's infra code is organised or if this could be achieved somehow else in a deployment-specific repo.

> 2. The cloud config defaults are somewhere anyone can see/edit so that anyone can make changes to it without having to do something somewhere else

I see this as something quite valuable if we communicate the idea that when someone develops a new config property, they should take cloud defaults into account as part of the PR that adds the config change.

We could even develop a `mattermod` check that posts a message in the PR when a config property is added notifying cloud in case we'd like to include a cloud default. With something like this we could avoid putting the configuration in the server bundle too and ensure that we're making cloud aware of the change instead of relying on the developer involved with the change.

I think that it's valuable having the configuration stored in an independent place from server, but if you think that it is the best solution, I understand the tradeoffs and the value of doing so as well :)

cmd/mattermost/commands/server.go
ethervoid: If possible we can check if the env. is cloud and make a hard fail in that case. Could be that too specific?

ethervoid:
Regarding where to put the configuration:

- In case we decide to put it in the server repo, I'll include it in the enterprise one given that is part of the cloud codebase
- My main concern regarding this is what Miguel said about having a disonance between when the changes are introduced and when the configuration is introduced when we want to freeze version to release the cloud one.
- I'm not a big fan of solving configuration issues inside the codebase repository beyond having some defaults to help the new users configure the server

Said this I understand the tradeoffs at this point and I'd like to know the downsides/caveats of having that file as part of the build pipeline (apart for being a bit error-prone to remember that you have to change that file if you make changes in the configuration)

cmd/mattermost/commands/server.go
marianunez: I don't believe there should be anything in the defaults that are absolutely necessary for pods to start, those type of settings are handled directly by env var in the groups. @jwilander can correct me here.  

Just want to make sure we don't stop the server by an improper use of the new custom defaults env var and/or defaults file outside of cloud but at the same time would like to keep this implementation generic and not include a cloud condition here.

cmd/mattermost/commands/server.go
marianunez: `MM_CUSTOM_DEFAULTS_PATH` sounds good to me. Thanks!

jwilander:
Thanks for the thoughts @mgdelacroix @ethervoid 

@ethervoid I'll leave it up to you what to do in the near term since we need to get this in for the coming release

For long-term/final solution, I do agree that tying configuration to the mattermost-server codebase is a little out of ordinary.  That said, it is just the defaults and not live configuration that's being updated. For self-hosted we already do set those defaults, it just happens to be directly part of the `model` package.

My main concern is that I'd like to have anyone submitting code to actively think about what that change means for cloud and if we have the defaults for cloud in a totally different location than the self-hosted defaults it's going to be easy to forget about it. I might be overthinking it a little if most of the time the cloud defaults won't be any different than the self-hosted defaults. Overall, I'm 0/5 on the actual method/process we choose but I'd say maybe 2/5 that we consider doing something that won't let the cloud defaults get forgotten. A `mattermod` check could be a nice way to do that 👍

cmd/mattermost/commands/server.go
jwilander: I agree with @marianunez. All the critical "must be set" settings will be set by env vars so I don't think it needs to be a hard fail if we can't load the custom defaults

ethervoid:
@jwilander @mgdelacroix @marianunez @jespino @agnivade 

Ok, I'll take it from here. I'll do the following the next Monday:

- For the short term I'll create the default configuration file in the cloud folder from the enterprise repository. That way the team edition doesn't have a clue about this new defaults package.
- I'll investigate/modify the way we're building the EE docker images so we include this new configuration file in order to use it
- Create a PR with the configuration changes rever and all the needed changes in the `cloud-ga` branch to prepare it for the fast-forward
- I'll create a new ticket for the follow-up in this subject

If any of the mentioned have thoughts about this plan of action, please don't hesitate and ask/suggest anything. FYI we need to have it for next monday

ethervoid:
@marianunez @jwilander may you please complete the review to merge this change?

marianunez (approved):
Thanks @mgdelacroix! Let's make sure there is a soft fail when there is a problem loading the custom defaults, besides that LGTM

jwilander (approved):

amyblais:
@mgdelacroix Please also help cherry pick this to cloud-ga branch.

PR #16266, opened 2020-11-12
[Reliability-improvement-POC] Make app initialisation tolerant to database failures

nronas:
## Summary

Currently, when the app gets through its initialization phase it tries to establish a connection with the storage. Then it tries for 3 minutes to ping in order to assess storage health(1 try every 10 seconds for a total of 18 tries). As soon as the max ping tries are exceeded, the process is exited, which will lead to pods being restarted and the whole cycle begins all over.

In the above situation, we know that storage availability is degraded, so restarting the app layer is not going to give us much value in solving or improving the problem at hand.

On the other hand, as soon as the app is initialized and connections are established, in the case of storage degradation, the app is resilient and just returns `500`s back to the web app.

### Proposal

Make storage access semantics homogeneous between the different stages of the app(initialization vs normal operation), by locking on an infinite loop while waiting for the storage to come up online vs having finite tries.

#### Benefits

- Focusing on fixing the issue at hand, which is storage unavailability.
- App restarts during storage unavailability can be risky.
- Reduces worst-case time to recovery, after the storage availability is restored. from previously `time taken for pods to restart` to now fixed `10s(time between pings)`.

@jwilander @crspeller will love your thoughts on this, cc'ed since I could see you two were part of some initial PRs around this part of the code.

Looked at some PRs for historical context but I couldn't build up any actionable context:
- https://github.com/mattermost/mattermost-server/pull/6693
- https://github.com/mattermost/mattermost-server/pull/6810

## Ticket Link
No associated ticket found.

## Release Note
```release-note
Remove max ping retries logic, when assessing storage health.
```

### DIFF
--- a/store/sqlstore/supplier.go
+++ b/store/sqlstore/supplier.go
@@ -230,7 +230,7 @@ func NewSqlSupplier(settings model.SqlSettings, metrics einterfaces.MetricsInter
 func setupConnection(connType string, dataSource string, settings *model.SqlSettings, ctx context.Context, config *model.Config) *gorp.DbMap {
 	var db *dbsql.DB
 
-	if config != nil && config.FeatureFlags.EnsureDatabaseConnection != "off" {
+	if config != nil && config.FeatureFlags.EnsureDatabaseConnectionEnabled {
 		db = ensureDatabaseConnection(settings, dataSource, connType, ctx)
 	} else {
 		db = checkDatabaseConnection(settings, dataSource, connType)
--- a/store/sqlstore/supplier.go
+++ b/store/sqlstore/supplier.go
@@ -1339,10 +1339,10 @@ func checkDatabaseConnection(settings *model.SqlSettings, dataSource, connType s
 func ensureDatabaseConnection(settings *model.SqlSettings, dataSource, connType string, lifecycleContext context.Context) *dbsql.DB {
 	for {
 		ctx, cancel := context.WithTimeout(context.Background(), DB_PING_TIMEOUT_SECS*time.Second)
+		defer cancel()
 
 		select {
 		case <-lifecycleContext.Done():
-			cancel()
 			os.Exit(EXIT_DB_OPEN)
 		default:
 			db, err := dbsql.Open(*settings.DriverName, dataSource)
--- a/store/sqlstore/supplier.go
+++ b/store/sqlstore/supplier.go
@@ -108,17 +108,19 @@ type SqlSupplierStores struct {
 type SqlSupplier struct {
 	// rrCounter and srCounter should be kept first.
 	// See https://github.com/mattermost/mattermost-server/v5/pull/7281
-	rrCounter      int64
-	srCounter      int64
-	master         *gorp.DbMap
-	replicas       []*gorp.DbMap
-	searchReplicas []*gorp.DbMap
-	stores         SqlSupplierStores
-	settings       *model.SqlSettings
-	lockedToMaster bool
-	context        context.Context
-	license        *model.License
-	licenseMutex   sync.RWMutex
+	rrCounter        int64
+	srCounter        int64
+	master           *gorp.DbMap
+	replicas         []*gorp.DbMap
+	searchReplicas   []*gorp.DbMap
+	stores           SqlSupplierStores
+	settings         *model.SqlSettings
+	lockedToMaster   bool
+	context          context.Context
+	license          *model.License
+	licenseMutex     sync.RWMutex
+	lifecycleContext context.Context
+	config           *model.Config
 }
 
 type TraceOnAdapter struct{}
@@ -131,11 +133,13 @@ func (t *TraceOnAdapter) Printf(format string, v ...interface{}) {
 	mlog.Debug(newString)
 }
 
-func NewSqlSupplier(settings model.SqlSettings, metrics einterfaces.MetricsInterface) *SqlSupplier {
+func NewSqlSupplier(settings model.SqlSettings, metrics einterfaces.MetricsInterface, lifecycleContext context.Context, config *model.Config) *SqlSupplier {
 	supplier := &SqlSupplier{
-		rrCounter: 0,
-		srCounter: 0,
-		settings:  &settings,
+		rrCounter:        0,
+		srCounter:        0,
+		settings:         &settings,
+		lifecycleContext: lifecycleContext,
+		config:           config,
 	}
 
 	supplier.initConnection()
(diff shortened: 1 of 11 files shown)


jwilander:
Thanks @nronas for the PR 👍 

First of all, I think there's a bunch we can and should do to the server, particularly around initialization, that will make it much friendlier to run in a containerized environment like Kubernetes. I also think this specific area you've brought up is a great place for us to improve.

As for your proposal, I think it makes sense. We've run into some similar issues related to the pod restarting in Kubernetes because of issues like this and then getting stuck in long back off timers from Kubernetes due to CrashLoopBackoff.

The first thing that came to mind when reading through your code/proposal was how this would affect start-up, particularly in relation to pod readiness probing. The behaviour will change a little but overall I don't think it will cause any issues and as you said will likely result in better overall time to recovery in the event of the database being unreachable. We might want to tweak the values we use for the readiness probe to get the best behaviour.

The other thing I'd want to think about some more is how this might affect our monitoring/alerting. As is currently it is an MM pod will exit out quicker than it would with your proposal in the event that the database isn't reachable. This will cause our alerting to fire off quicker since pod gets in CrashLoopBackoff or a high number of restarts quickly. We could set the readiness probe appropriately to catch it in alerting and/or we could also alert on a number of failed DB pings in the logs.

Any thoughts on those?

---

One other area related to this that we currently see an issue is on first ever start-up of multiple MM pods at the same time. Essentially what happens is they race to create the schema/tables/indexes for the DB. One wins the race and gets to create them, while the other loses out and often ends up restarting 1-2 times before finally coming up. We have a Kubernetes-side fix to it here in our operator https://github.com/mattermost/mattermost-operator/pull/194 that uses an init job to create the tables/etc. first but I think this is something MM should be resilient too itself. This is where it errors out https://github.com/mattermost/mattermost-server/blob/master/store/sqlstore/supplier.go#L179 with something like this: 
```json
{"level":"error","ts":1605107528.6208477,"caller":"sqlstore/supplier.go:179","msg":"Error creating database tables.","error":"pq: duplicate key value violates unique constraint \"pg_type_typname_nsp_index\""}
```

nronas:
@jwilander thanks a bunch for your thoughts and context here.

> The first thing that came to mind when reading through your code/proposal was how this would affect start-up, particularly in relation to pod readiness probing. The behaviour will change a little but overall I don't think it will cause any issues and as you said will likely result in better overall time to recovery in the event of the database being unreachable. We might want to tweak the values we use for the readiness probe to get the best behaviour.

This is a valid point since I had the impression that we start the webserver without requiring a database connection to be established, so in my incorrect assumption the server was `pingable`, hence k8s will think that the pod is healthy, but when it will try to establish a DB connection will error and return `500`. But looking deeper into the code, we do require a healthy database connection to be established when initializing the server. So I think first we need to decouple our webserver to be resilient to database unavailability and then fix the DB connection health check during initialization.

**Potential solution:** We can spin up the webserver just fine, and when the first call needs to talk to the database it then synchronizes the whole connection logic. If that sounds like a good thing to do, I can draft something up.

> The other thing I'd want to think about some more is how this might affect our monitoring/alerting. As is currently it is an MM pod will exit out quicker than it would with your proposal in the event that the database isn't reachable. This will cause our alerting to fire off quicker since the pod gets in CrashLoopBackoff or a high number of restarts quickly. We could set the readiness probe appropriately to catch it in alerting and/or we could also alert on a number of failed DB pings in the logs.

How I see it is that using alerts based on `CrashLoopBackoff` will tell us that the application pods cannot start. In general, this is a good way of alerting when we do not know the reason for the failure, but I think in the scenario that we are talking about we defo know that is something wrong with the storage, so as you suggested we can alert based number of failed pings, we can either send counters to our aggregated metrics(Prometheus) and put a monitor for them or use logs.

<hr />

> One other area related to this that we currently see an issue is on first ever start-up of multiple MM pods at the same time. Essentially what happens is they race to create the schema/tables/indexes for the DB. One wins the race and gets to create them, while the other loses out and often ends up restarting 1-2 times before finally coming up. We have a Kubernetes-side fix to it here in our operator mattermost/mattermost-operator#194 that uses an init job to create the tables/etc. first but I think this is something MM should be resilient to itself. This is where it errors out https://github.com/mattermost/mattermost-server/blob/master/store/sqlstore/supplier.go#L179 with something like this:

This is interesting and IMHO a low-hanging fruit since we can check if the error is `duplicate key value...` and assume that someone else beat us to it. 


Will love your thoughts on the above.

agnivade:
> But looking deeper into the code, we do require a healthy database connection to be established when initializing the server. So I think first we need to decouple our webserver to be resilient to database unavailability and then fix the DB connection health check during initialization.

> Potential solution: We can spin up the webserver just fine, and when the first call needs to talk to the database it then synchronizes the whole connection logic. If that sounds like a good thing to do, I can draft something up.

This is a great idea, but I think it might be hard to implement. The first obstacle is database schema changes. Right now, the app startup does the schema changes. But if we delay them to be done lazily, it might cause unexpected drops in latency when the first request hits the server. Even if we get past that, there are a number of services in the app startup phase which depend on the DB to be up. Making _all_ of them work in an asynchronous fashion might turn out to be complicated.

> This will cause our alerting to fire off quicker since pod gets in CrashLoopBackoff or a high number of restarts quickly. We could set the readiness probe appropriately to catch it in alerting and/or we could also alert on a number of failed DB pings in the logs.

Setting a Prom metric for failed DB pings is a great idea. At the same time I would also expect the actual alert to be triggered from the DB itself because it's down?

store/sqlstore/supplier.go
```diff
@@ -1319,3 +1305,59 @@ func convertMySQLFullTextColumnsToPostgres(columnNames string) string {
...
+	return db
+}
+
+func ensureDatabaseConnection(settings *model.SqlSettings, dataSource, connType string, lifecycleContext context.Context) *dbsql.DB {
+	for {
+		ctx, cancel := context.WithTimeout(context.Background(), DB_PING_TIMEOUT_SECS*time.Second)
+		defer cancel()
+
+		select {
+		case <-lifecycleContext.Done():
```
agnivade, line 1345: I am wondering about the use of the lifecycleContext here. If the DB is not available, `app.NewServer` won't even start. And therefore, the signal handler code wouldn't even execute. So the context will never be cancelled, and therefore, this loop will actually never exit gracefully unless I am mistaken.

If you think about it, even the earlier version suffered from the same problem. If the DB was down, the server would be stuck pinging the DB, and if you send SIGINT, it wouldn't exit gracefully, but just abort.

At the same time, having an infinite loop in the code is not a great idea.

What are your thoughts on making the DB_PING_ATTEMPTS and pingInterval to be configurable? That way the code behaves as before in dev environments, and in production, the cloud team has full control over how long a DB can be expected to be down and set the ping_attempts and ping_interval appropriately. This also avoids the need for it to be a feature flag, as setting the old settings effectively makes it behave like the old code.

nronas:
> This is a great idea, but I think it might be hard to implement. The first obstacle is database schema changes. Right now, the app startup does the schema changes. But if we delay them to be done lazily, it might cause unexpected drops in latency when the first request hits the server. Even if we get past that, there are a number of services in the app startup phase which depend on the DB to be up. Making all of them work in an asynchronous fashion might turn out to be complicated.

Oh, interesting. what's the historical context around the whys for running schema changes as part of the boot-up phase, presumably we are wasting compute cycles by checking for potential data schema updates to be applied on every boot-up since the likelihood to have a schema change is lower than the frequency of app deploys? I would expect boot up vs data schema changes to be decoupled and have something like a one-off command(migration) that will update the schema to it's latest version.

> Setting a Prom metric for failed DB pings is a great idea. At the same time I would also expect the actual alert to be triggered from the DB itself because it's down?

💯  on this, I would also expect a DB alert to be triggered first. The whole vision here is that our application should tolerate database degradation, and by tolerating I mean do not crash the process just return 500 since we can't talk to the DB.

@agnivade will love your thoughts on this

store/sqlstore/supplier.go
nronas: aha, that is true since we are waiting for the os signals after we have the server successfully initialized on the `commands/server.go`. I guess we can work around this by passing the signals channel to the supplier, but thinking this even more it's starting to be a bit hairy.

Maybe if we step back and define what access semantics we would like from our storage layer, can help us shape what needed here.

My view is something like this: `Server` needs a `storage service` object. The storage service owns the storage connection lifecycle. So I guess that boils down to, avoid obtaining a connection when initializing the supplier? With that, we can "configure" the server with its required dependencies just fine. The app will lock on waiting for `os.Signal` and when a request comes and we will need something from storage, then we can either return an error(cannot obtain connection) which will result in a `500` or fulfill the request.

> What are your thoughts on making the DB_PING_ATTEMPTS and pingInterval to be configurable? That way the code behaves as before in dev environments, and in production, the cloud team has full control over how long a DB can be expected to be down and set the ping_attempts and ping_interval appropriately. This also avoids the need for it to be a feature flag, as setting the old settings effectively makes it behave like the old code.

Totally, that is the path of least resistance.

agnivade:
> Oh, interesting. what's the historical context around the whys for running schema changes as part of the boot-up phase, presumably we are wasting compute cycles by checking for potential data schema updates to be applied on every boot-up since the likelihood to have a schema change is lower than the frequency of app deploys? I would expect boot up vs data schema changes to be decoupled and have something like a one-off command(migration) that will update the schema to it's latest version.

I would say convenience is a big factor. We were mainly an on-prem software historically. So it's much more simpler to ask the customers to just run the new binary, rather than ask them to read release notes, and run a separate schema migration tool if needed. Less chances of mistakes and more robust if the software handles this automatically.

However, we do have https://github.com/mattermost/mattermost-server/pull/13654 which addresses the blocking nature of app startup. That gives a nice compromise between startup speed and automatic schema upgrades.

store/sqlstore/supplier.go
agnivade: I think this goes back to my comment [here](https://github.com/mattermost/mattermost-server/pull/16266#issuecomment-728281461). In short, the server needs a storage object _with an active connection_. That's the thing here. The DB _is_ a required dependency for the server to start.

Even if you ignore the schema migrations, there are other services which depend on the DB when the server starts.

nronas:
Closing this, since the problem at hand is a bit bigger in scope from what was originally thought it would be. Also, it seems that we need to have a broader discussion around what type(s) of redundancies do we need from the app's point of view when it talks to storage. Will create another PR that makes the ping interval configurable.

PR #16270, opened 2020-11-13
Add logic to skip duplicate table errors when initializing db

nronas:
#### Summary
Based on the discussion [here](https://github.com/mattermost/mattermost-server/pull/16266#issuecomment-726398876) make database initialization resilient to race conditions, when creating tables.

#### Ticket Link
No associated ticket found

#### Release Note
```release-note
NONE
```

**NOTE** Writing tests for this is a bit tricky.

/cc @jwilander since we chatted about this


store/sqlstore/supplier.go
```diff
@@ -176,9 +178,12 @@ func NewSqlSupplier(settings model.SqlSettings, metrics einterfaces.MetricsInter
 	supplier.stores.productNotices = newSqlProductNoticesStore(supplier)
 	err := supplier.GetMaster().CreateTablesIfNotExists()
 	if err != nil {
-		mlog.Critical("Error creating database tables.", mlog.Err(err))
-		time.Sleep(time.Second)
-		os.Exit(EXIT_CREATE_TABLE)
+		if dupError, ok := errors.Cause(err).(*pq.Error); ok && dupError.Code.Name() == DUP_ENTRY_ERROR {
+			mlog.Warn("Duplicate key error occurred.", mlog.Err(dupError))
```
gabrieljackson, line 182: nit: perhaps something like this would be a little clearer to anyone seeing this log message

```suggestion
			mlog.Warn("Duplicate key error occurred; assuming table already created and proceeding.", mlog.Err(dupError))
```

Not a required change. Just thinking it may give some clarity.

gabrieljackson (approved):
👍

store/sqlstore/supplier.go
```diff
@@ -176,9 +178,12 @@ func NewSqlSupplier(settings model.SqlSettings, metrics einterfaces.MetricsInter
 	supplier.stores.productNotices = newSqlProductNoticesStore(supplier)
 	err := supplier.GetMaster().CreateTablesIfNotExists()
 	if err != nil {
-		mlog.Critical("Error creating database tables.", mlog.Err(err))
-		time.Sleep(time.Second)
-		os.Exit(EXIT_CREATE_TABLE)
+		if dupError, ok := errors.Cause(err).(*pq.Error); ok && dupError.Code.Name() == DUP_ENTRY_ERROR {
```
agnivade, line 181: This would only work with Postgres I believe. Can we modify it to support [MySQL](https://dev.mysql.com/doc/mysql-errors/5.7/en/server-error-reference.html#error_er_table_exists_error) too?

It would also be great if we can refactor this entire thing to a function to just check for the error and return whether it's a duplicate error or not. That way we can also write a test for this.

Take a look at https://github.com/mattermost/mattermost-server/blob/master/store/retrylayer/retrylayer.go#L365-L379 for an example of how we check for repeatable errors coming from the database.

It's also more robust to use `errors.As` rather than to just typecast in case the error was wrapped.

agnivade (commented):
Thanks for the change @nronas ! It's a great idea and was much needed.

store/sqlstore/supplier.go
nronas: Superb, thanks for the context @agnivade. Pretty new to the codebase so i wasn't aware that this code path is exercised with multiple storages. Will amend to handle MySQL as well.

nronas: amended in https://github.com/mattermost/mattermost-server/pull/16270/commits/4d980e8dc6900157fdf5ce8d62491d9bf88d60cb

store/sqlstore/supplier.go
```diff
@@ -1319,3 +1325,22 @@ func convertMySQLFullTextColumnsToPostgres(columnNames string) string {
 
 	return concatenatedColumnNames
 }
+
+// isDuplicate checks whether an error is a duplicate key error, which comes when processes are competing on creating the same
```
agnivade, line 1329: Nit: IsDuplicate

store/sqlstore/supplier_test.go
```diff
@@ -314,6 +318,23 @@ func TestGetAllConns(t *testing.T) {
...
+	testErrors := map[error]bool{
+		&pq.Error{Code: "42P06"}:                                       false,
+		&pq.Error{Code: sqlstore.PG_DUP_TABLE_ERROR_CODE}:              true,
+		&mysql.MySQLError{Number: uint16(1000)}:                        false,
+		&mysql.MySQLError{Number: sqlstore.MYSQL_DUP_TABLE_ERROR_CODE}: true,
+		errors.New("Random error"):                                     false,
+	}
+
+	for err, expected := range testErrors {
+		t.Run(fmt.Sprintf("Should return %t for %s", expected, err.Error()), func(t *testing.T) {
```
agnivade, line 331: A separate test name would have been better. But this is fine too :)

store/sqlstore/supplier.go
```diff
@@ -28,10 +29,12 @@ import (
...
 
 const (
-	INDEX_TYPE_FULL_TEXT = "full_text"
-	INDEX_TYPE_DEFAULT   = "default"
-	DB_PING_ATTEMPTS     = 18
-	DB_PING_TIMEOUT_SECS = 10
+	INDEX_TYPE_FULL_TEXT       = "full_text"
+	INDEX_TYPE_DEFAULT         = "default"
+	PG_DUP_TABLE_ERROR_CODE    = "42P07"      // see https://github.com/lib/pq/blob/master/error.go#L268
+	MYSQL_DUP_TABLE_ERROR_CODE = uint16(1050) // see https://fromdual.com/mysql-error-codes-and-messages-1050-1099#error_er_table_exists_error
```
agnivade, line 35: I would prefer referring to the official MySQL reference- https://dev.mysql.com/doc/mysql-errors/5.7/en/server-error-reference.html#error_er_table_exists_error

agnivade (commented):
Looks amazing now. Just some minor nits to take care of. And then it should be good to go.

Thanks @nronas !

store/sqlstore/supplier.go
nronas: I do not know why i pasted that link there :D. Thanks, fo catching this up 🦅 👀

agnivade (approved):

agnivade:
@lindalumitchell - can be qa-deferred.

lindalumitchell (approved):

jwilander:
Thanks @nronas !

PR #16271, opened 2020-11-13
[MM-29515] Don't allow slack imports in cloud installations

nickmisasi:
#### Summary
See https://github.com/mattermost/mattermost-webapp/pull/7069

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-29515

#### Release Note
```release-note
Slack imports are no longer allowed in Cloud installations.
```


api4/team.go
```diff
@@ -1079,6 +1079,11 @@ func teamExists(c *Context, w http.ResponseWriter, r *http.Request) {
 }
 
 func importTeam(c *Context, w http.ResponseWriter, r *http.Request) {
+	if *c.App.Config().ExperimentalSettings.RestrictSystemAdmin {
+		c.Err = model.NewAppError("getLogs", "api.restricted_system_admin", nil, "", http.StatusForbidden)
```
marianunez, line 1083:
```suggestion
		c.Err = model.NewAppError("importTeam", "api.restricted_system_admin", nil, "", http.StatusForbidden)
```

marianunez (changes requested):
Could we cover this case in the unit tests? https://github.com/mattermost/mattermost-server/blob/46a409aef6b958b39109888fd1adff2edf35bf6f/api4/team_test.go#L2672

nickmisasi:
@marianunez added a test case to make sure it returns 403 when the setting is true.

api4/team_test.go
```diff
@@ -2709,6 +2709,20 @@ func TestImportTeam(t *testing.T) {
 		require.Equal(t, posts.Posts[posts.Order[3]].Message, "This is a test post to test the import process", "missing posts in the import process")
 	})
 
+	t.Run("CloudForbidden", func(t *testing.T) {
```
marianunez, line 2712:
```suggestion
	t.Run("RestrictSystemAdmin Forbidden", func(t *testing.T) {
```

marianunez (approved):

api4/team.go
```diff
@@ -1079,6 +1079,11 @@ func teamExists(c *Context, w http.ResponseWriter, r *http.Request) {
 }
 
 func importTeam(c *Context, w http.ResponseWriter, r *http.Request) {
+	if *c.App.Config().ExperimentalSettings.RestrictSystemAdmin {
```
ethervoid, line 1082: This would apply to on-premise installations too. Are we ok with it?

nickmisasi: @marianunez thoughts here? I think we've only been using this RestrictSystemAdmin setting for toggling things on and off, rather than checking for cloud?

marianunez: Correct, we have relied on this on other endpoints like locking the system logs. This was created initially for Cloud but in a generic way others can leverage. 1/5 we keep consistency and keep using but I'm ok as well with looking at the cloud license. Thoughts @joram?

Would like to get this merged today before the FF

nickmisasi: +1 on consistency, will wait for thoughts from Joram :)

nickmisasi: So I think we're good here? @ethervoid @marianunez

marianunez: As discussed in today's planning, let's do the check on cloud licenses here instead of `RestrictSystemAdmin` as suggested by Mario

ethervoid (approved):
LGTM!

PR #16286, opened 2020-11-16
MM-30544 - Fix interactive dialogs called from post integration actions

cpoile:
#### Summary
- The Incident Management plugin has the following use case: We have a reminder post with a button `Update Status`, when the user clicks on it, it should call the slash command `/incident update` which opens an interactive dialog. However, the slash command is called from the plugin's code, and there is no way to return the `trigger_id` to the client. 
- This change allows an integration request to return a `trigger_id` field in the response. If the `DoPostActionWithCookie` method sees a `trigger_id` field returned, it will return that (instead of the `trigger_id` it generated).
- I am pretty sure this won't break anything, since the trigger_id is being used for what it's supposed to. And the plugin should only return a trigger_id when /it/ is the code that is creating the dialog. But please let me know if I've overlooked something.

#### Ticket Link
- https://mattermost.atlassian.net/browse/MM-30544

#### Release Note
```release-note
NONE
```

### DIFF
--- a/app/integration_action.go
+++ b/app/integration_action.go
@@ -291,7 +291,14 @@ func (a *App) DoPostActionWithCookie(postId, actionId, userId, selectedOption st
 		a.SendEphemeralPost(userId, ephemeralPost)
 	}
 
-	return clientTriggerId, nil
+	// Allow the integration action to override the triggerId, e.g. when a plugin opens an
+	// interactive dialog and needs to return that triggerId to the client.
+	returnTriggerId := clientTriggerId
+	if response.TriggerId != "" {
+		returnTriggerId = response.TriggerId
+	}
+
+	return returnTriggerId, nil
 }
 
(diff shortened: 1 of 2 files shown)


levb (approved):
Adding @hanzei since he understands the TriggerID business.

lieut-data:
@cpoile, is there a way to avoid the `/incident update` slash command execution in our pieline? I had hoped the framework would allow the interactive buttons, themselves, to directly reach the server (i.e. the plugins endpoint) and transition to an interactive dialog. I too look forward to @hanzei's understanding of the limitations here, since I'm surprised we haven't run into this before.

cpoile:
@lieut-data Yes, we can call the open interactive dialog directly and pass through the encoded `triggerId` we receive from the post action. I think I might have passed in a decoded `triggerId` when I first tried that. 🤦 

I suppose we don't need this PR anymore. Sorry for the trouble!

PR #16288, opened 2020-11-16
Fix db migration

ethervoid:
#### Summary

This PR is a temporal fix to upgrade the database for some cloud installations that have version 5.30, because we get that from the [currentModelVersion](https://github.com/mattermost/mattermost-server/blob/cloud/store/sqlstore/upgrade.go#L117), instead of 5.29 and they missed some needed migrations.

#### Release Note

```release-note
NONE
```

### DIFF
--- a/store/sqlstore/upgrade.go
+++ b/store/sqlstore/upgrade.go
@@ -929,22 +929,22 @@ func precheckMigrationToVersion528(sqlStore SqlStore) error {
 }
 
 func upgradeDatabaseToVersion529(sqlStore SqlStore) {
-	if shouldPerformUpgrade(sqlStore, VERSION_5_28_1, VERSION_5_29_0) {
-		sqlStore.AlterColumnTypeIfExists("SidebarCategories", "Id", "VARCHAR(128)", "VARCHAR(128)")
-		sqlStore.AlterColumnDefaultIfExists("SidebarCategories", "Id", model.NewString(""), nil)
-		sqlStore.AlterColumnTypeIfExists("SidebarChannels", "CategoryId", "VARCHAR(128)", "VARCHAR(128)")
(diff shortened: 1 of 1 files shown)


jwilander (approved):
I don't think it's a huge deal but these are going to run on every start-up of an MM pod now. It should be fine since they're all idempotent but something to be aware of.

agnivade:
Was it the `ChannelId` column in the threads table? If so, then I think it's https://community-daily.mattermost.com/core/channels/developers-server/i6qmt617pt8y8b6dqnf4t9ypwc.

I agree with Joram that this change is a bit unnecessary for fresh pods, because they will always be run. For existing installations, I would suggest changing the schema version in the Systems table manual and just restarting the pod.

To be clear, this was a mistake made while bumping up schema versions and the actual schema change. There's no bug in the migration logic per-se.

lieut-data:
Could we would just "duplicate" the migration steps in the more recent migration step instead?

ethervoid:
@lieut-data until we face the migration changes we're going to find this kind of issue due to having the version coupled to the database changes. 

Moreover, we have an additional issue which is that due to our current versioning system we're bumping from one feature version to another, eg. from 5.28.0 to 5.30.0, in the version model. Because of that, any fresh install in the cloud is going to populate the database using gorp and put the version for the schema in 5.30 so all the changes we made in the database between 5.28 to 5.30 are going to be missing because we're using master, this doesn't happen in release because we're setting the model version to the release one.

So, in summary, we need to tackle multiple fronts here versioning and migrations for the cloud installations. Right now I'm working on a migration checker that would check if the migration is applied, despite the system version stored, and applied if some migration has missing changes.

I'm going to close this PR in favor of the new one

PR #16290, opened 2020-11-17
MM-29487: Fix racy TestMuteCommandSpecificChannel

agnivade:
The code was doing `go fakeApp.UpdateProductNotices()`
due to which the Store object was accessed in a racy manner
when it was reset if the localcachelayer flag was set.

We fix this by using the StoreOverride functionality
to pre-define what store to use while initializing the server
itself so that there isn't a need to reset the store _after_
the server is set.

We also apply the same fix in the app and api4 layers too
to prevent similar things from happening there.

https://mattermost.atlassian.net/browse/MM-29487

```release-note
NONE
```


reflog (approved):

streamer45 (approved):
Nice :+1:

PR #16303, opened 2020-11-18
[MM-28883] Migration checker

ethervoid:
#### Summary

This PR includes a "migration checker". This allows us to include a function, in the code where we check if we need to apply a migration, that checks if the migration changes are already applied or we have missing changes and if we have missing changes, we apply that migration again.

Also, we include the option to check for version prior to change the `Systems.Version` field so we can check if is necessary to update it or not. We introduce this because to avoid changing the version to the previous one if we have missing migration changes while we're on a greater version.

In the logs we reflect those exceptional events like this:

```json
{"level":"info","ts":1605699867.6229799,"caller":"sqlstore/upgrade.go:254","msg":"Missing migrations detected for version 5.28.1","version_to_upgrade":"5.29.0","db_stored_version":"5.30.0"}
{"level":"info","ts":1605699867.9481034,"caller":"sqlstore/upgrade.go:228","msg":"No need to update the DB schema version","system_version":"5.30.0","upgrade_version":"5.29.0"}
```

Check the "No need to update..." log event where we detect that a request for schema version update in the DB but given that the system version is greater or equal to the requested upgrade, we don't proceed.

#### Ticket Link

https://mattermost.atlassian.net/browse/MM-28883

#### Release Note

```release-note
NONE
```


store/sqlstore/upgrade.go
```diff
@@ -19,7 +19,7 @@ import (
 )
 
 const (
-	CURRENT_SCHEMA_VERSION   = VERSION_5_29_0
+	CURRENT_SCHEMA_VERSION   = VERSION_5_30_0
```
ethervoid, line 22: For cloud, we're going to set this as the next DB schema until we close the version and move to the next one. We're going to leverage the use of this checker to include or re-execute the migrations for 5.30.

Why? This way we keep the migrations reapply for every rollout/restart we made in the pods to the minimum

lieut-data (dismissed):
I don't think this is the right way to go about solving the current problem. I'm worried about a new class of bugs in the checking functions. I think it's easier to reason about a migration that can only be applied once, and push the complexity into the framework for writing the migrations instead of executing them.

Could we pursue an approach that allows each commit to effectively add a discrete migration? Maybe this means abandoning the current numbering scheme altogether, or maybe we just rely on the patch versions in the existing model?

ethervoid:
@lieut-data Yeah totally agree and that's the idea in the short-term. To include a new framework, go-migrate, to decouple the migrations from versioning and only apply migrations as discrete changes.

But we're facing this problem now where we have installations pointing to version 5.30 but with the schema from 5.28.1. This is a temporal fix, only for cloud branch, to check for missing migrations and apply when needed even if the version is greater.

This way we're able to update those installations and synchronize them until we have proper versioning and migration tools

lieut-data:
Given the transient nature of this problem, and that our migrations are usually idempotent, would it feasible just to comment out the `shouldPerformUpgrade` checks for `upgradeDatabaseToVersion529` and `upgradeDatabaseToVersion530`, deploy the change to cloud (recognizing that it may run more often than necessary), wait for the rollout to complete, and then just revert these changes in the cloud branch?

ethervoid:
@lieut-data yep that was on the table, when Joram and I talked about that, but was discarded because:

- This looks like is not going to be a one-time thing. This could happen more times, for example with the open 5.30 migration we're going to have installations with some migrations that could not be applied when we close the migration.
- Those migrations for 5.29 are going to be applied ad infinitum until we do the rollback (not a big problem but all the installations are going to re-apply those migrations)
- From my side, I want to avoid having an open migration, right now in master we have the 5.30 open, that would be applied every time we make a rollout
- Every time we make a fast forward, and due to the change made un the upgrades.go file, we're going to face conflicts so we can make the needed changes, and also we can tell the company about this little change so we can include checks for the migrations.

Those are the reasons for introducing this checker instead of doing manually and revert changes. And also manual changes are more error-prone so if we can remove operational work would be awesome too

agnivade (commented):
Looks good in general. Just some minor comments.

I do have one concern though which didn't occur to me earlier.

- So for all future migrations, do we need to have a separate commit for the cloud branch and a different one for master?
- How do we want to handle merge conflicts?
- How do we actually enforce this and ensure that nobody pushes a schema change without a `missingMigrationCheck` function?

store/sqlstore/upgrade.go
```diff
@@ -921,7 +964,7 @@ func precheckMigrationToVersion528(sqlStore SqlStore) error {
 }
 
 func upgradeDatabaseToVersion529(sqlStore SqlStore) {
-	if shouldPerformUpgrade(sqlStore, VERSION_5_28_1, VERSION_5_29_0) {
+	if shouldPerformUpgrade(sqlStore, VERSION_5_28_1, VERSION_5_29_0, hasMissingMigrationsVersion529) {
```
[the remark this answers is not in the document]
ethervoid: Sure, no problem

store/sqlstore/supplier.go
```diff
@@ -537,6 +542,42 @@ func (ss *SqlSupplier) DoesColumnExist(tableName string, columnName string) bool
 	}
 }
 
+func (ss *SqlSupplier) GetColumnInfo(tableName, columnName string) (*ColumnInfo, error) {
```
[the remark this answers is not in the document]
ethervoid: Yes, of course

ethervoid:
@agnivade 

> So for all future migrations, do we need to have a separate commit for the cloud branch and a different one for master?

That would be the better option for us but is not necessary if we don't want to (see next point)

> How do we want to handle merge conflicts?

We need to handle those anyway after we make the fast forward of the cloud branch to master so we can do that as part of the manual process of making the release artifact to be tested (kind of the work we make now to uncomment the migration code and fix the model version in the release branch)

> How do we actually enforce this and ensure that nobody pushes a schema change without a missingMigrationCheck function?

Well, the cloud release people are in control of that. We should introduce a new check in the workflow and if possible another one in mattermod so we can be aware of it

The idea is to fix the OAuth problem and the migration/version problem so we can be able to use master for cloud releases but in the mean time...

lieut-data:
I'm still struggling to understand why we need the complexity of checking the schema. It seems to me that we're already so close to where we want to end up. What if pushed towards something like this instead?
```go
func upgradeDatabaseToVersion530(sqlStore SqlStore) {
        applyMigrationOnce(sqlStore, "5_30_file_info_changes", func() {
                sqlStore.CreateColumnIfNotExistsNoDefault("FileInfo", "Content", "longtext", "text")
        })
        applyMigrationOnce(sqlStore, "5_30_sidebar_categories", func() {
		sqlStore.CreateColumnIfNotExists("SidebarCategories", "Muted", "tinyint(1)", "boolean", "0")
        })

        // Since v5.30, the recorded version schema is now just a coarse checkpoint 
        // to signal the state of the schema, and not itself the gate to running a migration.
	if shouldPerformUpgrade(sqlStore, VERSION_5_29_0, VERSION_5_30_0) {
  		saveSchemaVersion(sqlStore, VERSION_5_30_0, true)
        }
}
```

where `applyMigrationOnce` conditionally runs the given callback. We manage a new table, `Migrations` that records the key name and time of migration of all keys, load that on first startup into a map, and use it to do the conditional check.

If we want to later eliminate the whole notion of `upgradeDatabaseToVersionX`, that's easily accomplished too.

agnivade:
> We need to handle those anyway after we make the fast forward of the cloud branch to master so we can do that as part of the manual process of making the release artifact to be tested (kind of the work we make now to uncomment the migration code and fix the model version in the release branch)

Hmm .. but will this really scale? I do realize that all of this is temporary, but just cautious if we aren't digging a hole for ourselves here. Manually changing code and maintaining a different branch sounds like a nightmare to me.

> We manage a new table, Migrations that records the key name and time of migration of all keys, load that on first startup into a map, and use it to do the conditional check.

I think that will also turn out to have equal complexity. Ideally, we should just move to use golang-migrate. But until then, we have to decide on a compromise. I am 0/5 on either the existing approach or having a named key for each migration.

One problem with the named key approach might be that, when we _do_ move to golang-migrate. Then we would have yet another Migrations table that captures migration versions based on a different logic. Not sure how would that turn out eventually.


---------

I want revisit the idea of bumping the model version _after_ the current release is tagged, not when the _previous_ release is tagged. You mentioned that mobile team needed it. I am wondering if we can just have a separate variable with 2 different values? The mobile team can use the usual one, and that can have the normal workflow. And the schema migrations can use the new one, and the cloud team can take care of bumping it _after_ the release is tagged. That way, both schema migrations and mobile team is taken care of. And you don't need to fix conflicts.

This removes all the complexity. You just need to uncomment the required code, let the migrations run and let the current problem fix itself. And then for all future code, we can use the new variable. And that should work for both master and cloud.

Only concern is schema changes backports. But good news is that cloud branch will always go forward, and not use backports. So we should be good there.

What do you think?

ethervoid:
> Hmm .. but will this really scale? I do realize that all of this is temporary, but just cautious if we aren't digging a hole for ourselves here. Manually changing code and maintaining a different branch sounds like a nightmare to me.

Well, I don't think is much more different that we are already doing for the release with the migrations and version bumping right?

> One problem with the named key approach might be that when we do move to golang-migrate. Then we would have yet another Migrations table that captures migration versions based on a different logic. Not sure how would that turn out eventually.

Yep for me including a new table means we stick with that table for the future, we deprecate that table and we have garbage in our schema or worse, we have to get rid of it. 

In any case, taking into account that we're going to tackle the new migration system in weeks, I'm not inclined to introduce new tables and/or changes in the schema that we need to revert or let there unused.

Also, the current idea on the table for the new migration system is to keep the current one to apply the old migrations and start using the new one after a designated version. So if we introduce a new table we have either to migrate the old migrations to use the new table or have two ways of dealing with migration and when the new system is added, we would have three.

In summary, from my perspective, we introduce more complexity by adding a new table and new migration methods than just checking the schema and in case of doubt reapply the migration given their idempotent nature.

---------

> You mentioned that mobile team needed it.

I'm not sure about this. I heard something but I'm not known for sure. I'll ping @hanzei because he's the one doing the version bumping

> And the schema migrations can use the new one, and the cloud team can take care of bumping it after the release is tagged. That way, both schema migrations and mobile team is taken care of. And you don't need to fix conflicts.

So you are suggesting to include new versioning for the cloud that use that to slice the migrations by "cloud versioning" instead of using the on-prem version, right?

If that's the case, we're in a similar position than before, we have to change the current workflow and manually prepare the versioning, which migrations enter in the new one, etc.

Something I want to stress is that we need to find a solution quickly, I know all are valid solutions but having 3 of them and not having consensus for days is not a desirable situation because we're in the middle of the release testing and we need to implement something to fix the current problem we're facing.

My proposal is to either find a common ground to work in the solution or just review this current solution and go ahead with it if doesn't have an empirical design flaw

lieut-data:
Given the time sensitivity here, I've lifted my requested changes. I love the idea of moving towards something like `golang-migrate`, though it's obviously out of scope for the changes needed here. 

I think my initial gut reaction is the "touch everything" approach in updating the signature on `shouldPerformUpgrade`, and the worry that this won't just be with us for the short-term.

Perhaps, as a compromise, we could just tidy things up to be more localized to these migrations:
```go
func upgradeDatabaseToVersion530(sqlStore SqlStore) {
        if hasMissingMigrationsVersion530(sqlStore) {
                sqlStore.CreateColumnIfNotExistsNoDefault("FileInfo", "Content", "longtext", "text")
		sqlStore.CreateColumnIfNotExists("SidebarCategories", "Muted", "tinyint(1)", "boolean", "0")
        }

        // In v5.30, the recorded version schema is just a coarse checkpoint until we address
        // to signal the state of the schema, and not itself the gate to running a migration.
	if shouldPerformUpgrade(sqlStore, VERSION_5_29_0, VERSION_5_30_0) {
  		saveSchemaVersion(sqlStore, VERSION_5_30_0, true)
        }
}
```

agnivade:
> If that's the case, we're in a similar position than before, we have to change the current workflow and manually prepare the versioning, which migrations enter in the new one, etc.

In my mind, all we do is just delay the bumping of model.CloudSchemaVersion till after the release is tagged. The `model.SchemaVersion` can be bumped in the usual manner.

Sorry, don't want to hold this up any longer. The current approach or Jesse's last comments also look good to me. Once you just rename the functions so that old code is left untouched, it's good from my side. :+1:

ethervoid:
Cool, on it. Thank you guys for being understanding the time restriction. I'm fully aware that is not an optimal solution and we could face problems while applying it :crossed_fingers:

I'll tackle those changes and ask for a review again

mgdelacroix (approved):
LGTM given the time constraints and besides the existing comments

store/sqlstore/upgrade.go
```diff
@@ -209,7 +209,8 @@ func saveSchemaVersion(sqlStore SqlStore, version string) {
 }
 
 func shouldPerformUpgrade(sqlStore SqlStore, currentSchemaVersion string, expectedSchemaVersion string) bool {
-	if sqlStore.GetCurrentSchemaVersion() == currentSchemaVersion {
+	currentStoredSchemaVersion := sqlStore.GetCurrentSchemaVersion()
```
agnivade, line 212: Hmm .. not sure I am seeing the point of this variable re-assignment.

agnivade (approved):
Looks good. Great work Mario given the time constraints!

store/sqlstore/upgrade.go
ethervoid: Oh that's part of the old code. Reverting that

store/sqlstore/upgrade_test.go
```diff
@@ -57,10 +57,10 @@ func TestStoreUpgrade(t *testing.T) {
 		})
 
 		t.Run("upgrade schema running later minor version", func(t *testing.T) {
-			saveSchemaVersion(sqlStore, "5.29.0")
```
lieut-data, line 60: Hah, I clearly didn't think big enough when writing this test :p

lieut-data (approved):
Really nice -- thanks, @ethervoid for the iteration!

PR #16307, opened 2020-11-18
Removing AppError check from the retry layer

jespino:
#### Summary
Removing AppError check from the retry layer. This is part of the clean up of
the AppError usage on the store layers.

### DIFF
--- a/store/layer_generators/retry_layer.go.tmpl
+++ b/store/layer_generators/retry_layer.go.tmpl
@@ -60,25 +60,21 @@ func (s *{{$.Name}}{{$substoreName}}Store) {{$index}}({{$element.Params | joinPa
         s.{{$substoreName}}Store.{{$index}}({{$element.Params | joinParams}})
     {{else}}
         {{if $element.Results | errorPresent}}
-            {{if $element.Results | isAppError}}
-                return s.{{$substoreName}}Store.{{$index}}({{$element.Params | joinParams}})
-            {{else}}
-                tries := 0
-                for {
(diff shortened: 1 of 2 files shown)


reflog (approved):

jespino:
I set the Tests/Not-Needed label because it doesn't affect the store at all because the changes doesn't get propagated to the real code, I only changed the template, so... this shouldn't need any test at all.

store/layer_generators/retry_layer.go.tmpl
```diff
@@ -60,25 +60,21 @@ func (s *{{$.Name}}{{$substoreName}}Store) {{$index}}({{$element.Params | joinPa
...
-                    }
-                    if !isRepeatableError({{$element.Results | errorVar}}) {
-                        return {{genResultsVars $element.Results false }}
-                    }
-                    tries++
-                    if tries >= 3 {
-                        {{$element.Results | errorVar}} = errors.Wrap({{$element.Results | errorVar}}, "giving up after 3 consecutive repeatable transaction failures")
-                        return {{genResultsVars $element.Results false }}
-                    }
+            tries := 0
```
agnivade, line 63: Did nothing change in the generated code?

jespino: Exactly, because all the code was already without AppError, so this path is no longer followed by any store method. That proof that is longer not needed, and that everything is already migrated :)

agnivade: Got it.

agnivade (approved):

PR #16311, opened 2020-11-18
MM-30740 Add support for boolean feature flags.

crspeller:
### Summary
Support for boolean feature flags. To set in split.io use "on" or "true" any other value will be interpreted as false. 

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-30740

#### Release Note
```release-note
NONE
```


nickmisasi (approved):
LGTM, thanks @crspeller!

config/feature_flags.go
```diff
@@ -73,7 +74,8 @@ func (f *FeatureFlagSynchronizer) Close() {
 // featureFlagsFromMap sets the feature flags from a map[string]string.
 // It starts with baseFeatureFlags and only sets values that are
 // given by the upstream management system.
-// Makes the assumption that all feature flags are strings for now.
+// Makes the assumption that all feature flags are strings or booleans.
+// Strings are converted to booleans by consitering "on" or "true" as true and any other value as false.
```
cpoile, line 78: Wondering if we shouldn't use the `strconv.ParseBool()` semantics, since that's the Go standard way of parsing (and pretty well understood)? The downside of that is that `on` would return an error.

cpoile (changes requested):
Just a minor question about the parsing logic.

config/feature_flags.go
crspeller: The defaults for split.io are "on" and "off" so I would really like "on" to be true. What I can do is check for "on" then use strconv.

config/feature_flags.go
```diff
@@ -75,7 +76,7 @@ func (f *FeatureFlagSynchronizer) Close() {
 // It starts with baseFeatureFlags and only sets values that are
 // given by the upstream management system.
 // Makes the assumption that all feature flags are strings or booleans.
-// Strings are converted to booleans by consitering "on" or "true" as true and any other value as false.
+// Strings are converted to booleans by consitering case insenstive "on" or any value consitered by strconv.ParseBool as true and any other value as false.
```
cpoile, line 79:
```suggestion
// Strings are converted to booleans by considering case insensitive "on" or any value considered by strconv.ParseBool as true and any other value as false.
```

cpoile (approved):
Works for me!

prapti (approved):
Tested on a cloud test server -- works as expected the boolean feature flag as well as the regular test feature flag get picked up as set in split.io.
Approving.

PR #16319, opened 2020-11-19
[MM-29845] Add a new handler to allow authentication via CWS API Key

nickmisasi:
#### Summary
When Stripe sends a webhook to the CWS in Cloud, we need a way for CWS to process that webhook, and send a notification to the correct installation. In order to do that, we need a new authentication handler to allow for authenticating via a token

Previously, we introduced a new environment variable `MM_CLOUD_API_KEY` that would allow mattermost-server/enterprise to make HTTP requests to CWS. We have chosen to leverage this API Key to allow bi-directional HTTP requests (ie, the API Key is now mutual, and can be used back and forth). 

This PR adds a new handler : `CloudApiKeyRequired` that will authenticate a request if the provided X-Cloud-Token matches the value in its environment variable `MM_CLOUD_API_KEY`

This PR does not contain any logic to use this new handler; I wanted to keep the focus on the handler flow itself.

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-29845

#### Release Note

I am not sure this release note should be publicly published?
```release-note
The Customer Web Server can now make HTTP requests to an installation by passing the X-Cloud-Token header, with the value being that installations Customer API Key.
```


web/handlers.go
```diff
@@ -190,13 +191,26 @@ func (h Handler) ServeHTTP(w http.ResponseWriter, r *http.Request) {
...
-				c.Err = model.NewAppError("ServeHTTP", "api.context.session_expired.app_error", nil, "token="+token, http.StatusUnauthorized)
+			if c.App.Srv().License() != nil && *c.App.Srv().License().Features.Cloud && tokenLocation == app.TokenLocationCloudHeader {
+				// Check to see if this provided token matches our CWS Token
+				session, err = c.App.GetCloudSession(token)
+				if err != nil {
+					c.Log.Info("Invalid CWS token", mlog.Err(err))
+					c.Err = err
+				} else {
+					c.App.SetSession(session)
+				}
```
marianunez, line 202: I wonder if we can move this specific case of creating a session into the `App.GetSession` and not act on an error case here

marianunez (changes requested):
Thanks @nickmisasi! Overall LGTM, just some requests for tests and minor changes

jupenur (approved):

web/handlers.go
ethervoid: Yes, probable makes sense to try to include that in the app package

web/handlers.go
```diff
@@ -190,13 +191,26 @@ func (h Handler) ServeHTTP(w http.ResponseWriter, r *http.Request) {
...
+					c.Err = err
+				} else if h.RequireSession {
+					c.RemoveSessionCookie(w, r)
+					c.Err = model.NewAppError("ServeHTTP", "api.context.session_expired.app_error", nil, "token="+token, http.StatusUnauthorized)
+				}
 			}
+
 		} else if !session.IsOAuth && tokenLocation == app.TokenLocationQueryString {
 			c.Err = model.NewAppError("ServeHTTP", "api.context.token_provided.app_error", nil, "token="+token, http.StatusUnauthorized)
 		} else {
```
ethervoid, line 216: Given that we don't have session in this case, probably we don't need the CSRFToken or handle in some other way

ethervoid (commented):
Some comments

app/session.go
```diff
@@ -34,6 +35,21 @@ func (a *App) CreateSession(session *model.Session) (*model.Session, *model.AppE
 	return session, nil
 }
 
+func (a *App) GetCloudSession(token string) (*model.Session, *model.AppError) {
+	apiKey := os.Getenv("MM_CLOUD_API_KEY")
+	if apiKey == token {
```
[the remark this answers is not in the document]
nickmisasi: Good call

app/session.go
```diff
@@ -34,6 +35,21 @@ func (a *App) CreateSession(session *model.Session) (*model.Session, *model.AppE
 	return session, nil
 }
 
+func (a *App) GetCloudSession(token string) (*model.Session, *model.AppError) {
```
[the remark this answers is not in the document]
nickmisasi: Had to refactor the test function to allow for the new header type/location but think it should be covered now.

app/authentication.go
```diff
@@ -281,5 +284,9 @@ func ParseAuthTokenFromRequest(r *http.Request) (string, TokenLocation) {
 		return token, TokenLocationQueryString
 	}
 
+	if token := r.Header.Get(model.HEADER_CLOUD_TOKEN); token != "" {
+		return token, TokenLocationCloudHeader
+	}
```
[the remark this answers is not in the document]
nickmisasi: Had to refactor the test function to allow for the new header type/location but think it should be covered now.

api4/handlers.go
```diff
@@ -51,6 +51,26 @@ func (api *API) ApiSessionRequired(h func(*Context, http.ResponseWriter, *http.R
...
+// CloudApiKeyRequired provides a handler for webhook endpoints to access Cloud installations from CWS
+func (api *API) CloudApiKeyRequired(h func(*Context, http.ResponseWriter, *http.Request)) http.Handler {
+	handler := &web.Handler{
+		GetGlobalAppOptions: api.GetGlobalAppOptions,
+		HandleFunc:          h,
+		HandlerName:         web.GetHandlerName(h),
+		RequireSession:      false,
+		RequireCloudKey:     true,
+		TrustRequester:      false,
+		RequireMfa:          true,
```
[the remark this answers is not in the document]
nickmisasi: Nope!

web/handlers.go
nickmisasi: I've changed up this code to be a little less weird, you can see it in the latest diff, I think it's much cleaner than the first go.

I am hesitant to include this logic inside the `App.GetSession` function because that function has no notion of where the token came from, so conditionals will be challenging. We could pass the `tokenLocation` variable into `GetSession` to allow the logic to live entirely in that function. However, we'd then be changing the signature of a function used 17 times inside mattermost-server, and potentially more in plugins. That sounds like a dangerous change I'd like to avoid doing, especially in such a fundamental-to-the-app function like `GetSession`

web/handlers.go
nickmisasi: This is no longer applicable as of the latest diff. Going to resolve this one and we can discuss here: https://github.com/mattermost/mattermost-server/pull/16319#discussion_r527795317

web/handlers.go
```diff
@@ -197,6 +198,7 @@ func (h Handler) ServeHTTP(w http.ResponseWriter, r *http.Request) {
 				c.RemoveSessionCookie(w, r)
 				c.Err = model.NewAppError("ServeHTTP", "api.context.session_expired.app_error", nil, "token="+token, http.StatusUnauthorized)
 			}
+
```
marianunez, line 201: Nit: new line

marianunez (approved):
Tests look great! Thanks!

web/handlers.go
marianunez: Looks much cleaner with the latest changes and +1 on changing the `App.GetSession` signature`

agnivade (approved):
Just a small suggestion, but nothing blocking.

ethervoid (approved):
LGTM

web/handlers.go
```diff
@@ -209,6 +210,15 @@ func (h Handler) ServeHTTP(w http.ResponseWriter, r *http.Request) {
 		}
 
 		h.checkCSRFToken(c, r, token, tokenLocation, session)
+	} else if len(token) != 0 && c.App.Srv().License() != nil && *c.App.Srv().License().Features.Cloud && tokenLocation == app.TokenLocationCloudHeader {
+		// Check to see if this provided token matches our CWS Token
+		session, err := c.App.GetCloudSession(token)
+		if err != nil {
+			c.Log.Info("Invalid CWS token", mlog.Err(err))
```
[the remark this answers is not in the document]
nickmisasi: Good point!

PR #16322, opened 2020-11-19
Remove unneeded Err field in the store result struct

jespino:
#### Summary
Remove unneeded Err field in the store result struct.

This is part of the clean up after the AppError removal from the store.

### DIFF
--- a/store/store.go
+++ b/store/store.go
@@ -14,7 +14,6 @@ import (
 
 type StoreResult struct {
 	Data interface{}
-	Err  *model.AppError
 
 	// NErr a temporary field used by the new code for the AppError migration. This will later become Err when the entire store is migrated.
 	NErr error
--- a/app/notification.go
+++ b/app/notification.go
@@ -15,6 +15,7 @@ import (
 	"github.com/mattermost/mattermost-server/v5/store"
 	"github.com/mattermost/mattermost-server/v5/utils"
 	"github.com/mattermost/mattermost-server/v5/utils/markdown"
+	"github.com/pkg/errors"
 )
 
 func (a *App) SendNotifications(post *model.Post, team *model.Team, channel *model.Channel, sender *model.User, parentPostList *model.PostList, setOnline bool) ([]string, error) {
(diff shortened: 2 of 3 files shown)


store/store.go
```diff
@@ -14,7 +14,6 @@ import (
 
 type StoreResult struct {
 	Data interface{}
-	Err  *model.AppError
 
 	// NErr a temporary field used by the new code for the AppError migration. This will later become Err when the entire store is migrated.
 	NErr error
```
rvillablanca, line 19: Two things:

* The comment is unnecessary now so we can get rid of it.
* Since `NErr`  is the only error field in the `StoreResult`, we could refactor this and rename it to Err (I can help with this if you prefer)

rvillablanca (changes requested):
Just a couple of minor suggestions, otherwise it looks good to me. Thanks for your contribution @jespino :woman_dancing: :yum:

store/store.go
agnivade: Yes, I agree. we should do the opposite. Change `Err` to be `error`.

jespino: Actually, this is the first one, that only remove the Err to verify it there are still usage of this, another one can do the renaming.

agnivade: Sounds good to me.

agnivade (approved):

store/store.go
jespino: @rvillablanca I think we can remove the comment in the other PR, is going to be more obvious what is going on.

rvillablanca: Sure

rvillablanca (approved):
LGTM

jespino:
@lindy65 The tests of this is very specific, the notifications for (ldap) groups should keep working correctly. A simple verification of ldap group notification been sent should be enough to test it.

lindy65:
Thanks @jespino :)  Can you link me to the Jira ticket for this PR too please?  Thanks!

jespino:
@lindy65 actually, there is not a jira ticket for this, but I going to create it, is part of a refactor, so the ticket didn't exists.

lindy65:
Thanks @jespino 👍

jespino:
@lindy65 sorry the delay, here: https://mattermost.atlassian.net/browse/MM-30843

lindy65 (approved):
Thanks @jespino - looks good to me 👍

PR #16323, opened 2020-11-19
MM-30800: prepackage incident management 1.1

lieut-data:
#### Summary
A few releases back, we pre-packaged incident _response_ v0.6, cherry-picking those changes into the `cloud-beta` branch for an early rollout. v5.28 and v5.29 did not pre-package incident response, since they were cut just after that commit in `master`, and we had intentions to revisit this prior to v5.30.

Since that time, we renamed the product, released a few more versions, and cherry-picked into `cloud-ga`. All of these changes have been in "production" for a while now, but we need to bring `master` back up-to-date, since a new `cloud` branch is pending release to production but it's still pointing at IR v0.6 instead of IM v1.1.

When IM v1.2 is released, we'll bump the version again, but this commit is important to include in the upcoming release of the `cloud` branch on Tuesday, November 24th.

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-30800

#### Release Note
```release-note
Pre-package and enable incident management v1.1.1
```

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -95,7 +95,7 @@ PLUGIN_PACKAGES += mattermost-plugin-antivirus-v0.1.2
 PLUGIN_PACKAGES += mattermost-plugin-jira-v2.3.2
 PLUGIN_PACKAGES += mattermost-plugin-gitlab-v1.1.0
 PLUGIN_PACKAGES += mattermost-plugin-jenkins-v1.0.0
-PLUGIN_PACKAGES += mattermost-plugin-incident-management-v1.1.0
+PLUGIN_PACKAGES += mattermost-plugin-incident-management-v1.1.1
 
 # Prepares the enterprise build if exists. The IGNORE stuff is a hack to get the Makefile to execute the commands outside a target
 ifeq ($(BUILD_ENTERPRISE_READY),true)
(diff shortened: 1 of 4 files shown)


lieut-data:
@prapti, I'm assigning you to this one before dev review, because I don't think there's anything to test here, and would like to get this into `master` and then immediately into `cloud` to ensure we don't start deploying v0.6 to new cloud instances next week. Do you agree we can skip QA testing here?

Makefile
```diff
@@ -95,7 +95,7 @@ PLUGIN_PACKAGES += mattermost-plugin-antivirus-v0.1.2
 PLUGIN_PACKAGES += mattermost-plugin-jira-v2.3.2
 PLUGIN_PACKAGES += mattermost-plugin-gitlab-v1.1.0
 PLUGIN_PACKAGES += mattermost-plugin-jenkins-v1.0.0
-PLUGIN_PACKAGES += mattermost-plugin-incident-response-v0.6.0
+PLUGIN_PACKAGES += mattermost-plugin-incident-management-v1.1.0
```
cpoile, line 98: v1.1.1? (for the telemetry fix)

cpoile (approved):
🎉

Makefile
lieut-data: Oh, goodness -- thank you!

crspeller (approved):

prapti (approved):
Yes, I think that totally makes sense @lieut-data. I can do a quick check after things are updated in cloud. Approving.

PR #16351, opened 2020-11-20
[MM-29845] Add CWS Webhook endpoint and payment failed email

nickmisasi:
#### Summary
When a payment fails in stripe, a webhook will be sent to CWS, and then to this new endpoint in the installation. This endpoint will trigger an email to all admins notifying them of the failed payment.

This PR is based off of #16319 as that PR contains a new authentication handler needed for this PR to work. 

### Testing
You can test this by creating an admin user for yourself (with your email) on the spinwick, and making a POST request to:
/api/v4/cloud/webhook
With body:
```
{
    "event": "failed-payment",
    "failed_payment": {
        "card_brand": "Visa",
        "last_four": 4242,
        "failure_message": "Credit card was declined."
    }
}
```
The failure_code can be anything from [this list](https://stripe.com/docs/declines/codes)

Here's a CURL command you can paste into your terminal to quickly test:
```
curl --location --request POST 'https://mattermost-server-pr-16351.test.mattermost.cloud/api/v4/cloud/webhook' \
--header 'X-Cloud-Token: e1d9yoexe3by9eg49guse4jkyw' \
--header 'Content-Type: application/json' \
--data-raw '{
    "event": "failed-payment",
    "failed_payment": {
        "card_brand": "Visa",
        "last_four": 4242,
        "failure_message": "Credit card was declined."
    }
}'
```

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-29845

#### Release Note
```release-note
System administrators can receive an email notifying them of a failed payment in mattermost cloud
```
image


nickmisasi:
@matthewbirtch will need a review on this one at some point, but had a question for you WRT [this comment](https://docs.google.com/document/d/1ZKwROOv0I6vzA_0oUVU026U2gHi01iUJXNW94jyTsTk/edit?disco=AAAAKV-p9yM) on the doc. 

We _do_ have access to the reason that a transaction failed. However, the reason is sent to us non-localized (in english), meaning that we would only be able to put it into the email in English. The reason this is a problem is currently we are sending the email to the administrators in the language that each administrator has set as their preferred locale. So in the event of a French administrator, they'd see the whole email in french except for the reason, which would be in English. 

Is this fine? Alternatively, we could just display a more generic "Your payment was declined" in that section, and it would be localized properly. This might lose some useful information to an English speaking administrator.

matthewbirtch:
> We _do_ have access to the reason that a transaction failed. However, the reason is sent to us non-localized (in english), meaning that we would only be able to put it into the email in English. The reason this is a problem is currently we are sending the email to the administrators in the language that each administrator has set as their preferred locale. So in the event of a French administrator, they'd see the whole email in french except for the reason, which would be in English.
> 
> Is this fine? Alternatively, we could just display a more generic "Your payment was declined" in that section, and it would be localized properly. This might lose some useful information to an English speaking administrator.

It's definitely weird to show the majority of content in the preferred language, but the reason for declined payment (which may be the most important part of the email) in English. I guess it's better to have it in English than no reason at all. Is there now way to get the reason in the preferred language? Surely Stripe has something to account for this?

nickmisasi:
@matthewbirtch Did a little more research into this, and there is no way to get a localized version from Stripe (Even if they did, it wouldn't technically be helpful because the event is processed in CWS, and then sent to the installation, meaning CWS doesn't have any notion of which locales it needs)

I've found these: https://stripe.com/docs/error-codes which seem to give you a way to map these errors to something localized on your own. I'm hesitant to have to keep a list of stripe codes that could possibly change/not all be user facing. Going to loop Adam and Maria on that part. 

> I guess it's better to have it in English than no reason at all.
This is basically what I was wondering in the event the above option doesn't pan out

marianunez:
> @matthewbirtch Did a little more research into this, and there is no way to get a localized version from Stripe (Even if they did, it wouldn't technically be helpful because the event is processed in CWS, and then sent to the installation, meaning CWS doesn't have any notion of which locales it needs)
> 
> I've found these: [stripe.com/docs/error-codes](https://stripe.com/docs/error-codes) which seem to give you a way to map these errors to something localized on your own. I'm hesitant to have to keep a list of stripe codes that could possibly change/not all be user facing. Going to loop Adam and Maria on that part.
> 
> > I guess it's better to have it in English than no reason at all.
> 
> This is basically what I was wondering in the event the above option doesn't pan out

I am inclined to leave the error message in English and we can later improve with the error codes if it makes senses/worth it in the future. I believe that given that this email is directed to admins, we can assume that they would understand at least the reason of the payment failure in English.

nickmisasi:
>I am inclined to leave the error message in English and we can later improve with the error codes if it makes senses/worth it in the future. I believe that given that this email is directed to admins, we can assume that they would understand at least the reason of the payment failure in English.

@marianunez I've changed this PR to have this behaviour for now.

nickmisasi:
@stevemudie @matthewbirtch matterwick _finally_ started cooperating so this one should be ready for a test/design review. I've left instructions in the PR on how to test It, but feel free to shoot me a DM and I can trigger the email to send once you've got your accounts setup!

matthewbirtch (approved):
Looks good to me

api4/cloud.go
```diff
@@ -324,3 +327,33 @@ func getSubscriptionInvoicePDF(c *Context, w http.ResponseWriter, r *http.Reques
 		return
 	}
 }
+
+func handleCWSWebhook(c *Context, w http.ResponseWriter, r *http.Request) {
+	if c.App.Srv().License() == nil || !*c.App.Srv().License().Features.Cloud {
+		c.Err = model.NewAppError("Api4.getSuscriptionInvoicePDF", "api.cloud.license_error", nil, "", http.StatusNotImplemented)
```
marianunez, line 333:
```suggestion
		c.Err = model.NewAppError("Api4.handleCWSWebhook", "api.cloud.license_error", nil, "", http.StatusNotImplemented)
```

api4/cloud.go
```diff
@@ -324,3 +327,33 @@ func getSubscriptionInvoicePDF(c *Context, w http.ResponseWriter, r *http.Reques
...
+		return
+	}
+
+	bodyBytes, err := ioutil.ReadAll(r.Body)
+	if err != nil {
+		c.Err = model.NewAppError("Api4.confirmCustomerPayment", "api.cloud.app_error", nil, err.Error(), http.StatusInternalServerError)
```
marianunez, line 339:
```suggestion
		c.Err = model.NewAppError("Api4.handleCWSWebhook", "api.cloud.app_error", nil, err.Error(), http.StatusInternalServerError)
```

api4/cloud.go
```diff
@@ -324,3 +327,33 @@ func getSubscriptionInvoicePDF(c *Context, w http.ResponseWriter, r *http.Reques
...
+		return
+	}
+
+	var event *model.CWSWebhook
+	if err = json.Unmarshal(bodyBytes, &event); err != nil {
+		c.Err = model.NewAppError("Api4.confirmCustomerPayment", "api.cloud.app_error", nil, err.Error(), http.StatusInternalServerError)
```
marianunez, line 345:
```suggestion
		c.Err = model.NewAppError("Api4.handleCWSWebhook", "api.cloud.app_error", nil, err.Error(), http.StatusInternalServerError)
```

model/cloud.go
```diff
@@ -112,3 +112,14 @@ type InvoiceLineItem struct {
 	Type         string                 `json:"type"`
 	Metadata     map[string]interface{} `json:"metadata"`
 }
+
+type CWSWebhook struct {
+	Event         string         `json:"event"`
+	FailedPayment *FailedPayment `json:"failed_payment"`
```
marianunez, line 118: It's ok for now given this is the only case we have, but in the future we may have this be a generic payload that we can unmarshall to the struct we need based on the `Event` string

marianunez (approved):
Thanks @nickmisasi! Minor comments below

model/cloud.go
nickmisasi: My thinking was we could just keep adding structs to this and let jsonUnmarshal handle the legwork there. Just need to be intelligent enough with the code that an Event with type `failed-payment` will only have the FailedPayment struct populated, etc.

stevemudie (approved):
Worked as expected. Thanks for the easy to follow steps, Nick!

api4/cloud.go
```diff
@@ -324,3 +327,33 @@ func getSubscriptionInvoicePDF(c *Context, w http.ResponseWriter, r *http.Reques
...
 	}
 }
+
+func handleCWSWebhook(c *Context, w http.ResponseWriter, r *http.Request) {
+	if c.App.Srv().License() == nil || !*c.App.Srv().License().Features.Cloud {
+		c.Err = model.NewAppError("Api4.handleCWSWebhook", "api.cloud.license_error", nil, "", http.StatusNotImplemented)
+		return
+	}
+
+	bodyBytes, err := ioutil.ReadAll(r.Body)
```
ethervoid, line 337: Include the defer body.close please

api4/cloud.go
```diff
@@ -324,3 +327,33 @@ func getSubscriptionInvoicePDF(c *Context, w http.ResponseWriter, r *http.Reques
...
+	}
+
+	var event *model.CWSWebhook
+	if err = json.Unmarshal(bodyBytes, &event); err != nil {
+		c.Err = model.NewAppError("Api4.handleCWSWebhook", "api.cloud.app_error", nil, err.Error(), http.StatusInternalServerError)
+		return
+	}
+
+	switch event.Event {
+	case "failed-payment":
```
ethervoid, line 350: Please use a constant

api4/cloud.go
```diff
@@ -324,3 +327,33 @@ func getSubscriptionInvoicePDF(c *Context, w http.ResponseWriter, r *http.Reques
...
+		return
+	}
+
+	bodyBytes, err := ioutil.ReadAll(r.Body)
+	if err != nil {
+		c.Err = model.NewAppError("Api4.handleCWSWebhook", "api.cloud.app_error", nil, err.Error(), http.StatusInternalServerError)
+		return
+	}
+
+	var event *model.CWSWebhook
```
ethervoid, line 343: 2/5 to rename this to `CWSWebhookPayload`

app/cloud.go
```diff
@@ -4,9 +4,20 @@
 package app
 
 import (
+	"github.com/mattermost/mattermost-server/v5/mlog"
 	"github.com/mattermost/mattermost-server/v5/model"
 )
 
+func (a *App) getEmailRecipients() ([]*model.User, *model.AppError) {
```
ethervoid, line 11:
```suggestion
func (a *App) getSysAdminsEmailRecipients() ([]*model.User, *model.AppError) {
```

app/email.go
```diff
@@ -766,3 +766,25 @@ func (es *EmailService) SendSuspensionEmailToSupport(email string, installationI
...
+	bodyPage.Props["Title"] = T("api.templates.payment_failed.title")
+	bodyPage.Props["Info1"] = T("api.templates.payment_failed.info1", map[string]interface{}{"CardBrand": failedPayment.CardBrand, "LastFour": failedPayment.LastFour})
+	bodyPage.Props["Info2"] = T("api.templates.payment_failed.info2")
+	bodyPage.Props["Info3"] = T("api.templates.payment_failed.info3")
+	bodyPage.Props["Button"] = T("api.templates.over_limit_fix_now")
+
+	bodyPage.Props["FailedReason"] = failedPayment.FailureMessage
+
+	if err := es.sendMail(email, subject, bodyPage.Render()); err != nil {
+		return false, model.NewAppError("SendOverUserLimitWarningEmail", "api.user.send_password_reset.send.app_error", nil, "err="+err.Message, http.StatusInternalServerError)
```
ethervoid, line 786:
```suggestion
		return false, model.NewAppError("SendPaymentFailedEmail", "api.user.send_password_reset.send.app_error", nil, "err="+err.Message, http.StatusInternalServerError)
```

ethervoid (changes requested):

ethervoid (approved):
LGTM!

PR #16354, opened 2020-11-23
MM-30319: Fix race in ensureAsymmetricSigningKey

agnivade:
A cluster changed listener always runs in a separate goroutine.
This causes a race condition when a new member gets added in the cluster.

We fix it by accessing the value atomically.

I tried to write a test for this, but it wasn't triggering at all.
The best way is to just start the server with cluster enabled in race mode,
and verify that the race doesn't happen.

https://mattermost.atlassian.net/browse/MM-30319

```release-notes
NONE
```


streamer45 (approved):
Nice :+1:

isacikgoz (approved):
LGTM 👍

PR #16355, opened 2020-11-23
Removing supplier concept from the sql store

jespino:
#### Summary
Removing supplier concept from the sql store

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-30897

#### Release Note
No release notes


store/sqlstore/store.go
```diff
@@ -4,6 +4,1346 @@
...
+
+func (t *TraceOnAdapter) Printf(format string, v ...interface{}) {
+	originalString := fmt.Sprintf(format, v...)
+	newString := strings.ReplaceAll(originalString, "\n", " ")
+	newString = strings.ReplaceAll(newString, "\t", " ")
+	newString = strings.ReplaceAll(newString, "\"", "")
+	mlog.Debug(newString)
+}
+
+func NewSqlStore(settings model.SqlSettings, metrics einterfaces.MetricsInterface) *SqlStore {
```
agnivade, line 140: While we are here, I would suggest to change the name to just `Store` to avoid the stutter in `sqlstore.NewSqlStore`. It should just be `sqlstore.New`, and the struct should be `Store`.

https://golang.org/doc/effective_go.html#package-names

> The importer of a package will use the name to refer to its contents, so exported names in the package can use that fact to avoid stutter.

> For instance, the buffered reader type in the bufio package is called Reader, not BufReader, because users see it as bufio.Reader, which is a clear, concise name.

agnivade (commented):
Just some small comments. Looks amazing.

store/sqlstore/store.go
jespino: I totally agree with the `New` method name, but I don't know if the `sqlstore.Store` is going to be confusing with `store.Store`, for this case I actually prefer the `SqlStore`.

agnivade (approved):

store/sqlstore/store.go
agnivade: Ok.

mgdelacroix (approved):
LGTM! :+1:

jespino:
@lindalumitchell This is the kind of refactor that affects a lot of parts, I think we should merge it and test it as part of the release tests.

lindalumitchell (approved):

PR #16357, opened 2020-11-23
MM-30706: Skip racy test TestWebSocket

agnivade:
We just skip the test for now

https://mattermost.atlassian.net/browse/MM-30706

```release-note
NONE
```

### DIFF
--- a/api4/websocket_norace_test.go
+++ b/api4/websocket_norace_test.go
@@ -1,0 +1,63 @@
+// +build !race
+// Copyright (c) 2015-present Mattermost, Inc. All Rights Reserved.
+// See LICENSE.txt for license information.
+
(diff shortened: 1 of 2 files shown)


api4/websocket_norace_test.go
```diff
@@ -0,0 +1,63 @@
+// +build !race
```
streamer45, line 1: The linter didn't like this.

streamer45 (approved):
Looks good :+1:

isacikgoz (approved):

api4/websocket_norace_test.go
agnivade: Ah, now I have to fix the linter.

PR #16359, opened 2020-11-23
Return when merging config errors

nronas:
#### Summary
Fixes an error reported https://mattermost.atlassian.net/browse/MM-30873 that was caused due to an unsafe pointer access.

#### Ticket Link
Fixes https://mattermost.atlassian.net/browse/MM-30873

#### Release Note

```release-note
Fixes panic on GetConfig, UpdateConfig and PatchConfig
```

I've tried to write some tests that will pass an empty config, but with our current setup, the tests require a config to be present(regardless that the config is a pointer to a struct, which can be nil). Looked around the code for prior art on testing missing(nil) configs but I run out empty. Any hints on how to cover this with specs would be much appreciated.

### DIFF
--- a/api4/config.go
+++ b/api4/config.go
@@ -44,7 +44,7 @@ func init() {
 	}
 }
 
-func getConfig(c *Context, w http.ResponseWriter, r *http.Request) {
+func getConfig(c *Context, w http.ResponseWriter, _ *http.Request) {
 	if !c.App.SessionHasPermissionToAny(*c.App.Session(), model.SysconsoleReadPermissions) {
 		c.SetPermissionError(model.SysconsoleReadPermissions...)
 		return
@@ -60,6 +60,7 @@ func getConfig(c *Context, w http.ResponseWriter, r *http.Request) {
 	})
 	if err != nil {
 		c.Err = model.NewAppError("getConfig", "api.config.get_config.restricted_merge.app_error", nil, err.Error(), http.StatusInternalServerError)
+		return
 	}
 
 	auditRec.Success()
@@ -125,6 +126,7 @@ func updateConfig(c *Context, w http.ResponseWriter, r *http.Request) {
 	})
 	if err1 != nil {
 		c.Err = model.NewAppError("updateConfig", "api.config.update_config.restricted_merge.app_error", nil, err1.Error(), http.StatusInternalServerError)
+		return
 	}
 
 	// Do not allow plugin uploads to be toggled through the API
@@ -154,6 +156,7 @@ func updateConfig(c *Context, w http.ResponseWriter, r *http.Request) {
 	})
 	if mergeErr != nil {
 		c.Err = model.NewAppError("getConfig", "api.config.update_config.restricted_merge.app_error", nil, err.Error(), http.StatusInternalServerError)
+		return
 	}
 
 	auditRec.Success()
@@ -264,6 +267,7 @@ func patchConfig(c *Context, w http.ResponseWriter, r *http.Request) {
 	})
 	if mergeErr != nil {
 		c.Err = model.NewAppError("getConfig", "api.config.patch_config.restricted_merge.app_error", nil, err.Error(), http.StatusInternalServerError)
+		return
 	}
 
 	w.Header().Set("Cache-Control", "no-cache, no-store, must-revalidate")


ethervoid (approved):
LGTM! Yes, to be able to test the configuration alone we need to move to use the [ConfigService](https://github.com/mattermost/mattermost-server/blob/master/services/configservice/configservice.go) so we can mock it up

agnivade:
Thanks @nronas! Yep, let's add a test following Mario's suggestion. Do let me know if that doesn't work somehow and I can take a look.

nronas:
@ethervoid @agnivade just to be sure that we are on the same page, you are saying to not cover with tests the `client.GetConfig()` instead to test the private method `getConfig` in the api4 and i can pass a context with a ConfigService in the state that we want.

agnivade:
I'll take a look and let you know. :+1:

agnivade:
It's possible, but will take some effort to achieve this. We need to call getConfig directly with our own `Context`, passing a mocked version of `App` which returns nil in `GetSanitizedConfig`. 

Something like

```
c := &Context{}
c.App = mockedAppInterfaceReturningNilSanitizedConfig

getConfig(c, w, _)
```

But more importantly, I am not able to reproduce this panic by manually returning `nil` in `GetSanitizedConfig`. The only error that can come from utils.Merge is the reflect check, and due to that, I get

```
http: panic serving 127.0.0.1:37289: reflect: call of reflect.Value.Field on zero Value\ngoroutine 3804 [running]:\nnet/http.(*conn).serve.func1(0xc000e5b860)\n\t/usr/local/go/src/net/http/server.go:1800 +0x139\npanic(0x200fe60, 0xc00219b2e0)\n\t/usr/local/go/src/runtime/panic.go:975 +0x3e3\nreflect.Value.Field(0x0, 0x0, 0x0, 0x0, 0x22db420, 0xc003a44000, 0x199)\n\t/usr/local/go/src/reflect/value.go:834 +0x123\ngithub.com/mattermost/mattermost-server/v5/utils.merge(0x22aa560, 0xc003a44000, 0x199, 0x0, 0x0, 0x0, 0xc003a4d1d0, 0x0, 0x0, 0x0, ...)\n\t/home/agniva/mattermost/mattermost-server/utils/merge.go:77 +0x17c8\ngithub.com/mattermost/mattermost-server/v5/utils.Merge(0x21aa600, 0xc003a44000, 0x21aa600, 0x0, 0xc003a4d1d0, 0x0, 0x0, 0x0, 0x0)\n\t/home/agniva/mattermost/mattermost-server/utils/merge.go:58 +0x43d\ngithub.com/mattermost/mattermost-server/v5/config.Merge(0xc003a44000, 0x0, 0xc003a4d1d0, 0x2301e51, 0x4, 0xc00365e480)\n\t/home/agniva/mattermost/mattermost-server/config/utils.go:163 +0x6e\ngithub.com/mattermost/mattermost-server/v5/api4.getConfig(0xc002027500, 0x283e3a0, 0xc0020274c0, 0xc003a00800)\n\t/home/agniva/mattermost/mattermost-server/api4/config.go:56 +0x238\ngithub.com/mattermost/mattermost-server/v5/web.Handler.ServeHTTP(0xc0016dc720, 0x23f0f60, 0x38dedb8, 0x9, 0x10001, 0x0, 0x0, 0x2840de0, 0xc003a31490, 0xc003a00800)\n\t/home/agniva/mattermost/mattermost-server/web/handlers.go:246 +0x1e02\ngithub.com/NYTimes/gziphandler.GzipHandlerWithOpts.func1.1(0x2840820, 0xc003a1a1c0, 0xc003a00800)\n\t/home/agniva/mattermost/mattermost-server/vendor/github.com/NYTimes/gziphandler/gzip.go:337 +0x211\nnet/http.HandlerFunc.ServeHTTP(0xc0036a6c60, 0x2840820, 0xc003a1a1c0, 0xc003a00800)\n\t/usr/local/go/src/net/http/server.go:2041 +0x44\ngithub.com/gorilla/mux.(*Router).ServeHTTP(0xc00217c000, 0x2840820, 0xc003a1a1c0, 0xc003a00600)\n\t/home/agniva/mattermost/mattermost-server/vendor/github.com/gorilla/mux/mux.go:210 +0xe2\nnet/http.serverHandler.ServeHTTP(0xc0002d07e0, 0x2840820, 0xc003a1a1c0, 0xc003a00600)\n\t/usr/local/go/src/net/http/server.go:2836 +0xa3\nnet/http.(*conn).serve(0xc000e5b860, 0x28496e0, 0xc0020273c0)\n\t/usr/local/go/src/net/http/server.go:1924 +0x86c\ncreated by net/http.(*Server).Serve\n\t/usr/local/go/src/net/http/server.go:2962 +0x35c"
```

@nronas - I am wondering if I am missing something here? Did you verify this panic in some other way?

nronas:
@agnivade thanks for looking into this. 

> It's possible, but will take some effort to achieve this. We need to call getConfig directly with our own Context, passing a mocked version of App which returns nil in GetSanitizedConfig.

Yeap that is exactly what I was thinking too 👍 

> @nronas - I am wondering if I am missing something here? Did you verify this panic in some other way?

Maybe my reproduction had a false positive since I remember getting a panic when I was passing nil to the merge, but I've re-run this and you are 💯 right the merge does not return an error when nil is passed. To be completely honest I didn't spent enough time reproducing since my editor has an unsafe pointer access detection and was matching the line of the panic that was reported in the ticket.

The panic reported was in line 68 on the config which I guess we need to find which sha of the code the pod was running. The only other part on that line that I can see that it could fail with nil pointer is `*c.App.Srv().License().Features.Cloud` when `Features` is nil.

ethervoid:
> It's possible but will take some effort to achieve this. We need to call getConfig directly with our own Context, passing a mocked version of App which returns nil in GetSanitizedConfig.

Yep!, I took a look this morning and I agree with that. Probably we can, in the future, try to have the `ConfigStore` as a dependency in the server creation instead of processing it inside the [NewServer function](https://github.com/mattermost/mattermost-server/blob/master/app/server.go#L207-L218)

Regarding the panic, I think the root cause of this problem is [this one](https://github.com/mattermost/mattermost-server/pull/16365)

nronas:
@ethervoid superb, i guess this is not needed anymore. Superseded from https://github.com/mattermost/mattermost-server/pull/16365

PR #16364, opened 2020-11-24
Add Billing Permissions

hahmadia:
#### Summary

Adds permissions for billing section in admin definition 

#### Ticket Link

https://mattermost.atlassian.net/browse/MM-30950

#### Release Note

```release-note
Adds billing permissions for billing section of admin definition
```


hahmadia:
@sbishel @marianunez 
Could you guys help provide what the correct description should be for the billing section? 

image

marianunez:
> Could you guys help provide what the correct description should be for the billing section?

I'll defer this to @matthewbirtch that surely can provide a better text that I can. Matt, this is for the new System Roles page to grant permissions to the System Console Billing area:

Screen Shot 2020-11-24 at 11 30 32 AM

matthewbirtch:
> I'll defer this to @matthewbirtch that surely can provide a better text that I can. Matt, this is for the new System Roles page to grant permissions to the System Console Billing area:

No problem. Here you go:

**Billing**
Access subscription details, billing history, company information and payment information.

model/permission.go
```diff
@@ -695,6 +698,18 @@ func initializePermissions() {
...
+	PERMISSION_SYSCONSOLE_READ_BILLING = &Permission{
+		"sysconsole_read_billing",
+		"authentication.permissions.use_group_mentions.name",
+		"authentication.permissions.use_group_mentions.description",
+		PermissionScopeSystem,
+	}
+	PERMISSION_SYSCONSOLE_WRITE_BILLING = &Permission{
+		"sysconsole_write_billing",
+		"authentication.permissions.use_group_mentions.name",
+		"authentication.permissions.use_group_mentions.description",
```
sbishel, line 710: Should the "use_group_mentions.name" and "use_group_mentions.description" be changed

app/permissions_migrations.go
```diff
@@ -516,6 +516,15 @@ func (a *App) getAddManageSharedChannelsPermissionsMigration() (permissionsMap,
...
 }
 
+func (a *App) getBillingPermissionsMigration() (permissionsMap, error) {
+	return permissionsMap{
+		permissionTransformation{
+			On:  isRole(model.SYSTEM_ADMIN_ROLE_ID),
+			Add: []string{model.PERMISSION_SYSCONSOLE_READ_BILLING.Id, model.PERMISSION_SYSCONSOLE_WRITE_BILLING.Id},
+		},
+	}, nil
+}
```
sbishel, line 526: Does this need added to any of the new roles?

hahmadia: I wasn't sure either but if it's not there by default , I was thinking it's not too hard for the system admin to assign it to those roles.

marianunez (approved):
LGTM, thanks again @hahmadia!

model/permission.go
hahmadia: To be honest, I'm not sure. If it does need to be changed, what should it be changed too?

app/permissions_migrations.go
hahmadia: @marianunez Any comments on this? Do you think by default any other role should have this permission?

model/permission.go
sbishel: I'm not familiar with how it is used, please investigate. The names just didn't look right to me.

hahmadia: Done. As discussed, passing empty string.

sbishel (approved):

app/permissions_migrations.go
marianunez: For the purposes of this PR, I agree that Admin would be enough and they can grant permissions to other roles as needed. We do have the concept Billing Admins as a feature in the roadmap so I will work with PM to see how this would fit with these new System Roles and if we would want to modify any others as well.

hahmadia: Cool, thanks for the quick response Maria!

hahmadia:
@furqanmlk 
You can test the new permission directly in the cloud server created above.

furqanmlk (approved):
Tested and verified with Read-Only / User Manager / System Manager / .
Working as expected

PR #16365, opened 2020-11-24
MM-30900 - Dropped condition of RestrictAdmin for filtering cloud_restrictable

marianunez:
#### Summary

5.30 introduces new System Roles (i.e. System Manager) that by default don't have read/write permissions to `ExperimentalSettings`. For cloud servers, when getting and updating the config, we were checking `ExperimentalSettings.RestrictedAdmin` to filter out `cloud_restrictable` settings. This was causing a panic for these new roles. See related bug. 

This PR drops the check for `RestrictedAdmin` and just filters `cloud_restrictable` for any cloud server to avoid these cases where roles can be granted/removed the Experimental Settings by admins in the System Console. 

#### Ticket Link
[MM-30900](https://mattermost.atlassian.net/browse/MM-30900)

#### Release Note
```release-note
NONE
```

### DIFF
--- a/api4/config.go
+++ b/api4/config.go
@@ -65,7 +65,7 @@ func getConfig(c *Context, w http.ResponseWriter, r *http.Request) {
 	auditRec.Success()
 
 	w.Header().Set("Cache-Control", "no-cache, no-store, must-revalidate")
-	if c.App.Srv().License() != nil && *c.App.Srv().License().Features.Cloud && *cfg.ExperimentalSettings.RestrictSystemAdmin {
+	if c.App.Srv().License() != nil && *c.App.Srv().License().Features.Cloud {
 		w.Write([]byte(cfg.ToJsonFiltered(model.ConfigAccessTagType, model.ConfigAccessTagCloudRestrictable)))
 	} else {
 		w.Write([]byte(cfg.ToJson()))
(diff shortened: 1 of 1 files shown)


marianunez:
> Mattermost test server created! 🎉
> 
> Access here: [mattermost-server-pr-16365.test.mattermost.cloud](https://mattermost-server-pr-16365.test.mattermost.cloud)
> 
> Account Type	Username	Password
> Admin	sysadmin	Sys@dmin123
> User	user-1	User-1@123

@furqanmlk I've added the cloud license to the test server so you can verify the fix in a cloud environment.

ethervoid (approved):
LGTM! But we should think if we want that option moved out of the `ExperimentalSettings` and have it in the `ServiceSettings` so we don't have this problem.

RIght now we're checking more options with `RestrictSystemAdmin`

nronas:
@ethervoid @marianunez What is the expected logical accuracy for the above case? Do we want to respect the `RestrictedSystemAdmin` flag when present?

ethervoid:
@nronas Not is needed by now so the fix is enough but it'd be great to have that flag present, I think could be useful in the future and doesn't do any harm :)

marianunez:
> @ethervoid @marianunez What is the expected logical accuracy for the above case? Do we want to respect the `RestrictedSystemAdmin` flag when present?

The `RestrictSystemAdmin` is always true in a cloud server so checking it after checking a cloud license was redundant in a sense.

@ethervoid I agree that we check this setting in other places and may no be tied to a license check like here. We should initially take a look at that on any other possibilities of panics for any user that doesn't have permission over those settings.

We could also move this setting out of `ExperimentalSettings` however I believe those users should be restricted only for access to the System Console Experimental Settings and not on the server side. They still should have read access there. @mkraft any thoughts on this?

jwilander (approved):

marianunez:
@furqanmlk Given the urgency to include this in today's release, I am going to merge to get a new build and we can validate once on the test environment.

furqanmlk (commented):
Working as expected

marianunez:
> Working as expected

Thanks @furqanmlk! Should've have waited 3 more minutes :)

PR #16376, opened 2020-11-24
MM-30861: Fix racy test TestSaml2Login

agnivade:
https://mattermost.atlassian.net/browse/MM-30861

```release-note
NONE
```

### DIFF
--- a/web/saml.go
+++ b/web/saml.go
@@ -118,14 +118,12 @@ func completeSaml(c *Context, w http.ResponseWriter, r *http.Request) {
 	switch action {
 	case model.OAUTH_ACTION_SIGNUP:
 		teamId := relayProps["team_id"]
-		if len(teamId) > 0 {
-			c.App.Srv().Go(func() {
-				if err = c.App.AddUserToTeamByTeamId(teamId, user); err != nil {
-					mlog.Error(err.Error())
-				} else {
-					c.App.AddDirectChannels(teamId, user)
-				}
-			})
+		if teamId != "" {
+			if err = c.App.AddUserToTeamByTeamId(teamId, user); err != nil {
+				mlog.Error(err.Error())
+			} else {
+				c.App.AddDirectChannels(teamId, user)
+			}
 		}
 	case model.OAUTH_ACTION_EMAIL_TO_SSO:
 		if err = c.App.RevokeAllSessions(user.Id); err != nil {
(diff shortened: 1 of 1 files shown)


web/saml.go
```diff
@@ -118,14 +118,12 @@ func completeSaml(c *Context, w http.ResponseWriter, r *http.Request) {
 	switch action {
 	case model.OAUTH_ACTION_SIGNUP:
 		teamId := relayProps["team_id"]
-		if len(teamId) > 0 {
-			c.App.Srv().Go(func() {
```
agnivade, line 122: @sbishel / @mkraft - Same question as earlier. I didn't see any particular cause for delays except for a few additional DB calls and websocket messages. Nothing that should cause any noticeable delay.

This is also achievable by deep copying the user and doing it in a goroutine. But I wanted to check if there was anything that I missed.

web/saml.go
```diff
@@ -118,14 +118,12 @@ func completeSaml(c *Context, w http.ResponseWriter, r *http.Request) {
...
-				if err = c.App.AddUserToTeamByTeamId(teamId, user); err != nil {
-					mlog.Error(err.Error())
-				} else {
-					c.App.AddDirectChannels(teamId, user)
-				}
-			})
+		if teamId != "" {
+			if err = c.App.AddUserToTeamByTeamId(teamId, user); err != nil {
+				mlog.Error(err.Error())
```
streamer45, line 123: We could just `break` here.

web/saml.go
```diff
@@ -118,14 +118,12 @@ func completeSaml(c *Context, w http.ResponseWriter, r *http.Request) {
...
 		teamId := relayProps["team_id"]
-		if len(teamId) > 0 {
-			c.App.Srv().Go(func() {
-				if err = c.App.AddUserToTeamByTeamId(teamId, user); err != nil {
-					mlog.Error(err.Error())
-				} else {
-					c.App.AddDirectChannels(teamId, user)
-				}
-			})
+		if teamId != "" {
```
streamer45, line 121: This could be done in the same line/scope.

streamer45 (approved):
Looks good :+1: 
Left a couple of minor syntaxing suggestions.

web/saml.go
sbishel: I'm not as familiar with this section, but tracing it through, I agree. Shouldn't cause noticable delay.

sbishel (approved):

mkraft (approved):

PR #16382, opened 2020-11-24
MM-30863: Fix race in LRU

agnivade:
After shortening the critical section, we missed out the fact
that the byte slice is still accessible after the element is returned.
So the lock needs to be active until the byte slice is fully read and
unmarshaled

https://mattermost.atlassian.net/browse/MM-30863

```release-note
NONE
```

### DIFF
--- a/services/cache/lru.go
+++ b/services/cache/lru.go
@@ -185,23 +185,14 @@ func (l *LRU) set(key string, value interface{}, ttl time.Duration) error {
 }
 
 func (l *LRU) get(key string, value interface{}) error {
-	l.lock.Lock()
-	defer l.lock.Unlock()
-
-	ent, ok := l.items[key]
-	if !ok {
-		return ErrKeyNotFound
+	val, err := l.getItem(key)
+	if err != nil {
+		return err
 	}
-	e := ent.Value.(*entry)
-	if e.generation != l.currentGeneration || (!e.expires.IsZero() && time.Now().After(e.expires)) {
-		l.removeElement(ent)
-		return ErrKeyNotFound
-	}
-	l.evictList.MoveToFront(ent)
 
 	// We use a fast path for hot structs.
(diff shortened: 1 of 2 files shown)


akshaychhajed:
@agnivade So it will extend to getItem function too ?

agnivade:
Yes, unfortunately there is no way out of it. The byte slice is part of the struct, hence needs to be inside the critical section.

akshaychhajed:
Thanks for the tag @agnivade.
Completely missed that slice.
I think deep copy would have similar performance to current marshalling.

Good catch!

services/cache/lru.go
```diff
@@ -226,23 +235,6 @@ func (l *LRU) get(key string, value interface{}) error {
 	return msgpack.Unmarshal(e.value, value)
 }
 
-func (l *LRU) getItem(key string) (*entry, error) {
```
streamer45, line 229: From what I can see the race is just about accessing the entry struct, not about the byte slice. We can guarantee the data is only written before any read so it shouldn't be a problem. Wouldn't returning `e.value` here instead of the whole struct be sufficient to fix this case?

agnivade: Yep, indeed. My bad.

streamer45 (approved):
Looks good :+1: Completely missed this one on the original PR.

services/cache/lru.go
akshaychhajed: Just out of curiosity, where does data race message tell you if race was for struct and not byte slice?

I am referring to this
https://mattermost.atlassian.net/browse/MM-30863

agnivade: It does not directly tell you. Needs a little bit of fiddling to understand it.

isacikgoz (approved):
Looks good tome 👍

PR #16383, opened 2020-11-24
[MM-30907]: Change receiver name for (me *AwayProvider) in app/slashcommands/command_away.go to be more idiomatic.

haardikdharma10:
#### Summary 
Updated receiver name for (me *AwayProvider) to be more idiomatic.

#### Ticket Link
Fixes: #16366 
JIRA Ticket: https://mattermost.atlassian.net/browse/MM-30907

#### Release Note
```release-note
NONE
```

### DIFF
--- a/app/slashcommands/command_away.go
+++ b/app/slashcommands/command_away.go
@@ -20,11 +20,11 @@ func init() {
 	app.RegisterCommandProvider(&AwayProvider{})
 }
 
-func (aw *AwayProvider) GetTrigger() string {
+func (*AwayProvider) GetTrigger() string {
 	return CMD_AWAY
 }
 
-func (aw *AwayProvider) GetCommand(a *app.App, T goi18n.TranslateFunc) *model.Command {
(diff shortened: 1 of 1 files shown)


agnivade (commented):
Hi @haardikdharma10, thanks for the PR!

I wanted to mention a couple of general comments that are applicable to the entire set of PRs that you sent.
- If we don't use the receiver value inside a method, such as here, we should just omit the receiver variable.
- The name is usually the first letters of each word that is abbreviated together. In this case, if the struct were to be used, it would be `ap`, not `aw`.

Let's make the changes accordingly to this PR and the rest of the PRs. Thanks for your time!

haardikdharma10:
Hi @agnivade, thank you for your time and feedback. I understood the points you mentioned and also understood that using the receiver variable here (in this case) is not necessary and hence omitted it. Can you please give it a look and I'll make changes to the rest of the PR's as per the second point you mentioned. Thank you!

agnivade (approved):
Looks great, thank you!

PR #16393, opened 2020-11-24
Syncronize access when updating config on search layer

nronas:
#### Summary
Fixes a race condition due to a non synchronized access of a critical section, in our case updating the config of the search layer. 

Solution: synchronise access via making config to be an `atomic.Value`

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-30868

#### Release Note

```release-note
Fix race condition when updating config on search layer.
```

### DIFF
--- a/store/searchlayer/layer.go
+++ b/store/searchlayer/layer.go
@@ -4,6 +4,8 @@
 package searchlayer
 
 import (
+	"sync"
+
 	"github.com/mattermost/mattermost-server/v5/mlog"
 	"github.com/mattermost/mattermost-server/v5/model"
 	"github.com/mattermost/mattermost-server/v5/services/searchengine"
@@ -18,6 +20,7 @@ type SearchStore struct {
 	channel      *SearchChannelStore
 	post         *SearchPostStore
 	config       *model.Config
+	configLock   sync.RWMutex
 }
 
 func NewSearchLayer(baseStore store.Store, searchEngine *searchengine.Broker, cfg *model.Config) *SearchStore {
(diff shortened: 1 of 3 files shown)


store/searchlayer/layer.go
```diff
@@ -18,6 +20,7 @@ type SearchStore struct {
 	channel      *SearchChannelStore
 	post         *SearchPostStore
 	config       *model.Config
+	configLock   sync.RWMutex
```
nronas, line 23: Also a `sync.Mutex` here will be just fine.

streamer45: I am wondering if using `atomic.Value` would be more appropriate in this case. We use that consistently in the `app.Server` struct.

store/searchlayer/layer_test.go
```diff
@@ -0,0 +1,41 @@
...
+// See LICENSE.txt for license information.
+
+package searchlayer_test
+
+import (
+	"os"
+	"testing"
+
+	"github.com/mattermost/mattermost-server/v5/store/searchlayer"
+
```
agnivade, line 11: Nit: there is no need for gaps here and on L16

store/searchlayer/layer_test.go
```diff
@@ -0,0 +1,41 @@
...
+	cfg := &model.Config{}
+	cfg.SetDefaults()
+	cfg.ClusterSettings.MaxIdleConns = model.NewInt(1)
+	searchEngine := searchengine.NewBroker(cfg, nil)
+	layer := searchlayer.NewSearchLayer(&testlib.TestStore{Store: store}, searchEngine, cfg)
+
+	for i := 0; i < 5; i++ {
+		go func() {
+			layer.UpdateConfig(cfg.Clone())
+		}()
```
agnivade, line 39: I think it might be better to wait for the goroutines to finish before exiting the test using a waitgroup.

agnivade (commented):
Looks good. Just a few small comments.

store/searchlayer/layer_test.go
nronas: damn this is my editor doing `goimports` on save.

store/searchlayer/layer_test.go
nronas: nice catch, totally missed that

agnivade (approved):
LGTM

streamer45 (approved):

PR #16398, opened 2020-11-24
Make mlogger atomic

nronas:
#### Summary
Makes mlogger atomic on its basic operation. Started investigating the ticket bellow and it seems that mlogger has many critical sections that a race can happen. This PR synchronizes access to `logger.logrLogger`.

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-30869

#### Release Note

```release-note
Make mlogger atomic
```

### DIFF
--- a/mlog/log.go
+++ b/mlog/log.go
@@ -74,7 +74,7 @@ type Logger struct {
 	consoleLevel zap.AtomicLevel
 	fileLevel    zap.AtomicLevel
 	logrLogger   *logr.Logger
-	mutex        *sync.Mutex
+	mutex        *sync.RWMutex
 }
 
 func getZapLevel(level string) zapcore.Level {
@@ -108,7 +108,7 @@ func NewLogger(config *LoggerConfiguration) *Logger {
 		consoleLevel: zap.NewAtomicLevelAt(getZapLevel(config.ConsoleLevel)),
 		fileLevel:    zap.NewAtomicLevelAt(getZapLevel(config.FileLevel)),
 		logrLogger:   newLogr(),
-		mutex:        &sync.Mutex{},
+		mutex:        &sync.RWMutex{},
 	}
 
 	if config.EnableConsole {
@@ -167,8 +167,8 @@ func (l *Logger) SetConsoleLevel(level string) {
 func (l *Logger) With(fields ...Field) *Logger {
 	newLogger := *l
 	newLogger.zap = newLogger.zap.With(fields...)
-	if newLogger.logrLogger != nil {
-		ll := newLogger.logrLogger.WithFields(zapToLogr(fields))
+	if newLogger.getLogger() != nil {
+		ll := newLogger.getLogger().WithFields(zapToLogr(fields))
 		newLogger.logrLogger = &ll
 	}
 	return &newLogger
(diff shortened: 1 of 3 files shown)


wiggin77:
Hi @nronas.  This race is very specific to server shutdown.  Let's focus on trying to solve that case specifically instead of removing concurrency from all logging operations, which will have a performance impact.

nronas:
@wiggin77 sure thing, it's defo a scope creep , my only concern is/was that the above package is racy. For example, the race condition is between the Shutdown and the Debug call, because the shutdown it assigns a new logger. But depending on the log level that can be any of the log level methods. So what you think for a minimum to synchronize all the log level methods and the `ShutdownAdvancedLogging` ?

wiggin77:
> @wiggin77 sure thing, it's defo a scope creep , my only concern is/was that the above package is racy. For example, the race condition is between the Shutdown and the Debug call, because the shutdown it assigns a new logger. But depending on the log level that can be any of the log level methods. So what you think for a minimum to synchronize all the log level methods and the `ShutdownAdvancedLogging` ?

I don't think the package is inherently racy.  That re-assignment of logrLogger is definitely racy.  That was added by me (doh!) just a few months ago.  I'm trying to figure out why a new Logr is assigned anyway, since shutdown is only called when server is shutting down, and one of the last things done before the server exits.  Removing the re-assignment fixes the race.  I'm checking what, if any, side effect that has.

nronas:
> I don't think the package is inherently racy. That re-assignment of logrLogger is definitely racy. That was added by me (doh!) just a few months ago. I'm trying to figure out why a new Logr is assigned anyway, since shutdown is only called when server is shutting down, and one of the last things done before the server exits. Removing the re-assignment fixes the race. I'm checking what, if any, side effect that has.


Superb, that is basically what i was saying but probably poorly :D. Due to the re-assignment any invocation on methods that require the `logrLogger` are open to races.

nronas:
> EDIT: ah, I see shutdown also gets called when re-configuring as a lazy way to start fresh. I suggest creating a getter method for (*Logger).logrLogger and guard that with a RW mutex. That will preserve 99% of the concurrency and fix the race for that re-assignment.


Amended in https://github.com/mattermost/mattermost-server/pull/16398/commits/e37cc41f6b04d6abc89472d8271a813e015a4440

wiggin77 (approved):
LGTM :+1:

mlog/log.go
```diff
@@ -190,9 +193,9 @@ func (l *Logger) StdLogWriter() io.Writer {
 }
 
 func (l *Logger) WithCallerSkip(skip int) *Logger {
-	newlogger := *l
-	newlogger.zap = newlogger.zap.WithOptions(zap.AddCallerSkip(skip))
-	return &newlogger
+	newLogger := *l
+	newLogger.zap = newLogger.zap.WithOptions(zap.AddCallerSkip(skip))
```
streamer45, line 197: Could this trigger similar issues?

mlog/log_test.go
```diff
@@ -0,0 +1,50 @@
...
+			select {
+			case <-ctx.Done():
+				return
+			default:
+				logger.Debug("testing...")
+			}
+		}
+	}()
+
+	require.True(t, <-started, "goroutine failed to start")
```
streamer45, line 43: Is the `require` check really needed? It's hard to see this ever failing :D

nronas: Fair :D.

mlog/log.go
nronas: hmmmm...I think this is fine since it operates on the main logger struct and not on the `logger.Logr` right?

streamer45: Looks fine since it's done on a newly created object and from that point on it should be all reads.
Even `StdLogWriter` does a copy before setting that so we are good.

nronas: Exactly!!!

streamer45 (approved):
Nicely done :+1:

mlog/log.go
```diff
@@ -297,13 +300,27 @@ func (l *Logger) RemoveTargets(ctx context.Context, f func(ti TargetInfo) bool)
...
 // EnableMetrics enables metrics collection by supplying a MetricsCollector.
 // The MetricsCollector provides counters and gauges that are updated by log targets.
 func (l *Logger) EnableMetrics(collector logr.MetricsCollector) error {
-	return l.logrLogger.Logr().SetMetricsCollector(collector)
+	return l.getLogger().Logr().SetMetricsCollector(collector)
+}
+
+// getLogger is a concurrent safe getter of the logr logger
+func (l *Logger) getLogger() *logr.Logger {
+	defer l.mutex.RUnlock()
```
agnivade, line 314: Wondering if we should also use the atomic approach we used in the earlier PR? 0/5 though.

mlog/log.go
```diff
@@ -73,6 +74,7 @@ type Logger struct {
 	consoleLevel zap.AtomicLevel
 	fileLevel    zap.AtomicLevel
 	logrLogger   *logr.Logger
+	mutex        *sync.RWMutex
```
agnivade, line 77: There is no need to use a pointer here. The receiver is already a pointer, so that should be sufficient.

nronas: I used a pointer here because we do a deep copy on the Shutdown which will leak resources right?

mlog/log.go
nronas: i do not see why not, also 0/5 here. I guess mutexes are slower and for a logger an atomic.Value will me more performant. I do not know how well atomic. Value plays with copies though

nronas: Since from the docs it says `Once Store has been called, a Value must not be copied.` ref https://golang.org/pkg/sync/atomic/#Value

mlog/log.go
agnivade: Ah I see. Indeed I missed that. Thanks!

mlog/log.go
agnivade: Yeah, doesn't make sense if there are copies involved. :+1:

agnivade (approved):

PR #16406, opened 2020-11-24
MM-30801: pre-package channel export plugin

lieut-data:
#### Summary
The [channel export plugin](https://github.com/mattermost/mattermost-plugin-channel-export) complements [incident management](https://github.com/mattermost/mattermost-plugin-incident-management).

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-30801

```release-note
Pre-package and pre-install the channel export plugin for enterprise-ready builds.
```


services/telemetry/telemetry.go
```diff
@@ -1283,6 +1283,7 @@ func (ts *TelemetryService) trackPluginConfig(cfg *model.Config, marketplaceURL
 		"com.mattermost.custom-attributes",
 		"com.mattermost.mscalendar",
 		"com.mattermost.nps",
+		"com.mattermost.plugin-channel-export",
```
agarciamontoro, line 1286: Do we need this, even if there is no telemetry in the plugin?

agarciamontoro (approved):
Nice! :tada: I left a non-blocking comment below.

services/telemetry/telemetry.go
lieut-data: I /think/ we whitelist the plugin here so that it gets included in the metrics as being installed.

hanzei: Correct. Please notes that this list is only used for airgapped installations.

hanzei (commented):
Two questions:

1. How does the auto enable part work on a non-licensed server? Is there a change that admins are confused about a functionality that they can't enable?
2. How much value does this plugin add for the average installation?

lieut-data:
Thanks for the query, @hanzei!

1. On an unlicensed server, the plugin enables normally, but explains the license requirement when the specific functionality is requested. Outside of a slash command, there are no user-facing changes, so confusion should be at a minimum.
2. I can't offer a quantified value-add proposition, but this plugin is intended to ship by default alongside Incident Management as part of the "DevOps Command Centre." In a sense, we could just fold this functionality into a single plugin, but have decided to keep the delivery separate for now. @itao, any other thoughts here?

hanzei:
1. Would we enable the plugin only if a E20 license is present?

lieut-data:
1. Unfortunately, I don't believe there's a good technical way to do to this right now: defaulting the configuration occurs well before we load a license, and we only have "one shot" to flip the `nil` state to `true` (allowing the plugin to be disabled without forcing it back on again). For better or worse, we're using a similar mechanism with incident management which fails more "harshly" during `OnActivate`.

itao:
1. 💯  aligned with @lieut-data 

2. I assume we're referring to average E20 installations here since it's only applicable for those customers right now. We've seen enough recurring requests as well as complicated workarounds from different customers to be confident that it's a useful feature. It also completes the incident management narrative and use case, hence why we decided it's worth to pre-install.

itao (approved):

model/config.go
```diff
@@ -2683,6 +2683,11 @@ func (s *PluginSettings) SetDefaults(ls LogSettings) {
 		s.PluginStates["com.mattermost.plugin-incident-management"] = &PluginState{Enable: true}
 	}
 
+	if s.PluginStates["com.mattermost.plugin-channel-export"] == nil {
+		// Enable the channel export plugin by default
+		s.PluginStates["com.mattermost.plugin-channel-export"] = &PluginState{Enable: true}
```
hanzei, line 2688: I'm still wondering if we can mitigate the impact of not working slash command for TE users. Maybe we only enable the plugin for EE installations because TE installations won't have the required license file anyway.

lieut-data: If we do this, it would have to be outside this code path -- possibly at runtime of the plugin itself? But that could get tricky if a license isn't yet installed on a fresh EE instance but is about to be.

I see this as one of those annoying "enterprise nag" sort of things, where there's a button on display but it clearly communicates a need to upgrade first. Maybe running the plugin at all in this context should be on the chopping block, so it behaves like IM?

Deferring to @tao here, but I believe this is working as desired for TE users at present.

hanzei: > If we do this, it would have to be outside this code path

Why can't we check that in this code path? Just to be clear here: I'm suggesting to enable the plugin based on the installation type, not based on a configured license.

lieut-data: Ah, TE vs. EE installations! Yes, that would validly that cut out a set of non-EE users, though it leaves the EE-but-TE users still affected. I think this is a solid improvement for both plugins, IMO.

Makefile
```diff
@@ -96,6 +96,7 @@ PLUGIN_PACKAGES += mattermost-plugin-jira-v2.3.2
 PLUGIN_PACKAGES += mattermost-plugin-gitlab-v1.1.0
 PLUGIN_PACKAGES += mattermost-plugin-jenkins-v1.0.0
 PLUGIN_PACKAGES += mattermost-plugin-incident-management-v1.1.1
+PLUGIN_PACKAGES += mattermost-plugin-channel-export-v0.2.2
```
hanzei, line 99: Should we use `v1.0.0`?

hanzei (approved):
Awesome :tada: 

One non blocking question

Makefile
lieut-data: Yes, I'll update as soon as we can build a new version.

amyblais:
@lieut-data Is this critical for v5.30 or can be pushed to the next release?

lieut-data:
@amyblais, if still possible, we'd like to include this pre-packaged plugin to complement Incident Management already being pre-packaged. 

@prapti, can you help verify the new behaviour here?
* Channel Export ready and available alongside Incident Management
* Incident Management and Channel Export only enable by default on EE installations (whether or not a license is present)

Ideally, we'll bump the included version to v1.0, but it's not strictly critical and we can pre-package as-is while we work towards listing the newer version in the marketplace. (Pipeline issues with the bumped version.)

prapti (approved):
Tested and verified on PR, cloud and local servers that both Channel Export and Incident Management plugins are installed and enabled in EE servers with or without license; and the plugins are not installed in TE.
LGTM!

lieut-data:
Thank you SO MUCH, @prapti!

prapti:
No problem, Jesse! Thanks for the ping!

lieut-data:
Syncing with @tao, and we're going to move ahead with this version as-is while we sort out the plugin build and signing issues. The `v1.0.0` moniker is just about optics at this point.

PR #16430, opened 2020-12-01
MM-30987- Fix race in setting clearNotify in Busy

agnivade:
A time.AfterFunc runs the function in its own goroutine. So we need
to guard that with a mutex too.

https://mattermost.atlassian.net/browse/MM-30987

```release-note
NONE
```


streamer45 (approved):
Nice :+1:

wiggin77 (approved):
Doh!  I should have caught that.  :+1:

PR #16431, opened 2020-12-01
MM-30988 - Fix racy test ServerSystemdNotification

agnivade:
The translateFunc is a global variable which was unguarded. This wasn't a problem from the code,
but when consecutive tests were run, one test would still be writing to it while the other would be reading.
We wrap the write with a sync.Once so that consecutive tests don't write to it again.

https://mattermost.atlassian.net/browse/MM-30988

```release-note
NONE
```

### DIFF
--- a/model/utils.go
+++ b/model/utils.go
@@ -73,9 +73,12 @@ func (sa StringArray) Equals(input StringArray) bool {
 }
 
 var translateFunc goi18n.TranslateFunc
+var translateFuncOnce sync.Once
 
 func AppErrorInit(t goi18n.TranslateFunc) {
-	translateFunc = t
+	translateFuncOnce.Do(func() {
+		translateFunc = t
+	})
 }
 
 type AppError struct {
--- a/model/utils.go
+++ b/model/utils.go
@@ -18,6 +18,7 @@ import (
 	"regexp"
 	"strconv"
 	"strings"
+	"sync"
 	"time"
 	"unicode"
 
--- a/model/utils.go
+++ b/model/utils.go
@@ -18,6 +18,7 @@ import (
 	"regexp"
 	"strconv"
 	"strings"
+	"sync/atomic"
 	"time"
 	"unicode"
 
@@ -72,10 +73,10 @@ func (sa StringArray) Equals(input StringArray) bool {
 	return true
 }
 
-var translateFunc goi18n.TranslateFunc = nil
+var translateFunc atomic.Value
 
 func AppErrorInit(t goi18n.TranslateFunc) {
-	translateFunc = t
+	translateFunc.Store(t)
 }
 
 type AppError struct {
@@ -148,7 +149,7 @@ func NewAppError(where string, id string, params map[string]interface{}, details
 	ap.DetailedError = details
 	ap.StatusCode = status
 	ap.IsOAuth = false
-	ap.Translate(translateFunc)
+	ap.Translate(translateFunc.Load().(goi18n.TranslateFunc))
 	return ap
 }
 
--- a/model/utils.go
+++ b/model/utils.go
@@ -18,7 +18,6 @@ import (
 	"regexp"
 	"strconv"
 	"strings"
-	"sync/atomic"
 	"time"
 	"unicode"
 
@@ -73,10 +72,10 @@ func (sa StringArray) Equals(input StringArray) bool {
 	return true
 }
 
-var translateFunc atomic.Value
+var translateFunc goi18n.TranslateFunc
 
 func AppErrorInit(t goi18n.TranslateFunc) {
-	translateFunc.Store(t)
+	translateFunc = t
 }
 
 type AppError struct {
@@ -149,11 +148,7 @@ func NewAppError(where string, id string, params map[string]interface{}, details
 	ap.DetailedError = details
 	ap.StatusCode = status
 	ap.IsOAuth = false
-	if fn, ok := translateFunc.Load().(goi18n.TranslateFunc); ok {
-		ap.Translate(fn)
-	} else {
-		ap.Translate(nil)
-	}
+	ap.Translate(translateFunc)
 	return ap
 }
 
--- a/model/utils.go
+++ b/model/utils.go
@@ -149,7 +149,11 @@ func NewAppError(where string, id string, params map[string]interface{}, details
 	ap.DetailedError = details
 	ap.StatusCode = status
 	ap.IsOAuth = false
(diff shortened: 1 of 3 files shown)


nronas (approved):
LGTM 👌

model/utils.go
```diff
@@ -72,10 +73,10 @@ func (sa StringArray) Equals(input StringArray) bool {
 	return true
 }
 
-var translateFunc goi18n.TranslateFunc = nil
+var translateFunc atomic.Value
```
streamer45, line 76: Do we really need this to be atomic? This looks like something that should be set before any `AppError` can be created and never touched again.
Wouldn't it be enough to move the call to `AppErrorInit()` further up in the server initialization function (before initializing things like jobs)?

agnivade: I agree. What made me think this needs to be atomic is that for this race itself, the setting of the value happens before the read code is executed. So I thought maybe there's something else going on. I couldn't actually repro the race locally. But I'll move it upwards to make it more robust.

streamer45: Yeah, looking at the race logs, it seems like `runServer` gets called at the same time from the two tests (`TestRunServerSuccess` and `TestRunServerSystemdNotification`). I don't think those would run in parallel, so maybe we are not properly waiting for every goroutine to be done on server shutdown? In any case, this race seems test specific as in production I'd assume we'd call that only once.

scripts/test.sh
```diff
@@ -16,7 +16,7 @@ echo "Packages to test: $PACKAGES"
 find . -name 'cprofile*.out' -exec sh -c 'rm "{}"' \;
 find . -type d -name data -not -path './vendor/*' -not -path './data' | xargs rm -rf
 
-$GO test $GOFLAGS -run=$TESTS $TESTFLAGS -v -timeout=2000s -covermode=count -coverpkg=$PACKAGES_COMMA -exec $DIR/test-xprog.sh $PACKAGES 2>&1 > >( tee output )
+$GO test $GOFLAGS -run=$TESTS $TESTFLAGS -v -race -timeout=5000s -covermode=atomic -coverpkg=$PACKAGES_COMMA -exec $DIR/test-xprog.sh $PACKAGES 2>&1 > >( tee output )
```
nronas, line 19: Just for my context, what's the reason for updating the timeout?

streamer45: Running with `-race` takes much more time so it makes sense to bump that a bit.

model/utils.go
agnivade: Ok, I think I know what's happening. I moved the call to the top, but I still see this race happening in some other tests. I made a slight mistake by putting it too high. But I think I get the general problem here.

Since this is a global variable, so the value persists between multiple tests. So as a result, even if there is a slight mistiming where one test does not finish completely, and another test starts, this race will manifest. It's not just `TestRunServerSuccess` and `TestRunServerSystemdNotification`. If you see https://app.circleci.com/pipelines/github/mattermost/mattermost-server/18451/workflows/f3065425-eba6-412e-8caf-10c897633637/jobs/105295 and https://app.circleci.com/pipelines/github/mattermost/mattermost-server/18451/workflows/f3065425-eba6-412e-8caf-10c897633637/jobs/105294, there are multiple failures of the same race.

All of it is because of consecutive tests running in the same process. The actual code is fine. I am thinking of covering the set call with a `sync.Once`, so that it's only set once during the lifetime of a process. That does not go the full way of guarding a variable, but should still fix the race. Thoughts?

The issue is more complicated by this `utils.TranslationsPreInit` which sets yet another global variable, and then we set that global variable to another variable in model package. Talk about complexity :)
```
if err := utils.TranslationsPreInit(); err != nil {
		return nil, errors.Wrapf(err, "unable to load Mattermost translation files")
	}
	model.AppErrorInit(utils.T)
```

nronas: hmm interesting findings. I would agree that `sync.Once` will fix the race, we must ensure that we are not copying that global var anywhere though.

streamer45: > Since this is a global variable, so the value persists between multiple tests. So as a result, even if there is a slight mistiming where one test does not finish completely, and another test starts, this race will manifest. It's not just `TestRunServerSuccess` and `TestRunServerSystemdNotification`. If you see https://app.circleci.com/pipelines/github/mattermost/mattermost-server/18451/workflows/f3065425-eba6-412e-8caf-10c897633637/jobs/105295 and https://app.circleci.com/pipelines/github/mattermost/mattermost-server/18451/workflows/f3065425-eba6-412e-8caf-10c897633637/jobs/105294, there are multiple failures of the same race.

yes, somehow we are not guaranteeing order between accesses.

> All of it is because of consecutive tests running in the same process. The actual code is fine. I am thinking of covering the set call with a `sync.Once`, so that it's only set once during the lifetime of a process. That does not go the full way of guarding a variable, but should still fix the race. Thoughts?

Sounds good to me :+1:

agnivade: Ok, pushed that. Will let it run through `-race` flag to confirm it fixes the race.

agnivade: Worked.

scripts/test.sh
```diff
@@ -16,7 +16,7 @@ echo "Packages to test: $PACKAGES"
 find . -name 'cprofile*.out' -exec sh -c 'rm "{}"' \;
 find . -type d -name data -not -path './vendor/*' -not -path './data' | xargs rm -rf
 
-$GO test $GOFLAGS -run=$TESTS $TESTFLAGS -v -timeout=2000s -covermode=count -coverpkg=$PACKAGES_COMMA -exec $DIR/test-xprog.sh $PACKAGES 2>&1 > >( tee output )
+$GO test $GOFLAGS -run=$TESTS $TESTFLAGS -v -timeout=1000s -covermode=count -coverpkg=$PACKAGES_COMMA -exec $DIR/test-xprog.sh $PACKAGES 2>&1 > >( tee output )
```
streamer45, line 19: - Is 1000s enough for CI tests these days?
- Should we switch to a more human friendly duration format? e.g. 20m

agnivade: Oh my bad, it should be 2000s. I'll change it to 20m

streamer45: Cool. Maybe 20m wasn't the best example as it's closer to 1000s than 2000s but quickly looking at CI times it seems that the average is around 15-16 minutes so it should be okay. I guess we could always could bump that to half an hour to be extra-safe.

streamer45 (approved):

PR #16447, opened 2020-12-02
After OAUTH, SAML auth completion, Redirect to App custom url scheme with token data as query params.

anurag6713:
#### Summary
MM-30480:
For mobile clients, After Auth process is completed, It redirects to the specified url with cookie data as query params.

This is used by mobile client for new OAuth implementation as we cannot read cookies from the mobile browser, we will pass them using query params 

MM-29644:
- Web: Ignores redirect URL passed from client
- Mobile: Validates the Scheme from redirect URL and makes sure it doesn't include default schemes (http, https etc)

Added screens to show auth success or on error:
Simulator Screen Shot - iPhone 11 - 2020-12-14 at 14 46 57
Simulator Screen Shot - iPhone 11 - 2020-12-14 at 15 06 40

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-30480
https://mattermost.atlassian.net/browse/MM-29644

#### Release Note
1. Added ability to redirect to the mobile app after OAuth & SAML auth completion.
2. Validating RedirectURL for web and mobile OAuth flows.


utils/utils.go
```diff
@@ -172,3 +173,13 @@ func GetUrlWithCache(url string, cache *RequestCache, skip bool) ([]byte, error)
 	cache.Date = resp.Header.Get("Date")
 	return cache.Data, err
 }
+
+func BuildUrlQueryStringFromCookies(baseUrl string, r *http.Request) string {
+	u, _ := url.Parse(baseUrl)
+	q, _ := url.ParseQuery(u.RawQuery)
```
mkraft, line 179: Could we please handle these errors? I know it's done in some existing places.

mkraft (changes requested):

sbishel (changes requested):

web/oauth.go
```diff
@@ -367,13 +370,15 @@ func mobileLoginWithOAuth(c *Context, w http.ResponseWriter, r *http.Request) {
 		return
 	}
 
+	redirectTo := r.URL.Query().Get("redirect_to")
```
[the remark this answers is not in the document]
anurag6713: As we have discussed about validating it even for web, I thought I would do it in a different PR. Should I include mobile validation here instead?

utils/utils.go
```diff
@@ -172,3 +173,32 @@ func GetUrlWithCache(url string, cache *RequestCache, skip bool) ([]byte, error)
...
+	}
+	u.RawQuery = q.Encode()
+	return u.String()
+}
+
+var DefaultUrlSchemes = []string{"http", "https", "ftp", "mailto", "tel"}
+
+// Validate .. if passed redirectUrl is a valid mobile app custom scheme url
+// Scheme should not match any default schemes
+func IsValidCustomSchemeUrl(customSchemeUrl string) bool {
```
mkraft, line 198: Shouldn't this just be looking for the prefix `mattermost://`? Won't, say, `file://foo` or `facetime://bar` incorrectly appear valid?

web/oauth.go
```diff
@@ -273,34 +273,45 @@ func completeOAuth(c *Context, w http.ResponseWriter, r *http.Request) {
...
 
 	action := ""
+	hasRedirectUrl := false
+	isActionMobileAuth := false
+	redirectUrl := ""
 	if props != nil {
 		action = props["action"]
+		isActionMobileAuth = action == model.OAUTH_ACTION_MOBILE
+		if _, ok := props["redirect_to"]; ok {
+			redirectUrl = props["redirect_to"]
```
mkraft, line 283: You can use the first parameter of the lookup instead of using the blank identifier (`_`) and then you don't have to look it up twice.

utils/utils.go
anurag6713: We have ["Build your own app"](https://developers.mattermost.com/contribute/mobile/build-your-own/) and they would have different scheme url. That's why I thought of making it a part of config on the mobile client.

If needed we can also think about making it a config like "Site url" ??

anurag6713: Added it in the config properties.

sbishel (changes requested):
Regarding the function `CompleteOAuth`, there appear to be 3 states -
- Old mobile (isActionMobileAuth == true && hasRedirectURL == false)
- New mobile (isActionMobileAuth == true && hasRedirectURL == true)
- Web App (isActionMobileAuth == true)
I wonder if we could restructure the code around these states and simplify/remove some of the if/then logic and make it more explicit what the settings mean.

Unit tests should be added to make sure that code is performing as expected. I realize unit tests haven't been added for this function, but this code is screaming for some.

web/oauth.go
```diff
@@ -274,38 +274,49 @@ func completeOAuth(c *Context, w http.ResponseWriter, r *http.Request) {
...
 
-	var redirectUrl string
 	if action == model.OAUTH_ACTION_EMAIL_TO_SSO {
-		redirectUrl = c.GetSiteURLHeader() + "/login?extra=signin_change"
+		redirectURL = c.GetSiteURLHeader() + "/login?extra=signin_change"
 	} else if action == model.OAUTH_ACTION_SSO_TO_EMAIL {
-		redirectUrl = app.GetProtocol(r) + "://" + r.Host + "/claim?email=" + url.QueryEscape(props["email"])
+		redirectURL = app.GetProtocol(r) + "://" + r.Host + "/claim?email=" + url.QueryEscape(props["email"])
 	} else {
 		isMobile, parseErr := strconv.ParseBool(props[model.USER_AUTH_SERVICE_IS_MOBILE])
```
sbishel, line 321: Are `isMobile` and `isActionMobileAuth` always the same, should we be using one or the other?

utils/utils.go
```diff
@@ -172,3 +175,55 @@ func GetUrlWithCache(url string, cache *RequestCache, skip bool) ([]byte, error)
 	cache.Date = resp.Header.Get("Date")
 	return cache.Data, err
 }
+
+// Append cookies to passed baseUrl into query params
+func BuildUrlQueryStringFromCookies(baseUrl string, r *http.Request) string {
```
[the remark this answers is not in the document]
anurag6713: I forgot to remove it after adding a different method.

web/oauth.go
anurag6713: Yes, they both are the same. I have removed it from this function.

sbishel (approved):

mkraft (approved):

sbishel (approved):

furqanmlk:
Looks good to me,
Verified 
**5.30.0  - OAUTH Gitlab, Google App, Office 365 ✅ 
5.32.0  - OAUTH Gitlab, Google App, Office 365 ✅ 
5.25.0 - OAUTH Gitlab, Google App, Office 365** ✅ 

https://mattermost.atlassian.net/browse/MM-29644 
image

 Login Complete
image

furqanmlk (approved):

amyblais:
@anurag6713 Should the new setting be added to Telemetry https://github.com/mattermost/mattermost-server/blob/master/services/telemetry/telemetry.go?

anurag6713:
> @anurag6713 Should the new setting be added to Telemetry https://github.com/mattermost/mattermost-server/blob/master/services/telemetry/telemetry.go?

@amyblais Yes we need to add this setting to telemetry. As I have merged this PR, will add it in a new PR. Thanks!

PR #16456, opened 2020-12-03
MM-30810 Remove special props from /me post

hmhealey:
Originally, /me posts were just styled by adding asterisks to force them to become italic, but we wanted to change how they were formatted, so we stored the non-formatted message in props to allow the client to do its formatting. We still needed the `Post.Message` to be wrapped in asterisks for compatibility with mobile and other clients.

That ended up causing problems because editing the post would only change the `Post.Message` without changing the copy in the props, so instead of trying to manage two copies of the post, we're just going to have the web app remove the asterisks from the message.

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-30810

#### Related Pull Requests
https://github.com/mattermost/mattermost-webapp/pull/7163

#### Release Note
```release-note
NONE
```


nickmisasi (approved):

hmhealey:
@jfrerich Can you review this when you have a chance? If you're busy, I can re-assign it

jfrerich (approved):
LGTM. sorry for getting to this a bit late!

jgilliam17 (approved):
Thank you @hmhealey  - apologize for the delay 
Tested, looks good to merge.
- Verified edited /me posts are behaving the same in webapp and mobile

PR #16503, opened 2020-12-06
MM-29988 - Update exising Post APIs with collapsed threads

reflog:
#### Summary

Enrich existing post APIs to return thread info (if exists).
collapsedThreads query parameter will cause only threads to be returned
collapsedThreadsExtended query parameter will cause the 'participants' part of the reply to include full user info

#### Ticket Link

https://mattermost.atlassian.net/browse/MM-29988
#### Release Note

```release-note
added collapsedThreads and collapsedThreadsExtended query parameters to:
'api/v4/channels/{channel_id:[A-Za-z0-9]+}/posts'
'api/v4/users/{user_id:[A-Za-z0-9]+}/channels/{channel_id:[A-Za-z0-9]+}/posts/unread'
'api/v4/posts/{post_id:[A-Za-z0-9]+}/thread'
```


store/sqlstore/post_store.go
```diff
@@ -484,9 +530,14 @@ type etagPosts struct {
 func (s *SqlPostStore) InvalidateLastPostTimeCache(channelId string) {
 }
 
-func (s *SqlPostStore) GetEtag(channelId string, allowFromCache bool) string {
+func (s *SqlPostStore) GetEtag(channelId string, allowFromCache, collapsedThreads bool) string {
+	q := s.getQueryBuilder().Select("Id", "UpdateAt").From("Posts").Where(sq.Eq{"ChannelId": channelId})
```
jespino, line 534: You haven't removed the `ORDER BY` here, what is key, because you are looking if there is any new message.

ashishbhate (approved):

jespino:
I've been trying to use this API and I'm having SQL errors like `pq: column reference \"channelid\" is ambiguous"` (related that you are using `channelId` without any prefix like `posts.channelId` in the ` getPostsCollapsedThreads` method.

And also this other one `Scan error on column index 18, name \"threadreplycount\": converting NULL to int64 is unsupported`

streamer45:
@reflog 

`threadmemberships.unreadmentions` does not exist in my local DB and it's causing errors.
Should we add a `CreateColumnIfNotExists()` during store upgrade?

reflog:
> 
> 
> @reflog
> 
> `threadmemberships.unreadmentions` does not exist in my local DB and it's causing errors.
> Should we add a `CreateColumnIfNotExists()` during store upgrade?

Great catch @streamer45 , will add

store/sqlstore/post_store.go
```diff
@@ -616,10 +659,105 @@ func (s *SqlPostStore) PermanentDeleteByChannel(channelId string) error {
...
+	} else {
+		l := len(posts)
+		for i := range posts {
+			if err := processPost(posts[l-i-1]); err != nil {
+				return nil, err
+			}
+			list.AddPost(&posts[l-i-1].Post)
+			list.AddOrder(posts[l-i-1].Id)
+		}
+	}
```
streamer45, line 721: Maybe we could simplify a bit here by having one loop only and setting the index accordingly.

store/sqlstore/post_store.go
```diff
@@ -616,10 +659,105 @@ func (s *SqlPostStore) PermanentDeleteByChannel(channelId string) error {
...
+	var userIds []string
+	userIdMap := map[string]bool{}
+	for _, thread := range posts {
+		for _, participantId := range thread.ThreadParticipants {
+			if _, ok := userIdMap[participantId]; !ok {
+				userIdMap[participantId] = true
+				userIds = append(userIds, participantId)
+			}
+		}
+	}
```
streamer45, line 673: This feels like it could be a standalone helper function in `model`.

streamer45 (commented):
Ran an initial load-test which (as expected) did not reveal any issue. 
We should be running a second one as soon as we've added load-test coverage for the updated API calls.

streamer45 (approved):
Ran a second comparative load-test with the added coverage (thank you @reflog for the help there!) plus a good dataset (3M posts) and didn't notice any significant worsening. So I think we are good to go from a performance perspective :+1: 

image

image

image

store/sqlstore/post_store.go
```diff
@@ -484,9 +522,14 @@ type etagPosts struct {
 func (s *SqlPostStore) InvalidateLastPostTimeCache(channelId string) {
 }
 
-func (s *SqlPostStore) GetEtag(channelId string, allowFromCache bool) string {
+func (s *SqlPostStore) GetEtag(channelId string, allowFromCache, collapsedThreads bool) string {
+	q := s.getQueryBuilder().Select("Id", "UpdateAt").From("Posts").Where(sq.Eq{"ChannelId": channelId}).OrderBy("UpdateAt DESC").Limit(1)
+	if collapsedThreads {
```
jespino, line 527: I don't fully understand why this is changing the Etag, this basically convert any  not-root-post in a not-etag-able data. The idea of the etag, is to not change if nothing has changed, and change if something has change. In this case, we need to find something that only changes is the expected result is changed. Maybe we can use something like the Max value between Post.UdpateAt and Thread.LastReplyAt, or something like that. Maybe that is not needed because it is supposed to not used the collapsed thread for not-root posts. I think this is a bet confusing, can work, but maybe we can add a comment with the intention.

jespino (approved):
A comment for improvement, but otherwise looks good to me.

store/sqlstore/post_store.go
reflog: @jespino - My thinking was: ETag is used to cache results of a channel. We calculate etag based on UpdateAt. If we are in collapsed mode, the channel shows only root posts. Thus I added the query clause. If we keep taking UpdateAt from regular posts I think the caching will be less efficient. Thoughts?

jespino: Yes, I think you are right, I think I miss understood the method itself, I though was about 1 post, and now I see is about the channel itself.

ogi-m (approved):
Smoke tested around posts and threads, no issues. LGTM 👍

store/sqlstore/post_store.go
```diff
@@ -616,10 +659,102 @@ func (s *SqlPostStore) PermanentDeleteByChannel(channelId string) error {
...
+	postFetchQuery, args, _ := s.getQueryBuilder().
+		Select(columns...).
+		From("Posts").
+		LeftJoin("Threads ON Threads.PostId = Id").
+		Where(sq.Eq{"DeleteAt": 0}).
+		Where(sq.Eq{"Posts.ChannelId": options.ChannelId}).
+		Where(sq.Eq{"RootId": ""}).
+		Limit(uint64(options.PerPage)).
+		Offset(uint64(offset)).
+		OrderBy("CreateAt DESC").ToSql()
```
nronas, line 740: If my reading of this is correct, this essentially computes the list of posts needed to fulfill a page on a paginated list. Since we are ordering by createat we are making the page computation non-deterministic in the scenario that we have multiple posts with the same exact createat i.e:

### Total dataset
```
posts
| id |     createat |
---------------------
| 1 | 1609859599058 |
| 2 | 1609859599059 |
| 3| 1609859599059 |
| 4 | 1609859599060 |
| 5 | 1609859599061 |
```

let's assume a page size of 2

### computing first page:
First ordering can return posts[1, 2] but it can also return posts[1,3]

Thoughts? Apologies if i am missing any context here.

reflog: @nronas - first of all, thanks for taking a look, always good to have another pair of eyes on the code :)
second - yes, I think you are correct, however, @calebroseland has already brought this up and we have a pending ticket https://mattermost.atlassian.net/browse/MM-31711 to move to cursor-based paging to avoid this issue.
If you don't think that that would be a sufficient fix - please let me know

nronas: awesome, wasn't aware of the ticket. Cursor based pagination should do the trick 👍

model/client4.go
```diff
@@ -2903,8 +2903,12 @@ func (c *Client4) DeletePost(postId string) (bool, *Response) {
 }
 
 // GetPostThread gets a post with all the other posts in the same thread.
-func (c *Client4) GetPostThread(postId string, etag string) (*PostList, *Response) {
-	r, err := c.DoApiGet(c.GetPostRoute(postId)+"/thread", etag)
+func (c *Client4) GetPostThread(postId string, etag string, collapsedThreads bool) (*PostList, *Response) {
```
streamer45, line 2906: I understand it's a bit late but I'd like to point out that all these are effectively **breaking** changes to our Golang client API. 
I am curious to know if this was discussed before and accepted.

reflog: @streamer45 - how do you figure we 'break' things? The additional parameter to GetPostThread function?

streamer45: Yes, any method/function signature change (as in changing the number of parameters or their type) would be breaking in Golang. This means that any client built on top of this will that makes use of those functions will fail to compile/execute as soon as its dependencies get updated. Usually breaking the API would require a major version bump (6.0 in our case). This kind of accidental breakage has happened before so it's not the end of the world but we should probably try to avoid this.

reflog: @streamer45 - you are right. Thankfully this method is rather obscure and hopefully nobody outside of Mattermost is actually using it

streamer45: Well, this applies to all the affected methods so I wouldn't be surprised if someone is using one of them since they fetch posts (especially `GetPostsSince`). Maybe we could add a warning in the release notes. I think it would be better if it was a conscious decision rather than an overlook.

PR #16508, opened 2020-12-07
[MM-30984] Change default footer for copyright in 2 emails

nickmisasi:
#### Summary
These emails were using the default footer which states `To change your notification preferences, log in to your team site and go to Account Settings > Notifications.` however you won't actually be able to silence these emails. This will put the copyright there. 

#### Release Note
```release-note
None
```


app/email.go
```diff
@@ -781,6 +781,8 @@ func (es *EmailService) SendPaymentFailedEmail(email string, locale string, fail
 	bodyPage.Props["Button"] = T("api.templates.over_limit_fix_now")
 	bodyPage.Props["EmailUs"] = T("api.templates.email_us_anytime_at")
 
+	bodyPage.Props["Footer"] = T("api.templates.copyright")
```
marianunez, line 784: Do we need to make thing change to the overage 7/14/30 days sequence emails as well?

nickmisasi: This is already in those emails. I just forgot to add it to these 2

marianunez (approved):

ethervoid (approved):

PR #16512, opened 2020-12-08
Limit configuration database connections

gabrieljackson:
When configuration is backed by a database, a separate pool of
connections is used for access from the primary database. This
change sets sensible config DB connection limits in order to
reduce database resource usage at scale.

Fixes https://mattermost.atlassian.net/browse/MM-31158

```release-note
NONE
```


nronas (approved):
LGTM

agnivade (approved):
Great stuff. Let's push it!

gabrieljackson:
Noting that QA testing of this change is complicated with very little behavior that could be checked. The test cloud account has been using this change for some days now with no issue.

PR #16527, opened 2020-12-09
Clean up message export redundant flows

nronas:
#### Summary
Counterpart of https://github.com/mattermost/enterprise/pull/876

#### Release Note
```release-note
Refactor message-export feature, improve accuracy, maintainability, and  performance
```


agnivade (approved):
Amazing

mkraft (approved):

store/sqlstore/compliance_store.go
```diff
@@ -243,10 +243,12 @@ func (s SqlComplianceStore) MessageExport(after int64, limit int) ([]*model.Mess
 		LEFT OUTER JOIN Users ON Posts.UserId = Users.Id
 		LEFT JOIN Bots ON Bots.UserId = Posts.UserId
 		WHERE
-			Posts.UpdateAt > :StartTime AND
+			Posts.UpdateAt >= :StartTime AND
+			Posts.UpdateAt <= :EndTime AND
 			Posts.Type NOT LIKE 'system_%'
-		ORDER BY PostUpdateAt
-		LIMIT :Limit`
+		ORDER BY PostUpdateAt, PostId
```
nronas, line 249: Since we are using a compound two-term ordering here, which has some implications to the query plan, it actually makes the query twice as slow, since we only have indexes on PostId and PostUpdateAt independently.

@agnivade @mkraft Before we merge this we will have to create a compound index on  `posts`. Will try to gather some telemetry around this on community in order to have some estimates around the time penalty while creating the index.

agnivade: This is a bit tricky. While maintaining this query's performance is important, I'd be _very_ careful against introducing any indexes to the Posts table. Table sizes for some customers range from 22M-150M posts. This means we'll have to ask all customers to apply the index prior to upgrading.

Unless it shows considerable improvement across the board in our load-tests, adding an index to improve just a single query has considerable tradeoffs. I am wondering if there are ways to work around this. Or perhaps, salvage the whole runtime by optimizing other parts of the workflow as we plan to.

Message export is still an asynchronous job which is expected to take some time. But adding an index will affect all writes to this table. And bigger the table, bigger the index size.

mkraft: I totally agree with what you're both saying, namely @nronas that the query is too slow to release as-is and @agnivade that we need to proceed with caution and be aware of the slowdown to the Posts table and the startup time upon upgrade. 

Great calls on identifying both of those things.

I wonder if we can mitigate the slowdown to Posts table writes by—at the same time as we add the new proposed multicolumn index on (UpdateAt, Id)—also doing the following: 

1. Remove the existing index on UpdateAt (namely "idx_posts_update_at") because it would be functionally handled by the new proposed index.
2. Remove the redundant index on ChannelId (namely "idx_posts_channel_id") which is already covered by two other multicolumn indexes, namely "idx_posts_channel_id_delete_at_create_at" and "idx_posts_channel_id_update_at". 

Sure it's kind of cheating to count #2 as a performance improvement related to this change, but on the other hand it can be legitimately counted towards "not making things slower". Cheating's fine when it comes to performance :)

@agnivade I like your idea of giving customers the option to pre-add the index prior to doing the server startup. Say it takes 5 minutes to add the new index, that's a small price to pay for export correctness without slowing it down. Granted, there are large instance that don't use compliance export. For them they'll be waiting the 5 minutes without seeing any benefits. But any instance with "5 minutes worth" of Post records is a large enterprise that isn't doing the upgrade willy-nilly.

I'm trying to think creatively of other solutions but nothing is coming to me. I don't think the query can be re-written to make it faster without needing the proposed index. The only other thing I can think of is going back to not using the `OFFSET` approach but as you pointed out before @nronas that means the possibility of duplicate posts exported when UpdateAt values are duplicated.

nronas: Cannot agree more with both of you. If we had the ability to run migrations without slowing down/blocking the application boot up that would be ideal. The fact that in the server team this quarter we are going to work on a db migration framework, that can potentially give us the ability to run an index creation while the app is running(meaning outside of the app's. lifecycle).

agnivade: Hmm .. after reading https://ikonicscale.com/your-legacy-database-is-outgrowing-itself, I thought an option might be to do some testing around what happens if we move the sorting part outside the query to the code itself. But even then, there is a correctness problem here because of the LIMIT and OFFSET.

This is really hard to solve without either dropping performance or modifying indexes. Which is very sad, because I was really looking forward to get this merged :(

nronas: The big issue that we have regardless of this work, is the ability to run migrations while the app is running. This work would have been so much easier to be applied if we had the ability to run the server and then issue a concurrent index creation. I know that the server team will work on standalone migrations, so that will unblock this work.

agnivade: Feel free to take up: https://mattermost.atlassian.net/browse/MM-30186 :)

mkraft:
Sorry, I just wanted to confirm this wasn't waiting for me. What was our next step with this one?

agnivade:
I believe we are waiting until we have asynchronous migrations ready.

nronas:
@mkraft what @agnivade said, basically i am going to push out of the door the async migrations PR and resurrect this, right after.

angeloskyratzakos:
Hey all, I will remove the tag `Setup Cloud Test Server` for maintenance reasons. Feel free to re-add it when you need it. Thank you

jasonblais:
@agnivade not sure if there are next steps for this PR?

agnivade:
The code for this PR is already merged via another PR. Closing.

PR #16528, opened 2020-12-09
MM-31171 Fork OpenGraph library to handle bad HTML from YouTube

hmhealey:
YouTube previews are currently broken sporadically on different servers because they started sending HTML with the OpenGraph metadata in the wrong place as well as some other invalid bits. It seems like it's a rolling change being propagated through all their servers since previews went from working on community to working sometimes to not working at all. Their HTML is actually invalid now, so I assume they'll fix it, but I've forked the OpenGraph library to be more relaxed.

The changes to the fork are here: https://github.com/dyatlov/go-opengraph/compare/master...hmhealey:master
In particular, it's these lines: https://github.com/dyatlov/go-opengraph/compare/master...hmhealey:master#diff-47ab48a2396307d97df39eabbd77905aad8cb1b808acfcab8a7a3b5523f08a6fL122-L124

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-31171

#### Release Note
```release-note
Fixed YouTube previews not being displayed sometimes
```

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -139,3 +139,5 @@ require (
 )
 
 replace github.com/NYTimes/gziphandler v1.1.1 => github.com/agnivade/gziphandler v1.1.2-0.20200815170021-7481835cb745
+
+replace github.com/dyatlov/go-opengraph => github.com/hmhealey/go-opengraph v0.0.0-20201209151214-f2d823730dba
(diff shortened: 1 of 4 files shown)


crspeller (commented):
Should we move this fork into the org?

hmhealey:
We could consider that, but I don't think it's entirely necessary since this should only be a temporary fork. YouTube should hopefully fix their HTML since it really looks like a bug in their markup generation, but I've also reached out to the maintainer of the upstream library to see if they would be interested in having this pushed upstream since it causes a slight performance degradation by parsing the whole document instead of just the header.

amyblais:
@hmhealey Agniva is currently in the middle of a customer incident, perhaps another dev can review this.

crspeller (approved):

mickmister (approved):

jgilliam17 (approved):
Thank you @hmhealey 
Tested, looks good to merge.
- Verified youtube link previews are displaying.

agnivade:
This broke master dependencies. If you run "go mod tidy" on master, it gives:

```
go: finding module for package github.com/hmhealey/go-opengraph/opengraph
go: found github.com/hmhealey/go-opengraph/opengraph in github.com/hmhealey/go-opengraph v0.0.0-20201209151214-f2d823730dba
go: github.com/hmhealey/go-opengraph@v0.0.0-20201209151214-f2d823730dba used for two different module paths (github.com/dyatlov/go-opengraph and github.com/hmhealey/go-opengraph)
```

I stumbled across this while doing https://mattermost.atlassian.net/browse/MM-31436. I assume you would have seen this error while running go mod tidy/go mod vendor. Was this something that got missed?

Please do not ignore warnings/errors for these commands. These cause unnecessary cycles for other developers to figure out what went wrong only to realize master itself is broken. If there is any difficulty in understanding why the error is happening, please feel free to reach out to the server team.

I have fixed it here: https://github.com/hmhealey/go-opengraph/pull/1. Please take a look. And do let me know if you have any questions.

tldr is that replace statements are not the same as forking a repo. Module paths are not supposed to change. That's what a replace statement is supposed to do. If you are changing a module path, directly replace the import statement too.

PR #16529, opened 2020-12-09
[MM-30851] Replace ExperimentalChannelSidebarOrganization with EnableLegacySidebar

devinbinnie:
#### Summary
Migrating to a new setting called `EnableLegacySidebar` to enable the old sidebar.
The new sidebar will be enabled by default system-wide from now on.

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-30851
#### Release Note
```release-note
Removed the `ExperimentalChannelSidebarOrganization` setting and added an `EnableLegacySidebar` setting. The new channel sidebar will be enabled system-wide by default.
```


nickmisasi (approved):
Exciting times! LGTM

config/client.go
```diff
@@ -61,7 +61,6 @@ func GenerateClientConfig(c *model.Config, telemetryID string, license *model.Li
 		props["ExperimentalChannelOrganization"] = strconv.FormatBool(false)
 	}
 
-	props["ExperimentalChannelSidebarOrganization"] = *c.ServiceSettings.ExperimentalChannelSidebarOrganization
```
hmhealey, line 64: Nothing to change here, but we'll have to be careful about merging this without the mobile changes since the mobile app still relies on this setting

hmhealey (approved):

config/client.go
hmhealey: Come to think of it, we might want to leave this in for a few releases and derive it from the new setting for new server/old mobile app compatibility

hmhealey (changes requested):

config/client.go
devinbinnie: I'm pretty sure the plan is to ignore the original setting and just force it on for everyone.
Either way, this won't be merged until after the rest of the pre-GA sidebar work is done, so we have some time to tweak if we want.

hmhealey: I'll let you know if this does need to change, but old mobile apps that won't support the new sidebar will still need some value for this or else they'll lose support for the new sidebar until they're updated.

hmhealey (approved):

PR #16530, opened 2020-12-10
Revert "implement and use striped LRU cache to lower mutex contention"

agnivade:
Reverts mattermost/mattermost-server#15764

Causes issues with machines having only 1 CPU.


agnivade:
Sent the actual fix here: https://github.com/mattermost/mattermost-server/pull/16531

PR #16531, opened 2020-12-10
MM-31326: Fix Striped LRU to have minimum of 1 bucket

agnivade:
https://mattermost.atlassian.net/browse/MM-31236

```release-note
NONE
```

### DIFF
--- a/app/server.go
+++ b/app/server.go
@@ -287,11 +287,15 @@ func NewServer(options ...Option) (*Server, error) {
 		return nil, errors.Wrapf(err, "Unable to connect to cache provider")
 	}
 
+	cpus := runtime.NumCPU() - 1
+	if cpus == 0 {
+		cpus = 1
+	}
 	var err error
 	if s.sessionCache, err = s.CacheProvider.NewCache(&cache.CacheOptions{
 		Size:           model.SESSION_CACHE_SIZE,
 		Striped:        true,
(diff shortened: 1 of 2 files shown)


app/server.go
```diff
@@ -287,11 +287,15 @@ func NewServer(options ...Option) (*Server, error) {
...
+	cpus := runtime.NumCPU() - 1
+	if cpus == 0 {
+		cpus = 1
+	}
 	var err error
 	if s.sessionCache, err = s.CacheProvider.NewCache(&cache.CacheOptions{
 		Size:           model.SESSION_CACHE_SIZE,
 		Striped:        true,
-		StripedBuckets: runtime.NumCPU() - 1,
+		StripedBuckets: cpus,
```
streamer45, line 298: If we define an integer max function this might become a bit cleaner:

```go
StripedBuckets: utils.MaxInt(runtime.NumCPU() - 1, 1)
```

`math.Max` would work too but it's definitely more cluttered:

```go
StripedBuckets: int(math.Max(float64(runtime.NumCPU() - 1), 1))
```

streamer45 (approved):
:+1:

app/server.go
agnivade: Went with a local function, and not adding more junk to `utils`. "A little copying is better than a little dependency" ;)

PR #16539, opened 2020-12-11
[MM-30539] New renewal link logic

ethervoid:
#### Summary

Endpoint and logic that returns the renewal link to be used to start the license renewal process.

#### Ticket Link

https://mattermost.atlassian.net/browse/MM-30539

#### Related PRs

https://github.com/mattermost/mattermost-api-reference/pull/594

#### Release Note

```release-note
Added the logic to generate a new renewal link that could be used to start the renewal process
```


app/license.go
```diff
@@ -245,3 +256,62 @@ func (s *Server) RequestTrialLicense(trialRequest *model.TrialLicenseRequest) *m
...
+		})
+		if err != nil {
+			return "", model.NewAppError("GenerateRenewalToken", "ent.license.app_error", nil, err.Error(), http.StatusInternalServerError)
+		}
+		return tokenString, nil
+	} else {
+		// Clean renewal token if there is no license present
+		_, _ = s.Store.System().PermanentDeleteByName(model.SYSTEM_LICENSE_RENEWAL_TOKEN)
+		return "", model.NewAppError("GenerateRenewalToken", "ent.license.no_license", nil, "", http.StatusBadRequest)
+	}
```
nickmisasi, line 300: Small refactor suggestion: Move this into an if statement at the top of the function (with a check for license == nil) and move the logic inside the current if statement outside of the if/else

app/license.go
```diff
@@ -245,3 +256,62 @@ func (s *Server) RequestTrialLicense(trialRequest *model.TrialLicenseRequest) *m
 
 	return nil
 }
+
+// GenerateRenewalToken returns the current active token or generate a new one if
+// the current active one has expired
+func (s *Server) GenerateRenewalToken(expiration time.Duration) (string, *model.AppError) {
+	if license := s.License(); license != nil {
+		currentToken, err := s.Store.System().GetByName(model.SYSTEM_LICENSE_RENEWAL_TOKEN)
```
nickmisasi, line 264: You don't actually check `err`, should either check it or `_` it

app/license.go
```diff
@@ -245,3 +256,62 @@ func (s *Server) RequestTrialLicense(trialRequest *model.TrialLicenseRequest) *m
...
+			Name:  model.SYSTEM_LICENSE_RENEWAL_TOKEN,
+			Value: tokenString,
+		})
+		if err != nil {
+			return "", model.NewAppError("GenerateRenewalToken", "ent.license.app_error", nil, err.Error(), http.StatusInternalServerError)
+		}
+		return tokenString, nil
+	} else {
+		// Clean renewal token if there is no license present
+		_, _ = s.Store.System().PermanentDeleteByName(model.SYSTEM_LICENSE_RENEWAL_TOKEN)
```
nickmisasi, line 298: Could be worth handling an err from this one. While unlikely, it's possible

nickmisasi (commented):
Overall LGTM - some suggestions for slight refactors

nickmisasi (approved):
LGTM!

totongcute (approved):

totongcute (approved):

api4/license.go
```diff
@@ -210,3 +213,36 @@ func requestTrialLicense(c *Context, w http.ResponseWriter, r *http.Request) {
 
 	ReturnStatusOK(w)
 }
+
+func requestRenewalLink(c *Context, w http.ResponseWriter, r *http.Request) {
+	auditRec := c.MakeAuditRecord("requestRenewalLink", audit.Fail)
+	defer c.LogAuditRec(auditRec)
+	c.LogAudit("attempt")
+
+	if !c.App.SessionHasPermissionTo(*c.App.Session(), model.PERMISSION_SYSCONSOLE_READ_ABOUT) {
```
marianunez, line 222: Are the notifications for license renewal limited on this `PERMISSION_SYSCONSOLE_READ_ABOUT ` permission? The banner notifications will have the button to "Renew Now" and would need to generate a link

api4/license.go
```diff
@@ -210,3 +213,36 @@ func requestTrialLicense(c *Context, w http.ResponseWriter, r *http.Request) {
...
+		c.Err = model.NewAppError("requestRenewalLink", "api.restricted_system_admin", nil, "", http.StatusForbidden)
+		return
+	}
+
+	renewalToken, err := c.App.Srv().GenerateRenewalToken(app.JWTDefaultTokenExpiration)
+	if err != nil {
+		c.Err = err
+		return
+	}
+	licenseRenewalURL := *c.App.Srv().Config().ServiceSettings.LicenseRenewalURL
```
marianunez: Would there be any benefit of using the base url from `CloudSettings.CWSUrl`?

app/license.go
```diff
@@ -5,21 +5,32 @@ package app
...
 	"github.com/mattermost/mattermost-server/v5/model"
 	"github.com/mattermost/mattermost-server/v5/utils"
+
+	"github.com/pkg/errors"
 )
 
 const (
-	requestTrialURL = "https://customers.mattermost.com/api/v1/trials"
-	LicenseEnv      = "MM_LICENSE"
+	requestTrialURL           = "https://customers.mattermost.com/api/v1/trials"
```
marianunez, line 22: I know this code was pre-existing, but this base URL for prod CWS should be configured somewhere

app/license.go
```diff
@@ -5,21 +5,33 @@ package app
...
+	LicenseEnv                = "MM_LICENSE"
+	LicenseRenewalURL         = "https://customers.mattermost.com/subscribe/renew"
```
marianunez, line 24: Why remove the configuration for the URL? I believe it's a better approach than having this hardcoded here

app/license.go
```diff
@@ -5,21 +5,33 @@ package app
...
+	JWTDefaultTokenExpiration = 7 * 24 * time.Hour // 7 days of expiration
```
marianunez, line 25: @jasonblais heads up that the license renewal links have a validity of 7 days and will not be valid after this timeframe for security. Customer Support/Sales can generate another if the customer didn't perform the renewal within that timeframe. 

Open to suggestions on the amount of time.

app/license.go
```diff
@@ -245,3 +257,66 @@ func (s *Server) RequestTrialLicense(trialRequest *model.TrialLicenseRequest) *m
 
 	return nil
 }
+
+// GenerateRenewalToken returns the current active token or generate a new one if
+// the current active one has expired
+func (s *Server) GenerateRenewalToken(expiration time.Duration) (string, *model.AppError) {
```
marianunez, line 263: Is there anything here we should include in the enterprise repo instead?

api4/license.go
```diff
@@ -210,3 +213,35 @@ func requestTrialLicense(c *Context, w http.ResponseWriter, r *http.Request) {
 
 	ReturnStatusOK(w)
 }
+
+func requestRenewalLink(c *Context, w http.ResponseWriter, r *http.Request) {
```
marianunez, line 217: Should we also check if they are licensed before continuing the license renewal link logic?

app/license.go
```diff
@@ -245,3 +257,66 @@ func (s *Server) RequestTrialLicense(trialRequest *model.TrialLicenseRequest) *m
...
+	license := s.License()
+	if license == nil {
+		// Clean renewal token if there is no license present
+		_, err := s.Store.System().PermanentDeleteByName(model.SYSTEM_LICENSE_RENEWAL_TOKEN)
+		if err != nil {
+			mlog.Error("error removing the renewal token", mlog.Err(err))
+		}
```
marianunez, line 270: nit:

```suggestion
		if _, err := s.Store.System().PermanentDeleteByName(model.SYSTEM_LICENSE_RENEWAL_TOKEN); err != nil {
			mlog.Error("error removing the renewal token", mlog.Err(err))
		}
```

app/license.go
```diff
@@ -245,3 +257,66 @@ func (s *Server) RequestTrialLicense(trialRequest *model.TrialLicenseRequest) *m
...
+		return "", model.NewAppError("GenerateRenewalToken", "ent.license.generate_renewal_token.no_license", nil, "", http.StatusBadRequest)
+	}
+
+	currentToken, _ := s.Store.System().GetByName(model.SYSTEM_LICENSE_RENEWAL_TOKEN)
```
marianunez, line 274: What is the purpose of storing the token? Will storing the token start counting towards the expiration time? When will we make the request to get the link with the generated token? when they click the "Renew Now" button or just to show the button in the System Console?

marianunez (commented):
Great stuff @ethervoid 🎉

api4/license.go
ethervoid: I think it should. TBH I don't know how the notifications work, should be addressed in another PR, but it should only be shown to the admin so I think that permission should do the trick

app/license.go
ethervoid: Agree but we should address that change in another PR

app/license.go
ethervoid: After talking with @srkgupta offline he suggested to better have it hardcoded to avoid the admins introduce a phishing URL or something similar.

app/license.go
ethervoid: The token will be available for an amount of time, as Maria said, but there is no problem to go into de system console and generate a new one. 

From my PoV this amount should be smaller, like 1 day for example but I know the enterprise times for renewal could be greater than expected.

app/license.go
ethervoid: No need. I've talked in the server guild and they want the license code in the team edition unless there is a good reason for it and security by obscurity is not one

api4/license.go
ethervoid: The `GenerateRenewalToken` checks for license

app/license.go
ethervoid: > What is the purpose of storing the token? 

Storing the token ensures that only one token is available at the same time, request from security team.

> Will storing the token start counting towards the expiration time?

Yes, when the token is generated we're adding the expiration time

> When will we make the request to get the link with the generated token? when they click the "Renew Now" button or just to show the button in the System Console?

The idea is when the button is clicked and the URL is shown or  the user is redirected

api4/license.go
marianunez: Looks odd to me that the user that has read rights (and possibly not write permissions) to the license section can generate the link. With the new System Roles, non-admins could have rights to `PERMISSION_SYSCONSOLE_READ_ABOUT`.

 I'm 1/5 that we should probably check `model.PERMISSION_SYSCONSOLE_WRITE_ABOUT` here

ethervoid: I'll check that

app/license.go
marianunez: Thanks for the insights here 👍

app/license.go
```diff
@@ -245,3 +257,65 @@ func (s *Server) RequestTrialLicense(trialRequest *model.TrialLicenseRequest) *m
 
 	return nil
 }
+
+// GenerateRenewalToken returns the current active token or generate a new one if
+// the current active one has expired
+func (s *Server) GenerateRenewalToken(expiration time.Duration) (string, *model.AppError) {
+	license := s.License()
```
marianunez, line 264: Should we restrict creating a renewal link if the license is cloud? Doesn't seem to make sense for that case

ethervoid: I'm 1/5 on that. I'd add it to be consistent

app/license.go
jasonblais: @marianunez @ethervoid At what point is the renewal link generated, and is a new one automatically created once the previous link expires?

ethervoid: @jasonblais is not automatic. The user should go to the screen and generate a new token

jasonblais: @ethervoid I tried to create a test server to see the screen, but looks like it's not created at the moment. Would you be open to sharing a screenshot of what it looks like?

As a first reaction, this doesn't seem obvious from the user's perspective. I think most admins are going to click on the 'Renew license now' button in the header and be confused why the link doesn't work.

image

Is there a possibility where after 7 days, if an admin clicks on the `renew license now` button, we auto-generate a new token for them, then redirect the admin to the renewal page? This would make the experience faster and more obvious, following our [design principles](https://docs.mattermost.com/developer/fx-guidelines.html#fast-obvious-forgiving). We could also add a note somewhere on the renewal page saying something like `This link is only valid for 7 days.`.

Let me know what you think.

cc @alexsiclari as well for thoughts.

marianunez: @jasonblais as soon as the user clicks that button, the link will be valid because we will call the endpoint added in this PR which will always return a valid url (if it generated one before and it's expired it will regenerate a new one).

The expiration affects mostly when the link is handed out directly, either by email by us directly (i.e the case for people that have not upgraded 5.32) or if we offer someway they can copy the link for airgapped cases (not scoped for V1). 

In the case they go to an expired link, we need to show them an appropriate error message in the customer portal to request a new link.

ethervoid: @jasonblais This PR is only the server side logic, we're not including yet the frontend/webapp part here.

For the rest of the questions, is what @marianunez said

jasonblais: Great, that sounds good to me. 7 days is fine in this use case - I'll check with Jeff Johnson as well, but no need to wait for that.

> In the case they go to an expired link, we need to show them an appropriate error message in the customer portal to request a new link.

Yes, please work with Alex Siclari on the design. We may also want to add a `This link is only valid for 7 days.` on the customer portal renewal page, but I'm 0/5 on that.

alexsiclari: Yes, let me know what you need. I will accommodate the text in some way.

jasonblais (approved):

srkgupta (approved):
Looks good from security perspective. Thanks @ethervoid

app/license.go
jasonblais: @ethervoid @marianunez confirmed with Jeff that 7-day expiry is fine. UX design is being discussed in the Customer Portal channel.

Resolving this comment.

marianunez (approved):
Thanks @ethervoid!

ethervoid:
@stevemudie what do you think on testing this post-merge with the CWS part merged too?

stevemudie:
@ethervoid Yes, that sounds good to me.

alexsiclari (approved):

PR #16542, opened 2020-12-14
MM-31323: Fix racy test TestHandleCommandResponsePost

agnivade:
In https://github.com/mattermost/mattermost-server/pull/16089,
we added logic to modify the notification text in the presence
of fallback text.

Unfortunately, that created a race condition during publishing
the post via websockets. To fix this, we create a clone of the post
before sending it off for push notifications.

https://mattermost.atlassian.net/browse/MM-31323

```release-note
NONE
```

### DIFF
--- a/app/notification.go
+++ b/app/notification.go
@@ -210,7 +210,7 @@ func (a *App) SendNotifications(post *model.Post, team *model.Team, channel *mod
 	}
 
 	notification := &PostNotification{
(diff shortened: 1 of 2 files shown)


isacikgoz (approved):
Looks good to me 🏁

streamer45 (approved):
:+1:

PR #16550, opened 2020-12-15
[MM-31357] Avoid returning empty appError when autocompleting users

mgdelacroix:
#### Summary
As part of the `UserStore` error migration, we started returning `error` from the `searchlayer`, inadvertently returning an empty `AppError` instance in the user autocomplete function, that was then caught as `!= nil` by the caller, causing a panic.

`AppError` strikes yet again.

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-31357

#### Release Note
```release-note
NONE
```


saturninoabril (approved):
Awesome, tested and passed!

jespino (approved):

isacikgoz (approved):
LGTM 👍

PR #16561, opened 2020-12-16
MM-31369: Fix racy test TestFileStoreWatcher

agnivade:
Since this test had t.Parallel set, it would run the sub-tests
in parallel. The issue was that the same file path was being written
from the 2 sub-tests causing duplicate firing of the event handlers.

This would cause a race where the store would be closing, and at the same time,
an event handler would fire, trying to read the fs.watcher field.

Having two separate file paths resolves this.

https://mattermost.atlassian.net/browse/MM-31369

```release-note
NONE
```

### DIFF
--- a/config/file_test.go
+++ b/config/file_test.go
@@ -884,10 +884,9 @@ func TestFileStoreWatcherEmitter(t *testing.T) {
 
 	t.Parallel()
 
-	path, tearDown := setupConfigFile(t, emptyConfig)
-	defer tearDown()
-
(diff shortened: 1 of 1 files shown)


streamer45 (approved):
Nice :+1:

agnivade:
Gentle ping on this @crspeller

crspeller (approved):

crspeller:
@agnivade Sorry for the holiday delay!

PR #16562, opened 2020-12-16
MM-31370: Fix racy TestClusterShutdownRace

agnivade:
This race wasn't specific to this test, but somehow got triggered by it.
This was a critical race because the PluginsLock is widely used
throughout the codebase to gate access to the PluginsEnvironment.

However, a config listener can often run in its own goroutine leading to
a wide variety of races.

https://mattermost.atlassian.net/browse/MM-31370

```release-note
NONE
```


agnivade:
@streamer45 / @crspeller - I am a bit wary of introducing recursive locking due to this. I know there are hundreds of codepaths that use `s.PluginsEnvironment` to do various things. Please give it a good look. Thanks!

app/server.go
```diff
@@ -484,6 +484,8 @@ func NewServer(options ...Option) (*Server, error) {
 		pluginsEnvironment.InitPluginHealthCheckJob(*s.Config().PluginSettings.Enable && *s.Config().PluginSettings.EnableHealthCheck)
 	}
 	s.AddConfigListener(func(_, c *model.Config) {
+		s.PluginsLock.RLock()
+		defer s.PluginsLock.RUnlock()
```
streamer45, line 488: It looks safe to me but we could also avoid the `defer` since we only need to lock to read the env from the server struct.

agnivade: makes sense to me

streamer45 (approved):
Looks good :+1:

app/server.go
```diff
@@ -484,7 +484,9 @@ func NewServer(options ...Option) (*Server, error) {
 		pluginsEnvironment.InitPluginHealthCheckJob(*s.Config().PluginSettings.Enable && *s.Config().PluginSettings.EnableHealthCheck)
 	}
 	s.AddConfigListener(func(_, c *model.Config) {
+		s.PluginsLock.RLock()
 		pluginsEnvironment := s.PluginsEnvironment
+		s.PluginsLock.RUnlock()
```
crspeller, line 489: Is this enough to prevent the race? This is a shallow copy of environment and the actual action we take here is to use one of the pointers in the struct.

agnivade: Yes, it does. The race is about accessing the `PluginsEnvironment` field from inside the Server struct.

crspeller (approved):

agnivade:
@lindalumitchell - would be helpful to have some plugin related tests done on this one. This touches some very hairy code.

agnivade:
Gentle reminder on this @lindalumitchell

lindalumitchell (approved):
Thanks for the ping and sorry for the delay! Did some plugins smoke testing, no issues found. 👍

PR #16564, opened 2020-12-16
MM-28549: Bump up postgres version

agnivade:
https://mattermost.atlassian.net/browse/MM-28549

```release-notes
Mattermost does not support Postgres 9.4 anymore. The minimum
required version for Postgres will be 10.
```

### DIFF
--- a/build/docker-compose.common.yml
+++ b/build/docker-compose.common.yml
@@ -17,7 +17,7 @@ services:
       timeout: 10s
       retries: 3
   postgres:
-    image: "postgres:9.4"
+    image: "postgres:10"
     restart: always
     networks:
       - mm-test


agnivade:
Finally it's here :tada: 

@metanerd - I checked gitlab and circleCI files. All of them seem to setup the environment from docker-compose. I believe there is nothing else to change. But please let me know if I am mistaken and there is some other part of the pipeline that needs modification.

metanerd (approved):
That should indeed be everything needed. LGTM

grundleborg (approved):
LGTM. @agnivade can you also add the other ticket to make startup of the server fail fast on old Postgres versions? I think it would be good to merge both around the same time so it's a really clean break.

agnivade:
@grundleborg - I was thinking to make a gradual change to soak this change first a little in CI. So that we are 100% confident that everything works fine with pg10. And then make the fail fast PR. But 0/5 on making it right now. Thoughts?

grundleborg:
@agnivade I see what you mean. Yes, I'm happy for this to be merged today and then you add the fail fast in say, 1 week's time assuming this didn't get reverted for any reason. Is that what you are thinking?

agnivade:
Yep, that's right. :)

amyblais:
@agnivade I noticed v5.31 milestone on the ticket. Since v5.31 will be based on the cloud release we did last week https://community.mattermost.com/private-core/pl/khyfkk1h1pbpmcon7g7auhjmnr, do I move this to v5.32 or is this a high priority to cherry-pick separately to v5.31?

agnivade:
Nope, no need to cherry-pick. The milestone was old I think.

PR #16568, opened 2020-12-16
MM-31401  Add Shared flag to Channels table

wiggin77:
#### Summary
This PR adds a new flag to the Channels table `Shared`.  

This is in support of the upcoming Shared Channels feature and is needed to efficiently determine if a channel needs to sync after Post added/edited/deleted.  Values may be nil, false, true.  `(*Channel).IsShared` has been added as well.

The flag will not be set or queried until the Shared Channels feature branch is merged.

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-31401

#### Release Note
```release-note
NONE
```

### DIFF
--- a/store/sqlstore/upgrade.go
+++ b/store/sqlstore/upgrade.go
@@ -956,6 +956,7 @@ func upgradeDatabaseToVersion531(sqlStore *SqlStore) {
 	// if shouldPerformUpgrade(sqlStore, VERSION_5_30_0, VERSION_5_31_0) {
 	// allow 10 files per post
 	sqlStore.AlterColumnTypeIfExists("Posts", "FileIds", "text", "varchar(300)")
+	sqlStore.CreateColumnIfNotExists("Channels", "Shared", "tinyint(1)", "boolean", "0")
 	// saveSchemaVersion(sqlStore, VERSION_5_31_0)
 	// }
 }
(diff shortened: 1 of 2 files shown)


store/sqlstore/upgrade.go
```diff
@@ -956,6 +956,7 @@ func upgradeDatabaseToVersion531(sqlStore *SqlStore) {
 	// if shouldPerformUpgrade(sqlStore, VERSION_5_30_0, VERSION_5_31_0) {
 	// allow 10 files per post
 	sqlStore.AlterColumnTypeIfExists("Posts", "FileIds", "text", "varchar(300)")
+	sqlStore.CreateColumnIfNotExists("Channels", "Shared", "tinyint(1)", "boolean", "0")
```
nronas, line 959: Not sure if am reading this correctly, but adding a new column with a default non-null value will lock the table, because it needs an exclusive lock, since `Shared` can be null, can we just not set a default value?

nronas: I think we should be fine since if I am reading our flow correctly, the migrations will run as the first step of an app boot up.

wiggin77: Yes, good point.  Makes me wonder about the other flags added in recent versions that do set default values.

nronas (approved):
LGTM

store/sqlstore/upgrade.go
nronas: I think so we operate under the assumption that migrations run first and then the app boots, so i guess there is no user traffic coming in to the tables while those are being modified

wiggin77: Correct, however some older Postgres (including 10) will rewrite the whole table with defaults.  If this was the Posts table that could take many minutes.  Channels table includes all archived channels - it's essentially a grow only table - so some installations may get large.  We don't need a default here so I'll remove it.  Good catch.

nronas: Yeap :100 %

wiggin77:
No QA needed at this time.  New column added to Channels table is not set or queried until Shared Channels feature branch is merge.

streamer45 (approved):

PR #16569, opened 2020-12-16
[MM-28000] Fix system manager can download export files without right permissions

hahmadia:
#### Summary

System manager can retrieve compliance export files and download them even though he can't see the compliance section in the system console.

#### Ticket Link

https://mattermost.atlassian.net/browse/MM-28000

#### Release Note

```release-note
NONE
```


api4/job.go
```diff
@@ -70,6 +70,11 @@ func downloadJob(c *Context, w http.ResponseWriter, r *http.Request) {
 		return
 	}
 
+	if job.Type == model.JOB_TYPE_MESSAGE_EXPORT && !c.App.SessionHasPermissionTo(*c.App.Session(), model.PERMISSION_SYSCONSOLE_READ_COMPLIANCE) {
+		c.SetPermissionError(model.PERMISSION_SYSCONSOLE_READ_COMPLIANCE)
+		return
+	}
```
sbishel, line 76: This should be moved above call to `GetJob`.

sbishel (changes requested):

api4/job.go
hahmadia: Hey @sbishel !

Not sure if I fully understand but if you meant put that statement above the GetJob call, I wanted too however I don't have access to the job.type property at that point which is why I put it after :slightly_frowning_face: We only have Job ID before then and we use that to get the job. Unless I've misunderstood your meaning 😅

sbishel: You're right. Thanks for explanation.

sbishel (approved):

mkraft (commented):
In this PR I think you've raised something important: the fact that there's an existing gap in the granularity of job with relation to the system console permissions. 

In a sense this is like declaring a mapping between `PERMISSION_SYSCONSOLE_*_COMPLIANCE` granting permission to `JOB_TYPE_MESSAGE_EXPORT` specifically (rather than to all jobs).

I'm wondering if there are other sections that also should also be limited. For example, in the `SysconsoleAncillaryPermissions` mapping [`PERMISSION_SYSCONSOLE_READ_ENVIRONMENT` and `PERMISSION_SYSCONSOLE_READ_AUTHENTICATION`](https://github.com/mattermost/mattermost-server/blob/master/model/role.go#L64-L69) also have access to all jobs (well not to `JOB_TYPE_MESSAGE_EXPORT` anymore but all the others). Should they continue to have access to all the other jobs? Or should we be addressing all of this section-to-job-type access granularity at once?

hahmadia:
Hey @sbishel ,

I had a meeting with Martin last week regarding this PR and he proposed a much better change to what I originally had. 

Martin explained that those SYSCONSOLE permissions are primarily for the frontend and if we ever want say a bot to download compliance files, we would need to give it this SYSCONSOLE permission which could make thing confusing.

We would rather map these SYSCONSOLE permission to a more specific permissions that way in the future if a bot needs it, we could give it this specific permission and this will falls in line with our current pattern. This happens through our current implementation of ancililary permissions (which prior to this PR, I wasn't aware this was its purpose).

Since this PR primarily is concerning the fact that we are allowed to download compliance reports when we shouldn't be allowed, I mapped `SYSCONSOLE_READ_COMPLIANCE` to `read_jobs` and a new permissions specifically allowing us to download the compliance report and `SYSCONSOLE_WRITE_COMPLIANCE` to `manage_jobs`. 

This new permission is used in the `downloadJob` api endpoint. Since that endpoint is currently only being used to download compliance reports, it makes sense to remove the current `read_jobs` permission check (since this could bring up security concerns in the future allowing us to download any new job that follows the same output structure as compliance reports) and check to see if we are downloading a compliance report and also check to see if we have the appropriate permission. If it's not a compliance report, then we throw an error saying that the job they are trying to download is not supported (this further secures us from accidentally giving user permission to download a job they shouldn't be allowed too)

Anyhow, requesting a re-review as I feel this implementation makes more sense than my previous one. Thanks!

mkraft (approved):
Looks good. Nice work.

api4/apitestlib.go
```diff
@@ -420,6 +436,10 @@ func (th *TestHelper) CreateWebSocketSystemAdminClient() (*model.WebSocketClient
 	return model.NewWebSocketClient4(fmt.Sprintf("ws://localhost:%v", th.App.Srv().ListenAddr.Port), th.SystemAdminClient.AuthToken)
 }
 
+func (th *TestHelper) CreateWebSocketSystemManagerClient() (*model.WebSocketClient, *model.AppError) {
```
sbishel, line 439: Is this used?

sbishel (approved):

api4/apitestlib.go
hahmadia: Nope. I just added it since it was added for System Admin. Is it better to remove or keep in case it needs to be used in the future?

hahmadia: @sbishel

furqanmlk (approved):
Working as expected

image


Added new test case
[MM-T3667](https://mattermost.atlassian.net/projects/MM?selectedItem=com.atlassian.plugins.atlassian-connect-plugin:com.kanoah.test-manager__main-project-page#!/testCase/MM-T3667)

PR #16574, opened 2020-12-17
MM-30836: Feature flag for remote cluster service

agnivade:
https://mattermost.atlassian.net/browse/MM-30836

```release-note
NONE
```

### DIFF
--- a/app/server.go
+++ b/app/server.go
@@ -721,6 +721,7 @@ func (s *Server) startInterClusterServices(license *model.License) {
 
 	if !*s.License().Features.SharedChannels {
 		mlog.Debug("License does not have shared channels enabled")
+		return
 	}
 
 	var err error
--- a/app/server.go
+++ b/app/server.go
@@ -719,19 +719,25 @@ func (s *Server) startInterClusterServices(license *model.License) {
 		return
 	}
 
-	var err error
-
-	// TODO: check remote cluster service feature flag and license (MM-30836 & MM-30838)
-	s.remoteClusterService, err = remotecluster.NewRemoteClusterService(s)
-	if err != nil {
-		mlog.Error("Error initializing Remote Cluster Service", mlog.Err(err))
-		return
+	if !*s.License().Features.SharedChannels {
+		mlog.Debug("License does not have shared channels enabled")
 	}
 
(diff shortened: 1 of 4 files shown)


nronas (commented):
Logic wise this looks good to me. But reading the ticket I got a bit confused, do we need just a license check here or an actual feature flag? https://github.com/mattermost/mattermost-server/blob/master/model/feature_flags.go

agnivade:
I think that's a separate thing, which is the new "feature flags" functionality we introduced. This is just about adding a dedicated flag in config.json, which also technically qualifies as a "feature flag".

> So basically we can have SharedChannels disabled with RemoteClusterService enabled since this is a possibility because right now those flags are not mutually exclusive.

I guess so? I'll defer to @wiggin77

wiggin77:
> I think that's a separate thing, which is the new "feature flags" functionality we introduced. This is just about adding a dedicated flag in config.json, which also technically qualifies as a "feature flag".
> 
> > So basically we can have SharedChannels disabled with RemoteClusterService enabled since this is a possibility because right now those flags are not mutually exclusive.
> 
> I guess so? I'll defer to @wiggin77

That is correct.  Remote Cluster Service can be enabled while Shared Channels disabled, however not the other way around.  Other services will share content using Remote Cluster Service hence its has independent flags.

wiggin77 (approved):
LGTM  :+1:

nronas (approved):
LGTM

wiggin77:
@agnivade, sorry to drag this PR back up, but I'm wondering if this really added a Feature flag.  I re-read the spec https://docs.google.com/document/d/1B2oOeQsH0TiWObAMbHw_jLOSUjMRNfd3j8Df-2QKkDc/ and apparently feature flags live in the Config.Features struct and are populated externally.  This PR added a config setting, which is good, but not a feature flag.  Of course I could be missing something about how feature flags work.

agnivade:
Yep, it's not a feature flag. That's what I said here https://github.com/mattermost/mattermost-server/pull/16574#issuecomment-747328390. I thought you wanted a dedicated config setting. So is it that you want all sites to have remote cluster service to be on eventually?

PR #16578, opened 2020-12-18
[MM-27622] Publish messages as WarnMetricsBot when a user is required

mgdelacroix:
#### Summary
This PR adds the channel restore and channel privacy endpoints to local mode using the Advisor Bot to post the relevant messages when the endpoints are used.

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-24864
https://mattermost.atlassian.net/browse/MM-24600
https://mattermost.atlassian.net/browse/MM-27440

#### Release Note
```release-note
Added channel restore and channel privacy change endpoints to local mode using the advisor bot
```

### DIFF
--- a/app/app.go
+++ b/app/app.go
@@ -324,6 +324,12 @@ func (a *App) getWarnMetricStatusAndDisplayTextsForId(warnMetricId string, T i18
 }
 
 func (a *App) notifyAdminsOfWarnMetricStatus(warnMetricId string, isE0Edition bool) *model.AppError {
+	// get warn metrics bot
+	warnMetricsBot, err := a.GetWarnMetricsBot()
+	if err != nil {
+		return err
+	}
+
 	perPage := 25
 	userOptions := &model.UserGetOptions{
 		Page:     0,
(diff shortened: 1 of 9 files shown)


app/app.go
```diff
@@ -324,6 +324,12 @@ func (a *App) getWarnMetricStatusAndDisplayTextsForId(warnMetricId string, T i18
 }
 
 func (a *App) notifyAdminsOfWarnMetricStatus(warnMetricId string, isE0Edition bool) *model.AppError {
+	// get warn metrics bot
+	warnMetricsBot, err := a.GetWarnMetricsBot()
```
catalintomai, line 328: Not sure how strict we are in general with extra DB calls, - just pointing out that by factoring this out we add an extra GetUsers() call here.

jespino: Yes, we are using this call in multiple places, and makes no sense to get over and over again the same boot, probably we can cache this as a module variable or as an `App` attribute. So we can get the metrics bot one time, and unless the metrics bot have to change for any reason, we don't need to reload it (actually, if we doubt about the cache, we can just simply make that cache automatically expire every minute or 2 minutes or 10 minutes or something like that).

jespino (approved):
LGTM, but I would like to see some optimization around the `GetWarnMetricsBot`

app/app.go
mgdelacroix: I think this is going to be a case of premature optimization as the feature is planned now. The admin notification is now going to be disabled, and the usage that `mmctl` is quite eventual, so a two minutes cache probably won't ever get hit.

I'd suggest to proceed with the current implementation and optimize if needed. Thoughts?

PR #16580, opened 2020-12-18
Update db schema version for MM-31401 and MM-27312

streamer45:
#### Summary

Since both mentioned PRs were not included in 5.31 I've bumped the schema version for the DB upgrade steps so that those happen in the expected version.


wiggin77 (approved):
:+1:   Thanks!

hmhealey (approved):

PR #16587, opened 2020-12-22
MM-31062: Rewrite empty string checks to be more idiomatic

agnivade:
https://mattermost.atlassian.net/browse/MM-31062

```release-note
NONE
```


agnivade:
A careful review needed on this one :P We should only be changing string checks.

isacikgoz (approved):
Looks more idiomatic to me 👍

PR #16595, opened 2020-12-23
MM-30836: Use feature flags for RemoteClusterService

agnivade:
We use a proper feature flag for toggling this feature.

### DIFF
--- a/app/server.go
+++ b/app/server.go
@@ -729,7 +729,7 @@ func (s *Server) startInterClusterServices(license *model.License) {
 	var err error
 
 	// TODO: check remote cluster service license (MM-30838)
-	if *s.Config().ExperimentalSettings.EnableRemoteClusterService {
+	if *s.Config().ExperimentalSettings.EnableRemoteClusterService && s.Config().FeatureFlags.EnableRemoteClusterService {
 		s.remoteClusterService, err = remotecluster.NewRemoteClusterService(s)
 		if err != nil {
 			mlog.Error("Error initializing Remote Cluster Service", mlog.Err(err))
--- a/config/client.go
+++ b/config/client.go
@@ -202,7 +202,7 @@ func GenerateClientConfig(c *model.Config, telemetryID string, license *model.Li
 
 		if *license.Features.SharedChannels {
 			props["ExperimentalSharedChannels"] = strconv.FormatBool(*c.ExperimentalSettings.EnableSharedChannels)
-			props["ExperimentalRemoteClusterService"] = strconv.FormatBool(*c.ExperimentalSettings.EnableRemoteClusterService)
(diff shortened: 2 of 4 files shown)


wiggin77:
Looks good @agnivade .  Can we also leave the `Config().ExperimentalSettings.EnableRemoteClusterService` in place as well?  We'll need both.  On-prem customers would enable/disble via that instead of feature flags.

agnivade:
On-prem customers can also use feature flags via environment variables as per my understanding. Do we explicitly want to have both? That might defeat the purpose of having a feature flag. Either it's a config, or a feature flag.

wiggin77:
> On-prem customers can also use feature flags via environment variables as per my understanding. Do we explicitly want to have both? That might defeat the purpose of having a feature flag. Either it's a config, or a feature flag.

Feature flags eventually get removed whereas we'll always need a way to disable these services.  I'm not clear on whether a config setting would suffice and obviate the need for feature flags.  Feature flags would allow cloud team to selectively enable via  Split. We can ask @crspeller what the indention is... do we need config setting and feature flags for enabling/disabling a service?

agnivade:
I see. Then using a config flag is the correct way for this. Feature flags are meant to be like temporary things for experimental stuff which is eventually meant to be stable and enabled. That's why I asked originally whether you want a feature flag or not in the first place. If this is a permanent thing, then it's not a feature flag.

wiggin77:
> I see. Then using a config flag is the correct way for this. Feature flags are meant to be like temporary things for experimental stuff which is eventually meant to be stable and enabled. That's why I asked originally whether you want a feature flag or not in the first place. If this is a permanent thing, then it's not a feature flag.

I still think we need both feature flags and config.  For example, when cloud team wants to try out these services they would enable in config, but use Split to activate for specific clusters.  The feature flag spec describes this use case.

agnivade:
Sure.

agnivade:
So which one should take priority? Feature flag or the config? Should we document this somewhere?

crspeller:
@wiggin77 Does an experimental config and a feature flag make sense? 
I can see a regular config setting plus a feature flag. The config setting is just part of the feature.  But I feel like feature flags replace experimental config settings. Like @agnivade pointed out on-prem customers can enable feature flags using environment variables. 

Maybe I don't understand the rollout plan though. What's the plan for rolling this feature out to customers?

wiggin77:
> @wiggin77 Does an experimental config and a feature flag make sense?
> I can see a regular config setting plus a feature flag. The config setting is just part of the feature. But I feel like feature flags replace experimental config settings. Like @agnivade pointed out on-prem customers can enable feature flags using environment variables.
> 
> Maybe I don't understand the rollout plan though. What's the plan for rolling this feature out to customers?

Thanks @crspeller.  There will always need to be a way to enable/disable these services via config, so eventually these will stop being experimental config items and move into the main config set.  IMO, the feature flags are not needed for on-prem but for cloud.  Cloud would enable the services in config, but use feature flags to enable for select customers.

wiggin77 (approved):
LGTM :+1:

PR #16597, opened 2020-12-23
MM-31356: Add a minimum required version check for Postgres

agnivade:
To keep conformance with our failing fast and obvious philosophy,
we add a check to prevent Mattermost server from starting
if the postgres version is below 10.0.

This gives customers a chance to upgrade their database before upgrading
their Mattermost version, than to run into weird compatibility issues
after they have finished the upgrade.

https://mattermost.atlassian.net/browse/MM-31356

```release-note
NONE
```

### DIFF
--- a/app/server.go
+++ b/app/server.go
@@ -16,6 +16,7 @@ import (
 	"path"
 	"path/filepath"
 	"runtime"
+	"strconv"
 	"strings"
 	"sync"
 	"sync/atomic"
@@ -319,6 +320,19 @@ func NewServer(options ...Option) (*Server, error) {
 	if s.newStore == nil {
 		s.newStore = func() (store.Store, error) {
 			s.sqlStore = sqlstore.New(s.Config().SqlSettings, s.Metrics)
+			if s.sqlStore.DriverName() == model.DATABASE_DRIVER_POSTGRES {
+				ver, err := s.sqlStore.GetDbVersion(true)
+				if err != nil {
+					return nil, errors.Wrap(err, "cannot get DB version")
+				}
+				intVer, err := strconv.Atoi(ver)
+				if err != nil {
+					return nil, errors.Wrap(err, "cannot parse DB version")
+				}
+				if intVer < sqlstore.MINIMUM_REQUIRED_POSTGRES_VERSION {
+					return nil, fmt.Errorf("minimum required postgres version is 10.0; found %s", ver)
+				}
+			}
 
 			lcl, err2 := localcachelayer.NewLocalCacheLayer(
 				retrylayer.New(s.sqlStore),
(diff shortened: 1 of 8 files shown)


isacikgoz (approved):
Nice one 👍

store/sqlstore/store.go
```diff
@@ -321,10 +326,17 @@ func (ss *SqlStore) GetCurrentSchemaVersion() string {
...
+// that can be parsed by callers.
+func (ss *SqlStore) GetDbVersion(numerical bool) (string, error) {
 	var sqlVersion string
 	if ss.DriverName() == model.DATABASE_DRIVER_POSTGRES {
-		sqlVersion = `SHOW server_version`
+		if numerical {
+			sqlVersion = `SHOW server_version_num`
+		} else {
+			sqlVersion = `SHOW server_version`
+		}
```
streamer45, line 339: I am wondering if a separate method that directly returns an `int` would be more appropriate. Alternatively, we could return three values `(string, int, error)` without the need for a parameter. Maybe we could use the same convention (2 digits per version) for other db types and provide a conversion accordingly? (this could be done as a separate, even help-wanted ticket).

app/server.go
```diff
@@ -319,6 +320,19 @@ func NewServer(options ...Option) (*Server, error) {
...
+				ver, err2 := s.sqlStore.GetDbVersion(true)
+				if err2 != nil {
+					return nil, errors.Wrap(err2, "cannot get DB version")
+				}
+				intVer, err2 := strconv.Atoi(ver)
+				if err2 != nil {
+					return nil, errors.Wrap(err2, "cannot parse DB version")
+				}
+				if intVer < sqlstore.MINIMUM_REQUIRED_POSTGRES_VERSION {
+					return nil, fmt.Errorf("minimum required postgres version is 10.0; found %s", ver)
```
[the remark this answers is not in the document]
agnivade: Yep, makes complete sense.

store/sqlstore/store.go
agnivade: I had thought about this one. I didn't go for a separate method because this was just for postgres. But now I checked MySQL and it directly returns a single version string for `select version()`. I am not terribly against doing this in this PR itself, but an HW ticket for a separate method along with the parsing and converting to int should also be okay. Since this is not called anywhere else in the codebase except for telemetry.

Let me know.

streamer45: Yes, I like the HW route.

streamer45 (approved):
Awesome :+1:

amyblais:
Does this have an expected fix version / milestone or this goes to v5.32?

agnivade:
5.32 is fine. Thanks @amyblais

PR #16604, opened 2020-12-24
Send user updated websocket event on plugin.UpdateUser

chetanyakan:
#### Summary

- Sends a WebSocket event when `plugin.UpdateUser` is called so it is updated immediately for all the users.

#### Release Note

```release-note
- UserUpdate WebSocket Event is now broadcast by 2 more APIs - plugin.UpdateUser and ConvertBotToUser
- Change in a User's NotifyProps during the user import process will not send an email notification. This is done to make it consistent with other parts of the import process where change in user's attributes would also not send any notifications.
```

### DIFF
--- a/app/app_iface.go
+++ b/app/app_iface.go
@@ -1018,7 +1018,7 @@ type AppIface interface {
 	UpdateUserActive(userId string, active bool) *model.AppError
 	UpdateUserAsUser(user *model.User, asAdmin bool) (*model.User, *model.AppError)
 	UpdateUserAuth(userId string, userAuth *model.UserAuth) (*model.UserAuth, *model.AppError)
-	UpdateUserNotifyProps(userId string, props map[string]string) (*model.User, *model.AppError)
+	UpdateUserNotifyProps(userId string, props map[string]string, sendNotifications bool) (*model.User, *model.AppError)
 	UpdateUserRoles(userId string, newRoles string, sendWebSocketEvent bool) (*model.User, *model.AppError)
 	UploadData(us *model.UploadSession, rd io.Reader) (*model.FileInfo, *model.AppError)
 	UploadEmojiImage(id string, imageData *multipart.FileHeader) *model.AppError
--- a/app/import_functions.go
+++ b/app/import_functions.go
@@ -493,7 +493,7 @@ func (a *App) importUser(data *UserImportData, dryRun bool) *model.AppError {
 			}
 		}
 		if hasNotifyPropsChanged {
-			if savedUser, err = a.UpdateUserNotifyProps(user.Id, user.NotifyProps); err != nil {
+			if savedUser, err = a.UpdateUserNotifyProps(user.Id, user.NotifyProps, false); err != nil {
 				return err
 			}
 		}
--- a/app/opentracing/opentracing_layer.go
+++ b/app/opentracing/opentracing_layer.go
@@ -15622,7 +15622,7 @@ func (a *OpenTracingAppLayer) UpdateUserAuth(userId string, userAuth *model.User
 	return resultVar0, resultVar1
 }
 
-func (a *OpenTracingAppLayer) UpdateUserNotifyProps(userId string, props map[string]string) (*model.User, *model.AppError) {
+func (a *OpenTracingAppLayer) UpdateUserNotifyProps(userId string, props map[string]string, sendNotifications bool) (*model.User, *model.AppError) {
 	origCtx := a.ctx
 	span, newCtx := tracing.StartSpanWithParentByContext(a.ctx, "app.UpdateUserNotifyProps")
 
@@ -15634,7 +15634,7 @@ func (a *OpenTracingAppLayer) UpdateUserNotifyProps(userId string, props map[str
 	}()
 
 	defer span.Finish()
-	resultVar0, resultVar1 := a.app.UpdateUserNotifyProps(userId, props)
+	resultVar0, resultVar1 := a.app.UpdateUserNotifyProps(userId, props, sendNotifications)
 
 	if resultVar1 != nil {
 		span.LogFields(spanlog.Error(resultVar1))
--- a/app/user.go
+++ b/app/user.go
@@ -1174,8 +1174,6 @@ func (a *App) UpdateUserAsUser(user *model.User, asAdmin bool) (*model.User, *mo
 		return nil, err
 	}
 
-	a.sendUpdatedUserEvent(*updatedUser)
-
 	return updatedUser, nil
 }
 
@@ -1191,8 +1189,6 @@ func (a *App) PatchUser(userId string, patch *model.UserPatch, asAdmin bool) (*m
 	if err != nil {
 		return nil, err
 	}
-
-	a.sendUpdatedUserEvent(*updatedUser)
 
 	return updatedUser, nil
 }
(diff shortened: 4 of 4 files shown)


app/plugin_api.go
```diff
@@ -279,7 +279,14 @@ func (api *PluginAPI) DeletePreferencesForUser(userId string, preferences []mode
 }
 
 func (api *PluginAPI) UpdateUser(user *model.User) (*model.User, *model.AppError) {
-	return api.app.UpdateUser(user, true)
+	updatedUser, err := api.app.UpdateUser(user, true)
+	if err != nil {
+		return nil, err
+	}
+
+	api.app.sendUpdatedUserEvent(*updatedUser)
```
agnivade, line 287: We should move this inside the `UpdateUser` method itself. There are 2 places where `sendUpdatedUserEvent` is called after UpdateUser. We should remove those as well.

agnivade (changes requested):
Thanks @chetanyakan. Please take a look at the comments.

app/plugin_api.go
chetanyakan: @agnivade This will send the UpdatedUser WebSocket event in other places a well like - 
`[UpdateUserNotifyProps](https://github.com/mattermost/mattermost-server/blob/c1dd23a3c80365c4ca4e2721e55c0c2c90038b25/app/user.go#L1348)` and `[ConvertBotToUser](https://github.com/mattermost/mattermost-server/blob/c1dd23a3c80365c4ca4e2721e55c0c2c90038b25/app/user.go#L2334)`

Would that be an acceptable change?

agnivade: Yes, I think that should be fine. Thanks.

chetanyakan: @agnivade done

agnivade:
Woops, @chetanyakan - please update the PR with the latest master. Thanks!

agnivade:
Also, let's flesh out the release notes properly by specifically mentioning which 2 API requests will now broadcast the user update messages via websocket.

chetanyakan:
Done @agnivade

agnivade (approved):

agarciamontoro (approved):
Thank you for your contribution, @chetanyakan :tada:

(and sorry for the late review, I was out on vacation!)

lindy65:
Hi @agnivade / @agarciamontoro - is there a Jira ticket for this PR? Please link me to it? Thanks!

agarciamontoro:
Hm, that's a good question I don't an answer for, I had not noticed when I reviewed it. Maybe @agnivade or @chetanyakan know.

agnivade:
There is no JIRA ticket @lindy65. This is something @chetanyakan created on his own. I am not sure of the QA process we follow for contributor generated PRs.

lindy65:
Hi @agnivade - would you help create a Jira ticket please?  This is required so that once the PR is merged, we don't lose track of it and have something to test against from resolved tickets.  Please also help add test steps to the ticket? TIA :)

agnivade (changes requested):
Actually, @chetanyakan - apologies for asking you to do this again, but on a closer look, I realized that we don't want to do this for UpdateNotifyProps. That's because it happens during the user import process and there is no need to send any notifications.

It should not be too much work. The method is just used in 2 places. We need to add another flag to `UpdateUserNotifyProps` for `sendNotifications` and pass false from app/import_functions.go.

And only send the user event if `sendNotifications` is set to true.

Everything else remains same. Sorry for the double work here.

agnivade:
@lindy65 - https://mattermost.atlassian.net/browse/MM-31915 is the ticket, but please hold off from testing for now. The PR needs to be updated a bit.

lindy65:
Thanks @agnivade - I'll hold off testing until you give me the go-ahead.  Can you tell me what plugin updates a user profile so I can test using it?

agnivade:
There are many like that but probably @agarciamontoro knows a suitable one to test this.

agarciamontoro:
I cannot think of one right now, actually :thinking: I checked that the Github plugin does call UpdateUser when disconnecting the Github account: https://github.com/mattermost/mattermost-plugin-github/blob/8f2addacdcf24636d697fc2dc015cbadbcbb2c8a/server/plugin/plugin.go#L331, but I'm not sure if that's very userful for your testing purposes, @lindy65.

lindy65:
Thanks @agarciamontoro :)  Perhaps I should test this after merge where it'll be easier with available plugins on `master`

agnivade:
Hey @chetanyakan - how is it going with this one? Let us know if you need any help.

chetanyakan:
Sorry, @agnivade, I have been really busy. Will update this shortly.

chetanyakan:
> Actually, @chetanyakan - apologies for asking you to do this again, but on a closer look, I realized that we don't want to do this for UpdateNotifyProps. That's because it happens during the user import process and there is no need to send any notifications.
> 
> It should not be too much work. The method is just used in 2 places. We need to add another flag to `UpdateUserNotifyProps` for `sendNotifications` and pass false from app/import_functions.go.
> 
> And only send the user event if `sendNotifications` is set to true.
> 
> Everything else remains same. Sorry for the double work here.

Hey @agnivade just to clarify, I will be making the following changes:

1. Update the `app.UpdateUser` method to call `a.sendUpdatedUserEvent(*user)` only when the value of the second parameter `sendNotifications` is true.
2. Update the method `app.UpdateUserNotifyProps` to add an extra parameter `sendNotifications` and pass that to `UpdateUser`.
3. Update the usages of `UpdateUserNotifyProps` method to pass the values of  `sendNotifications` and pass `false` in `app/import_functions.go`.

Is that correct?

agnivade:
That is correct! A small note there is that `app/import_functions.go` is the only place that calls `UpdateUserNotifyProps`.

I should add here for posterity that this change will make `UpdateUserNotifyProps` not send mail notifications from import_functions.go. But there are other areas, where `UpdateUser` / `UpdateUserRoles` are called during the import process with notifications disabled. And it sounds like the correct process to me. So we would be actually making the behavior more consistent.

And for full transparency, we should add this in release notes too, that change in a user's notify props during the import process will not send a mail notification. This is done to make it consistent with changes in other attributes of the user.

agnivade:
Are the release notes still correct? `UpdateUserNotifyProps` won't broadcast an event any more correct?

/update-branch

agnivade:
Woops, please update the PR with the latest master.

chetanyakan:
@agnivade I have updated the PR to the head of `master` and updated the release notes.

Your note was correct, `UpdateUserNotifyProps` won't broadcast the user updated event.

agnivade (approved):
Nice work @chetanyakan !

agnivade:
@lindy65 - over to you :)

lindy65:
> I cannot think of one right now, actually 🤔 I checked that the Github plugin does call UpdateUser when disconnecting the Github account: https://github.com/mattermost/mattermost-plugin-github/blob/8f2addacdcf24636d697fc2dc015cbadbcbb2c8a/server/plugin/plugin.go#L331, but I'm not sure if that's very userful for your testing purposes, @lindy65.

Hey @agnivade - would this PR be better tested after merge based on what `@agarciamontoro` mentioned above?  I'm not sure which plugin tests this fix properly but there is more choice on `community-daily`? Thoughts?

agnivade:
I agree.

lindy65 (approved):
This will be tested on `community-daily` after merge 👍

PR #16606, opened 2020-12-24
[MM-31289] Add some metrics for remote cluster service

streamer45:
#### Summary

PR implements some Prometheus metrics for the remote cluster service.

#### Ticket Link

https://mattermost.atlassian.net/browse/MM-31289

#### Release Note

```release-note
none
```

#### Related PRs

https://github.com/mattermost/enterprise/pull/885

### DIFF
--- a/app/server.go
+++ b/app/server.go
@@ -1177,7 +1177,9 @@ func (s *Server) Start() error {
 		}
 	}
 
-	s.startInterClusterServices(s.License())
+	if license := s.License(); license != nil {
+		s.startInterClusterServices(license)
+	}
 
 	return nil
 }
--- a/app/server.go
+++ b/app/server.go
@@ -1658,3 +1658,9 @@ func (s *Server) GetStore() store.Store {
 func (s *Server) GetRemoteClusterService() *remotecluster.Service {
 	return s.remoteClusterService
 }
+
+// GetMetrics returns the server's Metrics interface. Exposing via a method
+// allows interfaces to be created with subsets of server APIs.
+func (s *Server) GetMetrics() einterfaces.MetricsInterface {
+	return s.Metrics
+}
--- a/i18n/en.json
+++ b/i18n/en.json
@@ -1207,10 +1207,6 @@
     "translation": "Unable to find remote id {{.RemoteId}}."
   },
   {
-    "id": "api.context.remote_token_invalid.app_error",
-    "translation": "Invalid remote token."
-  },
-  {
     "id": "api.context.server_busy.app_error",
     "translation": "Server is busy, non-critical services are temporarily unavailable."
   },
@@ -1957,10 +1953,6 @@
   {
     "id": "api.remote_cluster.invalid_token.app_error",
     "translation": "Invalid token."
-  },
-  {
-    "id": "api.remote_cluster.invalid_topic.app_error",
-    "translation": "Invalid topic."
   },
   {
     "id": "api.remote_cluster.save.app_error",
--- a/i18n/en.json
+++ b/i18n/en.json
@@ -1955,6 +1955,10 @@
     "translation": "Invalid token."
   },
   {
+    "id": "api.remote_cluster.invalid_topic.app_error",
+    "translation": "Invalid topic."
+  },
+  {
     "id": "api.remote_cluster.save.app_error",
     "translation": "We encountered an error saving the remote cluster."
   },
--- a/model/remote_cluster.go
+++ b/model/remote_cluster.go
@@ -149,10 +149,6 @@ func (m *RemoteClusterMsg) IsValid() *AppError {
 		return NewAppError("RemoteClusterMsg.IsValid", "api.remote_cluster.invalid_id.app_error", nil, "Id="+m.Id, http.StatusBadRequest)
 	}
 
-	if m.Topic == "" {
-		return NewAppError("RemoteClusterMsg.IsValid", "api.remote_cluster.invalid_topic.app_error", nil, "Topic empty", http.StatusBadRequest)
-	}
-
 	if !IsValidId(m.Token) {
 		return NewAppError("RemoteClusterMsg.IsValid", "api.remote_cluster.invalid_token.app_error", nil, "", http.StatusBadRequest)
 	}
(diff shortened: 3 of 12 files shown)


model/remote_cluster.go
```diff
@@ -149,10 +149,6 @@ func (m *RemoteClusterMsg) IsValid() *AppError {
 		return NewAppError("RemoteClusterMsg.IsValid", "api.remote_cluster.invalid_id.app_error", nil, "Id="+m.Id, http.StatusBadRequest)
 	}
 
-	if m.Topic == "" {
-		return NewAppError("RemoteClusterMsg.IsValid", "api.remote_cluster.invalid_topic.app_error", nil, "Topic empty", http.StatusBadRequest)
-	}
-
```
streamer45, line 155: I've temporarily removed this since it was causing failures. We actually send and receive some messages without topic (ping and invitation).
Let me know how you'd like to handle these cases.

services/remotecluster/invitation.go
```diff
@@ -52,6 +52,7 @@ func makeConfirmFrame(rc *model.RemoteCluster, siteURL string) (*model.RemoteClu
 	msg := model.RemoteClusterMsg{
 		Id:       model.NewId(),
 		CreateAt: model.GetMillis(),
+		Token:    rc.RemoteToken,
```
streamer45, line 55: Here I had to include the Token since it was again causing some more failures during validation. 
Do we need the redundant token in both frame and message structs?

services/remotecluster/ping.go
```diff
@@ -97,6 +97,12 @@ func (rcs *Service) pingRemote(rc *model.RemoteCluster) error {
 		return err
 	}
 
+	if metrics := rcs.server.GetMetrics(); metrics != nil {
+		sentAt := time.Unix(0, ping.SentAt*int64(time.Millisecond))
+		elapsed := time.Since(sentAt).Seconds()
+		metrics.ObserveRemoteClusterPingDuration(rc.RemoteId, elapsed)
+	}
```
streamer45, line 104: Here we are storing the actual round-trip time for the ping message which doesn't suffer from clock skew.
If you want, we can add an additional metric for `ping.RecvAt - ping.SentAt`.

services/remotecluster/ping.go
```diff
@@ -118,6 +124,7 @@ func makePingFrame(rc *model.RemoteCluster) (*model.RemoteClusterFrame, error) {
 	msg := model.RemoteClusterMsg{
 		Id:       model.NewId(),
 		CreateAt: model.GetMillis(),
+		Token:    rc.RemoteToken,
```
streamer45, line 127: Same thing as above. Added the redundant token to avoid failure.

services/remotecluster/send.go
```diff
@@ -121,6 +122,13 @@ func (rcs *Service) sendFrameToRemote(timeout time.Duration, frame *model.Remote
 	req.Header.Set("Content-Type", "application/json")
 
 	resp, err := rcs.httpClient.Do(req.WithContext(ctx))
+	if metrics := rcs.server.GetMetrics(); metrics != nil {
+		if err != nil || resp.StatusCode != http.StatusOK {
+			metrics.IncrementRemoteClusterMsgErrorsCounter(frame.RemoteId, os.IsTimeout(err))
+		} else {
+			metrics.IncrementRemoteClusterMsgSentCounter(frame.RemoteId)
+		}
+	}
```
streamer45, line 131: Here I've opted for incrementing counters at the frame level. This will also include pings. Let me know if you prefer something more granular (e.g. differentiate between message types).

model/remote_cluster.go
wiggin77: I have a PR coming up that includes a topic for pings and invitations.  I'll add this check back when that gets submitted.

services/remotecluster/ping.go
wiggin77: I included the extra fields in the messaging in case we needed to detect clock skew.  We know how long the round-trip was, so the difference between that and RecAt-SentAt should be clock skew.

If the clocks are way off, then we'll get messages in different orders at each site.  If you could add a clock-skew metric that would be great.  If not we can add it later.

services/remotecluster/invitation.go
wiggin77: Hmm... sorry about that.  Not sure how the feature branch ended up like that.  I'll remove the token from msg when I merge next PR.  FYI, a message can be sent to multiple sites and the tokens are site specific.

services/remotecluster/send.go
wiggin77: Frame level makes sense. :+1:

i18n/en.json
```diff
@@ -1206,10 +1206,6 @@
     "id": "api.context.remote_id_invalid.app_error",
     "translation": "Unable to find remote id {{.RemoteId}}."
   },
-  {
```
wiggin77, line 1209: These strings are still needed.

wiggin77 (changes requested):
Looks great @streamer45 .  Just need the strings added back to en.json.  :+1:  Thanks!

services/remotecluster/ping.go
agnivade: To clarify, we would be assuming that each side of the roundtrip takes the same time right? Currently doing some research on this, and one way that's suggested in the literature is to query the NTP servers themselves to know the clock skew.

streamer45: Yes, there are certainly more reliable ways to get the actual clock skew. Here we would be doing an approximation with the data we have:

```
RT/2 - (RecvAt-SentAt)
```

app/server.go
```diff
@@ -1177,7 +1177,9 @@ func (s *Server) Start() error {
 		}
 	}
 
-	s.startInterClusterServices(s.License())
+	if license := s.License(); license != nil {
+		s.startInterClusterServices(license)
+	}
```
streamer45, line 1182: Added this since it was causing a panic on an unlicensed server.

wiggin77 (approved):
Looks great.  Thanks!  :+1:

isacikgoz (approved):
LGTM 🎉

PR #16608, opened 2020-12-28
MM-31063: Change constants to use CamelCase

agnivade:


agnivade:
@isacikgoz - Need some help from you in finishing this for the rest of the packages :stuck_out_tongue_closed_eyes: Here are some commands which might be of help. We need to do this for all packages except model. And add an exclude directive to skip the model directory for golint.

```
golangci-lint run --max-same-issues 1000 <all_directories_except_model> > ALLCAPS
grep " = " ALLCAPS | cut -f 1 -d '=' | tr -d ' \t' | sort | uniq > words.txt
```

And then 
```
#!/bin/bash

while IFS= read -r line
do
  src=`echo "$line" | cut -f 1 -d ' '`
  dst=`echo "$line" | cut -f 2 -d ' '`
  git grep -rl $src . | xargs sed -i "s/$src/$dst/g"
done < words.txt
```

`go fmt ./...`

agnivade:
@isacikgoz - I completed the store package too, and changed the golangci file to ignore model package.

isacikgoz (approved):
Awesome 🎉

PR #16610, opened 2020-12-28
[MM-31555] Print golang version on server startup

Ayanrocks:
#### Summary
Add logging to Print Golang Version before server

#### Ticket Link
Fixes: https://github.com/mattermost/mattermost-server/issues/16609
JIRA: https://mattermost.atlassian.net/browse/MM-31555

#### Release Note

```release-note
NONE
```

### DIFF
--- a/app/server.go
+++ b/app/server.go
@@ -224,7 +224,7 @@ func NewServer(options ...Option) (*Server, error) {
 	}
 
 	// This is called after initLogging() to avoid a race condition.
-	mlog.Info("Server is initializing...", mlog.String(" Go version", runtime.Version()))
+	mlog.Info("Server is initializing...", mlog.String("go_version", runtime.Version()))
 
 	// It is important to initialize the hub only after the global logger is set
 	// to avoid race conditions while logging from inside the hub.
--- a/app/server.go
+++ b/app/server.go
@@ -224,7 +224,7 @@ func NewServer(options ...Option) (*Server, error) {
 	}
 
 	// This is called after initLogging() to avoid a race condition.
-	mlog.Info("Server is initializing with Go version " + runtime.Version())
+	mlog.Info("Server is initializing...", mlog.String(" Go version", runtime.Version()))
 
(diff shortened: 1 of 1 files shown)


app/server.go
```diff
@@ -223,6 +223,8 @@ func NewServer(options ...Option) (*Server, error) {
 		mlog.Error(err.Error())
 	}
 
+	mlog.Info("Go Version Detected: " + runtime.Version())
+
 	// This is called after initLogging() to avoid a race condition.
 	mlog.Info("Server is initializing...")
```
agnivade, line 229: Maybe add the info as a field to this line. It can be thought of as that the server is initializing with certain parameters. And we can just say "Go version".

Ayanrocks: Fixed it. now it says 
```go
Server is initializing with Go version 
```

agnivade: Sorry I meant something like `mlog.Info("Server is initializing...", mlog.String("Go version", runtime.Version()))`. It is better suited for log processing tools.

Ayanrocks: Done

agnivade (approved):

app/server.go
```diff
@@ -224,7 +224,7 @@ func NewServer(options ...Option) (*Server, error) {
 	}
 
 	// This is called after initLogging() to avoid a race condition.
-	mlog.Info("Server is initializing...")
+	mlog.Info("Server is initializing...", mlog.String(" Go version", runtime.Version()))
```
streamer45, line 227:
```suggestion
	mlog.Info("Server is initializing...", mlog.String("go_version", runtime.Version()))
```

This would be more consistent with how we use the mlog fields. First string is similar to a variable name and then the actual value.

streamer45 (commented):
Thank you @Ayanrocks :+1:  Just left a small suggestion.

streamer45 (approved):

agnivade:
No testing needed for this as it's just a log line change.

PR #16618, opened 2021-01-02
MM-24530-Add support for search terms with underscore using postgresql engine

### LINKED ISSUE #14436: Search terms with underscore using PostgreSQL engine should work
Right we’re using *to_tsquery* in the Postgresql search but when the word has underscore is returning more results than it should.

The problem is that `to_tsquery` for a word with an underscore is splitting that word into multiple parts:

```sql
mattermost_test=# select to_tsquery('english', '("test_with_underscore")');
      to_tsquery      
----------------------
 'test' & 'underscor'
(1 row)
```

Using for example something like:

```sql
mattermost*test=# select $$test_with*underscore$$::tsquery;
        tsquery         
~~----------------------~~
 'test_with_underscore'
(1 row)
```

could work but we are losing the dictionary option (the ‘English’ part in to_tsquery)

----

New contributors, please see our [Developer's Guide](https://developers.mattermost.com/contribute/getting-started/).

JIRA: https://mattermost.atlassian.net/browse/MM-24530

antwigambrah:
#### Summary

Search terms with underscore using PostgreSQL engine

#### Ticket Link
 Fixes   #14436

#### Jira ticket
https://mattermost.atlassian.net/browse/MM-24530

#### Release Note

``` 
The DB search using Postgresql now supports searching for terms with underscore
```


antwigambrah:
> @antwigambrah: Adding the "do-not-merge/release-note-label-needed" label because no release-note block was detected, please follow our [release note process](https://github.com/mattermost/chewbacca#release-notes-process) to remove it.
/release-note-none

antwigambrah:
@ethervoid ready for initial review.

agnivade:
@antwigambrah - Looks like some tests are failing. Please take a look.

antwigambrah:
@agnivade Update provided

antwigambrah:
/release-note-none


> @antwigambrah: Adding the "do-not-merge/release-note-label-needed" label because no release-note block was detected, please follow our [release note process](https://github.com/mattermost/chewbacca#release-notes-process) to remove it.

store/sqlstore/post_store.go
```diff
@@ -1313,11 +1313,18 @@ func (s *SqlPostStore) search(teamId string, userId string, params *model.Search
...
 		if params.OrTerms {
 			queryParams["Terms"] = "(" + strings.Join(strings.Fields(terms), " | ") + ")" + excludeClause
+
 		} else {
 			queryParams["Terms"] = "(" + strings.Join(strings.Fields(terms), " & ") + ")" + excludeClause
 		}
+		searchClause := ""
+
+		if !params.OrTerms && strings.Contains(terms, "_") {
+			searchClause = fmt.Sprintf("AND to_tsvector('english', replace(%s, '_', '/')) @@ plainto_tsquery('english', replace(:Terms, '_', '/'))", searchType)
```
agnivade, line 1323: I am wondering what are the benefits that `plainto_tsquery` is giving us if we have to replace underscores with `/` anyways, which also works the same way for `to_tsquery`. And what if the search clause itself contains `/`? Wouldn't we need to escape that?

antwigambrah: it works in both options and if the search contains /  it returns the same results.  i believe checking if it contains underscores and the else statement is not needed if that's the case then. Let me run some test on it .

agnivade: `plainto_tsquery` is fine for handling the underscores, but it can't express boolean logic other than plain `AND`. I am wondering if we should just fall back on just casting to `::tsquery` and lose out on the language config part if a text contains underscores.

antwigambrah: Casting to ::tsquery works best if there will be only single search terms with underscores.In case of multiple terms with one having underscores it seems that is where the language config will be needed to cater for the stop words and text normalization.

agnivade: That is right. It's a tradeoff between expressing boolean logic versus text normalization. I was thinking that expressing the boolean logic is more important. But completely open to opinions on this one.

Wondering if @ethervoid or @mgdelacroix has any thoughts on this one.

ethervoid: First of all thank you very much @antwigambrah for taking care of this, really appreciated :)

I agree with @agnivade my idea on this was to make a pre-normalization of the terms so we can use boolean operations as we've been doing. Changing the way we're returning results would result in complaints from users.

The idea is to provide extra features instead of different ones

antwigambrah: @ethervoid Then i believe checking if it contains underscores and direct casting will be the best option if that's the case .

agnivade: Yes, I agree.

ethervoid: @antwigambrah yes that would be the option IMHO

antwigambrah: @agnivade @ethervoid Update provided

agnivade (commented):
Looking much better now! Just some small suggestions.

agnivade:
@antwigambrah - I believe the reason why the CI build is not able to run is because you are following your personal fork rather than the upstream Mattermost repo: https://support.circleci.com/hc/en-us/articles/360008097173-Why-aren-t-pull-requests-triggering-jobs-on-my-organization-. Can you check that and unfollow if that's the case?

antwigambrah:
@agnivade Update provided

agnivade (approved):
Nicely done!

ethervoid (approved):
Great work! Thank you very much for your work, this change will make a noticeable improvement in the search for the MM users :)

lindy65:
Hi @antwigambrah 

Thanks so much for your contribution!

While testing, I created a post `post_Underscore` and searched for `post_underscore` and no results were found - most probably because the search is case-sensitive? 

image

If I search for `post_Underscore` the post **is** returned in the search results

image

Are you able to make the search non-case-sensitive?

Also, I'm noticing a never-ending "loading" animation but this reproduces on `master` too so I doubt it's related to your PR. I filed a Jira ticket [here](https://mattermost.atlassian.net/browse/MM-32121) for this issue.

image


/cc @ethervoid

antwigambrah:
By default , lexemes are lowercased any time to_tsquery() and to_tsvector() is used which makes it case insensitive . But the feature requirement needs direct casting of ::tsquery and ::tsvector which preserves the term in its original form and makes it case sensitive . However there is a workaround to use lower() on them directly or create and index on the message field with lower() or use the citext type on it . Great you noticed it.  @ethervoid  @agnivade

lindy65:
Thanks for the feedback @antwigambrah :)

Let's wait for `ethervoid` and/or `agnivade` to respond and take it from there 👍

agnivade:
Using `lower()` works for me.

ethervoid:
Works for me too!

antwigambrah:
@ethervoid @agnivade The direct approach or the index approach ?

ethervoid:
I'll use it directly

lindy65:
Thanks all :)  @antwigambrah - let me know when you've made the updates and I'll retest 👍

antwigambrah:
@lindy65 You can test now

lindy65:
Thanks @antwigambrah - seems the test server is failing to update though...  Can you take a look?

//cc @agnivade

antwigambrah:
@lindy65 Does it have anything to do from my end ?

agnivade:
@lindy65 - The cloud team would be the best to reach out on this.

lindy65:
Thanks @antwigambrah @agnivade - seems the test server has built so I will test again 👍

lindy65:
Awesome @antwigambrah - the case sensitivity issue I saw before is fixed and search term containing underscore is returned in results regardless of case 👍

lindy65 (approved):
Thanks @antwigambrah @agnivade - looks good to me 👍

agnivade:
This PR has a bug where if the database contains messages containing `:`, the query will fail because `:` are meant to indicate integer positions in the [lexeme](https://www.postgresql.org/docs/current/datatype-textsearch.html#DATATYPE-TSVECTOR).

`$$` can be used to treat it as a literal string, but embedded quotes and backslashes must be doubled.

Reverting this for now, as any search for text containing underscores will fail on Postgres. We can revisit this again.

/cc @ethervoid @antwigambrah

PR #16620, opened 2021-01-03
MM-21012: Revamp websocket implementation

agnivade:
We replace the old gorilla/websocket implementation with the
gobwas/ws library. The gorilla library was in maintenance mode
and had a high level API due to which we cannot use that for
situations where a large number of concurrent connections needs
to be supported.

The ws library is a very low-level library that allows us
to work with raw net.Conns. We make several improvements:

- We completely remove the reader goroutines, and instead
replace them with a manual epoll implementation which sends off
messages to be read when it receives any data on the connection.
This lets us scale to a much larger number of connections.
- The reader buffer is eliminated, because we directly read
from the connection now.

https://mattermost.atlassian.net/browse/MM-21012

```release-notes
Improved the websocket implementation by using epoll to manually
read from a websocket connection. As a result, the number of goroutines
is expected to go down by half.

This implementation is only available on Linux and FreeBSD based distributions.
```


agnivade:
There are 2 things pending:
- Implement max read limit. (This needs to be done in ws package itself. I'll look into it).
- Goroutine throttling.

Just opening it up for review so that we can iterate on the things that are already done.

api4/websocket.go
```diff
@@ -16,23 +20,47 @@ func (api *API) InitWebSocket() {
...
+		// on goroutine creation.
+		go func() {
+			err := wc.ReadMsg()
+			if err != nil {
+				mlog.Debug("Error while reading message from websocket", mlog.Err(err))
+				c.App.Srv().Poller().Stop(desc)
+				wc.Close()
+			}
+		}()
+	})
```
streamer45, line 65: Should this be part of `wc.Pump()`? It feels strange to have such a low level logic in the api package.

api4/websocket.go
```diff
@@ -5,9 +5,13 @@ package api4
...
 import (
 	"net/http"
+	"time"
 
-	"github.com/gorilla/websocket"
+	"github.com/mattermost/mattermost-server/v5/mlog"
 	"github.com/mattermost/mattermost-server/v5/model"
+
+	"github.com/gobwas/ws"
+	"github.com/mailru/easygo/netpoll"
```
streamer45, line 14: Given this is a change touching a very delicate part of our backend I have to ask:

What's your general opinion on these dependencies? Are they used in production elsewhere? Are they considered stable? Should we do a quick review before merging or do you think it's enough to test it internally (plus possibly a few long running load-tests)?

api4/websocket.go
```diff
@@ -16,23 +20,47 @@ func (api *API) InitWebSocket() {
...
-		WriteBufferSize: model.SOCKET_MAX_MESSAGE_SIZE_KB,
-		CheckOrigin:     c.App.OriginChecker(),
+	fn := c.App.OriginChecker()
+	if fn != nil && !fn(r) {
+		c.Err = model.NewAppError("origin_check", "api.web_socket.connect.check_origin.app_error", nil, "", http.StatusBadRequest)
+		return
+	}
+
+	upgrader := ws.HTTPUpgrader{
+		Timeout: 5 * time.Second,
```
streamer45, line 30: This looks like an addition. Just wondering if we actually need to set this.

app/web_conn.go
```diff
@@ -122,43 +128,45 @@ func (wc *WebConn) SetSession(v *model.Session) {
...
-
-	for {
-		var req model.WebSocketRequest
-		if err := wc.WebSocket.ReadJSON(&req); err != nil {
-			wc.logSocketErr("websocket.read", err)
-			return
+	default:
+		// Default case of data message.
+		if err := decoder.Decode(&req); err != nil {
+			return errors.Wrap(err, "error during decoding websocket message")
```
streamer45, line 164: Are there cases where we would need to call [`r.Discard()`](https://godoc.org/github.com/gobwas/ws/wsutil#Reader.Discard)? In other words, can the decoder return error having failed to read the whole message content?

streamer45 (commented):
Very nice :+1: I've gave this a first look and left some comments, mainly questions.

api4/websocket.go
agnivade: Yeah, the gorilla implementation doesn't actually implement any timeout while writing to a network connection. So it's good to have this.

app/web_conn.go
agnivade: Good point. :+1:

api4/websocket.go
agnivade: Yeah, I'll see what I can do.

api4/websocket.go
agnivade: Yes, I think we should do a review of the package. It's essentially an epoll API. The code is fairly self-contained here: https://github.com/mailru/easygo/blob/3c14a0dc985f4cd7899588176a176809055d2b3b/netpoll/epoll.go.

I don't have hands-on experience with implementing event loops myself. So it would be good to have multiple pair of eyes go through it :)

The only catch that I just realized now is that we would be losing Windows support. I don't know how strongly we supported native Windows. But since it doesn't have kqueue/epoll, this will straightaway throw an error.

api4/websocket.go
streamer45: Interesting. I thought this was only about the handshake process and that it essentially matched [`HandshakeTimeout`](https://pkg.go.dev/github.com/gorilla/websocket#Upgrader) which we never used, hence my question.

agnivade: Oh no you are right. That timeout does exactly that. I must have looked at somewhere else. It's just that we weren't using it.

app/web_conn.go
agnivade: Done

api4/websocket.go
agnivade: Done

api4/websocket.go
agnivade: I took a look. Looks simple enough where you just add/remove fds to an epoll instance. The core event loop logic is here: https://github.com/mailru/easygo/blob/3c14a0dc985f4cd7899588176a176809055d2b3b/netpoll/epoll.go#L245-L279. Let me know if you see any issues with it.

agnivade:
@streamer45 / @nronas - This is good to be reviewed now. All concerns are addressed.

I am wondering if this needs to go in a feature flag? IMO with sufficient QA and load-testing, it doesn't need to be. And it's also a bit cumbersome to maintain two implementations side-by-side. I think we can just revert if needed and fix issues, rather than going through the pain of having two implementations, and then remove the old one again.

Any thoughts?

app/web_conn.go
```diff
@@ -121,45 +131,95 @@ func (wc *WebConn) SetSession(v *model.Session) {
...
+	desc := netpoll.Must(netpoll.HandleRead(wc.WebSocket))
+	wc.App.Srv().Poller().Start(desc, func(wsEv netpoll.Event) {
+		if wsEv&(netpoll.EventReadHup|netpoll.EventHup) != 0 {
+			wc.App.Srv().Poller().Stop(desc)
+			wc.Close()
+			return
+		}
+
+		// Block until we have a token.
+		wc.App.Srv().GetWebConnToken()
```
streamer45, line 155: What happens if this returns when the connection/poller has been closed/stopped already?

api4/websocket.go
streamer45: Alright, getting back to Windows support (yeah), should we include a PM in the decision?

app/server.go
```diff
@@ -26,6 +26,7 @@ import (
 	"github.com/getsentry/sentry-go"
 	sentryhttp "github.com/getsentry/sentry-go/http"
 	"github.com/gorilla/mux"
+	"github.com/mailru/easygo/netpoll"
```
nronas, line 29: This means that we are dropping support for Windows NT right, don't know if we ever support it, but just highlighting this

app/web_conn.go
```diff
@@ -121,45 +131,95 @@ func (wc *WebConn) SetSession(v *model.Session) {
...
-	for {
-		var req model.WebSocketRequest
-		if err := wc.WebSocket.ReadJSON(&req); err != nil {
-			wc.logSocketErr("websocket.read", err)
-			return
+	default:
+		// Default case of data message.
+		if err := decoder.Decode(&req); err != nil {
+			// We discard any remaining data left in the socket.
+			r.Discard()
```
nronas, line 216: Do we need to do that? Is it not handled already in the receiver? since the receiver of this when it gets an error closes the poller all together.

nronas: Ignore me, i see that this is the socket itself.

nronas:
> I am wondering if this needs to go in a feature flag? IMO with sufficient QA and load-testing, it doesn't need to be. And it's also a bit cumbersome to maintain two implementations side-by-side. I think we can just revert if needed and fix issues, rather than going through the pain of having two implementations, and then remove the old one again.

I would say that it would be super hard to maintain both versions in parallel. Based on my understanding they are not compatible with each other(meaning you cannot disable and enable one for the other without a server restart right?)

api4/websocket.go
agnivade: Yeah, let me talk to Katie. I do see windows builds being part of https://latest.mattermost.com/. So there might be some pushback :confounded: Let's see.

app/server.go
agnivade: We do actually. I never realized. I will talk with PM.

agnivade:
> I would say that it would be super hard to maintain both versions in parallel. Based on my understanding they are not compatible with each other(meaning you cannot disable and enable one for the other without a server restart right?)

Yeah, it's hard to run both side by side. You technically can do away without restarting, but it's going to be hard.

app/web_conn.go
agnivade: The poller is only stopped in 2 places: L149 and L162 in this file. So if the connection is closed, `ReadMsg()` will fail, and then we stop poller and close connection, and release the token on our way out.

If it happens earlier, then the token is not taken in the first place.

app/server.go
```diff
@@ -98,8 +99,11 @@ type Server struct {
 
 	EmailService *EmailService
 
-	hubs     []*Hub
-	hashSeed maphash.Seed
+	hubs          []*Hub
+	hashSeed      maphash.Seed
+	poller        netpoll.Poller
+	webConnSema   chan struct{}
+	webConnSemaWg *sync.WaitGroup
```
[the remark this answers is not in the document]
agnivade: I was a bit concerned to prevent inadvertently copying the waitgroup. But the server methods are pointer-based. So it should be fine.

streamer45:
> > I would say that it would be super hard to maintain both versions in parallel. Based on my understanding they are not compatible with each other(meaning you cannot disable and enable one for the other without a server restart right?)
> 
> Yeah, it's hard to run both side by side. You technically can do away without restarting, but it's going to be hard.

We might actually have to do that in the long term if we want to keep Windows support :grimacing:

agnivade:
Actually, thinking a bit more, we only need to keep the readPump running for Windows support. So not _that_ big of a deal. Just need some more refactoring, and if conditions. But I will wait first and see how well my "drop Windows" proposal goes :P

agnivade:
Aaand boom. Windows support!

nronas:
> Aaand boom. Windows support!

I can't see where I can download the CD-ROM with the latest update? 😄

nronas (approved):
LGTM 🚢 - it

app/web_conn.go
```diff
@@ -140,6 +149,23 @@ func (wc *WebConn) Pump() {
 	close(wc.pumpFinished)
 }
 
+// BlockingPump is the Windows alternative of Pump.
+// It creates two goroutines - one for reading, another
+// for writing.
```
streamer45, line 154: So we are creating one goroutine for writing and using the current one for reading which essentially blocks until disconnection.

streamer45 (approved):
Looks good :+1: It would be nice to see some load-tests now.

app/web_conn.go
agnivade: Yeah, this is the current behavior. No change from that.

agnivade:
> It would be nice to see some load-tests now.

Where is the label when you need it? :laughing:

streamer45:
> > It would be nice to see some load-tests now.
> 
> Where is the label when you need it?

Don't get me started :stuck_out_tongue:

agnivade:
Load test done. Caught an edge case and fixed it! Apparently the netpoller can return a data event after a HUP event within 1ms apart.

And it looks like we have some nice memory improvements. There are some spikes in the store response times, I think that's because I doubled the rate of user addition in the websocket test. The original base test was using the replica lag configuration which used a slow ramp up. In any case, this code does not change any other part of the code, so spikes in store response times are unrelated.

websocket-connections
p99-api-times
avg-api-times
p99-store-times
avg-store-times
stack-in-use
heap-in-use
cpu-utilization-app

streamer45:
Love that stack in use graph :grin: :clap:

agnivade:
Yep, drastic change from O(n) to O(k).

agnivade:
@lindy65 - I wanted to get some testing done in a Windows environment for this PR. Since you use a Windows setup, will you be able to help with this? I can attach the exe file that you would need to run, or maybe even let the CI generate it for us.

lindy65:
Hi @agnivade - I don't have a local setup like the SDETs but if I'm able to test this on a spinwick on the PR then all is good :)

agnivade:
You can't test this on a spinwick. Are there any SDETs with a Windows setup?

lindy65:
I'll ask in our QA channel and post back here 👍

agnivade:
@lindy65 - Any word on this one? ;)

lindy65:
Hey @agnivade - 2 SDETs have responded saying they don't have Windows...  I'm checking with the other 2...  Alternatively, if no SDETs have Windows, is there anyone with Windows on the Server team who could "peer-review"?

lindy65:
Hey @agnivade - none of the SDETs have a Windows environment but @furqanmlk will chat to you about what's required for testing 👍

agnivade:
Hey @furqanmlk - how is it going with this one? :)

furqanmlk:
Hi @agnivade thank you so much for you help.
Finally I was able to verify in Windows :)




https://user-images.githubusercontent.com/6909890/107824042-c979a000-6d4e-11eb-8f9a-59de5c52251e.mp4




https://user-images.githubusercontent.com/6909890/107824292-3b51e980-6d4f-11eb-95aa-ca5594f0769c.mp4

furqanmlk (approved):

PR #16628, opened 2021-01-04
MM-31625: System user role should be assumed if nothing is specified

### LINKED ISSUE #16577: Slash command autocomplete data - system user role should be assumed if nothing is specified
#### Summary
Slash command autocomplete data is filtered based on the user's role and command's specified role. However, the default value of the empty string is not handled in Mattermost server code, thus causing such a slash command to never be displayed for a system user.

#### Steps to reproduce
1. Register a slash command with autocomplete data from a plugin. Don't specify anything for `RoleID`.
2. Install and enable the plugin.
3. Log in to Mattermost from a system user account (not system admin).
4. Try typing `/` in the message box to bring up the list of auto complete slash command suggestions. Notice that your slash command is not in the list.

#### Expected behavior
Slash command autocomplete data with no role specified should be assumed to be for the system user role. It should only be required to specify a role when the developer needs to restrict it.

Slash command autocomplete data with no role specified works only for system admins, not for system users.

#### Possible fixes

In the function [`getSuggestions()`](https://github.com/mattermost/mattermost-server/blob/277f886b3ffb483cc1f1d9f077672bc8d687efa2/app/command_autocomplete.go#L50) at this line - 

https://github.com/mattermost/mattermost-server/blob/277f886b3ffb483cc1f1d9f077672bc8d687efa2/app/command_autocomplete.go#L56

we check if `roleID == ""`, instead we should be checking for `command.RoleID == ""`. The current check of `roleID == ""` will never succeed because at this line -

https://github.com/mattermost/mattermost-server/blob/e974b7b9be5ec499b2e48a86038b6edbb1fa8bbd/api4/command.go#L380

we always set it to something, so it can never be empty string.

haardikdharma10:
#### Summary

This PR aims to fix #16577 

#### Ticket Link

Fixes: #16577 
JIRA: https://mattermost.atlassian.net/browse/MM-31625

#### Release Note

```release-note
NONE
```

### DIFF
--- a/app/command_autocomplete.go
+++ b/app/command_autocomplete.go
@@ -53,7 +53,7 @@ func (a *App) getSuggestions(commandArgs *model.CommandArgs, commands []*model.A
 
 	if index == -1 { // no space in input
 		for _, command := range commands {
-			if strings.HasPrefix(command.Trigger, strings.ToLower(inputToBeParsed)) && (command.RoleID == roleID || roleID == model.SYSTEM_ADMIN_ROLE_ID || command.RoleID == "") {
+			if strings.HasPrefix(command.Trigger, strings.ToLower(inputToBeParsed)) && (command.RoleID == roleID || roleID == model.SYSTEM_ADMIN_ROLE_ID) {
 				s := model.AutocompleteSuggestion{
 					Complete:    inputParsed + command.Trigger,
 					Suggestion:  command.Trigger,
--- a/app/command_autocomplete.go
+++ b/app/command_autocomplete.go
@@ -53,7 +53,7 @@ func (a *App) getSuggestions(commandArgs *model.CommandArgs, commands []*model.A
 
 	if index == -1 { // no space in input
 		for _, command := range commands {
-			if strings.HasPrefix(command.Trigger, strings.ToLower(inputToBeParsed)) && (command.RoleID == roleID || roleID == model.SYSTEM_ADMIN_ROLE_ID || roleID == "") {
+			if strings.HasPrefix(command.Trigger, strings.ToLower(inputToBeParsed)) && (command.RoleID == roleID || roleID == model.SYSTEM_ADMIN_ROLE_ID || command.RoleID == "") {
 				s := model.AutocompleteSuggestion{
 					Complete:    inputParsed + command.Trigger,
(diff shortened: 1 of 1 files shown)


agnivade (changes requested):
I think instead of this approach, we should remove the `roleID == ""` check and by default populate the role to be `model.SYSTEM_USER_ROLE_ID` while creating the slash command if it's not specified.

And also update the API documentation with it.

haardikdharma10:
> I think instead of this approach, we should remove the `roleID == ""` check and by default populate the role to be `model.SYSTEM_USER_ROLE_ID` while creating the slash command if it's not specified.
> 
> And also update the API documentation with it.

Hi @agnivade, Aren't we already doing that [here?](https://github.com/mattermost/mattermost-server/blob/e974b7b9be5ec499b2e48a86038b6edbb1fa8bbd/api4/command.go#L380-L383) Did I miss something?

agnivade:
That API is to list the autocomplete suggestions which sets the roleID of the user making the request. What I suggested was to set the allowed roleID of the slash command itself while creating it. That's being done here: https://github.com/mattermost/mattermost-server/blob/e89b26e8f306f27489d2fda946130ef8d16ef308/app/command.go#L623. Check for RoleID inside AutoCompleteData, and if it's not specified, set it to `SYSTEM_USER_ROLE_ID`.

jfrerich:
@haardikdharma10, do you need any further clarification.  BTW, thanks for the many PRs!

emilyacook:
> https://github.com/mattermost/mattermost-server/pull/16628#issuecomment-756522426
> That API is to list the autocomplete suggestions which sets the roleID of the user making the request. What I suggested was to set the allowed roleID of the slash command itself while creating it. That's being done here:
> 
> https://github.com/mattermost/mattermost-server/blob/e89b26e8f306f27489d2fda946130ef8d16ef308/app/command.go#L623
> 
> . Check for RoleID inside AutoCompleteData, and if it's not specified, set it to `SYSTEM_USER_ROLE_ID`.

@haardikdharma10 Please let us know if there's anything we can help you with on this :) thank you so much for the contribution!

haardikdharma10:
Sorry for the delay on this. I will close this for now.

PR #16637, opened 2021-01-05
Merge latest cloud into master

jwilander:


jwilander:
Because we didn't do a merge commit the last couple times (it's disabled in the UI) the old commits are showing up. I'll do a manual merge commit when the PR is approved to fix it for the future.

@marianunez @mgdelacroix if you could take a look at the `ea7468b` commit to make sure it looks good (it's all been tested) that'd be great

store/sqlstore/upgrade.go
```diff
@@ -940,31 +941,107 @@ func upgradeDatabaseToVersion529(sqlStore *SqlStore) {
...
+		)
+		return true
+	}
+
+	if sqlStore.DriverName() == model.DATABASE_DRIVER_POSTGRES {
+		if !sqlStore.IsVarchar(scIdInfo.DataType) || scIdInfo.CharMaximumLength != 300 {
+			return true
+		}
+	} else if sqlStore.DriverName() == model.DATABASE_DRIVER_MYSQL {
+		if scIdInfo.DataType != "text" {
```
marianunez, line 1037: Do we want to check the length in this case as well to be consistent?

marianunez (approved):
Checked migration checker [commit](https://github.com/mattermost/mattermost-server/commit/ea7468be52e550271d2dce3c3955bf0a6fecf5c4)

LGTM

store/sqlstore/upgrade.go
jwilander: There's no length for `text` columns in MySQL

marianunez: 🤦‍♀️

marianunez:
I believe Miguel is OOO this week, so we may want to swap out the reviewer with possibly someone else in platform?

jwilander:
@agnivade this is mainly a merge commit (please excuse the number of commits that was the result of not merge committing and squashing previously) but if you could give https://github.com/mattermost/mattermost-server/commit/ea7468be52e550271d2dce3c3955bf0a6fecf5c4 a review that would be great

jwilander:
There's an issue with the merge causing some tests to fail. I'm looking into it.

jwilander:
Tests should be fixed now.

agnivade:
@jwilander - The code could use some more refactoring. The `GetColumnInfo` and the subsequent check should be in a single method like `DoesColumnExist`. Perhaps something like `IsColumnOfType`. But I guess you just wanted a sanity check and not a PR review right?

Looks good to me otherwise.

jwilander:
Yes, looking mainly for a sanity check but the feedback is always appreciated. The special cloud migration code should be temporary and go away once we land on our official cloud friendly schema migration strategy.

agnivade (approved):

PR #16639, opened 2021-01-05
MM-31699 Don't downgrade plugins with feature flag for on-prem

crspeller:
#### Summary
For on-prem, we don't need downgrade control. 

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-31699

#### Release Note
```release-note
NONE
```

### DIFF
--- a/app/plugin.go
+++ b/app/plugin.go
@@ -930,7 +930,7 @@ func (a *App) installFeatureFlagPlugins() {
 				}
 
 				if parsedVersion.LTE(parsedExistingVersion) {
-					a.Log().Debug("Skip installation because given version was a downgrade and this isn't cloud.", mlog.String("plugin_id", pluginId), mlog.Err(err), mlog.String("version", pluginStatus.Version))
+					a.Log().Debug("Skip installation because given version was a downgrade and on-prem installations should not downgrade.", mlog.String("plugin_id", pluginId), mlog.Err(err), mlog.String("version", pluginStatus.Version))
 					return
 				}
 			}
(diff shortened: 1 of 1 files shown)


lieut-data (approved):
Looks good -- any way to unit test?

app/plugin.go
```diff
@@ -914,6 +914,27 @@ func (a *App) installFeatureFlagPlugins() {
...
+					return
+				}
+				parsedExistingVersion, err := semver.Parse(pluginStatus.Version)
+				if err != nil {
+					a.Log().Debug("Bad version from plugin manifest", mlog.String("plugin_id", pluginId), mlog.Err(err), mlog.String("version", pluginStatus.Version))
+					return
+				}
+
+				if parsedVersion.LTE(parsedExistingVersion) {
+					a.Log().Debug("Skip installation because given version was a downgrade and this isn't cloud.", mlog.String("plugin_id", pluginId), mlog.Err(err), mlog.String("version", pluginStatus.Version))
```
cpoile, line 933: We might want to say something like:
```suggestion
					a.Log().Debug("Skip installation because given version was a downgrade and on-prem installations should not downgrade.", mlog.String("plugin_id", pluginId), mlog.Err(err), mlog.String("version", pluginStatus.Version))
```

cpoile (approved):
Nice. Non-blocking wording suggestion (sys admins might not understand the reason if we just say "this isn't cloud")

crspeller:
@prapti Testing the same as before, except that for a local instance (no cloud license) it should not downgrade an installed version, only upgrade. Up/Down should work in cloud.

prapti:
Test in local environment in progress.

prapti (approved):
Tested in local server. The IM plugin doesn't get downgraded anymore with these changes.
LGTM!

PR #16644, opened 2021-01-07
MM-31715: Flip ExperimentalGossip to be true by default

agnivade:
https://mattermost.atlassian.net/browse/MM-31715

```release-notes
The UseExperimentalGossip field under ClusterSettings is now true by default.
This means that new installations will use the gossip protocol for cluster communication.
There will be no changes to existing installations.

The Gossip protocol is now considered to be in General Availability and is the recommended clustering mode.
```

### DIFF
--- a/model/config.go
+++ b/model/config.go
@@ -848,7 +848,7 @@ func (s *ClusterSettings) SetDefaults() {
 	}
 
 	if s.UseExperimentalGossip == nil {
-		s.UseExperimentalGossip = NewBool(false)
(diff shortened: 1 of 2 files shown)


grundleborg (approved):

grundleborg:
@agnivade could you edit the release note to say that Gossip is now considered to be General Availability and is the recommended clustering mode? Just to make it clear that it's not experimental any more.

agnivade:
Yep, done.

wiersgallak (approved):

PR #16650, opened 2021-01-08
Job metrics & dynamic Grafana annotations

lieut-data:
#### Summary
Extend the work in https://github.com/mattermost/mattermost-server/pull/16649 to track active jobs by type, introducing a new gauge vector, `mattermost_jobs_active` that increments when a job starts and decrements when it ends.

Use this to display an active jobs chart:
image

As well as dynamically addding a range annotation corresponding to jobs being active:
image

In addition to the job metrics above, track server start time by introducing a new gauge, `mattermost_system_server_start_time`, that's set to the current time on server start. Use this to dynamically add an annotation corresponding to the event:
image

The dynamic annotations builds on the work of https://www.inovex.de/blog/grafana-annotations-prometheus-deep-dive/. Under the covers, Grafana is just doing range queries over the raw Prometheus timeseries data and effectively rendering an event whenever the value changes. This means that jobs whose runtime is less than the Prometheus polling interval are unlikely to be "seen" at all -- but the most interesting use of these annotations is to streamline analysis when the job /is/ long running (like an LDAP sync).

The changes in this PR make these new settings available to developers, but if if we approve of the changes, we'll want to push this upstream to grafana.com as well.

Note that I have not yet tested this experience in an HA environment, where it might need some further tweaking in the charts.

#### Ticket Link
None, yet.

#### Release Note
```release-note
Introduce new mattermost_system_server_start_time and mattermost_jobs_active metrics for improved debugging with Grafana dashboards.
```


jobs/jobs.go
```diff
@@ -83,6 +87,9 @@ func (srv *JobServer) SetJobSuccess(job *model.Job) *model.AppError {
 	if _, err := srv.Store.Job().UpdateStatus(job.Id, model.JOB_STATUS_SUCCESS); err != nil {
 		return model.NewAppError("SetJobSuccess", "app.job.update.app_error", nil, err.Error(), http.StatusInternalServerError)
 	}
+
```
agnivade, line 90: We'll also need to decrement for canceled, error, and requestCancellation

agnivade (changes requested):
Nice work! So IIUC, essentially any Prometheus metric can be converted to an annotation? Or is it only for Gauges?

jobs/jobs.go
lieut-data: Ah, yes, thank you! I had meant to ask about these and address them preemptively because I wasn't confident about the state changes, but I've pushed a change now that amounts to:
* When `SetJobError` transitions to `JOB_STATUS_ERROR` (either unilaterally, or successfully after an optimistic update), we decrement.
* On `SetJobCanceled`, we decrement.
* If `RequestCancellation` successfully transitions to `JOB_STATUS_CANCELED`, we decrement.

In all other cases, in particular when we only transition to `JOB_STATUS_CANCEL_REQUESTED`, we don't decrement, but wait for the end state transition from a `SetJobCanceled` or a `SetJobError`. Let me know if my assumptions are valid here :)

lieut-data:
> So IIUC, essentially any Prometheus metric can be converted to an annotation? Or is it only for Gauges?

Assuming I understand correctly, Grafana can interpret any Prometheus query as a potential data source for annotations. Any non-zero/non-null value in the timeseries is used to break up regions. Gauges that sit at zero for most of the time are ideally suited to drive this, but in the linked blog post there are examples of using `rate` against a counter. There is also special handling for time series whose value is a constant timestamp, as illustrated with the server start time.

It will be interesting to see the performance of the dashboard (rendering, etc.) against a Prometheus instance with more data (and against a longer time range).

lieut-data:
(Sigh, pushed that up prematurely. Will ping when ready.)

agnivade (commented):
Looking great now @lieut-data ! Would it be possible to add some tests to this? :see_no_evil: Just for the individual methods. Shouldn't be too hard because we already have a mocked store, and mocked metrics interface. We just need to verify that the proper metrics methods are called.

We are a bit short on QA in the server team, so wherever possible, any tests would be very appreciated :)

lieut-data:
Absolutely! The parallel enterprise PR has some updates to a job which happens to exercise a few of the metrics, but I agree we need something more holistic here. I'll submit something accordingly :)

lieut-data:
@agnivade, mocked unit tests added. I admit these feel "less than useful", in that they don't materially test anything other than call sequence. If you have any suggestions on how to improve, I'll all ears, but I find this is usually an inevitable outcome from everything being mocked.

agnivade:
@lieut-data - I think it depends on what is it we want to test. I mainly wanted to verify that the metrics calls were happening. This just leaves out the semantic verification of the value of the calls themselves, which would actually mean we need to test with a live Prometheus instance. I would call that a decent tradeoff.

agnivade (approved):
Looks great

streamer45 (approved):
Awesome :+1:

agnivade:
@lindalumitchell - can be qa-deferred.

lindalumitchell (approved):

lieut-data:
While waiting for the root `dev-metrics` branch and merge to complete, I did some testing in HA during a loadtest and tweaked the annotation queries slightly to include the instance as a tag, as well as constrain to the selected instances. It's pretty neat to see this "in action":

image

PR #16654, opened 2021-01-09
MM-31875 soft delete reactions

wiggin77:
#### Summary
This PR changes the way reactions are deleted with respect to database storage.  Previously, reactions were hard deleted from the Reactions table.  With this PR, reactions are soft deleted by setting the new `DeleteAt` field to non-zero.  Re-adding the same reaction (same user, post, emoji) simply sets `DeleteAt` back to zero. 

An `UpdateAt` field was also added to track last update timestamp.

This change is in support of the upcoming shared channels feature and is needed so that reactions can be correctly synchronized with remote sites after deletion.  Synchronization is achieved by fetching all reactions with an `UpdateAt` greater than a remembered `LastSyncAt` timestamp tracked for each remote cluster.  

Reactions are still hard deleted when removed due to data retention policy.

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-31875

#### Release Note

```release-note
Deleting a reaction is now a soft delete in the Reactions table. A schema update is required and may take up to 15 seconds on first run with large data sets.
```


agnivade:
Ahh I knew this would happen sooner or later. With posts, we were already soft-deleting which helped us avoid this complexity so far.

I am wondering if we can use a generic way to sync more things in future? It may be just posts and reactions for now, but later, we'd need to synchronize arbitrary objects between multiple sites. It might become tedious to be adding Deleteat and Updateat to other tables that don't have it. And things will get _really_ complicated if an object becomes editable by multiple sites. This might also pollute installations that don't actually use shared channels, but still need those columns because the code is same.

We need a generic way to sync objects and data within multiple sites in a distributed manner. The "proper" way would be to use a CRDT. And if you squint closely, this is _exactly_ a 2P-Set, or even better an OR set. But perhaps a temporary middle-ground can be some sort of a WAL log table? Because for now, reactions can only be edited/deleted by the author/local sysadmin. So there are no conflicts to resolve (which is primarily where a CRDT shines). Perhaps we could have a table like (sequential_primary_key, action, table_name, JSON representation of object, timestamp)? And then when a sync happens, all of these entries could be sent to all remote sites, and then each site can parse the JSON, know what is the action being performed, and then apply that. And the local site, after sending the records successfully, can vacuum the old records.

This also serves as a generic scratch table for any future possible synchronization requirements that come up. Eventually at some point, we'd need to use a CRDT when there are objects that become editable by any site.

To be completely transparent, I just ripped off this idea from: https://hero35.com/dotjs/2019/crdts-for-mortals.

Anyways, I fully realize that what I propose completely disregards all of the work that you have put in this. Completely fine with the existing approach if there are deadlines that need to be met. But we might want to think of a generic sync mechanism between sites, because more things are likely going to be synced in future. And this also relates to my ordering proposal for shared channels.

wiggin77:
> The "proper" way would be to use a CRDT.

Maybe. As you know there is a initiative started to look at reliable sync for clients, which ideally would be a subset of inter-server sync. I like CRDT, however it can be complicated and cumbersome to implement on non-trivial data sets.  Once implemented correctly it is very robust. We should prototype something however it is way out of scope for shared channels MVP or even V1.

Thanks for the feedback @agnivade.

store/sqlstore/reaction_store.go
```diff
@@ -110,22 +112,31 @@ func (s *SqlReactionStore) BulkGetForPosts(postIds []string) ([]*model.Reaction,
 
 func (s *SqlReactionStore) DeleteAllWithEmojiName(emojiName string) error {
```
jespino, line 113: This function is only executed when you remove the Custom emoji, so... should it be "permanent delete" instead of mark as deleted? Maybe we can rename it to "PermanentDeleteAllWithEmojiName" to be explicit about the behavior, but I think this should keep permanent deleting the reactions, because the underneath custom emoji is no longer there.

store/sqlstore/reaction_store.go
```diff
@@ -167,23 +178,80 @@ func (s *SqlReactionStore) PermanentDeleteBatch(endTime int64, limit int64) (int
...
+			`INSERT INTO
+				Reactions
+				(UserId, PostId, EmojiName, CreateAt, UpdateAt, DeleteAt)
+			VALUES
+				(:UserId, :PostId, :EmojiName, :CreateAt, :UpdateAt, 0)
+			ON DUPLICATE KEY UPDATE
+				UpdateAt = :UpdateAt, DeleteAt = 0`, params); err != nil {
+			return err
+		}
+	} else if s.DriverName() == model.DATABASE_DRIVER_POSTGRES {
```
jespino, line 201: With the last min-requirements update on postgres, can we now use the "ON CONFLICT" to solve this? (Probably out of the scope of the PR, but maybe is interesting to create a ticket for that).

jespino (changes requested):
O couple of tiny comments, otherwise looks good to me.

store/sqlstore/reaction_store.go
wiggin77: Yes, I was unaware that was the only use of that method.  It should be renamed.  Still, soft delete here does no harm since there is no way to undelete since the emoji no longer exists and the soft deleted record will never appear in any queries.

store/sqlstore/reaction_store.go
wiggin77: Yes, there are three areas we do upserts that would benefit from ON CONFLICT.  If our minimum is at 9.5 or better I'll create a ticket for those.

wiggin77:
> O couple of tiny comments, otherwise looks good to me.

Thanks for revewing @jespino

store/sqlstore/reaction_store.go
wiggin77: Thinking about this some more...  I think perhaps `DeleteAllWithEmojiName` should do a soft delete.  Since we will be keeping reaction history it should be complete.  It can be used for compliance/audit to prove someone read a post.  Despite the emoji not longer existing the records are safe in that they can never be undeleted, unless the emoji is re-added.

store/sqlstore/reaction_store.go
agnivade: Our minimum is 10.0 now.

wiggin77: Ticket to use ON CONFLICT  https://mattermost.atlassian.net/browse/MM-31906

agnivade:
I wasn't suggesting to implement a CRDT. Rather the WAL-log table approach. But I completely understand if we don't want to go that route. My only concern is that this will pollute tables for sites that don't actually use shared channels, and that this might lead to a slippery slope of using the same soft-delete approach for all other objects that we might want to sync in future.

store/sqlstore/reaction_store.go
```diff
@@ -167,23 +178,80 @@ func (s *SqlReactionStore) PermanentDeleteBatch(endTime int64, limit int64) (int
...
+		// postgres has no way to upsert values until version 9.5 and trying inserting and then updating causes transactions to abort
+		count, err := transaction.SelectInt(
+			`SELECT
```
agnivade, line 204: I'd prefer to use an `ON CONFLICT` here. It doesn't seem like a lot of work. Otherwise, we can simply use a CTE. There is no need to have 2 separate queries.

store/sqlstore/upgrade.go
```diff
@@ -965,6 +965,11 @@ func upgradeDatabaseToVersion532(sqlStore *SqlStore) {
 	// allow 10 files per post
 	sqlStore.AlterColumnTypeIfExists("Posts", "FileIds", "text", "varchar(300)")
 	sqlStore.CreateColumnIfNotExistsNoDefault("Channels", "Shared", "tinyint(1)", "boolean")
+	sqlStore.CreateColumnIfNotExistsNoDefault("Reactions", "UpdateAt", "bigint", "bigint")
+	sqlStore.CreateColumnIfNotExistsNoDefault("Reactions", "DeleteAt", "bigint", "bigint")
```
agnivade, line 969: The reactions table has a very important multi-column index which is _very_ sensitive to performance. If we want to add these columns, we need to verify performance and possibly add new indexes/update existing ones.

Customers have extremely large reactions tables, so we might also want to run some simulations on a 10M reactions table and see how long these column additions take.

Basically, just some due diligence on index performance, and verifying that there are no regressions.

agnivade (commented):
Wondering if we can add some tests in the store layer to verify the new column values?

wiggin77:
> Wondering if we can add some tests in the store layer to verify the new column values?

The existing tests exercise DeleteAt well.  I added some testing for UpdateAt via https://github.com/mattermost/mattermost-server/pull/16654/commits/d672f78f31d6d9c43e2e787e6aa6d87118b7afa5

store/sqlstore/reaction_store.go
wiggin77: Now using ON CONFLICT  https://github.com/mattermost/mattermost-server/pull/16654/commits/474429d278b1475ed9b7ec1690b3611b291b0360

store/sqlstore/reaction_store.go
wiggin77: Now using ON CONFLICT https://github.com/mattermost/mattermost-server/pull/16654/commits/474429d278b1475ed9b7ec1690b3611b291b0360

store/sqlstore/upgrade.go
nronas: To second what @agnivade said, would be awesome if we can have some explains on a moderately large reactions table.

store/sqlstore/reaction_store.go
jespino: Sounds sensible. I think now the reactions are not included in the compliance export, but something to consider, of course.

jespino (approved):
Now looks good to me.

store/sqlstore/upgrade.go
wiggin77: This being handled via the Perf Review tag that Claudio is helping with.

streamer45: I've run the upgrade steps on both db engines (running on `db.r5.4xlarge`) using a rather big reactions table (~8.5M entries). Here are the results:

#### PostgreSQL

```
=> ALTER TABLE Reactions ADD COLUMN UpdateAt bigint;
ALTER TABLE
Time: 3.851 ms
=> ALTER TABLE Reactions ADD COLUMN DeleteAt bigint;
ALTER TABLE
Time: 3.670 ms
=> EXPLAIN ANALYZE UPDATE Reactions SET UpdateAt=CreateAt, DeleteAt=0 WHERE DeleteAt IS NULL;
                                                         QUERY PLAN                                                         
----------------------------------------------------------------------------------------------------------------------------
 Update on reactions  (cost=0.00..191756.34 rows=42907 width=88) (actual time=137371.651..137371.651 rows=0 loops=1)
   ->  Seq Scan on reactions  (cost=0.00..191756.34 rows=42907 width=88) (actual time=0.013..2539.940 rows=8581334 loops=1)
         Filter: (deleteat IS NULL)
 Planning Time: 0.186 ms
 Execution Time: 137371.675 ms
(5 rows)

Time: 137461.137 ms (02:17.461)
```

#### MySQL

```
> ALTER TABLE Reactions ADD COLUMN UpdateAt bigint;
Query OK, 0 rows affected (12.75 sec)

> ALTER TABLE Reactions ADD COLUMN DeleteAt bigint;
Query OK, 0 rows affected (13.05 sec)

> UPDATE Reactions SET UpdateAt=CreateAt, DeleteAt=0 WHERE DeleteAt IS NULL;
Query OK, 8581334 rows affected (4 min 30.41 sec)
Rows matched: 8581334  Changed: 8581334  Warnings: 0
```

As expected the `UPDATE` query can take a significant amount of time. I guess we need to decide whether having the server stuck for a few minutes is considered acceptable (although this time could significantly change between different db setups). If not, with a bit more logic I think we should be able to handle this on the application side, essentially checking for zero `UpdateAt`/`DeleteAt` values (through the use of `COALESCE`).

streamer45: Also ran a comparative load-test on the same dataset which did **not** reveal any performance issues/regressions:

```
==================================================
Comparison result:
Report: report_0_postgresql_unbounded.md
Grafana Dashboard: http://54.162.212.152:3000/d/6ZPSLWBGz/comparison-0-postgresql-unbounded
5.30.1:
  DB Engine: postgresql
  Supported Users: 2611
  Errors: 3094
MM-31875:
  Type: unbounded
  DB Engine: postgresql
  Supported Users: 2698
  Errors: 2983
==================================================

==================================================
Comparison result:
Grafana Dashboard: http://52.200.68.215:3000/d/_0NNYZBGz/comparison-0-mysql-unbounded
5.30.1:
  Type: unbounded
  DB Engine: mysql
  Supported Users: 2557
  Errors: 2896
MM-31875:
  Type: unbounded
  DB Engine: mysql
  Supported Users: 2677
  Errors: 2520
==================================================
```

As far as db indexes I am going to suggest **not** adding/modifying anything. The existing one on the multi-column primary key is working extremely well. 

I've benchmarked a possible addition of a `Reactions.PostId, Reactions.DeleteAt` index which did not show any performance improvement. This is somewhat expected given that the costliest part is getting reactions for a given `postId` after which the number of entries is already drastically reduced. 

Therefore, unless we expect degenerate cases of tens of thousands of reactions per post, I'd say we should be okay with what we already have.

```
                       base time/op  with_added_index time/op  delta
Reactions/postgres/GetForPost-8   685µs ± 1%   732µs ± 4%  +6.92%  (p=0.000 n=8+10)
                       base time/op  with_added_index time/op  delta
Reactions/mysql/GetForPost-8      753µs ± 3%   805µs ± 3%  +6.97%  (p=0.000 n=10+10)
```

agnivade: Amazing work @streamer45 ! I'll join in with the recommendation of handling the UPDATE in the application logic itself via COALESCE.

And even the column additions are in the order of seconds. So I'll suggest to go the route of offline SQL queries - what we did when we changed the multi-column index for this table.

Essentially, have a very EXPLICIT mention in the release notes to run the upgrade query offline during low traffic hours. And _then_ upgrade the server. Otherwise, the server will be blocked for a while.

wiggin77: I've removed the code to clear out the NULLs on database upgrade, and updated all the SELECTS to use COALESCE via https://github.com/mattermost/mattermost-server/pull/16654/commits/8a66a4dac02d59091e6b4cab09969574443ccf7d

streamer45 (approved):

store/sqlstore/reaction_store.go
```diff
@@ -110,22 +123,36 @@ func (s *SqlReactionStore) BulkGetForPosts(postIds []string) ([]*model.Reaction,
...
 
 	if _, err := s.GetReplica().Select(&reactions,
 		`SELECT
-				*
+				UserId,
+				PostId,
+				EmojiName,
+				CreateAt,
+				COALESCE(UpdateAt, CreateAt) As UpdateAt,
+				COALESCE(DeleteAt, 0) As DeleteAt
```
agnivade, line 141: We can still use `*` here, can't we? Because in the for loop later, we only use the `PostId`. UpdateAt and DeleteAt are completely ignored.

store/sqlstore/reaction_store.go
```diff
@@ -167,23 +194,60 @@ func (s *SqlReactionStore) PermanentDeleteBatch(endTime int64, limit int64) (int
...
+				(:UserId, :PostId, :EmojiName, :CreateAt, :UpdateAt, 0)
+			ON CONFLICT (UserId, PostId, EmojiName) 
+				DO UPDATE SET UpdateAt = :UpdateAt, DeleteAt = 0`, params); err != nil {
+			return err
+		}
+	}
 	return updatePostForReactionsOnInsert(transaction, reaction.PostId)
 }
 
 func deleteReactionAndUpdatePost(transaction *gorp.Transaction, reaction *model.Reaction) error {
```
agnivade, line 232: Since you converted one function to a method above, should we also do the same here?

agnivade (commented):
Looking great now! Just some small comments.

A problem that I realized with offline schema update is that this is not backward compatible. So I'd suggest explicitly mentioning in the release notes that for large data sets, it is expected that the startup phase of Mattermost might take upto 15 seconds. Just so that users are aware of it, and avoid scenarios of botched upgrades.

store/sqlstore/reaction_store.go
wiggin77: Gorp still has to convert the NULL to a Int64 to fill the struct and that is where it fails on NULL.  I made the unit test force NULLs just in case someone sets it back to `*` at some point.

store/sqlstore/reaction_store.go
wiggin77: I only converted the function above to a method because I needed access to the db driver name.  I left this one alone as it was unrelated.

agnivade (approved):
Amazing stuff! Just a small reminder on the release notes clarification. Everything else looks great.

mgdelacroix (approved):

nronas (approved):
Awesome stuff. LGTM 🚀

wiggin77:
Hi @wiersgallak.  This one's a bit time sensitive in support of shared channels.

wiersgallak (approved):

wiggin77:
Ready for QA.  Test steps are in the ticket.

lindalumitchell (approved):
Tested around reactions as described. No issues found. 👍

PR #16656, opened 2021-01-10
MM-31876: Fix racy test GetReplica

agnivade:
After starting a db connection, we would actually compare
the dbConn struct using reflection. This would race
with the connection cleaner goroutine which would modify
the internal structure of the db conn object.

There is no direct way to compare database connections safely
while it is being used. A hack might be to use the inveitable time.Sleep.
But it's cleaner to just remove the code altogether. It's only
testing a small part of the logic of assigning masters and replicas.

https://mattermost.atlassian.net/browse/MM-31876


streamer45:
If the issue is just with the deep equality can't we just compare the pointers? Either using `require.Same/NotSame` or even `require.True(t, ptr1 == ptr2)`.

agnivade:
Good idea :bulb:

agnivade:
Only one didn't work- a map iteration compare. Since order of map iteration is undefined, it fails. We could obviously work around it by either taking a slice of keys, or just using an ordered slice. But since it's just a pointer comparison of a sub-test, thought of leaving it out. Let me know if you think it's important enough.

streamer45:
> Only one didn't work- a map iteration compare. Since order of map iteration is undefined, it fails. We could obviously work around it by either taking a slice of keys, or just using an ordered slice. But since it's just a pointer comparison of a sub-test, thought of leaving it out. Let me know if you think it's important enough.

I guess we could assert the lengths of the two maps match first and then range over one and verify the keys are equal. Not a blocker though.

streamer45 (approved):
Looks good :+1:

agnivade:
> I guess we could assert the lengths of the two maps match first and then range over one and verify the keys are equal. Not a blocker though.

We'll need to sort them first. Sorting pointers just feel weird to me :/ Or use a slice.

streamer45:
> > I guess we could assert the lengths of the two maps match first and then range over one and verify the keys are equal. Not a blocker though.
> 
> We'll need to sort them first. Sorting pointers just feel weird to me :/ Or use a slice.

Why do we need to sort them? I thought we just had to verify the two maps contained the same elements. More like a set equality.

nronas (approved):
LGTM 🚀

agnivade:
> Why do we need to sort them? I thought we just had to verify the two maps contained the same elements. More like a set equality.

The assert.Same uses reflection to compare objects. So while comparing maps, it fails (I haven't dug too deep into the testify code). That's why set equality becomes bit complicated without writing some boilerplate code first to handle it.

Did you have something else in mind?

streamer45:
> Did you have something else in mind?

I was thinking of converting:

```go
assert.Equal(t, replicas, searchReplicas)
```

into something like:

```go
assert.Equal(t, len(replicas), len(searchReplicas))
for k := range replicas {
  assert.Contains(t, searchReplicas, k)
}
```

agnivade:
Sorry, `Contains` triggers the race again :P Can't fight reflection.

streamer45:
> Sorry, `Contains` triggers the race again :P Can't fight reflection.



Mmm, does 

```go
assert.True(t, searchReplicas[k])
``` 
also trigger the race?

agnivade:
Hehe, no it doesn't. Nice one.

PR #16659, opened 2021-01-11
MM-31713 Server/API: GetUserThread method

reflog:
#### Summary

added single thread GET

#### Release Note

```release-note
API method for retrieving single thread added: GET /{team_id}/threads/{thread_id}
```


ashishbhate (approved):

calebroseland (commented):
When the thread is not currently being followed by the user (but it was previously followed), it returns a 500. Could this maybe instead return the user thread with `{is_following: false}`?

```json
{"id":"app.user.get_threads_for_user.app_error","message":"Unable to get user threads","detailed_error":"sql: no rows in result set","request_id":"cfzdkpqoebramqemqik9xoz5gr","status_code":500}
``` 

And each participant seems to only contain the Id (everything else seems nil/default)

```json
{"id":"997xsqoi77dq8k74n1ubyxnmra","delete_at":0,"username":"","auth_data":"","auth_service":"","email":"","nickname":"","first_name":"","last_name":"","position":"","roles":"","locale":"","timezone":null}
```

ashishbhate:
>And each participant seems to only contain the Id (everything else seems nil/default)

I've seen this behaviour for getting all threads as well. Using `extended=true` fills out all the data. I assumed this was by design.

calebroseland:
> > And each participant seems to only contain the Id (everything else seems nil/default)
> 
> I've seen this behaviour for getting all threads as well. Using `extended=true` fills out all the data. I assumed this was by design.

I tried that, but it didn't seem to fill things in and the method in `client4.go` doesn't seem to have any of those extra parameters.

With `extended=false`, I didn't realize that it it would include the empty properties beyond `participants: [{id}]`.

ashishbhate:
>I tried that, but it didn't seem to fill things in and the method in client4.go doesn't seem to have any of those extra parameters.

I meant in the case of getting all threads. In this PR (getting a single thread), the `extended` parameter hasn't been implemented.

>With extended=false, I didn't realize that it it would include the empty properties beyond participants: [{id}].

I'm assumed this is how it was meant to be. I'm not sure if it is.

reflog:
@calebroseland @ashishbhate - I indeed forgot the 'extended' parameter.
I've added implementation for it.

Regarding threads that are not followed - the GetUserThreads and GetUserThread are specifically for getting followed threads, so there's not point in adding {is_followed:false} to response.
Instead, when getting a non-followed thread you will now get a 404

calebroseland (approved):
Thanks, sounds good!

PR #16660, opened 2021-01-11
MM-31712 Server/API: Unreads-only filter for GetUserThreads

reflog:
#### Summary

Added 'unread' field in query

#### Release Note

```release-note
updated `/users/{user_id}/teams/{team_id}/threads` api to support unread=true query parameter
```


ashishbhate (approved):

calebroseland (approved):
LGTM, though I have question on `since` handling

PR #16663, opened 2021-01-11
MM-31710 Server/API: change total_unread_replies to total_unread_threads

reflog:
#### Summary

total_unread_threads is now calculated instead of total_unread_replies

#### Release Note

```release-note
Changed output of Get Threads API to include total_unread_threads instead of total_unread_replies
```


calebroseland (commented):
So I'm getting some weirdness, I don't fully know how to explain it, but `&deleted=` doesn't change `total`, and `total_unread_threads` doesn't match how many unread threads are returned in `threads` (only one page, so not a paging issue). Not quite sure what's up.

This is with `threads?page=0&pageSize=200&extended=true&deleted=false&since=0`

(removed props except `id` and `unread_replies`)

CleanShot 2021-01-14 at 12 25 15@2x

ashishbhate:
With the latest changes the totals appear to be correct.

calebroseland (approved):
This is working better and I don't seem to be able to reproduce the incorrect `total_unread_threads `, so hopefully we're in the clear there.

ashishbhate (approved):

PR #16667, opened 2021-01-11
[MM-31790] Support Packet Generation BACKEND

hahmadia:
#### Summary

Support Packet Generation Backend Implementation

#### Ticket Link

https://mattermost.atlassian.net/browse/MM-31790

#### Release Note

```release-note
Allows any system admin to download a support packet which provides helpful information to our support team
```


sbishel (changes requested):

api4/system.go
```diff
@@ -64,6 +68,150 @@ func (api *API) InitSystem() {
 	api.BaseRoutes.ApiRoot.Handle("/warn_metrics/trial-license-ack/{warn_metric_id:[A-Za-z0-9-_]+}", api.ApiHandler(requestTrialLicenseAndAckWarnMetric)).Methods("POST")
 	api.BaseRoutes.System.Handle("/notices/{team_id:[A-Za-z0-9]+}", api.ApiSessionRequired(getProductNotices)).Methods("GET")
 	api.BaseRoutes.System.Handle("/notices/view", api.ApiSessionRequired(updateViewedProductNotices)).Methods("PUT")
+
+	api.BaseRoutes.System.Handle("/generate_support_packet", api.ApiSessionRequired(generateSupportPacket)).Methods("GET")
```
sbishel, line 72: Can we drop "generate" on the RestAPI.  Essentially it is a "Get" "support_packet". I'm ok with `generateSupportPacket` for internal function name.

hahmadia: Okay! Done. Thanks :)

api4/system.go
```diff
@@ -64,6 +68,150 @@ func (api *API) InitSystem() {
...
+			Filename: "support_packet.yaml",
+			Body:     supportPacketYaml,
+		},
+	}
+
+	// Getting the plugins installed on the server, prettify it, and then add them to the file data array
+	pluginsResponse, appErr := c.App.GetPlugins()
+	if appErr == nil {
+		pluginsPrettyJSON, err := json.MarshalIndent(pluginsResponse, "", "    ")
+		if err == nil {
```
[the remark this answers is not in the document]
hahmadia: Yes! Great idea! I had this in mind but wasn't sure about it.

api4/system.go
```diff
@@ -64,6 +68,150 @@ func (api *API) InitSystem() {
...
+				Body:     sanitizedConfigPrettyJson,
+			})
+	}
+
+	// Getting mattermost.log and notifications.log
+	if *c.App.Srv().Config().LogSettings.EnableFile {
+		// mattermost.log
+		mattermostLog := utils.GetLogFileLocation(*c.App.Srv().Config().LogSettings.FileLocation)
+
+		mattermostLogFileData, mattermostLogFileDataErr := ioutil.ReadFile(mattermostLog)
```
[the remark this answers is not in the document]
hahmadia: I guess I'm a little confused by this. So I did notice that in `utils/logger.go` we have a `LogRotateSize = 10000` which matches with what the front-end description says "It gets rotated every 10,000) but I couldn't ever find that variable being used in either the server repo or enterprise repo. 

I spent some time trying to figure out how the rotating log files works but really couldn't. Maybe we can setup a meeting to go over this together.

api4/system.go
```diff
@@ -64,6 +68,150 @@ func (api *API) InitSystem() {
 	api.BaseRoutes.ApiRoot.Handle("/warn_metrics/trial-license-ack/{warn_metric_id:[A-Za-z0-9-_]+}", api.ApiHandler(requestTrialLicenseAndAckWarnMetric)).Methods("POST")
 	api.BaseRoutes.System.Handle("/notices/{team_id:[A-Za-z0-9]+}", api.ApiSessionRequired(getProductNotices)).Methods("GET")
 	api.BaseRoutes.System.Handle("/notices/view", api.ApiSessionRequired(updateViewedProductNotices)).Methods("PUT")
+
+	api.BaseRoutes.System.Handle("/generate_support_packet", api.ApiSessionRequired(generateSupportPacket)).Methods("GET")
+}
+
+func generateSupportPacket(c *Context, w http.ResponseWriter, r *http.Request) {
```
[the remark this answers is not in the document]
hahmadia: Great Idea!

api4/system.go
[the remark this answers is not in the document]
hahmadia: I looked into this and I don't think we do. 
I noticed that we never set the rotating of the files by the lines (unless I missed something here) but rather by file size (https://github.com/mattermost/mattermost-server/blob/master/mlog/log.go#L141). We can see we are setting the Max Size of the file to 100mb (which means when the logger file reaches 100mb, it gets rotated). If there is several of these, the zip file size could become big pretty fast.

I think the extra work that needs to be done for this doesn't make sense. Having the current log (which can hold up to a 100mb of information) is enough I believe. Otherwise, we may get into situations where the zip file is too big to attach and may take a while to download.

api4/system.go
```diff
@@ -64,6 +66,61 @@ func (api *API) InitSystem() {
...
+
+	// Checking to see if the server has a e10 or e20 license (this feature is only permitted for servers with licenses)
+	if c.App.Srv().License() == nil {
+		c.Err = model.NewAppError("Api4.generateSupportPacket", "api.no_license", nil, "", http.StatusForbidden)
+		return
+	}
+
+	fileDatas := c.App.GenerateSupportPacket()
+
+	// Constructing the ZIP file name as per spec (mattermost_support_packet_YYYY-MM-DD-HH-MM.zip)
```
sbishel, line 91: Can we create a generic utility to create the zip file? We already do this for compliance.

app/server.go
```diff
@@ -1614,3 +1624,142 @@ func (s *Server) HttpService() httpservice.HTTPService {
 func (s *Server) SetLog(l *mlog.Logger) {
 	s.Log = l
 }
+
+func (a *App) GenerateSupportPacket() []model.FileData {
+	// If any errors we come across within this function, we will log it in a warning.txt file so that we know why certain files did not get produced if any
+	var warnings string
```
sbishel, line 1630: Would this be better as a []string

app/server.go
```diff
@@ -1614,3 +1624,142 @@ func (s *Server) HttpService() httpservice.HTTPService {
...
+
+	// Creating an array of files that we are going to be adding to our zip file
+	fileDatas := []model.FileData{}
+
+	a.generateSupportPacketYaml(&warnings, &fileDatas)
+	a.createPluginsFile(&warnings, &fileDatas)
+	a.createSanitizedConfigFile(&warnings, &fileDatas)
+	a.getMattermostLog(&warnings, &fileDatas)
+	a.getNotificationsLog(&warnings, &fileDatas)
```
sbishel, line 1639: Prefer to return warnings and filedatas, keeps the other functions more pure.

sbishel (changes requested):

app/server.go
hahmadia: Done Thanks!

app/server.go
hahmadia: Good idea! Thanks!

api4/system.go
hahmadia: I noticed the compliance utility functions however I'm not exactly sure how you'd like me to approach this. I was thinking of creating a utility function in `model/utils` which just creates a zip and attaches an array of `fileDatas` and then returns a reference to the zip file? Let me know your thoughts. 

Example file signature
`func CreateZipFileAndAttachFiles (fileName string, files []model.FileDatas) (*os.File, error)`

sbishel (changes requested):

app/server.go
```diff
@@ -1625,100 +1625,157 @@ func (s *Server) SetLog(l *mlog.Logger) {
...
-	a.createSanitizedConfigFile(&warnings, &fileDatas)
-	a.getMattermostLog(&warnings, &fileDatas)
-	a.getNotificationsLog(&warnings, &fileDatas)
+	fileData, warning := a.generateSupportPacketYaml()
+
+	if warning == "" {
+		fileDatas = append(fileDatas, fileData)
+	} else {
+		warnings = append(warnings, warning)
+	}
```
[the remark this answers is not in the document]
hahmadia: Sorry! This is a OOPSY on my end 😅 The idea was that if warning has text, we will ignore the return of fileData. The reason I had done it like this was so we don't have multiple return statements. Anyhow, I can see the confusion here and have refactored it to better fit my thinking. Sorry about the confusion.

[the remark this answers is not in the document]
hahmadia: I've refactored it such that now function return a `(*fileData, string)` and then we check if `fileData != nil` instead of `warning == ""`. If `fileData != nil`, then something went wrong and therefore we append our warning to string array. Hope this makes more sense.

app/server.go
```diff
@@ -1625,100 +1625,157 @@ func (s *Server) SetLog(l *mlog.Logger) {
...
-		} else {
-			*warnings = *warnings + fmt.Sprintf("ioutil.ReadFile(notificationsLog) Error: %s\n\n", notificationsLogFileDataErr.Error())
+		if notificationsLogFileDataErr != nil {
+			warning = fmt.Sprintf("ioutil.ReadFile(notificationsLog) Error: %s", notificationsLogFileDataErr.Error())
+		}
+
+		fileData = model.FileData{
+			Filename: "notifications.log",
+			Body:     notificationsLogFileData,
 		}
```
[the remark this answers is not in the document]
hahmadia: Refactored to make more clear. Initial implementation was not clear. Sorry for confusion.

api4/system.go
```diff
@@ -64,6 +66,62 @@ func (api *API) InitSystem() {
...
+	YYYYMMDDHHSSFormat := fmt.Sprintf("%d-%02d-%02d-%02d-%02d", now.Year(), now.Month(), now.Day(), now.Hour(), now.Minute())
+	outputZipFilename := fmt.Sprintf("mattermost_support_packet_%s.zip", YYYYMMDDHHSSFormat)
+
+	fileStorageBackend, fileBackendErr := c.App.FileBackend()
+	if fileBackendErr != nil {
+		c.Err = fileBackendErr
+		return
+	}
+
+	err := c.App.CreateZipFileAndAddFiles(fileStorageBackend, fileDatas, outputZipFilename, OutputDirectory)
```
hahmadia, line 103: Hey @sbishel !
Finally got this utility function working! Here it is being used.

api4/system.go
```diff
@@ -64,6 +66,62 @@ func (api *API) InitSystem() {
...
+	if err != nil {
+		c.Err = model.NewAppError("Api4.generateSupportPacket", "api.unable_to_create_zip_file", nil, err.Error(), http.StatusForbidden)
+		return
+	}
+
+	fileBytes, err := fileStorageBackend.ReadFile(path.Join(OutputDirectory, outputZipFilename))
+	defer fileStorageBackend.RemoveDirectory(OutputDirectory)
```
hahmadia, line 110: After reading the file, since it suppose to be temporary, I remove the entire directory.

app/file.go
```diff
@@ -1245,3 +1248,49 @@ func (a *App) CopyFileInfos(userId string, fileIds []string) ([]string, *model.A
 
 	return newFileIds, nil
 }
+
+// This function zip's up all the files in fileDatas array and then saves it to the directory specified with the specified zip file name
+// Ensure the zip file name ends with a .zip
+func (a *App) CreateZipFileAndAddFiles(fileBackend filesstore.FileBackend, fileDatas []model.FileData, zipFileName, directory string) error {
```
hahmadia, line 1254: Here is the utility function as we discussed 🎉 !! 
It follows what we had discussed.

hahmadia (commented):
@sbishel Comments regarding discussion around zip file utility function

app/server.go
```diff
@@ -1612,3 +1622,204 @@ func (s *Server) HttpService() httpservice.HTTPService {
...
+
+	} else {
+		warning = "Unable to retrieve mattermost.log because LogSettings: EnableFile is false in config.json"
+	}
+
+	return nil, warning
+}
+
+func (a *App) createSanitizedConfigFile() (*model.FileData, string) {
+	var warning string
```
mkraft, line 1739:
```suggestion
```

app/server.go
```diff
@@ -1612,3 +1622,204 @@ func (s *Server) HttpService() httpservice.HTTPService {
...
+	sanitizedConfigPrettyJSON, err := json.MarshalIndent(a.GetSanitizedConfig(), "", "    ")
+	if err == nil {
+		fileData := model.FileData{
+			Filename: "sanitized_config.json",
+			Body:     sanitizedConfigPrettyJSON,
+		}
+		return &fileData, ""
+	}
+
+	warning = fmt.Sprintf("json.MarshalIndent(c.App.GetSanitizedConfig()) Error: %s", err.Error())
```
mkraft, line 1751:
```suggestion
	warning := fmt.Sprintf("json.MarshalIndent(c.App.GetSanitizedConfig()) Error: %s", err.Error())
```

services/searchengine/bleveengine/bleve.go
```diff
@@ -231,6 +231,14 @@ func (b *BleveEngine) GetVersion() int {
 	return 0
 }
 
+func (b *BleveEngine) GetFullVersion() string {
```
mkraft, line 234: What was the reason for the `(*BleveEngine).GetFullVersion` vs `(*BleveEngine).GetVersion`?

api4/system.go
sbishel: Typically, if this is going to be a temp directory and be removed, it should be unique. You want to make sure that directory isn't being used for anything else. You could simply append `YYYYMMDDHHSSFormat` to `OutputDirectory`, like you did for the filename.

sbishel (changes requested):
Small change...looks good.

app/server.go
```diff
@@ -636,6 +640,12 @@ func (s *Server) AppOptions() []AppOption {
 	}
 }
 
+// Return Database type (postgres or mysql) and current version
+func (s *Server) DatabaseTypeAndVersion() (string, string) {
+	driverVersion, _ := s.Store.System().GetByName("Version")
```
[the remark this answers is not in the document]
hahmadia: Renamed + Updated comment to clarify

app/server.go
hahmadia: Removed the string variable

app/server.go
hahmadia: Done Thanks

api4/system.go
```diff
@@ -64,6 +66,62 @@ func (api *API) InitSystem() {
...
+	if c.App.Srv().License() == nil {
+		c.Err = model.NewAppError("Api4.generateSupportPacket", "api.no_license", nil, "", http.StatusForbidden)
+		return
+	}
+
+	fileDatas := c.App.GenerateSupportPacket()
+
+	// Constructing the ZIP file name as per spec (mattermost_support_packet_YYYY-MM-DD-HH-MM.zip)
+	now := time.Now()
+	YYYYMMDDHHSSFormat := fmt.Sprintf("%d-%02d-%02d-%02d-%02d", now.Year(), now.Month(), now.Day(), now.Hour(), now.Minute())
```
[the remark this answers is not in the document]
hahmadia: Shoot! I didn't know you could modify it to the way you like. Support wanted it in the format of `YYYY-MM-DD-HH-MM` with the dashes inbetween. I didn't know it could be changed to the way you want it. Thanks for this point out.

app/server.go
```diff
@@ -1612,3 +1622,204 @@ func (s *Server) HttpService() httpservice.HTTPService {
...
+	// Adding a warning.txt file to the fileDatas if any warning
+	if len(warnings) > 0 {
+		finalWarning := strings.Join(warnings, "\n")
+		fileDatas = append(fileDatas, model.FileData{
+			Filename: "warning.txt",
+			Body:     []byte(finalWarning),
+		})
+	}
+
+	return fileDatas
```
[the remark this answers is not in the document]
hahmadia: Ah! Nice catch!! I don't know why but I assumed we couldn't do this in Golang for some reason. After some further investigation, totally possible and code is much neater now :) Thanks!

api4/system.go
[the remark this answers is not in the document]
hahmadia: No worries! I don't think we need to use `TempDir` or `TempFile`. I think it's fine as it is. 
To subside the concurrent problem, I've appended a `model.NewId()`  to the output directory which returns a `globally unique identifier.` so every output directory will have a unique ID at the end meaning no concurrency issues.

hahmadia: The problem is that if I spam the download button at `11:00` , since we aren't taking into account the seconds in the naming, it still possible to come across concurrency problem. 

To subside the concurrent problem, I've appended a `model.NewId()`  to the output directory which returns a `globally unique identifier.` so every output directory will have a unique ID at the end meaning we shouldn't run into any concurrency issues.

sbishel (changes requested):
Should we have unit tests for CreateZipFileAndAddFiles? Also the functions in app/server.go

mkraft (approved):

hahmadia:
> Should we have unit tests for CreateZipFileAndAddFiles? Also the functions in app/server.go
> Otherwise, looking good.

Thanks @sbishel! Tests have now been added.

sbishel (approved):

furqanmlk (approved):
All look good ✅

PR #16671, opened 2021-01-11
Mm 28090 user settings api when ldap sync

maxerenberg:
#### Summary
This pull request prevents users from modifying some of their user settings via the Mattermost API if they authenticate via LDAP, SAML or OAUTH. It aims to have identical restrictions to those of the web UI.

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-28090

#### Release Note
```release-note
* For LDAP and SAML users, the following fields cannot be changed via the API if the corresponding
  LDAP/SAML attributes have been set: first name, last name, position, nickname, email, profile picture
* For OAUTH users (i.e. Gitlab, Google, Office365 and OpenID), the following fields cannot be changed
  via the API: first name, last name
* All users who authenticate via a method other than email cannot change their username via the API
```


i18n/en.json
```diff
@@ -3322,6 +3322,10 @@
     "id": "api.user.oauth_to_email.not_available.app_error",
     "translation": "Authentication Transfer not configured or available on this server."
   },
+  {
+    "id": "api.user.patch_user.login_provider_attribute_set.app_error",
+    "translation": "Field '{{.Field}}' must be set through user's login provider"
```
mkraft, line 3327:
```suggestion
    "translation": "Field '{{.Field}}' must be set through user's login provider."
```

api4/user.go
```diff
@@ -1129,6 +1140,93 @@ func updateUser(c *Context, w http.ResponseWriter, r *http.Request) {
...
+		FirstNameAttribute *string
+		LastNameAttribute  *string
+		NicknameAttribute  *string
+		PositionAttribute  *string
+		EmailAttribute     *string
+	}{
+		{model.USER_AUTH_SERVICE_LDAP, LS.FirstNameAttribute, LS.LastNameAttribute, LS.NicknameAttribute,
+			LS.PositionAttribute, LS.EmailAttribute},
+		{model.USER_AUTH_SERVICE_SAML, SS.FirstNameAttribute, SS.LastNameAttribute, SS.NicknameAttribute,
+			SS.PositionAttribute, SS.EmailAttribute},
```
mkraft, line 1167: Are there any differences between the logic applied for LDAP vs SAML? If there are no differences, what are your thoughts on whether this can be done in a single loop or is it best to keep two loops?

api4/user.go
```diff
@@ -1129,6 +1140,93 @@ func updateUser(c *Context, w http.ResponseWriter, r *http.Request) {
...
+			continue
+		}
+		for _, attr := range []struct {
+			name       string
+			patchValue *string
+			userValue  *string
+			selector   *string
+		}{
+			{firstNameField, patch.FirstName, &ouser.FirstName, provider.FirstNameAttribute},
+			{lastNameField, patch.LastName, &ouser.LastName, provider.LastNameAttribute},
```
mkraft, line 1179: You know how we had noticed that the front-end disables both first name and last name updates if either of those fields are configured... do you think this logic needs to match that? My only thought for making this logic match that is for consistency with the front-end and with the existing production app. What do you think?

api4/user.go
```diff
@@ -1129,6 +1140,93 @@ func updateUser(c *Context, w http.ResponseWriter, r *http.Request) {
...
+			http.StatusConflict)
+		return
+	}
+	// If an OAUTH provider is used then neither the first name nor the last
+	// name may be changed
+	for _, oauthService := range []string{
+		model.USER_AUTH_SERVICE_GITLAB,
+		"google",
+		"office365",
+		"openid",
```
mkraft, line 1206: Looks like we have some constants that we could leverage here: https://github.com/mattermost/mattermost-server/blob/master/model/config.go#L46-L49

api4/user.go
```diff
@@ -1129,6 +1140,93 @@ func updateUser(c *Context, w http.ResponseWriter, r *http.Request) {
...
+		if ouser.AuthService != oauthService {
+			continue
+		}
+		for _, attr := range []struct {
+			name       string
+			patchValue *string
+			userValue  *string
+		}{
+			{firstNameField, patch.FirstName, &ouser.FirstName},
+			{lastNameField, patch.FirstName, &ouser.LastName},
```
mkraft, line 1217:
```suggestion
			{lastNameField, patch.LastName, &ouser.LastName},
```
Do you think we should add a unit test to ensure this is covered?

mkraft (changes requested):
This is looking really good. What are your thoughts about `updateUser`? Can that endpoint be used to update the user fields? Should we also be applying the new business logic to that endpoint?

api4/user.go
maxerenberg: There are no differences, but one loop would be simpler.

api4/user.go
maxerenberg: Good point. Back-end logic should match front-end logic.

api4/user.go
maxerenberg: Unit tests for LDAP users are in the enterprise repo. Tests for SAML and OAUTH users will be added once we figure out how to run Keycloak during our tests.

hahmadia:
We usually only require 2 developers to check the code. Since Martin and Scott had already been assigned, I'm going to un-assign myself.

api4/user.go
```diff
@@ -1129,6 +1153,80 @@ func updateUser(c *Context, w http.ResponseWriter, r *http.Request) {
...
+			(ouser.AuthService == SAML && (*SS.FirstNameAttribute != "" || *SS.LastNameAttribute != "")) ||
+			isOauth(ouser.AuthService)) {
+		c.Err = model.NewAppError(
+			methodName, errName,
+			map[string]interface{}{"Field": "full name"}, "", http.StatusConflict)
+		return
+	}
+	// If LDAP or SAML is used, and any of the following attributes have been specified
+	// in the settings, then their corresponding fields may not be changed
+	for _, attr := range []struct {
```
sbishel, line 1196: Rather than doing this loop 6 times, every time. Create the array necessary for LDAP and/or SAML and loop through that array. Which means you won't loop at all for OAUTH or "".

api4/user.go
```diff
@@ -1129,6 +1153,80 @@ func updateUser(c *Context, w http.ResponseWriter, r *http.Request) {
...
+		selector    *string
+	}{
+		{LDAP, "nickname", patch.Nickname, &ouser.Nickname, LS.NicknameAttribute},
+		{LDAP, "position", patch.Position, &ouser.Position, LS.PositionAttribute},
+		{LDAP, "email", patch.Email, &ouser.Email, LS.EmailAttribute},
+		{SAML, "nickname", patch.Nickname, &ouser.Nickname, SS.NicknameAttribute},
+		{SAML, "position", patch.Position, &ouser.Position, SS.PositionAttribute},
+		{SAML, "email", patch.Email, &ouser.Email, SS.EmailAttribute},
+	} {
+		if ouser.AuthService != attr.authService {
```
sbishel, line 1210: Perform this check before ever looping.

api4/user.go
```diff
@@ -1060,6 +1071,13 @@ func autocompleteUsers(c *Context, w http.ResponseWriter, r *http.Request) {
 	w.Write([]byte((autocomplete.ToJson())))
 }
 
+func convertToUserPatch(user *model.User) *model.UserPatch {
```
sbishel, line 1074: This should probably be a method on a user type and be moved to models/users.go.
func (u *User) IsUserOAuth(){}

sbishel (changes requested):
Coming together well. Let me know if any of my comments are too vague and you have questions.

api4/user.go
```diff
@@ -464,8 +464,19 @@ func setProfileImage(c *Context, w http.ResponseWriter, r *http.Request) {
...
-	if user, err := c.App.GetUser(c.Params.UserId); err == nil {
-		auditRec.AddMeta("user", user)
+	user, err := c.App.GetUser(c.Params.UserId)
+	if err != nil {
+		c.SetInvalidParam("user_id")
+		return
+	}
+	auditRec.AddMeta("user", user)
+
+	if (user.AuthService == model.USER_AUTH_SERVICE_LDAP || user.AuthService == model.USER_AUTH_SERVICE_SAML) &&
```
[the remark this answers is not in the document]
maxerenberg: Sure. We don't seem to be doing that check in the web app though.

api4/user.go
```diff
@@ -1129,6 +1156,61 @@ func updateUser(c *Context, w http.ResponseWriter, r *http.Request) {
...
+			}
+		}
+	}
+	// If any login provider is used, then the username may not be changed
+	if ouser.AuthService != "" && patch.Username != nil && *patch.Username != ouser.Username {
+		return errors.New("username")
+	}
+	return nil
+}
+
```
sbishel, line 1213: The flow of this function could be improved to make it more understandable. The current flow is checking basically in field order. 
Check Firstname/Lastname (then check Auth)
Check Nickname/Position/Email (if auth = saml, ldap)
Check Username (if auth !"")
Then each of these have additional checks to determine whether or not to process. (ie. ouser.IsLdap()). 

I would suggest rewriting this function to be based on the users authentication and check all fields need to be checked for that authentication. Then write a helper function that each "authentication" type will call with an array of fields to check. Remember for SAML and LDAP, the fields don't need checked if the attribute isn't set.

sbishel (changes requested):
Hey Max. Just need to see if the function `checkLoginProviderAttributes` can be simplified.Let me know if you want to jump on a call to discuss.

api4/user.go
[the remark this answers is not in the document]
sbishel: If that is the case, would you write up a ticket for WebApp. Also, I think we should also be checking - LdapSettings.EnableSync. Sorry I missed that one before. Here is another place we do this type of check. We should probably make a single function. 
https://github.com/mattermost/enterprise/blob/master/ldap/ldap.go#L101

model/user.go
```diff
@@ -693,7 +693,10 @@ func (u *User) IsSSOUser() bool {
 }
 
 func (u *User) IsOAuthUser() bool {
-	return u.AuthService == USER_AUTH_SERVICE_GITLAB
+	return u.AuthService == SERVICE_GITLAB ||
+		u.AuthService == SERVICE_GOOGLE ||
+		u.AuthService == SERVICE_OFFICE365 ||
+		u.AuthService == SERVICE_OPENID
```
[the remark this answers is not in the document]
maxerenberg: The only other place where this function is used is [here](https://github.com/mattermost/mattermost-server/blob/master/store/sqlstore/user_store.go#L191), and it doesn't appear to be Gitlab-specific.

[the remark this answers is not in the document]
sbishel: And where is `Update()` called from. Make sure you understand the impact of this change. It may even be fixing a bug. Luckily its only used in one place.

api4/user.go
```diff
@@ -1129,6 +1156,69 @@ func updateUser(c *Context, w http.ResponseWriter, r *http.Request) {
 	w.Write([]byte(ruser.ToJson()))
 }
 
+// checkLoginProviderAttributes returns nil if the patch can be applied without overriding
+// attributes set by the user's login provider; otherwise, an error is returned whose text
+// is the name of the offending field.
+func checkLoginProviderAttributes(cfg *model.Config, ouser *model.User, patch *model.UserPatch) error {
```
sbishel, line 1162: No reason to return an error here. You can just return a string.

api4/user.go
```diff
@@ -1129,6 +1156,69 @@ func updateUser(c *Context, w http.ResponseWriter, r *http.Request) {
...
+		for _, attr := range []attrParam{
+			{"nickname", patch.Nickname, &ouser.Nickname, SS.NicknameAttribute},
+			{"position", patch.Position, &ouser.Position, SS.PositionAttribute},
+			{"email", patch.Email, &ouser.Email, SS.EmailAttribute},
+		} {
+			if isConflict(attr.patchValue, attr.userValue, attr.providerValue) {
+				return errors.New(attr.name)
+			}
+		}
+	}
```
sbishel, line 1212: Don't we need to check isNameConflict for OAuth? I thought the prior code did.

sbishel (changes requested):
BOOM!! I like the changes for `checkLoginProviderAttributes`. Just a couple small comments and some unresolved from prior reviews...

api4/user.go
maxerenberg: Yes, you're right. Completely forgot.

sbishel (changes requested):
Hey Max, just the one comment regarding `convertToUserPatch` and make sure the changes to `IsOauthUser` won't cause issues.

model/user.go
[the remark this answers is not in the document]
maxerenberg: The IsOAuthUser function was written 4-5 years ago and I'm guessing it just wasn't updated when we added more OAuth providers. As mentioned above, this function is only used in one other place, and it doesn't seem to be specific to Gitlab.
As for the `updateUser()` function, that is only used as the HTTP handler for the `users/{user_id}` endpoint. So the changes are specific to this endpoint.

sbishel:
@maxerenberg - Hey this PR is looking good. Please add some unit tests for functions added/changed in users.go. Also yuor `Enterprise` PR should be brought into here. Sorry, I wasn't paying too much attention to why that was in Enterprise. We should be able test that in the server repo.

sbishel (approved):

maxerenberg:
I'm going to close this PR and reopen it on the upstream repo so that the enterprise tests run against it.

PR #16675, opened 2021-01-12
MM-31433/shared channel sync notifier

nronas:
Implements: https://mattermost.atlassian.net/browse/MM-31433

#### Summary
Implements a shared channel sync notifier, which describes the functionality around when a sync of a shared channel needs to be triggered.

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-31433

#### Release Note
```release-note
NONE
```

### Pending tasks

- Remove /sync chat-op command.


nronas:
*NOTE* THe conflicts will be fixed in the parent branch, since this PR cannot be merged without it.

app/shared_channel_service_iface.go
```diff
@@ -0,0 +1,14 @@
+// Copyright (c) 2015-present Mattermost, Inc. All Rights Reserved.
+// See LICENSE.txt for license information.
+
+package app
+
+import "github.com/mattermost/mattermost-server/v5/model"
+
+// SharedChannelServiceIFace is the interface to the shared channel service
+type SharedChannelServiceIFace interface {
```
wiggin77, line 9: Idiomatic Go suggests this should be defined with the mock, and the mock should accept the concrete type from `Server.GetSharedChannelSyncService`.  e.g.:

`func newMockRemoteClusterService(service ServiceIFace) *mockRemoteClusterService {...}`

I reluctantly did something similar with the logging engine which I've noted to go back and fix.

agnivade:
@nronas - This is a bit cumbersome to review in this form. It's hard to keep track of which commit changed what. Wondering if we can wait for a while to get the pending things sorted out and then move with review?

nronas:
@agnivade totally that is exactly what i suggested to @wiggin77 yesterday as well.

model/cluster_message.go
```diff
@@ -42,6 +42,7 @@ const (
 	CLUSTER_EVENT_REMOVE_PLUGIN                                     = "remove_plugin"
 	CLUSTER_EVENT_INVALIDATE_CACHE_FOR_TERMS_OF_SERVICE             = "inv_terms_of_service"
 	CLUSTER_EVENT_BUSY_STATE_CHANGED                                = "busy_state_change"
+	CLUSTER_EVENT_SYNC_SHARED_CHANNEL                               = "sync_shared_channel"
```
nronas, line 45: i couldn't find a better name, so open to suggestions 😄 . Since this action is to signal the shared channel service to start syncing

app/server.go
```diff
@@ -146,7 +146,7 @@ type Server struct {
 	telemetryService *telemetry.TelemetryService
 
 	remoteClusterService     *remotecluster.Service
-	sharedChannelSyncService *sharedchannel.Service
+	sharedChannelSyncService SharedChannelServiceIFace
```
wiggin77, line 149: What's the reason for turning this into an interface?

app/server.go
```diff
@@ -1704,7 +1704,7 @@ func (s *Server) GetRemoteClusterService() *remotecluster.Service {
 
 // GetSharedChannelSyncService returns the `SharedChannelSyncService` instantiated by the server.
 // May be nil if the service is not enabled via license.
-func (s *Server) GetSharedChannelSyncService() *sharedchannel.Service {
+func (s *Server) GetSharedChannelSyncService() SharedChannelServiceIFace {
```
wiggin77, line 1707: This should return concrete type.

app/shared_channel_notifier.go
```diff
@@ -0,0 +1,72 @@
...
+package app
+
+import (
+	"github.com/mattermost/mattermost-server/v5/mlog"
+	"github.com/mattermost/mattermost-server/v5/model"
+)
+
+// NotifySharedChannelSync signals the syncService to start syncing shared channel updates
+func (a *App) NotifySharedChannelSync(channel *model.Channel, event string) {
+	if channel.IsShared() {
```
wiggin77, line 13: Would be cleaner to just return if `!channel.IsShared()` and not indent the rest of of the method.

app/shared_channel_notifier.go
```diff
@@ -0,0 +1,72 @@
...
+		// When the sync service is not running on the node and cluster is enabled, broadcast shared sync message
+		if a.Cluster() != nil {
+			mlog.Debug(
+				"Shared channel sync service is not running on this node. Broadcasting sync cluster event.",
+				mlog.String("channel_id", channel.Id),
+				mlog.String("event", event),
+			)
+
+			a.Cluster().SendClusterMessage(&model.ClusterMessage{
+				Event:    model.CLUSTER_EVENT_SYNC_SHARED_CHANNEL,
```
wiggin77, line 36: Is there an existing event that gets broadcast on each post?  I'm concerned we just doubled cluster traffic, which was already a pain point.

model/cluster_message.go
wiggin77: Name looks fine, but we may not need a new event type.  See comment above.

app/shared_channel_notifier.go
nronas: That is such a good point. Silly me not relying on the initial events, since we just need to react to those events, which already exist for post(create, update, delete) and reaction(add, remove). Will amend.

app/server.go
nronas: it was done in order to be able to mock the service. Maybe i am missing something but returning a concrete type of `*sharedchannel.Service` will not allow us to do so.

agnivade:
Wow, I just realised this PR has 138 commits.

nronas:
> Wow, I just realised this PR has 138 commits.

I was surprised as well, but I think the reason is that was branched off from another branch that was squashed merged on the main feature branch of shared channels

app/shared_channel_notifier.go
nronas: Amended in https://github.com/mattermost/mattermost-server/pull/16675/commits/422512b78c0bce9a9281bb4b82ed0da134b04897

app/post.go
```diff
@@ -359,6 +359,9 @@ func (a *App) CreatePost(post *model.Post, channel *model.Channel, triggerWebhoo
 		a.SendEphemeralPost(post.UserId, ephemeralPost)
 	}
 
+	// Notify shared channel sync service for the new change.
+	a.NotifySharedChannelSync(channel, model.WEBSOCKET_EVENT_POSTED)
```
nronas, line 363: We need to explicitly invoke a sync here, in order to cover the case were, the Leader is the one that handled the user intention(i.e create a post)

app/shared_channel_notifier.go
```diff
@@ -0,0 +1,73 @@
...
+
+package app
+
+import (
+	"github.com/mattermost/mattermost-server/v5/mlog"
+	"github.com/mattermost/mattermost-server/v5/model"
+)
+
+var SharedChannelEventsForSync model.StringArray = []string{
+	model.WEBSOCKET_EVENT_POSTED,
```
nronas, line 12: Not sure if we need to handle ephemeral events as well.

wiggin77: I can't think any reason why we would need to.

app/shared_channel_notifier.go
```diff
@@ -0,0 +1,73 @@
...
+	model.WEBSOCKET_EVENT_POST_EDITED,
+	model.WEBSOCKET_EVENT_POST_DELETED,
+	model.WEBSOCKET_EVENT_REACTION_ADDED,
+	model.WEBSOCKET_EVENT_REACTION_REMOVED,
+}
+
+// NotifySharedChannelSync signals the syncService to start syncing shared channel updates
+func (a *App) NotifySharedChannelSync(channel *model.Channel, event string) {
+	if !SharedChannelEventsForSync.Contains(event) || !channel.IsShared() {
```
wiggin77, line 21: Minor optimization... since most channels are not shared, moving `!channel.IsShared()` to be the first expression will avoid walking the string array for most messages/reactions.

app/server.go
wiggin77: Within the shared channel service package an IFace for the service can be created.  You can auto-generate a mock for that Iface there, or just manually create a simple one.  There is already a mock for server in `services/remotecluster/mock_test.go` that can return your service mock.  That would keep with the idiom of returning concrete types while accepting interfaces, and reduce proliferation of interfaces between packages.

That said, we're under time crunch so I'll accept as is. Thanks.

wiggin77 (approved):
Looks great.  One minor optimization and good to go. :+1:

app/shared_channel_notifier.go
```diff
@@ -0,0 +1,73 @@
...
+			mlog.String("event", event.EventType()),
+		)
+		return
+	}
+
+	if !SharedChannelEventsForSync.Contains(event.EventType()) {
+		mlog.Debug(
+			"Received websocket message that is not eligible to trigger shared channel sync, skipping...",
+			mlog.String("event", event.EventType()),
+		)
```
agnivade, line 56: I think you are missing a return statement here. And also, IMO, this log line might become spammy. There are just too many messages that will get filtered out. Perhaps best to just remove this line.

app/shared_channel_notifier.go
```diff
@@ -0,0 +1,73 @@
...
+	syncService.NotifyChannelChanged(channel.Id)
+}
+
+// ServerSyncSharedChannelHandler is called when a websocket event is received by a cluster node.
+// Only on the leader node it will notify the sync service to start syncing content for the given
+// shared channel.
+func (a *App) ServerSyncSharedChannelHandler(event *model.WebSocketEvent) {
+	syncService := a.srv.GetSharedChannelSyncService()
+	if syncService == nil || !syncService.Active() {
+		mlog.Debug(
```
agnivade, line 45: IMO we can remove this too. This will dump log lines for all non-leader nodes.

app/post.go
agnivade: I don't believe this is actually needed. This will actually generate duplicate notifications. You would only need to call `ServerSyncSharedChannelHandler` at a single place - `PublishSkipClusterSend`. All events are routed through this method. There's no need to call `NotifySharedChannelSync` here or anywhere else.

In short, this is how the cluster publishing happens:
- Everything starts off with the `(a *App) Publish` call.
- Now this call splits into two - `s.PublishSkipClusterSend` and `s.Cluster.SendClusterMessage`.
- The `PublishSkipClusterSend` handles propagating of events for the local machine itself. And `s.Cluster.SendClusterMessage` sends it off to the cluster.
- Now in the cluster handler, everything is routed through `PublishSkipClusterSend` again.
- So therefore, _all_ events, irrespective of being generated locally, or coming from another node, will be routed through `PublishSkipClusterSend`.
- So if we call `a.ServerSyncSharedChannelHandler(event)` inside `PublishSkipClusterSend`, we will automatically get free handling of all events. All we need to do is filter the right event, and check if it's a leader or not, which we are already doing.

agnivade (commented):
I think the event handler placement can be improved a bit. Right now, there's a lot of duplication and manual work to try and find out the places where to call the handler.

app/slashcommands/command_share.go
```diff
@@ -52,14 +52,11 @@ func (sp *ShareProvider) GetCommand(a *app.App, T goi18n.TranslateFunc) *model.C
 
 	status := model.NewAutocompleteData("status", "", "Displays status for this shared channel")
 
-	sync := model.NewAutocompleteData("sync", "", "Forces a sync with remote instances this shared channel")
```
agnivade, line 55: Not sure if this is intentional, or a bad merge?

agnivade (approved):
Nicely done!

app/slashcommands/command_share.go
nronas: My understanding is that the manual sync command now is not needed, since we are now automatically handling that. I can put it back, will defer to @wiggin77 on this

nronas: Discussed this during planning and the decision was to put the manual one back for an escape hatch when things fail to work.

PR #16690, opened 2021-01-13
MM-29605 Do not allow system managers to create S3 buckets

maxerenberg:
#### Summary
This pull request prevents the `/file/s3_test` endpoint from creating new S3 buckets.

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-29605

#### Release Note
```release-note
Previously, testing an S3 connection would create a new bucket if the bucket name specified in the
request did not exist. Now, the connection test will simply fail.
```


services/filesstore/s3store.go
```diff
@@ -142,11 +142,8 @@ func (b *S3FileBackend) TestConnection() error {
 	if exists {
 		mlog.Debug("Connection to S3 or minio is good. Bucket exists.")
 	} else {
-		mlog.Warn("Bucket specified does not exist. Attempting to create...")
-		err := b.client.MakeBucket(context.Background(), b.bucket, s3.MakeBucketOptions{Region: b.region})
-		if err != nil {
-			return errors.Wrap(err, "unable to create the s3 bucket")
-		}
+		mlog.Warn("Bucket specified does not exist.")
```
agnivade, line 145: This changes the behavior at a global level than just for system managers. The testConnection also happens when the server boots up, and it creates a bucket if that does not exist. That is intentional behavior and the permissions are controlled by the s3 user.

I think what we want here is to not create the bucket _only if_ the user is not a sysadmin or it's not made via an API call.

maxerenberg: If that's the case, then I think we should be explicitly creating a bucket (e.g. via a CreateBucket method) during startup. A function called "TestConnection" should really just be testing a connection. Similarly, an API endpoint titled "Test AWS S3 connection" (according to our API reference) should only be testing a connection. If we need to create a new bucket, we should explicitly state so. Our API reference does not mention the fact that a new bucket will be created if it does not exist.

If a system admin wants to use a new S3 bucket via the REST API, they can do so via the `/config/patch` endpoint.

agnivade: I like that.

We can let TestConnection return the status of the exists variable (or maybe have an `ErrNoBucket` error). And then on startup, create the bucket if that's false. And on the API call, just ignore the variable.

agnivade:
Just confirming- did you want for me to do a code review? It does not look like the code contains the changes that we discussed. :thinking:

maxerenberg:
My bad. I re-requested your review too early. The commit I just pushed should have the changes now.

services/filesstore/s3store.go
```diff
@@ -142,16 +144,21 @@ func (b *S3FileBackend) TestConnection() error {
 	if exists {
 		mlog.Debug("Connection to S3 or minio is good. Bucket exists.")
```
agnivade, line 145: This is out of scope for this PR. But we should move these log lines from this package to the app layer. There is similar logging happening with the localstore also.

services/filesstore/s3store.go
```diff
@@ -142,16 +144,21 @@ func (b *S3FileBackend) TestConnection() error {
...
 	} else {
-		mlog.Warn("Bucket specified does not exist. Attempting to create...")
-		err := b.client.MakeBucket(context.Background(), b.bucket, s3.MakeBucketOptions{Region: b.region})
-		if err != nil {
-			return errors.Wrap(err, "unable to create the s3 bucket")
-		}
+		mlog.Warn("Bucket specified does not exist.")
+		return ErrNoS3Bucket
 	}
```
agnivade, line 149: I am wondering if we can simplify it to:

```go
if !exists {
  return ErrNoS3Bucket
}
mlog.Debug("Connection to S3 or minio is good. Bucket exists.")
return nil
```

services/filesstore/s3store.go
```diff
@@ -41,6 +41,8 @@ const (
 	bucketNotFound = "NoSuchBucket"
 )
 
+var ErrNoS3Bucket = errors.New(bucketNotFound)
```
agnivade, line 44: The constant string was an error code. We usually have sentences for error strings - "bucket not found"

app/server.go
```diff
@@ -483,8 +483,15 @@ func NewServer(options ...Option) (*Server, error) {
...
 		nErr := backend.TestConnection()
 		if nErr != nil {
+			if errors.Is(nErr, filesstore.ErrNoS3Bucket) {
+				nErr = backend.(*filesstore.S3FileBackend).MakeBucket()
+				if nErr == nil {
+					goto backendTestSucceeded
+				}
+			}
 			mlog.Error("Problem with file storage settings", mlog.Err(nErr))
 		}
```
agnivade, line 493: The goto statement feels a bit unnecessary. We do run into some extra indentation, but it still keeps the code readable IMO.

```go
nErr := backend.TestConnection()
if nErr != nil {
  if errors.Is(nErr, filesstore.ErrNoS3Bucket) {
    mlog.Warn("Bucket specified does not exist. Attempting to create...")
    nErr = backend.(*filesstore.S3FileBackend).MakeBucket()
  }
  if nErr != nil {
    mlog.Error("Problem with file storage settings", mlog.Err(nErr))
  }
}
```

agnivade (commented):
Looking great! Just some small suggestions.

services/filesstore/s3store.go
maxerenberg: I feel like backend-specific log messages should be placed close to the relevant code though?

agnivade (approved):
Looks great now!

services/filesstore/s3store.go
agnivade: Logging is typically a concern of the higher level layer of an application. Low level packages should just be returning errors to indicate success or failures along with their context. That's why libraries rarely do logging on their own unless there are asynchronous tasks being performed somewhere else.

We don't need to do it in this PR. It was just a mental note.

mkraft (approved):

jupenur:
Should this change also address the equivalent behavior (directory creation) in `localstore.go`? Or will that be addressed separately?

maxerenberg:
> Should this change also address the equivalent behavior (directory creation) in `localstore.go`? Or will that be addressed separately?

I believe we should also implement the same behaviour in `localstore.go`. That'll probably have to go in a separate PR though.

jupenur (approved):

furqanmlk:
image

@maxerenberg is updating it

furqanmlk (approved):

maxerenberg:
I deleted my fork last week and I didn't realize that this PR was attached to it. Unfortunately I'm going to have to close this PR and re-open it upstream so that I can resolve the merge conflicts.

PR #16692, opened 2021-01-14
MM-30314 Don't delete all SidebarChannels when removed from team

### LINKED ISSUE #16517: Categories lose child items periodically
I confirm (by marking "x" in the [ ] below: [x]):

- [x ] This is not a troubleshooting question. [Troubleshooting questions go here: http://www.mattermost.org/troubleshoot/](http://www.mattermost.org/troubleshoot/).
- [ ] This doesn't reproduce on web browsers (such as in Chrome). If it does, [issue reports go to the Mattermost Server repository](https://github.com/mattermost/platform/issues).
- [x ] I have read [contributing guidelines](https://github.com/mattermost/desktop/blob/master/CONTRIBUTING.md).

---

**Summary**

After an indeterminate amount of time, Categories lose child channels and direct messages. 

**Environment**
- Operating System: Windows 10
- Mattermost Desktop App version: 4.6.1 
- Mattermost Server version: 5.28.1 

**Steps to reproduce**
Organize groups and direct messages into categories.  
Use as normal.  After some period of time, channels and direct messages within the category will return back to the Channels and Direct Messages group

**Expected behavior**
Items within Categories retain/persist indefinitely 

**Observed behavior**

More information about environment.  Our company is using a branded enterprise version. The webapp is older than the desktop application vs the Desktop version that I downloaded from Mattermost. 

The vanishing children (channels and direct messages) is intermittent.  I've seen the categories retain items through a reboot.  Not all categories lose their children at the same time.  It might be a specific amount of time since the category was created.  The category name persists, but all items within the category return back to their initial place.  I can't give a specific time, however, I noticed that category that I created yesterday has just lost its children.  It could be around 24 hours time. 

**Possible fixes**

hmhealey:
This fixes a strange bug where sometimes a user's custom categories would lose all the channels in them.

It turns out that this happens because of the code that removes the sidebar data when a user leaves or is removed from a team while they have a DM channel in a custom category on that team. This also only affected Postgres, so we weren't going to find it on community even if we did join/leave teams more often.

The reason this happened is because of the DELETE FROM query used to delete the SIdebarChannels, although I'm honestly not sure why. If I ran that query myself, it would reproduce the issue, but if I changed it to a SELECT with the same JOIN and WHERE, it only returned rows on the team that the user would've been leaving. I'm not entirely sure it has to do with a DM channel in particular, but that was the simplest repro steps I could find.

Since I didn't understand exactly why that caused a problem, I rewrote those queries in the most naive way I could think of. This isn't a hot API route, so I think we can afford the extra DB hit to be extra clear about what we're doing.

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-30314
Fixes #16517 

#### Release Note
```release-note
Fixed bug where channels would sometimes be removed from custom categories when a user leaves a team
```

### DIFF
--- a/store/sqlstore/channel_store_categories.go
+++ b/store/sqlstore/channel_store_categories.go
@@ -953,23 +953,50 @@ func (s SqlChannelStore) UpdateSidebarChannelCategoryOnMove(channel *model.Chann
 
 func (s SqlChannelStore) ClearSidebarOnTeamLeave(userId, teamId string) error {
 	// if user leaves the team, clean his team related entries in sidebar channels and categories
-	params := map[string]interface{}{
-		"UserId": userId,
-		"TeamId": teamId,
+	transaction, err := s.GetMaster().Begin()
+	if err != nil {
+		return errors.Wrap(err, "begin_transaction")
+	}
+	defer finalizeTransaction(transaction)
+
+	s.GetMaster().TraceOn("HARRISON", &TraceOnAdapter{})
+	defer s.GetMaster().TraceOff()
+
+	selectQuery, selectParams, _ := s.getQueryBuilder().
+		Select("Id").
+		From("SidebarCategories").
+		Where(sq.Eq{
+			"TeamId": teamId,
+			"UserId": userId,
+		}).ToSql()
+
+	var categoryIds []string
+	if _, err := transaction.Select(&categoryIds, selectQuery, selectParams...); err != nil {
+		return errors.Wrap(err, "Failed to get sidebar category IDs for deletion")
 	}
 
-	var deleteQuery string
-	if s.DriverName() == model.DATABASE_DRIVER_MYSQL {
-		deleteQuery = "DELETE SidebarChannels FROM SidebarChannels LEFT JOIN SidebarCategories ON SidebarCategories.Id = SidebarChannels.CategoryId WHERE SidebarCategories.TeamId=:TeamId AND SidebarCategories.UserId=:UserId"
-	} else {
-		deleteQuery = "DELETE FROM SidebarChannels USING SidebarChannels AS chan LEFT OUTER JOIN SidebarCategories AS cat ON cat.Id = chan.CategoryId WHERE cat.UserId = :UserId AND   cat.TeamId = :TeamId"
+	if len(categoryIds) > 0 {
+		deleteCategoryQuery, deleteCategoryParams, _ := s.getQueryBuilder().
+			Delete("SidebarCategories").
+			Where(sq.Eq{"Id": categoryIds}).ToSql()
+
+		if _, err := transaction.Exec(deleteCategoryQuery, deleteCategoryParams...); err != nil {
+			return errors.Wrap(err, "Failed to delete sidebar categories")
+		}
+
+		deleteChannelsQuery, deleteChannelsParams, _ := s.getQueryBuilder().
+			Delete("SidebarChannels").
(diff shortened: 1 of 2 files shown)


agnivade:
This looks like a problem with the query itself. We shouldn't need to run multiple queries. Let me take a look.

agnivade:
The issue is with how we are using the `USING` statement. This query:

```sql
 DELETE FROM SidebarChannels USING SidebarChannels AS chan LEFT OUTER JOIN SidebarCategories AS cat ON cat.Id = chan.CategoryId WHERE cat.UserId = '4fwk696m7bfbdgwpwro7kp8ihr' AND   cat.TeamId = 'q9caeua1tpyjuqsx5gqmceansy';
```

repeats the source table again in the USING clause. This causes a self-join with the table itself and will automatically match everything. As mentioned here: https://www.postgresql.org/docs/10/sql-delete.html

> Do not repeat the target table as a from_item unless you wish to set up a self-join (in which case it must appear with an alias in the from_item).

The USING syntax is actually not a standard syntax as recommended here:

>  This syntax is not standard. A more standard way to do it is:

If we want to keep using `USING`, this would be the query:

```sql
DELETE FROM SidebarChannels chan USING SidebarCategories cat WHERE cat.Id = chan.CategoryId AND cat.UserId = '4fwk696m7bfbdgwpwro7kp8ihr' AND   cat.TeamId = 'q9caeua1tpyjuqsx5gqmceansy';
```

Alternatively, this is what feels simplest and most readable to me, although being a bit more verbose:

```sql
delete from sidebarchannels where categoryid in (select categoryid from sidebarchannels chan, sidebarcategories cat where chan.categoryid=cat.id and cat.userid='4fwk696m7bfbdgwpwro7kp8ihr' and cat.teamid='q9caeua1tpyjuqsx5gqmceansy');
```

I'll let you decide which to choose. :)

agnivade:
I also think we should backport this. Leaving team is a pretty common activity.

reflog (approved):

hmhealey:
I think I'll go with the DELETE with the nested SELECT since having it all in one query does make it more atomic and reduce the chance for more deadlocking like that one case we still haven't reproduced. I know when writing some of this, I was trying to avoid nesting queries, but we might have gone too far with that.

Also, I just realized I left some debug code there that needs to be removed.

hmhealey:
@agnivade The delete from with a select nested in it doesn't work on MySQL by the looks of it since it doesn't want to . Since this was to fix a Postgres-specific issue, I'd rather keep both DBs using the same query going forward, so I'm going to stick with what I originally pushed.

As for backporting, I can ask Eric if he thinks we should do that, but I'm not super convinced this needs to be done since the sidebar is still experimental at this point.

agnivade:
Since we already were using 2 separate queries, I think it should be fine to keep using that. I don't see a huge benefit in using 2 queries just so that we can remove an if condition. The PR essentially becomes a one-line change (except the tests ofcourse). And as you said, longer transactions create more chances for deadlocks. So it seems beneficial to keep using a single query.

Agree on the backport.

agnivade (approved):
Thanks!

jgilliam17 (approved):
Thank you for adding test steps @hmhealey 
Tested, looks good to merge.
- Verified custom categories containing DMs/GMs remain populated on one team when user is removed from a different team.

amyblais:
@hmhealey We have some candidates for v5.31 ESR backport lined up, please help cherry-pick this when you can.

PR #16701, opened 2021-01-15
MM-31993: Add flag to disable gossip compression

agnivade:
Load tests have shown that our workload is not very suitable
to LZW compression. And in fact, compressing leads to more
network bandwidth than less. So we are spending more CPU cycles,
and creating more traffic, leading to a lose-lose situation.

We add a flag to control this behavior. Ideally, this should not
be a flag in the first place, since there is never a need to enable this
because clearly there is no benefit.

But to keep our community servers working, we need to be able
to configure this.

https://mattermost.atlassian.net/browse/MM-31993

```release-notes
Add a flag to disable compression in the Gossip protocol.

By default the value of the flag is true, which is the existing
default. This is done to maintain compatibility with old servers in a cluster.

Once all servers in a cluster are upgraded, it is recommended to disable
this flag for better performance.
```

Load-test results:
network-rx
network-tx
p99-store-times
p99-api-times
cpu-utilization-app (1)

Grafana:
CPU (1)
NW (1)


agnivade:
@wiersgallak - I'd need some guidance from you regarding the default value of the flag.

The issue is that if we keep the default value of the flag to be false. Then there will be incompatibility between servers in a cluster during an upgrade scenario. Because older servers will be true, but newer servers will be false. The solution would be to ask customers to completely shutdown a cluster, and then upgrade. This is exactly what we did when we released gossip encryption, but here things will break silently, so it's slightly more dangerous.

On the other hand, if we go with the default value to be true. Then we'd need to _strongly_ suggest in the release notes to disable compression, because it negatively affects cluster performance. This might start to become more prominent down the line when we make gossip to be the only option.

This looks to be a safer option. But I'm leaning towards you for any advice that you might have.

There is actually no need to keep this as a configurable value in the first place. But only to make our community servers running, we need to add a flag here.

nronas (commented):
This LGTM. Just to double-check my understanding here, this is currently enabled for everyone, but because it has a negative perf footprint we want to allow customers to have the ability to disable it(after a proper cluster drain/halt) in order to get the perf benefits of not compressing. Essentially, we shouldn't have compression on in the first place right?

If the above is correct then I would defo vote for the latter of the two options(enabled by default) but enrich the text on the configuration toggle to highlight the impact and importance of this toggle.

agnivade:
Yeah, I think I agree with you. Just wanted to lay out all options on the table.

nronas (approved):
LGTM

grundleborg (commented):
Deferring my review until Katie has considered the admin experience of this flow. But no objection to the flag in principle.

agnivade:
Gentle ping on this @wiersgallak

wiersgallak:
@agnivade I agree, it seems best to set to True and add release notes for the customer.  I also like John's suggestion of augmenting the text on the configuration setting as well.

agnivade:
Great, thanks! I will make the changes accordingly.

agnivade:
Gentle ping @grundleborg, @wiersgallak

grundleborg (approved):

wiersgallak (approved):

agnivade:
@lindalumitchell - This PR adds a new admin console setting. There are Cypress and unit tests for that. Nothing more than that to verify. I'll leave it to you if you want to defer QA.

saturninoabril:
@agnivade fyi, LM is OoO. I can chime in for a review if you don't mind.

saturninoabril (approved):
Thanks @agnivade! Agree, deferring QA review with impressive load test results and analysis.

PR #16702, opened 2021-01-15
Add make command for running the app in an HA topology

nronas:
#### Summary
While testing shared channels, it came to me that we do have features that rely on an HA topology but we do not seem to have an easy way of establishing that topology in dev. Added a make command separate from the canonical `run-server` for doing exactly that

### Usage

`$ make run-haserver` starts Mattermost with 2 server containers load balanced via round-robin through nginx.
`$ make restart-haserver` since docker-compose services have dynamic IPs in order for the server to pick up new changes and the haproxy to know where the new services live.

Tearing down should be handled by the existing `make stop`, since for an HA scenario we rely fully on docker-compose, in which `make stop` will call `docker-compose stop`

#### Ticket Link
N/A

#### Release Note
```release-note
NONE
```

## End Result
Screenshot 2021-01-18 at 3 11 48 PM


docker-compose.yaml
```diff
@@ -74,6 +74,69 @@ services:
...
+    healthcheck:
+      test: ["CMD", "curl", "-f", "http://leader:8065/api/v4/system/ping"]
+      interval: 5s
+      timeout: 10s
+      retries: 30
+    command: ['make', 'run-server-no-docker']
+    expose:
+      - "8065"
+
+  follower:
```
nronas, line 102: This is identical to the `leader` service, but cannot use `--scale service=2` because upon initialization those two will compete for migrations.

nronas:
Maybe worth adding someone from the releases team to review this. I would like some points on who to ping here 🙏

nginx.default.conf
```diff
@@ -0,0 +1,35 @@
+upstream app_cluster {
```
agnivade, line 1: I'd suggest placing this under build/docker directory. It has the relevant assets for docker related stuff.

Makefile
```diff
@@ -415,6 +420,13 @@ debug-server: start-docker ## Compile and start server using delve.
 		-X github.com/mattermost/mattermost-server/v5/model.BuildHashEnterprise=$(BUILD_HASH_ENTERPRISE)\
 		-X github.com/mattermost/mattermost-server/v5/model.BuildEnterpriseReady=$(BUILD_ENTERPRISE_READY)'"
 
+run-server-no-docker: prepackaged-binaries validate-go-version
```
agnivade, line 423: Hehe, we actually don't need this. `run-server` does not start docker if you have the `MM_NO_DOCKER` env var set :P

Makefile
```diff
@@ -156,6 +156,11 @@ ifneq (,$(findstring openldap,$(ENABLED_DOCKER_SERVICES)))
 endif
 endif
 
+run-haserver: run-client
+	@echo Starting docker containers in HA formation
+
```
agnivade, line 161: We should add a check here for `ifeq ($(BUILD_ENTERPRISE_READY),true)`, and only run this in that case. Mainly to help contributors not run it accidentally and have it not work.

docker-compose.yaml
```diff
@@ -74,6 +74,69 @@ services:
       - elasticsearch
     command: postgres:5432 mysql:3306 minio:9000 inbucket:10080 openldap:389 elasticsearch:9200
 
+  leader:
+    build:
+      context: .
+      dockerfile: ./build/Dockerfile.buildenv
+    working_dir: '/home/mattermost-server'
+    environment:
+      - "MM_SQLSETTINGS_DATASOURCE=postgres://mmuser:mostest@postgres/mattermost_test?sslmode=disable\u0026connect_timeout=10"
```
agnivade, line 83: I am wondering if we can just skip this and rely on the config.json value? By default, it is this value, and should work out of the box. For Mattermost staff, they should know to revert back to the docker settings, when running the Make command.

docker-compose.yaml
agnivade: Hmm .. yeah good point. We should work on adding a DB lock and only then move towards making schema changes. Oh well, for another time I guess!

agnivade (commented):
This is fantastic stuff @nronas ! You are now officially booked for next week's R&D demo.

agnivade:
> Maybe worth adding someone from the releases team to review this. I would like some pointers on who to ping here pray

I think you meant the Cloud team. :wink:  The releases team only handles releases. I will ping @stylianosrigas / @gabrieljackson for some expert eyes on the changes.

nginx.default.conf
nronas: Thought about that, wasn't sure if the build directory was supposed to only hold non dev release stuff. WIll move there

Makefile
nronas: Duh, you are so right on this.

Makefile
nronas: totally

docker-compose.yaml
nronas: Not sure about this, since this supports both environments since the follower and leader services are only used in an HA scenario. So traditional `make run-server or make run` they should just work.

Makefile
agnivade: I think it should be `BUILD_ENTERPRISE_READY`. `BUILD_ENTERPRISE` is just an env var to control whether to build enterprise or not. The `BUILD_ENTERPRISE_READY` checks the existence of the `enterprise` directory which is more robust and used elsewhere in the Makefile.

nronas: did i paste the wrong thing?... I did. Sorry about that.

docker-compose.yaml
agnivade: Yep, I think it should work even without setting the env var. I'll leave it up to you. 0/5.

gabrieljackson: Agreed that it would be nice to rely on the standard server config file for this, but also 0/5.

gabrieljackson (approved):
This looks good to me. Nice addition!

Makefile
agnivade: Hmm .. I still don't see the changes though :thinking: Not sure if I am missing something.

nronas: haven't pushed them yet 😄 , battling with Cypress atm

agnivade: Ah I see. When you requested my review, I thought the changes were there. No worries, just give me a ping when ready.

wiggin77 (approved):
LGTM :+1:   Will be very useful.

PR #16746, opened 2021-01-18
MM-31596: Support permalink in shared channels

agnivade:
We match permalinks in posts that are to be sent in shared channels.
If the permalink is not for the shared channel, we send an ephemeral
post to the post author notifying them that the link may not be
available to all remote users.
    
On the receiving side, we replace the site URL in the post
with the local site's URL.
    
```release-note
NONE
```

https://mattermost.atlassian.net/browse/MM-31596


services/sharedchannel/permalink.go
```diff
@@ -0,0 +1,70 @@
...
+			return msg
+		}
+
+		// Otherwise, modify pl to plshared as a marker to be replaced by remote sites
+		return strings.Replace(msg, "/pl/", "/"+permalinkMarker+"/", 1)
+	})
+}
+
+// processPermalinkFromRemote processes all permalinks coming from a remote site.
+func (scs *Service) processPermalinkFromRemote(msg string) string {
```
agnivade, line 55: @wiggin77 - I assume I'll need to replace the remote team name with the local team name as well right?

services/sharedchannel/permalink.go
```diff
@@ -0,0 +1,70 @@
...
+
+const (
+	permalinkMarker = "plshared"
+)
+
+// processPermalinkToRemote processes all permalinks going towards a remote site.
+func (scs *Service) processPermalinkToRemote(p *model.Post) string {
+	return permaLinkRegex.ReplaceAllStringFunc(p.Message, func(msg string) string {
+		// Extract the postID (This is simple enough not to warrant full-blown URL parsing.)
+		lastSlash := strings.LastIndexByte(msg, '/')
```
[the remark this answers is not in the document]
agnivade: Yes, absolutely we can. And there is a test case for that too. Are you seeing any particular problem with this code?

services/sharedchannel/permalink.go
wiggin77: Correct.  RemoteTeamId is in the RemoteClusters table.

wiggin77: Ah, looking closer I see this part is executed on the remote side.  Ignore my previous comment.  Do team names appear in perma-links?

wiggin77: Ah they do. Yes, they would have to be replaced.

services/sharedchannel/permalink.go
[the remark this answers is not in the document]
nronas: i didn't see the test case in the first pass. But I can defo see it now, so all good

services/sharedchannel/permalink.go
agnivade: So to confirm, I should still take the team name from "RemoteTeamId is in the RemoteClusters table." ?

wiggin77: Right, local team name for channel.

services/sharedchannel/permalink.go
```diff
@@ -0,0 +1,75 @@
...
+		}
+		if len(postList.Order) == 0 {
+			scs.server.GetLogger().Log(mlog.LvlSharedChannelServiceWarn, "No post found for permalink", mlog.String("postID", postID))
+			return msg
+		}
+
+		// If postID is for a different channel
+		if postList.Posts[postList.Order[0]].ChannelId != p.ChannelId {
+			// Send ephemeral message to OP (only once per message).
+			// Note: Does this need to be translated?
```
wiggin77, line 45: There is a ticket to have all ephemeral posts in the slash commands translated.  This should be as well, but I'm 0/5 whether that happens before or after MVP.  A ticket will suffice if that helps.

agnivade: Did that in this ticket itself.

nronas (approved):
LGTM

wiggin77 (approved):
Looks great!  :+1:

PR #16748, opened 2021-01-18
MM-31711 - Implement cursor paging for threads

reflog:
#### Ticket Link

https://mattermost.atlassian.net/browse/MM-31711

#### Release Note

```release-note
 `/api/v4/users/{user_id}/teams/{team_id}/threads` now accepts "before" or "after" parameters instead of page index
```


api4/user.go
```diff
@@ -2844,18 +2845,15 @@ func getThreadsForUser(c *Context, w http.ResponseWriter, r *http.Request) {
...
-		if parseError != nil {
-			c.SetInvalidParam("page")
-			return
-		}
-		options.Page = page
+	options.Before = r.URL.Query().Get("before")
+	options.After = r.URL.Query().Get("after")
+	// parameters are mutually exclusive
+	if options.Before != "" && options.After != "" {
+		c.Err = model.NewAppError("MfaRequired", "api.getThreadsForUser.bad_params", nil, "", http.StatusBadRequest)
```
ashishbhate, line 2852: >`MfaRequired`

Should probably be something else.

ashishbhate (approved):

store/sqlstore/thread_store.go
```diff
@@ -182,7 +182,20 @@ func (s *SqlThreadStore) GetThreadsForUser(userId, teamId string, opts model.Get
 		if opts.Since > 0 {
 			newFetchConditions = sq.And{newFetchConditions, sq.GtOrEq{"Threads.LastReplyAt": opts.Since}}
```
calebroseland, line 183: (Same as other comment) Should this factor in `ThreadMemberships.LastEditAt`?

calebroseland (approved):

store/sqlstore/thread_store.go
reflog: @calebroseland - there's no LastEditAt in thread membership, so not sure what you meant :)

calebroseland: @reflog, doh! I meant `lastupdated` (on `threadmemberships` in DB).  `since` in my interpretation would include changes to `following` (and similar) for sync'ing cached UserThreads in the mobile app.

ashishbhate: yep, for mobile it would be good to have all changes since `since`, if possible.

PR #16753, opened 2021-01-19
MM-32033: Check if a remote is already added before inviting

agnivade:
We check the store if the remote already exists in the DB.
If it does, we return an error.

https://mattermost.atlassian.net/browse/MM-32033

```release-note
NONE
```

### DIFF
--- a/app/slashcommands/command_share.go
+++ b/app/slashcommands/command_share.go
@@ -240,6 +240,14 @@ func (sp *ShareProvider) doInviteRemote(a *app.App, args *model.CommandArgs, mar
 		return responsef("Must specify a valid remote cluster id to invite.")
 	}
 
+	all, err := a.GetSharedChannelRemotes(args.ChannelId)
+	if err != nil {
+		return responsef("Error fetching remote clusters: %v", err)
+	}
+	if len(all) > 0 {
+		return responsef("The remote cluster has already been invited")
+	}
+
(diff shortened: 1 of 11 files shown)


app/slashcommands/command_share.go
```diff
@@ -240,6 +240,14 @@ func (sp *ShareProvider) doInviteRemote(a *app.App, args *model.CommandArgs, mar
 		return responsef("Must specify a valid remote cluster id to invite.")
 	}
 
+	all, err := a.GetSharedChannelRemotes(args.ChannelId)
```
nronas, line 243: Not a blocker, 1/5, but this will fetch all the columns for a given channel when the channel exists in the `SharedChannelRemotes` table, I think a more appropriate query here would be an existence check like `select 1 as one from SharedChannelRemotes where ChannelId = x`

nronas (approved):
A non-blocking comment. but this LGTM.

app/slashcommands/command_share.go
agnivade: Agree. We will need to add a new method. Will do. :+1:

agnivade: Done.

agnivade:
@wiggin77 - Good to be reviewed now.

wiggin77 (approved):
Looks great!  :+1:

PR #16755, opened 2021-01-19
MM-32034: Auto share a channel if not shared

agnivade:
When inviting a remote, if the channel is not shared,
we automatically share it.

https://mattermost.atlassian.net/browse/MM-32034

```release-note
NONE
```


app/slashcommands/command_share.go
```diff
@@ -240,6 +241,15 @@ func (sp *ShareProvider) doInviteRemote(a *app.App, args *model.CommandArgs, mar
 		return responsef("Must specify a valid remote cluster id to invite.")
 	}
 
+	// Check if channel is shared or not.
+	if _, err := a.GetSharedChannel(args.ChannelId); err != nil {
```
nronas, line 245: Will be good to do an existence check here as well, since we explicitly ignoring the result of the query. But logic wise this LGTM

nronas (approved):
No blocking comment, this LGTM

app/slashcommands/command_share.go
agnivade: I think that should be a test case of the store function itself? Checking for `ErrNotFound` should automatically negate the existence of the channel.

nronas: I was referring to the underlying query. Since the existing query is doing a `select * from SharedChannels where id = x`. Similar to what we were discussing in a previous PR

agnivade: Oh my bad :facepalm: gotcha .. will look into it.

wiggin77: I'm 0/5 on whether the optimized existence check is needed, however I will say that inviting a remote cluster to a channel is a very rare event.

app/slashcommands/command_share.go
```diff
@@ -240,9 +240,19 @@ func (sp *ShareProvider) doInviteRemote(a *app.App, args *model.CommandArgs, mar
...
-	rc, err := a.GetRemoteCluster(remoteId)
+	// Check if channel is shared or not.
+	has, err := a.HasSharedChannel(args.ChannelId)
 	if err != nil {
-		return responsef("Remote cluster id is invalid: %v", err)
+		return responsef("Error while checking if shared channel exists: %v", err)
+	}
+	if !has {
+		// If it doesn't exist, then create it.
+		sp.doShareChannel(a, args, margs)
```
wiggin77, line 250: `doShareChannel` returns a string that becomes an ephemeral message notifying the user the channel has been shared, or any error.  Can we add that to the response returned below?  Just prepending it to the response below with a newline separator should be enough.

store/storetest/shared_channel_store.go
```diff
@@ -138,6 +140,34 @@ func testGetSharedChannel(t *testing.T, ss store.Store) {
 	})
 }
 
+func testExistsSharedChannel(t *testing.T, ss store.Store) {
```
wiggin77, line 143: Should be `testHasSharedChannel(...)` ?

wiggin77 (approved):
Looks great :+1: 

Prepending to named return value via defer... nice.

agnivade:
> Prepending to named return value via defer... nice.

I was ambivalent on whether to do that or not. It's a trick. Go doesn't like tricks. But the function was small enough not to get lost in the code flow.

PR #16763, opened 2021-01-20
MM-31721: Fix Cache-Control directives

agnivade:
We were incorrectly setting the cache control to public
when it should be rather private.

https://mattermost.atlassian.net/browse/MM-31721

```release-notes
Fix Cache-Control headers to instruct that responses may only be cached
on browsers.
```


srkgupta (approved):

mgdelacroix (approved):
LGTM! :+1:

PR #16769, opened 2021-01-21
MM-29584: Make apps plugin hook invocation/registration conditional

catalintomai:
#### Summary
Add a variant of the `ShouldProcessMessage` helper method that does not ry to retrieve/check for the presence of the `botId` in the KV store. The `ShouldProcessMessageNoBotCheck` is being used by the `mattermost-plugin-apps` proxy plugin to skip processing the `MessageHasBeenPosted` handler functionality (`AllowWebhook` is `false`)

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-29584

#### Related PRs
https://github.com/mattermost/mattermost-plugin-apps/pull/50

```release-note
NONE
```


levb (approved):
Do you think we could possibly make a `NoBotCheck` option for the original ShouldProcessMessage? (the option function could also accept the BotID to use in other checks, if needed)

Otherwise, LGTM

levb:
image
@hanzei would know how to fix

catalintomai:
> Do you think we could possibly make a `NoBotCheck` option for the original ShouldProcessMessage? (the option function could also accept the BotID to use in other checks, if needed)
> 
> Otherwise, LGTM

Adding the `NoBotCheck` option would be ok, but did not want to update the `ShouldProcessMessage` signature by adding another param - botId (I don't know what our upgrade API method story is), and adding the botId to the list of options doesn't fit (it is not a general set of options, they are called `ShouldProcessMessageOption`)

plugin/helpers.go
```diff
@@ -79,6 +79,19 @@ type Helpers interface {
 	// Minimum server version: 5.2
 	ShouldProcessMessage(post *model.Post, options ...ShouldProcessMessageOption) (bool, error)
```
hanzei, line 80: `ShouldProcessMessage` uses a function options. Hence, we can add a new option without changing the signature. What do you think about a `BotID` option?

hanzei (changes requested):

plugin/helpers.go
catalintomai: As I was mentioning in the previous comment - the options set is not general, it is called `ShouldProcessMessageOption` (and the current list of options: `AllowWebhook`, etc, support this naming), so BotID doesn't really fit (to me).

hanzei: Could you please elaborate why `BotID` isn't a fitting option to you? The whole idea of functional options is to support addtional options when they are needed without changing the signature.

catalintomai: @hqnzei - I guess I see BotID (and other `id`s) as essentially function params versus options (like a "should I do this or not" option). Wasn't aware that the primary intent of the Options set is solve the signature issue, don't really agree with that but I'll align to it since that is the existing pattern.

hanzei (approved):

levb (approved):

DHaussermann:
Hi @catalintomai  sorry for the delay on this. I can deploy it and briefly regression test. But seem I would want to do this along side the plugin change here. https://github.com/mattermost/mattermost-plugin-apps/pull/50

When I tried to pull master into https://github.com/mattermost/mattermost-plugin-apps/pull/50 in seems the CI checks are failing. Can you please take a look when you have a moment?

catalintomai:
@DHaussermann , the "mattermost-apps" CI tests are failing because of "BotID not declared by package plugin" error message, which basically says that the server code being referenced does not contain the `BodID` variable, which is expected, and will disappear once the server code is checked in. The way around it is to modify the `go.mod` file and make it reference the local server code by adding a statement smth. like:
`replace github.com/mattermost/mattermost-server/v5 v5.3.2-0.20210120031517-5a7759f4d63b => /Users/<username>/go/src/github.com/mattermost/mattermost-server`

DHaussermann (approved):
Tested and passed
- Regression tested user managed bots making posts 
- Regression tested plugin managed bots making posts  
- No new functional testing applicable here 

LGTM!

DHaussermann:
I did deploy this and run through regression tests. I was unable to successfully do the go.mod replacement as instructed above. I tried the replacement line a few different ways a continue getting parsing errors when it tries to import the module. 

@catalintomai, would it be reasonable to merge this wih the testing done so far and then have me test https://github.com/mattermost/mattermost-plugin-apps/pull/50 once this is merged?

catalintomai:
Thanks @DHaussermann , I usually add the replace statement at the end of the `go.mod` file - one thing that can be missed - the version like `v5.3.2-0.20210120031517-5a7759f4d63b` needs to be the same as the one in the `require` clause. I'll check-in the server part and then you should be able to validate the `apps` part.

catalintomai:
> Thanks @DHaussermann , I usually add the replace statement at the end of the `go.mod` file - one thing that can be missed - the version like `v5.3.2-0.20210120031517-5a7759f4d63b` needs to be the same as the one in the `require` clause. I'll check-in the server part and then you should be able to validate the `apps` part.

@DHaussermann , FYI, the server PR got merged.

PR #13654, opened 2020-01-20
MM-21674: Asynchronous database migrations

Pomyk:
#### Summary
Mechanism for asynchronous database migrations. Currently only create index and drop index implemented.

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-21674

PR #16777, opened 2021-01-22
Add the abillity to run db operations asyncronsously

nronas:
#### Summary
A vanilla implementation that will allow us to run database operations in an async way. Is not implementing migrations semantics(like up, down, run etc), but I think is good enough for async index creation on big tables, clean up tasks etc.

*NOTE* If direction wise this looks good, we will need to create a `create index concurrently` function that will allow concurrent index creation within the DB while being executed asynchronously from the server.

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-30186

#### Release Note
```release-note
Allows async database operations to be executed without interfering with server boot up.
```

### DIFF
--- a/model/system.go
+++ b/model/system.go
@@ -37,6 +37,7 @@ const (
 	USER_LIMIT_OVERAGE_CYCLE_END_DATE             = "UserLimitOverageCycleEndDate"
 	OVER_USER_LIMIT_FORGIVEN_COUNT                = "OverUserLimitForgivenCount"
 	OVER_USER_LIMIT_LAST_EMAIL_SENT               = "OverUserLimitLastEmailSent"
+	ASYNC_MIGRATION_RUNNER_ID                     = "async_migration_runner_id"
 )
 
 const (
--- a/store/sqlstore/store.go
+++ b/store/sqlstore/store.go
@@ -117,20 +117,24 @@ type SqlStoreStores struct {
 	linkMetadata         store.LinkMetadataStore
 }
 
+type AsyncMigration func()
+
 type SqlStore struct {
 	// rrCounter and srCounter should be kept first.
 	// See https://github.com/mattermost/mattermost-server/v5/pull/7281
-	rrCounter      int64
-	srCounter      int64
-	master         *gorp.DbMap
-	replicas       []*gorp.DbMap
-	searchReplicas []*gorp.DbMap
-	stores         SqlStoreStores
-	settings       *model.SqlSettings
-	lockedToMaster bool
-	context        context.Context
-	license        *model.License
-	licenseMutex   sync.RWMutex
+	rrCounter       int64
+	srCounter       int64
+	master          *gorp.DbMap
+	replicas        []*gorp.DbMap
+	searchReplicas  []*gorp.DbMap
+	stores          SqlStoreStores
+	settings        *model.SqlSettings
+	lockedToMaster  bool
+	context         context.Context
+	license         *model.License
+	licenseMutex    sync.RWMutex
+	asyncMigrations []AsyncMigration
+	wg              sync.WaitGroup
 }
 
 type TraceOnAdapter struct{}
@@ -145,9 +149,10 @@ func (t *TraceOnAdapter) Printf(format string, v ...interface{}) {
 
 func New(settings model.SqlSettings, metrics einterfaces.MetricsInterface) *SqlStore {
 	store := &SqlStore{
-		rrCounter: 0,
-		srCounter: 0,
-		settings:  &settings,
+		rrCounter:       0,
+		srCounter:       0,
+		settings:        &settings,
+		asyncMigrations: []AsyncMigration{},
 	}
 
 	store.initConnection()
@@ -233,7 +238,10 @@ func New(settings model.SqlSettings, metrics einterfaces.MetricsInterface) *SqlS
 	store.stores.linkMetadata.(*SqlLinkMetadataStore).createIndexesIfNotExists()
 	store.stores.group.(*SqlGroupStore).createIndexesIfNotExists()
 	store.stores.scheme.(*SqlSchemeStore).createIndexesIfNotExists()
-	store.stores.preference.(*SqlPreferenceStore).deleteUnusedFeatures()
+
+	store.asyncMigrations = append(store.asyncMigrations, store.stores.preference.(*SqlPreferenceStore).generateAsyncMigrations()...)
+
+	store.dispatchAsyncMigrations()
 
 	return store
 }
@@ -933,7 +941,6 @@ func (ss *SqlStore) createIndexIfNotExists(indexName string, tableName string, c
 			os.Exit(ExitCreateIndexPostgres)
 		}
 	} else if ss.DriverName() == model.DATABASE_DRIVER_MYSQL {
-
 		count, err := ss.GetMaster().SelectInt("SELECT COUNT(0) AS index_exists FROM information_schema.statistics WHERE TABLE_SCHEMA = DATABASE() and table_name = ? AND index_name = ?", tableName, indexName)
 		if err != nil {
 			mlog.Critical("Failed to check index", mlog.Err(err))
@@ -1070,6 +1077,10 @@ func (ss *SqlStore) RecycleDBConnections(d time.Duration) {
 }
 
 func (ss *SqlStore) Close() {
+	mlog.Debug("Waiting for active async migrations to finish...")
+	ss.wg.Wait()
+	mlog.Debug("Async migrations completed...")
+
 	ss.master.Db.Close()
 	for _, replica := range ss.replicas {
 		replica.Db.Close()
@@ -1242,6 +1253,16 @@ func (ss *SqlStore) UpdateLicense(license *model.License) {
 	ss.licenseMutex.Lock()
 	defer ss.licenseMutex.Unlock()
 	ss.license = license
+}
+
+func (ss *SqlStore) dispatchAsyncMigrations() {
+	ss.wg.Add(len(ss.asyncMigrations))
+	for _, migration := range ss.asyncMigrations {
+		go func(m func()) {
+			defer ss.wg.Done()
+			m()
+		}(migration)
+	}
 }
 
 type mattermConverter struct{}
--- a/store/sqlstore/store.go
+++ b/store/sqlstore/store.go
@@ -133,6 +133,7 @@ type SqlStore struct {
 	context         context.Context
 	license         *model.License
 	licenseMutex    sync.RWMutex
+	id              string
 	asyncMigrations []AsyncMigration
 	wg              sync.WaitGroup
 }
@@ -153,6 +154,7 @@ func New(settings model.SqlSettings, metrics einterfaces.MetricsInterface) *SqlS
 		srCounter:       0,
 		settings:        &settings,
 		asyncMigrations: []AsyncMigration{},
+		id:              model.NewId(),
 	}
 
 	store.initConnection()
@@ -241,7 +243,9 @@ func New(settings model.SqlSettings, metrics einterfaces.MetricsInterface) *SqlS
 
 	store.asyncMigrations = append(store.asyncMigrations, store.stores.preference.(*SqlPreferenceStore).generateAsyncMigrations()...)
 
-	store.dispatchAsyncMigrations()
+	if store.tryToDispatchAsyncMigrations() {
+		store.dispatchAsyncMigrations()
+	}
 
 	return store
 }
@@ -1081,6 +1085,16 @@ func (ss *SqlStore) Close() {
 	ss.wg.Wait()
 	mlog.Debug("Async migrations completed...")
 
+	if _, err := ss.System().UpsertWithCond(&model.System{
+		Name:  model.ASYNC_MIGRATION_RUNNER_ID,
+		Value: "",
+	}, map[string]string{
+		"Name":  model.ASYNC_MIGRATION_RUNNER_ID,
+		"Value": ss.id,
+	}); err != nil {
+		mlog.Warn("Failed updating system async migrations", mlog.Err(err))
+	}
+
 	ss.master.Db.Close()
 	for _, replica := range ss.replicas {
 		replica.Db.Close()
@@ -1263,6 +1277,28 @@ func (ss *SqlStore) dispatchAsyncMigrations() {
 			m()
 		}(migration)
 	}
+}
+
+func (ss *SqlStore) tryToDispatchAsyncMigrations() bool {
+	system, err := ss.System().UpsertWithCond(&model.System{
+		Name:  model.ASYNC_MIGRATION_RUNNER_ID,
+		Value: ss.id,
+	}, map[string]string{
+		"Name":  model.ASYNC_MIGRATION_RUNNER_ID,
+		"Value": "",
+	})
+
+	if err != nil {
+		mlog.Warn("Error while updating async migrations", mlog.Err(err))
+		return false
+	}
+
+	if system == nil {
+		mlog.Debug("Skipping async migration dispatching, another node has already started the migrations.")
+		return false
+	}
+
+	return true
 }
 
 type mattermConverter struct{}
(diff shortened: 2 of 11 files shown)


nronas:
@agnivade @grundleborg will love a 30% review on this. I've tried to keep it dead simple in order to not interfere with future plans that we have to implement migrations. This just allows running db operations in a supervised async way.

agnivade:
Are we sure spawning the migration in a goroutine is all we need? https://github.com/mattermost/mattermost-server/pull/13654 needed an atomic lock on the DB to prevent conflicting migrations in a cluster upgrade scenario.

https://github.com/mattermost/mattermost-server/pull/13654#issuecomment-579723098.

The core issue with that PR is that it's untested. So we would need to do some heavy testing with cluster upgrades, crashes, timeouts and ensure that everything works as expected.

nronas:
> Are we sure spawning the migration in a goroutine is all we need? #13654 needed an atomic lock on the DB to prevent conflicting migrations in a cluster upgrade scenario.

I guess I can check if cluster is enabled and run them only on the leader. That will guarantee, I think atomicity.

> The core issue with that PR is that it's untested. So we would need to do some heavy testing with cluster upgrades, crashes, timeouts and ensure that everything works as expected.

Yeap yeap that is still a 30% ready.

nronas:
Was thinking if we can push a worker job and that to run the migrations, since we have atomicity when picking up jobs(if we don't then I would be surprised 😄 )

agnivade:
The worker job approach sounds perfect for this. Completely asynchronous and well-tested, and works independently of leader or not.

Re: running on leader, I think that was discussed too on that PR. But then I think we somehow needed a DB lock to counter for some other cases. George/Jesse might remember.

grundleborg:
> The worker job approach sounds perfect for this. Completely asynchronous and well-tested, and works independently of leader or not.

The workers stuff does guarantee scheduled jobs will be run only once, but the scheduling doesn't have any guarantee of atomicity, so you need to do something to make sure a sudden change of cluster leader might result in it scheduling the same job twice?

> Re: running on leader, I think that was discussed too on that PR. But then I think we somehow needed a DB lock to counter for some other cases. George/Jesse might remember.

I don't remember the specific issues, but I do recall there being some concerns that just using the cluster leader wasn't atomic enough.

That said, can't we just use some optimistic locking to get the "exactly once" semantics (either by adding a "schedule this exactly once" primitive to the jobs scheduler/worker framework, or just specifically for these db migrations)?

Friday evening is the wrong time for me to be thinking about these kind of things, so sorry for the incoherent braindump. Hopefully you guys can make something useful out of it.

nronas:
@agnivade @grundleborg that's some wonderful context, thank you. I'll dig deeper into the use cases that we have and we can take it from there. I need to double-check the clustering logic in order to be sure, but the `isLeader` will not suffice because leader re-election does not necessarily mean that the old leader has been torn down successfully(which would have covered the race, since on teardown we are waiting for the migrations to finish), i think the clustering logic is using a heartbeat in order to determine if the leader is healthy.

lieut-data:
I think @grundleborg and @agnivade have already hit the highlights, but just wanted to re-emphasize something I recall from the original discussion. In general, most schema changes must be synchronous, since we can't actually start running the new code until the changes have been made (think new tables, columns, etc.)

The two use for asynchronous migrations that come to mind are 1) data migration, and 2) index creation. I think we have an established paradigm for data migration already, which amounts to kicking off a one-time job and using the database row as a mutex to avoid concurrent execution. Some of the discussion in https://github.com/mattermost/mattermost-server/pull/13654 focused on commoditizing that functionality. It wasn't clear to me that we had enough use cases to justify the overall effort, but I'm not opposed to improving this.

With index creation, if we already have `CONCURRENTLY`, do we need asynchronous invocation? (I'm using the force, since it's been a while, but I thought `CREATE INDEX CONCURRENTLY ...` returned immediately anyway). @agnivade's already summarized the potential pain point of trying to do this twice from two different servers, but I /suspect/ the database will just reject the second request since it "already exists."

Lastly, I wanted to highlight that, traditionally, we have relied on a /functioning/ index in the newly deployed code. I'd be worried about temporary performance issues from trying to use a feature without the index having finished. (Think full table scans, etc.)

</end_of_week_braindump>

agnivade:
`CREATE INDEX CONCURRENTLY` does not return immediately. You _could_ run it asynchronously yourself, but that's a different matter.

IMO the main use-case that this unlocks is the ability for us to freely upgrade/drop indexes on large tables (mainly `Posts`). As of now, the most blocking thing is the improvement to message export which cannot be done until we add a new index on Posts. But we also wanted to remove some redundant indexes. We weren't able to even start removing/optimizing indexes because of this problem. And now that Postgres version is bumped, this allows us to use even more advanced indexing approaches.

All of these improve on existing indexes, and does not touch the query at all. So there shouldn't be any temporary performance issues.

I am 0/5 on the implementation, but 4/5 that we will definitely need some manner of asynchronous DB operations.

nronas:
https://github.com/mattermost/mattermost-server/pull/16777/commits/d1073546283486d81e4e58f5b92a6be122cd0b68 tried to synchronize different runners on a new system row, that indicates which node is the active runner. Have in mind that the synchronization is efficient enough and does not wait for the migration to finish, it just checks the value of the row with a `select ... for update`. There is the possibility that two nodes can run the migrations if the `model.NewId()` returns the same id for two runners, I would rank that as a scenario that we do not want to cover since we use that method to generate pkeys in our tables.

store/sqlstore/system_store.go
```diff
@@ -116,7 +116,7 @@ func (s SqlSystemStore) PermanentDeleteByName(name string) (*model.System, error
 
 // InsertIfExists inserts a given system value if it does not already exist. If a value
```
agnivade, line 117: Also change the comment here.

store/sqlstore/system_store.go
```diff
@@ -147,3 +147,36 @@ func (s SqlSystemStore) InsertIfExists(system *model.System) (*model.System, err
 	}
 	return system, nil
 }
+
+// UpsertWithCond selects for update a system record with the provided condArgs and updates it's value
+// from the given system arg.
+func (s SqlSystemStore) UpsertWithCond(system *model.System, condArgs map[string]string) (*model.System, error) {
+	tx, err := s.GetMaster().BeginTx(context.Background(), &sql.TxOptions{
+		Isolation: sql.LevelReadCommitted,
```
agnivade, line 155: Is this just to capture the additional gap lock? Since Id will anyways be unique, wondering if we need this?

nronas: I think we need this, for Mysql versions that have auto commit enabled.

agnivade:
Looks good wrt atomicity.

I think another case that we need to handle is what if a server is killed halfway an index migration is happening. If we use the jobs framework, on restart, we can detect that a job was in-progress, but not finished. But now we need to decide how to restart the job. All index migrations may not be completely idempotent (or maybe they are?). If that's not the case, we might need to add additional commands to drop the index again, or issue the reindex command, or do something else.

nronas:
> I think another case that we need to handle is what if a server is killed halfway an index migration is happening. If we use the jobs framework, on restart, we can detect that a job was in-progress, but not finished. But now we need to decide how to restart the job. All index migrations may not be completely idempotent (or maybe they are?). If that's not the case, we might need to add additional commands to drop the index again, or issue the reindex command, or do something else.

Yeap i thought about that, but my understanding is that if we kill the routine midway through it will rollback. So i think we have idempotency.

model/system.go
```diff
@@ -37,6 +37,7 @@ const (
 	USER_LIMIT_OVERAGE_CYCLE_END_DATE             = "UserLimitOverageCycleEndDate"
 	OVER_USER_LIMIT_FORGIVEN_COUNT                = "OverUserLimitForgivenCount"
 	OVER_USER_LIMIT_LAST_EMAIL_SENT               = "OverUserLimitLastEmailSent"
+	ASYNC_MIGRATION_RUNNER_ID                     = "async_migration_runner_id"
```
lieut-data, line 40: nit: do we need to export this in the "public" `model` package?

store/sqlstore/store.go
```diff
@@ -233,7 +240,12 @@ func New(settings model.SqlSettings, metrics einterfaces.MetricsInterface) *SqlS
 	store.stores.linkMetadata.(*SqlLinkMetadataStore).createIndexesIfNotExists()
 	store.stores.group.(*SqlGroupStore).createIndexesIfNotExists()
 	store.stores.scheme.(*SqlSchemeStore).createIndexesIfNotExists()
-	store.stores.preference.(*SqlPreferenceStore).deleteUnusedFeatures()
+
+	store.asyncMigrations = append(store.asyncMigrations, store.stores.preference.(*SqlPreferenceStore).generateAsyncMigrations()...)
```
lieut-data, line 244: Wondering if it makes sense to keep the the migrations high-level vs. anchoring them to a single store? This makes sense for things like indices, but I suspect some of the data migrations we've already written (and ones to come) will span stores, and it might be awkward to have them scattered.

store/sqlstore/store.go
```diff
@@ -1070,6 +1081,20 @@ func (ss *SqlStore) RecycleDBConnections(d time.Duration) {
 }
 
 func (ss *SqlStore) Close() {
+	mlog.Debug("Waiting for active async migrations to finish...")
+	ss.wg.Wait()
```
lieut-data, line 1085: I think `systemd` will kill a process if it doesn't shut down in a timely fashion. Are we assuming the server stays up long enough to finish the upgrade? Might this be problematic in a cloud environment with pods that are expected to get killed more frequently?

store/sqlstore/store.go
```diff
@@ -233,7 +240,12 @@ func New(settings model.SqlSettings, metrics einterfaces.MetricsInterface) *SqlS
 	store.stores.linkMetadata.(*SqlLinkMetadataStore).createIndexesIfNotExists()
 	store.stores.group.(*SqlGroupStore).createIndexesIfNotExists()
 	store.stores.scheme.(*SqlSchemeStore).createIndexesIfNotExists()
-	store.stores.preference.(*SqlPreferenceStore).deleteUnusedFeatures()
+
+	store.asyncMigrations = append(store.asyncMigrations, store.stores.preference.(*SqlPreferenceStore).generateAsyncMigrations()...)
+
+	if store.tryToDispatchAsyncMigrations() {
```
lieut-data, line 246: This pattern seems weird to me, followed by our having to clean up in `Close`. Should we just `tryToDispatchAsyncMigrations()` and have it contain the entirety of any setup / teardown?

store/store.go
```diff
@@ -461,8 +461,9 @@ type SystemStore interface {
 	Get() (model.StringMap, error)
 	GetByName(name string) (*model.System, error)
 	PermanentDeleteByName(name string) (*model.System, error)
-	InsertIfExists(system *model.System) (*model.System, error)
+	InsertIfNotExists(system *model.System) (*model.System, error)
 	SaveOrUpdateWithWarnMetricHandling(system *model.System) error
+	UpsertWithCond(system *model.System, condArgs map[string]string) (*model.System, error)
```
lieut-data, line 466: This interface seems a bit low-level for the caller: can we just pass an `oldValue string` instead?

store/sqlstore/system_store.go
lieut-data: I'm rusty on this: what's the fundamental difference between an explicit transaction and a `FOR UPDATE` vs. just doing an `UPDATE ... WHERE Value = :Value` and measuring the number of rows changed?

lieut-data (changes requested):
A handful of comments, but also wondering if it would it make sense to "rewrite" https://github.com/mattermost/mattermost-server/blob/master/app/migrations.go using these primitives to verify the interface for known old use cases?

agnivade:
> Yeap i thought about that, but my understanding is that if we kill the routine midway through it will rollback. So i think we have idempotency.

Hmm .. not very sure about that. From reading the docs, `CREATE INDEX CONCURRENTLY` cannot be performed within a transaction block. And due to the 2 pass approach, a failure during creation of an index always results in the corrupted index to remain in the DB. (https://blog.dbi-services.com/create-index-concurrently-in-postgresql/)

Also,

> If a problem arises while scanning the table, such as a deadlock or a uniqueness violation in a unique index, the CREATE INDEX command will fail but leave behind an “invalid” index. This index will be ignored for querying purposes because it might be incomplete; however it will still consume update overhead. 

> The recommended recovery method in such cases is to drop the index and try again to perform CREATE INDEX CONCURRENTLY. (Another possibility is to rebuild the index with REINDEX INDEX CONCURRENTLY).

----

There's also another small point which I just realized: we need to increase the timeout of these async queries. The usual DB query timeout from the config will apply to all these too, which might be too low. So we would need to override them to have possibly infinite timeout.

nronas:
> There's also another small point which I just realized: we need to increase the timeout of these async queries. The usual DB query timeout from the config will apply to all these too, which might be too low. So we would need to override them to have possibly infinite timeout.

Yeap yeap, all these are on my todo list.

> Hmm .. not very sure about that. From reading the docs, CREATE INDEX CONCURRENTLY cannot be performed within a transaction block. And due to the 2 pass approach, a failure during creation of an index always results in the corrupted index to remain in the DB. (https://blog.dbi-services.com/create-index-concurrently-in-postgresql/)

Totally, you are right,  it does not support ddl transactions, darn it.

store/sqlstore/store.go
nronas: yeap it will kill the parent process, this implementation does not change those semantics because we still have in master a situation that if you add an index on the posts table, for example, the boot up will hand until a timeout is reached. The strategy that we've followed with some customers/cloud is telling them to run a script to add the index before they boot the server

store/sqlstore/store.go
nronas: Totally with you on this, i've just followed what we follow for the sync indexes. We basically call `createIndexesIfNotExists` for each of the stores. Happy to keep it in one place, 0/5 here

model/system.go
nronas: Totally!

store/sqlstore/store.go
nronas: if we do setup + teardown here, that does not cover all the race conditions, since the teardown empties the column, and as soon as it commits the transaction the next thing that waits on that row can pick it up and start running the same migrations again(of course this time will be a no-op)

lieut-data: Maybe I'm misunderstanding the flow -- if one server starts and holds onto the lock indefinitely, how will a subsequent server upgrade successfully migrate unless the whole cluster is first powered off? (It's my understanding we support online, zero downtime upgrades at present.)

store/sqlstore/store.go
lieut-data: Wondering if we shouldn't forcibly stop our attempted migration here instead of holding on "just a bit longer" in the hopes we don't get killed anyway.

store/sqlstore/store.go
nronas: The servers do not hold any lock while performing the migrations. The lock is only obtained in order to syncronise reads after writes(essentially guarantee that everyone will see the same value that has been written in `async_migration_runner_id` key). If the value on that key is equal to the store id then that server will run the migrations. Only 1 server will actively run the migrations and the N-1 will skip them and boot up.

There is a scenario in which we have an existing cluster that one node actively runs the async migrations, if that node hasn't finished running the migrations if a new node is added will not run any new migrations, but I think this is not something that is communicated to the customers or our customers follow, since if we deploy a new node that runs a new version of the code without rebooting the rest then the cluster becomes heterogeneous.  Do our customers follow that approach for having zero downtime upgrades? My understanding was that customers in HA formation reboot the whole cluster, and subsequently wait for all pending migrations to finish during boot up.

nronas:
Just for posterity, we synced up with @lieut-data & @agnivade and we reached the conclusion that we should keep the server behaviour as is when it comes to applying migrations(we called it "blocking" mode), and give the ability to customers/users to apply long running migrations/transitions by using a command(CLI). The latter should pave the way for the OKR that we have in the server team around creating a migrations schema framework.

PR #16780, opened 2021-01-22
MM-32248 streamline serialization auto-generation

wiggin77:
#### Summary
Currently there are a handful of structs in model that have serialization methods auto-generated via `make gen-serialized`.  For this to work, all the structs that require serialization methods generated must be temporarily copied into session.go, otherwise running the command will break the build.

This PR streamlines that process by having the make file run the generator against the files that need it, and annotating structs in those files that don't require it so they are ignored.

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-32248

#### Release Note
```release-note
NONE
```


agnivade:
You would need to add the right files to `-inconsistentReceiverName.ignore=serialized_gen.go` in the `make vet` target.

Also, I think I might've had to insert the license string manually in the generated files.

wiggin77:
> You would need to add the right files to `-inconsistentReceiverName.ignore=serialized_gen.go` in the `make vet` target.
> 
> Also, I think I might've had to insert the license string manually in the generated files.

Thanks for the heads up.  BTW, `make vet` is broken for me:
```
❯ make vet
MM_VET_OPENSPEC_PATH not set or spec yaml path in it is incorrect. Skipping API check
flag provided but not defined: -inconsistentReceiverName.ignore
usage: go vet [-n] [-x] [-vettool prog] [build flags] [vet flags] [packages]
Run 'go help vet' for details.
Run '/home/dlauder/Development/mattermost/mattermost-server/bin/mattermost-govet -help' for the vet tool's flags.
make: *** [Makefile:576: vet] Error 2
```

agnivade:
I think you need to upgrade your local installation of mattermost-govet to the latest.

wiggin77:
Thanks @agnivade.  Should be good to go now.

Makefile
```diff
@@ -591,10 +591,10 @@ endif
 
 gen-serialized: ## Generates serialization methods for hot structs
 	# This tool only works at a file level, not at a package level.
```
agnivade, line 593: I'd suggest adding a line to the effect that:
> There will be some warnings about "unresolved identifiers", but that is because of the above problem. Since we are generating methods for all the relevant files at a package level, all identifiers will be resolved. An alternative to remove the warnings would be to temporarily move all the structs to the same file, but that involves a lot of manual work.

to preserve the context.

agnivade (approved):

PR #16783, opened 2021-01-25
MM-31709 Store posts/reactions for remote clusters

wiggin77:
#### Summary
This PR implements a sync strategy for shared channels.  When sync messages are received by a remote cluster any posts, reactions and users are upserted using app layer APIs.  Clients are notified of changes.

There remains one or two edge cases where a redundant sync message is sent which will be addressed in a subsequent PR.

This should be the last large PR for Shared Channels MVP.  The rest should tweaks, bug fixes and smaller features left to the end.

**To be merged into Shared Channels feature branch.**

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-31709

#### Release Note
```release-note
NONE
```


services/sharedchannel/msg.go
```diff
@@ -6,48 +6,104 @@ package sharedchannel
...
+			continue
+		}
+
 		if sm, ok := cache[p.Id]; ok {
 			syncMessages = append(syncMessages, sm)
 			continue
 		}
 
-		reactions, err := scs.server.GetStore().Reaction().GetForPost(p.Id, true)
+		// any reactions originating from the remote cluster are filtered out
```
streamer45, line 58: Can filtering things out on a remote cluster basis create problems with the cached message?

services/sharedchannel/sync_recv.go
```diff
@@ -28,223 +28,208 @@ func (scs *Service) onReceiveSyncMessage(msg model.RemoteClusterMsg, rc *model.R
...
+			mlog.String("post_id", sm.PostId),
+			mlog.String("channel_id", sm.ChannelId),
+			mlog.Int("reaction_count", len(sm.Reactions)),
+			mlog.Int("user_count", len(sm.Users)),
+			mlog.Bool("has_post", sm.Post != nil),
+		)
 
 		if channel == nil {
-			if channel, err = scs.server.GetStore().Channel().Get(sm.Post.ChannelId, true); err != nil {
+			if channel, err = scs.server.GetStore().Channel().Get(sm.ChannelId, true); err != nil {
```
streamer45, line 59: Can the channel change between loop iterations?

streamer45 (commented):
Things are getting serious now :) Very nice :+1: 
I've gave this a quick first look and left some comments and questions.

model/post.go
```diff
@@ -97,6 +97,7 @@ type Post struct {
 	FileIds       StringArray     `json:"file_ids,omitempty"`
 	PendingPostId string          `json:"pending_post_id" db:"-"`
 	HasReactions  bool            `json:"has_reactions,omitempty"`
+	RemoteId      *string         `json:"remote_id,omitempty"`
```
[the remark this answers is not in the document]
wiggin77: If it is not a pointer then all the `SELECT *` queries need to be replaced with lists of fields and COALESCE for RemoteId, or we need to remove all NULLs from the table on upgrade.  I've gotten pushback in the past about adding non-nullable fields on the big tables.

services/sharedchannel/msg.go
```diff
@@ -6,48 +6,104 @@ package sharedchannel
...
 // msgCache caches the work of converting a change in the Posts table to a remote cluster message.
 // Maps Post id to syncMsg.
 type msgCache map[string]syncMsg
 
-// syncMsg represents a change in content (post add/edit/delete, reaction add/remove).
+// syncMsg represents a change in content (post add/edit/delete, reaction add/remove, users).
 // It is sent to remote clusters as the payload of a `RemoteClusterMsg`.
 type syncMsg struct {
+	ChannelId string            `json:"channel_id"`
+	PostId    string            `json:"post_id"`
```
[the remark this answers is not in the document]
wiggin77: Correct.

services/sharedchannel/msg.go
```diff
@@ -6,48 +6,104 @@ package sharedchannel
...
 		}
 
+		postSync := p
+
+		// TODO:  don't include the post if only the reactions changed. Unfortunately there is no way to reliably know the
+		//        difference between an existing (synchronized) post with new reaction, and a brand new post (un-synchronized)
+		//        with a reaction.
+		//if p.EditAt < p.UpdateAt && p.CreateAt < p.UpdateAt && p.DeleteAt == 0 {
+		//	postSync = nil
+		//}
```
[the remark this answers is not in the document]
wiggin77: Yes.  Currently the logic is assume everything needs to be sent unless we can detect it doesn't need to.  I need to re-visit this as it wasn't working as expected when I wrote it, but I have fixed other issues since then which might have been the cause.  I have another ticket to fix the redundant post sends.

services/sharedchannel/msg.go
wiggin77: Not sure I follow.  The filter ensures that when a remote sends a reaction we don't send it back.  The remote already has it so I don't see any cache issues.  I may be missing something.

services/sharedchannel/sync_recv.go
wiggin77: I don't see how.  The snapshot of posts we're working with can only come from the same channel.  Even if a post was moved to another channel (wrangler) it will just appear deleted in a subsequent sync.

model/post.go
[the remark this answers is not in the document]
streamer45: Oh I see, that would be messy.

services/sharedchannel/msg.go
streamer45: Mainly I was worried that we could be excluding data just for one remote and since that would get cached we would be using always the same message for a given post.

services/sharedchannel/sync_recv.go
streamer45: But we are using `sm.ChannelId`. Does that mean that all sync messages are for the same channel?

wiggin77: Yes.  On the sending side, an entire sync message is populated from a single GetPostsSince, which is channel specific.  It is a contract of the message that it can only contain post/reaction/user from a single channel.

services/sharedchannel/msg.go
wiggin77: Ah, I think you're on to something.  If the remote cluster from which a post was sent happens to be the first one getting sync'd then an incomplete sync message is cached.  Also, the cache can have issues as users are filtered out.  The caching is a premature optimization that I will remove until a reliable mechanism can be devised.  That cache is likely also  why I'm getting redundant updates occasionally - a remote caches a full sync message that then gets sent to a remote that only needs a partial update.

wiggin77: Done via https://github.com/mattermost/mattermost-server/pull/16783/commits/b4c8d01f8adc1ac87bc789e6c033569c13fc2a9b

services/sharedchannel/sync_recv.go
```diff
@@ -28,223 +28,208 @@ func (scs *Service) onReceiveSyncMessage(msg model.RemoteClusterMsg, rc *model.R
...
-			// update post
-			rpost, appErr = scs.app.UpdatePost(sm.Post, false)
-			scs.server.GetLogger().Log(mlog.LvlSharedChannelServiceError, "Updating sync post",
-				mlog.String("post_id", sm.Post.Id),
-				mlog.String("channel_id", sm.Post.ChannelId))
+	if euser == nil {
+		if userSaved, err = scs.server.GetStore().User().Save(user); err != nil {
+			if _, ok := err.(errConflict); !ok {
+				return nil, fmt.Errorf("error inserting sync user: %w", err)
+			}
```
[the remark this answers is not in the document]
wiggin77: Right. UserStore.Save is converting unique constraint errors into InvalidInput errors (not even wrapping).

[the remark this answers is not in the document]
wiggin77: Thanks.  Should be fixed via https://github.com/mattermost/mattermost-server/pull/16783/commits/e075b132de5c9b764647817104f5d58763262a0e

services/sharedchannel/msg.go
```diff
@@ -56,3 +103,92 @@ func (scs *Service) postsToMsg(posts []*model.Post, cache msgCache) (model.Remot
...
+		// user not in the SharedChannelUsers table, so we must add them.
+		scu = &model.SharedChannelUser{
+			UserId:          user.Id,
+			RemoteClusterId: rc.RemoteId,
+		}
+		if _, err = scs.server.GetStore().SharedChannel().SaveUser(scu); err != nil {
+			scs.server.GetLogger().Log(mlog.LvlSharedChannelServiceError, "Error adding user to shared channel users",
+				mlog.String("remote_id", rc.RemoteId),
+				mlog.String("user_id", user.Id))
+		}
```
mgdelacroix, line 178: Should we `return false, err` here instead of marking the user for sync without a `SharedChannelUser` entry?

store/sqlstore/reaction_store.go
```diff
@@ -75,26 +78,56 @@ func (s *SqlReactionStore) Delete(reaction *model.Reaction) (*model.Reaction, er
 	return reaction, nil
 }
 
+// GetForPost returns all reactions associated with `postId` that are not deleted.
 func (s *SqlReactionStore) GetForPost(postId string, allowFromCache bool) ([]*model.Reaction, error) {
-	var reactions []*model.Reaction
+	queryString, args, err := s.getQueryBuilder().
```
mgdelacroix, line 83: :heart:

mgdelacroix (commented):
This is going to be awesome!! :rocket:

model/channel_shared.go
```diff
@@ -176,3 +176,37 @@ type SharedChannelRemoteStatus struct {
...
+	scu.CreateAt = GetMillis()
+}
+
+func (scu *SharedChannelUser) IsValid() *AppError {
+	if !IsValidId(scu.Id) {
+		return NewAppError("SharedChannelUser.IsValid", "model.channel.is_valid.id.app_error", nil, "Id="+scu.Id, http.StatusBadRequest)
+	}
+
+	if !IsValidId(scu.UserId) {
+		return NewAppError("SharedChannelUser.IsValid", "model.channel.is_valid.id.app_error", nil, "UserId="+scu.UserId, http.StatusBadRequest)
```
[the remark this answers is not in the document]
wiggin77: `model.channel.is_valid.id.app_error` is a translated string that in English says, "Invalid Id."    I don't think we need to create additional strings with the same translation.

model/channel_shared.go
```diff
@@ -176,3 +176,37 @@ type SharedChannelRemoteStatus struct {
...
+	}
+
+	if !IsValidId(scu.RemoteClusterId) {
+		return NewAppError("SharedChannelUser.IsValid", "model.channel.is_valid.id.app_error", nil, "RemoteClusterId="+scu.RemoteClusterId, http.StatusBadRequest)
```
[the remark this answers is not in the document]
wiggin77: Same as above.

services/sharedchannel/msg.go
```diff
@@ -56,3 +103,92 @@ func (scs *Service) postsToMsg(posts []*model.Post, cache msgCache) (model.Remot
...
+	}
+	return users
+}
+
+func sanitizeUserForSync(user *model.User) *model.User {
+	user.Password = model.NewId()
+	user.AuthData = nil
+	user.AuthService = ""
+	user.Roles = "system_user"
```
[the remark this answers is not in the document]
wiggin77: I'm not sure that's needed.  These dummy accounts cannot be logged into and are only created to satisfy the app layer and client.  My understanding about guest accounts (based on this: https://docs.mattermost.com/deployment/guest-accounts.html) is their limitations are entirely about being logged in (i.e. cannot discover public channels, join open teams, DM, invite) and making them guest accounts on the remote sites would not provide any additional security.

If we did make them guest accounts then we would have to handle them being promoted/demoted.  That could be updated via user sync but that is a lazy sync (currently max once per hour).  Not sure if that would bite us in some way.

services/sharedchannel/msg.go
```diff
@@ -56,3 +103,92 @@ func (scs *Service) postsToMsg(posts []*model.Post, cache msgCache) (model.Remot
...
+func (scs *Service) shouldUserSync(user *model.User, rc *model.RemoteCluster) (bool, error) {
+	// don't sync users with the remote they originated from.
+	if user.RemoteId != nil && *user.RemoteId == rc.RemoteId {
+		return false, nil
+	}
+
+	scu, err := scs.server.GetStore().SharedChannel().GetUser(user.Id, rc.RemoteId)
+	if err != nil {
+		if _, ok := err.(errNotFound); !ok {
+			return false, nil
```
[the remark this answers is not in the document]
wiggin77: Good catch.

services/sharedchannel/msg.go
wiggin77: If we don't sync the user then the post create at the remote side will fail (assuming the user doesn't already exist there).  

Allowing the user to sync here just means we'll try to create the `SharedChannelUser` record again next time.  That table is really about not sync'ing users redundantly for every post they make, however it's better to sync them too often than not enough.

services/sharedchannel/msg.go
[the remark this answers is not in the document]
wiggin77: Fixed via https://github.com/mattermost/mattermost-server/pull/16783/commits/559245156afcb14c29ebf3d9352a18730683a37f

services/sharedchannel/msg.go
```diff
@@ -56,3 +103,92 @@ func (scs *Service) postsToMsg(posts []*model.Post, cache msgCache) (model.Remot
...
+	}
+
+	// TODO: extract @mentions to local users and sync those as well?
+
+	users := make([]*model.User, 0)
+
+	for id := range userIds {
+		user, err := scs.server.GetStore().User().Get(id)
+		if err == nil {
+			if sync, err := scs.shouldUserSync(user, rc); err != nil {
```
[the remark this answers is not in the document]
wiggin77: Fixed via https://github.com/mattermost/mattermost-server/pull/16783/commits/559245156afcb14c29ebf3d9352a18730683a37f

services/sharedchannel/msg.go
```diff
@@ -56,3 +103,92 @@ func (scs *Service) postsToMsg(posts []*model.Post, cache msgCache) (model.Remot
...
+		userIds[r.UserId] = struct{}{}
+	}
+
+	// TODO: extract @mentions to local users and sync those as well?
+
+	users := make([]*model.User, 0)
+
+	for id := range userIds {
+		user, err := scs.server.GetStore().User().Get(id)
+		if err == nil {
```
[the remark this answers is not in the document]
wiggin77: Fixed via https://github.com/mattermost/mattermost-server/pull/16783/commits/559245156afcb14c29ebf3d9352a18730683a37f

wiggin77:
@streamer45 and @mgdelacroix, thanks very much for your continued support reviewing these PR's. You've provided great feedback and caught some issues that would have been a headache later.

model/channel_shared.go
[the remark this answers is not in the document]
mgdelacroix: makes sense

services/sharedchannel/msg.go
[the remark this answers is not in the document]
mgdelacroix: Understood. I was thinking on how the client will show the shared guests (more than effective limitations), but if we acknowledge that all remote users will be shown as "just" shared users with no other distinction, I think it's fine to keep it as is

mgdelacroix (approved):

services/sharedchannel/sync_recv.go
streamer45: One last thing, is it possible that `sm.ChannelId` differs from `sm.Post.ChannelId`?
I am trying to understand why we use the former to get the channel and the latter to get the team.

services/sharedchannel/sync_recv.go
```diff
@@ -28,223 +28,211 @@ func (scs *Service) onReceiveSyncMessage(msg model.RemoteClusterMsg, rc *model.R
...
-			scs.server.GetLogger().Log(mlog.LvlSharedChannelServiceError, "Error saving sync Post",
-				mlog.String("remote", rc.DisplayName),
-				mlog.String("ChannelId", sm.Post.ChannelId),
-				mlog.String("PostId", sm.Post.Id),
-				mlog.Err(err),
-			)
-			postErrors = append(postErrors, sm.Post.Id)
-			continue
+		// add post (may be nil if only reactions changed)
+		if sm.Post != nil {
```
streamer45, line 99: Is the order important or can this `if sm.Post != nil {` block be merged with the previous one above?

services/sharedchannel/sync_recv.go
```diff
@@ -28,223 +28,211 @@ func (scs *Service) onReceiveSyncMessage(msg model.RemoteClusterMsg, rc *model.R
...
+			rpost, err := scs.upsertSyncPost(sm.Post, channel, rc)
+			if err != nil {
+				postErrors = append(postErrors, sm.Post.Id)
+				scs.server.GetLogger().Log(mlog.LvlSharedChannelServiceError, "Error upserting sync post",
+					mlog.String("post_id", sm.Post.Id),
+					mlog.String("channel_id", sm.Post.ChannelId),
+					mlog.Err(err))
+			} else if lastSyncAt < rpost.UpdateAt {
+				lastSyncAt = rpost.UpdateAt
+			}
```
streamer45, line 109: So if somehow the `upsertSyncPost` fails and a subsequent one succeeds, could setting `lastSyncAt` cause losing posts? (same argument for reactions below)

services/sharedchannel/sync_recv.go
wiggin77: Ah, I see where the confusion is.  When Agniva added the perma-link processing he didn't realize there could be only one channel per batch of sync messages and he created a map of channels to teams.  My fault - I completely missed that when reviewing.  I will simply that part (remove the map) and also sanity check that `sm.Post.ChannelId == sm.ChannelId`.  I'll then make all reference to channel id use `sm.ChannelId`.

services/sharedchannel/sync_recv.go
wiggin77: Yes it can.  I thought a lot about that, and sometime in the future I will need a retry on sync'ing posts/reactions with a max count for retries.  On the receiving side, if we don't move on by setting lasySyncAt and the error repeats we will be stuck at that point.  We do currently send the failed post ids back to the sender, so the retry can be added there.  I don't think we want to block it here.

services/sharedchannel/sync_recv.go
wiggin77: They can be merged.  Fixed via https://github.com/mattermost/mattermost-server/pull/16783/commits/9bace6eb1613a4e5f34f48b99b964ac0fac95bab

store/sqlstore/shared_channel_store.go
```diff
@@ -42,13 +42,20 @@ func newSqlSharedChannelStore(sqlStore *SqlStore) store.SharedChannelStore {
 		tableSharedChannelRemotes.ColMap("CreatorId").SetMaxSize(26)
 		tableSharedChannelRemotes.ColMap("RemoteClusterId").SetMaxSize(26)
 		tableSharedChannelRemotes.SetUniqueTogether("ChannelId", "RemoteClusterId")
+
+		tableSharedChannelUsers := db.AddTableWithName(model.SharedChannelUser{}, "SharedChannelUsers").SetKeys(false, "Id")
+		tableSharedChannelUsers.ColMap("Id").SetMaxSize(26)
+		tableSharedChannelUsers.ColMap("UserId").SetMaxSize(26)
+		tableSharedChannelUsers.ColMap("RemoteClusterId").SetMaxSize(64)
```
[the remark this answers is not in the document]
wiggin77: No idea.  Fixed via https://github.com/mattermost/mattermost-server/pull/16783/commits/9bace6eb1613a4e5f34f48b99b964ac0fac95bab

services/sharedchannel/sync_recv.go
wiggin77: Fixed via https://github.com/mattermost/mattermost-server/pull/16783/commits/9bace6eb1613a4e5f34f48b99b964ac0fac95bab

streamer45 (approved):
Looks great :+1:

PR #16786, opened 2021-01-25
[MM-31497] - Prevent end users from inviting people and exceeding the free tier limits

AGMETEOR:
#### Summary
This PR adds a new cloud only endpoint to obtain stats about the kind of subscription a user has.

#### Ticket Link
[MM-31497](https://mattermost.atlassian.net/browse/MM-31497?atlOrigin=eyJpIjoiMDJhMDQwOWM1ODZhNDMzM2IyNDE5ZjAxNDM1OTg4Y2UiLCJwIjoiaiJ9)

#### Release Note

```release-note
Added new API endpoint GET /api/v4/cloud/subscription/stats
```


api4/cloud.go
```diff
@@ -46,11 +46,6 @@ func getSubscription(c *Context, w http.ResponseWriter, r *http.Request) {
 		return
 	}
 
-	if !c.App.SessionHasPermissionTo(*c.App.Session(), model.PERMISSION_SYSCONSOLE_READ_BILLING) {
```
jespino, line 49: Removing this permission you are exposing way too much data to any user. If you want the "remaining users available" I think you should expose only that.

api4/user.go
```diff
@@ -599,6 +600,21 @@ func getFilteredUsersStats(c *Context, w http.ResponseWriter, r *http.Request) {
 	w.Write([]byte(stats.ToJson()))
 }
 
+func getGeneralUserStats(c *Context, w http.ResponseWriter, r *http.Request) {
```
jespino, line 603: I think exposing this is not strictly needed for the functionality, I would say here we should expose the "remaining available slots for new users".

api4/cloud.go
AGMETEOR: Hi Jesùs. Thanks a lot for the review :) This is needed for normal users to access subscription information. From some discussions,  @marianunez said it was alright to remove these permissions.

api4/user.go
AGMETEOR: Oh yeah! I think that's much better. Thanks a lot

api4/cloud.go
AGMETEOR: With the your follow up review, I now understand. I could leave these permissions here and shift the part of calculation from the webapp to the new end point.

api4/cloud.go
```diff
@@ -38,6 +38,9 @@ func (api *API) InitCloud() {
 
 	// POST /api/v4/cloud/webhook
 	api.BaseRoutes.Cloud.Handle("/webhook", api.CloudApiKeyRequired(handleCWSWebhook)).Methods("POST")
+
+	// GET /api/v4/cloud/free_tier/stats
+	api.BaseRoutes.Cloud.Handle("/free_tier/stats", api.ApiSessionRequired(getFreeTierStats)).Methods("GET")
```
marianunez, line 43: Given that these stats are related to their subscription, maybe we can follow the existing convention and have the endpoint be `/api/v4/cloud/subscription/stats`?
```suggestion
	api.BaseRoutes.Cloud.Handle("/subscription/stats", api.ApiSessionRequired(getSubscriptionStats)).Methods("GET")
```

api4/cloud.go
marianunez: Given that we will now pull the information from the new stats endpoint, we could return the permission back here?

api4/cloud.go
```diff
@@ -67,6 +65,31 @@ func getSubscription(c *Context, w http.ResponseWriter, r *http.Request) {
...
+		c.Err = model.NewAppError("Api4.getFreeTierStats", "app.user.get_total_users_count.app_error", nil, err.Error(), http.StatusInternalServerError)
+		return
+	}
+	cloudUserLimit := *c.App.Config().ExperimentalSettings.CloudUserLimit
+
+	s := cloudUserLimit - count
+
+	stats, _ := json.Marshal(model.FreeTierStats{
+		RemainingSeats: int(s),
+		TierStatus:     subscription.IsPaidTier,
```
marianunez, line 87: Also here for consistency we could keep the same name as `IsPaidTier`
```suggestion
		IsPaidTier:     subscription.IsPaidTier,
```

marianunez (changes requested):
Thanks @AGMETEOR, great approach with this new endpoint. Some feedback on naming.

marianunez:
Also, can you update the PR with the new implementation?

api4/cloud.go
```diff
@@ -67,6 +68,31 @@ func getSubscription(c *Context, w http.ResponseWriter, r *http.Request) {
 	w.Write(json)
 }
 
+func getSubscriptionStats(c *Context, w http.ResponseWriter, r *http.Request) {
+	subscription, appErr := c.App.Cloud().GetSubscription()
```
marianunez, line 72: We also need to make sure they are on a cloud license before we try to get the subscription something like: 

```
if c.App.Srv().License() == nil || !*c.App.Srv().License().Features.Cloud {
		c.Err = model.NewAppError("Api4.getSubscriptionStats", "api.cloud.license_error", nil, "", http.StatusNotImplemented)
		return
	}
```

jespino (approved):
Looks good to me. Thanks! :)

api4/cloud.go
```diff
@@ -67,6 +68,36 @@ func getSubscription(c *Context, w http.ResponseWriter, r *http.Request) {
 	w.Write(json)
 }
 
+func getSubscriptionStats(c *Context, w http.ResponseWriter, r *http.Request) {
+	if c.App.Srv().License() == nil || !*c.App.Srv().License().Features.Cloud {
+		c.Err = model.NewAppError("Api4.getSubscriptionStats", "api.cloud.license_error", nil, "", http.StatusNotImplemented)
```
marianunez, line 73:
```suggestion
		c.Err = model.NewAppError("Api4.getSubscriptionStats", "api.cloud.license_error", nil, "", http. StatusInternalServerError)
```

marianunez (changes requested):
LGTM, can you update the PR description to the final approach taken with the new endpoint for `/subscription/stats`?

Also, let's add a test case to cover this new endpoint. Cloud tests are currently in enterprise here: https://github.com/mattermost/enterprise/blob/master/cloud/cloud_test.go

AGMETEOR:
> LGTM, can you update the PR description to the final approach taken with the new endpoint for `/subscription/stats`?
> 
> Also, let's add a test case to cover this new endpoint. Cloud tests are currently in enterprise here: https://github.com/mattermost/enterprise/blob/master/cloud/cloud_test.go

Thanks @marianunez for the review. Here's the test case in the enterprise repo https://github.com/mattermost/enterprise/pull/914

marianunez (approved):
Great work @AGMETEOR! 🎉

marianunez:
I've removed the `QA Review` label for this, given that it will be tested as part of https://github.com/mattermost/mattermost-webapp/pull/7439 and it includes the unit tests at https://github.com/mattermost/enterprise/pull/914

marianunez:
@AGMETEOR you also need to add the API documentation for this new link here https://github.com/mattermost/mattermost-api-reference to be able to get past the build error. You can see this example https://github.com/mattermost/mattermost-api-reference/pull/587

ethervoid:
@AGMETEOR you need to add the new endpoint to the [API reference project](https://github.com/mattermost/mattermost-api-reference) as you can see in the build error `https://github.com/mattermost/mattermost-api-reference` in the `mattermost-server` branch

PR #16792, opened 2021-01-26
MM-31391: Add max connection idle timeout to config

agnivade:
https://mattermost.atlassian.net/browse/MM-31391

```release-note
A new config field called `ConnMaxIdleTimeMilliseconds` was added to `DatabaseSettings` which allows to control the maximum time a database connection can remain idle. The default value is set to 5 minutes.
```

Closes https://github.com/mattermost/mattermost-server/issues/16601

### DIFF
--- a/model/config.go
+++ b/model/config.go
@@ -3247,7 +3247,7 @@ func (s *SqlSettings) isValid() *AppError {
 	}
 
 	if *s.ConnMaxIdleTimeMilliseconds < 0 {
-		return NewAppError("Config.IsValid", "model.config.is_valid.sql_conn_max_lifetime_milliseconds.app_error", nil, "", http.StatusBadRequest)
+		return NewAppError("Config.IsValid", "model.config.is_valid.sql_conn_max_idle_time_milliseconds.app_error", nil, "", http.StatusBadRequest)
 	}
 
 	if *s.QueryTimeout <= 0 {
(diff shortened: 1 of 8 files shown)


model/config.go
```diff
@@ -1091,6 +1091,7 @@ type SqlSettings struct {
 	DataSourceSearchReplicas    []string `access:"environment,write_restrictable,cloud_restrictable"`
 	MaxIdleConns                *int     `access:"environment,write_restrictable,cloud_restrictable"`
 	ConnMaxLifetimeMilliseconds *int     `access:"environment,write_restrictable,cloud_restrictable"`
+	ConnMaxIdleTimeMilliseconds *int     `access:"environment,write_restrictable,cloud_restrictable"`
```
streamer45, line 1094: I understand we want to be consistent with the other setting but I am curious to know if the millisecond precision makes any sense. Could there be reasons to go below 1 second?

streamer45 (approved):

model/config.go
agnivade: Yeah I agree. It's mainly legacy. I doubt anyone would know why specifically millisecond precision was chosen. :P

isacikgoz (approved):
LGTM 👍

agnivade:
@lindalumitchell - Another config field that got added. Covered with unit tests. Should be okay to be QA-deferred.

lindalumitchell (approved):

PR #16794, opened 2021-01-26
[MM-32253] Fix "Cannot mark unread a DM message in a thread"

larkox:
#### Summary
When marking a channel post unread, if there is a rootID, we do some extra checks. Among those checks, is counting the thread mentions. We used to check the team at the beginning of this function, but since DMs don't have a team, it failed. We moved the team check after the DM check to avoid this error.

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-32253

#### Release Note
```release-note
NONE
```

### DIFF
--- a/app/post.go
+++ b/app/post.go
@@ -1343,10 +1343,6 @@ func (a *App) MaxPostSize() int {
 
 // countThreadMentions returns the number of times the user is mentioned in a specified thread after the timestamp.
 func (a *App) countThreadMentions(user *model.User, post *model.Post, teamId string, timestamp int64) (int64, *model.AppError) {
-	team, err := a.GetTeam(teamId)
-	if err != nil {
-		return 0, err
-	}
 	channel, err := a.GetChannel(post.ChannelId)
 	if err != nil {
 		return 0, err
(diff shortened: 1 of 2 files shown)


reflog (approved):

app/post.go
```diff
@@ -1379,6 +1375,11 @@ func (a *App) countThreadMentions(user *model.User, post *model.Post, teamId str
 		return int64(count), nil
 	}
 
+	team, err := a.GetTeam(teamId)
+	if err != nil {
+		return 0, err
+	}
```
streamer45, line 1381: Would group channels still fail the check?

larkox: I don't understand it 100%, but apparently, `Thread().GetMembershipsForUser` returns nil for GM, and therefore exits the condition before reaching `countThreadMentions`.

larkox: My bad, you are right. It will also fail for GMs. Will fix.

amyblais:
@larkox Did you find if this is a recent regression and from which version / or which PR may have caused this?

streamer45 (approved):
Looks good :+1:

larkox:
@amyblais After doing some research, it seems these functions were included into the code on this commit: https://github.com/mattermost/mattermost-server/commit/c2036f614e1b36c89186a2ca7b2d5c0b19f53c3c
It was merged to master early December.

reflog (approved):

ogi-m (approved):
Tested and passed - DM messages in a thread are consistently marked as unread

PR #16796, opened 2021-01-26
MM-32047:Added AppCustomURLSchemes settings to telemetry

anurag6713:
#### Summary
Add AppCustomURLSchemes setting to telemetry

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-32047

#### Release Note
```release-note
NONE
```


services/telemetry/telemetry.go
```diff
@@ -696,6 +697,7 @@ func (ts *TelemetryService) trackConfig() {
 	})
 
 	ts.sendTelemetry(TrackConfigNativeApp, map[string]interface{}{
+		"isdefault_app_custom_url_schemes":    reflect.DeepEqual(cfg.NativeAppSettings.AppCustomURLSchemes, model.GetDefaultAppCustomURLSchemes()),
```
agnivade, line 700: Couple of things here:
- Do we care about the order of elements in the slice? `reflect.DeepEqual` doesn't.
- I'd suggest to include this functionality in `isDefault` itself. We can just check for the type, and then check the 2 slices. Importing the reflect package seems too heavy for this.

sbishel: Agreed.

anurag6713: Thanks! I have added the equality check in `isDefault`. I don't think we should cover the order of elements in the slice as it could mean value has been set something other than default.

services/telemetry/telemetry.go
```diff
@@ -183,6 +184,19 @@ func (ts *TelemetryService) sendTelemetry(event string, properties map[string]in
 }
 
 func isDefault(setting interface{}, defaultValue interface{}) bool {
+	if reflect.TypeOf(setting).Kind() == reflect.Slice && reflect.TypeOf(defaultValue).Kind() == reflect.Slice {
```
agnivade, line 187: Oh sorry for any confusion, I just meant to go with our immediate case of slice of strings and work with that. Importing the reflect package is effectively the same as using `reflect.DeepEqual` itself. If we see ourselves using slices of other things as well, we can use `reflect.DeepEqual` then.
```go
func comp(setting, defaultValue interface{}) bool {
	slice1, ok1 := setting.([]string)
	slice2, ok2 := defaultValue.([]string)
	if ok1 && ok2 {
		if len(slice1) != len(slice2) {
			return false
		}
		for i := 0; i < len(slice1); i++ {
			if slice1[i] != slice2[i] {
				return false
			}
		}
		return true
	}
	return setting == defaultValue
}
```

Apologies for the double work.

anurag6713: Understood! I think you have explained it right but my limited exposure to golang is to blame for this 😄 I thought that's the only available way to check the type and get the value. Thanks for the review and help! 🙂

anurag6713: Also I was trying to cover int slices as well (that's not in the use case but thought when we have it will be covered automatically). I guess that's where I went in the wrong direction into using reflect package again 😅

sbishel (changes requested):
Thoughts on creating a separate function `isDefaultArray()`?
Please add unit tests for array comparison.

agnivade (commented):
A separate function sounds good to me.

And also please add tests.

After that, it's good to go. Thanks for your patience with this, and apologies for the to and fro!

agnivade (approved):
Nicely done :sparkles:

sbishel (approved):
Thanks Anurag.

PR #16801, opened 2021-01-27
MM-31094: Adds tooling to develop and test using a MySQL instance with replicat…

mkraft:
…ion lag. Adds some lazy lookups to fallback to master if results are not found.

#### Summary

Does a lazy fallback to master in a few cases to solve four issues related to replication lag:

1. First login of a new LDAP user (user gets created and group, team, channel memberships are created in the same request).
2. Newly logged-in user reads from a node without the session replicated.
3. LDAP sync job syncs a user to a new channel.
4. Reply to new post where the reply request reads from a node that does not yet have the new post replicated.

Related PR: https://github.com/mattermost/enterprise/pull/916

#### Ticket Link

https://mattermost.atlassian.net/browse/MM-31094

#### Release Note

```release-note
Fixes bugs related to replication lag for Enterprise Edition instances configured to use read replicas. System administrators can enable the fix using the new SqlSettings.ReplicaLazyReads configuration setting.
```


config.mk
```diff
@@ -10,7 +10,7 @@
 # Must be space separated names.
 #
 # Example: mysql postgres elasticsearch
-ENABLED_DOCKER_SERVICES ?= mysql postgres inbucket
+ENABLED_DOCKER_SERVICES ?= mysql postgres inbucket mysql-read-replica
```
mkraft, line 13: Remove this so that it doesn't start by default.

store/sqlstore/post_store.go
```diff
@@ -529,7 +529,7 @@ func (s *SqlPostStore) GetEtag(channelId string, allowFromCache, collapsedThread
 	}
 	sql, args, _ := q.ToSql()
 	var et etagPosts
-	err := s.GetReplica().SelectOne(&et, sql, args...)
+	err := s.GetReplica().SelectOne(&et, sql, args)
```
mkraft, line 532: What?

store/sqlstore/session_store.go
```diff
@@ -76,15 +76,26 @@ func (me SqlSessionStore) Save(session *model.Session) (*model.Session, error) {
 func (me SqlSessionStore) Get(sessionIdOrToken string) (*model.Session, error) {
 	var sessions []*model.Session
 
-	if _, err := me.GetReplica().Select(&sessions, "SELECT * FROM Sessions WHERE Token = :Token OR Id = :Id LIMIT 1", map[string]interface{}{"Token": sessionIdOrToken, "Id": sessionIdOrToken}); err != nil {
+	retryUsingMasterF := func(result interface{}, err error) bool {
```
mkraft, line 79: Rename to `retryUsingMasterFunc`

app/group.go
```diff
@@ -212,7 +212,7 @@ func (a *App) UpsertGroupSyncable(groupSyncable *model.GroupSyncable) (*model.Gr
 
 	// reject the syncable creation if the group isn't already associated to the parent team
 	if groupSyncable.Type == model.GroupSyncableTypeChannel {
-		channel, nErr := a.Srv().Store.Channel().Get(groupSyncable.SyncableId, true)
+		channel, nErr := a.Srv().Store.Channel().Get(groupSyncable.SyncableId, false)
```
sbishel, line 215: What was the reason for this change?

sbishel (approved):
just a couple questions. Also wondering if we should run this through performance testing.

PR #16803, opened 2021-01-27
Make channel readonly when shared as such

nronas:
#### Summary
Implements: https://mattermost.atlassian.net/browse/MM-32304

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-32304

#### Release Note
```release-note
NONE
```


services/sharedchannel/service.go
```diff
@@ -177,3 +178,16 @@ func (scs *Service) pause() {
 
 	scs.server.GetLogger().Debug("Shared Channel Service inactive")
 }
+
+// Makes the remote channel to be read only, only for guests.
+func (scs *Service) makeChannelReadOnly(channel *model.Channel) *model.AppError {
```
nronas, line 183: Couldn't find the product requirements for this one anywhere, so will defer to @wiggin77 for this one. As is right now we define a remote channel as read-only, when guests cannot create a post to it(they can though react), also all members of that channel can post and react. There are essentially 4 layers of "readonlyness".

1) Guests cannot create a post
2) Guests cannot create a post and reactions
3) Guests and Members cannot create a post
4) Guests and Member cannot create a post and reactions(Essentially only admins can edit the channel)

wiggin77: I thought the link in the ticket pointed directly to the right section but I see now it points to the section above it.  This is meant to be like an announcement channel, the second section in that link.  The use case from design meetings was sharing a Town Square like channel with other orgs.

services/sharedchannel/service.go
```diff
@@ -177,3 +178,16 @@ func (scs *Service) pause() {
...
+	readonlyChannelModerations := []*model.ChannelModerationPatch{
+		{
+			Name:  &model.ChannelModeratedPermissions[0],
```
agnivade, line 186: Why not just have `PERMISSION_CREATE_POST.Id`? Easy to remember and read.

services/sharedchannel/service.go
```diff
@@ -177,3 +178,16 @@ func (scs *Service) pause() {
 
 	scs.server.GetLogger().Debug("Shared Channel Service inactive")
 }
+
+// Makes the remote channel to be read only, only for guests.
```
agnivade, line 182: Did you mean remote users here? Guests have a different meaning.

services/sharedchannel/channelinvite_test.go
```diff
@@ -0,0 +1,138 @@
...
+package sharedchannel
+
+import (
+	"encoding/json"
+	"fmt"
+	"net/http"
+	"testing"
+
+	"github.com/stretchr/testify/assert"
+
```
agnivade, line 13: Nit: looks a bit odd with so many newlines in between them.

services/sharedchannel/channelinvite.go
```diff
@@ -98,7 +98,7 @@ func combineErrors(err error, serror string) string {
 	return sb.String()
 }
 
-func (scs *Service) onReceiveChannelInvite(msg model.RemoteClusterMsg, rc *model.RemoteCluster, response remotecluster.Response) error {
+func (scs *Service) onReceiveChannelInvite(msg model.RemoteClusterMsg, rc *model.RemoteCluster, _ remotecluster.Response) error {
```
agnivade, line 101: Can we change the method signature altogether to not pass the param? Or is is satisfying some interface?

services/sharedchannel/service.go
nronas: For sure, will amend

services/sharedchannel/service.go
nronas: Took the reference for the definition of read-only-ness that we have in the channel permissions, we have a different layer for Guest users and Channel members, and I've implemented the read-only aspect to apply to Guests. Will amend that since Doug clarified that we want the channel to be an announcements channel so only admins can Post or React to the channel.

services/sharedchannel/channelinvite.go
nronas: Yeap it conforms to function type https://github.com/mattermost/mattermost-server/blob/MM-27493_shared_channels/services/remotecluster/service.go#L55

services/sharedchannel/channelinvite_test.go
nronas: Don't know why goimports did that. Will amend

nronas: amended in https://github.com/mattermost/mattermost-server/pull/16803/commits/749ec630122e22d8f49487d7c88457cbc7ba70f4

services/sharedchannel/service.go
nronas: amended in https://github.com/mattermost/mattermost-server/pull/16803/commits/749ec630122e22d8f49487d7c88457cbc7ba70f4

services/sharedchannel/service.go
nronas: amended in https://github.com/mattermost/mattermost-server/pull/16803/commits/749ec630122e22d8f49487d7c88457cbc7ba70f4

services/sharedchannel/service.go
nronas: All done in https://github.com/mattermost/mattermost-server/pull/16803/commits/749ec630122e22d8f49487d7c88457cbc7ba70f4

services/sharedchannel/service.go
agnivade: Ah so it's actually for guests. Thanks.

agnivade (approved):

services/sharedchannel/service.go
nronas: it was yes, but after the clarification, we want the readonly-ness to apply to all apart from admins.

wiggin77 (approved):
LGTM :+1:

PR #16805, opened 2021-01-27
[MM-32234] Fix config.json getting reset

streamer45:
#### Summary

PR fixes an issue with `config.json` getting sporadically reset by adding support for read-only config stores which I enabled for CLI commands (with some exceptions, e.g. `config` command). 

#### Ticket

https://mattermost.atlassian.net/browse/MM-32234

#### Release Notes

```release-note
Fixed an issue where config.json was sporadically getting reset upon CLI command execution.
```

### DIFF
--- a/api4/apitestlib.go
+++ b/api4/apitestlib.go
@@ -96,7 +96,7 @@ func setupTestHelper(dbStore store.Store, searchEngine *searchengine.Broker, ent
 	}
 	memoryStore.Set(memoryConfig)
 
-	configStore, err := config.NewStoreFromBacking(memoryStore, nil)
+	configStore, err := config.NewStoreFromBacking(memoryStore, nil, false)
 	if err != nil {
 		panic(err)
 	}
--- a/api4/config_test.go
+++ b/api4/config_test.go
@@ -691,11 +691,11 @@ func TestMigrateConfig(t *testing.T) {
 	})
 
 	th.TestForSystemAdminAndLocal(t, func(t *testing.T, client *model.Client4) {
-		f, err := config.NewStore("from.json", false, nil)
+		f, err := config.NewStore("from.json", false, false, nil)
 		require.NoError(t, err)
 		defer f.RemoveFile("from.json")
 
-		_, err = config.NewStore("to.json", false, nil)
+		_, err = config.NewStore("to.json", false, false, nil)
 		require.NoError(t, err)
 		defer f.RemoveFile("to.json")
 
--- a/app/config.go
+++ b/app/config.go
@@ -46,6 +46,9 @@ func (a *App) EnvironmentConfig() map[string]interface{} {
 }
 
 func (s *Server) UpdateConfig(f func(*model.Config)) {
+	if s.configStore.IsReadOnly() {
+		return
+	}
 	old := s.Config()
 	updated := old.Clone()
 	f(updated)
--- a/app/options.go
+++ b/app/options.go
@@ -43,9 +43,9 @@ func StoreOverride(override interface{}) Option {
 // or a database connection string. It receives as well a set of
 // custom defaults that will be applied for any unset property of the
 // config loaded from the dsn on top of the normal defaults
-func Config(dsn string, watch bool, configDefaults *model.Config) Option {
+func Config(dsn string, watch, readOnly bool, configDefaults *model.Config) Option {
 	return func(s *Server) error {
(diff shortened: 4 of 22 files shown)


crspeller (approved):

agnivade:
I'd like to take a step back and really understand the root of the issue.

From what I understand, the issue happens when a server is running in the usual config watch mode, and a CLI command is run at the same time. The CLI writes to config (irrespective of it needing to write), which triggers the server's config watcher. But somehow, the `ioutil.ReadAll` returns 0 bytes and then it writes back the default config due to it.

Couple of questions come up here:
- Why is the CLI trying to write the config in the first place when it doesn't need to?
- Can't this race anyways happen in the normal intended use-case where a user manually edits the config.json and the server picks up the changes? In this case, it seems like the server will write back the default config, wiping all the changes that the user did.

The answer to the first question is here: https://github.com/mattermost/mattermost-server/blob/d9a890e73fd90d09deed43fb7887ae529b16cc48/config/store.go#L262.

At this point, `s.config` is nil, because it's not even loaded from the backing store. So we are actually calling `Clone` with a nil config. But there's some unexpected magic happening in the `Clone` method:

```go
var ret Config
	if err := json.Unmarshal([]byte(o.ToJson()), &ret); err != nil {
		panic(err)
	}
	return &ret
```

If `o` is nil, it returns a pointer to an empty Config struct. So therefore, `loadLockedWithOld` is _always_ called with the default config. The problem doesn't actually appear unless the backing store also returns 0 bytes, in which case, the default config gets written instead of the actual config.

IMO, we should try to fix this behavior if possible. Because it looks like the racy behavior is somewhat unavoidable. What we can avoid is writing back the same config again to the store, just because we are reading an empty config.

I guess you must have already thought of this, and came to the conclusion that it's not feasible? I haven't dug any deeper than this, so looking forward to your thoughts on this. :)

streamer45:
The primary reason for this approach is that I don't feel comfortable messing with the config internals while rushing for a fix.

I don't have enough context/experience here and therefore I am not sure I could predict all possible edge cases.
Also, given the sense of urgency for the fix, I opted for limiting the immediate effects of the issue rather than fixing the (possibly erroneous?) logic.

As you said, the racy behaviour is hard to completely eradicate unless some file-locking mechanism is used but that would require additional complexity. So I decided to eliminate the race by eliminating the writes, at least for those commands that don't have reasons to modify the config. I've wrapped everything into allowing a store to be read-only which might not be as bad as a hack and could potentially find some use.

I share the sense of unsatisfaction here but I think a proper solution should be thought of as a separate process with enough time to validate it (config logic overhaul?). That said, if you are confident about a definitive solution to this, then I am happy to discuss it further and reconsider the approach.

agnivade (approved):
Sounds good to me then. We should probably file a ticket to address this tech-debt for config overhaul.

And it seems to me that we might have to address that sooner than later. Customers are already starting to experience issues: https://community-daily.mattermost.com/core/pl/c8qdzs3nxjn3ixqihc91uwofqc.

streamer45:
> Sounds good to me then. We should probably file a ticket to address this tech-debt for config overhaul.
> 
> And it seems to me that we might have to address that sooner than later. Customers are already starting to experience issues: https://community-daily.mattermost.com/core/pl/c8qdzs3nxjn3ixqihc91uwofqc.

Absolutely :+1: I am also happy to own this. I've made a ticket [(MM-32390)](https://mattermost.atlassian.net/browse/MM-32390) to revisit (and possibly improve) the whole configuration logic.

streamer45:
@lindy65 

A couple of things to check on this. First of all that the issue does not reproduce anymore (I've updated QA steps).
Then a general smoke-test around config changes would also be helpful.

furqanmlk (approved):
**Environment**
OS: Mac
MM Build : Master

**Steps to verify**

1. Ran Server
2. Make few changes in `config.json`
2. Open terminal run `while ./mattermost version; do :; done`
3. Open two more terminal and ran same command
4. Examined that `config.json` file did not get reset

Successfully verified ✅

amyblais:
Unless this is risky, a good candidate for v5.32 and v5.31 ESR.

streamer45:
> Unless this is risky, a good candidate for v5.32 and v5.31 ESR.

Anything touching config can be probably thought as potentially risky but the PR was specifically made to try and limit any side effect which would be limited to CLI commands. Also, given the annoyance of the issue I'd say it makes sense to cherry-pick.

PR #16808, opened 2021-01-28
MM-5639: Write panic to log on the way out

agnivade:
We introduce a panic pass-through layer which logs
the panic at a critical level before crashing.
This helps customers to easily get the panic output
in their logging infrastructure, rather than having to
go through journaltctl logs.

https://mattermost.atlassian.net/browse/MM-5639

```release-note
Server crashes due to runtime panics are now captured
as a log line.
```

### DIFF
--- a/cmd/mattermost/commands/server.go
+++ b/cmd/mattermost/commands/server.go
@@ -67,19 +67,6 @@ func runServer(configStore *config.Store, usedPlatform bool, interruptChan chan 
 	// and also preserve a crash dump for later investigation.
 	debug.SetTraceback("crash")
 
-	defer func() {
-		// A panic pass-through layer which just logs it
-		// and sends it upwards.
-		if x := recover(); x != nil {
-			var buf bytes.Buffer
-			pprof.Lookup("goroutine").WriteTo(&buf, 2)
-			mlog.Critical("A panic occurred",
-				mlog.Any("error", x),
-				mlog.String("stack", buf.String()))
-			panic(x)
-		}
-	}()
-
 	options := []app.Option{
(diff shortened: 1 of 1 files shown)


isacikgoz (approved):
Nice one! 👍

cmd/mattermost/commands/server.go
```diff
@@ -65,6 +67,19 @@ func runServer(configStore *config.Store, usedPlatform bool, interruptChan chan
 	// and also preserve a crash dump for later investigation.
 	debug.SetTraceback("crash")
 
+	defer func() {
```
wiggin77, line 70: In a normal method exit this would get executed after `server.Shutdown` due to lifo nature of the defer queue.  I'm not clear on if/how that changes in a panic scenario.  If `server.Shutdown` does get called first then the logging engine will be shutdown before this log call.

agnivade: That's a great point. From what I can see it shuts down the advanced logging part, which I didn't actually test. I am a bit rusty on the logging part, but does it disable the normal logger too, when the advanced logging shuts down?

wiggin77: It doesn't shut down the Zap logger, nothing does - it never exits cleanly but that doesn't matter so much since it writes synchronously.  However the Zap logger will be removed later this year so I think it's best to output the panic log before shutdown so it gets flushed.

wiggin77 (approved):
LGTM :+1:

agnivade:
It's difficult to test this manually. I have tested myself by modifying the code to return a panic.

PR #16812, opened 2021-01-28
config->util_test->TestIsDatabaseDSN duplice test case fix

Quaqmre:
#### Summary

Utils_test.go in config package TestIsDatabaseDSN test case fix.

#### Ticket Link

I dont have any Ticket

#### Release Note
<!--

```release-note
Fix the duplicated test case.  TestIsDatabaseDSN method check mysql and postgres prefix.But the test case checks 2 times mysql prefix.one of should be mysql prefix the other one should be postgres prefix
```

### DIFF
--- a/config/utils_test.go
+++ b/config/utils_test.go
@@ -158,8 +158,8 @@ func TestIsDatabaseDSN(t *testing.T) {
 			Expected: true,
 		},
 		{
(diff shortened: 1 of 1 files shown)


Quaqmre:
singin CLA

PR #16816, opened 2021-01-28
Reducing the filestore dependencies from the rest of the source code

jespino:
#### Summary
Reducing the filestore dependencies from the rest of the source code

#### Ticket Link
[MM-32481](https://mattermost.atlassian.net/browse/MM-32481)

#### Related PRs
enterprise: mattermost/enterprise#912

#### Release Note
```release-note
NONE
```


testlib/resources.go
```diff
@@ -151,7 +152,12 @@ func SetupTestResources() (string, error) {
 
 		if testResource.action == actionCopy {
 			if testResource.resType == resourceTypeFile {
-				err = utils.CopyFile(testResource.src, resourceDestInTemp)
+				var fileBackend filesstore.FileBackend
+				fileBackend, err = filesstore.NewFileBackend(filesstore.FileBackendSettings{DriverName: "local", Directory: ""})
```
wiggin77, line 156: Nit:  model.IMAGE_DRIVER_LOCAL  ?

services/filesstore/filesstore.go
```diff
@@ -8,8 +8,11 @@ import (
 	"time"
 
 	"github.com/pkg/errors"
+)
 
-	"github.com/mattermost/mattermost-server/v5/model"
+const (
+	driverS3    = "amazons3"
+	driverLocal = "local"
```
wiggin77, line 15: Nit:  model.IMAGE_DRIVER_LOCAL  ?

wiggin77 (approved):
LGTM  :+1:

services/filesstore/filesstore_test.go
```diff
@@ -14,16 +14,19 @@ import (
...
 	"github.com/stretchr/testify/require"
 	"github.com/stretchr/testify/suite"
+	"github.com/xtgo/uuid"
 
 	"github.com/mattermost/mattermost-server/v5/mlog"
-	"github.com/mattermost/mattermost-server/v5/model"
-	"github.com/mattermost/mattermost-server/v5/utils"
 )
 
+func RandomString() string {
```
streamer45, line 22: I think we can unexport this.

services/filesstore/s3store.go
```diff
@@ -42,19 +41,42 @@ const (
...
+
+func isFileExtImage(ext string) bool {
+	ext = strings.ToLower(ext)
+	for _, imgExt := range imageExtensions {
+		if ext == imgExt {
+			return true
+		}
+	}
+	return false
+}
```
streamer45, line 57: Does it make sense to use a map here as well?

model/config.go
```diff
@@ -1449,6 +1450,28 @@ func (s *FileSettings) SetDefaults(isUpdate bool) {
...
+		AmazonS3AccessKeyId:     *s.AmazonS3AccessKeyId,
+		AmazonS3SecretAccessKey: *s.AmazonS3SecretAccessKey,
+		AmazonS3Bucket:          *s.AmazonS3Bucket,
+		AmazonS3PathPrefix:      *s.AmazonS3PathPrefix,
+		AmazonS3Region:          *s.AmazonS3Region,
+		AmazonS3Endpoint:        *s.AmazonS3Endpoint,
+		AmazonS3SSL:             s.AmazonS3SSL == nil || *s.AmazonS3SSL,
+		AmazonS3SignV2:          s.AmazonS3SignV2 != nil && *s.AmazonS3SignV2,
+		AmazonS3SSE:             s.AmazonS3SSE != nil && *s.AmazonS3SSE && enableComplianceFeature,
+		AmazonS3Trace:           s.AmazonS3Trace != nil && *s.AmazonS3Trace,
```
streamer45, line 1471: Wondering if we should converge on either assuming the fields are not `nil` or check for `nil` on all of them. As it is it seems arbitrary (except maybe for `AmazonS3SSL` which defaults to `true`).

streamer45 (commented):
Very nicely done :+1: Left a couple of comments.

model/config.go
jespino: It is based on the pre-existing code, we can check everything there, but that, in theory is checked before save, in theory some of that fields can't be nil (can be empty string, or a value, but the pointer is to allow config.json to not-have the field). Once is in the system we expect some fields to be set. Probably we should ensure all of them and remove the nil check for everything else. (Out of the scope of this PR, a good thing to decide and do all over the code).

services/filesstore/filesstore_test.go
jespino: sure.

services/filesstore/s3store.go
jespino: :facepalm: sure, just copied the previous existing approach. I'll change it.

services/filesstore/filesstore_test.go
jespino: done

services/filesstore/s3store.go
jespino: done

streamer45 (approved):

jespino:
@lindalumitchell this requires smoke testing around attachments usage. I think this can be deferred to be covered by the RC testing.

lindalumitchell (approved):

agnivade (commented):
Everything looks good to me. I just have 2 thoughts:
- I am wondering whether the `ToFileBackendSettings` method is a good fit inside the config.go file. It seems like we are generating an input config for an external service, but we are doing it in a place which should not concern itself with what/how external packages should consume their config. Should the `FileBackendSettings` config be created at the call site itself? We can have a separate function to create the config in the app layer itself.
- The changing of `utils.CopyFile` which now mandates to first declare a struct seems to laborious to me from an API design standpoint. There is no state that is needed in that struct which is used to copy the input files, except the decision that they should be copied from the local filesystem. 
I am wondering if it makes sense to keep the function signature but just move it from `utils.CopyFile` to `filestore.CopyFIle`?

It doesn't block this PR from going in. But it can be iterated on later if we agree on these points.

jespino:
@agnivade about the `ToFileBackendSettings` can be move to the `app` layer, and I think makes a lot of sense (the mailservice also uses it, but we going to remove that soon, so It is not a big deal).

jespino:
I don't think  `filesstore` should expose that, is not related with the goal of the package, abstract the file storage using different backends. I removed that from utils because I didn't want to have repeated code, and I realize that using the filestorage for that was enough.

Also the `utils.CopyFile` it is only use on tests, so In any case I would remove it from `utils` to `testlib`, and then probably use something like `testlib.CopyFile` that under the hood does the filestore instanciation.

I going to try that.

agnivade:
Love the changes. What about the `ToFileBackendSettings` ? Did you plan to do it later or in this PR? Works for me either ways.

jespino:
@agnivade yes, I'll do it later, because I prefer to wait until the mail service is cleaned.

agnivade (approved):

PR #16818, opened 2021-01-28
Fix initialisms/app

SimonSimonB:
#### Summary
Fixes initialism errors, mainly in `app/`.

#### Ticket Link
Partly fixes https://github.com/mattermost/mattermost-server/issues/16623
(Since fixing all initialism errors requires many thousands of changes, I decided together with @agnivade to split this up into multiple PRs.)

#### Release Note
```release-note
NONE
```


agnivade:
Fantastic piece of work @SimonSimonB ! It seems like there are some merge conflicts in the PR. Once you fix them, we can move forward with reviewing it.

SimonSimonB:
Great! I opened this as a draft PR because I wanted to run the tests to make sure that the (many) changes so far don't break anything. Once that is confirmed, I will fix a few more initialism errors, resolve conflicts, and then mark this as ready for review.

agnivade:
Yeah I think even this PR is a bit too big to review. :P But it's okay if you have already done it. Just ensure to keep sending smaller chunks from next time onwards. :smile:

SimonSimonB:
That's fine! I am happy to put other changes in a separate PR.

(One thing to be aware of is that there will be a temporary loss of consistency in the code, with `userID` and `teamId` next to each other, for example. I will try to focus on fixing all errors in `app/` quickly, so that at least that part of the code base uses consistent and correct capitalization soon.)

agnivade:
That sounds great to me. Thanks a lot for spending your time on this @SimonSimonB ! Really appreciate it.

wiggin77:
This is great work @SimonSimonB!  One thing I want to make you aware of is the use of some fields in the translation files.  For example:  https://github.com/mattermost/mattermost-server/blob/0ea3b0f17d71889b77bad4247e73c68d73d0ae16/i18n/en.json#L3131

Any field names that get corrected will have to be corrected in the translation files as well.  Fortunately there only appears to be on instance for `TeamId`.

SimonSimonB:
Thanks for the heads-up, @wiggin77!

Just to clarify, I have so far not touched translation files and the string keys used to substitute data into strings in translation files. For example, I have changed the name of the Go variable `teamId` to `teamID` but not the string `"TeamId"` to `"TeamID"` in the following line of code:

`return model.NewAppError(..., "api.user.add_direct_channels_and_forget.failed.error", map[string]interface{}{"UserId": user.Id, "TeamId": teamID, "Error": err.Error()}, ...)`

For consistency, it'd probably be desirable to change string constants like `"TeamId"`, too, in the code and in the translation files. If I'm not mistaken, this can be done independently of fixing initialism errors in the Go code, so it might be worth opening a separate issue just for changing these strings.

agnivade:
That's great @SimonSimonB ! It should not be an issue then.

wiggin77 (approved):
LGTM :+1:   Great work!

agnivade (approved):
Wow, astounding job there @SimonSimonB !

Glad that it's not just a `sed -i 's/userId/userID/g'` :P I'm curious, how did you go about doing this? Did you use some automation to detect variables, or just selected each variable and invoked the editor's rename functionality?

agnivade:
Aaand we have more conflicts :(

SimonSimonB:
To do this, I did an interactive find & replace across all files in the `app/` folder. That way, I could check for each occurrence I was replacing whether it should actually be replaced, but it's pretty fast nevertheless since it usually takes less than a second to make that decision.

Thanks for reviewing!

PR #16819, opened 2021-01-28
Optimise creation of DM

nronas:
#### Summary
While working on https://mattermost.atlassian.net/browse/MM-32279 I stumble on the direct channel creation logic, and I think we can instead of running goroutines and synchronize them for fetching the users, optimize dm creation by fetching the users from the DB in a single call.

**Optimization gains:**

From 2 DB calls for each DM, to 1 DB for each DM. Essentially halving the DB pressure when creating DMs

#### Ticket Link
N/A

#### Release Note
```release-note
Optimize DM creation by fetching all users involved in a single db call.
```

## QA Acceptance Criteria
- [ ] Creating a DM should work as expected


store/sqlstore/user_store.go
```diff
@@ -325,6 +325,24 @@ func (us SqlUserStore) UpdateMfaActive(userId string, active bool) error {
 	return nil
 }
 
+func (us SqlUserStore) GetMany(ids []string) ([]*model.User, error) {
```
nronas, line 328: Didn't update the existing `Get` method since in a lot of places we rely on the error to be NoRowsErr.

app/channel.go
```diff
@@ -322,69 +322,70 @@ func (a *App) CreateChannel(channel *model.Channel, addMember bool) (*model.Chan
...
-		user, err := a.Srv().Store.User().Get(userId)
-		uc1 <- store.StoreResult{Data: user, NErr: err}
-		close(uc1)
-	}()
-	go func() {
-		user, err := a.Srv().Store.User().Get(otherUserId)
-		uc2 <- store.StoreResult{Data: user, NErr: err}
-		close(uc2)
-	}()
+	users, err := a.Srv().Store.User().GetMany([]string{userId, otherUserId})
```
jespino, line 364: This looks like a great idea, but it is not using the users cache and the get method does it, probably this is less performant in the real environment. We can load test this to verify it. Anyway I would suggest measure it, or add the cache usage to the new GetMany method.

agnivade: +1 to this. Let's add the `GetMany` method to the cache layer which gets the values from multiple ids.

store/sqlstore/user_store.go
agnivade, line 328: Please add a godoc to this, and some store tests. :)

app/channel.go
nronas: oh yeah, completely forgot about the cache. Will added to the cache.

store/sqlstore/user_store.go
nronas: Amended in https://github.com/mattermost/mattermost-server/pull/16819/commits/3ba1726037e26841ecf2b9ca8afd5e6d8a289c7a

app/channel.go
nronas: amended in https://github.com/mattermost/mattermost-server/pull/16819/commits/3ba1726037e26841ecf2b9ca8afd5e6d8a289c7a

store/localcachelayer/user_layer.go
```diff
@@ -155,3 +155,21 @@ func (s LocalCacheUserStore) Get(id string) (*model.User, error) {
 	s.rootStore.doStandardAddToCache(s.rootStore.userProfileByIdsCache, id, user)
 	return user, nil
 }
+
+// GetMany is a cache wrapper around the SqlStore method to get a user profiles by ids.
+// It checks if the user entries are present in the cache, returning the entries from cache
+// if it is present. Otherwise, it fetches the entries from the store and stores it in the
+// cache.
+func (s LocalCacheUserStore) GetMany(ids []string) ([]*model.User, error) {
```
jespino, line 163: This is just going back to the previous implementation. This layer is going to be executed always, so... GetMany is not going to be called ever, because here you use the `s.Get` to get the user.

My Expectation is to loop over the ids, try to get them from the cache, and do a `s.rootStore.GetMany` of the `notFoundIds`, then update the cache with that new obtained users, and return the final list.

nronas: Damn you are right, silly me. Will amend. i wanted to keep only the caching check for the Get and I somehow thought that it will do that.

nronas: That should do it 🤞 😄 https://github.com/mattermost/mattermost-server/pull/16819/commits/db9d12df918a14eadcc742ed11a43eb9b3688b51

store/localcachelayer/user_layer.go
```diff
@@ -155,3 +155,65 @@ func (s LocalCacheUserStore) Get(id string) (*model.User, error) {
...
+			cachedUsers = append(cachedUsers, cachedUser)
+			cachedUserIds = append(cachedUserIds, cachedUser.Id)
+			continue
+		}
+		if s.rootStore.metrics != nil {
+			s.rootStore.metrics.AddMemCacheMissCounter("Profile By Id", float64(1))
+		}
+	}
+
+	notCachedUserIds := arrayDiff(ids, cachedUserIds)
```
jespino: I think this "arrayDiff" is completely innecesary, you can ge the "notCachedUserIds" based on the cache misses , that is anything after the if clause in the for loop. You can do another append there. That way you avoid looping again in the same data.

store/localcachelayer/user_layer_test.go
```diff
@@ -258,3 +258,53 @@ func TestUserStoreGetCache(t *testing.T) {
...
+		assert.Contains(t, gotUsers, fakeUser)
+		assert.Contains(t, gotUsers, otherFakeUser)
+
+		gotUsers, err = cachedStore.User().GetMany([]string{fakeUser.Id, otherFakeUser.Id})
+		require.Nil(t, err)
+		assert.Len(t, gotUsers, 2)
+		mockStore.User().(*mocks.UserStore).AssertNumberOfCalls(t, "GetMany", 1)
+	})
+
+	t.Run("first call not cached, invalidate one user, and then check that one is cached and one is fetched from db", func(t *testing.T) {
```
jespino, line 291: Would be great to verify this is getting only the no-longer-cached element, and not query for all the requested ids.

store/localcachelayer/user_layer.go
nronas: Thanks for looking, I am clearly not seeing things right today. Will amend 😄

jespino (dismissed):
Small improvements needed, but otherwise is looking good :)

store/sqlstore/user_store.go
```diff
@@ -325,6 +325,25 @@ func (us SqlUserStore) UpdateMfaActive(userId string, active bool) error {
...
+	query := us.usersQuery.Where(sq.Eq{"Id": ids})
+	queryString, args, err := query.ToSql()
+	if err != nil {
+		return nil, errors.Wrap(err, "users_get_many_tosql")
+	}
+
+	var users []*model.User
+	if _, err := us.GetReplica().Select(&users, queryString, args...); err != nil {
+		if errors.As(err, sql.ErrNoRows) {
```
agnivade, line 338: This should be `errors.Is` I believe.

Also, any reason to check for no rows in a range query? We are not doing this anywhere else in user_store.

app/channel.go
```diff
@@ -322,69 +322,70 @@ func (a *App) CreateChannel(channel *model.Channel, addMember bool) (*model.Chan
...
+	// After we counted for direct channels with the same user, if we do not have two users then we failed to find one
+	if len(users) != 2 {
+		return nil, model.NewAppError("CreateDirectChannel", "api.channel.create_direct_channel.invalid_user.app_error", nil, fmt.Sprintf("No users found for ids: %s. %s", userId, otherUserId), http.StatusBadRequest)
 	}
-	user := result.Data.(*model.User)
 
-	result = <-uc2
-	if result.NErr != nil {
-		return nil, model.NewAppError("CreateDirectChannel", "api.channel.create_direct_channel.invalid_user.app_error", nil, otherUserId, http.StatusBadRequest)
+	user := users[0]
```
agnivade, line 383: A comment here explaining what the if condition is doing would be helpful. It's not immediately apparent why this is needed.

store/localcachelayer/user_layer.go
```diff
@@ -155,3 +155,58 @@ func (s LocalCacheUserStore) Get(id string) (*model.User, error) {
...
+// GetMany is a cache wrapper around the SqlStore method to get a user profiles by ids.
+// It checks if the user entries are present in the cache, returning the entries from cache
+// if it is present. Otherwise, it fetches the entries from the store and stores it in the
+// cache.
+func (s LocalCacheUserStore) GetMany(ids []string) ([]*model.User, error) {
+	// we are doing a loop instead of caching the full set in the cache because the number of permutations that we can have
+	// in this func is making caching of the total set not beneficial.
+	var cachedUsers []*model.User
+	var notCachedUserIds []string
+	uniqIds := newSet(ids)
```
agnivade, line 168: Nit: our new style guide suggests to use `uniqIDs`. https://developers.mattermost.com/contribute/server/style-guide/#initialismshttpsgithubcomgolanggowikicodereviewcommentsinitialisms

app/channel.go
```diff
@@ -322,69 +322,70 @@ func (a *App) CreateChannel(channel *model.Channel, addMember bool) (*model.Chan
 
 func (a *App) GetOrCreateDirectChannel(userId, otherUserId string) (*model.Channel, *model.AppError) {
```
agnivade, line 323: Nit on initialisms. Every `Id` should be `ID`

store/localcachelayer/user_layer.go
```diff
@@ -155,3 +155,58 @@ func (s LocalCacheUserStore) Get(id string) (*model.User, error) {
...
+		for _, user := range dbUsers {
+			s.rootStore.doStandardAddToCache(s.rootStore.userProfileByIdsCache, user.Id, user)
+			cachedUsers = append(cachedUsers, user)
+		}
+	}
+
+	return cachedUsers, nil
+}
+
+func newSet(elements []string) []string {
```
agnivade, line 200: We can use the zero-alloc in-place dedup algorithm here: https://github.com/golang/go/wiki/SliceTricks#in-place-deduplicate-comparable.

agnivade, line 200: I'd just call it `dedup`, or if you prefer to use it, then just `set`.

agnivade (commented):
One final round of comments. Looking great now!

store/localcachelayer/user_layer.go
nronas: For sure, but we are trading space for time here. The existing one is O(N) in time with O(N) in space and the suggested one is O(NlogN) in time and O(1) in space. I've chosen the faster in time one since the N is going to be considered small.

store/sqlstore/user_store.go
nronas: Totally 👍 . I've handled the error, because I think is more reasonable for this methdo to return an empty slice when the query didn't return anything(in a sense treating that error as a non-error)

agnivade: Yes, `Select` does a `Query`, so it will return zero rows if it doesn't find one.

app/channel.go
```diff
@@ -322,69 +322,74 @@ func (a *App) CreateChannel(channel *model.Channel, addMember bool) (*model.Chan
...
-			channel, err = a.createDirectChannel(userId, otherUserId)
-			if err != nil {
-				if err.Id == store.ChannelExistsError {
-					return channel, nil
-				}
-				return nil, err
-			}
-
-			a.WaitForChannelMembership(channel.Id, userId)
+	if nErr == nil {
```
jespino, line 325: This change looks like a simplification, but IMO is against the idiomatic way of writing it. The idiomatic way is "the happy path is at the lowest level of indentation", you are short circuiting there to return the happy path right away and handling the error after that. I think is more readable to handle the error cases, because is what I going to expect.

store/localcachelayer/user_layer.go
```diff
@@ -155,3 +157,62 @@ func (s LocalCacheUserStore) Get(id string) (*model.User, error) {
...
+// GetMany is a cache wrapper around the SqlStore method to get a user profiles by ids.
+// It checks if the user entries are present in the cache, returning the entries from cache
+// if it is present. Otherwise, it fetches the entries from the store and stores it in the
+// cache.
+func (s LocalCacheUserStore) GetMany(ids []string) ([]*model.User, error) {
+	// we are doing a loop instead of caching the full set in the cache because the number of permutations that we can have
+	// in this func is making caching of the total set not beneficial.
+	var cachedUsers []*model.User
+	var notCachedUserIds []string
+	uniqIDs := dedup(ids)
```
jespino, line 170: I think we can avoid the "double loop" here reusing the loop in the for, and storing the "already seen ids" in a map, that way you don't need a new function, and you don't need loop over the ids twice.

jespino (changes requested):
A couple of improvements suggestions, but now looks pretty good.

store/localcachelayer/user_layer.go
nronas: Sure thing, I find the existing implementation easier to read IMO, and also they have the same asymptotic complexity as the one you are suggesting. 2/5 here since we do not take a perf hit and the code is more readable.

app/channel.go
nronas: hmm, so are you suggesting to amended to something like this:

```go
	channel, nErr := a.Srv().Store.Channel().GetByName("", model.GetDMNameFromIds(userId, otherUserId), true)
	var nfErr *store.ErrNotFound
	if !errors.As(nErr, &nfErr) {
		return nil, model.NewAppError("GetOrCreateDirectChannel", "web.incoming_webhook.channel.app_error", nil, nErr.Error(), http.StatusInternalServerError)
	} else if nErr == nil {
		return channel, nil
	}
```

Since in the case that the error is notFound we need to create the channel. This method has 2 happy paths technically.

store/localcachelayer/user_layer.go
jespino: Yes, my concern wasn't about performance in this case, I don't think is going to have any performance implications there, I was thinking more about readability, for me is easier to read the whole code in one single place, but in this case the "dedup" function is pretty self-explanatory, so... I not 100% sure about that. So... however yo prefer, it is ok for me. :)

app/channel.go
nronas: actual the above will not work when the `nErr` is nil, so I will revert back to short-circuiting

store/localcachelayer/user_layer.go
agnivade: IMO existing implementation looks more readable.

store/storetest/user_store.go
```diff
@@ -296,6 +297,43 @@ func testUserStoreGet(t *testing.T, ss store.Store) {
...
+		users[i] = user
+	}
+	defer func() {
+		require.Nil(t, ss.User().PermanentDelete(users[0].Id))
+		require.Nil(t, ss.User().PermanentDelete(users[1].Id))
+	}()
+
+	t.Run("fetch empty ids", func(t *testing.T) {
+		res, err := ss.User().GetMany([]string{})
+		require.Nil(t, err)
```
agnivade, line 317: It is more correct to use `require.(No)Error` family of functions to check for variables of type `error`.

agnivade (approved):
Nicely done! Good to go from my side

store/storetest/user_store.go
nronas: I've tried to keep it consistent with the rest of the tests in this file. Will amend all of those to use `NoError`

app/channel.go
jespino: No, I'm suggesting something like keeping the old "big if clause" wrapping all the "error handling" case.

For me here, the happy path is "I try to get the existing channel, and is there, so I don't have to do anything, just returning the channel" (that should be at the lowest level of indentation), then you have "If I'm not able to get the channel, how I handle that case?" that should be inside the if body.

But this is again my own interpretation of how it should be. I always expect the `if err != nil` after something an error, so... having that `if err == nil`, maybe is going to confuse me (and probably I directly don't read that, a go straight forward to the if body expecting the error handling).

nronas: Have no strong pref here, let me see what I can do in order to make it more readable, since IMO  the "big if clause" I do not think provides higher readability, maybe if I use local functions we can clean this up a bit. Let me see how that looks and we can take it from there.

nronas: @jespino how that looks https://github.com/mattermost/mattermost-server/pull/16819/commits/5085aa64be956bb1c1618a869e1fe4c6338b99ad ?

store/storetest/user_store.go
agnivade: Yes, this is an artifact from the apperror migration from model.AppError to error. We actually need to change all `require.Nil` of error to `require.NoError`.

nronas: Already done for this file :)

angeloskyratzakos (approved):
transferring approvals

jespino (approved):
LGTM

prapti (approved):
Tested and verified in local and the PR server that DM still works as expected (with DM + on LHS, profile popover, channel switcher, members in channel header, RHS, with normal users and bots).
Thank you for including the QA acceptance criteria @nronas, that helped!
Approving.

PR #16822, opened 2021-01-28
MM-28090 User settings api when ldap sync

maxerenberg:
#### Summary
This pull request prevents users from modifying some of their user settings via the Mattermost API if they authenticate via LDAP, SAML or OAUTH. It aims to have identical restrictions to those of the web UI.

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-28090

#### Release Note
```release-note
* For LDAP and SAML users, the following fields cannot be changed via the API if the corresponding
  LDAP/SAML attributes have been set: first name, last name, position, nickname, email, profile picture
* For OAUTH users (i.e. Gitlab, Google, Office365 and OpenID), the following fields cannot be changed
  via the API: first name, last name
* All users who authenticate via a method other than email cannot change their username via the API
```

#### PR for unit tests
https://github.com/mattermost/enterprise/pull/898


api4/user.go
```diff
@@ -465,8 +465,19 @@ func setProfileImage(c *Context, w http.ResponseWriter, r *http.Request) {
 		auditRec.AddMeta("filename", imageArray[0].Filename)
 	}
 
-	if user, err := c.App.GetUser(c.Params.UserId); err == nil {
-		auditRec.AddMeta("user", user)
+	user, err := c.App.GetUser(c.Params.UserId)
+	if err != nil {
+		c.SetInvalidParam("user_id")
```
sbishel, line 470: This is not the correct error to return. 
'Invalid or missing user_id in request body.'
It should return 
{
    "id": "api.context.invalid_url_param.app_error",
    "translation": "Invalid or missing {{.Name}} parameter in request URL."
  },

sbishel (changes requested):
Just some minor issues...looking good.

sbishel (approved):
Looks good.

model/user.go
```diff
@@ -724,6 +727,29 @@ func (u *User) GetPreferredTimezone() string {
...
+	"Nickname":  func(p *UserPatch) **string { return &p.Nickname },
+	"Email":     func(p *UserPatch) **string { return &p.Email },
+	"Position":  func(p *UserPatch) **string { return &p.Position },
+	"Username":  func(p *UserPatch) **string { return &p.Username },
+}
+
+func (u *UserPatch) SetField(fieldName string, fieldValue string) {
+	*patchFieldSelector[fieldName](u) = &fieldValue
+}
+
```
hahmadia, line 752: Not sure if I'm the only one but I'd opt to sacrifice lines of code and expand on this in return for readability. It took me a while to understand and although it's compact, I'm not sure how readable it is (and could just be me 😅 ).
Would it be cool to change it to something like the following (and helping us get rid of the global variable too).
```suggestion
func (u *UserPatch) SetField(fieldName string, fieldValue string) {
	switch fieldName {
	case "FirstName":
		u.FirstName = &fieldValue
	case "LastName":
		u.LastName = &fieldValue
	case "Nickname":
		u.Nickname = &fieldValue
	case "Email":
		u.Email = &fieldValue
	case "Position":
		u.Position = &fieldValue
	case "Username":
		u.Username = &fieldValue
	}
}

```
I know this a little bit more drawn out and less compact then what you have but I think it's more readable and easier to catch on to.

api4/user.go
```diff
@@ -1131,6 +1151,37 @@ func updateUser(c *Context, w http.ResponseWriter, r *http.Request) {
 	w.Write([]byte(ruser.ToJson()))
 }
 
+// checkProviderAttributes returns the empty string if the patch can be applied without
+// overriding attributes set by the user's login provider; otherwise, the name of the offending
+// field is returned.
+func checkProviderAttributes(a app.AppIface, ouser *model.User, patch *model.UserPatch) string {
```
hahmadia, line 1157: nit : Maybe we can just call it `user` or `originalUser` instead of `ouser` here?
```suggestion
func checkProviderAttributes(a app.AppIface, originalUser *model.User, patch *model.UserPatch) string {
```

api4/user.go
```diff
@@ -1131,6 +1151,37 @@ func updateUser(c *Context, w http.ResponseWriter, r *http.Request) {
...
+	tryingToChange := func(patchValue *string, userValue *string) bool {
```
hahmadia, line 1158: nit: lets change the order here to match the function to give the idea of `trying to change users value to patch value`.
```suggestion
	tryingToChange := func(userValue *string, patchValue *string) bool {
```

api4/user.go
```diff
@@ -1131,6 +1151,37 @@ func updateUser(c *Context, w http.ResponseWriter, r *http.Request) {
...
+		return patchValue != nil && *patchValue != *userValue
+	}
+	LdapSettings := &a.Config().LdapSettings
+	SamlSettings := &a.Config().SamlSettings
+
+	conflictField := ""
+	if a.Ldap() != nil &&
```
hahmadia, line 1165: We could do the following for more clarity. Personal preference here but it may be more clear.
```suggestion
	if LdapSettings.Enable &&
```

api4/user.go
```diff
@@ -1131,6 +1151,37 @@ func updateUser(c *Context, w http.ResponseWriter, r *http.Request) {
...
+		(ouser.IsLDAPUser() || (ouser.IsSAMLUser() && *SamlSettings.EnableSyncWithLdap)) {
+		conflictField = a.Ldap().CheckProviderAttributes(LdapSettings, ouser, patch)
+	} else if a.Saml() != nil && ouser.IsSAMLUser() {
```
hahmadia, line 1168: We could do the following for more clarity. Personal preference here but it may be more clear.
```suggestion
	} else if SamlSettings.Enable && ouser.IsSAMLUser() {
```

api4/user.go
```diff
@@ -1131,6 +1151,37 @@ func updateUser(c *Context, w http.ResponseWriter, r *http.Request) {
...
+		conflictField = a.Saml().CheckProviderAttributes(SamlSettings, ouser, patch)
+	} else if ouser.IsOAuthUser() {
+		if tryingToChange(patch.FirstName, &ouser.FirstName) ||
+			tryingToChange(patch.LastName, &ouser.LastName) {
+			conflictField = "full name"
+		}
+	}
```
hahmadia, line 1175: We can avoid nested if statements here
```suggestion
	} else if ouser.IsOAuthUser() && (tryingToChange(patch.FirstName, &ouser.FirstName) || tryingToChange(patch.LastName, &ouser.LastName)) {
			conflictField = "full name"
	}
```

api4/user.go
```diff
@@ -1131,6 +1151,37 @@ func updateUser(c *Context, w http.ResponseWriter, r *http.Request) {
...
+	// If any login provider is used, then the username may not be changed
+	if ouser.AuthService != "" {
+		if tryingToChange(patch.Username, &ouser.Username) && conflictField == "" {
+			conflictField = "username"
+		}
+	}
```
hahmadia, line 1181: I think this can be move to the top of the function right before you check if ldap is enabled and etc. I would say you can move there and do something like this
```suggestion
	// If any login provider is used, then the username may not be changed
	if ouser.AuthService != "" && tryingToChange(patch.Username, &ouser.Username) {
	        return "username"
	}
```
It makes it clear to say something like "If you're using any auth service and you are trying to change username, we will return since you're not allowed and not check for specifics cause the rest of the code doesn't matter" Hopefully that made sense. I feel this order will make it more clear from a developers standpoint.

hahmadia (changes requested):
Nice job! Just some questions and small adjustments. Thanks

api4/user.go
maxerenberg: I personally prefer to keep `a.Ldap() ~= nil` since it guarantees that the LDAP interface is present.

api4/user.go
maxerenberg: I'd prefer to keep it nested because that way we have one if clause for each authentication type. If we decide to add another clause for a new authentication type in the future, OAuth users wouldn't have to check the new clause.

api4/user.go
maxerenberg: Good idea. I like this a lot better.

hahmadia (approved):
Nice changes! LGTM 👍

mkraft (approved):
LGTM! 👍🏻

furqanmlk (approved):

adamjclarkson (approved):

PR #16824, opened 2021-01-28
[GH-15906][MM-22844] Redesign welcome and verify email.

jp0707:
#### Summary
Redesign welcome and verify email. Rest of the emails are untouched.

This change adds MJML into the build tool chain. Relevant discussion at https://community.mattermost.com/core/pl/jz9ry3cp97y4bf1jokxff4ui8w

#### Ticket Link
https://github.com/mattermost/mattermost-server/issues/15906


app/server.go
```diff
@@ -365,7 +365,7 @@ func NewServer(options ...Option) (*Server, error) {
 		}
 	}
 
-	if htmlTemplateWatcher, err2 := utils.NewHTMLTemplateWatcher("templates"); err2 != nil {
+	if htmlTemplateWatcher, err2 := utils.NewHTMLTemplateWatcher("templates", "templates/compiled"); err2 != nil {
```
jespino, line 368: some people relay on modified version of the template files in their own systems, so this this changes of routes can change unexpectedly the behavior for them. Can we maintain the same paths for the generated templates?

services/mailservice/mail.go
```diff
@@ -307,9 +309,14 @@ func sendMailUsingConfigAdvanced(mail mailData, config *model.Config, enableComp
 func SendMail(c smtpClient, mail mailData, fileBackend filesstore.FileBackend, date time.Time) *model.AppError {
 	mlog.Debug("sending mail", mlog.String("to", mail.smtpTo), mlog.String("subject", mail.subject))
 
-	htmlMessage := "\r\n<html><body>" + mail.htmlBody + "</body></html>"
+	htmlMessage := ""
+	if (mail.isBody) {
```
jespino, line 313: This is a bit of glue code that is affecting everything else, can we just add the `<html><body>` tags directly to all the old templates, and leave the code as is? There is not reason after everything is migrated to have this differentiation. so I think is better to just leave the code as should be at the end.

templates/welcome.mjml
```diff
@@ -0,0 +1,26 @@
+<mjml>
```
jespino, line 1: I love this, way way way nicer :)

utils/html.go
```diff
@@ -26,10 +26,7 @@ type HTMLTemplateWatcher struct {
 	stopped   chan struct{}
 }
 
-func NewHTMLTemplateWatcher(directory string) (*HTMLTemplateWatcher, error) {
-	templatesDir, _ := fileutils.FindDir(directory)
-	mlog.Debug("Parsing server templates", mlog.String("templates_directory", templatesDir))
-
+func NewHTMLTemplateWatcher(directories ...string) (*HTMLTemplateWatcher, error) {
```
jespino, line 29: If we just generate the old files with the exact same path, we don't need anything changing here.

app/email.go
```diff
@@ -201,17 +204,18 @@ func (es *EmailService) sendWelcomeEmail(userId string, email string, verified b
 		map[string]interface{}{"SiteName": es.srv.Config().TeamSettings.SiteName,
 			"ServerURL": serverURL})
 
-	bodyPage := es.newEmailTemplate("welcome_body", locale)
+	bodyPage := es.newEmailTemplate("welcome", locale)
```
jespino, line 207: I know is no longer "body" because you are generating the whole HTML, but as I said in other comment, we have users that relays on modifications on this files, I think i prefer to keep the filename exactly the same for backward compatibility.

app/email.go
```diff
@@ -154,14 +154,17 @@ func (es *EmailService) sendVerifyEmail(userEmail, locale, siteURL, token, redir
 	subject := T("api.templates.verify_subject",
 		map[string]interface{}{"SiteName": es.srv.Config().TeamSettings.SiteName})
 
-	bodyPage := es.newEmailTemplate("verify_body", locale)
+	bodyPage := es.newEmailTemplate("verify", locale)
```
jespino, line 157: same here.

agnivade:
I just finished reviewing a load of PRs and need to get back to other stuff. If this requires quick feedback, then please feel free to assign someone else.

reflog (approved):
Looks good, except for @jespino comments

wiggin77:
This looks great @jp0707.  One question:  when a developer runs `make run-server` it will build the email templates each time.  What happens if they do no have mjml installed?  I can see where it gets installed for CI but not developers running locally.

app/server.go
jp0707: Yes done. But now this requires us to mention each generated html file in `.gitignore`. In future when all email templates have migrated to mjml then we can simplify this with `templates/*.html`

services/mailservice/mail.go
jp0707: Great idea! Done

utils/html.go
jp0707: Yup, done.

app/email.go
jp0707: Yes done.

app/email.go
jp0707: done

jp0707:
> One question: when a developer runs `make run-server` it will build the email templates each time. What happens if they do no have mjml installed? I can see where it gets installed for CI but not developers running locally.

@wiggin77 We have added this into `check-prereqs`. So in `make run` it will check pre-requisites and fail with error message if it's not already installed. In `make run-server`, it's already not running `check-prereqs` -- i.e. it's not checking for other things like node, npm, go, docker etc any way.

Should we add this to documentation for setting up the machine? If so, please let me know which file and I can do it.

wiggin77:
> it's not checking for other things like node, npm, go, docker etc any way.

It's true those dependencies are not checked, but developers commonly have them installed already. Developers typically won't have a specific and uncommon npm package installed and every new developer will trip over it going forward.

IMO, this is more like the packages and images that are automatically fetched for Go and Docker.  Or the npm packages automatically fetched for webapp.

I'm suggesting a simple `npm install ...` as part of `make run-server` so most developers won't need to concern themselves with mjml.

app/server.go
jespino: Yes, that is ok for now.

jespino:
@wiggin77 @jp0707 a solution for this would be not "enforce" the generation of the files, and just submit the generated files too, and check in the CI that the generated files matches the submitted files (We do that for other generated go files, like the layers or the mocks).

jp0707:
@wiggin77 @jespino Implemented option 2. Please take a look. Can you also trigger a CI build so I can get a feedback on whether that's working -- haven't really tested changes on CI yet.

wiggin77 (approved):
LGTM :+1:    Great work @jp0707.  There is one code scan warning regarding a link - can you double check that?

jespino (approved):
Looks good to me. Thanks @jp0707! awesome work! :tada:

jp0707:
@matthewbirtch Here is how the email looks. Please take a look.
verify_body
verify_body
welcome_body
welcome_body

matthewbirtch:
> @matthewbirtch Here is how the email looks. Please take a look.

@jp0707 is there any way you can send me test emails so I can review the emails at actual size in my inbox and inspect elements/text sizes? It's tough to review with just screenshots.

esethna:
@jp0707 thanks so much for your great work here!

matthewbirtch (changes requested):
Thanks for the contributions @jp0707. I reviewed the welcome email and most of my changes below apply to both emails since they share the same layout. Let me know if you have any questions as you work through the below. Hopefully it's all clear.

1. The email should be contained in a white container with 8px radius rounded corners. There should be a grey background (`#F7F7F7`) for everything outside the main container. And this container should be 32px from the edges (see screenshot below.) 

**32px margin all the way around:**
image 

**how it should look with the browser width expanded**
image

**for smaller viewports:**
  - between 540px and 401px  we can reduce this margin to 16px
  - smaller than 400px in width remove the margin (and border radius) altogether

2. The logo needs to to be reduced by 6px. (the logo should be 24px from the top based on the design).
3. Need to add 20px more space between the logo the main headline
4. The `font-weight` on the main headline is correctly set to `600`, however, it doesn't look like that weight is being imported in the head of the HTML.
5. Also, the `welcome.html` file you provided has "Verify your email address" as the headline. Meanwhile the `verify.html` file provided had "Welcome to the team" as the headline. Should those be switched?
6. The main 'Verify Email' button is smaller than designed. It should be 48px in height and the font size should be 16px
7. The button also appears to be using the wrong font. It seems to be using `Ubuntu` 
8. On smaller viewports, we should reduce the padding on the "This email address was used…" body copy. See screenshot below



9. Also, `Ubuntu` is being used in the imported in the head of the HTML. We shouldn't need this font at all. It should all be `Open Sans`
8. Reduce the space between the 'Verify Email' button and the illustration by 24px;
9. In the "Download the desktop and mobile apps" area, can we tighten up the spacing between lines to match design better?

image

10. Need some styling updates on the "Download" button as well:
  - Font on the button seems to be using `Ubuntu`
  - Height should be 40px - looks like its
  - `font-size: 14px;`
  - `line-height: 14px;`
  - `font-weight: 600;`
  - font should be `Open Sans` not `Ubuntu`
  - The `font-weight` on "Download" is not rendering as `600` - likely for the same reason as stated above.
11. Reduce the space between the footer copyright line and the content above by about 32px.
12. There seems to be a `<br>` tag in front of the copyright line for some reason? this might solve some of the spacing in the note above.

jespino:
Hi @matthewbirtch, how much of this is needed to be changed in this PR? I mean, this is including all the infrastructure needed for sending emails using the new templates, so... this PR is complex enough by itself. Can we merge this as is now and do a second PR to fix the problems there? I guess this changes are more about "pixel perfect" than something conceptually broken. And actually, the current changes are a huge improvement over the previous existing emails anyway. So.. what do you think? can we merge this and iterate right away on fixing that things?

matthewbirtch:
> Hi @matthewbirtch, how much of this is needed to be changed in this PR?

Isn't the whole point of this PR to implement redesigned emails? I'm not sure why we would merge it if it's not complete. I spoke with @jp0707 and she is confident she can fix the above.

jp0707:
@matthewbirtch Done. Fixed all the issues. Sent you generated html files on mattermost community

jespino:
@matthewbirtch Improving he emails is the whole point, but it doesn't mean needs to be done everything in one shot. Actually, this PR includes the improvements on some of the emails, and the improvements on the emails generation system. If we try to make it all in one shot, it takes more time and maintenance of the branches. I'm not suggesting merging something broken, but because it includes the initial version, I think we should aim for good, not for perfection, and we achieve the perfection later in a subsequent PR.

Anyway, that is my opinion, and I asked you because of that, to know how you feel about merging it and iterating.

matthewbirtch:
> @matthewbirtch Improving he emails is the whole point, but it doesn't mean needs to be done everything in one shot. Actually, this PR includes the improvements on some of the emails, and the improvements on the emails generation system. If we try to make it all in one shot, it takes more time and maintenance of the branches. I'm not suggesting merging something broken, but because it includes the initial version, I think we should aim for good, not for perfection, and we achieve the perfection later in a subsequent PR.
> 
> Anyway, that is my opinion, and I asked you because of that, to know how you feel about merging it and iterating.

It should be fine. @jp0707 has already fixed the issues with only a couple minor updates. We chatted in Mattermost about a few things, but I imagine it will be updated here momentarily. I don't think this will slow things down.

srkgupta (approved):
All CodeQL alerts triggered are dismissed. Looks good from security perspective.

matthewbirtch (approved):
Nice work @jp0707!

ogi-m (approved):
Tested and passed 👍🏻

kalvdans:
After this change, Thunderbird warns about external content in the email. Is the exact font in the emails that important?

bild

PR #16827, opened 2021-01-29
MM-32299 Add sync'd users to team & ChannelMembers

wiggin77:
#### Summary
When a post or reaction from a remote side is received, it includes any users involved in the post/reaction.  A dummy user account is created locally which cannot be logged into.  

This PR adds the user to the shared channel's team and ChannelMembers table.

**To be merged into Shared Channel feature branch.**

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-32299

#### Release Note
```release-note
NONE
```


services/sharedchannel/sync_recv.go
```diff
@@ -197,6 +197,21 @@ func (scs *Service) upsertSyncUser(user *model.User, rc *model.RemoteCluster) (*
 		}
 		userSaved = userUpdated.New
 	}
+
+	// add user to team. We do this here regardless of whether the user was
+	// just created or patched since there are three steps to adding a user
+	// (insert rec, add to team, add to channel) and any one could fail.
+	// Instead of undoing what succeeded on any failure we simply do all steps each
+	// time. AddUserToChannel & AddUserToTeamByTeamId do not error if user already
```
agnivade, line 205: > AddUserToChannel & AddUserToTeamByTeamId do not error if user already
	// added and exit quickly.

IIUC, this will query the DB for every user in every post right? So even if the user is same for 10 posts, it will issue 10 DB queries. It's out of scope for this PR, but perhaps we might look into some batching strategies to handle each single batch of posts?
- find out all set of unique users from the current batch of posts
- batch add them
- Then process all posts/reactions.

agnivade (approved):

services/sharedchannel/sync_recv.go
wiggin77: The sender keeps tracks of what users have been sync'd already and are not supposed to send them redundantly.  That said, there is an issue because more user sync's happen then expected, and it looks like the reason is because Users.Updated gets set a lot more often than expected.  I will address this for V1 by tracking the handful of fields we care about for Users and only sync if they change.  On the receiving side we'll redundantly add users to team and channel to keep error handling simple, but this will be pretty rare.

wiggin77: ... and the sender will cache the results of sender lookup for any batch of posts being processed.

agnivade: Ah I see, so the dedup and batching is done on the sender side?

wiggin77: Yes.  Besides not wanting to hit the database needlessly we also don't want to transmit the bytes needlessly so the sender might dedup.

mgdelacroix (approved):
LGTM! :+1:

PR #16835, opened 2021-02-01
Custom status feature

chetanyakan:
#### Summary

- Create a system console setting for custom status feature enabled.
- Create server APIs for custom status.
- Create slash commands to set/unset status.
- Add logic to set/unset status.

- ~~Note - Call to `a.sendUpdatedUserEvent(*user)` can be removed after https://github.com/mattermost/mattermost-server/pull/16604 is merged.~~

#### Release Note

* Config changes - additions
* API additions—new endpoints
* Websocket additions or changes

```release-note
Added new API endpoints PUT /api/v4/users/<id>/status/custom, DELETE /api/v4/users/<id>/status/custom, and DELETE  /api/v4/users/<id>/status/custom/recent.
```

```release-note
Added a new config setting TeamSettings.EnableCustomUserStatuses.
```

### DIFF
--- a/api4/status.go
+++ b/api4/status.go
@@ -118,7 +118,7 @@ func updateUserCustomStatus(c *Context, w http.ResponseWriter, r *http.Request) 
 	}
 
 	customStatus := model.CustomStatusFromJson(r.Body)
-	if customStatus == nil {
+	if customStatus == nil || (customStatus.Text == "" && customStatus.Emoji == "") {
 		c.SetInvalidParam("custom_status")
 		return
 	}
--- a/api4/status.go
+++ b/api4/status.go
@@ -118,7 +118,7 @@ func updateUserCustomStatus(c *Context, w http.ResponseWriter, r *http.Request) 
 	}
 
 	if !c.App.Config().FeatureFlags.CustomUserStatuses || !*c.App.Config().TeamSettings.EnableCustomUserStatuses {
-		c.Err = model.NewAppError("updateUserCustomStatus", "api.custom_status.disabled", nil, "", http.StatusForbidden)
+		c.Err = model.NewAppError("updateUserCustomStatus", "api.custom_status.disabled", nil, "", http.StatusNotImplemented)
 		return
 	}
 
@@ -150,7 +150,7 @@ func removeUserCustomStatus(c *Context, w http.ResponseWriter, r *http.Request) 
 	}
 
 	if !c.App.Config().FeatureFlags.CustomUserStatuses || !*c.App.Config().TeamSettings.EnableCustomUserStatuses {
-		c.Err = model.NewAppError("removeUserCustomStatus", "api.custom_status.disabled", nil, "", http.StatusForbidden)
+		c.Err = model.NewAppError("removeUserCustomStatus", "api.custom_status.disabled", nil, "", http.StatusNotImplemented)
 		return
 	}
 
@@ -174,7 +174,7 @@ func removeUserRecentCustomStatus(c *Context, w http.ResponseWriter, r *http.Req
 	}
 
 	if !c.App.Config().FeatureFlags.CustomUserStatuses || !*c.App.Config().TeamSettings.EnableCustomUserStatuses {
-		c.Err = model.NewAppError("removeUserRecentCustomStatus", "api.custom_status.disabled", nil, "", http.StatusForbidden)
+		c.Err = model.NewAppError("removeUserRecentCustomStatus", "api.custom_status.disabled", nil, "", http.StatusNotImplemented)
 		return
 	}
 
--- a/api4/status.go
+++ b/api4/status.go
@@ -117,6 +117,11 @@ func updateUserCustomStatus(c *Context, w http.ResponseWriter, r *http.Request) 
 		return
 	}
 
+	if !c.App.Config().FeatureFlags.CustomUserStatuses || !*c.App.Config().TeamSettings.EnableCustomUserStatuses {
+		c.Err = model.NewAppError("updateUserCustomStatus", "api.custom_status.disabled", nil, "", http.StatusForbidden)
+		return
+	}
+
 	customStatus := model.CustomStatusFromJson(r.Body)
 	if customStatus == nil || (customStatus.Text == "" && customStatus.Emoji == "") {
 		c.SetInvalidParam("custom_status")
@@ -144,6 +149,11 @@ func removeUserCustomStatus(c *Context, w http.ResponseWriter, r *http.Request) 
 		return
 	}
 
+	if !c.App.Config().FeatureFlags.CustomUserStatuses || !*c.App.Config().TeamSettings.EnableCustomUserStatuses {
+		c.Err = model.NewAppError("removeUserCustomStatus", "api.custom_status.disabled", nil, "", http.StatusForbidden)
+		return
+	}
+
 	if !c.App.SessionHasPermissionToUser(*c.App.Session(), c.Params.UserId) {
 		c.SetPermissionError(model.PERMISSION_EDIT_OTHER_USERS)
 		return
@@ -160,6 +170,11 @@ func removeUserCustomStatus(c *Context, w http.ResponseWriter, r *http.Request) 
 func removeUserRecentCustomStatus(c *Context, w http.ResponseWriter, r *http.Request) {
 	c.RequireUserId()
 	if c.Err != nil {
+		return
+	}
+
+	if !c.App.Config().FeatureFlags.CustomUserStatuses || !*c.App.Config().TeamSettings.EnableCustomUserStatuses {
+		c.Err = model.NewAppError("removeUserRecentCustomStatus", "api.custom_status.disabled", nil, "", http.StatusForbidden)
 		return
 	}
 
--- a/api4/status.go
+++ b/api4/status.go
@@ -128,6 +128,7 @@ func updateUserCustomStatus(c *Context, w http.ResponseWriter, r *http.Request) 
 		return
 	}
 
+	customStatus.TrimMessage()
 	err := c.App.SetCustomStatus(c.Params.UserId, customStatus)
 	if err != nil {
 		c.SetInvalidParam("custom_status")
--- a/api4/status.go
+++ b/api4/status.go
@@ -131,7 +131,7 @@ func updateUserCustomStatus(c *Context, w http.ResponseWriter, r *http.Request) 
 	customStatus.TrimMessage()
 	err := c.App.SetCustomStatus(c.Params.UserId, customStatus)
 	if err != nil {
-		c.SetInvalidParam("custom_status")
+		c.Err = err
 		return
 	}
 
@@ -150,7 +150,7 @@ func removeUserCustomStatus(c *Context, w http.ResponseWriter, r *http.Request) 
 	}
 
 	if err := c.App.RemoveCustomStatus(c.Params.UserId); err != nil {
-		c.SetInvalidParam("custom_status")
+		c.Err = err
 		return
 	}
 
@@ -175,7 +175,7 @@ func removeUserRecentCustomStatus(c *Context, w http.ResponseWriter, r *http.Req
 	}
 
 	if err := c.App.RemoveRecentCustomStatus(c.Params.UserId, recentCustomStatus); err != nil {
-		c.SetInvalidParam("custom_status")
+		c.Err = err
 		return
 	}
 
--- a/api4/status.go
+++ b/api4/status.go
@@ -13,6 +13,9 @@ func (api *API) InitStatus() {
 	api.BaseRoutes.User.Handle("/status", api.ApiSessionRequired(getUserStatus)).Methods("GET")
 	api.BaseRoutes.Users.Handle("/status/ids", api.ApiSessionRequired(getUserStatusesByIds)).Methods("POST")
 	api.BaseRoutes.User.Handle("/status", api.ApiSessionRequired(updateUserStatus)).Methods("PUT")
+	api.BaseRoutes.User.Handle("/status/custom", api.ApiSessionRequired(updateUserCustomStatus)).Methods("PUT")
+	api.BaseRoutes.User.Handle("/status/custom", api.ApiSessionRequired(removeUserCustomStatus)).Methods("DELETE")
+	api.BaseRoutes.User.Handle("/status/custom/recent", api.ApiSessionRequired(removeUserRecentCustomStatus)).Methods("DELETE")
 }
 
 func getUserStatus(c *Context, w http.ResponseWriter, r *http.Request) {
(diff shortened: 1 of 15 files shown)


esethna:
Test server from brightscout that includes webapp, redux and server changes: http://ec2-3-137-148-232.us-east-2.compute.amazonaws.com:8065/signup_user_complete/?id=559fit8sj7bm8n7mcpqcz9cy5y

chetanyakan:
Rebased to the head of master.

reflog (commented):
.

api4/status.go
```diff
@@ -13,6 +13,9 @@ func (api *API) InitStatus() {
 	api.BaseRoutes.User.Handle("/status", api.ApiSessionRequired(getUserStatus)).Methods("GET")
 	api.BaseRoutes.Users.Handle("/status/ids", api.ApiSessionRequired(getUserStatusesByIds)).Methods("POST")
 	api.BaseRoutes.User.Handle("/status", api.ApiSessionRequired(updateUserStatus)).Methods("PUT")
+	api.BaseRoutes.User.Handle("/status/custom", api.ApiSessionRequired(updateUserCustomStatus)).Methods("PUT")
+	api.BaseRoutes.User.Handle("/status/custom", api.ApiSessionRequired(removeUserCustomStatus)).Methods("DELETE")
+	api.BaseRoutes.User.Handle("/status/custom/recent", api.ApiSessionRequired(removeUserRecentCustomStatus)).Methods("DELETE")
```
reflog, line 18: A PR to https://github.com/mattermost/mattermost-api-reference is needed to document new endpoints, and to fix the failing CI

api4/status.go
```diff
@@ -107,3 +110,74 @@ func updateUserStatus(c *Context, w http.ResponseWriter, r *http.Request) {
...
+		return
+	}
+
+	if !c.App.SessionHasPermissionToUser(*c.App.Session(), c.Params.UserId) {
+		c.SetPermissionError(model.PERMISSION_EDIT_OTHER_USERS)
+		return
+	}
+
+	if err := c.App.RemoveCustomStatus(c.Params.UserId); err != nil {
+		c.SetInvalidParam("custom_status")
```
reflog, line 153: seems like this should be a different error

api4/status.go
```diff
@@ -107,3 +110,74 @@ func updateUserStatus(c *Context, w http.ResponseWriter, r *http.Request) {
...
+		return
+	}
+
+	if !c.App.SessionHasPermissionToUser(*c.App.Session(), c.Params.UserId) {
+		c.SetPermissionError(model.PERMISSION_EDIT_OTHER_USERS)
+		return
+	}
+
+	if err := c.App.RemoveRecentCustomStatus(c.Params.UserId, recentCustomStatus); err != nil {
+		c.SetInvalidParam("custom_status")
```
reflog, line 178: seems like this should be a different error

model/config.go
```diff
@@ -1831,6 +1831,7 @@ type TeamSettings struct {
 	EnableOpenServer                                          *bool    `access:"authentication"`
 	EnableUserDeactivation                                    *bool    `access:"experimental"`
 	RestrictCreationToDomains                                 *string  `access:"authentication"`
+	EnableCustomUserStatuses                                  *bool    `access:"site"`
```
reflog, line 1834: Each new feature setting should be behind a Feature Flag, please following instructions here: https://developers.mattermost.com/contribute/server/feature-flags/

model/config.go
```diff
@@ -1884,6 +1885,10 @@ func (s *TeamSettings) SetDefaults() {
 		s.RestrictCreationToDomains = NewString("")
 	}
 
+	if s.EnableCustomUserStatuses == nil {
+		s.EnableCustomUserStatuses = NewBool(true)
```
reflog, line 1889: Shouldn't it be off by default?

reflog (changes requested):
looks great, added some notes

model/config.go
chetanyakan: @reflog I assumed the default value is true by the value set here: https://mattermost.atlassian.net/wiki/spaces/CS/pages/1107034113/Configuration+Setting

But maybe that wasn't meant to be the default value. I think it makes sense to have this off by default.
What do you think?

esethna: @reflog @chetanyakan the feature should certainly be **enabled by-default**. Admins can disable it if they desire

app/slashcommands/command_custom_status.go
```diff
@@ -0,0 +1,81 @@
...
+			ResponseType: model.COMMAND_RESPONSE_TYPE_EPHEMERAL,
+			Text:         args.T("api.command_custom_status.clear.success"),
+		}
+	}
+
+	customStatus := &model.CustomStatus{
+		Emoji: DefaultCustomStatusEmoji,
+		Text:  message,
+	}
+	firstEmojiLocations := model.EMOJI_PATTERN.FindIndex([]byte(message))
```
chetanyakan, line 61: @reflog @jespino Maybe you can pitch in here.

There is an issue with t

chetanyakan:
Rebased to the HEAD of master.

model/config.go
chetanyakan: @reflog The doc mentions that each feature flag should be off by default.
But, we need to make sure the setting is enabled for all users.

Any advice there?

chetanyakan:
Rebased to HEAD of master

api4/status.go
chetanyakan: @reflog PR for API documentation here: https://github.com/mattermost/mattermost-api-reference/pull/609

(cc @esethna)

model/config.go
reflog: @chetanyakan Feature flags override the Config defaults. Currently, any new feature that is added to Mattermost is first 'hidden' behind a feature flag, rolled out gradually, tested, and then the feature flag is removed completely, leaving the defaults that you've set in the Config to take effect

app/slashcommands/command_custom_status.go
reflog: @chetanyakan I would probably fix the model.EMOJI_PATTERN itself, if it's actually an incorrect regex. I doubt that will break things.
@jespino - thoughts?

model/config.go
chetanyakan: @reflog I can set the feature flag to turn the custom status feature off.

Is there a way I can set the feature flag to enable it locally and on the test server?

reflog: @chetanyakan - you take https://github.com/mattermost/mattermost-server/pull/16598 as an example of how to add a feature flag

model/feature_flags.go
```diff
@@ -27,6 +31,7 @@ func (f *FeatureFlags) SetDefaults() {
 	f.TestBoolFeature = false
 	f.CloudDelinquentEmailJobsEnabled = false
 	f.CollapsedThreads = false
+	f.CustomUserStatuses = true
```
reflog, line 34: this should be false by default

reflog (changes requested):

app/slashcommands/command_custom_status.go
hmhealey: The pattern is technically correct since it was written to be used for custom emojis, and like you found, `:+1:` is a bit of a special case there. That said, it keeps causing us problems like this, so I think we allow plus signs in custom emojis to simplify things. I have wanted to make system emojis more similar to custom emojis for a while now. I think we'd need to consider what else needs to change for this though since at least the web app and some help text would need updates from what I see.

Maybe for now, we could add a new pattern for use here? It's not great, and I imagine naming it something like `ALL_EMOJI_PATTERN` might confuse some people in the future, but I think that's the safest thing to do for now.

model/feature_flags.go
esethna: @reflog this feature should be enabled by default for all users and servers

hmhealey: We might need to talk to Cloud (or whoever owns feature flags) about that then since the intent was to have these features off by default and then gradually rolled out to Cloud servers before the flag finally gets removed and they're available in on-prem. On-prem users can still bypass the feature flag by setting an environment variable, but it wouldn't be on for them by default until the flag is removed.

esethna: The gradual roll-out to cloud makes sense, I just want to make sure that when it hits on prem, it's enabled by default

esethna:
@jespino please review here when you can so we aren't blocked on merging

chetanyakan:
Rebased to the HEAD of master

model/feature_flags.go
reflog: @esethna - this is opposite to current practices as outlined by @crspeller .
Until feature is deemed 'complete' the feature flag by default should be 'false'.

app/status.go
```diff
@@ -363,3 +364,86 @@ func (a *App) GetStatus(userID string) (*model.Status, *model.AppError) {
...
+	}
+
+	user.SetCustomStatus(cs)
+	_, updateErr := a.UpdateUser(user, true)
+	if updateErr != nil {
+		return err
+	}
+
+	if err := a.addRecentCustomStatus(userID, cs); err != nil {
+		return err
```
jespino, line 381: I'm not 100% sure if we should fail entirely if you are not able to add the custom status to the recent custom statuses. Maybe we can log that. The custom status is already changed, so a "partial" update is better than a confusing response.

chetanyakan: @jespino Agree that it may be a bit confusing to return an error as the `API` or `Slash command` response even after the status is already changed in lieu of failed recent statuses update.

As per your suggestion, I have updated to only log the error in this case.

model/feature_flags.go
crspeller: You are correct, feature flags should be off by default. When ready to release to on-prem, they should be removed.

reflog:
@chetanyakan TeamSettings.EnableCustomUserStatuses should be added to telemetry, that's why CI is failing

chetanyakan:
> @chetanyakan TeamSettings.EnableCustomUserStatuses should be added to telemetry, that's why CI is failing

@reflog Can you point out any related ducumentation or example on how to add the setting telemetry?

reflog:
> 
> 
> > @chetanyakan TeamSettings.EnableCustomUserStatuses should be added to telemetry, that's why CI is failing
> 
> @reflog Can you point out any related ducumentation or example on how to add the setting to telemetry?

Take a look at services/telemetry/telemetry.go , all you need to do is to add an entry that reports the state of the config.

api4/status.go
```diff
@@ -107,3 +110,89 @@ func updateUserStatus(c *Context, w http.ResponseWriter, r *http.Request) {
...
 }
+
+func updateUserCustomStatus(c *Context, w http.ResponseWriter, r *http.Request) {
+	c.RequireUserId()
+	if c.Err != nil {
+		return
+	}
+
+	if !c.App.Config().FeatureFlags.CustomUserStatuses || !*c.App.Config().TeamSettings.EnableCustomUserStatuses {
+		c.Err = model.NewAppError("updateUserCustomStatus", "api.custom_status.disabled", nil, "", http.StatusForbidden)
```
jespino, line 121: Probably the status should be `http.StatusNotImplemented` here.

api4/status.go
```diff
@@ -107,3 +110,89 @@ func updateUserStatus(c *Context, w http.ResponseWriter, r *http.Request) {
...
+}
+
+func removeUserCustomStatus(c *Context, w http.ResponseWriter, r *http.Request) {
+	c.RequireUserId()
+	if c.Err != nil {
+		return
+	}
+
+	if !c.App.Config().FeatureFlags.CustomUserStatuses || !*c.App.Config().TeamSettings.EnableCustomUserStatuses {
+		c.Err = model.NewAppError("removeUserCustomStatus", "api.custom_status.disabled", nil, "", http.StatusForbidden)
```
jespino, line 153: same here

api4/status.go
```diff
@@ -107,3 +110,89 @@ func updateUserStatus(c *Context, w http.ResponseWriter, r *http.Request) {
...
+}
+
+func removeUserRecentCustomStatus(c *Context, w http.ResponseWriter, r *http.Request) {
+	c.RequireUserId()
+	if c.Err != nil {
+		return
+	}
+
+	if !c.App.Config().FeatureFlags.CustomUserStatuses || !*c.App.Config().TeamSettings.EnableCustomUserStatuses {
+		c.Err = model.NewAppError("removeUserRecentCustomStatus", "api.custom_status.disabled", nil, "", http.StatusForbidden)
```
jespino, line 177: Probably the status should be `http.StatusNotImplemented` here.

app/slashcommands/command_custom_status.go
```diff
@@ -0,0 +1,85 @@
...
+		AutoComplete:     true,
+		AutoCompleteDesc: T("api.command_custom_status.desc"),
+		AutoCompleteHint: T("api.command_custom_status.hint"),
+		DisplayName:      T("api.command_custom_status.name"),
+	}
+}
+
+func (*CustomStatusProvider) DoCommand(a *app.App, args *model.CommandArgs, message string) *model.CommandResponse {
+	if !a.Config().FeatureFlags.CustomUserStatuses || !*a.Config().TeamSettings.EnableCustomUserStatuses {
+		return &model.CommandResponse{}
```
jespino, line 46: Maybe a more explanatory error about why is not doing nothing? //cc @esethna

jespino (approved):
Some comments for improvement but nothing mayor. Looks good to me. Thanks @chetanyakan!

app/slashcommands/command_custom_status.go
chetanyakan: @esethna This is the case when the custom status setting is off and the `/status` slash command is executed.

Although we don't show any autocomplete text, the user can still execute the command. We don't take any action in this case. It probably makes sense to show a helpful error message instead.

What should be the slash command error message in this case?

P.S. probably not a blocker for this PR.

esethna: I think it should be the standard error message for an unrecognized slash command:
 
image

chetanyakan: @esethna This is done with the latest commit.

image

chetanyakan:
@reflog I've added telemetry settings as per your suggestions.

Please let me know if anything else is required to get this PR merged.

(cc: @esethna)

ogi-m (approved):
Approved by QA

chetanyakan:
Rebased to the HEAD of master.

reflog (approved):
Looks good!

chetanyakan:
Rebased to the HEAD of master.

i18n/en.json
```diff
@@ -703,6 +703,34 @@
     "id": "api.command_code.name",
     "translation": "code"
   },
+  {
```
jespino, line 706: this should be ordered alphabetically, if not it fails on CI.

esethna: @chetanyakan ^

chetanyakan: Done

jespino: Nope, you have executed the `make i18n-extract` and you removed all the enterprise translations, because you don't have the enterprise code.

chetanyakan: @jespino Thanks for pointing this out. 
I got confused as the behaviour of `make i18n-extract` is not the same as webapp. (which makes sense as there's no private enterprise code for the webapp)

But maybe, in a future PR, it might make sense to create a Makefile command for community's benefit to just reorder strings and not remove enterprise ones.

jespino: Actually, I think there is some changes proposed for the i18n-extract to preserve the enterprise strings.

amyblais:
To be cherry-picked to ``cloud`` branch.

jespino:
@amyblais Looks like the 5.33 branch doesn't exists, where should I cherry-pick this?

PR #16836, opened 2021-02-01
[MM-32389] Fix FeatureFlags section erroneously getting written to config

streamer45:
#### Summary

PR fixes an issue where the `FeatureFlags` section of the config was getting written back to the store even when `persistFeatureFlags = false` by adding to `loadLockedWithOld()` the same check present in `Set()` to avoid persisting the change.

#### Ticket

https://mattermost.atlassian.net/browse/MM-32389

#### Release Note

```release-note
Fixed FeatureFlags section erroneously getting written to config.
```

### DIFF
--- a/config/store.go
+++ b/config/store.go
@@ -214,6 +214,8 @@ func (s *Store) loadLockedWithOld(oldCfg *model.Config, unlockOnce *sync.Once) e
 		}
 	}
 
+	loadedFeatureFlags := loadedConfig.FeatureFlags
+
 	// If we have custom defaults set, the initial config is merged on
 	// top of them and we delete them not to be used again in the
 	// configuration reloads
@@ -257,7 +259,7 @@ func (s *Store) loadLockedWithOld(oldCfg *model.Config, unlockOnce *sync.Once) e
 		// MM cloud uses config in the DB as a cache of the feature flag
 		// settings in case the management system is down when a pod starts.
 		if !s.persistFeatureFlags {
-			s.configNoEnv.FeatureFlags = nil
+			s.configNoEnv.FeatureFlags = loadedFeatureFlags
 		}
 		toStoreBytes, err := json.Marshal(s.configNoEnv)
 		if err != nil {
--- a/config/store.go
+++ b/config/store.go
@@ -249,10 +249,17 @@ func (s *Store) loadLockedWithOld(oldCfg *model.Config, unlockOnce *sync.Once) e
 		return errors.Wrap(err, "failed to marshal loaded config")
 	}
 	if !s.readOnly && (len(configBytes) == 0 || !bytes.Equal(oldCfgBytes, newCfgBytes)) {
-		if err := s.backingStore.Set(s.configNoEnv); err != nil {
-			if !errors.Is(err, ErrReadOnlyConfiguration) {
-				return errors.Wrap(err, "failed to persist")
-			}
+		featureFlags := s.configNoEnv.FeatureFlags
+		// Don't persist feature flags unless we are on MM cloud
+		// MM cloud uses config in the DB as a cache of the feature flag
+		// settings in case the management system is down when a pod starts.
+		if !s.persistFeatureFlags {
+			s.configNoEnv.FeatureFlags = nil
+		}
+		err := s.backingStore.Set(s.configNoEnv)
+		s.configNoEnv.FeatureFlags = featureFlags
+		if err != nil && !errors.Is(err, ErrReadOnlyConfiguration) {
+			return errors.Wrap(err, "failed to persist")
 		}
 	}
 
--- a/config/store.go
+++ b/config/store.go
@@ -248,7 +248,10 @@ func (s *Store) loadLockedWithOld(oldCfg *model.Config, unlockOnce *sync.Once) e
 	if err != nil {
 		return errors.Wrap(err, "failed to marshal loaded config")
 	}
-	if !s.readOnly && (len(configBytes) == 0 || !bytes.Equal(oldCfgBytes, newCfgBytes)) {
+
+	var shouldStore bool
+	hasChanged := len(configBytes) == 0 || !bytes.Equal(oldCfgBytes, newCfgBytes)
+	if hasChanged {
 		featureFlags := s.configNoEnv.FeatureFlags
 		// Don't persist feature flags unless we are on MM cloud
 		// MM cloud uses config in the DB as a cache of the feature flag
@@ -256,10 +259,19 @@ func (s *Store) loadLockedWithOld(oldCfg *model.Config, unlockOnce *sync.Once) e
 		if !s.persistFeatureFlags {
 			s.configNoEnv.FeatureFlags = nil
 		}
-		err := s.backingStore.Set(s.configNoEnv)
-		s.configNoEnv.FeatureFlags = featureFlags
-		if err != nil && !errors.Is(err, ErrReadOnlyConfiguration) {
-			return errors.Wrap(err, "failed to persist")
+		toStoreBytes, err := json.Marshal(s.configNoEnv)
+		if err != nil {
+			return errors.Wrap(err, "failed to marshal old config")
+		}
+		shouldStore = !bytes.Equal(toStoreBytes, configBytes)
+		// We write back to the backing store only if
+		// the config has changed and the store is not read-only.
+		if !s.readOnly && shouldStore {
+			err := s.backingStore.Set(s.configNoEnv)
+			s.configNoEnv.FeatureFlags = featureFlags
+			if err != nil && !errors.Is(err, ErrReadOnlyConfiguration) {
+				return errors.Wrap(err, "failed to persist")
+			}
 		}
 	}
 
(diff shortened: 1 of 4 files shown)


crspeller (approved):
Good catch.

streamer45:
Adding the `Do Not Merge` label since PR needs some additional changes. We should also avoid invoking the listeners if no change has happened to alleviate the websocket buffer pressure in a clustered environment (limit the `config_changed` event spam).

agnivade (commented):
Dummy review to push it out of my queue.

config/store.go
```diff
@@ -248,9 +248,28 @@ func (s *Store) loadLockedWithOld(oldCfg *model.Config, unlockOnce *sync.Once) e
...
+	var shouldStore bool
+	hasChanged := len(configBytes) == 0 || !bytes.Equal(oldCfgBytes, newCfgBytes)
+	if hasChanged {
+		featureFlags := s.configNoEnv.FeatureFlags
+		// Don't persist feature flags unless we are on MM cloud
+		// MM cloud uses config in the DB as a cache of the feature flag
+		// settings in case the management system is down when a pod starts.
+		if !s.persistFeatureFlags {
+			s.configNoEnv.FeatureFlags = nil
+		}
```
agnivade, line 261: Should we refactor this to a separate method? Seems to be repeated in the `Get` method too.

config/store.go
```diff
@@ -248,9 +248,28 @@ func (s *Store) loadLockedWithOld(oldCfg *model.Config, unlockOnce *sync.Once) e
...
+		toStoreBytes, err := json.Marshal(s.configNoEnv)
+		if err != nil {
+			return errors.Wrap(err, "failed to marshal old config")
+		}
+		shouldStore = !bytes.Equal(toStoreBytes, configBytes)
```
agnivade, line 266: Wondering if it's possible to check if the `FeatureFlags` wasn't nil previously and was set to nil, and use that as boolean result for `shouldStore` rather than doing a `json.Marshal` and a `bytes.Equal`?

agnivade (commented):
Just a few clarifications. Looking great otherwise.

config/store.go
streamer45: That thought crossed my mind. The `shouldStore` isn't only about `FeatureFlags` though. Again the approach here was very conservative and I opted to repeat some logic rather than possibly breaking something (to be properly reviewed in [MM-32390](https://mattermost.atlassian.net/browse/MM-32390) of course).

Mostly I want to make sure that what we read from the backing store (`configBytes`) is different from what we are going to write (`s.configNoEnv`) to avoid triggering the set. The db store actually does this already internally but I am thinking a `json.Marshal` + `bytes.Equal` while not ideal should be more efficient (at least time-wise) than a db query or reading a file.

That said, I have a feeling a lot of this logic can be simplified (and possibly removed) but since it's just a feeling I am opting for the dead simple approach of actually checking the data (even if twice).

config/store.go
streamer45: You mean in `Set()`? I've purposely repeated but I guess we can make it method since it's all pointers.

crspeller (approved):

config/store.go
agnivade: Gotcha. Overall, the `Set` flow looks very complicated and hard to read for me. I thought the existing json.Marshal comparison was for that purpose. But it looks like it's not? That seems to compare `newCfgBytes, err := json.Marshal(loadedConfig)`, but it actually writes `s.configNoEnv`. IMO, there should only be 2 `json.Marshal`s, not 3. 

There seems to be 3 sets of configs that we are comparing:
- The old config struct
- The old config loaded from the backing store
- The new config that's going to be written.

And then there are multiple filters and environment maps being applied to each of them.

Totally agree that a lot of this logic is redundant and needs to be cleaned up. So will be looking forward to MM-32390 :)

config/store.go
agnivade: 0/5. Since the comment was repeated, it seemed like an important piece of logic that shouldn't be missed out if this needs to be done elsewhere again. But can be part of the cleanup.

agnivade (approved):
To be entirely honest, I still haven't understood the entire flow :P But tests are there where human beings fail :) Great work on fixing this!

streamer45:
@lindy65 

This is a bit delicate so two things would be needed:

- Make sure the reported issue doesn't reproduce anymore (we can involve someone from SDETs if you think it's appropriate).
- A smoke test around general functioning of the application, especially aimed at config changes.

config/store.go
streamer45: Yes, I didn't really trust all those transformations hence the added serialization + check on `configNoEnv` but I am confident that when I get to refactor this we should be able to avoid several checks and limit the marshaling to a minimum.

lindy65:
Thanks @streamer45 - I'll test the functionality of the app, including config changes in the system console and ask an SDET to check that the reported issue doesn't reproduce.

lindy65 (approved):
Hi @streamer45 - I've done a good smoke test of the app and system console and did not find any issues.  Saturn will be testing the reported issue. I've added him as an additional QA reviewer on this PR.

streamer45:
> Hi @streamer45 - I've done a good smoke test of the app and system console and did not find any issues. Saturn will be testing the reported issue. I've added him as an additional QA reviewer on this PR.

Thank you @lindy65 :+1:

saturninoabril:
Thanks @streamer45!

Tested the following and passed:

__A. On both TE and EE:__
1. On server start/restart,
      - (passed) `FeatureFlags` is being removed in config if `FeatureFlags` section was initially present before start/restart.
      - (passed) It remains without `FeatureFlags` if config is originally without such section.
2. On config update,
      - (passed) `FeatureFlags` is now not getting written back into the config. Specific config change is updated correctly.

__B. On Cloud:__
1. On server start/restart,
      - (passed) `FeatureFlags` remain as is when keys/values are the same as the default.
2. On config update,
      - (passed) `FeatureFlags` remain as is. Specific config change is updated correctly.

However, please help check the following if they are expected to Cloud server.
1. if I set `FeatureFlags.TestBoolFeature` to `true` and start/restart the server, the value in config automatically changed to `false`.
2. If I add new key/value pair to `FeatureFlags` like `”AnyFeature”: true`, it’s being removed on server start/restart and all I'm seeing are the default key/value pairs.

Other observations as it confused me on what to expect in both TE and EE:
1. `FeatureFlags` are not in config.json after server start which is expected. However, default flags and its default values are listed and shown in UI at `http://localhost:8065/admin_console/experimental/feature_flags`. Is that expected?
2. Or since `FeatureFlags` are not in config.json, should the section in System Console > EXPERIMENTAL > Feature Flags with link `http://localhost:8065/admin_console/experimental/feature_flags` be available/accessible at all? That seemed to be confusing to me if I am the system administrator.

Note:
Below are the default key/value pairs I'm seeing currently.
```json
    "FeatureFlags": {
        "TestFeature": "off",
        "TestBoolFeature": false,
        "CloudDelinquentEmailJobsEnabled": false,
        "CollapsedThreads": false,
        "PluginIncidentManagement": "1.1.1"
    },
```

streamer45:
Thank you @saturninoabril , terrific work on this.

> 1. if I set `FeatureFlags.TestBoolFeature` to `true` and start/restart the server, the value in config automatically changed to `false`.
> 2. If I add new key/value pair to `FeatureFlags` like `”AnyFeature”: true`, it’s being removed on server start/restart and all I'm seeing are the default key/value pairs.

Did this behaviour change from `master`?

> 1. `FeatureFlags` are not in config.json after server start which is expected. However, default flags and its default values are listed and shown in UI at `http://localhost:8065/admin_console/experimental/feature_flags`. Is that expected?

I'd say yes. The way the logic is now, the in-memory config can still have the `FeatureFlags` section (although set to defaults). We are just avoiding writing it to the backing store unless told so (if the server is running in cloud).

> 2. Or since `FeatureFlags` are not in config.json, should the section in System Console > EXPERIMENTAL > Feature Flags with link `http://localhost:8065/admin_console/experimental/feature_flags` be available/accessible at all? That seemed to be confusing to me if I am the system administrator.

Probably for non-cloud installations it should not be available. I admit I don't have much context on feature flags but it seems like they are mainly (only?) built for cloud.

saturninoabril:
Thanks @streamer45 for looking.

> Did this behaviour change from master?

No, same on master. With that, I created a ticket to discuss in triage and maybe address separately - [MM-32574](https://mattermost.atlassian.net/browse/MM-32574)

Also filed [MM-32575](https://mattermost.atlassian.net/browse/MM-32575) to triage for UI/UX.

I'm good (tested and passed) with the changes here per ticket description. Let me know how to proceed.

streamer45:
> > Did this behaviour change from master?
> 
> No, same on master. With that, I created a ticket to discuss in triage and maybe address separately - [MM-32574](https://mattermost.atlassian.net/browse/MM-32574)

That's very interesting. It seems to me like it would defeat the purpose of caching them in the config if they get reset on startup.
I've pushed a small change. I'd like to re-test with this on cloud and verify the reset doesn't happen.

saturninoabril (approved):
Tested and passed. Also tested the issue filed as MM-32574 and all worked as expected using split.io.  Other ticket, MM-32575, is closed as expected during Triage meeting.

crspeller:
Good QA @saturninoabril! Thanks for taking a look at everything!

streamer45:
> Good QA @saturninoabril! Thanks for taking a look at everything!

Yeah, brilliant work :clap:

PR #16847, opened 2021-02-02
MM-32553 hard delete channel when channel invite fails

wiggin77:
#### Summary
When accepting a channel invite a new channel is created. If the subsequent creation of `SharedChannelRemote` fails then the new channel is deleted.

Previously the channel was soft deleted.   This PR changes that to permanent delete the channel otherwise a retry will fail due to id collision.

**To be merged with feature branch**

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-32553

#### Release Note

```release-note
NONE
```


streamer45 (approved):
Nice :+1:

PR #16851, opened 2021-02-03
Add feature flag for apps

larkox:
#### Summary
Add a feature flag for the Apps framework. This feature flag is used in two main places:
- All over the user agents (Webapp and Mobile) as a kill switch for all the App functionality.
- At the plugin activation/deactivation process, to tie the proxy to the feature flag instead of the active status.

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-32467

#### Release Note
```release-note
NONE
```


model/feature_flags.go
```diff
@@ -20,13 +20,17 @@ type FeatureFlags struct {
...
+	// AppsEnabled toggle the Apps framework functionalities both in server and client side
+	AppsEnabled bool
 }
 
 func (f *FeatureFlags) SetDefaults() {
 	f.TestFeature = "off"
 	f.TestBoolFeature = false
 	f.CloudDelinquentEmailJobsEnabled = false
 	f.CollapsedThreads = false
+	f.AppsEnabled = true
```
catalintomai, line 33: Typically a feature starts in the "disable" (default) state and after one gets enough confidence wrt to its functioning then it gets permanently set to the "enable" state or the feature flag is removed - though maybe it doesn't make much of a difference now since the number of cloud deployments is relatively small. This is different from a killswitch for which the new behavior is by default "on" and when the killswitch is turned "on" the old behavior is "on" . Do we know what is the mechanism for turning the feature flag "on" on a subset of the cloud instances?

model/feature_flags.go
```diff
@@ -20,13 +20,17 @@ type FeatureFlags struct {
 	CollapsedThreads bool
 	// Feature flags to control plugin versions
 	PluginIncidentManagement string `plugin_id:"com.mattermost.plugin-incident-management"`
+
+	// AppsEnabled toggle the Apps framework functionalities both in server and client side
+	AppsEnabled bool
```
crspeller, line 25: The intended use of feature flags for plugins is to use the functionally like Incident is doing above. Where it will install a specified version from the marketplace and automatically enable it. This way you can roll out new version of the plugin, not just enable and disable it. 
It also gives the user control to enable/disable the plugin and prevent automatic upgrade from taking place.

crspeller (changes requested):

model/feature_flags.go
larkox: @catalintomai @levb Is this what we want?
@crspeller We might not want to allow the user to disable / enable this plugin. It is an open discussion right now.

larkox: @crspeller after internal discussion, we have thought the following:
- At the moment, we are not interested in using the features flags the way Incident Management uses them. We are not exactly forcing a version.
- We don't want users to be able to enable or disable the plugin. We will also add the needed steps so the plugin becomes a "special" plugin like NPS. The reason for this is that even if the apps framework is implemented as a plugin for isolation, we prefer it being considered as part of the product. Therefore it can only be turned off by feature flags.

Does this make sense?

crspeller: @larkox Sounds like you just want a configuration setting then. Or maybe just a special environment variable?  Feature flags are for rolling out new features or plugin versions, then are designed to be removed. This sounds more perminant?
When do you see someone using this flag?

larkox: We plan to have this feature flag until we are confident with the feature. When we are confident with the results of the feature, we can remove the feature flag. We may add more tooling to keep this "always enabled" functionality in the future, but right now this is the plan, at least the way I see it.
@catalintomai @levb Any comment?

crspeller: Presumably you plan to do this more than once? What are you going to do when you have 1.0 of your plugin and you want to roll out 1.1?

catalintomai: We can do smth. similar - add a different feature flag for the new functionality and remove it when we are confident that the functionality is ok. Is it mandatory to reuse an existing feature flag for a particular area of the product (this plugin in this case)?

crspeller: > Is it mandatory to reuse an initial/existing feature flag for a particular area of the product (this plugin in this case)?

No

> the way we have things setup now - it triggers a version change

Ya. It's less granular then the feature flags can be on the server. But since plugins generally have less per release and are managed by versions, it makes sense to flag by version. 

> and since we don't have them yet, then we'll have to use a feature flag

Feature flags are kill switches. 

> or should we use a new one for that purpose (and not tie it to the version one)

You can. if you want more granular control, in addition to version control, you can use additional feature flags to control specific features on the plugin. The feature flags come in though the config to plugins.

catalintomai (approved):

model/feature_flags.go
larkox: I made another Push. As a summary:
- We want apps to behave as part of the system. Therefore, we want it to be always active.
- We want the version enforcing and the kill switch.
- Version enforcing will allow us to make sure all instances are at the version we want.
- Kill switch will allow us to turn off the apps if necessary.

Does this make sense @crspeller ? @catalintomai , am I mising anything?

crspeller (approved):
Looks good! Seems like the kill switch would get deprecated as you release more versions. (not ideal if customers are actually using it to kill all their apps)

larkox:
Exact. The kill switch is a "safety measure" until we are comfortable enough with the apps code. It is supposed to be "short-lived", and eventually it might be "replaced" by other feature flags that will turn on/off other minor new features in the code, and when the project is mature enough, all feature flags can be removed (but the version enforcing one, probably).

Thanks for the review and all the comments!

app/plugin.go
```diff
@@ -97,6 +97,11 @@ func (a *App) SyncPluginsActiveState() {
 				pluginEnabled = state.Enable
 			}
 
+			// Tie Apps proxy features to the feature flag.
+			if pluginId == "com.mattermost.apps" {
+				pluginEnabled = a.Config().FeatureFlags.AppsEnabled
+			}
```
hanzei, line 103: Does that mean an admin on Cloud can't disable the plugin? If so, is that obvious to an admin?

hanzei (changes requested):
One question on the UX

app/plugin.go
larkox: We need to add some handling to this on the frontend. My approach would be to hide the plugin disable/enable button from the system console, or even the whole plugin.

hanzei: Then the questions becomes if admins should be able to remove the Apps plugin. @aaronrothschild What do you think?

DHaussermann:
@larkox I was able to test this on split.io with some help from @saturninoabril 👍 

1. Functionally, the flag does work:
- `On` enables the plugin and `Off` disables the pluign.
- `On` prevents the plugin from disabling and `Off` prevents the plugin from disabling. 
But the UX for this is odd in that the `enable`/`disable` remains visible and the UI just hangs when you you click it and on refresh you can see that it failed.
Screen Shot 2021-03-05 at 11 37 00 PM
Screen Shot 2021-03-05 at 11 38 41 PM

0/5 not sure how the IR plugin does it but, could we just hide the button?

2. Any thoughts on how the version can be tested. 
A. Do you know what the mechanism is that does the upgrade when the version on split.io is updated?
B. Can we build a token release wit a version incremented? (maybe this only works off of official releases from GitHub?)

larkox:
@DHaussermann Regarding UX, I created this PR: https://github.com/mattermost/mattermost-webapp/pull/7651
Regarding 2, I feel we can trust this feature to work as expected, and move forward with merging to master. (For this part we are just adding the feature flag, not adding any extra functionality).

DHaussermann (approved):
Tested and passed.
- The flag is working as per comment above.
- As we have only one version, I agree that it seems reasonable to assume the split io version deployment functionality would work as it is not unique to Cloud Apps and is already in use by IR. 
- Confirmed follow-up Webapp PR will resolve the concerns for 1. 
- @larkox  I will test this again if there are more changes but for now, it is not possible to remove the plugin. Seems there is an unresolved conversation in the reviews about this. Do we still need input from @aaronrothschild?
- Created test case to cover this change here https://mattermost.atlassian.net/projects/MM?selectedItem=com.atlassian.plugins.atlassian-connect-plugin:com.kanoah.test-manager__main-project-page#!/testCase/MM-T3870

app/plugin.go
aaronrothschild: > Then the questions becomes if admins should be able to remove the Apps plugin. @aaronrothschild What do you think?

I have no objection on Cloud to this handling (ability to turn off the Plugin), but why does it need to be like that on Self-Managed?    I think Admins may expect to be able to turn off all their plugins if they want to.  We'd need to explain why the apps plugin, MUST remain running.

hanzei: @larkox Does it make sense to limit https://github.com/mattermost/mattermost-webapp/pull/7651 to cloud?

larkox: I see value in that when we get out of the feature flag stage. Right now, the only way to keep the plugin running or stopped is by setting or unsetting the feature flag, doesn't matter if it is cloud or on-prem.

hanzei (approved):
It's a bit of a PITA for devs, but works fine for now

aaronrothschild:
@larkox is it possible for us to add a strong warning message that disabling the plugin will result in all apps no longer working, and leave the ability for an admin to turn the Proxy Plugin Enabled/Disabled, but not remove it.

aaronrothschild (approved):

larkox:
@crspeller reasking for review since we slightly change the behaviour.

Now if the feature flag is set to false, the app will get disabled.
If the feature flag is on, will remember the user selection (whether it was enabled or disabled). If the user has not changed the status, it will stay as enabled.

larkox:
A minor update on my previous comment.

When the feature flag is set to false, we ignore the value in the config and keep the plugin disabled.
When the feature flag is on, we use the config value (which is maintained even if we change the feature flag).
When the instance starts for the first time, it will use the version flag to install the plugin and enable for the first time (since there is no information of enable or disable on the config).

crspeller (approved):
Looks good. It should work the way you describe.

app/plugin.go
```diff
@@ -97,6 +97,13 @@ func (a *App) SyncPluginsActiveState() {
 				pluginEnabled = state.Enable
 			}
 
+			// Tie Apps proxy disabled status to the feature flag.
+			if pluginID == "com.mattermost.apps" {
```
catalintomai, line 101: glad we didn't have to store another state = )

catalintomai (approved):

DHaussermann (approved):
This is working as expected. I've been using this branch locally for some time.
I can also see the new webapp changes working as expected.

LGTM!

PR #16854, opened 2021-02-03
Save access on plugin env during ServePluginPublicRequest

nronas:
#### Summary
Fixes unsafe access of properties of plugin environment during `ServePluginPublicRequest`

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-32475

#### Release Note
```release-note
* Fixes unsafe access of properties of plugin environment during `ServePluginPublicRequest`
```


app/plugin_requests.go
```diff
@@ -89,6 +89,11 @@ func (a *App) ServePluginPublicRequest(w http.ResponseWriter, r *http.Request) {
 	vars := mux.Vars(r)
 	pluginID := vars["plugin_id"]
 
+	if a.GetPluginsEnvironment() == nil {
+		http.NotFound(w, r)
+		return
+	}
```
streamer45, line 95: I guess my question would be if we have any idea on why the plugin environment would be nil here. Is it a race condition of some sort? (e.g. not yet initialized).

nronas: I should have written that down on the description, my bad. The `GetPluginsEnvironment` check the config to see if `PluginSettings` are enabled. If not it will return nil. So we have a "race" between an active request that hits the server for the plugin api and someone disabling the plugin settings. https://github.com/mattermost/mattermost-server/blob/master/app/plugin.go#L45-L47

I think that Is not worth synchronizing those conditions because they have different critical sections.

streamer45: Awesome, thank you for the added context :+1: . Would it make sense to leave a small comment in the code? I am sure it would help future readers.

streamer45 (approved):

agnivade (commented):
Let's also add a test please :)

agnivade (changes requested):
Hmm .. actually on a deeper look, I don't believe it's due to `PluginSettings` being disabled. From the stack trace, I see that it's coming from `(*Environment) GetPluginState`, which means that it's the `*Environment` struct which is nil.

So there can still be a race here if `a.GetPluginsEnvironment` does NOT return nil, and then it gets set to nil, and then it proceeds to execute `a.GetPluginsEnvironment().PublicFilesPath(pluginID)`.

We probably need to use the `s.PluginsLock` somewhere.

nronas:
@agnivade yeap good point.  Will see what I can do on synchronizing access on the *env pointer.

nronas:
@agnivade  this is the cleanest way that I could come up with. https://github.com/mattermost/mattermost-server/pull/16854/commits/961c8a827cfa84dc7b7880fedabb7c9cb2f20125 the main issue is that the lock can be accessed only from the `Server` context so a double check will suffice here. Also the only place that sets the env to nil is server shutdown https://github.com/mattermost/mattermost-server/blob/master/app/plugin.go#L314

Let me know what you think

agnivade:
I am still trying to understand how this can happen in the first place. Can we write a test for this? If the only place where it can be nil is during server shutdown, how did this crash happen?

IIRC, @streamer45 fixed a similar bug a long time back. But this seems to be from a recent build from the Go version.

nronas:
> I am still trying to understand how this can happen in the first place. Can we write a test for this? If the only place where it can be nil is during server shutdown, how did this crash happen?

I can defo try to cover it with a test case, it's a bit tricky though since my theory judging from the error trace is that we had an active request on the plugins API and at that time the server got a SIGTERM. The shutdown method was called and because the API is not concurrently safe it reached for the plugin env, it got it back and then the server shutdown set it to nil which raised the nil pointer panic. I think the scenario is super edge-casey.

streamer45:
Interestingly this is easily reproducible. You just need to go do an HTTP GET to `/plugins/some.fake.plugin/public/somefile` after server start but **before** ever loading the root. Loading the root (e.g. `http://localhost:8065/`) does trigger all the initialization (plugins included). This seems to be explained by the fact that the specific plugin route doesn't go through our `ServeHTTP` handler which is what eventually triggers the initialization (through `App.InitServer`).

nronas:
> Interestingly this is easily reproducible. You just need to go do an HTTP GET to /plugins/some.fake.plugin/public/somefile after server start but before ever loading the root. Loading the root (e.g. http://localhost:8065/) does trigger all the initialization (plugins included). This seems to be explained by the fact that the specific plugin route doesn't go through our ServeHTTP handler which is what eventually triggers the initialization (through App.InitServer).

Ohhhhhhh that i didn't know, TIL. This is indeed super interesting

agnivade:
Oh somehow I missed this too. `InitServer` is only called from websocket router. Somehow I thought that any request via web/handlers.go would call it. Nice one :+1:

nronas:
Right, so I think there are two issues here. 1) what @streamer45 highlighted and the other one is the race condition during a shutdown. So the initial "fix" without the lock, by just checking for the presence of the environment it should work. If we use the lock we are also covering the race on the shutdown as well.

The request_plugins are being initialized in `NewServer` with a comment that says:

`// FakeApp: remove this when we have the ServePluginRequest and ServePluginPublicRequest migrated in the server`

agnivade:
I think I would be good with just the "fix" for this exact bug along with a test case that shows it.

app/plugin_requests.go
```diff
@@ -89,11 +89,24 @@ func (a *App) ServePluginPublicRequest(w http.ResponseWriter, r *http.Request) {
...
+
+	publicFilesPath, err := pluginsEnv.PublicFilesPath(pluginID)
 	if err != nil {
 		http.NotFound(w, r)
+		a.srv.PluginsLock.Unlock()
 		return
 	}
+	a.srv.PluginsLock.Unlock()
```
streamer45, line 109: Do we really need to lock? Once we've got the environment and verified it's not `nil` I am thinking we should be good.
Also if we go with the lock I believe it should be enough to read-lock.

nronas: For the issues of an active server instance, we do not need the lock for sure. But there is a race condition during a shutdown. Good point regards the Rlock.

streamer45: What is the resource affected by the race? Getting the plugin environment through `GetPluginsEnvironment` is safe. If that returns a non-nil value then that's what we are eventually calling methods on. The server can still set to `nil` its own value but I don't see a big problem with that.

nronas: Oh shoot, you are 💯 right. I thought the server value was the same pointer. Awesome, amending.

app/plugin_requests_test.go
```diff
@@ -0,0 +1,50 @@
...
+		app.UpdateConfig(func(cfg *model.Config) { *cfg.PluginSettings.Enable = true })
+
+		req, err := http.NewRequest("GET", "/plugins", nil)
+		require.NoError(t, err)
+
+		rr := httptest.NewRecorder()
+		handler := http.HandlerFunc(app.ServePluginPublicRequest)
+		handler.ServeHTTP(rr, req)
+
+		assert.Equal(t, rr.Code, http.StatusNotFound)
```
streamer45, line 48: Very small detail but at times this can cause headaches when debugging test failures: the expected value (`http.StatusNotFound` in this case) should go first.

streamer45 (approved):
Looks good :+1:

app/plugin_requests_test.go
nronas: Amended, quick fingers on my end.

app/plugin_requests_test.go
```diff
@@ -0,0 +1,50 @@
...
+	"testing"
+
+	"github.com/mattermost/mattermost-server/v5/config"
+
+	"github.com/gorilla/mux"
+
+	"github.com/mattermost/mattermost-server/v5/model"
+
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
```
agnivade, line 19: Nit: imports messed up

app/plugin_requests.go
```diff
@@ -89,7 +89,15 @@ func (a *App) ServePluginPublicRequest(w http.ResponseWriter, r *http.Request) {
 	vars := mux.Vars(r)
 	pluginID := vars["plugin_id"]
 
-	publicFilesPath, err := a.GetPluginsEnvironment().PublicFilesPath(pluginID)
+	pluginsEnv := a.GetPluginsEnvironment()
+
+	// Check if someone has nullify the pluginsEnv in the meantime
```
agnivade, line 94: Nit: nullified

agnivade (approved):
Sorry for the delay.

nronas:
Test cases were written to test the bug resolution. No QA needed. QA deferred

PR #16855, opened 2021-02-03
[MM-32437] Make error text for message props more explanatory

larkox:
#### Summary
Error text for message props was only stating that there was an error. Now it says the error is due to the length of the message.

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-32437
#### Release Note
```release-note
NONE
```


mgdelacroix (approved):
LGTM! :+1:

justinegeffen (approved):

wiersgallak:
@larkox @DHaussermann Can you guys give an update on this PR?

DHaussermann:
Thanks for the ping @wiersgallak and putting this back on my radar. I will make room in my queue to test this PR over the next few days.

wiersgallak:
@DHaussermann repinging this one.  Thanks!

DHaussermann:
@larkox I wasn't able to verify the existing behavior on this. The jira ticket says " Send a message to an incoming webhook with an attachments field more than 8,000 characters long"
But, what part of `  "attachments": []` is affected
I put 16K+ characters in the attachment text but can't repor the original issue.
image
 
 Can you please provide some test steps or guidance on this?

larkox:
@DHaussermann I am not sure from your screenshot if the props are properly formed. The complete post object should look something like:
```
{
  channel_id: 'fasffewrfsadfad',
  message: 'some message',
  props: {
    attachments: {
      text: 'some text',
    }
  }
}
```
Does your payload generate a post in the system?

DHaussermann:
After testing the behavior on 5.35 and testing with this code change, I see a different behavior from the orginal bug report. 
When the attachment text exceed 8000 characters, it seems to silently truncate the text string and make the post via the webhook with the end of the text missing.

@larkox is investigating why the post does not fail as expected.

larkox:
After looking into this, there are two different things:
- The plugin API (or the standard API) CreatePost function will show this error. This is what has been fixed.
- The Webhook API will try to truncate the text of the props to fit the size limitations. The error shown by this when no truncation is possible (when you have more than 8000 characters anywhere that is not the text field) is correct.

CC. @DHaussermann

DHaussermann:
Thanks @larkox so, just to clarify here the code change is not related to the webhooks functionality but rather the api for making posts directly?

@icelander can you confirm that this will resolve the client issue? I may have misunderstood something but it seems this solution may not address the problem described in the related Jira ticket.

larkox:
I confirm that the webhook functionality works in a different way and it is handling the error in a different way in a different place of the code.

larkox:
@DHaussermann friendly reminder to test this.

DHaussermann:
@larkox I'm working to clean-out stale PRs and came across this. 
This issue was tested but never reproduced as outlined by support. When I attempted to test this by exceeding the post maximum - the payload was simply divided across multiple posts and so no error was returned. 

Is this still needed and can you provide some test guidance on this?

larkox:
As far as I know, this is for the plugin API. May also be for the rest API, but when trying to check that on community, I got a timeout :sweat_smile:

The main problem for testing, is that they bumped the size up to 800.000 runes, and it seems it takes a while to process such a big message in the server.

For webhooks, for sure it works in a different way.

All that being said... 800k is a limit hard to reach, so we may be safe closing this PR, but if anyone ever reach that limit, we may have to bring this back.

isacikgoz:
Hi @larkox, Note that we [merged multiple repositories into one monorepo](https://github.com/mattermost/mattermost-server/issues/22420) which may require you to resubmit these changes. I'm closing in the favor of reducing load on mattermod. Also the CI checks should run on our new pipelines. You can update the branch and re-open if you feel necessary.

PR #16865, opened 2021-02-04
unparam lint

mahmud2011:
#### Summary
* fix "... is always nil" lint err

#### Ref

#### Ticket Link
https://github.com/mattermost/mattermost-server/issues/16622

#### Release Note
```release-note
None
```


agnivade:
Does this error also come from `unparam`?

mahmud2011:
@agnivade Yes. For instance:

```
app/plugin_api_tests/test_update_user_active_plugin/main.go:24:75: 
(*MyPlugin).MessageWillBePosted - result 0 (*github.com/mattermost/mattermost-server/v5/model.Post) 
is always nil (unparam)
```

plugin/interface_generator/main.go
```diff
@@ -265,6 +265,7 @@ func getPluginInfo(dir string) (*PluginInterfaceInfo, error) {
 	packages, err := parser.ParseDir(pluginInfo.FileSet, dir, nil, parser.ParseComments)
 	if err != nil {
 		log.Println("Parser error in dir "+dir+": ", err)
+		return nil, err
```
agnivade, line 268: @crspeller - Any thoughts on this?

app/plugin_api_tests/manual.test_load_configuration_defaults_plugin/main.go
```diff
@@ -40,7 +40,8 @@ func (p *MyPlugin) MessageWillBePosted(_ *plugin.Context, _ *model.Post) (*model
 	if !p.configuration.MyBoolSetting {
 		return nil, "MyBoolSetting has invalid value"
 	}
-	return nil, "OK"
+
+	return &model.Post{}, "OK"
```
agnivade, line 44: I'm not sure I understand why are we doing this. What is the error you are seeing?

plugin/interface_generator/main.go
crspeller: Looks like the return was just forgotten?

app/plugin_api_tests/manual.test_load_configuration_defaults_plugin/main.go
mahmud2011: `is always nil (unparam)`. 

The method is returning `nil` in every if condition and also in final return. So, the linter is giving `is always nil error`.

plugin/interface_generator/main.go
agnivade: Yeah, the docs say that on an error, a partial map is returned with whatever was parsed. So just checked whether it was intentional or not. Thanks for the clarification!

app/plugin_api_tests/manual.test_load_configuration_defaults_plugin/main.go
agnivade: I see the issue now. Since we are not using the post param at all, we don't have an option. I think returning `nil` makes more sense than returning an empty post. Let's add an ignore rule in .golangci.yml for this.

Something like for all file paths under `plugin_api_tests/`, ignore the `is always nil` error.

agnivade (approved):
Awesome work!

wiggin77 (approved):
LGTM :+1:

agnivade:
@mahmud2011 - Please run `make i18n-extract` to remove a translation string from en.json file.

agnivade:
Woops sorry, the tool also removes the ones if it cannot access the EE repo. We only need to remove the one that's removed due to this PR. All the "ent.*" ones should remain.

@rvillablanca - Wasn't this problem fixed IIRC? Was it to pass a flag or something to prevent the EE translations from being deleted?

rvillablanca:
@mahmud2011 @agnivade yup, it is a flag called `--contributor` which is intended for that purpose.

image

So I believe this is not enabled in the Makefile so you have to run it manually with this command:

```shell
mmgotool i18n extract --contributor
```

Take into account `--portal-dir` flag is required.

agnivade:
Ah right, I had an older version of the tool.

agnivade:
These are just internal changes and don't require any QA.

PR #16875, opened 2021-02-05
pre-package incident management v1.3.2

lieut-data:
#### Summary
v1.3.1 proved to be incompatible with custom permission schemes. Bump the version to v1.3.2, reverting the underlying feature that caused a problem.

#### Ticket Link
Fixes: https://mattermost.atlassian.net/browse/MM-32569

#### Release Note
```release-note
Pre-package Incident Collaboration v1.3.2
```


crspeller (changes requested):
Need to change the default in `model/feature_flags.go`. Otherwise auto install won't install the newer version.

lieut-data:
Oh, interesting -- thanks @crspeller. I'd forgottten about the nuances here. Would it make sense to have the default be the empty string, and let split fully control the deployment, if any?

I'm hoping to have a future change that obviates the need to bump the pre-packaged versions at all (do it automatically), but this default will always need to be bumped.

crspeller:
@lieut-data The default controls what on-prem instances get. Those that are not hooked up to split or in cloud. If it was set to "" nothing would be automatically installed.

lieut-data:
Ah, I see. I'd love to see us consolidate this process somehow, since NPS self-installs itself the way incident collaboration used to: by defaulting to "enabled" in the config, and pulling from whatever version is in the pre-packaged cache.

Would it make sense to disable all of this for self-managed customers and revert to the old model? Or can we have a special "version" value that says, "just install what's locally available"?

lieut-data:
This PR is blocked until we can identify why the `/mb cutplugin` process didn't push the necessary artifacts to plugins-store.test.mattermost.com (the reason the CI is failing).

prapti (approved):

Makefile
```diff
@@ -100,7 +100,7 @@ PLUGIN_PACKAGES += mattermost-plugin-channel-export-v0.2.2
 PLUGIN_PACKAGES += mattermost-plugin-custom-attributes-v1.3.0
 PLUGIN_PACKAGES += mattermost-plugin-github-v2.0.0
 PLUGIN_PACKAGES += mattermost-plugin-gitlab-v1.3.0
-PLUGIN_PACKAGES += mattermost-plugin-incident-management-v1.3.1
+PLUGIN_PACKAGES += mattermost-plugin-incident-collaboration-v1.3.2
```
crspeller, line 103: We didn't change the plugin ID. This should just stay the same right?

crspeller (changes requested):

crspeller (approved):
Oh wait, nevermind, that's the fix.

lieut-data:
@crspeller, I've pushed a change that disables the remote marketplace in the prepackaged plugin tests: with the change above, it started downloading and installing v1.3.2 in the tests. Can you take a look before we finish the merge?

lieut-data:
For context, the reason the original tests were passing was because the tests couldn't find `v1.1.1` from the marketplace. With a bit of extra logging:
```
   testing.go:22: {"level":"debug","ts":1612814689.458827,"caller":"marketplace/client.go:74","msg":"GetPlugin","filter":{"Page":0,"PerPage":0,"Filter":"","ServerVersion":"5.32.0","BuildEnterpriseReady":false,"EnterprisePlugins":false,"Cloud":false,"LocalOnly":false,"Pla
tform":"darwin-amd64","PluginId":"com.mattermost.plugin-incident-management","ReturnAllVersions":true},"version":"1.1.1","count":2}
    testing.go:22: {"level":"debug","ts":1612814689.458946,"caller":"marketplace/client.go:76","msg":"GetPlugin - found","version":"1.3.2"}
    testing.go:22: {"level":"debug","ts":1612814689.459051,"caller":"marketplace/client.go:76","msg":"GetPlugin - found","version":"1.3.1"}
    testing.go:22: {"level":"debug","ts":1612814689.45914,"caller":"mlog/log.go:211","msg":"Unable to install plugin from FF manifest","plugin_id":"com.mattermost.plugin-incident-management","error":"GetMarketplacePlugin: Could not find the requested marketplace plugin., plugin not found","version":"1.1.1"}
```

I can reproduce this manually:
```
http https://api.integrations.mattermost.com/api/v1/plugins plugin_id==com.mattermost.plugin-incident-management return_all_versions==true cloud==false local_only==false platform==darwin-amd64 server_version==5.32.0 build_enterprise_ready==false | jq '.[].manifest.version'
"1.3.2"
"1.3.1"
```

It looks like, without the enterprise source, we never find v1.1.1 -- but v1.3.* are incorrectly marked as not requiring enterprise, and so get installed, thus failing the tests. As discussed with @crspeller, the path to the tests still appears to be correct (to be robust against such things), but I'll also correct the metadata in the production marketplace to consider these plugins as enterprise-only.

crspeller (approved):

Issue #16623, opened 2021-01-04
Fix all initialism errors in the codebase


agnivade:
#### Summary
This ticket is about fixing any variable naming issues that do not follow the initialism convention as laid out in: https://github.com/golang/go/wiki/CodeReviewComments#initialisms.

The way to go about it would be to make the following change to .golangci.yml

```diff
diff --git a/.golangci.yml b/.golangci.yml
index 29d9a4622..662931923 100644
--- a/.golangci.yml
+++ b/.golangci.yml
@@ -49,7 +49,7 @@ issues:
 
     - linters:
       - golint
-      text: "should have|should be|should replace|stutters|underscore|annoying|error strings should not be capitalized"
+      text: "should replace|stutters|underscore|annoying|error strings should not be capitalized"
 
     - linters:
       - golint      
```

And then fix all issues that arise until make golangci-lint passes with no errors.

SimonSimonB:
I'd like to work on this.

agnivade:
All yours @SimonSimonB !

haardikdharma10:
Hi @agnivade, If work has not been started on this ticket yet, I would like to give it a go!

agnivade:
Reopening this is not yet fully fixed.

haardikdharma10:
Hi @agnivade, I have managed to fix few of the errors. Can I open a draft PR?

agnivade:
You are most welcome to do that. But ensure to coordinate with @SimonSimonB so that work is not duplicated. :)

SimonSimonB:
Thanks @haardikdharma10 ! I'll make another commit related to this later today or tomorrow---just so you know and I'm not interfering with any changes you might make.

SimonSimonB:
I'm afraid I won't have time for another PR for this issue in the next weeks, so I unassigned myself in case someone else would like to take over.

haardikdharma10:
Thanks for letting us know @SimonSimonB. @agnivade, you can assign this to me, I'll open a bunch of PR's for this later.

ethervoid:
We need to revert [PR-3](https://github.com/mattermost/mattermost-server/pull/17062) because it changes the plugins API and leads to those plugins failing.

TealWater:
Hi team... to confirm this is about fixing errors like this: 
https://github.com/mattermost/mattermost/blob/4a4bf8244081c0abab7f92e93b8bbff94c6b626d/server/channels/api4/apitestlib.go#L726

[...]

PR #16877, opened 2021-02-05
[Partial Fix - #16623]: Fix initialism errors in codebase.

haardikdharma10:
#### Summary

Aims to solve #16623 partially.

#### Ticket Link

#16623 

#### Release Note

```release-note
NONE
```


haardikdharma10:
@agnivade, ready for review:)

plugin/hooks.go
```diff
@@ -15,27 +15,27 @@ import (
 // Feel free to add more, but do not change existing assignments. Follow the naming convention of
 // <HookName>Id as the autogenerated glue code depends on that.
 const (
-	OnActivateId             = 0
-	OnDeactivateId           = 1
+	OnActivateID             = 0
+	OnDeactivateID           = 1
 	ServeHTTPId              = 2
```
streamer45, line 20: Should this become `ServeHTTPID`?

plugin/hooks.go
```diff
@@ -15,27 +15,27 @@ import (
 // Feel free to add more, but do not change existing assignments. Follow the naming convention of
 // <HookName>Id as the autogenerated glue code depends on that.
```
streamer45, line 16: Let's also update the comment.

utils/html.go
```diff
@@ -120,14 +120,14 @@ func (t *HTMLTemplate) RenderToWriter(w io.Writer) error {
...
-	message := t(translationID, escapeForHtml(args))
+func TranslateAsHTML(t i18n.TranslateFunc, translationID string, args map[string]interface{}) template.HTML {
+	message := t(translationID, escapeforHTML(args))
 	message = strings.Replace(message, "[[", "<strong>", -1)
 	message = strings.Replace(message, "]]", "</strong>", -1)
 	return template.HTML(message)
 }
 
-func escapeForHtml(arg interface{}) interface{} {
+func escapeforHTML(arg interface{}) interface{} {
```
streamer45, line 130: Maybe this should become `escapeForHTML`.

streamer45 (commented):
Thank you @haardikdharma10 , very nice work on this :tada: 
There are some conflicts to address and I've left a couple of suggestions but otherwise it looks good :+1:

plugin/hooks.go
```diff
@@ -13,29 +13,29 @@ import (
...
-	UserHasLeftTeamId        = 12
-	ChannelHasBeenCreatedId  = 13
-	FileWillBeUploadedId     = 14
-	UserWillLogInId          = 15
-	UserHasLoggedInId        = 16
-	UserHasBeenCreatedId     = 17
-	ReactionHasBeenAddedId   = 18
-	ReactionHasBeenRemovedId = 19
-	TotalHooksId             = iota
+	OnActivateID             = 0
```
agnivade, line 18: I think we should leave them as-is. We saw downstream breakages because certain consumers import the `plugin` package too.

Anything which is not `model` or `plugin` should be good to be changed.

agnivade (changes requested):
Only one comment about the plugin changes. Looks good otherwise. Thanks @haardikdharma10 !

plugin/hooks.go
streamer45: Interesting, I am sure the plugin package is imported but what would be the use of those ids from the outside? As far as I can see they are only used by the RPC layer. I am okay with being extra cautious of course but I've thought about this as well and couldn't really imagine a use of them.

agnivade: Gotcha. All good from my side if they aren't being used.

Quick FYI to @levb @hanzei on this. Since we broke it last time with the variable case change.

agnivade (approved):

plugin/hooks.go
hanzei: I also _suppose_ that we can freely change them. @crspeller Could you please take a look given that you added this code?

agnivade:
@haardikdharma10 - There are some conflicts now :)

haardikdharma10:
> @haardikdharma10 - There are some conflicts now :)

Resolved @agnivade!

agnivade:
Oh wait, my apologies, @streamer45 is yet to review this. :facepalm:

streamer45:
> Oh wait, my apologies, @streamer45 is yet to review this.

Mainly waiting for an additional confirmation on the plugin hooks IDs.

plugin/hooks.go
crspeller: I don't think it will break existing plugins, since those IDs are only used within the server and the rpc client. The plugins don't use them directly, so they will all be updated at once. Cant guarantee that, but they shouldn't be. 
What will break however is exactly what the comment above this warns about. The generator of the RPC glue at `plugin/interface_generator/main.go` will break and needs to be modified to match.

streamer45: Thank you @crspeller. So we've essentially modified the generated file instead of the generator.

@haardikdharma10 , can you help changing the Id suffix in the generator file as well:

https://github.com/mattermost/mattermost-server/blob/8e71ab1ecb4d05df5c13652bd3f0c8c4e8d9b0e6/plugin/interface_generator/main.go#L294-L309

`{{.Name}}Id` becoming `{{.Name}}ID`

Then you can run `make pluginapi` to verify everything is generated properly.

haardikdharma10:
Hi @streamer45, is this PR good enough to get merged?

streamer45:
> Hi @streamer45, is this PR good enough to get merged?

Looks like it, although there's still a failure with translation.

streamer45:
Waiting for https://github.com/mattermost/mattermost-utilities/pull/50 to get merged. Changing `TranslateAsHtml` needs a change on the translation generator.

haardikdharma10:
All checks are passing @streamer45 🎉

streamer45 (approved):

PR #16882, opened 2021-02-08
[MM-32543] - Add ability to send email to admin that users are trying to invite others

AGMETEOR:
#### Summary
This PR adds the ability for users to alert sysadmins about the free tier user count limitations. This is used in the scenario on the webapp where users try to invite more users but fail.

Screenshot 2021-02-11 at 21 03 41

#### Ticket Link
[MM-32543](https://mattermost.atlassian.net/secure/RapidBoard.jspa?rapidView=79&projectKey=MM&modal=detail&selectedIssue=MM-32543&quickFilter=382)

#### Related Pull Requests
[Enterprise](https://github.com/mattermost/enterprise/pull/920)
[API SPEC](https://github.com/mattermost/mattermost-api-reference/pull/607)

#### Release Note
```release-note
Added new API endpoint GET /api/v4/cloud/subscription/limitreached/invite
```


api4/cloud.go
```diff
@@ -393,3 +395,30 @@ func handleCWSWebhook(c *Context, w http.ResponseWriter, r *http.Request) {
...
+		return
+	}
+
+	user, err := c.App.GetUser(c.App.Session().UserId)
+	if err != nil {
+		c.Err = err
+		return
+	}
+
+	sub, subErr := c.App.Cloud().GetSubscription()
```
marianunez, line 411: Why not just err here for consistency?

```suggestion
	sub, err := c.App.Cloud().GetSubscription()
```

app/cloud.go
```diff
@@ -18,6 +21,75 @@ func (a *App) getSysAdminsEmailRecipients() ([]*model.User, *model.AppError) {
...
+		return model.NewAppError("SendAdminsOverTheLimitAlertEmail",
+			"app.email.rate_limit_exceeded.app_error", map[string]interface{}{"RetryAfter": result.RetryAfter.String(), "ResetAfter": result.ResetAfter.String()},
+			fmt.Sprintf("username=%s, retry_after_secs=%f, reset_after_secs=%f",
+				username, result.RetryAfter.Seconds(), result.ResetAfter.Seconds()),
+			http.StatusRequestEntityTooLarge)
+	}
+
+	if subscription != nil && subscription.IsPaidTier == "true" {
+		return nil
+	}
```
marianunez, line 51: Should we move this before the rate limit check above and return early if not they are in paid tier?

app/cloud.go
```diff
@@ -18,6 +21,75 @@ func (a *App) getSysAdminsEmailRecipients() ([]*model.User, *model.AppError) {
...
+
+	sysAdmins, adminsErr := a.getSysAdminsEmailRecipients()
```
marianunez, line 53:
```suggestion
	sysAdmins, err := a.getSysAdminsEmailRecipients()
```

app/cloud.go
```diff
@@ -18,6 +21,75 @@ func (a *App) getSysAdminsEmailRecipients() ([]*model.User, *model.AppError) {
 	return a.GetUsers(userOptions)
 }
 
+// SendAdminsOverTheLimitAlertEmail takes the username of user trying to alert admins and then applies rate limit of n (number of admins) emails per user per day
+// before sending the emails.
+func (a *App) SendAdminsOverTheLimitAlertEmail(username string, subscription *model.Subscription) *model.AppError {
```
marianunez, line 26: Also, given that we do actually send an email to the admin if they go over the limit, could we rename here to maybe `sendAdminUpgradeRequestEmail`? This would cover the case for `invite` and `joining` as well

app/cloud.go
```diff
@@ -18,6 +21,75 @@ func (a *App) getSysAdminsEmailRecipients() ([]*model.User, *model.AppError) {
...
+	if subscription != nil && subscription.IsPaidTier == "true" {
+		return nil
+	}
+
+	sysAdmins, adminsErr := a.getSysAdminsEmailRecipients()
+	if adminsErr != nil {
+		return adminsErr
+	}
+
+	count, e := a.Srv().Store.User().Count(model.UserCountOptions{})
```
marianunez, line 58: We should be consistent on using `err` naming convention overall in the PR

```suggestion
	count, e := a.Srv().Store.User().Count(model.UserCountOptions{})
```

app/cloud.go
```diff
@@ -18,6 +21,75 @@ func (a *App) getSysAdminsEmailRecipients() ([]*model.User, *model.AppError) {
...
+		return model.NewAppError("app.SendAdminsOverTheLimitAlertEmail", "app.user.get_total_users_count.app_error", nil, e.Error(), http.StatusInternalServerError)
+	}
+
+	cloudUserLimit := *a.Config().ExperimentalSettings.CloudUserLimit
+
+	// we want to atleast have one email sent out to an admin
+	c := make(chan bool, len(sysAdmins))
+	for admin := range sysAdmins {
+		go func(admin int) {
+			ok, _ := a.Srv().EmailService.SendAdminOverTheLimitAlertEmail(count, int(cloudUserLimit), username, sysAdmins[admin].Email, sysAdmins[admin].Locale, *a.Config().ServiceSettings.SiteURL)
```
marianunez, line 69: Could we handle the err here so we can log why it's not sending them properly? 

Also, considering that the amount of system Admins is probably not high, what is the benefit of running those asynch?

app/email.go
```diff
@@ -619,6 +631,28 @@ func (es *EmailService) SendAtUserLimitWarningEmail(email string, locale string,
...
 
+// SendAdminOverTheLimitAlertEmail formats an email template and sends an email to an admin specified in the email arg
+func (es *EmailService) SendAdminOverTheLimitAlertEmail(tu int64, cul int, user, email, locale, siteURL string) (bool, *model.AppError) {
+	T := utils.GetUserTranslations(locale)
+
+	subject := T("api.templates.over_limit_subject", map[string]interface{}{"UserName": user})
+
+	bodyPage := es.newEmailTemplate("user_limit_admin_notification", locale)
+	bodyPage.Props["Title"] = T("api.templates.at_limit_title")
+	bodyPage.Props["Info4"] = T("api.templates.at_limit_stats", map[string]interface{}{"TotalUsers": tu, "CloudUserLimit": cul})
```
marianunez, line 642: Which email template is used for this email? It seems that the [Figma](https://www.figma.com/file/uv9A6dZsekQHHhOhYC7yc2/Cloud-UX?node-id=7861%3A888) linked in the ticket didn't have any refernece to the amount of seats taken / limit: 

Screen Shot 2021-02-09 at 5 43 07 PM

app/email_test.go
```diff
@@ -61,3 +61,38 @@ func TestSendInviteEmailRateLimits(t *testing.T) {
 	assert.Equal(t, "app.email.rate_limit_exceeded.app_error", err.Id)
 	assert.Equal(t, http.StatusRequestEntityTooLarge, err.StatusCode)
 }
+
+func TestSendAdminsOverTheLimitAlertEmail(t *testing.T) {
```
marianunez, line 65: Nice test 👍

marianunez (changes requested):
Awesome work @AGMETEOR! 

Wondering if it would make sense to include in this same PR https://mattermost.atlassian.net/browse/MM-32691 which is just a minor variation of the email you have implemented here. If you want to address in a separate PR that works too, but left some comments to start taking that into account.

app/email.go
AGMETEOR: Sorry. I just thought it would be useful for the admin to have some more context on why they're getting the email. But I have removed it now

app/cloud.go
AGMETEOR: This particular one a problem. I don't know why the compiler won't allow me to reuse ```err```

api4/cloud.go
```diff
@@ -39,6 +39,8 @@ func (api *API) InitCloud() {
 
 	// POST /api/v4/cloud/webhook
 	api.BaseRoutes.Cloud.Handle("/webhook", api.CloudApiKeyRequired(handleCWSWebhook)).Methods("POST")
+
+	api.BaseRoutes.Cloud.Handle("/alert/admin", api.ApiSessionRequired(sendOverLimitAlert)).Methods("POST")
```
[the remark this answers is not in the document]
AGMETEOR: For https://mattermost.atlassian.net/browse/MM-32691 I can address it in a separate PR

app/cloud.go
```diff
@@ -18,6 +21,59 @@ func (a *App) getSysAdminsEmailRecipients() ([]*model.User, *model.AppError) {
...
+	if e != nil {
+		return e
+	}
+
+	// we want to atleast have one email sent out to an admin
+	countNotOks := 0
+
+	for admin := range sysAdmins {
+		ok, err := a.Srv().EmailService.SendUpgradeEmail(username, sysAdmins[admin].Email, sysAdmins[admin].Locale, *a.Config().ServiceSettings.SiteURL)
+		if !ok {
```
marianunez, line 63: Not sure what the different scenarios are between `ok` and `err` but for completeness should we check here `err != nil` ?
```suggestion
		ok, err := a.Srv().EmailService.SendUpgradeEmail(username, sysAdmins[admin].Email, sysAdmins[admin].Locale, *a.Config().ServiceSettings.SiteURL)
		if !ok || err != nil {
```

templates/cloud_upgrade_request_email.html
```diff
@@ -0,0 +1,88 @@
...
+                            </tr>
+                        </table>
+                    </td>
+                </tr>
+            </table>
+        </td>
+    </tr>
+</table>
+
+{{end}}
```
marianunez, line 88: nit: missing end of file line

marianunez (approved):
LGTM, great work @AGMETEOR! 

Could you add a screenshot of the email sent for UX review?

api4/cloud.go
```diff
@@ -389,3 +390,30 @@ func handleCWSWebhook(c *Context, w http.ResponseWriter, r *http.Request) {
...
+
+func sendAdminUpgradeRequestEmail(c *Context, w http.ResponseWriter, r *http.Request) {
+	if c.App.Srv().License() == nil || !*c.App.Srv().License().Features.Cloud {
+		c.Err = model.NewAppError("Api4.sendAdminUpgradeRequestEmail", "api.cloud.license_error", nil, "", http.StatusNotImplemented)
+		return
+	}
+
+	user, err := c.App.GetUser(c.App.Session().UserId)
+	if err != nil {
+		c.Err = err
```
ethervoid, line 402: Please wrap this with a new model.AppError

api4/cloud.go
```diff
@@ -389,3 +390,30 @@ func handleCWSWebhook(c *Context, w http.ResponseWriter, r *http.Request) {
...
+
+	sub, err := c.App.Cloud().GetSubscription()
+	if err != nil {
+		c.Err = model.NewAppError("Api4.sendAdminUpgradeRequestEmail", "api.cloud.request_error", nil, err.Error(), http.StatusInternalServerError)
+		return
+	}
+
+	err = c.App.SendAdminUpgradeRequestEmail(user.Username, sub)
+	if err != nil {
+		c.Err = err
```
ethervoid, line 414: Same here

api4/cloud.go
```diff
@@ -389,3 +390,30 @@ func handleCWSWebhook(c *Context, w http.ResponseWriter, r *http.Request) {
...
+	}
+
+	sub, err := c.App.Cloud().GetSubscription()
+	if err != nil {
+		c.Err = model.NewAppError("Api4.sendAdminUpgradeRequestEmail", "api.cloud.request_error", nil, err.Error(), http.StatusInternalServerError)
+		return
+	}
+
+	err = c.App.SendAdminUpgradeRequestEmail(user.Username, sub)
+	if err != nil {
```
ethervoid, line 413:
```suggestion
	if err = c.App.SendAdminUpgradeRequestEmail(user.Username, sub); err != nil
```

app/cloud.go
```diff
@@ -18,6 +21,59 @@ func (a *App) getSysAdminsEmailRecipients() ([]*model.User, *model.AppError) {
...
+	if a.Srv().License() == nil || (a.Srv().License() != nil && !*a.Srv().License().Features.Cloud) {
+		return nil
+	}
+
+	if subscription != nil && subscription.IsPaidTier == "true" {
+		return nil
+	}
+
+	if a.Srv().EmailService.PerDayEmailRateLimiter == nil {
+		return model.NewAppError("app.SendAdminUpgradeRequestEmail", "app.email.no_rate_limiter.app_error", nil, fmt.Sprintf("username=%s", username), http.StatusInternalServerError)
```
ethervoid, line 36: Why don't use the params parameter of NewAppError to include the username?

app/cloud.go
```diff
@@ -18,6 +21,59 @@ func (a *App) getSysAdminsEmailRecipients() ([]*model.User, *model.AppError) {
...
+	if err != nil {
+		return model.NewAppError("app.SendAdminUpgradeRequestEmail", "app.email.setup_rate_limiter.app_error", nil, fmt.Sprintf("username=%s, error=%v", username, err), http.StatusInternalServerError)
+	}
+
+	if rateLimited {
+		return model.NewAppError("app.SendAdminUpgradeRequestEmail",
+			"app.email.rate_limit_exceeded.app_error", map[string]interface{}{"RetryAfter": result.RetryAfter.String(), "ResetAfter": result.ResetAfter.String()},
+			fmt.Sprintf("username=%s, retry_after_secs=%f, reset_after_secs=%f",
+				username, result.RetryAfter.Seconds(), result.ResetAfter.Seconds()),
+			http.StatusRequestEntityTooLarge)
```
ethervoid, line 50: Probably here fits better 429 instead of 413

app/cloud.go
```diff
@@ -18,6 +21,59 @@ func (a *App) getSysAdminsEmailRecipients() ([]*model.User, *model.AppError) {
...
+	}
+
+	sysAdmins, e := a.getSysAdminsEmailRecipients()
+	if e != nil {
+		return e
+	}
+
+	// we want to atleast have one email sent out to an admin
```
ethervoid, line 58:
```suggestion
	// we want to at least have one email sent out to an admin
```

app/email.go
```diff
@@ -41,37 +41,49 @@ func condenseSiteURL(siteURL string) string {
 }
 
 type EmailService struct {
-	srv              *Server
-	EmailRateLimiter *throttled.GCRARateLimiter
-	EmailBatching    *EmailBatchingJob
+	srv                    *Server
+	EmailRateLimiter       *throttled.GCRARateLimiter
```
ethervoid, line 45: Probably it's better to rename this one to `PerHourEmailRateLimiter`

app/email.go
```diff
@@ -41,37 +41,49 @@ func condenseSiteURL(siteURL string) string {
...
-	quota := throttled.RateQuota{
+	perHourQuota := throttled.RateQuota{
 		MaxRate:  throttled.PerHour(emailRateLimitingPerHour),
 		MaxBurst: emailRateLimitingMaxBurst,
 	}
 
-	rateLimiter, err := throttled.NewGCRARateLimiter(store, quota)
-	if err != nil || rateLimiter == nil {
+	perDayQuota := throttled.RateQuota{
+		MaxRate:  throttled.PerDay(1),
```
ethervoid, line 71: We're only allowing to send 1 email of this type per day right?

app/email.go
```diff
@@ -619,6 +631,28 @@ func (es *EmailService) SendAtUserLimitWarningEmail(email string, locale string,
...
+	bodyPage.Props["Title"] = T("api.templates.upgrade_request_title", map[string]interface{}{"UserName": user})
+	bodyPage.Props["Info4"] = T("api.templates.upgrade_request_info4")
+	bodyPage.Props["Info5"] = T("api.templates.at_limit_info5")
+	bodyPage.Props["SiteURL"] = siteURL
+	bodyPage.Props["Button"] = T("api.templates.upgrade_mattermost_cloud")
+	bodyPage.Props["EmailUs"] = T("api.templates.email_us_anytime_at")
+	bodyPage.Props["Footer"] = T("api.templates.copyright")
+
+	if err := es.sendMail(email, subject, bodyPage.Render()); err != nil {
+		return false, model.NewAppError("SendUpgradeEmail", "api.user.send_upgrade_request_email.error", nil, "err="+err.Error(), http.StatusInternalServerError)
```
ethervoid, line 650:
```suggestion
		return false, model.NewAppError("SendUpgradeEmail", "api.user.send_upgrade_request_email.error", nil, err.Error(), http.StatusInternalServerError)
```

app/email_test.go
```diff
@@ -61,3 +61,38 @@ func TestSendInviteEmailRateLimits(t *testing.T) {
 	assert.Equal(t, "app.email.rate_limit_exceeded.app_error", err.Id)
 	assert.Equal(t, http.StatusRequestEntityTooLarge, err.StatusCode)
 }
+
+func TestSendAdminUpgradeRequestEmail(t *testing.T) {
```
ethervoid, line 65: Could be possible to include a test for the rate limiter error?

templates/cloud_upgrade_request_email.html
```diff
@@ -0,0 +1,88 @@
...
+                                                    <tr>
+                                                        <td>
+                                                            <h2 style="margin-bottom: 0; text-align: center; color: #3D3C40; font-weight: bold; margin-top: 10px; line-height: 32px; font-size: 28px; font-family: Arial">{{ .Props.Title }}</h2>
+                                                        </td>
+                                                    </tr>
+                                                </table>
+                                                <p style="font-weight: normal; color: #3D3C40; font-size: 16px; line-height: 24px; font-family: Arial">
+                                                        {{ .Props.Info4 }}</p>
+                                                <p style="margin: 31px 0 31px">
+                                                    <a href="{{.Props.SiteURL}}/admin_console/billing/subscription" target="_blank"
```
ethervoid, line 36: Maybe we can pass the path as a property?

ethervoid (changes requested):
Great work! Some notes

app/cloud.go
marianunez: I believe this is Agnivas design for email rate limit errors: https://github.com/mattermost/mattermost-server/pull/15230

ethervoid: ^ @agnivade about the reason of using 413 instead of 429 for rate limiting?

app/cloud.go
AGMETEOR: Our translation id doesn't have any params included
```json
  {
    "id": "app.email.no_rate_limiter.app_error",
    "translation": "Rate limiter is not set up."
  }

```
In this case the username is just used by AppError's ```Error``` method

ethervoid: Yeah I mean, Why don't use `map[string]interface{}{"username": username}` in the parameters param and `err.Error()` in the error part of the `AppError` constructor?

app/email.go
AGMETEOR: That's true.

app/email_test.go
AGMETEOR: Do you mean with the time included? If so, it would be difficult to test an error like so 
```
expected: "app.SendAdminUpgradeRequestEmail: Invite emails rate limit exceeded. Timer will be reset after 23h59m59.864537s seconds. Please retry after 23h59m59.864537s seconds., username=un_eanktitjd38z7ngrdk6jzkirso, retry_after_secs=86399.864537, reset_after_secs=86399.864537
```

app/cloud.go
AGMETEOR: Something like this?

```go
	if a.Srv().EmailService.PerDayEmailRateLimiter == nil {
		return model.NewAppError("app.SendAdminUpgradeRequestEmail", "app.email.no_rate_limiter.app_error", map[string]interface{}{"username": username}, fmt.Sprintf("for username=%s", username), http.StatusInternalServerError)
	}
```

ethervoid: Oh, didn't see that this comes from a nil comparison. I'll use something like:

```
if a.Srv().EmailService.PerDayEmailRateLimiter == nil {
		return model.NewAppError("app.SendAdminUpgradeRequestEmail", "app.email.no_rate_limiter.app_error", map[string]interface{}{"username": username}, "", http.StatusInternalServerError)
	}
```

because it doesn't come from an error

app/email_test.go
ethervoid: Yeah we should be able to test that part but taking a closer look at how it works and we need to make a bigger refactor so no need for it by now :(

app/cloud.go
agnivade: It's explained in the PR body ;)

> The status code is chosen as 413 (entity too large) instead of 429 (too many requests)
because it's not the request which is rate limited, but the payload inside it which is.

agnivade: See https://github.com/mattermost/mattermost-server/pull/15230#issue-466523280 for a detailed context.

ethervoid: Noted! We can keep it as it is but one thing worth mentioning is that we also rate limiting by the number of emails sent (requests) so I kind of a mixed scenario

marianunez: @AGMETEOR can you revert your changes as you originally had them returning `http.StatusRequestEntityTooLarge` per Agniva's design.

app/cloud.go
AGMETEOR: Doing it like that would give us an error like so ```"detailed_error": "app.SendAdminUpgradeRequestEmail: Rate limiter is not set up., ",``` that's missing the username instead of the original that would be ```"detailed_error": "app.SendAdminUpgradeRequestEmail: Rate limiter is not set up., for username=test2",```

4/5 we leave it as it is

ethervoid: No hard thoughts on that. I think that if the constructor has the params it should appear in the logs, probably something is not in place but not important though. Resolving this one

ethervoid (approved):
LGTM! Great work buddy!

app/cloud.go
agnivade: > We can keep it as it is but one thing worth mentioning is that we also rate limiting by the number of emails sent (requests) so I kind of a mixed scenario

Yep, I explain it in more detail in the PR message. But essentially, it is rate limiting within rate limiting. There is no such response code to indicate that. I wanted to keep 429 purely to indicate rate limiting errors by the main rate limiter, and not by anything else internally. And returning a 429 error is also accompanied by `X-Retry-After` and `X-Reset-After`. We _could_ do `X-Email-Invite-Reset-After`, but I decided against it to keep the API clean.

ethervoid: Understood, thank you for the explanation Agniva :)

ethervoid:
@marianunez are we going to test this with the PR or after merge? // @stevemudie

marianunez:
> @marianunez are we going to test this with the PR or after merge? // @stevemudie

This should be tested post merge when @AGMETEOR completes the webapp portion that calls this endpoint to send the email.

stevemudie:
Will test post merge

PR #16892, opened 2021-02-08
Fixes environment variable override.

mkraft:
#### Summary

Allows `minio`, `elasticsearch`, and `openldap` to be disabled with the `ENABLED_DOCKER_SERVICES` environment variable.

#### Ticket Link

n/a

#### Release Note

```release-note
NONE
```


hahmadia (approved):
Ah so this was why I was getting services started that I hadn't initialized! Thanks for the fix

Makefile
```diff
@@ -138,14 +138,14 @@ ifeq ($(RUN_SERVER_IN_BACKGROUND),true)
 endif
 
 start-docker-check:
-ifeq (,$(findstring minio,$(ENABLED_DOCKER_SERVICES)))
+ifeq ($(findstring minio,$(ENABLED_DOCKER_SERVICES)),true)
```
jespino, line 141: I think this is not doing what you expect. This was added on purpose from somebody (I don't remember who) to enforce certain services (I didn't agree with that approach, but I don't think is wrong).

Accordingly to the GNU Make `findstring` documentation, `findstring minio,$(ENABLED_DOCKER_SERVICES)` should return `'minio'` if the `minio` string is there, or `''`, so... comparing it to `true` probably means "never take this path", so yes, it goes back to the previous behavior (disable/enable by hand) so... it would be easier to remove the whole `ifeq` blocks

jespino: I'm not suggesting anything here (my opinion is, it should be configured by default as expected, but should allow you to disable them, now is not allowed). For me that would mean remove entirely it, and verify the correct behavior on the situation where `config.override.mk` doesn't exist.

Specially for the case of enterprise code (if you have `BUILD_ENTERPRISE_READY`) it should add openldap and elasticsearch as defaults (but you should be able to disable it).

Again, all this is "my opinion", I suggest you to do a git blame and talk with the person that did this change.

mkraft: You're right. Wow [that's a super strange syntax](https://www.gnu.org/software/make/manual/html_node/Testing-Flags.html#Testing-Flags).

mkraft: Yeah, I think that `start-docker-check` can just be removed.

jespino: This was added by @agnivade and @streamer45 to enable that always in the tests. I don't think we should remove that without asking them before.

mkraft: For those of you just joining us, the goal is to be able to `make run-server` without elasticsearch, openldap, and minio always being enabled. That's where this is coming from.

agnivade: I think what I did was add the ldap container to run when EE directory was present. This was done to let the tests pass or something like that :thinking: 

But if CI passes, it's all good from my side :+1:

streamer45: This was added in https://github.com/mattermost/mattermost-server/pull/15217 which was fixing [MM-27485](https://mattermost.atlassian.net/browse/MM-27485).
Essentially some tests require services so the idea was to make sure these were running before starting the test. 
It would be also okay to fail fast and throw an error if the user is going to execute a set of tests requiring services which are not set to run. In any case I think we shouldn't assume the user knows what requires what when running (possibly lengthy) sets of tests.

agnivade: >  the goal is to be able to make run-server without elasticsearch, openldap, and minio always being enabled.

But this PR changes `make test-server` right? I don't think it's unreasonable to enforce a required dependency for a test to pass, if we want to run the test in the first place.

PR #16895, opened 2021-02-09
MM-32471: Use created user to update roles

agnivade:
During user creation via CLI, we would create the user,
but pass the user ID instead when updating the roles.

This falls into the category of read-after-write within a single request.
We fix this by passing the already created user and
directly update the roles.

https://mattermost.atlassian.net/browse/MM-32471

```release-note
NONE
```


wiggin77 (approved):
LGTM :+1:

streamer45 (commented):
Nice :+1: 
Could the same issue happen if the same action is performed through the API, e.g. with `mmctl`?

agnivade:
The API does not call the `UpdateUserRoles` so it's fine. I guess the CLI was meant to handle the edge case of creating the first user as sysadmin. And all of this happens because of the `--sysadmin` flag.

streamer45 (approved):

PR #16896, opened 2021-02-09
MM31503 - screen to restrict access because workspace capacity

pablovelezvidal:
#### Summary
This PR removes the session required from the /subscription/stats so it can be used to retrieve subscription information when users try to join a workspace via the invitation link and then decide if the user can join or not based on the stats information

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-31503

#### Release Note
```release-note
Removed session requered restriction from, GET api/v4/subscription/stats
```


ethervoid (approved):
LGTM! Those are not very security-concerning fields but in any case I'd ask @srkgupta to be sure

nickmisasi (approved):
LGTM

PR #16898, opened 2021-02-09
Reduce the coupling of the mailservice with the rest of the application

jespino:
#### Summary
Reduce the coupling of the mailservice with the rest of the application

#### Ticket Link
[MM-32885](https://mattermost.atlassian.net/browse/MM-32885)

#### Related PR:
enterprise: mattermost/enterprise#924

#### Release Note
```release-note
NONE
```


streamer45:
@jespino 

just so you know @isacikgoz is OoO in case you wanted this merged soon.

jespino:
@streamer45 there is no rush, but I'll change it anyway.

services/mailservice/mail.go
```diff
@@ -345,20 +357,6 @@ func SendMail(c smtpClient, mail mailData, fileBackend filesstore.FileBackend, d
 		m.EmbedReader(name, reader)
 	}
 
-	for _, fileInfo := range mail.attachments {
```
streamer45, line 348: So are we dropping support for attachments?

jespino: We are dropping "not exposed/not used" file attachments in the emails (except for the embedded files) that are supported and handled through other way. If we want to add back attachments in the future, I would recommend to add them without depending explicitly from the filestoreage package by itself (the coupling there wasn't needed event if we wanted to attach files).

streamer45: Sounds good :+1: We can always add a method in the future that takes files/readers.

streamer45 (approved):
Looks good :+1:

wiggin77 (approved):
LGTM :+1:

services/mailservice/mail.go
```diff
@@ -18,11 +18,29 @@ import (
 	gomail "gopkg.in/mail.v2"
 
 	"github.com/mattermost/mattermost-server/v5/mlog"
-	"github.com/mattermost/mattermost-server/v5/model"
-	"github.com/mattermost/mattermost-server/v5/services/filesstore"
-	"github.com/mattermost/mattermost-server/v5/utils"
 )
 
+const (
+	connSecurityTls      = "TLS"
```
agnivade, line 24: TLS and StartTLS

services/mailservice/mail.go
```diff
@@ -140,15 +157,15 @@ func ConnectToSMTPServerAdvanced(connectionInfo *SmtpConnectionInfo) (net.Conn,
 	return conn, nil
 }
 
-func ConnectToSMTPServer(config *model.Config) (net.Conn, error) {
+func ConnectToSMTPServer(config *SMTPConfig) (net.Conn, error) {
 	return ConnectToSMTPServerAdvanced(
 		&SmtpConnectionInfo{
```
agnivade, line 162: I am wondering if there is a need to duplicate the config in two structs? It seems like `SmtpConnectionInfo` already contains what we need. If we are exposing `SMTPConfig` to the outside world, can we just clean up `SmtpConnectionInfo` and use only one config struct throughout?

agnivade (commented):
Looks great. Spotted an opportunity for further cleanup.

services/mailservice/mail.go
jespino: Yes, that makes a lot of sense, for sure we can make it simpler.

jespino: done

services/mailservice/mail.go
jespino: Done

services/mailservice/mail.go
```diff
@@ -62,15 +80,15 @@ type SmtpConnectionInfo struct {
```
agnivade, line 80: We can actually remove the struct declaration above. L69.

agnivade (approved):
Looks amazing.

services/mailservice/mail.go
jespino: :facepalm:

streamer45 (approved):

lindy65 (approved):
Thanks @jespino - Global Relay email is working as expected as are email invitations, notification emails and emails received when changing password, etc. Looks good to me 👍

PR #16899, opened 2021-02-09
NO-TICKET Remove unused code and fix code smells

cupakob:
#### Summary
Remove unused code and fix some code smells

#### Release Note
```release-note
NONE
```

### DIFF
--- a/app/user.go
+++ b/app/user.go
@@ -352,19 +352,6 @@ func (a *App) CreateOAuthUser(service string, userData io.Reader, teamID string,
 		user.AuthService = service
 	}
 
-	suchan := make(chan store.StoreResult, 1)
-	euchan := make(chan store.StoreResult, 1)
-	go func() {
-		userByAuth, err := a.Srv().Store.User().GetByAuth(user.AuthData, service)
-		suchan <- store.StoreResult{Data: userByAuth, NErr: err}
-		close(suchan)
-	}()
-	go func() {
-		userByEmail, err := a.Srv().Store.User().GetByEmail(user.Email)
-		euchan <- store.StoreResult{Data: userByEmail, NErr: err}
-		close(euchan)
-	}()
-
 	found := true
 	count := 0
 	for found {
@@ -374,16 +361,17 @@ func (a *App) CreateOAuthUser(service string, userData io.Reader, teamID string,
 		}
 	}
 
-	if result := <-suchan; result.NErr == nil {
-		return result.Data.(*model.User), nil
+	userByAuth, _ := a.Srv().Store.User().GetByAuth(user.AuthData, service)
+	if userByAuth != nil {
+		return userByAuth, nil
 	}
 
-	if result := <-euchan; result.NErr == nil {
-		authService := result.Data.(*model.User).AuthService
-		if authService == "" {
+	userByEmail, _ := a.Srv().Store.User().GetByEmail(user.Email)
+	if userByEmail != nil {
+		if userByEmail.AuthService == "" {
 			return nil, model.NewAppError("CreateOAuthUser", "api.user.create_oauth_user.already_attached.app_error", map[string]interface{}{"Service": service, "Auth": model.USER_AUTH_SERVICE_EMAIL}, "email="+user.Email, http.StatusBadRequest)
 		}
-		return nil, model.NewAppError("CreateOAuthUser", "api.user.create_oauth_user.already_attached.app_error", map[string]interface{}{"Service": service, "Auth": authService}, "email="+user.Email, http.StatusBadRequest)
+		return nil, model.NewAppError("CreateOAuthUser", "api.user.create_oauth_user.already_attached.app_error", map[string]interface{}{"Service": service, "Auth": userByEmail.AuthService}, "email="+user.Email, http.StatusBadRequest)
 	}
 
 	user.EmailVerified = true


cupakob:
@hanzei the `work in progress` label can be removed (i think, i don't have the permissions for that)

utils/utils.go
```diff
@@ -36,7 +36,7 @@ func RemoveStringFromSlice(a string, slice []string) []string {
 
 // RemoveStringsFromSlice removes all occurrences of strings from slice.
 func RemoveStringsFromSlice(slice []string, strings ...string) []string {
-	newSlice := []string{}
+	var newSlice []string
```
reflog, line 39: `var newSlice []string` and `newSlice := []string{}` are not the same thing if you run it through json.Marshal, one will serialize as `nil` and the other as `[]`. Please revert this and other changes similar to it

cupakob: that's right, it's not the same, but the tests were successful. i reverted all similar changes.

reflog (approved):
LGTM

app/audit.go
```diff
@@ -85,9 +85,9 @@ func (a *App) LogAuditRecWithLevel(rec *audit.Record, level mlog.LogLevel, err e
 // MakeAuditRecord creates a audit record pre-populated with defaults.
 func (a *App) MakeAuditRecord(event string, initialStatus string) *audit.Record {
 	var userID string
-	user, err := user.Current()
+	currentUser, err := user.Current()
```
agnivade, line 88: I am wondering whether this is needed? `user` seems readable enough to me.

utils/markdown/inlines.go
```diff
@@ -549,7 +549,7 @@ func (p *inlineParser) parseAutolink(c rune) bool {
 		RawDestination: linkRange,
 		markdown:       p.markdown,
 	})
-	p.position += (link.End - link.Position)
+	p.position += link.End - link.Position
```
agnivade, line 552: If this is not changed by "gofmt -s -w", I am not sure if this is worth the code churn.

utils/textgeneration.go
```diff
@@ -42,7 +42,7 @@ ___
 	` **[2] - **[More Markdown Tests]**
 > i am a blockquote!
 
-> i am a 2nd multiline 
+> i am a 2nd multiline
```
agnivade, line 45: Please revert these. These are multi-line markdown strings, and the trailing space is for a reason.

utils/testutils/testutils.go
```diff
@@ -31,44 +26,3 @@ func ReadTestFile(name string) ([]byte, error) {
 	}
 	return data.Bytes(), nil
 }
-
```
agnivade, line 34: Please revert this. You cannot see this yet, but this function is called from EE code. :)

utils/textgeneration.go
```diff
@@ -425,19 +425,6 @@ var FuzzyStringsNames = []string{
 	"H҉̵̞̟̠̖̗̘Ȅ̐̐̑̒̚OMI҉̵̞̟̠",
 }
 
-// Strings that should pass as acceptable emails
```
agnivade, line 428: We should also remove L463-472.

agnivade (changes requested):
Nice work @cupakob. However, we might need to revert a few changes.

app/audit.go
cupakob: there is import "os/user", that is imho not a good idea

utils/markdown/inlines.go
cupakob: no, that was suggested by Goland and there is no additional value by the brackets

utils/testutils/testutils.go
cupakob: fixed

utils/textgeneration.go
cupakob: fixed

utils/textgeneration.go
cupakob: fixed

utils/textgeneration.go
agnivade: Hmm .. I still don't see them :thinking:

app/audit.go
agnivade: I agree that it can indeed be a bit confusing. However, in the context of the code, which is just 3 lines, I think it does not cause that much of a readability issue. The user can immediately see that `user.Uid` and `user.Username` come from the variable above. If it was quite high up in the code, then certainly it would make sense.

What do you think?

cupakob: I'm just following suggestion from the jetbrains knowledge (and i think, this community has a reason for this suggestion)  and the clean code rules of uncle Bob. I still think, `currentUser` is better identifier than `user`. But that is not my own code, i'm not the maintainer, so i've reverted the change.

utils/textgeneration.go
cupakob: sorry, my mistake

app/audit.go
agnivade: IMO `currentUser` makes complete sense if the variable was used in a long method/function. However, for just 3 lines of code, Go prefers to have shorter variable names.

To be more pedantic, the length of the variable name should be directly proportional to the longest distance between its instantiation and usage. https://github.com/golang/go/wiki/CodeReviewComments#variable-names.

Given that, ideally it should have been `usr`. But considering the fact that the code is already written, and every change has a cost, IMO changing it wouldn't be worth the cost it would have.

agnivade (approved):
Looks great!

app/audit.go
cupakob: i must say, the part with the variable names is new for me. But my own coding style is a little bit different. I still preffer longer variable names instad of `i` or `c`. But that is just my coding style :) everything is fine for me!

PR #16900, opened 2021-02-09
Changed require.Nil to require.NoError when error type is used

jespino:
#### Summary
Changed require.Nil to require.NoError when error type is used

#### Ticket Link
Not yet

#### Release Note
```release-note
NONE
```


wiggin77 (approved):
LGTM :+1:

agnivade (approved):

PR #16901, opened 2021-02-09
unparam lint

mahmud2011:
#### Summary
* fix "is never used" lint err

#### Ref

#### Ticket Link
https://github.com/mattermost/mattermost-server/issues/16622

#### Release Note
```release-note
None
```


config/utils.go
```diff
@@ -87,28 +87,20 @@ func desanitize(actual, target *model.Config) {
...
-			changed = true
 		}
 	}
 
-	if FixInvalidLocales(cfg) {
-		changed = true
-	}
-
-	return changed
+	_ = FixInvalidLocales(cfg)
```
streamer45, line 103:
```suggestion
	FixInvalidLocales(cfg)
```

Why the empty assignment?

app/channel.go
```diff
@@ -2245,7 +2245,10 @@ func (a *App) RemoveUserFromChannel(userIDToRemove string, removerUserId string,
 	}
 
 	if userIDToRemove == removerUserId {
-		a.postLeaveChannelMessage(user, channel)
+		err = a.postLeaveChannelMessage(user, channel)
+		if err != nil {
+			return err
+		}
```
streamer45, line 2251:
```suggestion
		if err = a.postLeaveChannelMessage(user, channel); err != nil {
			return err
		}
```

config/utils.go
```diff
@@ -87,28 +87,20 @@ func desanitize(actual, target *model.Config) {
 }
 
 // fixConfig patches invalid or missing data in the configuration, returning true if changed.
-func fixConfig(cfg *model.Config) bool {
-	changed := false
-
+func fixConfig(cfg *model.Config) {
```
streamer45, line 90: We should also update the comment.

streamer45 (commented):
Thank you @mahmud2011 :+1:  I've left a couple of suggestions.

config/utils.go
mahmud2011: explicitly ignoring return value

streamer45 (approved):

config/utils.go
```diff
@@ -86,29 +86,21 @@ func desanitize(actual, target *model.Config) {
 	}
 }
 
-// fixConfig patches invalid or missing data in the configuration, returning true if changed.
-func fixConfig(cfg *model.Config) bool {
-	changed := false
-
+// fixConfig patches invalid or missing data in the configuration.
+func fixConfig(cfg *model.Config) {
```
agnivade, line 90: @streamer45 - Just to confirm, the `changed` result is not required at the call site right? I can see that it's not used, but should it be?

streamer45: I am guessing it was needed originally but not since the last major change (feature flags). I am okay with removing it for now. All this code is getting reviewed soon anyway and if needed it's easy to add it back.

agnivade (approved):
Nice work!

PR #16906, opened 2021-02-10
Replacing require.nil in api layer

jespino:
#### Summary
Replacing require.nil in api layer

#### Ticket Link
[MM-32938](https://mattermost.atlassian.net/browse/MM-32938)

#### Release Note
```release-note
NONE
```


api4/role_test.go
```diff
@@ -26,7 +26,7 @@ func TestGetRole(t *testing.T) {
 	}
 
 	role, err := th.App.Srv().Store.Role().Save(role)
-	assert.Nil(t, err)
+	assert.NoError(t, err)
```
agnivade, line 29: These should be `require.NoError` IMO to prevent the tests from moving further.

api4/team_test.go
```diff
@@ -3075,19 +3075,19 @@ func TestSetTeamIcon(t *testing.T) {
...
-	teamAfter, err := th.App.GetTeam(team.Id)
-	require.Nil(t, err)
+	teamAfter, appErr := th.App.GetTeam(team.Id)
+	require.Nil(t, appErr)
 	assert.True(t, teamBefore.LastTeamIconUpdate < teamAfter.LastTeamIconUpdate, "LastTeamIconUpdate should have been updated for team")
 
 	info := &model.FileInfo{Path: "teams/" + team.Id + "/teamIcon.png"}
 	err = th.cleanupTestFile(info)
-	require.Nil(t, err, err)
+	require.NoError(t, err, err)
```
agnivade, line 3090: `require.NoError(t, err)`

api4/user_test.go
```diff
@@ -1720,7 +1720,7 @@ func TestPatchUser(t *testing.T) {
 	require.Empty(t, ruser.Timezone["manualTimezone"], "manualTimezone should update properly")
 
 	err := th.App.CheckPasswordAndAllCriteria(ruser, *patch.Password, "")
-	assert.Error(t, err, "Password should not match")
+	assert.NotNil(t, err, "Password should not match")
```
agnivade, line 1723: `require.NotNil`

agnivade (approved):
Looks great. No blocking comments.

api4/integration_action_test.go
```diff
@@ -237,7 +237,7 @@ func TestSubmitDialog(t *testing.T) {
 	ts := httptest.NewServer(http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
 		var request model.SubmitDialogRequest
 		err := json.NewDecoder(r.Body).Decode(&request)
-		require.Nil(t, err)
+		require.NoError(t, err)
 		assert.NotNil(t, request)
```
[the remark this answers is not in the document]
jespino: Actually, the pointer is always to be not-nil because is a pointer to a variable that always exists. I think we can remove this line because later is checking the values of request (so... in case of no error, it should have the proper data).

wiggin77 (approved):
LGTM :+1:

PR #16911, opened 2021-02-11
MM-30882: Fix read-after-write issue for demoting user

agnivade:
In (*App).DemoteUserToGuest, we would demote a user, and then immediately
read it back to do future operations from the user. This reading back
of the user had the effect of sticking the old value into the cache
after which it would never be updated.

There was another issue along with this, which was when the invalidation
message would broadcast across the cluster, it would hit the cache invalidation
problem where an unrelated store call would miss the cache because
it was invalidated, and then again read from replica and stick the old value.

To fix all these, we return the new value directly from the store method
to avoid having the app to read it again.

And we add a map in the localcache layer which tracks invalidations made,
and then switch to use master if it's true.

The core change is fairly limited, but due to changing the store method signatures,
a lot of code needed to be updated to pass "context.Background". Therefore the PR
just "appears" to be big, but the main changes are limited to app/user.go,
sqlstore/user_store.go and user_layer.go

https://mattermost.atlassian.net/browse/MM-30882

```release-note
Fix an issue where demoting a user to guest would not take effect in
an environment with read replicas.
```


store/sqlstore/user_store.go
```diff
@@ -326,28 +327,41 @@ func (us SqlUserStore) UpdateMfaActive(userId string, active bool) error {
...
 	if err != nil {
 		return nil, errors.Wrap(err, "users_get_many_tosql")
 	}
 
+	var db *gorp.DbMap
+	if hasMaster(ctx) {
+		db = us.GetMaster()
+	} else {
+		db = us.GetReplica()
+	}
```
streamer45, line 342: Should this become a function?

app/bot.go
```diff
@@ -4,6 +4,7 @@
 package app
 
 import (
+	"context"
```
streamer45, line 7: Do we have a future use for the context other than signaling where to run the query?
All this use of `context.Background()` feels a bit heavy on the code and possibly unnecessary in most cases.

store/sqlstore/user_store.go
agnivade: Totally, I am just refraining from too much DRY right at this moment, since we have just started to do this. But it's clear that this needs to be refactored. I will send another PR after this.

app/bot.go
agnivade: Indeed, it looks a bit heavy. But I think passing context is the right choice here. It's just that we aren't using context properly in our code. I would like to pass the context with a timeout to the actual DB query itself so that they can be properly cancelled if needed.

All store APIs should have passed a context anyways to start with. But it just looks like a lot of work because we never did that.

Thoughts?

streamer45: Sounds reasonable but the context should always come from the upper layer, correct? I see several additions in the store/search layer where a context is born out of nothing. So I am guessing it's really more of a `context.TODO()` for now.

agnivade: Fully agree. The context should come from the API layer, and _then_ passed on to the app layer. This would also help the open tracing layer which solves this by dynamically inserting contexts into the generated methods, but because of the fact that the contexts are not passed from the app layer to the store layer, we now have two different tracing layers, whereas ideally only a single context would need to be passed.

So yeah .. the context should come from upper layers. And I would very much like to see that in the near future :)

store/localcachelayer/user_layer.go
```diff
@@ -66,22 +85,34 @@ func (s LocalCacheUserStore) InvalidateProfilesInChannelCacheByUser(userId strin
...
 		if err := s.rootStore.doStandardReadCache(s.rootStore.profilesInChannelCache, channelId, &cachedMap); err == nil {
 			return cachedMap, nil
 		}
 	}
 
-	userMap, err := s.UserStore.GetAllProfilesInChannel(channelId, allowFromCache)
+	// If it was invalidated, then we need to query master.
+	s.profilesInChannelMut.Lock()
+	if s.profilesInChannelInvalidations[channelId] {
+		ctx = sqlstore.WithMaster(ctx)
```
streamer45, line 109: I guess it's okay and cleaner this way but technically getting the context could be done outside the critical section.

streamer45 (approved):
Looks good :+1:

store/localcachelayer/user_layer.go
agnivade: Hmm .. how so? Do you mean having an additional flag and then if that flag is true, then set the context? :thinking:

streamer45: Yes, like you are doing with `fromMaster` in some places. I don't know if getting the context is an expensive operation or not so I think it's fine to leave it as is unless we have some data showing we should move it out.

agnivade: Ah I see. This code does not seem to be in the hot path, because it'll only happen for cache misses. And `context.WithValue` only returns a struct, so there's not really that much of a time spent in contention.

```
func WithValue(parent Context, key, val interface{}) Context {
	if parent == nil {
		panic("cannot create context from nil parent")
	}
	if key == nil {
		panic("nil key")
	}
	if !reflectlite.TypeOf(key).Comparable() {
		panic("key is not comparable")
	}
	return &valueCtx{parent, key, val}
}
```

But you are right, I'll make a note to run a load test when this gets merged. Just to confirm.

nronas (approved):
Noice, LGTM

agnivade:
@srkgupta - Would you want to QA this PR? I already verified this manually myself, and on master I was able to repro pretty reliably in 2-3 times. But with this PR, I tried like 10 times and failed to repro.

Here are my cloud servers if you want to take a look:
PR: guestuser.test.mattermost.cloud
Master: agnivamain.test.mattermost.cloud

Usual sysadmin credentials.

srkgupta:
Ok sure @agnivade. Let me take a quick look at this and test this. Thanks.

srkgupta (approved):
Tested and the changes are working fine. Approving the PR. Thanks @agnivade

PR #16912, opened 2021-02-11
Cleanup unnecessary concurrent db calls on user

nronas:
#### Summary
Remove unnecessary concurrent DB calls, and replace them with JIT fetches instead. For scenarios that the OAuth user is present, we do not need to fetch the user by email at all, so instead of always fetching both, we use a JIT retrieval.

#### Ticket Link
N/A

#### Release Note

```release-note
* Use JIT db fetches on User. CreateOAuthUser
```

### DIFF
--- a/app/user.go
+++ b/app/user.go
@@ -352,19 +352,6 @@ func (a *App) CreateOAuthUser(service string, userData io.Reader, teamID string,
 		user.AuthService = service
 	}
 
-	suchan := make(chan store.StoreResult, 1)
-	euchan := make(chan store.StoreResult, 1)
-	go func() {
-		userByAuth, err := a.Srv().Store.User().GetByAuth(user.AuthData, service)
-		suchan <- store.StoreResult{Data: userByAuth, NErr: err}
-		close(suchan)
-	}()
-	go func() {
-		userByEmail, err := a.Srv().Store.User().GetByEmail(user.Email)
-		euchan <- store.StoreResult{Data: userByEmail, NErr: err}
-		close(euchan)
-	}()
-
 	found := true
 	count := 0
 	for found {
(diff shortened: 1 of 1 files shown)


reflog (approved):
LGTM

app/user.go
```diff
@@ -374,16 +361,17 @@ func (a *App) CreateOAuthUser(service string, userData io.Reader, teamID string,
 		}
 	}
 
-	if result := <-suchan; result.NErr == nil {
-		return result.Data.(*model.User), nil
+	userByAuth, _ := a.Srv().Store.User().GetByAuth(user.AuthData, service)
```
agnivade, line 364: Let's handle the error

app/user.go
```diff
@@ -374,16 +361,17 @@ func (a *App) CreateOAuthUser(service string, userData io.Reader, teamID string,
...
+	if userByAuth != nil {
+		return userByAuth, nil
 	}
 
-	if result := <-euchan; result.NErr == nil {
-		authService := result.Data.(*model.User).AuthService
-		if authService == "" {
+	userByEmail, _ := a.Srv().Store.User().GetByEmail(user.Email)
```
agnivade, line 369: Here too.

app/user.go
nronas: Actually, we do handling the error implicitly, if both user fetches are nil then we proceed on creating a user from scratch

agnivade: Ah yes I see now. Perhaps might be good to comment that to help with readability.

agnivade (approved):

app/user.go
mkraft: I don't think this code is exactly equivalent. The old code said "if retrieving the user by service x didn't throw an error then return the user". The new code says "if the returned object is not nil then return the user". One must look into the implementation of `(SqlUserStore).GetByAuth` to know whether a `nil` value is definitely returned in error cases. It wouldn't be unheard of for an instantiated struct to be returned with an error, in which case this logic would break unbeknownst to the developer refactoring `(SqlUserStore).GetByAuth`.

PR #16913, opened 2021-02-11
Cleanup unnecessary concurrent db calls on post and team

nronas:
#### Summary
Continuation of the sequel of async DB call retrievals. See for context: https://github.com/mattermost/mattermost-server/pull/16912

#### Ticket Link
N/A

#### Release Note
```release-note
* Replace async DB calls on team and post with JIT retrievals
```

### DIFF
--- a/app/post.go
+++ b/app/post.go
@@ -181,16 +181,6 @@ func (a *App) CreatePost(post *model.Post, channel *model.Channel, triggerWebhoo
 
 	post.SanitizeProps()
 
-	var pchan chan store.StoreResult
-	if post.RootId != "" {
-		pchan = make(chan store.StoreResult, 1)
-		go func() {
-			r, pErr := a.Srv().Store.Post().Get(post.RootId, false, false, false)
-			pchan <- store.StoreResult{Data: r, NErr: pErr}
-			close(pchan)
-		}()
-	}
-
 	user, nErr := a.Srv().Store.User().Get(post.UserId)
 	if nErr != nil {
 		var nfErr *store.ErrNotFound
@@ -231,12 +221,11 @@ func (a *App) CreatePost(post *model.Post, channel *model.Channel, triggerWebhoo
 
 	// Verify the parent/child relationships are correct
 	var parentPostList *model.PostList
-	if pchan != nil {
-		result := <-pchan
-		if result.NErr != nil {
+	if post.RootId != "" {
+		parentPostList, nErr = a.Srv().Store.Post().Get(post.RootId, false, false, false)
+		if nErr != nil {
 			return nil, model.NewAppError("createPost", "api.post.create_post.root_id.app_error", nil, "", http.StatusBadRequest)
 		}
-		parentPostList = result.Data.(*model.PostList)
 		if len(parentPostList.Posts) == 0 || !parentPostList.IsChannelId(post.ChannelId) {
 			return nil, model.NewAppError("createPost", "api.post.create_post.channel_root_id.app_error", nil, "", http.StatusInternalServerError)
 		}
--- a/app/team.go
+++ b/app/team.go
@@ -480,43 +480,10 @@ func (a *App) sendUpdatedMemberRoleEvent(userID string, member *model.TeamMember
 }
 
 func (a *App) AddUserToTeam(teamID string, userID string, userRequestorId string) (*model.Team, *model.AppError) {
-	tchan := make(chan store.StoreResult, 1)
-	go func() {
-		team, err := a.Srv().Store.Team().Get(teamID)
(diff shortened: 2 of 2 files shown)


reflog (approved):
nice!

app/post.go
```diff
@@ -231,12 +221,11 @@ func (a *App) CreatePost(post *model.Post, channel *model.Channel, triggerWebhoo
 
 	// Verify the parent/child relationships are correct
 	var parentPostList *model.PostList
-	if pchan != nil {
-		result := <-pchan
-		if result.NErr != nil {
+	if post.RootId != "" {
```
agnivade, line 224: Hmm .. creating posts are a hot path and should be fast. I _think_ I am okay with it, but it would be good to do a quick micro-benchmark to ensure we aren't losing out on performance.

app/team.go
```diff
@@ -1805,33 +1632,20 @@ func (a *App) RestoreTeam(teamID string) *model.AppError {
...
-		tchan <- store.StoreResult{Data: totalMemberCount, NErr: err}
-		close(tchan)
-	}()
-	achan := make(chan store.StoreResult, 1)
-	go func() {
-		memberCount, err := a.Srv().Store.Team().GetActiveMemberCount(teamID, restrictions)
-		achan <- store.StoreResult{Data: memberCount, NErr: err}
-		close(achan)
-	}()
+	totalMemberCount, err := a.Srv().Store.Team().GetTotalMemberCount(teamID, restrictions)
```
agnivade, line 1635: This one we should keep as-is. These queries are heavy and gives a good performance boost to keep it asynchronous.

app/team.go
```diff
@@ -480,43 +480,10 @@ func (a *App) sendUpdatedMemberRoleEvent(userID string, member *model.TeamMember
...
-	result = <-uchan
-	if result.NErr != nil {
-		var nfErr *store.ErrNotFound
-		switch {
-		case errors.As(result.NErr, &nfErr):
-			return nil, model.NewAppError("AddUserToTeam", MissingAccountError, nil, nfErr.Error(), http.StatusNotFound)
-		default:
-			return nil, model.NewAppError("AddUserToTeam", "app.user.get.app_error", nil, result.NErr.Error(), http.StatusInternalServerError)
-		}
+	user, team, err := a.prepareInviteNewUsersToTeam(teamID, userID)
```
agnivade, line 483: So as I see it, all of these changes basically make these calls synchronous. These actions aren't in the hot path, so it should be fine. But I am not sure if I would call it "unnecessary". They are just written in this way because we are checking for the error values one after the other.

But if we did something like:

```go
g := new(errgroup.Group)
var a,b Item
  g.Go(func() error {
  	var err error
  	a, err = Foo()
  	return err
	})

g.Go(func() error {
  	var err error
  	b, err = Bar()
  	return err
	})

g.Wait()
```

Then the intention is much more clear.

nronas: Yeah we can defo do that, the usage of the word "unnecessary" maybe is overpowering here or incorrect, but my sole intention with these was the fact that in case of an error on the top level query we are unnecessarily fetching everything else, also the code does not read really well, with all the channel wait calls.

app/team.go
nronas: will see if we can optimise the underlying call on this before reverting back to the original implementation.

app/post.go
nronas: Will defo do a benchmark for these, good point.

app/team.go
agnivade: > in case of an error on the top level query we are unnecessarily fetching everything else, 

Yeah, so that will be the case for all concurrent calls that we are doing. I think maybe you wanted to say that our "implementation" of concurrent calls is not very readable with the way we pass errors with channels and all?

That is very true and I completely agree with it. We can move to using the `errgroup` pattern, which makes it much more readable IMO. But making all of them synchronous is a different solution I think.

agnivade: Also, your first PR fixed a different problem. In that case, we were completely discarding the second query when we got the results from the first one. That case is a completely valid case of conditionally issuing synchronous calls because that's the happy path. But these are just normal concurrent calls from all of which, we need the results.

nronas: Totally, my only view here, is if those calls need to be async in the first place. Like if all take tens of ms, a user will not notice the difference, between promise-based calls and sync calls. But you mentioned that some of those are heavy once, so let me profile them and see what we can do there. Maybe just optimising the queries will be a good start.

agnivade: Ah yes, optimising those queries definitely sounds like a good plan. Thanks John!

nronas:
Closing this since, further profiling on those queries needed.

PR #16914, opened 2021-02-11
Extracting i18n functionality to i18n core library

jespino:
#### Summary
Extracting i18n functionality to i18n core library.

This also isolates the rest of the mattermost code from the go-i18n library, so if we need to change or update it in the future, it should be easier.

#### Ticket Link
[MM-32873](https://mattermost.atlassian.net/browse/MM-32873)

#### Related PR
enterprise: mattermost/enterprise#923

#### Release Note
```release-note
NONE
```


agnivade:
I want to decide the name of the directory first. I think we can do better than `corelibs` :stuck_out_tongue: Any suggestions?

jespino:
@agnivade my first instinct was "core", but core has other meaning. I can think about `libs` (that is really boring and doesn't say much). Other option can be `sharedlibs` but can lead to some confusion too. For me `corelibs` is the best option that I have though so far. But of course, I'm open to any suggestion.

agnivade:
Indeed, naming is hard! I would want to have a single word as the name. Following [effective go](https://golang.org/doc/effective_go.html#package-names):

> By convention, packages are given lower case, single-word names; there should be no need for underscores or mixedCaps.

Here are some of my thoughts:
- shared
- common

jespino:
In that case I would prefer `libs` or `shared` (`common` sounds like `utils` to me, a great excuse to put random stuff there without any good reason).

Anyway, all the proposed names are very general.

agnivade:
We can go with `shared`.

mgdelacroix:
Should we consider [using `pkg` for this directory](https://github.com/golang-standards/project-layout#pkg)? //cc @nronas

jespino:
@agnivade I like the pkg idea proposed by @mgdelacroix, what do you think?

mgdelacroix:
Merit goes to @nronas on this one :smile:

agnivade:
> This is a common layout pattern, but it's not universally accepted and some in the Go community don't recommend it.

I am one of them. :wink: To give some history on this, `pkg` was initially used by the Go team when there were no standards around how to design packages. And then later it was removed because it did not go with the Go philosophy of not creating grab-bag of packages, but rather have well-named packages which explain what the code inside them contain. Using `pkg` is an anti-pattern now.

Also note:

> Note that this is not a universally accepted pattern and for every popular repo that uses it you can find 10 that don't. It's up to you to decide if you want to use this pattern or not. 

I am 0/5 that `shared` is the right choice for this, but as long as it's a one word name that explains what the code under the directory contains, I am good.

nronas:
> it did not go with the Go philosophy of not creating grab-bag of packages, but rather have well-named packages which explain what the code inside them contain

So I wonder, are we not doing the same with using a folder named `shared`, essentially we are using shared as pkg, like shared is a grab-bag of packages :D.

jespino:
@nronas we are "namespacing" the packages that are expected to be shared with the rest of the world, I want to have them organized to be able to apply special rules if needed, and make it easy to discover the packages that are aim to be reused, by other Mattermost Inc. apps or plugins, for example.

nronas:
@jespino totally get the aim, but i was commenting on the whole grab-bag situation. Since the definition of the pkg directory is everything that can be shared/used from external applications, which essentially is what we are doing. Again -1/5 on this 😄. Please go ahead with whatever makes more sense.

mgdelacroix:
Same here, I didn't know the `pkg` approach and it did sound like what we're doing, but if we want to avoid using it not to confuse people with something that is now an anti-pattern, I'm totally happy with using `shared`.

jespino:
Ok, I think we'll go with `shared` and revisit this later if is needed.

agnivade:
> Since the definition of the pkg directory is everything that can be shared/used from external applications, which essentially is what we are doing.

Yes, so it's a made up definition. The word `pkg` by itself does not tell the reader that all the packages under this is okay to be shared by the outside world. Whereas `shared` is much clearer in it's intention. Somebody new to Go does not need to read through the patterns to understand what `shared` means. It is immediately obvious.

agnivade (approved):

nronas (approved):

mgdelacroix (approved):

furqanmlk (approved):
### English
image

image


### Spanish 
image

image

### French 
image

image

PR #16915, opened 2021-02-11
Moving mlog to corelibs

jespino:
#### Summary
Moving mlog to corelibs

#### Ticket Link
Not yet

#### Related PR
enterprise: mattermost/enterprise#922

#### Release Note
```release-note
NONE
```


agnivade:
I gave some more thought into this. And I don't think moving `mlog` to a separate package semantically makes sense to me. I don't want to have the core package be a dumping ground for all possible refactorable packages.

Logging should be a top level package for a repo, and the current hierarchy seems perfect to me. If any downstream consumers want to import it, they should use `mattermost-server/v5/mlog`. As you said earlier in the call, other packages like i18n or mail can depend on logging, it feels like a base package which provides logging service to all other packages.

Keep those in mind, I think this should remain where it is, atleast in this repo.

jespino:
I think moving this would easy the discovery of things that are expected to be shared with other projects, but other than that, I don't have strong feelings about keeping it or moving it.

The dependency of the libraries between each other, is something that can happen anyway, so I don't see that as a necessary restriction.

agnivade:
One thing that I am a bit concerned about is the downstream breakage of this move. This package is used by a lot of consumers, including load-test, mattermod, bifrost. And moving this will break all of them trying to upgrade their Mattermost dependency. The other packages like i18n and markdown are being extracted out of utils, so downstream consumers won't be affected. But the mlog package is a bit different.

What do you think?

jespino:
I understand that point, but it is exactly what I would expect. We should let the people know that we are moving that there, and everybody that is using this from the outside, should update his code.

The whole point of having this set of libraries there is exactly having the right context inside and outside mattermost-server about that library, "this is shared".

We can defer it to a bit later, to make everybody onboard before disrupting them with this. But I think, eventually, this library makes sense with the rest of the "shared" libraries.

agnivade:
> We can defer it to a bit later, to make everybody onboard before disrupting them with this.

Alright, I think if we announce this broadly in the dev meeting, and then do it, it should be fine.

> But I think, eventually, this library makes sense with the rest of the "shared" libraries.

I agree with that. Logging can and should be a shared and standardized component. :)

I only disagree with the location of this package being inside a sub-package of `mattermost-server`. :( If we truly want `mlog` to be shared by others, the proper way would be to move `mlog` to its own shared repository.

jespino:
@agnivade Yes, from a purist point of view, I totally agree with you, we should move this into its own package, but the maintenance burden of this can slowdown the mattermost-server development process, and it is the main restriction that I've set for this project is to not add maintenance burden to the mattermost-server team.

agnivade:
Yep, completely understand. And that's why I preferred to leave it in it's own place if it's not causing any harm. We are not gaining much except the "observability" part by moving this inside `shared`. If reusing packages by other repos is our main objective, then I think that's already fulfilled.

jespino:
@agnivade observability is important too, we can easy forget what is the plans for the `mlog` library and start adding unnecesary dependencies with the rest of the code. Again, I'm not against that, but I think everything that we want to be shared should be in the same place because consistency. If we we have a `mattermost/mattermost-server/v5/shared/` list of packages, and then something completely independent like `mattermost/mattermost-server/v5/mlog`, is `mlog` aim to be shared? should I hesitate before using it?

agnivade:
Fair enough, point taken :) No objections from my side, but let's announce it in the dev meeting so that everyone is aware. Thanks a lot Jesús for the discussion.

wiggin77:
@jespino is this ready to go?  I thought I read somewhere you were going to change `corelibs` to `shared`.

jespino:
@wiggin77 now is ready.

wiggin77 (approved):
LGTM

agnivade (approved):

mgdelacroix (approved):

jespino:
@lindalumitchell This should be affecting everything in the source code, but at the same time should be catch everything by the compiler. I don't think this need specific testing, I would say this can be QA deffered and confirm the correct behavior as part of the regular release testing.

If this gets broken we will notice that pretty quickly :)

lindalumitchell (approved):

PR #16922, opened 2021-02-12
Set required go version to 1.15

rvillablanca:
#### Summary
`SetConnMaxIdleTime` was added in go 1.15 so we need to upgrade it in the module.

This method is being used [here](https://github.com/mattermost/mattermost-server/blob/2846ad9f934c72109976a118e7336e26c33aeb43/store/sqlstore/store.go#L271)

#### Ticket Link

NONE

#### Release Note

```release-note
NONE
```

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -1,6 +1,6 @@
 module github.com/mattermost/mattermost-server/v5
 
-go 1.14
+go 1.15
 
 require (
 	code.sajari.com/docconv v1.1.1-0.20200701232649-d9ea05fbd50a


rvillablanca:
@agnivade :eyes:

agnivade:
This isn't strictly speaking necessary as the change was internal to the store package and no downstream consumers of the model or the plugin APIs are affected. Whereas a lot of consumers of our packages might not be updated to the latest.

I am wondering if you actually ran across this to be a problem somewhere?

rvillablanca:
I ran into this when compiling the whole server project where I was using v1.14 by default

agnivade:
Ok, so you saw a compilation error right? I am wondering why does the `go.mod` version needs to be bumped? That only affects downstream consumers of the package.

rvillablanca:
mmm ok you are right, I can compile the code just using go 1.15 or above so that means it is not required to change the `go` directive in the go.mod file. Anyway, according to the documentation of the directive, it seems to indicate which features are available when compiling the code https://golang.org/ref/mod#go-mod-file-go.

I believe I'm misunderstanding the usage of go directive in the go.mod file :thinking:, not sure if that is only for consumers of the package though.

agnivade:
No you are right Rodrigo. My apologies for the confusion. Sorry about that.

agnivade (approved):
Thanks Rodrigo!

PR #16924, opened 2021-02-12
NO-TICKET cleanup, fix possible resource leaks and bad smells

cupakob:
#### Summary
cleanup, fix possible resource leaks and bad smells

#### Release Note
```release-note
NONE
```

### DIFF
--- a/app/audit.go
+++ b/app/audit.go
@@ -18,13 +18,6 @@ import (
 	"github.com/mattermost/mattermost-server/v5/store"
 )
 
-const (
-	RestLevelID        = 240
-	RestContentLevelID = 241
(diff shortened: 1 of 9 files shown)


agnivade:
Hello @cupakob - thank you so much for taking initiatives like these! For the sake of making reviews easier, I'd request you to  kindly split the changes into groups which do only a single thing. For example, if we are removing unused constants or making error strings start with a small letter, then they are unrelated things and should go in separate PRs. Thanks!

cupakob:
@agnivade sure, i'll do it. One more question - should i make one PR for each group or one commit for each group?

cupakob:
i will separate the changes into different PRs

PR #16925, opened 2021-02-12
Isolate more the mfa service

jespino:
#### Summary
Isolate more the mfa service

#### Ticket Link
[MM-33010](https://mattermost.atlassian.net/browse/MM-33010)

#### Release Note
```release-note
NONE
```

### DIFF
--- a/services/mfa/mfa.go
+++ b/services/mfa/mfa.go
@@ -4,6 +4,8 @@
 package mfa
 
 import (
+	"crypto/rand"
+	"encoding/base32"
 	"fmt"
 	"net/http"
 	"net/url"
@@ -13,8 +15,6 @@ import (
 	"github.com/mattermost/rsc/qr"
 
 	"github.com/mattermost/mattermost-server/v5/model"
-	"github.com/mattermost/mattermost-server/v5/services/configservice"
-	"github.com/mattermost/mattermost-server/v5/store"
 )
 
 const (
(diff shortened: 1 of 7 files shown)


services/mfa/mfa.go
```diff
@@ -93,47 +90,51 @@ func (m *Mfa) Activate(user *model.User, token string) *model.AppError {
...
-		return err
-	}
+// Deactivate set the mfa as deactive, remove the mfa secret, store it with the StoreActive and StoreSecret functions provided
+func Deactivate(storeSecretFunc StoreSecret, storeActiveFunc StoreActive, userId string) error {
+	var wg sync.WaitGroup
+	wg.Add(2)
 
-	schan := make(chan error, 1)
+	var secretErr error
+	var activeErr error
```
agnivade, line 113: This is not a hot path by any means. Let's just make these synchronous.

services/mfa/mfa.go
```diff
@@ -52,39 +55,33 @@ func getIssuerFromUrl(uri string) string {
 	return url.QueryEscape(issuer)
 }
 
-func (m *Mfa) GenerateSecret(user *model.User) (string, []byte, *model.AppError) {
-	if err := m.checkConfig(); err != nil {
-		return "", nil, err
-	}
+// GenerateSecret generates a new user mfa secret and store it with the StoreSecret function provided
```
agnivade, line 58: I am wondering if we can use the consumer interface approach here? Using methods is the right approach when there is shared state involved. And the store certainly seems to be a shared state. The way we are using the package right now is incorrect. We are dynamically creating the struct and calling its methods, whereas we should just create the struct once and store it in `*Server` and call its methods.

Getting rid of the config service seems okay because we only use the siteURL. But for the store, I'd suggest to use a consumer interface so that consumers of the package can pass a proper store interface.

Here is what I would suggest:

```
type Store interface{
// add stuff here
}

type MFA struct{
  store Store
}

func New(store Store) *MFA {
}

func (m *MFA) GenerateSecret(siteURL, userEmail, userID string) (string, []byte, error) {}
func (m *MFA) Activate(userMFASecret, userID string, token string) error {}
func (m *MFA) Deactivate(userID string) error {}
func (m *MFA) ValidateToken(secret, token string) (bool, error) {}
```

jespino: That was my first approach, but I liked the idea of having just methods, and remove the need of the instantiation of the library itself. Specially because the dependency with the config, but now, if we only depend on the store, makes a lot of sense. Let me rewrite that.

services/mfa/mfa.go
jespino: makes sense, lets do it synchronous.

services/mfa/mfa.go
```diff
@@ -4,40 +4,43 @@
...
 	"github.com/mattermost/rsc/qr"
-
-	"github.com/mattermost/mattermost-server/v5/model"
-	"github.com/mattermost/mattermost-server/v5/services/configservice"
-	"github.com/mattermost/mattermost-server/v5/store"
+	"github.com/pkg/errors"
 )
 
+// InvalidToken indicates the case where the token validation has failed.
+type InvalidToken struct{}
```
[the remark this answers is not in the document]
jespino: sure

services/mfa/mfa.go
jespino: done

services/mfa/mfa.go
jespino: done

services/mfa/mfa.go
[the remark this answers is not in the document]
jespino: done

agnivade (approved):
Looking great :ok_hand:

mgdelacroix (approved):

lindy65 (approved):
Thanks @jespino - tested MFA and enforced MFA (enabling/disabling) and logging in/out with both.  Did not find any issues.

PR #16926, opened 2021-02-12
Migrate channel to waitgroup for access the store in parallel

jespino:
#### Summary
Migrate channel to waitgroup for access the store in parallel.

This is a proof of concept of something that can be end up being a campaign to migrate all the cases that still use channels. We talk about this a long time ago, but never was done.

#### Ticket Link
[MM-32865](https://mattermost.atlassian.net/browse/MM-32865)

#### Release Note
```release-note
NONE
```

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -91,7 +91,7 @@ require (
 	github.com/pierrec/lz4/v4 v4.1.3 // indirect
 	github.com/pkg/errors v0.9.1
 	github.com/prometheus/client_golang v1.9.0
-	github.com/prometheus/client_model v0.2.0 // indirect
+	github.com/prometheus/client_model v0.2.0
 	github.com/prometheus/procfs v0.3.0 // indirect
 	github.com/reflog/dateconstraints v0.2.1
 	github.com/rivo/uniseg v0.2.0 // indirect
--- a/go.mod
+++ b/go.mod
@@ -91,7 +91,7 @@ require (
 	github.com/pierrec/lz4/v4 v4.1.3 // indirect
 	github.com/pkg/errors v0.9.1
 	github.com/prometheus/client_golang v1.9.0
-	github.com/prometheus/client_model v0.2.0
+	github.com/prometheus/client_model v0.2.0 // indirect
 	github.com/prometheus/procfs v0.3.0 // indirect
 	github.com/reflog/dateconstraints v0.2.1
 	github.com/rivo/uniseg v0.2.0 // indirect
@@ -123,6 +123,7 @@ require (
 	golang.org/x/image v0.0.0-20201208152932-35266b937fa6
 	golang.org/x/mod v0.4.1 // indirect
 	golang.org/x/net v0.0.0-20210119194325-5f4716e94777
+	golang.org/x/sync v0.0.0-20201020160332-67f06af15bc9
 	golang.org/x/text v0.3.5
 	golang.org/x/tools v0.1.0
 	google.golang.org/genproto v0.0.0-20210119180700-e258113e47cc // indirect
--- a/store/sqlstore/user_store.go
+++ b/store/sqlstore/user_store.go
@@ -10,12 +10,11 @@ import (
 	"fmt"
 	"sort"
 	"strings"
-	"sync"
 
 	sq "github.com/Masterminds/squirrel"
-	"github.com/hashicorp/go-multierror"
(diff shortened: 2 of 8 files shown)


store/sqlstore/user_store.go
```diff
@@ -1942,35 +1943,36 @@ func (us SqlUserStore) DemoteUserToGuest(userID string) (*model.User, error) {
 }
 
 func (us SqlUserStore) AutocompleteUsersInChannel(teamId, channelId, term string, options *model.UserSearchOptions) (*model.UserAutocompleteInChannel, error) {
-	autocomplete := &model.UserAutocompleteInChannel{}
-	uchan := make(chan store.StoreResult, 1)
+	var wg sync.WaitGroup
+	wg.Add(2)
```
lieut-data, line 1947: nit: can we `wg.Add(1)` right before each `go func() { ... }`? I find this helps with the copy-paste problem of forgetting to increment/decrement this when changing the semantics.

store/sqlstore/user_store.go
```diff
@@ -1942,35 +1943,36 @@ func (us SqlUserStore) DemoteUserToGuest(userID string) (*model.User, error) {
...
+	var usersInChannel []*model.User
+	var usersInChannelErr error
+	var usersNotInChannel []*model.User
+	var usersNotInChannelErr error
```
lieut-data, line 1951: To further simplify, would it make sense to leverage something like https://github.com/hashicorp/go-multierror and accumulate the errors instead, especially if we don't handle them "specially"?

lieut-data (approved):
That definitely seems cleaner, eliminating the need for a `store.StoreResult` altogether. Just a few random thoughts below.

store/sqlstore/user_store.go
jespino: Good point, I'll change that

store/sqlstore/user_store.go
jespino: Nice one, I didn't know go-multierror and I didn't know that we were already using it. So... I added it. I still keep the "multierror.Append" after the `wg.Wait()` and not inside the gorouting because I think `multierror.Append` is not thread safe.

Other option is to use a `Group` (https://pkg.go.dev/github.com/hashicorp/go-multierror#Group) of goroutines from the multierror library.

agnivade:
Manually writing waitgroups like this is hard to get right. There is no way to cancel other tasks which might take longer when one task has already thrown an error. I'd suggest to use the https://pkg.go.dev/golang.org/x/sync/errgroup#pkg-overview package which has better semantics and exposes a nice API to do it correctly.

```go
var usersInChannel, usersNotInChannel Item
g, ctx := errgroup.WithContext(ctx)
g.Go(func() (err error) {
  usersInChannel, err = us.SearchInChannel(channelId, term, options)
  return err
})
g.Go(func() (err error) {
  usersNotInChannel, err = us.SearchNotInChannel(teamId, channelId, term, options)
  return err
})
err := g.Wait()
if err != nil {
  return err
}
```

jespino:
@agnivade I added the error group but without context, because we are no using context anywhere in the store now.

agnivade:
Good for now. But we are starting to use contexts in store methods. Around 5 of them use contexts now, and more will have soon :stuck_out_tongue:

go.mod
```diff
@@ -91,7 +91,7 @@ require (
 	github.com/pierrec/lz4/v4 v4.1.3 // indirect
 	github.com/pkg/errors v0.9.1
 	github.com/prometheus/client_golang v1.9.0
-	github.com/prometheus/client_model v0.2.0
+	github.com/prometheus/client_model v0.2.0 // indirect
```
agnivade, line 94: We can revert this. This is an unfortunate problem of our codebase.

agnivade (approved):
Looks so clean!

go.mod
jespino: sure

store/sqlstore/user_store.go
mgdelacroix: Just FYI, coincidentally we're adding `multierror` to `mmctl` for commands that manipulate multiple entities too

mgdelacroix (approved):
LGTM! :+1:

lindy65 (approved):
Thanks @jespino - looks good to me 👍

PR #16927, opened 2021-02-12
unparam lint

mahmud2011:
#### Summary
* fix "always receives ..." lint err
* add unparam lint check

#### Ref

#### Ticket Link
Fixes https://github.com/mattermost/mattermost-server/issues/16622

#### Release Note
```release-note
None
```


app/file.go
```diff
@@ -708,7 +708,7 @@ func (a *App) UploadFileX(channelId, name string, input io.Reader,
 
 	if *a.Config().FileSettings.DriverName == "" {
 		return nil, t.newAppError("api.file.upload_file.storage.app_error",
-			"", http.StatusNotImplemented)
+			"api.file.upload_file.storage.app_error", http.StatusNotImplemented)
```
agnivade, line 711: This seems to be unintentional. Let's revert this. That parameter is meant to be the error string. In the absence of an error string, an empty field is the correct way.

app/svg_test.go
```diff
@@ -64,6 +64,7 @@ func TestParseInvalidSVGData(t *testing.T) {
 	invalidSVGs := []io.Reader{
 		generateSVGData(width, height, false, false, false), // missing viewBox, width & height
 		generateSVGData(width, 0, false, true, false),       // missing viewBox, malformed width & height
+		generateSVGData(0, height, false, true, false),      // missing viewBox, malformed width & height
```
agnivade, line 67: This seems unrelated?

services/searchengine/bleveengine/bleve_test.go
```diff
@@ -113,7 +114,7 @@ func (s *BleveEngineTestSuite) TestDeleteChannelPosts() {
 		channelToAvoidID := model.NewId()
 		posts := make([]*model.Post, 0)
 		for i := 0; i < 10; i++ {
-			post := createPost(userID, channelID, "test one two three")
+			post := createPost(userID, channelID, "test one two three "+time.Now().String())
```
agnivade, line 117: unrelated?

agnivade (commented):
Great work!

It looks like there are some additional changes which add new test coverage/fix flaky tests? Can we make them as part of a different PR? I would like this to be only about the `unparam` linter.

app/file.go
mahmud2011: Actually you need to provide something in `details` otherwise you will get this error `always receives "" ...`. In every test case `details` is empty string.

mahmud2011:
> Great work!
> 
> It looks like there are some additional changes which add new test coverage/fix flaky tests? Can we make them as part of a different PR? I would like this to be only about the `unparam` linter.

I only made changes to satisfy unparam lint nothing more.

app/svg_test.go
mahmud2011: In every case `width` is same that's why the error `always receives ...`.

services/searchengine/bleveengine/bleve_test.go
mahmud2011: In every case `"test one two three"`

app/file.go
agnivade: Oh I see. Let's remove the `details` parameter then.

services/searchengine/bleveengine/bleve_test.go
agnivade: Let's use `model.NewRandomString` for this.

app/svg_test.go
agnivade: Let's add an ignore statement for this now, and fix this in a separate PR.

mahmud2011:
services/mailservice/mail.go:260:69: `sendMailUsingConfigAdvanced` - `enableComplianceFeatures` is unused (unparam)
func sendMailUsingConfigAdvanced(mail mailData, config *SMTPConfig, enableComplianceFeatures bool) error {

Local golangci-lint does not catch this. -_-

agnivade (approved):
Great work!

agnivade:
This is pending changes in the EE codebase too. I will take care of it.

services/searchengine/bleveengine/bleve_test.go
```diff
@@ -138,7 +138,7 @@ func (s *BleveEngineTestSuite) TestDeleteChannelPosts() {
 		userID := model.NewId()
 		channelID := model.NewId()
 		channelToDeleteID := model.NewId()
-		post := createPost(userID, channelID, "test one two three")
+		post := createPost(userID, channelID, model.NewRandomString(15))
```
streamer45, line 141: I think here (and in the other related cases) the best approach would be to remove the parameter and generate the random message directly inside `createPost`.

streamer45 (approved):
Thank you @mahmud2011 :+1: 
Left one minor suggestion but otherwise looks good!

services/searchengine/bleveengine/bleve_test.go
agnivade: @mahmud2011 - can you please look into this?

agnivade:
Hey @mahmud2011 - Just a small thing needed to be done here: https://github.com/mattermost/mattermost-server/pull/16927#discussion_r589964103.

services/searchengine/bleveengine/bleve_test.go
mahmud2011: @agnivade Sorry, I missed it. Happy to do that.

PR #16928, opened 2021-02-12
remove redundant transaction

mahmud2011:
#### Summary
* remove transaction from `SaveMultipleMembers` which is no longer necessary after merging https://github.com/mattermost/mattermost-server/pull/16771

#### Release Note
```release-note
NONE
```


reflog (approved):

jespino (approved):
Looks good to me. Thanks @mahmud2011

jespino:
@ogi-m from QA perspective, this should be keep working for importing bulk exports. So the memberships have to be properly imported.

agnivade:
I think this can be qa-deferred as if there are any regressions it should get caught as part of normal testing.

lindalumitchell (approved):

PR #16930, opened 2021-02-13
NO-TCIKET fix bad smells - error strings should not be capitalized

cupakob:
#### Summary
fix bad smells - error strings should not be capitalized or end with punctuation

#### Release Note
```release-note
NONE
```


mahmud2011:
There's an assert.
https://github.com/mattermost/mattermost-server/blob/master/web/response_writer_wrapper_test.go#L63

agnivade:
Hi @cupakob ! Thank you for taking the time to send these fixes. I will be reviewing them shortly. Just a quick note that in our review process, we usually refrain from doing force pushes, as it prevents the reviewers from seeing what changed since the last commit. It is conventional to just add a new commit, and when getting changes from master, use merge, rather than rebase. Thanks!

hahmadia (commented):
Hey @agnivade ,

I'm okay with approving this PR as it's minor text change but should we create a ticket for it regardless as part of the SOC 2 requirements?
Also, was there a requirement for making these changes? I wasn't sure since there is no ticket attached.

**Note** Sorry about that , I was mistaken and double checked and had gotten confused.

hahmadia (approved):
Changes look good but I'm 0/5 on if they are required or not.
Will let Agniva make the ultimate decision on that.

web/response_writer_wrapper.go
```diff
@@ -50,7 +50,7 @@ func (rw *responseWriterWrapper) Write(data []byte) (int, error) {
 // it loses the access to the implementation for Hijack or Flush
 func (rw *responseWriterWrapper) Hijack() (net.Conn, *bufio.ReadWriter, error) {
 	if rw.hijacker == nil {
-		return nil, nil, errors.New("Hijacker interface not supported by the wrapped ResponseWriter")
+		return nil, nil, errors.New("hijacker interface not supported by the wrapped ResponseWriter")
```
agnivade, line 53: This is incorrect. The name of the interface is `Hijacker`.

utils/i18n.go
```diff
@@ -80,12 +80,12 @@ func GetTranslationsBySystemLocale() (i18n.TranslateFunc, error) {
 	}
```
agnivade, line 80: There is one more error at L57: `fmt.Errorf(fmt.Sprintf("Unable to find i18n directory at %q", dir))`. Let's change that too, and directly use `fmt.Errorf`.

plugin/supervisor.go
```diff
@@ -123,7 +123,7 @@ func (sup *supervisor) PerformHealthCheck() error {
 		}
 		if pingErr != nil {
 			mlog.Debug("Error pinging plugin", mlog.Err(pingErr))
-			return fmt.Errorf("Plugin RPC connection is not responding")
+			return fmt.Errorf("plugin RPC connection is not responding")
```
agnivade, line 126: Let's also return the internal error.

@larkox  - Are there any concerns if we remove the `mlog.Debug` statement? We are logging the same thing twice.

plugin/client_rpc.go
```diff
@@ -436,7 +436,7 @@ func (s *apiRPCServer) PluginHTTP(args *Z_PluginHTTPArgs, returns *Z_PluginHTTPR
 		returns.Response = response
 		returns.ResponseBody = responseBody
 	} else {
```
agnivade, line 438: Please also run `make pluginapi` to ensure the generated code hasn't changed.

cmd/mattermost/commands/channel.go
```diff
@@ -562,7 +562,7 @@ func modifyChannelCmdF(command *cobra.Command, args []string) error {
 
 	user := getUserFromUserArg(a, username)
 	if user == nil {
-		return fmt.Errorf("Unable to find user: '%v'", username)
+		return fmt.Errorf("unable to find user: '%v'", username)
```
agnivade, line 565: The cmd/mattermost packages are a bit special. The error returned is directly displayed in the console. So we have to keep the error strings as-is because there is nothing prefixing them with a proper user facing string.

Please revert all changes in cmd/mattermost.

agnivade (changes requested):
Thanks @cupakob. There are some small changes that need to be done.

plugin/supervisor.go
larkox: My guts tell me that it can be removed, but I don't have the knowledge on this part of the Toolkit enough to make the decision. Maybe @lieut-data knows better or can lead to the person that can answer this?

lieut-data: I can say that I've never relied on the debug log in the entirety of its existence. If we surface `pingErr` via the return value instead and it gets logged, I don't anticipate any future problems.

agnivade: Brilliant, thank you all. @cupakob - let's also remove the `mlog.Debug` statement above this.

web/response_writer_wrapper.go
cupakob: fixed

utils/i18n.go
cupakob: fixed

plugin/supervisor.go
cupakob: fixed

cmd/mattermost/commands/channel.go
cupakob: fixed

plugin/client_rpc.go
cupakob: check - no changes after `make pluginapi`

agnivade (approved):
Thanks @cupakob !

PR #16934, opened 2021-02-14
NO-TICKET rename variables with reserved word as identifier

cupakob:
#### Summary
rename variables with reserved word as identifier

#### Release Note
```release-note
NONE
```


hahmadia:
@agnivade 
The code looks good but I was curious where the requirement for this renaming came up since there is no ticket for it. 
Are we okay to merge PR's without any tickets? I think as part of the SOC 2 requirement, all PR's need to be associated to a ticket.
I'm just curious if you feel this change is required or if we can keep the code as it is?

hahmadia (commented):
@agnivade
The code looks good but I was curious where the requirement for this renaming came up since there is no ticket for it.
Are we okay to merge PR's without any tickets? I think as part of the SOC 2 requirement, all PR's need to be associated to a ticket.
I'm just curious if you feel this change is required or if we can keep the code as it is?

agnivade:
I haven't gone through the PR yet, but I am not aware of any requirement that all PRs need to have a ticket. I have been sending PRs myself without tickets for small things. Was there any announcement that I missed?

AFAIK, only the QA team needs tickets associated with PRs so that they can track them during release.

hahmadia (approved):
I'm 0/5 on these changes. Look good but will let Agniva make the ultimate decision.

agnivade (changes requested):
Thanks @cupakob for the PR. The err/appErr change looks good to me. But for the others, unfortunately I think the cost outweighs the benefits. It is the job of the editor to correctly distinguish between variable names and reserved words. If this was forbidden, it should have been done from the spec itself. But since it's not, I don't see a huge benefit in renaming them now.

I am happy to merge this if we can reduce the changes only to the err/appErr ones.

Thank you again for your time and sorry about this. If you think something might be controversial, it's always a good thing to open an issue to discuss it before spending too much time on it.

cupakob:
> I am happy to merge this if we can reduce the changes only to the err/appErr ones.

i will close it ;)

PR #16938, opened 2021-02-15
Create basic make commands for configuring golang-migrate

nronas:
#### Summary
Implements: 
- https://mattermost.atlassian.net/browse/MM-32869
- https://mattermost.atlassian.net/browse/MM-32870
- https://mattermost.atlassian.net/browse/MM-32871

Adds two make commands `make migrations` that bundle migrations files to the binary and `make new-migrations name=xxx` that creates a new migration for MySQL and postgresql.

For better context, have a look at the deliverables of the schema migrations here: https://mattermost.atlassian.net/browse/MM-32868

#### Ticket Link
- https://mattermost.atlassian.net/browse/MM-32869
- https://mattermost.atlassian.net/browse/MM-32870
- https://mattermost.atlassian.net/browse/MM-32871

#### Release Note

```release-note
* Adds Schema Migrations Phase 0 (Teams, TeamMembers)
```


store/sqlstore/store.go
```diff
@@ -1245,6 +1265,50 @@ func (ss *SqlStore) UpdateLicense(license *model.License) {
...
+		driver, err = postgres.WithInstance(conn, &postgres.Config{})
+		if err != nil {
+			return err
+		}
+	}
+
+	var assetNamesForDriver []string
+	for _, assetName := range migrations.AssetNames() {
+		if strings.HasPrefix(assetName, ss.DriverName()) {
+			assetNamesForDriver = append(assetNamesForDriver, strings.TrimPrefix(assetName, ss.DriverName()+"/"))
```
nronas, line 1287: We need to do this weird prefix trimming and then prefix add, because golang migrate has a regex that does not allow paths on go-bindata files

Makefile
```diff
@@ -250,6 +250,24 @@ telemetry-mocks: ## Creates mock files.
 store-layers: ## Generate layers for the store
 	$(GO) generate $(GOFLAGS) ./store
 
+migration-prereqs: ## Builds prerequisite packages for migrations
+ifeq ($(PLATFORM),Windows)
+	@powershell ./exe/migration_prereqs.ps1
+else
+	@./scripts/migration-prereqs
+endif
```
agnivade, line 258: Let's use the standard `go get` method to install tools and pin the dependencies in go.tools.mod. It avoids different tool versions and any custom shell/powershell scripts to install these tools.

The only exception was `golangci-lint` where the README explicitly discouraged from using go-get to install the tool.

go.mod
```diff
@@ -93,7 +96,7 @@ require (
 	github.com/pierrec/lz4/v4 v4.1.3 // indirect
 	github.com/pkg/errors v0.9.1
 	github.com/prometheus/client_golang v1.9.0
-	github.com/prometheus/client_model v0.2.0
+	github.com/prometheus/client_model v0.2.0 // indirect
```
agnivade, line 99: Please revert this. This is an unfortunate issue with Go.

store/sqlstore/store.go
```diff
@@ -15,10 +16,20 @@ import (
 	"sync/atomic"
 	"time"
 
+	"github.com/mattermost/mattermost-server/v5/db/migrations"
+
+	bindata "github.com/golang-migrate/migrate/v4/source/go_bindata"
+
```
agnivade, line 22: Nit: imports are jumbled up

store/sqlstore/store.go
agnivade: Can you point me to which regex are we talking about? Just for my understanding. When doing this for mattermod, I did not notice any such limitation.

store/sqlstore/store.go
```diff
@@ -152,6 +163,15 @@ func New(settings model.SqlSettings, metrics einterfaces.MetricsInterface) *SqlS
 
 	store.initConnection()
 
+	for _, conn := range store.GetAllConns() {
```
agnivade, line 166: Won't doing it only for master be sufficient?

store/sqlstore/store.go
```diff
@@ -1245,6 +1265,50 @@ func (ss *SqlStore) UpdateLicense(license *model.License) {
...
+	migrations, err := migrate.NewWithInstance("go-bindata",
+		sourceDriver,
+		ss.DriverName(),
+		driver)
+
+	if err != nil {
+		return err
+	}
+
+	return migrations.Up()
```
agnivade, line 1309: We need to ignore certain errors: https://github.com/mattermost/mattermost-mattermod/blob/c6bdaa816030f8d3c361e80b7578fce8d0208eba/store/sql_store.go#L123-L129

agnivade (commented):
My first question would be: What is the minimum version of Mattermost that we have decided, on top of which these schema migrations would be applied?

store/sqlstore/store.go
nronas: I think so it will be, i just didn't want to change too many things here. We currently are applying migrations for all connections, so I followed the same.

store/sqlstore/store.go
nronas: will amend

Makefile
nronas: oh good point, on here https://github.com/golang-migrate/migrate/tree/master/cmd/migrate i didn;t notice the toolchain support. Will amend.

store/sqlstore/store.go
nronas: I am using folder namespacing in order to separate files for MySQL and Postgresql. So the bindata files are looking something like: `mysql/123_name.up.sql` and `postgres/123_name.up.sql` the golang-migrate has a parses that expects just a file name https://github.com/golang-migrate/migrate/blob/master/source/parse.go#L21 without paths.

agnivade: Ohh got you now, so it's trimming the folder prefix. I thought it was something else. Wondering if we can use [filepath.Base](https://golang.org/pkg/path/filepath/#Base) for this? That seems like a more robust alternative.

store/sqlstore/store.go
agnivade: Fair enough. Perhaps work to be done in a separate ticket?

nronas: Amended on this PR, since we actually need a separate connection for migrations altogether.

store/sqlstore/store.go
nronas: Will try and see if it works.

store/sqlstore/store.go
```diff
@@ -146,8 +154,30 @@ func New(settings model.SqlSettings, metrics einterfaces.MetricsInterface) *SqlS
...
+	// in order to make migrations work.
+	if store.DriverName() == model.DATABASE_DRIVER_MYSQL {
+		u, err := url.Parse(*settings.DataSource)
+		if err != nil {
+			mlog.Critical("Invalid database url found", mlog.Err(err))
+			time.Sleep(time.Second)
+			os.Exit(ExitGenericFailure)
+		}
+		q := u.Query()
+		q.Set("multiStatements", "true")
```
agnivade, line 167: Just checking, IIUC, the `multiStatements` option will be applied for the entire runtime of the application right? Because the migrations use the same connection as the rest of the app, and it won't be _only_ used as a CLI app.

Since we are already calling `setupConnection` inside the migrate method, I am wondering if it's possible to create a separate connection with `multiStatements` to true and then run migrations for it for MySQL?

If we don't want to go to such complexity, I'd just suggest to let the option be added from the DSN itself. So this will need to be part of the release notes and something the customer has to do. IMO, it's better than us changing it under the hood, which might be surprising and not something the customer would expect.

store/sqlstore/store.go
```diff
@@ -1187,6 +1217,61 @@ func (ss *SqlStore) UpdateLicense(license *model.License) {
...
+
+	var assetNamesForDriver []string
+	for _, assetName := range migrations.AssetNames() {
+		if strings.HasPrefix(assetName, ss.DriverName()) {
+			assetNamesForDriver = append(assetNamesForDriver, filepath.Base(assetName))
+		}
+	}
+
+	source := bindata.Resource(assetNamesForDriver, func(name string) ([]byte, error) {
+		return migrations.Asset(ss.DriverName() + "/" + name)
```
agnivade, line 1249: `filepath.Join`

store/sqlstore/store.go
```diff
@@ -1187,6 +1217,61 @@ func (ss *SqlStore) UpdateLicense(license *model.License) {
...
+		sourceDriver,
+		ss.DriverName(),
+		driver)
+
+	if err != nil {
+		return err
+	}
+
+	err = migrations.Up()
+	migrations.Close()
```
agnivade, line 1267: `defer migrations.Close()` above this?

store/sqlstore/upgrade.go
```diff
@@ -551,12 +550,8 @@ func upgradeDatabaseToVersion50(sqlStore *SqlStore) {
 	//    DELETE from Systems WHERE Name = 'migration_advanced_permissions_phase_2';
```
agnivade, line 551: I can see a line `sqlStore.CreateColumnIfNotExists("TeamMembers", "DeleteAt", "bigint(20)", "bigint", "0")` in `upgradeDatabaseToVersion32`.

But in our case, the `DeleteAt` column is part of the create table command. Wouldn't this fail for installations upgrading from 3.1 to 5.34 for example?

My understanding was that we would run _all_ migrations for Teams/TeamMembers from version 3.0. But it looks like some are missing.

agnivade (commented):
Great work!

I just have two high level questions:
- How should a developer add new migrations from this point? Should they use the new model?
- What is the complete plan going forward? Are we going to take each table and add _all_ of its migrations from start to end, and do it in each PR? Or some other way?

Makefile
```diff
@@ -250,6 +250,22 @@ telemetry-mocks: ## Creates mock files.
 store-layers: ## Generate layers for the store
 	$(GO) generate $(GOFLAGS) ./store
 
+migration-prereqs: ## Builds prerequisite packages for migrations
+	$(GO) get -tags 'postgres' -modfile=go.tools.mod github.com/golang-migrate/migrate/v4/cmd/migrate
```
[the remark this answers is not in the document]
nronas: No reason at all, i just copied it from the docs, haven't looked deeper tbh. Will amend to omit the tag

Makefile
```diff
@@ -250,6 +250,22 @@ telemetry-mocks: ## Creates mock files.
...
+new-migration: migration-prereqs ## Creates a new migration
+	@echo "Generating new migration for mysql"
+	$(GOBIN)/migrate create -ext sql -dir db/migrations/mysql -seq $(name)
+
+	@echo "Generating new migration for postgres"
+	$(GOBIN)/migrate create -ext sql -dir db/migrations/postgres -seq $(name)
+
+	@echo "When you are done writing your migration, run 'make migrations'"
+
+migrations: ## Generates bindata migrations
```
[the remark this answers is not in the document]
nronas: Same here, will amend

store/sqlstore/upgrade.go
nronas: Totally, i've missed that! 🦅 👀

db/migrations/postgres/000001_create_teams.down.sql
```diff
@@ -0,0 +1,12 @@
+ALTER TABLE teams
+DROP COLUMN IF EXISTS allowopeninvite;
+DROP COLUMN IF EXISTS lastteamiconupdate;
```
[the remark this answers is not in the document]
nronas: forgot to change the down files, for pg we have to have each alter command to have 1 statement; Will amend

store/sqlstore/store.go
nronas: Since this is a migrations-only flag and also is a client-owned flag, I prefer to move that into the migrations connection and then tear it down when we are done with executing the migrations. Will amend.

store/sqlstore/store.go
nronas: yeap, silly me

nronas:
> I just have two high level questions:
> - How should a developer add new migrations from this point? Should they use the new model?

@mgdelacroix to correct me if my understanding here is wrong, but from initial discussions that we had, developers at this stage of the migrations project should not see/feel that anything is different. Since we are pretty early stages we do not want to flood the gates open and require any new migrations to come through the new way of doing things. We will start migrating each table over and then any new migrations that have added we will do them too(the premise is that, we acknowledge that this is a moving target but the rate of newly added migrations is << than the rate that we will migrate things over to the new framework).

> - What is the complete plan going forward? Are we going to take each table and add all of its migrations from start to end, and do it in each PR? Or some other way?

I think I semi-answered this, but just for clarity, yes we will try to do 1 PR for each table, keeping things small and fast to review. Also, another reason for this high-granularity is to avoid any issues that come when people reviewing raw SQL(fewer SQL lines per PR the better 😄). For detailed info see the EPIC link https://mattermost.atlassian.net/browse/MM-31622

@agnivade happy to hear if you have any thoughts on ☝️

agnivade:
Makes sense to me. Thanks @nronas for the clarification!

mgdelacroix (approved):
Amazing work!! :rocket:

store/sqlstore/store.go
```diff
@@ -1245,6 +1274,59 @@ func (ss *SqlStore) UpdateLicense(license *model.License) {
 	ss.license = license
 }
 
+func (ss *SqlStore) migrate() error {
+	var driver database.Driver
+	var err error
+
+	// W hen WithInstance is used in golang-migrate, the underlying driver connections are not tracked.
```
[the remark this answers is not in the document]
nronas: Already fixed on a future commit ;)

store/sqlstore/store.go
```diff
@@ -44,6 +52,9 @@ const (
 	// 10.1 would be 100001.
 	// 9.6.3 would be 90603.
 	MinimumRequiredPostgresVersion = 100000
+
+	migrationsDirectionUp   = "up"
+	migrationsDirectionDown = "down"
```
agnivade, line 57: We can even make these to be typed constants for extra type safety, and use `migrationDirection` as the type for migrate method. 0/5.

```
type migrationDirection string

const (
 migrationsDirectionUp migrationDirection  = "up"
 migrationsDirectionDown migrationDirection = "down"
)
```

store/sqlstore/store.go
```diff
@@ -1187,6 +1205,88 @@ func (ss *SqlStore) UpdateLicense(license *model.License) {
...
+	}
+	defer migrations.Close()
+
+	if direction == migrationsDirectionUp {
+		err = migrations.Up()
+	} else if direction == migrationsDirectionDown {
+		err = migrations.Down()
+	} else {
+		return errors.New(fmt.Sprintf("unsupported migration direction %s", direction))
+	}
```
agnivade, line 1262: 0/5, might be better with a switch statement.

store/sqlstore/store.go
```diff
@@ -1187,6 +1205,88 @@ func (ss *SqlStore) UpdateLicense(license *model.License) {
...
+}
+
+func (ss *SqlStore) appendMultipleStatementsFlag(dataSource string) string {
+	// We need to tell the MySQL driver that we want to use multiStatements
+	// in order to make migrations work.
+	if ss.DriverName() == model.DATABASE_DRIVER_MYSQL {
+		u, err := url.Parse(dataSource)
+		if err != nil {
+			mlog.Critical("Invalid database url found", mlog.Err(err))
+			time.Sleep(time.Second)
```
agnivade, line 1278: Do we really need to sleep here?

db/migrations/postgres/000001_create_teams.up.sql
```diff
@@ -0,0 +1,27 @@
+CREATE TABLE IF NOT EXISTS teams (
```
agnivade, line 1: There is a groupconstrained column which seems to be missing: 

`sqlStore.CreateColumnIfNotExistsNoDefault("Teams", "GroupConstrained", "tinyint(4)", "boolean")`

agnivade (commented):
Just some finishing touches to be applied. Looking great now.

.circleci/config.yml
```diff
@@ -140,6 +140,18 @@ jobs:
             cd mattermost-server
             make einterfaces-mocks store-mocks
             if [[ -n $(git status --porcelain) ]]; then echo "Please update the mocks using `make einterfaces-mocks store-mocks`"; exit 1; fi
+  check-migrations:
+    docker:
+      - image: mattermost/mattermost-build-server:20200731_golang-1.14.6
```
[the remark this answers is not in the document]
nronas: oh true, just copied it from above. Didn't see the version number. Thanks

db/migrations/postgres/000001_create_teams.up.sql
```diff
@@ -0,0 +1,27 @@
...
+    id VARCHAR(26) PRIMARY KEY,
+    createat bigint,
+    updateat bigint,
+    deleteat bigint,
+    displayname VARCHAR(64),
+    name VARCHAR(64),
+    description VARCHAR(255),
```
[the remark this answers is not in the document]
nronas: It is needed, since it is part of the initial table create from Gorp, and then the upgrade will basically be noop.

db/migrations/mysql/000001_create_teams.up.sql
```diff
@@ -0,0 +1,68 @@
+CREATE TABLE IF NOT EXISTS Teams (
+    Id varchar(26) NOT NULL,
+    CreateAt bigint(20) DEFAULT NULL,
+    UpdateAt bigint(20) DEFAULT NULL,
+    DeleteAt bigint(20) DEFAULT NULL,
+    DisplayName varchar(64) DEFAULT NULL,
+    Name varchar(64) DEFAULT NULL,
+    Description varchar(255) DEFAULT NULL,
```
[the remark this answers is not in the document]
nronas: See reply above. 😄

store/sqlstore/store.go
nronas: sure, 0/5 as well.

store/sqlstore/store.go
nronas: nope, 💤

.circleci/config.yml
```diff
@@ -140,6 +140,18 @@ jobs:
             cd mattermost-server
             make einterfaces-mocks store-mocks
             if [[ -n $(git status --porcelain) ]]; then echo "Please update the mocks using `make einterfaces-mocks store-mocks`"; exit 1; fi
+  check-migrations:
```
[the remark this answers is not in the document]
nronas: 🤦

.circleci/config.yml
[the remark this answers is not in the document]
agnivade: Hmm .. doesn't seem to be updated? :thinking:

store/sqlstore/store.go
agnivade: Still seems to be here.

agnivade (commented):
@metanerd - Heads up that we are adding a new stage in the CI. We'll need to copy it over to Gitlab as well once it hits master.

store/sqlstore/store.go
nronas: oh shoot, i removed it on one place and not on the other.

.circleci/config.yml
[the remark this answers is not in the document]
nronas: i completely miss-understood your comment, i thought you were going to update that in a separate PR.

db/migrations/postgres/000001_create_teams.down.sql
```diff
@@ -0,0 +1,13 @@
...
+ALTER TABLE teams DROP COLUMN IF EXISTS groupconstrained;
+
+DROP INDEX IF EXISTS idx_teams_name;
+DROP INDEX IF EXISTS idx_teams_invite_id;
+DROP INDEX IF EXISTS idx_teams_update_at;
+DROP INDEX IF EXISTS idx_teams_create_at;
+DROP INDEX IF EXISTS idx_teams_delete_at;
+DROP INDEX IF EXISTS idx_teams_scheme_id;
+
+DROP TABLE IF EXISTS Teams;
```
mgdelacroix, line 13:
```suggestion
DROP TABLE IF EXISTS teams;
```

.circleci/config.yml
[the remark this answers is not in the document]
agnivade: Ohhh so sorry there @nronas ! My brain completely forgot that I said I'll send the PR separately. So sorry for the confusion there. And thanks for making the change in this PR. :)

agnivade (approved):
Nicely done! This is a great first step. Thanks @nronas for your patience with the review.

PR #16943, opened 2021-02-16
MM-32893: Refactor the getting of DB handle

agnivade:
We add a method in *SqlStore that returns the correct DB handle
depending on the context.

https://mattermost.atlassian.net/browse/MM-32893

```release-note
NONE
```

### DIFF
--- a/store/sqlstore/context.go
+++ b/store/sqlstore/context.go
@@ -36,9 +36,9 @@ func hasMaster(ctx context.Context) bool {
 }
 
 // DBFromContext is a helper utility that returns the DB handle from a given context.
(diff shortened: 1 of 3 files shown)


isacikgoz (approved):
Cool stuff 👍

store/sqlstore/context.go
```diff
@@ -5,6 +5,8 @@ package sqlstore
 
 import (
 	"context"
+
+	"github.com/mattermost/gorp"
```
streamer45, line 9: We should be removing these not adding them :stuck_out_tongue_closed_eyes:

streamer45 (approved):
Nicely done :+1:

PR #16946, opened 2021-02-16
Extracting html templates into a library 

jespino:
#### Summary
Extracting html templates into a library 

#### Ticket Link
[MM-33421](https://mattermost.atlassian.net/browse/MM-33421)

#### Related PR
enterprise: mattermost/enterprise#929

#### Release Note
```release-note
NONE
```


jespino:
@agnivade @mgdelacroix this is more a draft than a PR, I would love to have your point of view here, I changed the API because I think now is cleaner and clearer from the usage perspective.

nronas:
👋 @jespino what is the actual use case that requires those templates to be shared?

shared/mtemplates/templates.go
```diff
@@ -0,0 +1,116 @@
...
+	"io"
+	"path/filepath"
+	"sync/atomic"
+
+	"github.com/fsnotify/fsnotify"
+
+	"github.com/mattermost/mattermost-server/v5/mlog"
+)
+
+type Templates struct {
```
nronas, line 18: Is this a generic template or it is an HTML template? Since from reading the struct name i assume that is a generic template, but reading the test cases i see that we are testing an HTML template. Maybe worth changing the struct name to HTMLTemplate, also worth dropping the plural i guess.

jespino:
@nronas we are not sharing the templates itself, we are sharing the mechanism to load/watch and render templates from mattermost. This is going to be used as part of the rendering of email. (And as in mattermost can be used to generate the html to static pages like the `unsupported_browser` page).

shared/mtemplates/templates.go
jespino: Actually, this is the "Templates" bucket, where you initialize the templates, and then, using that templates you render things. So the plurar makes sense. I think adding the `HTML` prefix can be more explicit.

nronas:
> @nronas we are not sharing the templates itself

Yeap that's what I meant. Thanks for describing the use case!

shared/mtemplates/templates.go
```diff
@@ -0,0 +1,116 @@
...
+	}
+}
+
+func (t *Templates) Render(templateName string, data *Data) string {
+	var text bytes.Buffer
+	t.RenderToWriter(&text, templateName, data)
+	return text.String()
+}
+
+func (t *Templates) RenderToWriter(w io.Writer, templateName string, data *Data) error {
```
agnivade, line 109: I would suggest the naming to be inverted. Since we are passing the writer, it's understood that it will be written to the writer. It's better to rename the other one as `RenderToString` instead.

agnivade (commented):
I gave a lot of comments in various places. Here's the full API that I am thinking, so that it can be viewed at a single glance:

```
type Container struct {
  Err chan <-error
  // private fields
}

func NewFromTemplate(t *template.Template) (*Container, error)
func New(dir string, watch bool) (*Container, error)

(c *Container) Templates() *template.Template
(c *Container) Close()
(c *Container) Render(w io.Writer, templateName string, data Data) (string, error)
(c *Container) RenderToString(templateName string, data Data) (string, error)
```

shared/mtemplates/templates.go
nronas: +1 on ☝️

shared/mtemplates/templates.go
jespino: Done

shared/mtemplates/templates.go
```diff
@@ -0,0 +1,116 @@
...
+}
+
+func (t *Templates) Close() {
+	if t.watch {
+		close(t.stop)
+		<-t.stopped
+	}
+}
+
+func (t *Templates) Render(templateName string, data *Data) string {
```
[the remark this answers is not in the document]
jespino: Done

shared/mtemplates/templates.go
```diff
@@ -0,0 +1,116 @@
...
+)
+
+type Templates struct {
+	templates atomic.Value
+	stop      chan struct{}
+	stopped   chan struct{}
+	watch     bool
+}
+
+type Data struct {
```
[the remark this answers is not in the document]
jespino: Added godoc there, and everywhere else.

shared/mtemplates/templates.go
```diff
@@ -0,0 +1,116 @@
...
+	"path/filepath"
+	"sync/atomic"
+
+	"github.com/fsnotify/fsnotify"
+
+	"github.com/mattermost/mattermost-server/v5/mlog"
+)
+
+type Templates struct {
+	templates atomic.Value
```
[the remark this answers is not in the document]
jespino: Done

shared/mtemplates/templates.go
```diff
@@ -0,0 +1,116 @@
+// Copyright (c) 2015-present Mattermost, Inc. All Rights Reserved.
+// See LICENSE.txt for license information.
+
+package mtemplates
```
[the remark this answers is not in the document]
jespino: Done

shared/mtemplates/templates.go
```diff
@@ -0,0 +1,116 @@
...
+import (
+	"bytes"
+	"html/template"
+	"io"
+	"path/filepath"
+	"sync/atomic"
+
+	"github.com/fsnotify/fsnotify"
+
+	"github.com/mattermost/mattermost-server/v5/mlog"
```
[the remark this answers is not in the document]
jespino: I haven't done this change, because I'm not 100% sure about the best way of doing it. I'll reach you out in https://community.mattermost.com to talk about it.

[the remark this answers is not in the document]
jespino: Now is ready

shared/mtemplates/templates.go
jespino: Done

shared/templates/templates.go
```diff
@@ -0,0 +1,152 @@
...
+
+func (t *Container) getTemplates() *template.Template {
+	t.mutex.RLock()
+	defer t.mutex.RUnlock()
+	return t.templates
+}
+
+// Close stops the templates watcher of the container in case you have created
+// it with watch parameter set to true
+func (t *Container) Close() {
```
agnivade, line 129: Are we actually calling this somewhere? I am not seeing it :thinking:

shared/templates/templates.go
```diff
@@ -0,0 +1,152 @@
...
+func (t *Container) RenderToString(templateName string, data Data) string {
+	var text bytes.Buffer
+	t.Render(&text, templateName, data)
+	return text.String()
+}
+
+// RenderToString renders the template referenced with the template name using
+// the data provided and write it to the writer provided
+func (t *Container) Render(w io.Writer, templateName string, data Data) error {
+	if err := t.getTemplates().ExecuteTemplate(w, templateName, data); err != nil {
```
agnivade, line 147: It's good to check if `t.getTemplates` return nil or not, and return an error accordingly. There can be a case where the initialization failed, and the consumer is yet to handle the error, but has called `t.Render` before that.

shared/templates/templates.go
```diff
@@ -0,0 +1,152 @@
...
+	}()
+
+	return ret, errors
+}
+
+func (t *Container) getTemplates() *template.Template {
+	t.mutex.RLock()
+	defer t.mutex.RUnlock()
+	return t.templates
+}
```
agnivade, line 125: Is there a need for this method? It's only used once. Maybe we can move it inside `Render` method

app/server.go
```diff
@@ -383,10 +385,19 @@ func NewServer(options ...Option) (*Server, error) {
...
-		s.htmlTemplateWatcher = htmlTemplateWatcher
+		if htmlTemplateWatcher, errorsChan := templates.NewWithWatcher(templatesDir); errorsChan != nil {
+			s.Go(func() {
+				for err2 := range errorsChan {
+					mlog.Error("Server templates error", mlog.Err(err2))
+				}
+			})
+		} else {
+			s.htmlTemplateWatcher = htmlTemplateWatcher
+		}
```
agnivade, line 400: I am not entirely sure if this is correct. In the normal case, `errorsChan` will not be nil. So it looks like we won't be assigning the `s.htmlTemplateWatcher` in that case.

I would say that there is no need to check for nil channel. The API _guarantees_ that the channel will always be non-nil. We can just range over it. The idea would be that whenever an error occurs, we close the watcher and set s.htmlTemplateWatcher to `nil` so that successive calls will fail.
```
var errChan chan error
s.htmlTemplateWatcher, errChan = templates.NewWithWatcher(dir)
s.Go(func() {
    err2 := <- range errorsChan 
    mlog.Error("Server templates error", mlog.Err(err2))
   s.htmlTemplateWatcher.Close()
  //some mutex
s.htmlTemplateWatcher = nil
 // end mutex
})
```

app/app.go
```diff
@@ -150,9 +150,9 @@ func (a *App) TelemetryId() string {
...
 }
 
-func (s *Server) HTMLTemplates() *template.Template {
+func (s *Server) TemplatesContainer() *templates.Container {
 	if s.htmlTemplateWatcher != nil {
-		return s.htmlTemplateWatcher.Templates()
+		return s.htmlTemplateWatcher
 	}
 
 	return nil
```
agnivade, line 158: This can return nil if the watcher is not set, or closed. In that case, calls using this will panic. We need to handle the nil case.

shared/templates/templates.go
jespino: Yes, during the server Shutdown (was already there, that is the reason it is not as part of the changes in this PR).

shared/templates/templates.go
jespino: I'll do the check, it should never return nil unless you create ti using `templates.Container{}` . Using the existing defined constructors, that is always true.

shared/templates/templates.go
jespino: Good point

shared/templates/templates.go
agnivade: If you use `NewWithWatcher` and `fsnotify.NewWatcher()` fails, then it will be nil. That's what I was trying to say in my original comment.

app/server.go
jespino: :facepalm:

jespino:
@agnivade PTAL, I'm changing a bit the approach, but I don't feel really confident about the channels part.

agnivade (commented):
Coming up to be a great shape now. I thought a bit more on how to simplify and I think we are nearly there. Take a look and let me know what you think.

shared/templates/templates.go
```diff
@@ -0,0 +1,155 @@
...
+// NewWithWatcher creates a new templates container scanning a directory and
+// watch the directory filesystem changes to apply them to the loaded
+// templates. The caller must consume the returned error channel to ensure not
+// blocking the watch process.
+func NewWithWatcher(directory string) (*Container, <-chan error) {
+	errors := make(chan error)
+	ret := &Container{}
+
+	htmlTemplates, err := template.ParseGlob(filepath.Join(directory, "*.html"))
+	if err != nil {
```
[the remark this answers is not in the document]
jespino: I though about that, but I think is not "fair" to not return the root cause of the error. Maybe we can return `(*Container, error, chan error)`, that way, the `error` is the synchronous error about initializing the templates with a watcher, and the `chan error` is the errors produced asynchronously during the watch.

About the caller, we weren't stopping entirely the server start because the templates failure, so... My expectation would be to allow you to keep going without the htmlTemplateWatcher even if that fails to start. Also, we can use directly the `templates.NewFromTemplate(template.New(""))` on the initialization as part of the error handling, so we always have a valid `htmlTemplateWatcher` object.

Probably if I'm not able to start the `NewWithWatcher` I would log that, and try to create with a regular `New`, and finally, if that also fails, create the `templates.NewFromTemplate(template.New(""))`.

shared/templates/templates.go
```diff
@@ -0,0 +1,155 @@
...
+		close(t.stop)
+		<-t.stopped
+	}
+}
+
+// RenderToString renders the template referenced with the template name using
+// the data provided and return a string with the result
+func (t *Container) RenderToString(templateName string, data Data) string {
+	var text bytes.Buffer
+	t.Render(&text, templateName, data)
```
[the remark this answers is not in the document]
jespino: That would make the API uncomfortable and wasn't like that in the past.

Other different approaches are create a new function like `RenderToStringOrNothing` that has that exact same behavior, let `RenderToString` return `(string, error)`.

Other option is a "fail fast" approach with something like `MustRenderToString`, that panic on rendering failure. Not the sexiest option, but would help us to discover problems before they really happen.

Anyway, I think we can defer that until this is merged.

shared/templates/templates.go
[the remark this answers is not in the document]
agnivade: > Maybe we can return (*Container, error, chan error), that way, the error is the synchronous error about initializing the templates with a watcher, and the chan error is the errors produced asynchronously during the watch.

Yeah .. I had thought about that. It feels a bit weird API wise to return error and chan error, as there's no clear distinction. I wanted to give a last shot at the "error chan being a struct field" idea. We would only initialize the channel in the watcher case, and send errors only from the watcher loop. And the constructor definition can remain `*Container, error`. What do you think about that?

> About the caller, we weren't stopping entirely the server start because the templates failure, 

Would there be a reason not to fail? One of our core principles is to "fail fast". I think from a robustness perspective, it's good to always start with a correct and stable state. It also removes the need to add all of those extra logic you mentioned just to have a non-nil templates object. I wanted to take this opportunity and improve some of the areas where we should be more robust, but we aren't.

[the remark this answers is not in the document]
jespino: That hidden field that you have to handled... I think is more hard to follow and understand, I think the error + errorChannel is more explicit about the need of handling both.

I'm ok about "failing fast" here.

shared/templates/templates.go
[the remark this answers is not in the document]
agnivade: Yes, the API wasn't like that. And I think we should fix it. Ignoring errors from inside the method is not good.

I don't think there is a need to add `RenderToStringOrNothing` or `MustRenderToString` just to have an API that returns no error.

The server can choose to ignore the error if it wants (although I think that it shouldn't). But we should always make the API return an error.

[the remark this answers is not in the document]
jespino: Fair enough, I'll do the changes, but this is going to be a big commit.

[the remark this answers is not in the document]
jespino: Done

shared/templates/templates.go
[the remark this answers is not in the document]
agnivade: Sounds good then, let's go with it and fail fast.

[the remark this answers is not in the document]
jespino: Done

app/app.go
```diff
@@ -150,12 +151,12 @@ func (a *App) TelemetryId() string {
...
 
-func (s *Server) HTMLTemplates() *template.Template {
+func (s *Server) TemplatesContainer() *templates.Container {
 	if s.htmlTemplateWatcher != nil {
-		return s.htmlTemplateWatcher.Templates()
+		return s.htmlTemplateWatcher
 	}
 
-	return nil
+	return templates.NewFromTemplate(template.New(""))
```
agnivade, line 159: Wondering if we still need this? We are failing fast now.

shared/templates/templates.go
```diff
@@ -0,0 +1,147 @@
...
+// RenderToString renders the template referenced with the template name using
+// the data provided and write it to the writer provided
+func (c *Container) Render(w io.Writer, templateName string, data Data) error {
+	c.mutex.RLock()
+	htmlTemplates := c.templates
+	c.mutex.RUnlock()
+
+	if htmlTemplates == nil {
+		return nil
+	}
```
agnivade, line 140: We don't need this anymore.

app/notification_email.go
```diff
@@ -98,13 +98,16 @@ func (a *App) sendNotificationEmail(notification *PostNotification, user *model.
...
-	var bodyText = a.getNotificationEmailBody(user, post, channel, channelName, senderName, team.Name, landingURL, emailNotificationContentsType, useMilitaryTime, translateFunc)
-
-	a.Srv().Go(func() {
-		if nErr := a.Srv().EmailService.sendNotificationMail(user.Email, html.UnescapeString(subjectText), bodyText); nErr != nil {
-			mlog.Error("Error while sending the email", mlog.String("user_email", user.Email), mlog.Err(nErr))
-		}
-	})
+	var bodyText, err = a.getNotificationEmailBody(user, post, channel, channelName, senderName, team.Name, landingURL, emailNotificationContentsType, useMilitaryTime, translateFunc)
+	if err != nil {
+		mlog.Error("Error building notification email", mlog.Err(err))
```
agnivade, line 103: If we are not returning, this should be a warning.

agnivade (approved):
Great work!

Thank you so much for your patience with the review @jespino. I know this PR wasn't as fast as you would have preferred. But I think the result of the discussion has been fantastic. The new library looks great, and the code has also improved.

app/app.go
jespino: I don't think so, the server has to have the container initialize a the beginning, so maybe we need to fix a couple of tests after this, bug let see what the CI thinks about that ;)

app/notification_email.go
jespino: I did some changes in the approach, please take a look to the last 2 commits.

shared/templates/templates.go
jespino: Done

app/app.go
jespino: Done

shared/templates/templates.go
```diff
@@ -0,0 +1,147 @@
...
+		return nil, err
+	}
+	c.templates = htmlTemplates
+
+	return c, nil
+}
+
+// NewWithWatcher creates a new templates container scanning a directory and
+// watch the directory filesystem changes to apply them to the loaded
+// templates. This function returns the container an a errors channel to pass
```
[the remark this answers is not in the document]
jespino: Done

app/notification_email.go
agnivade: Looks brilliant to me. Superb work!

agnivade:
@jespino - I can offer to help fix the CI issues. You have done a tremendous job. It's the least I can do.

jespino:
@agnivade no problem, you did an amazing job reviewing and giving the right shape to this too :)

nronas (approved):
Bello 👌 LGTM, awesome work you two.

mgdelacroix (approved):

lindy65:
Hi @jespino - I've tested on the spinwick and have a few questions:

1. Is the layout in the emails within the scope of this PR?  I'm asking because, as you will see in the screenshot below (received when joining a server), the name of the server runs onto 2 lines and looks disjointed. I think it would look better if "Thanks for joining" was on the first line and the entire name of the server is on the second line.

image

2. Inviting a user to the team using the email invitation seems to send a blank email with no details contained in it:

image

jespino:
1 is not related to this Pr, but the 2 could be. I'll investigate it. Thanks! :)

jespino:
@lindy65 this was failing in master to, I'm submitting another PR for that, but I think we can merge this one.

lindy65 (approved):
Thanks @jespino - sorry about the delay - I was out last Thur/Fri...

I believe the email layouts are being looked at by the product team so I haven't created a ticket for the UI issue.

Otherwise looks good to me 👍

PR #16948, opened 2021-02-16
[MM-32922] Cloud release for 20210224

ethervoid:
#### Summary

This is the FF branch for the cloud release for Feb 24th 2021

#### Ticket Link

https://mattermost.atlassian.net/browse/MM-32922

#### Release Note

```release-note
NONE
```


jwilander (approved):

store/sqlstore/upgrade.go
```diff
@@ -940,34 +941,122 @@ func upgradeDatabaseToVersion529(sqlStore *SqlStore) {
...
-	sqlStore.CreateColumnIfNotExists("ThreadMemberships", "UnreadMentions", "bigint", "bigint", "0")
-	sqlStore.CreateColumnIfNotExistsNoDefault("Channels", "Shared", "tinyint(1)", "boolean")
-	sqlStore.CreateColumnIfNotExistsNoDefault("Reactions", "UpdateAt", "bigint", "bigint")
-	sqlStore.CreateColumnIfNotExistsNoDefault("Reactions", "DeleteAt", "bigint", "bigint")
-
-	// saveSchemaVersion(sqlStore, Version5320)
-	// }
+	if hasMissingMigrationsVersion532(sqlStore) {
+		// allow 10 files per post
+		sqlStore.AlterColumnTypeIfExists("Posts", "FileIds", "text", "varchar(300)")
```
marianunez, line 1012: I believe this change was reverted here https://github.com/mattermost/mattermost-server/pull/16908 and not added back?

marianunez: That's probably the reason the test is failing

ethervoid: Yeah but that change could lead to a problem with the migration because that column change was already in production since December.

I've talked with @streamer45  about it and we need to treat this as a new migration change that could be applied to production so we have the same code in master and cloud branches

// @jwilander

marianunez (approved):

store/sqlstore/upgrade.go
ethervoid: Changed the limit of the fields to 300. This should be tackled by the server team, given that [the ticket](https://mattermost.atlassian.net/browse/MM-27312) is from that team

// @grundleborg

PR #16951, opened 2021-02-17
MM-32693: Use master when emoji_layer is invalidated

agnivade:
We add a map which tracks invalidations to the emoji caches
and use master whenever a fresh item is fetched from DB.

https://mattermost.atlassian.net/browse/MM-32693

```release-note
NONE
```


streamer45 (approved):
Looks good :+1:

mgdelacroix (approved):
LGTM! :+1:

PR #16954, opened 2021-02-17
Replacing require.nil in web package

jespino:
#### Summary
Replacing require.nil in web package

#### Ticket Link
Not yet

#### Release Note
```release-note
NONE
```


web/webhook_test.go
```diff
@@ -294,7 +294,7 @@ func TestCommandWebhooks(t *testing.T) {
 
 	for i := 0; i < 5; i++ {
 		response, appErr2 := http.Post(ApiClient.Url+"/hooks/commands/"+hook.Id, "application/json", bytes.NewBufferString(`{"text":"this is a test"}`))
```
wiggin77, line 296: This one shouldn't be called `appErr2` as it's an `error`.

wiggin77 (approved):
LGTM :+1:    One var rename requested.

web/webhook_test.go
jespino: Done :)

agnivade (approved):

PR #16958, opened 2021-02-17
[MM-32691] - Send email when the users are not able to join a workspace

AGMETEOR:
#### Summary
This PR adds a new endpoint ``` /cloud/subscription/limitreached/join``` to send upgrade request emails to admins when users try to join workspace but fail

Screenshot 2021-02-22 at 16 20 10

#### Related PRs
[Enterprise](https://github.com/mattermost/enterprise/pull/931)
[API Spec](https://github.com/mattermost/mattermost-api-reference/pull/613)

#### Ticket Link
[MM-32691](https://mattermost.atlassian.net/browse/MM-32691?atlOrigin=eyJpIjoiYmU1ZjdiMWJlZDRiNGE5ZmJjNGNjYjA2Y2VjZjA5YTEiLCJwIjoiaiJ9)

#### Release Note
```release-note
Added new endpoint  /cloud/subscription/limitreached/join
```


api4/cloud.go
```diff
@@ -409,7 +409,25 @@ func sendAdminUpgradeRequestEmail(c *Context, w http.ResponseWriter, r *http.Req
...
+		c.Err = model.NewAppError("Api4.sendAdminUpgradeRequestEmail", "api.cloud.app_error", nil, e.Error(), http.StatusInternalServerError)
+		return
+	}
+	defer r.Body.Close()
+
+	var upgReqT *model.FreeTierUpgradeRequestType
+	if e = json.Unmarshal(body, &upgReqT); err != nil {
+		c.Err = model.NewAppError("Api4.sendAdminUpgradeRequestEmail", "api.cloud.app_error", nil, e.Error(), http.StatusInternalServerError)
+		return
+	}
```
marianunez, line 423: The current endpoint `/cloud/subscription/limitreached/invite` it's explicit that is for when a user is trying to `invite` others when the limit is reached. 

I believe the intent was to have another one under that handler like:

`/cloud/subscription/limitreached/join` for the case that someone is trying to join when the limit is reached. This way we can simplify here the need for this extra payload and other checks below.

marianunez (changes requested):

api4/cloud.go
AGMETEOR: I have updated the code

api4/cloud.go
```diff
@@ -409,7 +410,33 @@ func sendAdminUpgradeRequestEmail(c *Context, w http.ResponseWriter, r *http.Req
...
+
+func sendAdminUpgradeRequestEmailOnJoin(c *Context, w http.ResponseWriter, r *http.Request) {
+	if c.App.Srv().License() == nil || !*c.App.Srv().License().Features.Cloud {
+		c.Err = model.NewAppError("Api4.sendAdminUpgradeRequestEmailOnJoin", "api.cloud.license_error", nil, "", http.StatusNotImplemented)
+		return
+	}
+
+	user, err := c.App.GetUser(c.App.Session().UserId)
+	if err != nil {
+		c.Err = model.NewAppError("Api4.sendAdminUpgradeRequestEmailOnJoin", err.Id, nil, err.Error(), err.StatusCode)
```
ethervoid, line 429:
```suggestion
		c.Err = model.NewAppError("Api4.sendAdminUpgradeRequestEmailOnJoin", err.Id, nil, err.Error(), err.StatusInternalServerError)
```

ethervoid (approved):
Just a minor comment. LGTM!

api4/cloud.go
```diff
@@ -409,7 +410,33 @@ func sendAdminUpgradeRequestEmail(c *Context, w http.ResponseWriter, r *http.Req
...
+	ReturnStatusOK(w)
+}
+
+func sendAdminUpgradeRequestEmailOnJoin(c *Context, w http.ResponseWriter, r *http.Request) {
+	if c.App.Srv().License() == nil || !*c.App.Srv().License().Features.Cloud {
+		c.Err = model.NewAppError("Api4.sendAdminUpgradeRequestEmailOnJoin", "api.cloud.license_error", nil, "", http.StatusNotImplemented)
+		return
+	}
+
+	user, err := c.App.GetUser(c.App.Session().UserId)
```
marianunez, line 427: In this case that the user is trying to join, they currently don't have user session given they haven't created their account yet. The email template does not include the username

marianunez (changes requested):
Could you add a screenshot of the email that is sent?

Also, is there a related webapp PR where this is called when users reach this new screen https://github.com/mattermost/mattermost-webapp/pull/7468)?

api4/cloud.go
AGMETEOR: Is there another way to get something like their ```id``` or ```username```. Need one of those as key to track the rate limiting

api4/cloud.go
```diff
@@ -36,6 +36,7 @@ func (api *API) InitCloud() {
 	api.BaseRoutes.Cloud.Handle("/subscription/invoices", api.ApiSessionRequired(getInvoicesForSubscription)).Methods("GET")
 	api.BaseRoutes.Cloud.Handle("/subscription/invoices/{invoice_id:in_[A-Za-z0-9]+}/pdf", api.ApiSessionRequired(getSubscriptionInvoicePDF)).Methods("GET")
 	api.BaseRoutes.Cloud.Handle("/subscription/limitreached/invite", api.ApiSessionRequired(sendAdminUpgradeRequestEmail)).Methods("POST")
+	api.BaseRoutes.Cloud.Handle("/subscription/limitreached/join", api.ApiSessionRequired(sendAdminUpgradeRequestEmailOnJoin)).Methods("POST")
```
ethervoid, line 39:
```suggestion
	api.BaseRoutes.Cloud.Handle("/subscription/limitreached/join", api.ApiHandler(sendAdminUpgradeRequestEmailOnJoin)).Methods("POST")
```

api4/cloud.go
ethervoid: Yeah, that's a good point @marianunez for some reason I thought the request was coming from inside the installation.

In this case, we're doing a request to the API without authentication so you need to change the type of handler from `APISessionRequired` to `APIHandler`. Also given that you're not authenticated and you haven't created any user you won't have the username or the id so is not possible to pass that information to the mail

AGMETEOR: @ethervoid any suggestions on how we can handle the rate limiting then?

AGMETEOR: @ethervoid we're going to limit to 1 email per day. So the first user to encounter a block to invitation or join will trigger the sending of email to admin. Other users after that will find the call to that endpoint rate limited. We're also going to be rate limting based on a combination of the date and action (join/invitation)
cc: @marianunez

ethervoid: Ok, if we limit to 1 per day by action and date looks good to me :)

marianunez (approved):

ethervoid:
@AGMETEOR can you take a look at the failing tests pls?

ethervoid (approved):
LGTM! Great work!

PR #16962, opened 2021-02-17
MM-32669 GetPostsSinceForSync

wiggin77:
#### Summary
This PR fixes issues around shared channel sync:

- handle collisions in Post.UpdateAt timestamps via new GetPostsSinceForSync store API
    - simplify query versus existing GetPostsSince
    
- fix redundantly synchronizing posts and users

- ensure GetPostsForSync is not called too often when posts are coming in fast by delaying some tasks up to 2 secs

- ensure sync is done when remote joins shared channel

- rename Post.LastSyncAt to Post.NextSyncAt to better reflect meaning.

Additional changes to Makefile (run-haserver):

- don't `make run-server` as root. It sets root perms for server plugins, parts of webapp, file store, logs, ..., anything the server writes, which messes up the dev environment

- nginx websocket pass-thru requires HTTP 1.1 header when sending upstream; defaults to HTTP 1.0

- don't launch proxy until server nodes are healthy; otherwise nginx will timeout expected upstreams that take longer to build

- increase first health timeout for full re-compiles on slower machines

- instantiate 3 node cluster

- don't rebuild client every time

**To be merged with Shared Channel feature branch.**

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-32669

#### Release Note
```release-note
NONE
```


store/sqlstore/post_store.go
```diff
@@ -517,6 +517,29 @@ func (s *SqlPostStore) GetSingle(id string) (*model.Post, error) {
 	return &post, nil
 }
 
+func (s *SqlPostStore) GetSingleIncDeleted(id string) (*model.Post, error) {
```
streamer45, line 520: Why not making `GetSingle()` accept an `includeDeleted bool` parameter?

model/post.go
```diff
@@ -238,6 +238,17 @@ type GetPostsSinceOptions struct {
 	SortAscending            bool
 }
 
+type GetPostsSinceForSyncOptions struct {
+	ChannelId       string
+	Since           int64 // exclusive
+	Until           int64 // inclusive
```
streamer45, line 244: I am not sure I understand. From the queries both `Since` and `Until` seem to be inclusive, first one is doing `UpdateAt >= ts` while the other `UpdateAt <= ts`.

services/sharedchannel/getpostssince.go
```diff
@@ -0,0 +1,85 @@
...
+// include all posts within that millisecond so that subsequent calls can use an incremented `since`. If this
+// method were to be called repeatedly with the same `since` value the same records would be returned each time
+// and the sync would never move forward.
+//
+// A boolean is also returned to indicate if there are more posts to be synchronized (true) or not (false).
+func (scs *Service) getPostsSince(channelId string, rc *model.RemoteCluster, since int64) (sinceResult, error) {
+	opts := model.GetPostsSinceForSyncOptions{
+		ChannelId: channelId,
+		Since:     since,
+		//ExcludeRemoteId: rc.RemoteId,
```
streamer45, line 30: Should this be removed?

services/sharedchannel/getpostssince.go
```diff
@@ -0,0 +1,85 @@
...
+
+	countPosts := len(posts)
+	if countPosts == 0 {
+		return sinceResult{nextSince: since}, nil
+	}
+
+	var hasMore bool
+	if countPosts > MaxPostsPerSync {
+		hasMore = true
+		peekUpdateAt := posts[countPosts-1].UpdateAt
```
streamer45, line 47: This is probably personal preference but it feels like `countPosts` isn't adding much value as opposed to the more direct `len(posts)`.

services/sharedchannel/getpostssince.go
```diff
@@ -0,0 +1,85 @@
...
+	// walk backward through the slice until we find a post with UpdateAt that differs from milli.
+	var count int
+	for i := len(posts) - 1; i >= 0; i-- {
+		if posts[i].UpdateAt != milli {
+			return count
+		}
+		count++
+	}
+	return count
+}
```
streamer45, line 85: I'd have used binary search here. I am just kidding :p

model/channel_shared.go
```diff
@@ -113,7 +113,7 @@ type SharedChannelRemote struct {
 	IsInviteAccepted  bool   `json:"is_invite_accepted"`
 	IsInviteConfirmed bool   `json:"is_invite_confirmed"`
 	RemoteId          string `json:"remote_id"`
-	LastSyncAt        int64  `json:"last_sync_at"`
+	NextSyncAt        int64  `json:"next_sync_at"`
```
streamer45, line 116: Can you explain the reasons behind the naming change? I am asking because I can't see it from these changes. Was it incorrect before?

streamer45 (commented):
Looks good :+1:  Mainly left questions trying to understand better some decisions.

store/sqlstore/post_store.go
wiggin77: The only reason I didn't do that was when I checked all usages of `GetSingle` it appeared in a bunch of places like integration and plugins.  I didn't take the time to see if was expose beyond the app layer.  (eg. do plugins have any direct access to store APIs?)

If it's safe I can merge the APIs.

model/post.go
wiggin77: Good catch. The comments were left over from when NextSyncAt was LastSyncAt which needed UpdateAt>ts.

services/sharedchannel/getpostssince.go
wiggin77: Hold-over from working in languages where finding length of an array was not inlined.

model/channel_shared.go
wiggin77: If was correct before in that it accurately described what it represented.  The new algorithm ensures all posts with matching UpdateAt values get put into a single sync batch, so the next query can always be the UpdateAt from the last post in the previous batch plus 1 millisecond.  So now it no longer describes the last UpdateAt sync'd - it is where we should sync from next.

This will be simplified once we have a monotonic flakeId for UpdateAt.

model/post.go
wiggin77: Fixed via https://github.com/mattermost/mattermost-server/pull/16962/commits/3f4dc12f54db024764ab8b953b8d947174283d18

services/sharedchannel/getpostssince.go
wiggin77: Fixed via https://github.com/mattermost/mattermost-server/pull/16962/commits/3f4dc12f54db024764ab8b953b8d947174283d18

services/sharedchannel/getpostssince.go
wiggin77: Fixed via https://github.com/mattermost/mattermost-server/pull/16962/commits/3f4dc12f54db024764ab8b953b8d947174283d18

wiggin77:
Whoops, I accidentally included my changes for Makefile and docker-compose.  We need those anyway for testing so I'll leave them here versus separate PR.

build/docker/nginx/default.conf
```diff
@@ -9,6 +10,7 @@ server {
   location ~ /api/v[0-9]+/(users/)?websocket$ {
     proxy_set_header Upgrade $http_upgrade;
     proxy_set_header Connection "upgrade";
+	proxy_http_version 1.1;
```
nronas, line 13: is there an extra space here?

docker-compose.yaml
```diff
@@ -152,8 +203,12 @@ services:
       - ./build/docker/nginx/default.conf:/etc/nginx/conf.d/default.conf
     restart: on-failure
     depends_on:
-      - leader
-      - follower
+      leader:
+        condition: service_healthy
```
nronas, line 207: good catch

build/docker/nginx/default.conf
wiggin77: Tab vs space.  Not sure why the file is not tabbed.  Fixed.

store/sqlstore/post_store.go
streamer45: The store has never been considered a stable API so I'd say it's safe. As far as I know plugins don't access the store directly. They can connect to the db but after that they usually re-implement the querying layer.

services/sharedchannel/getpostssince.go
```diff
@@ -0,0 +1,83 @@
...
+//
+// A boolean is also returned to indicate if there are more posts to be synchronized (true) or not (false).
+func (scs *Service) getPostsSince(channelId string, rc *model.RemoteCluster, since int64) (sinceResult, error) {
+	opts := model.GetPostsSinceForSyncOptions{
+		ChannelId:      channelId,
+		Since:          since,
+		IncludeDeleted: true,
+		Limit:          MaxPostsPerSync + 1, // ask for 1 more than needed to peek at first post in next batch
+	}
+	posts, err := scs.server.GetStore().Post().GetPostsSinceForSync(opts, true)
```
nronas, line 33: Since we are ordering here by just the updateAt in SQL that makes the result non-deterministic since the dbplanner cannot return the same resultsets for records that have the same `UpdateAt`.  That can lead to potentially missed or duplicate syncs, right?

**Records: [A: updateAt = X, B: updateAt = X, C: updateAt = X, D: updateAt = X]**

Limit = 2

If i understand the algo correctly it will go like this:

- GetPostsSinceForSync() return [A,  B] + [C]
- Offset is calculated to 2
- GetPostsSinceForSync() return [D, C, A, B] limit is 1000 with offset 2 essentially the return list is [A, B] so we append to the initial list A and B.

wiggin77: Whoa.  I would have expected natural order (order they were inserted) after ORDER BY is exhausted.  That's how Oracle works.  Thanks for the heads up.  I'll add `Id` as a second OrderBy field.

wiggin77: Fixed via https://github.com/mattermost/mattermost-server/pull/16962/commits/3824abe805b8718cd6a130f4b15027e368148d69

nronas: Bello 👌

nronas (approved):
Superb!!! LGTM

services/sharedchannel/sync_send.go
```diff
@@ -133,70 +197,99 @@ func (scs *Service) processTask(task syncTask) error {
 		remotes = []*model.RemoteCluster{rc}
 	}
 
-	errs := merror.New()
 	for _, rc := range remotes {
-		if err := scs.updateForRemote(task.channelId, rc); err != nil {
-			errs.Append(err)
+		rtask := task
```
streamer45, line 201: Why the copy?

services/sharedchannel/sync_send.go
```diff
@@ -133,70 +197,99 @@ func (scs *Service) processTask(task syncTask) error {
...
 	}
 
 	if len(syncMessages) == 0 {
+		scs.server.GetLogger().Log(mlog.LvlSharedChannelServiceDebug, "sync task, all messages filtered out; skipping sync",
+			mlog.String("remote", rc.DisplayName),
+			mlog.String("channel_id", task.channelId),
+			mlog.Bool("repeat", repeat),
+		)
+
+		// All posts where filtered out, meaning no need to send them. Fast forward SharedChannelRemote's NextSyncAt.
```
streamer45, line 287:
```suggestion
		// All posts were filtered out, meaning no need to send them. Fast forward SharedChannelRemote's NextSyncAt.
```

streamer45 (approved):
Looks good :+1: Mostly left minor wording fixes and a couple of suggestions.

build/docker/nginx/default.conf
```diff
@@ -1,6 +1,7 @@
 upstream app_cluster {
-  server leader:8065 fail_timeout=5s max_fails=10;
-  server follower:8065 fail_timeout=5s max_fails=10;
+  server leader:8065 fail_timeout=10s max_fails=10;
+  server follower:8065 fail_timeout=10s max_fails=10;
+  server peon:8065 fail_timeout=10s max_fails=10;
```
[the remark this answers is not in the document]
wiggin77: Fair enough.  Admittedly my knowledge of the word primarily comes from WoW (original) - not the best source on which to model ones cultural sensitivities.

services/sharedchannel/sync_send.go
wiggin77: I am modifying the source task within a loop and passing the result to a method within that loop.  While technically I could modify the original and let pass-by-value do the copy I think that makes the code more fragile. If for some reason someone makes task a pointer it will break in a subtle way.

wiggin77: I am modifying the original task within a loop and using the copy in a method.  Minor style thing in that changing the incoming task contents feels wrong from a code fragility perspective. I guess it doesn't matter.

streamer45: I guess the question is: should the retry counter increment per remote?

wiggin77: Yes, it must since it fails per remote.

store/sqlstore/post_store.go
wiggin77: I'm going to do this in a different PR since it touches so many files.

PR #16969, opened 2021-02-18
Moving MFA service into shared libs

jespino:
#### Summary
Moving MFA service into shared libs

#### Ticket Link
[MM-33007](https://mattermost.atlassian.net/browse/MM-33007)

#### Release Note
```release-note
NONE
```


utils/html.go
```diff
@@ -120,7 +120,7 @@ func (t *HTMLTemplate) RenderToWriter(w io.Writer) error {
 	return nil
 }
 
-func TranslateAsHtml(t i18n.TranslateFunc, translationID string, args map[string]interface{}) template.HTML {
+func TranslateAsHTML(t i18n.TranslateFunc, translationID string, args map[string]interface{}) template.HTML {
```
agnivade, line 123: Can we do this change in a separate PR? I'd like to keep this PR only for moving the service.

jespino: Because the mattermost-utilities i18n extraction tool has changed to search for `TranslateAsHTML` this is needed to pass the CI's i18n-extract part.

jespino: Master was already changing that, so... now is fixed anyway.

agnivade (approved):

mgdelacroix (approved):

lindy65 (approved):
Thanks @jespino - tested as per test steps on the Jira ticket and did not find any issues 👍

Issue #16968, opened 2021-02-18
TLS: Connection to the server for the organization's internal root certificate and self-signed certificate fails.


3kami3:
#### Summary
I tried to use Docker Image.
I was not able to connect to the server with the organization's internal root certificate and self-signed certificate.

We believe this is not a problem with the mattermost server feature, but with the container image.
The current container image cannot be used with additional certificates for internal root certificates and self-signed certificates.

#### Steps to reproduce
Start the container as usual.
Configure the server to work with the internal root certificate and self-signed certificate.
For example, GitLab SSO, Webhook, etc.
Error log will be output.
```json
{"level":"error","ts":1613627491.40733,"caller":"mlog/log.go:232","msg":"Token request failed.","path":"/signup/gitlab/complete","request_id":"xm8qr4d9opfkbyq4eifd8iay3w","ip_addr":"172.18.0.1","user_id":"","method":"GET","err_where":"AuthorizeOAuthUser","http_code":500,"err_details":"Post \"https://gitlab.example.com/auth/realms/sample/oauth/token\": x509: certificate signed by unknown authority"}
```
#### Expected behavior
There are not anymore details.

#### Observed behavior (that appears unintentional) 
See log above.

#### Possible fixes
- Mount the root certificate to be added to the container
- Run update-ca-certificates

hanzei:
After multiple years of inactivity, we're closing this to help keep our issue tracker focused on active work.

PR #16970, opened 2021-02-18
[GH-16968] Add updating CA certificates

3kami3:
#### Summary
Add updating CA certificates.

#### Ticket Link
Fixes https://github.com/mattermost/mattermost-server/issues/16968
ref. https://github.com/mattermost/mattermost-docker-preview/pull/29

#### Release Note
```release-note
NONE
```

### DIFF
--- a/build/Dockerfile
+++ b/build/Dockerfile
@@ -18,6 +18,7 @@ RUN apk add --no-cache \
   netcat-openbsd \
   xmlsec-dev \
   tzdata \
+  sudo \
   && rm -rf /tmp/*
 
 # Get Mattermost
@@ -26,7 +27,8 @@ RUN mkdir -p /mattermost/data /mattermost/plugins /mattermost/client/plugins \
   else echo "please set the MM_PACKAGE" ; fi \
   && addgroup -g ${PGID} mattermost \
   && adduser -D -u ${PUID} -G mattermost -h /mattermost -D mattermost \
-  && chown -R mattermost:mattermost /mattermost /mattermost/plugins /mattermost/client/plugins
+  && chown -R mattermost:mattermost /mattermost /mattermost/plugins /mattermost/client/plugins \
+  && echo 'mattermost ALL=NOPASSWD: /usr/sbin/update-ca-certificates' >> /etc/sudoers
 
 USER mattermost
 
--- a/build/entrypoint.sh
+++ b/build/entrypoint.sh
@@ -1,5 +1,7 @@
 #!/bin/sh
 
+sudo update-ca-certificates --fresh >/dev/null
+
 if [ "${1:0:1}" = '-' ]; then
     set -- mattermost "$@"
 fi


build/Dockerfile
```diff
@@ -18,6 +18,7 @@ RUN apk add --no-cache \
   netcat-openbsd \
   xmlsec-dev \
   tzdata \
+  sudo \
```
khos2ow, line 21: `sudo` is not needed here.

build/Dockerfile
```diff
@@ -26,7 +27,8 @@ RUN mkdir -p /mattermost/data /mattermost/plugins /mattermost/client/plugins \
   else echo "please set the MM_PACKAGE" ; fi \
   && addgroup -g ${PGID} mattermost \
   && adduser -D -u ${PUID} -G mattermost -h /mattermost -D mattermost \
-  && chown -R mattermost:mattermost /mattermost /mattermost/plugins /mattermost/client/plugins
+  && chown -R mattermost:mattermost /mattermost /mattermost/plugins /mattermost/client/plugins \
+  && echo 'mattermost ALL=NOPASSWD: /usr/sbin/update-ca-certificates' >> /etc/sudoers
```
khos2ow, line 31: Instead of adding `mattermost` user to `sudoers` file you can just directly execute `update-ca-certificates` and move it out of entrypoint.

build/entrypoint.sh
```diff
@@ -1,5 +1,7 @@
 #!/bin/sh
 
+sudo update-ca-certificates --fresh >/dev/null
```
khos2ow, line 3: This should be in `Dockerfile` rather than entrypoint.

khos2ow (changes requested):
Thanks @3kami3, the idea is really solid but please look at comments below for a slight change of execution.

build/entrypoint.sh
3kami3: Thanks for the code review and comments. @khos2ow 
I think it's important to run it in an entrypoint instead of a Dockerfile.

If team A and team B require different organizational root certificates and self-certificates, running Dockerfile will require multiple images with each certificate statically embedded, and many image checks will be required.

If you run update-ca-certificates dynamically at container execution time with entrypoint, you only need one container image.

I think it is better to use certificates that expire and revoke dynamically and flexibly, instead of statically including them in the container image.
This way, there is no need to rebuild the container image every time a certificate expires and revokes.

Do you have any other good ideas to solve this problem?

khos2ow: Sorry for the massive delay! Your suggestions sounds good to me.

khos2ow (approved):
Thank you @3kami3, lgtm!

3kami3:
Thank you for your review.

Please review this PR. @jwilander @stylianosrigas

build/entrypoint.sh
khos2ow: @cpanato wdyt?

3kami3: How I think it should be used.
Example of docker-compose.yml for use with a single container image
```yaml
services:
  db:
    image: postgres:latest
<snip>
 mattermost:
    image: mattermost/mattermost-team-edition:same-version
    volumes:
      - ./certs/team-A.crt:/usr/local/share/ca-certificates/team-A.crt ### <- root certificates for team-A
      #- ./certs/team-B.crt:/usr/local/share/ca-certificates/team-B.crt ### <- root certificates for team-B
      - ./volumes/app/mattermost/config:/mattermost/config:rw
      - ./volumes/app/mattermost/data:/mattermost/data:rw
      - ./volumes/app/mattermost/logs:/mattermost/logs:rw
      - ./volumes/app/mattermost/plugins:/mattermost/plugins:rw
      - ./volumes/app/mattermost/client-plugins:/mattermost/client/plugins:rw
      - /etc/localtime:/etc/localtime:ro
    environment:
      - MM_USERNAME=mmuser
      - MM_PASSWORD=mmuser_password
      - MM_DBNAME=mattermost
```

cpanato: I'm not too fond of the idea to have the mattermost user a sudo permission to execute any command as root. This can lead to some malicious attack if the attacker has access to the container.

If need to add custom certificates, you can have a pipeline to rebuild the docker image and adding your custom CA there. I see this as common practices for other images

any thoughts @jwilander @stylianosrigas

spirosoik: +1 that we should not have mattermost user a sudoer

spirosoik:
I believe we can close this actually.

PR #16977, opened 2021-02-18
Mm 29605 read permission s3 bucket

maxerenberg:
#### Summary
This pull request prevents the `/file/s3_test endpoint` from creating new S3 buckets.

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-29605

#### Release Note
```release-note
Previously, testing an S3 connection would create a new bucket if the bucket name specified in the
request did not exist. Now, the connection test will simply fail.
```

#### Original PR
https://github.com/mattermost/mattermost-server/pull/16690


sbishel (commented):
Does `localstore.go` require the same type of changes to prevent directory creation?

maxerenberg:
> Does `localstore.go` require the same type of changes to prevent directory creation?

Yes (this was mentioned in the old PR). I think local store changes would have to go in a separate PR.

shared/filestore/s3store.go
```diff
@@ -41,6 +41,7 @@ const (
 	bucketNotFound = "NoSuchBucket"
 )
 
+var ErrNoS3Bucket = errors.New("no such bucket")
```
hahmadia, line 44: Nit: 
Just add a small comment explaining it. Ex.
```suggestion
// ErrNoS3Bucket is returned when testing connection and no S3 bucket is found (feel free to word it better)
var ErrNoS3Bucket = errors.New("no such bucket")
```

app/server.go
```diff
@@ -502,8 +502,14 @@ func NewServer(options ...Option) (*Server, error) {
 	if appErr != nil {
 		mlog.Error("Problem with file storage settings", mlog.Err(appErr))
 	} else {
-		if nErr := backend.TestConnection(); nErr != nil {
-			mlog.Error("Problem with file storage settings", mlog.Err(nErr))
+		nErr := backend.TestConnection()
+		if nErr != nil {
+			if errors.Is(nErr, filestore.ErrNoS3Bucket) {
+				nErr = backend.(*filestore.S3FileBackend).MakeBucket()
```
hahmadia, line 508: I'm just curious but what is this notation called `.(*filestore.S3FileBackend).` . Interested to understand how you can call the function `MakeBucket` when it isn't defined in the `FileBackend` interface. 

Also, let's add a test for this where no S3 bucket is found and it attempts to make a new one. Ensure that this really does happen.

shared/filestore/s3store.go
```diff
@@ -157,16 +158,18 @@ func (b *S3FileBackend) TestConnection() error {
...
-			return errors.Wrap(err, "unable to create the s3 bucket")
-		}
+	if !exists {
+		return ErrNoS3Bucket
 	}
+	mlog.Debug("Connection to S3 or minio is good. Bucket exists.")
+	return nil
+}
 
+func (b *S3FileBackend) MakeBucket() error {
```
hahmadia, line 168: Can we write tests for this function? One where it succeeds and one where it fails (mock the fail).

.gitignore
```diff
@@ -10,6 +10,7 @@ webapp/yarn-error.log
 mattermost.mattermost-license
 config/mattermost.mattermost-license
 config/config.json
+config/*.crt
```
hahmadia, line 13: What is this we are adding to gitignore?

app/server.go
maxerenberg: It's a [type assertion](https://tour.golang.org/methods/15).

I'll make sure to add a test for the server startup.

.gitignore
maxerenberg: It's for the Keycloak SAML setup. It's not related to this PR but I think it would be useful.

shared/filestore/s3store.go
maxerenberg: This function is fairly short - what kind of test did you have in mind?

hahmadia: See if you can mock the error and make sure it returns that. Although pretty sure, just wanna make sure we have some test coverage for any new function introduced even if something simple.

sbishel:
> Yes (this was mentioned in the old PR). I think local store changes would have to go in a separate PR.

Is there another PR/ticket for localstorage.go. Seems both should be fixed at the same time.

hahmadia (approved):
Awesome. Thanks! LGTM 👍

mkraft (approved):

sbishel (approved):
LGTM - Please create Jira Ticket for localstorage.go.

furqanmlk (commented):
Are we ok with this user message ?

image

maxerenberg:
> Are we ok with this user message ?
> 
> image

I'll have to create a webapp PR for this. See https://mattermost.atlassian.net/browse/MM-32679 as well.

furqanmlk (approved):

PR #16992, opened 2021-02-21
MM-32882: Add an unread badge/marker to the Main Menu icon and the ‘Plugin Marketplace’ Menu Item

catalintomai:
#### Summary
Add an unread badge/marker to the Main Menu icon and the ‘Plugin Marketplace’ Menu Item until an admin user visits the marketplace for the first time

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-32882

#### Related PRs
api: https://github.com/mattermost/mattermost-api-reference/pull/616
webapp: https://github.com/mattermost/mattermost-webapp/pull/7563

```release-note
NONE
```


api4/system.go
```diff
@@ -68,6 +69,8 @@ func (api *API) InitSystem() {
 	api.BaseRoutes.System.Handle("/notices/view", api.ApiSessionRequired(updateViewedProductNotices)).Methods("PUT")
 
 	api.BaseRoutes.System.Handle("/support_packet", api.ApiSessionRequired(generateSupportPacket)).Methods("GET")
+	api.BaseRoutes.ApiRoot.Handle("/first_admin_visit_marketplace", api.ApiHandler(setFirstAdminVisitMarketplaceStatus)).Methods("POST")
+	api.BaseRoutes.ApiRoot.Handle("/first_admin_visit_marketplace", api.ApiSessionRequired(getFirstAdminVisitMarketplaceStatus)).Methods("GET")
```
hanzei, line 73: Should these be part of the Marketplace routes? While the code makes use of the system store the API still has to do with more the Marketplace then with the system.

hanzei, line 73: Nit pick Usually `GET` routes are placed before `POST` routes.

api4/system.go
```diff
@@ -802,6 +805,59 @@ func requestTrialLicenseAndAckWarnMetric(c *Context, w http.ResponseWriter, r *h
...
+		c.SetPermissionError(model.PERMISSION_MANAGE_SYSTEM)
+		return
+	}
+
+	firstAdminVisitMarketplaceObj, err := c.App.Srv().Store.System().GetByName(model.SYSTEM_FIRST_ADMIN_VISIT_MARKETPLACE)
+	if err != nil {
+		var nfErr *store.ErrNotFound
+		switch {
+		case errors.As(err, &nfErr):
+			err = model.NewAppError("getFirstAdminVisitMarketplaceStatus", "api.no_first_admin_visit_marketplace_status", nil, nfErr.Error(), http.StatusNotFound)
```
hanzei, line 845: Why does `err` need to be set here? I doesn't seem to be used later.

api4/system.go
```diff
@@ -802,6 +805,59 @@ func requestTrialLicenseAndAckWarnMetric(c *Context, w http.ResponseWriter, r *h
...
+			resp := &model.System{
+				Name:  model.SYSTEM_FIRST_ADMIN_VISIT_MARKETPLACE,
+				Value: strconv.FormatBool(false),
+			}
```
hanzei, line 849: Does it make sense to set `firstAdminVisitMarketplaceObj` here and rely on the `Write` in L858?

api4/system.go
```diff
@@ -802,6 +805,59 @@ func requestTrialLicenseAndAckWarnMetric(c *Context, w http.ResponseWriter, r *h
...
+	if !c.App.SessionHasPermissionTo(*c.App.Session(), model.PERMISSION_MANAGE_SYSTEM) {
+		c.SetPermissionError(model.PERMISSION_MANAGE_SYSTEM)
+		return
+	}
+
+	firstAdminVisitMarketplaceObj, err := c.App.Srv().Store.System().GetByName(model.SYSTEM_FIRST_ADMIN_VISIT_MARKETPLACE)
+	if err != nil {
+		var nfErr *store.ErrNotFound
+		switch {
+		case errors.As(err, &nfErr):
```
hanzei, line 844: Nit pick: Should we use `errors.Is` here as we don't need any special property from `store.ErrNotFound`?

store/sqlstore/system_store.go
```diff
@@ -98,7 +99,9 @@ func (s SqlSystemStore) Get() (model.StringMap, error) {
 
 func (s SqlSystemStore) GetByName(name string) (*model.System, error) {
 	var system model.System
-	if err := s.GetMaster().SelectOne(&system, "SELECT * FROM Systems WHERE Name = :Name", map[string]interface{}{"Name": name}); err != nil {
+	if err := s.GetMaster().SelectOne(&system, "SELECT * FROM Systems WHERE Name = :Name", map[string]interface{}{"Name": name}); err == sql.ErrNoRows {
+		return nil, store.NewErrNotFound("System", fmt.Sprintf("name=%s", system.Name))
+	} else if err != nil {
```
hanzei, line 104: Should we stick with the standard pattern that is used in the code base here?

```suggestion
	if err := s.GetMaster().SelectOne(&system, "SELECT * FROM Systems WHERE Name = :Name", map[string]interface{}{"Name": name}); err != nil {
	if err == sql.ErrNoRows {
		return nil, store.NewErrNotFound("System", fmt.Sprintf("name=%s", system.Name))
	}
```

hanzei (changes requested):
The approach looks good to me. A couple of questions below.

store/sqlstore/system_store.go
larkox: If we were going forward with this change, I would be more comfortable with a Server Platform team member review.

larkox (approved):

model/system.go
```diff
@@ -37,6 +37,7 @@ const (
 	USER_LIMIT_OVERAGE_CYCLE_END_DATE             = "UserLimitOverageCycleEndDate"
 	OVER_USER_LIMIT_FORGIVEN_COUNT                = "OverUserLimitForgivenCount"
 	OVER_USER_LIMIT_LAST_EMAIL_SENT               = "OverUserLimitLastEmailSent"
+	SYSTEM_FIRST_ADMIN_VISIT_MARKETPLACE          = "FirstAdminVisitMarketplace"
```
jfrerich, line 40: Drive-by review here. Nit: move the new entry up into the other `SYSTEM_` constants.

api4/system.go
catalintomai: fwis, this is the current pattern used in the code.

api4/system.go
catalintomai: I've seen it done differently in diff places though (including our own `.../api4/plugin.go`)

catalintomai: @hanzei, @jfrerich  - are you guys proposing having a `marketplace` prefix, like `/marketplace/first_admin_visit` or moving the handle statements to the `api4/plugin.go` file, or both?

store/sqlstore/system_store.go
catalintomai: Adding @streamer45 from the Server team to the review.

api4/system.go
hanzei: Right now all Marketplace related routes are in `api4/plugin.go`. I think the two new routes should be places in the same file. Given that the routes are under `/plugins/` right now, I leaning toward thinking `api4/plugin.go` is the right place.

Should the new routes be under `/plugins/marketplace/first_admin_visit` despite `/plugins/marketplace/` being an a handler?

hanzei: Good point. Then it's just me wishing things are more consistent :wink:

hanzei (changes requested):
LGTM, beside open discussion on the URL path: https://github.com/mattermost/mattermost-server/pull/16992#discussion_r581877683

api4/system.go
```diff
@@ -802,6 +805,57 @@ func requestTrialLicenseAndAckWarnMetric(c *Context, w http.ResponseWriter, r *h
...
+			firstAdminVisitMarketplaceObj = &model.System{
+				Name:  model.SYSTEM_FIRST_ADMIN_VISIT_MARKETPLACE,
+				Value: strconv.FormatBool(false),
+			}
+		default:
+			c.Err = model.NewAppError("getFirstAdminVisitMarketplaceStatus", "api.error_get_first_admin_visit_marketplace_status", nil, err.Error(), http.StatusInternalServerError)
+		}
+		return
```
hanzei, line 852: Should the code return in all cases or only in the error case?

hanzei (commented):
Two additional questions:

1. Do we need to update `model/Client4.go`?
2. Is it worth adding a small test on the API layer?

api4/system.go
catalintomai: good catch. we still want to return in the "no entries case" - it is a special error case.

api4/system.go
catalintomai: Don't think there is an issue with that.

catalintomai:
> Two additional questions:
> 
> 1. Do we need to update `model/Client4.go`?
> 2. Is it worth adding a small test on the API layer?

#1 - don't see why
#2 - there isn't really any business/mid-tier  logic to test (the store part should already be covered).

api4/plugin.go
```diff
@@ -383,3 +388,55 @@ func installPlugin(c *Context, w http.ResponseWriter, plugin io.ReadSeeker, forc
...
+	c.LogAudit("attempt")
+
+	if !c.App.SessionHasPermissionTo(*c.App.Session(), model.PERMISSION_MANAGE_SYSTEM) {
+		c.SetPermissionError(model.PERMISSION_MANAGE_SYSTEM)
+		return
+	}
+
+	if err := c.App.Srv().Store.System().SaveOrUpdate(&model.System{
+		Name:  model.SYSTEM_FIRST_ADMIN_VISIT_MARKETPLACE,
+		Value: strconv.FormatBool(true),
```
streamer45, line 404:
```suggestion
		Value: "true",
```

No need to convert a static value.

api4/plugin.go
```diff
@@ -383,3 +388,55 @@ func installPlugin(c *Context, w http.ResponseWriter, plugin io.ReadSeeker, forc
...
+	}
+
+	firstAdminVisitMarketplaceObj, err := c.App.Srv().Store.System().GetByName(model.SYSTEM_FIRST_ADMIN_VISIT_MARKETPLACE)
+	if err != nil {
+		var nfErr *store.ErrNotFound
+		switch {
+		case errors.As(err, &nfErr):
+			firstAdminVisitMarketplaceObj = &model.System{
+				Name:  model.SYSTEM_FIRST_ADMIN_VISIT_MARKETPLACE,
+				Value: strconv.FormatBool(false),
```
streamer45, line 431:
```suggestion
				Value: "false",
```

Same thing.

streamer45 (approved):
Looks good, nothing blocking :+1:

hanzei (approved):
I think the policy for `model/client4.go` is that it should cover all API routes. But I'm going to deffer to @streamer45 on that.

Otherwise LGTM :+1:

streamer45:
> I think the policy for `model/client4.go` is that it should cover all API routes. But I'm going to deffer to @streamer45 on that.

Usually that's the case but there are methods not covered and since it's easy enough to add them but much more difficult to remove/change them I'd say it's okay for such a UI specific method to be left out for now.

DHaussermann:
Sorry for the delay in getting to this @catalintomai. I did do a round of testing. 
Seems like there may be an issue in Firefox. I see the hamburger menu marker but not the Marketplace marker. 
Screen Shot 2021-03-12 at 6 06 55 PM
This screen shows the same user logged in to Firefox on the left and Chrome on the right. It seems like it's rendering differently and just giving me the truncation `...` Can you please take a look? If it's relevant - I'm using FireFox 85.0.2 (it's recent)

catalintomai:
Thanks @DHaussermann , I'll take a look - it could be just the amount of spacing between the text and the marker. cc: @abhijit-singh

catalintomai:
> Thanks @DHaussermann , I'll take a look - it could be just the amount of spacing between the text and the marker. cc: @abhijit-singh

@DHaussermann , I tried to repro on (the latest) Firefox 86.0.1 on my Macbook PRO and wasn't able to. What OS are you using - and could you try with the latest Firefox?

DHaussermann:
@catalintomai 
1. Just to have a common point of comparison -  When I login to the webapp PR server created here https://github.com/mattermost/mattermost-webapp/pull/7563#issuecomment-783297228 I see the issue still. (not sure how this is even working without the server side code. But, seems like a very quick way to try and repro what I was seeing.) Wonder if you see the same thing as me here. (this server was update with the latest commit)
I am running Mac OS Catalina 10.15.6 and latest version of FireFox v 86.0.1 It does not look like I have a newer version available. 
Screen Shot 2021-03-15 at 10 03 34 AM

2. Unfortunately I'm no longer able to deploy this locally. I update all 3 PRs to pull in master and when I deploy this the webapp no longer loads and the console shows:
```
middleware.js?772e:8 Uncaught Error: Cannot find module 'redux-action-buffer'
    at webpackMissingModule (webpack-internal:///./node_modules/mattermost-redux/store/middleware.js:8)
    at eval (webpack-internal:///./node_modules/mattermost-redux/store/middleware.js:8)
    at Object../node_modules/mattermost-redux/store/middleware.js (main.b6fc5cb744c67754e8ec.js:13932)
    at __webpack_require__ (main.b6fc5cb744c67754e8ec.js:18316)
    at eval (webpack-internal:///./node_modules/mattermost-redux/store/configureStore.dev.js:26)
    at Object../node_modules/mattermost-redux/store/configureStore.dev.js (main.b6fc5cb744c67754e8ec.js:13888)
    at __webpack_require__ (main.b6fc5cb744c67754e8ec.js:18316)
    at eval (webpack-internal:///./node_modules/mattermost-redux/store/index.js:6)
    at Object../node_modules/mattermost-redux/store/index.js (main.b6fc5cb744c67754e8ec.js:13910)
    at __webpack_require__ (main.b6fc5cb744c67754e8ec.js:18316)
    ```

catalintomai:
Thanks @DHaussermann , I'll fix the local deployment (prob the redux hash is the old one, though if you install all three PRs things should work automatically - just run "npm run dev" in the mattermost-redux repo). 
My system is Mac OS Big Sur 11.2.3  - I'll paste my screenshot here next.

DHaussermann:
Thanks @catalintomai Seems this is working when all branches are properly deployed now.
There is a very minor UI oddity in that sysadmins will see the badge briefly while the UI loads when it should no longer be visible. In my judgment this is minor and I will create a follow-up task to address it later

Tested and passed
- Badge is visible on all browser and desktop
- Badge is only visible to sysadmins
- Badge will be removed once 1 system admin opens market place 
- Created follow-up task to address https://mattermost.atlassian.net/browse/MM-33830

LGTM!

DHaussermann (approved):

DHaussermann:
@catalintomai I have done more testing on this with the latest webapp. 
 
I am unable to reproduce https://mattermost.atlassian.net/browse/MM-33830 Tested various users, browsers and network throttling.  

I did have one follow-up question about this. The badge will appear for all system admins but once a system admin open the marketplace the other system admins will continue to see the badge until they do something that causes a refresh.
The admins will see the badge until they navigate to system console or one of the "backstage UIs" such as Integrations and then return to the user side. 

Just to make sure - We don't need something to dynamically remove the badge for other admins here?

catalintomai:
Thanks @DHaussermann -  the "remove the badge for the other admins" issue should be fixe in the latest iteration.

DHaussermann (approved):
Tested and passed
- Regression tested that badge works in a in all browsers and desktop
- Regression tested that only system admins see the badge
- Confirmed new functionality is working to clear the badge for all system admins once any admin opens the modal.  There is no longer a need to navigate to system console or perform any other action to trigger a refresh. That badge disappears i real time

LGTM!
Thanks @catalintomai 

Documented functional test for this in Zephyre here as MM-T3940

PR #16994, opened 2021-02-22
MM-29607 allow user to deactivate MFA even when disabled

maxerenberg:
#### Summary
This PR allows users to deactivate MFA for their account, even when MFA is disabled on the server.

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-29607

#### Release Note
```release-note
Users can now deactivate MFA for their accounts even if MFA is disabled on the server.
```


app/user_test.go
```diff
@@ -1394,3 +1394,18 @@ func TestUpdateUserRolesWithUser(t *testing.T) {
 	_, err = th.App.UpdateUserRolesWithUser(user, "does not exist", false)
 	require.NotNil(t, err)
 }
+
+func TestDeactivateMfa(t *testing.T) {
+	t.Run("MFA is disabled", func(t *testing.T) {
+		th := Setup(t).InitBasic()
+		defer th.TearDown()
+
```
hahmadia, line 1402: Is it possible to enable MFA here and assign the user some fake MFA code and then turn off MFA and see if the user can deactivate?
This will more closely mimic the scenario explained in the ticket. Not sure if it's possible but would recommend to look into if it is.

hahmadia (approved):
This change makes sense to me. A small comment for test but good otherwise!

Activating MFA should rely on `EnableMultifactorAuthentication` since you shouldn't be allowed to enable if that's false.

However, disabling should be allowed to be done anytime especially since if you enable it prior to your server disabling multifactor authentication.

Nice job!

sbishel (approved):

app/user_test.go
maxerenberg: The `App.ActivateMfa` method calls the MFA module directly, which then calls `dgoogauth.OTPConfig.Activate`. We would probably have to create mock interfaces for the MFA module if we want to enable MFA in our test.

furqanmlk (approved):
**If MFA is disabled, API enable sends error message**
image

**If MFA is disable, API disable user MFA**
image

PR #17004, opened 2021-02-23
[MM-32621] Remove app.waitForSidebarCategories

streamer45:
#### Summary

PR removes the need for `app.waitForSidebarCategories()` by making `ChannelStore.createInitialSidebarCategories()` return what it created instead of querying categories (from replica) through `GetSidebarCategories()`.

Interestingly `GetSidebarCategories()` was already querying from master so I am not sure why `app.waitForSidebarCategories()` was ever needed. 

#### Ticket

https://mattermost.atlassian.net/browse/MM-32621

```release-note
NONE
```

### DIFF
--- a/store/sqlstore/channel_store_categories.go
+++ b/store/sqlstore/channel_store_categories.go
@@ -14,22 +14,27 @@ import (
 	"github.com/mattermost/mattermost-server/v5/store"
 )
 
-func (s SqlChannelStore) CreateInitialSidebarCategories(userId, teamId string) error {
+func (s SqlChannelStore) CreateInitialSidebarCategories(userId, teamId string) (*model.OrderedSidebarCategories, error) {
 	transaction, err := s.GetMaster().Begin()
 	if err != nil {
-		return errors.Wrap(err, "CreateInitialSidebarCategories: begin_transaction")
+		return nil, errors.Wrap(err, "CreateInitialSidebarCategories: begin_transaction")
 	}
 	defer finalizeTransaction(transaction)
 
-	if err := s.createInitialSidebarCategoriesT(transaction, userId, teamId); err != nil {
-		return errors.Wrap(err, "CreateInitialSidebarCategories: createInitialSidebarCategoriesT")
+	if err = s.createInitialSidebarCategoriesT(transaction, userId, teamId); err != nil {
+		return nil, errors.Wrap(err, "CreateInitialSidebarCategories: createInitialSidebarCategoriesT")
+	}
+
+	oc, err := s.getSidebarCategoriesT(transaction, userId, teamId)
+	if err != nil {
+		return nil, errors.Wrap(err, "CreateInitialSidebarCategories: getSidebarCategoriesT")
 	}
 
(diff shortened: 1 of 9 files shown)


store/sqlstore/channel_store_categories.go
```diff
@@ -14,22 +14,27 @@ import (
...
 	}
 	defer finalizeTransaction(transaction)
 
-	if err := s.createInitialSidebarCategoriesT(transaction, userId, teamId); err != nil {
-		return errors.Wrap(err, "CreateInitialSidebarCategories: createInitialSidebarCategoriesT")
+	if err = s.createInitialSidebarCategoriesT(transaction, userId, teamId); err != nil {
+		return nil, errors.Wrap(err, "CreateInitialSidebarCategories: createInitialSidebarCategoriesT")
+	}
+
+	oc, err := s.getSidebarCategoriesT(transaction, userId, teamId)
```
agnivade, line 28: Hmm .. so this is doing read after write again. We want to return the data from `createInitialSidebarCategoriesT` itself. We already know what we are inserting, so there is no need to read from the DB again.

agnivade (commented):
Also cc @grundleborg since he was part of the initial review.

store/sqlstore/channel_store_categories.go
streamer45: We are reading from master so I don't see issues as far as consistency goes. If we are concerned about the additional query I guess we can just transform the data directly in `createInitialSidebarCategoriesT()`.

agnivade: Correct. That's what I was trying to say. Sorry for any confusion. For any set of queries, that write and then read the same data back again, we should return the written value from the write method itself. There is no need for an extra read query.

Somewhat similar to this case here: https://community-daily.mattermost.com/core/pl/nphtmkowcjd8ic76tbqtapx6nc. Except that the create method doesn't even return anything here.

streamer45: I can try to do that. My only worry is that we might not have enough information in the create method to return all the data we require in which case we would need to run additional queries anyway. There's also the risk that the db state differs from what we expect so again the create wouldn't return a true representation of it.

agnivade: Discussed offline. It's not exactly a read-after-write problem. The existing approach looks great.

agnivade (approved):
Brilliant :sparkles:

hmhealey (approved):
Having `CreateSidebarCategories` just return what it's created makes sense to me. I think we didn't do that originally because of the implicitly added channels and potential replica lag (since I thought `GetSidebarCategories` used the replica, whoops), but just calling `getSidebarCategoriesT` in the transaction makes sense to me

streamer45:
@lindalumitchell 

PR shouldn't really alter the existing functionality. A quick smoke test around sidebar categories (making sure default ones get created) should be enough.

lindalumitchell (approved):
LGTM!

PR #17010, opened 2021-02-24
Move commands table over to new schema migrations

nronas:
#### Summary
implements: https://mattermost.atlassian.net/browse/MM-33063

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-33063

#### Release Note

```release-note
* Moves Commands gorp driven migrations to schema-based migrations
```


db/migrations/mysql/000003_create_commands.up.sql
```diff
@@ -0,0 +1,42 @@
...
+    Method varchar(1) DEFAULT NULL,
+    Username varchar(64) DEFAULT NULL,
+    IconURL text,
+    AutoComplete tinyint(1) DEFAULT NULL,
+    AutoCompleteDesc text,
+    AutoCompleteHint text,
+    DisplayName varchar(64) DEFAULT NULL,
+    Description varchar(128) DEFAULT NULL,
+    URL text,
+    PluginId text,
```
mgdelacroix, line 19: We should remove this from the create table as it's added right after in the alter

db/migrations/postgres/000003_create_commands.up.sql
```diff
@@ -0,0 +1,29 @@
...
+    method VARCHAR(1),
+    username VARCHAR(64),
+    iconurl VARCHAR(1024),
+    autocomplete bool,
+    autocompletedesc VARCHAR(1024),
+    autocompletehint VARCHAR(1024),
+    displayname VARCHAR(64),
+    description VARCHAR(128),
+    url VARCHAR(1024),
+    pluginid VARCHAR(190)
```
mgdelacroix, line 19: Same here

db/migrations/mysql/000003_create_commands.up.sql
nronas: Full disclosure on this, i was trying to mimic what gorp + upgrade is doing. But happy to drop that column from here

PR #17026, opened 2021-02-26
MM-33233: Fix double close of webconn pump

agnivade:
There would be an edge case where an epoll HUP event
would occur at the same time as a connection close event,
and both would race to close the connection.

We fix this by enclosing the Close method with a sync.Once
thereby guaranteeing idempotency.

https://mattermost.atlassian.net/browse/MM-33233

```release-notes
Fixed a race condition which would crash the app server due
to improper handling of websocket closing.
```

### DIFF
--- a/app/web_conn.go
+++ b/app/web_conn.go
@@ -58,6 +58,7 @@ type WebConn struct {
 	isWindows                 bool
 	endWritePump              chan struct{}
 	pumpFinished              chan struct{}
+	closeOnce                 sync.Once
 }
 
 // NewWebConn returns a new WebConn instance.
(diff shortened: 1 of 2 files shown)


agnivade:
@streamer45 - It would be good to have your magic touch with a load test just to see if it reproes again. :wink:  No rush of course.

nronas (approved):
LGTM

streamer45 (approved):
Nice :+1: Stress-tested this for a while and couldn't reproduce :rocket:

agnivade:
Verified with load tests. So no QA needed for this.

PR #17028, opened 2021-02-26
Add cloud attributes to feature flags and allow non-cloud servers to use split.io sync

jwilander:
#### Summary
Three things in this PR:

1. Add some cloud attributes to the feature flags so that we can use those to isolate treatments.
2. Allow non-cloud servers to use the split.io feature flag management. Primary motivator there is allowing our test servers to use split.io for feature flags even if they don't have a cloud license.
3. Add a bit more debug logs to feature flags.

#### Release Note

```release-note
Add attributes to split.io feature flags
```

### DIFF
--- a/app/feature_flags.go
+++ b/app/feature_flags.go
@@ -4,6 +4,7 @@
 package app
 
 import (
+	"os"
 	"time"
 
 	"github.com/mattermost/mattermost-server/v5/config"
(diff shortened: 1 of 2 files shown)


app/feature_flags.go
```diff
@@ -15,16 +17,15 @@ import (
...
 	defer s.featureFlagSynchronizerMutex.Unlock()
-	license := s.License()
-	inCloud := license != nil && *license.Features.Cloud
 	splitKey := *s.Config().ServiceSettings.SplitKey
-	syncFeatureFlags := inCloud && splitKey != "" && s.IsLeader()
+	splitConfigured := splitKey != ""
+	syncFeatureFlags := splitConfigured && s.IsLeader()
 
-	s.configStore.PersistFeatures(inCloud)
+	s.configStore.PersistFeatures(splitConfigured)
```
jwilander, line 24: @crspeller this is the only part I'm unsure about. Should we only persist them for Cloud still?

crspeller: This issue with persisting outside cloud environments is undesirable config file changes. Since you still need to have split.io configured, and end-users shouldn't have that configured, I think this is probably fine.

crspeller (approved):
Good to see this getting configured for production cloud :)

gabrieljackson (approved):
LGTM

PR #17031, opened 2021-02-26
[MM-33333] Fix GetSidebarCategories() failing on db replicas

streamer45:
#### Summary

PR fixes `GetSidebarCategories()` from failing on db replicas. Now I understand why it was using master before :)
It was running everything in a transaction which isn't going to work on read-only replicas. To avoid code duplication I've added a small interface which makes `getSidebarCategoriesT` accept both transactions and regular db handlers.

#### Ticket

https://mattermost.atlassian.net/browse/MM-33333

#### Release Note

```release-note
NONE
```

### DIFF
--- a/store/sqlstore/channel_store_categories.go
+++ b/store/sqlstore/channel_store_categories.go
@@ -508,12 +508,7 @@ func (s SqlChannelStore) getSidebarCategoriesT(db dbSelecter, userId, teamId str
 }
 
 func (s SqlChannelStore) GetSidebarCategories(userId, teamId string) (*model.OrderedSidebarCategories, error) {
-	oc, err := s.getSidebarCategoriesT(s.GetReplica(), userId, teamId)
-	if err != nil {
-		return nil, err
-	}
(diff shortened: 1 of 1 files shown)


store/sqlstore/channel_store_categories.go
```diff
@@ -504,23 +508,7 @@ func (s SqlChannelStore) getSidebarCategoriesT(transaction *gorp.Transaction, us
...
-	if err != nil {
-		return nil, err
-	}
-
-	if err = transaction.Commit(); err != nil {
-		return nil, errors.Wrap(err, "commit_transaction")
-	}
-
-	return oc, nil
+	return s.getSidebarCategoriesT(s.GetReplica(), userId, teamId)
```
agnivade, line 511: Let's add a comment this time mentioning the reason for the introduction of the interface, so that the context is preserved.

agnivade (approved):

jwilander (approved):

PR #17044, opened 2021-03-02
removed redundant CreateMembershipIfNeeded call

reflog:
#### Summary
A potential race was being caused by a redundant call to CreateMembershipIfNeeded .

#### Ticket Link

https://mattermost.atlassian.net/browse/MM-33396

#### Release Note

```release-note
NONE
```

### DIFF
--- a/app/post.go
+++ b/app/post.go
@@ -458,12 +458,6 @@ func (a *App) handlePostEvents(post *model.Post, user *model.User, channel *mode
 		return err
 	}
 
-	if *a.Config().ServiceSettings.ThreadAutoFollow && post.RootId != "" {
-		if err := a.Srv().Store.Thread().CreateMembershipIfNeeded(post.UserId, post.RootId, true, false, true); err != nil {
-			return err
-		}
-	}
-
 	if post.Type != model.POST_AUTO_RESPONDER { // don't respond to an auto-responder
 		a.Srv().Go(func() {
 			_, err := a.SendAutoResponseIfNecessary(channel, user, post)
--- a/app/notification.go
+++ b/app/notification.go
@@ -184,13 +184,7 @@ func (a *App) SendNotifications(post *model.Post, team *model.Team, channel *mod
 			mac := make(chan *model.AppError, 1)
 			go func(userID string) {
 				defer close(mac)
-				incrementMentions := false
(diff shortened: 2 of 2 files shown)


streamer45 (approved):
Nice :+1:

agnivade (changes requested):
@reflog - Sorry for not having any prior context on this, but can you please explain what exactly is the race here in the PR message?

Can we also add a test for this?

streamer45:
> @reflog - Sorry for not having any prior context on this, but can you please explain what exactly is the race here in the PR message?

@agnivade 

This was spotted during load-tests. The logs were getting filled with "duplicate entries" warnings on inserting `ThreadMembership` rows. It turned out there were two subsequent calls (one of which in a separate goroutine) triggered by `handlePostEvents`. That's where the race originated from. Since we are making db calls it didn't have any real side-effect other than causing a spam of logs (and possibly a few more queries).

reflog:
> 
> 
> > @reflog - Sorry for not having any prior context on this, but can you please explain what exactly is the race here in the PR message?
> 
> @agnivade
> 
> This was spotted during load-tests. The logs were getting filled with "duplicate entries" warnings on inserting `ThreadMembership` rows. It turned out there were two subsequent calls (one of which in a separate goroutine) triggered by `handlePostEvents`. That's where the race originated from. Since we are making db calls it didn't have any real side-effect other than causing a spam of logs (and possibly a few more queries).
☝🏻 
What Claudio said :)

app/post.go
```diff
@@ -458,12 +458,6 @@ func (a *App) handlePostEvents(post *model.Post, user *model.User, channel *mode
 		return err
 	}
```
agnivade, line 460: It seems like there is some difference in the two. This query wouldn't be executed if `ThreadAutoFollow` is set to false and if it's a root post. Whereas the other one will be executed always, along with the `ThreadAutoFollow` passed to it. 

I guess the end result is the same, but I am wondering if we should add the if condition to whichever query we are keeping?

reflog:
@agnivade @streamer45 - I am closing this PR in favor of https://github.com/mattermost/mattermost-server/pull/17068 since both Agniva's comment was a correct one and there were other minor changes.

PR #17051, opened 2021-03-02
Mm 33260/move preferences to schema migrations

nronas:
#### Summary
implements: https://mattermost.atlassian.net/browse/MM-33260

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-33260

#### Release Note

```release-note
* Moves preferences gorp driven migrations to schema-based migrations
```


db/migrations/mysql/000012_create_preferences.up.sql
```diff
@@ -0,0 +1,105 @@
...
+) ENGINE=InnoDB DEFAULT CHARSET=utf8mb4;
+
+SET @preparedStatement = (SELECT IF(
+    (
+        SELECT COUNT(*) FROM INFORMATION_SCHEMA.COLUMNS
+        WHERE table_name = 'Preferences'
+        AND table_schema = DATABASE()
+        AND column_name = 'Value'
+    ) > 0,
+    'ALTER TABLE Preferences MODIFY Value text;',
```
mgdelacroix, line 16: What would happen if the value column already exists and is text? Should we not check on the type instead just on the existence?

db/migrations/mysql/000012_create_preferences.up.sql
```diff
@@ -0,0 +1,105 @@
...
+    ) > 0,
+    'SELECT 1',
+    'CREATE INDEX idx_preferences_name ON Preferences(Name);'
+));
+
+PREPARE createIndexIfNotExists FROM @preparedStatement;
+EXECUTE createIndexIfNotExists;
+DEALLOCATE PREPARE createIndexIfNotExists;
+
+CREATE PROCEDURE RenameSolarizedThemeWithUnderscore()
```
mgdelacroix, line 69: :heart_eyes:

db/migrations/mysql/000012_create_preferences.up.sql
nronas: i think it will be an noop, but better to double check this in order to be sure.

nronas: i've checked what store used to do, and we only care if the column exists: https://github.com/mattermost/mattermost-server/blob/master/store/sqlstore/store.go#L793

So i think we are achieving parity here, but i will spend some time to see if we can improve the check here in order to avoid cycles

PR #17052, opened 2021-03-02
Revert schema change

streamer45:
#### Summary

This migration was reverted in https://github.com/mattermost/mattermost-server/pull/16908 for performance reasons. We diverged from cloud since they had already applied the changes by the time we decided to revert. Still we should not go ahead with this migration on on-prem releases.

#### Release Note

```release-note
NONE
```

### DIFF
--- a/model/post.go
+++ b/model/post.go
@@ -45,7 +45,7 @@ const (
 	POST_EPHEMERAL              = "system_ephemeral"
 	POST_CHANGE_CHANNEL_PRIVACY = "system_change_chan_privacy"
 	POST_ADD_BOT_TEAMS_CHANNELS = "add_bot_teams_channels"
-	POST_FILEIDS_MAX_RUNES      = 300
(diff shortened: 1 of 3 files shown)


ethervoid (approved):
LGTM!

metanerd (approved):

ethervoid:
@streamer45 Given that we have this revert in master we're going to face a continuous merge problem when we do the Feature branching to cloud release, which has the change applied in all the installations DB schema. 

Could you please provide a safer way to revert or re-apply this change in the cloud installations without harming them?

Thank you!

// @jwilander

streamer45:
@ethervoid 

We can probably bump `POST_FILEIDS_MAX_RUNES` and make the migration Postgres only. That should fix the merging issues.  I'll bring this up with my team and if deemed appropriate will make a PR.

ethervoid:
@streamer45 Great! thank you! Really appreciated

PR #17054, opened 2021-03-02
[MM-32343] Environment SubSection

maxerenberg:
#### Summary
This PR adds subsections in the system console for ENVIRONMENT permissions.

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-32343

#### Release Note
```release-note
The Environment section permission has been deprecated and broken down into Environment subsections.
```

### Webapp and Redux PRs
https://github.com/mattermost/mattermost-webapp/pull/7622
https://github.com/mattermost/mattermost-redux/pull/1389


app/permissions_migrations.go
```diff
@@ -632,5 +728,12 @@ func (a *App) DoPermissionsMigrations() error {
 			return err
 		}
 	}
+	// Need to increase the size of the Permissions column in the Roles table
+	for _, db := range a.Srv().sqlStore.GetAllConns() {
```
BenCookie95, line 732: I'm unsure if this is the right place, do we have an area where we run one time schema update?

BenCookie95 (changes requested):

model/config.go
```diff
@@ -2698,7 +2698,7 @@ func (s *JobSettings) SetDefaults() {
 }
 
 type CloudSettings struct {
-	CWSUrl *string `access:"environment,write_restrictable"`
+	CWSUrl *string `access:"write_restrictable"`
```
hahmadia, line 2701: Are we sure we want to remove this? Was this confirmed?

app/permissions_migrations.go
hahmadia, line 732: I spoke to Claudio from our Server team and here is what he had to say:
> You could either increase it or just convert it to `text` and be done with it. In MySQL it would already be `text` anyway since it's more than 255 bytes. I don't see problems either way because I assume the table is usually small. As far as changes, you'd have to change `store/sqlstore/role_store.go` for table creation and add a migration step in `store/sqlstore/upgrade.go`.

In `role_store.go`, you will want to remove the `SetMaxSize` on `Permission`.

In `upgrade.go`, you'll want to write the migration there. You can remove it from here. Also, you can write it only for postgres since MYSQL is already a text. There should already be examples of ALTER in that file for you to follow.

You can see in our gorp library where the decision for text, varchar are happening.
Postgres:
https://github.com/mattermost/gorp/blob/master/dialect_postgres.go#L65

Mysql:
https://github.com/mattermost/gorp/blob/master/dialect_mysql.go#L86

hahmadia (changes requested):
Great work! You've taken on a beast of a section and you've done great! 

I have some minor points to look out for:

- Could we look at `getClusterStatus , getEnvironmentConfig, testEmail` which use `ENVIRONMENT` permissions but haven't been addressed in this PR. All in the `api4` directory files.

api4/config.go
```diff
@@ -76,8 +76,8 @@ func configReload(c *Context, w http.ResponseWriter, r *http.Request) {
 	auditRec := c.MakeAuditRecord("configReload", audit.Fail)
 	defer c.LogAuditRec(auditRec)
 
-	if !c.App.SessionHasPermissionToAny(*c.App.Session(), model.SysconsoleReadPermissions) {
-		c.SetPermissionError(model.SysconsoleReadPermissions...)
+	if !c.App.SessionHasPermissionTo(*c.App.Session(), model.PERMISSION_RELOAD_CONFIG) {
```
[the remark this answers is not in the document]
maxerenberg: This function is only called for the `/config/reload` endpoint. It's used when the user clicks on the "Reload Configuration From Disk" button in the Web Server subsection.
I don't think read permission are sufficient to click this button because it's not a passive operation: the server's config values could change as a result of this API call.

api4/elasticsearch.go
```diff
@@ -32,9 +32,9 @@ func testElasticsearch(c *Context, w http.ResponseWriter, r *http.Request) {
 	}
 
 	// no sniffing for ports or passwords
-	if !c.App.SessionHasPermissionTo(*c.App.Session(), model.PERMISSION_SYSCONSOLE_WRITE_ENVIRONMENT) {
+	if !c.App.SessionHasPermissionTo(*c.App.Session(), model.PERMISSION_SYSCONSOLE_WRITE_ENVIRONMENT_ELASTICSEARCH) {
 		if (*cfg.ElasticsearchSettings.ConnectionUrl != *c.App.Config().ElasticsearchSettings.ConnectionUrl) || (*cfg.ElasticsearchSettings.Password != model.FAKE_SETTING) {
```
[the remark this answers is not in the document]
maxerenberg: Here's how I interpret it...
Usually "testing a connection" would be considered a READ operation, but in this case the user is allowed to specify credentials when testing a connection, which means they could use the Mattermost server to perform a brute force search on someone's else Elasticsearch instance on their behalf.
I think the best thing to do here would be to make PERMISSION_TEST_ELASTICSEARCH a WRITE permission, so only sysadmins and system managers can use it. Thoughts?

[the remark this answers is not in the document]
maxerenberg: And the same reasoning applies to the PERMISSION_TEST_S3 permissions etc.

api4/system.go
```diff
@@ -232,8 +232,8 @@ func testSiteURL(c *Context, w http.ResponseWriter, r *http.Request) {
 		return
 	}
 
-	if !c.App.SessionHasPermissionTo(*c.App.Session(), model.PERMISSION_SYSCONSOLE_WRITE_ENVIRONMENT) && siteURL != *c.App.Config().ServiceSettings.SiteURL {
-		c.SetPermissionError(model.PERMISSION_SYSCONSOLE_READ_ENVIRONMENT)
+	if !c.App.SessionHasPermissionTo(*c.App.Session(), model.PERMISSION_SYSCONSOLE_WRITE_ENVIRONMENT_WEB_SERVER) && siteURL != *c.App.Config().ServiceSettings.SiteURL {
```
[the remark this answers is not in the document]
maxerenberg: Good idea.

api4/config.go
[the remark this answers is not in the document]
hahmadia: In that case, this change makessense.

api4/elasticsearch.go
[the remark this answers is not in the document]
hahmadia: Yes I agree with that change. If the user has write permission to a section, they will always have the read permission too. Please make these a ancillary of the write permissions and remove the duplicated permission checks for both. Thanks @maxerenberg

hahmadia:
When you’ve made the necessary changes, please don’t forget to request a re-review from us. Thank you.

model/config.go
maxerenberg: Yes, I spoke with Maria Nunez. She said it was probably a copy/paste error.

maxerenberg:
@hahmadia 

>     * Could we look at `getClusterStatus , getEnvironmentConfig, testEmail` which use `ENVIRONMENT` permissions but haven't been addressed in this PR. All in the `api4` directory files.

For `getEnvironmentConfig`, we'll have to use the same logic as we do for `getConfig` to make sure that the caller can only read values for the subsections for which they're authorized. This might be a bit messy implementation-wise because we want to omit null fields, but we don't have the `omitempty` tag on any of the Config fields right now.

hahmadia:
> @hahmadia 
> 
> >     * Could we look at `getClusterStatus , getEnvironmentConfig, testEmail` which use `ENVIRONMENT` permissions but haven't been addressed in this PR. All in the `api4` directory files.
> 
> For `getEnvironmentConfig`, we'll have to use the same logic as we do for `getConfig` to make sure that the caller can only read values for the subsections for which they're authorized. This might be a bit messy implementation-wise because we want to omit null fields, but we don't have the `omitempty` tag on any of the Config fields right now.

@maxerenberg 
Could you figure out who uses the grtEnvironmentConfig endpoint anyways? I'm curious how critical it is and where it is used. 

I think we are okay to return null fields in config. Don't we do that already in getconfig? I've noticed when getconfig is called, for sections I don't have permission, I get null fields which would be the same thing in this case. I don't think you have to omit the nulls since we aren't doing that with getconfig.

Let me know, thanks!

app/job.go
```diff
@@ -85,3 +87,27 @@ func (a *App) SessionHasPermissionToCreateJob(session model.Session, job *model.
...
+		model.JOB_TYPE_IMPORT_PROCESS,
+		model.JOB_TYPE_IMPORT_DELETE,
+		model.JOB_TYPE_EXPORT_PROCESS,
+		model.JOB_TYPE_EXPORT_DELETE,
+		model.JOB_TYPE_CLOUD:
+		return a.SessionHasPermissionTo(session, model.PERMISSION_MANAGE_JOBS), model.PERMISSION_MANAGE_JOBS
+	}
+
+	return false, nil
+}
```
sbishel, line 113: I have the same function in my PR. :) Just FYI when we merge.

store/sqlstore/upgrade.go
```diff
@@ -963,6 +963,7 @@ func upgradeDatabaseToVersion532(sqlStore *SqlStore) {
 
 func upgradeDatabaseToVersion533(sqlStore *SqlStore) {
 	if shouldPerformUpgrade(sqlStore, Version5320, Version5330) {
+		sqlStore.AlterColumnTypeIfExists("Roles", "Permissions", "longtext", "text")
```
hahmadia, line 966: You will want to put this in 534 not 533 (5.33 is already out).

model/role.go
```diff
@@ -140,7 +161,21 @@ func init() {
...
+		PERMISSION_SYSCONSOLE_READ_ENVIRONMENT_SMTP.Id,
+		PERMISSION_SYSCONSOLE_READ_ENVIRONMENT_PUSH_NOTIFICATION_SERVER.Id,
+		PERMISSION_SYSCONSOLE_READ_ENVIRONMENT_HIGH_AVAILABILITY.Id,
+		PERMISSION_SYSCONSOLE_READ_ENVIRONMENT_RATE_LIMITING.Id,
+		PERMISSION_SYSCONSOLE_READ_ENVIRONMENT_LOGGING.Id,
+		PERMISSION_SYSCONSOLE_READ_ENVIRONMENT_SESSION_LENGTHS.Id,
+		PERMISSION_SYSCONSOLE_READ_ENVIRONMENT_PERFORMANCE_MONITORING.Id,
+		PERMISSION_SYSCONSOLE_READ_ENVIRONMENT_DEVELOPER.Id,
+		PERMISSION_READ_ELASTICSEARCH_POST_INDEXING_JOB.Id,
+		PERMISSION_READ_ELASTICSEARCH_POST_AGGREGATION_JOB.Id,
```
hahmadia, line 178: You can remove these.
```suggestion
```
Line 171 in this file already gives the ancillary permissions.

model/role.go
```diff
@@ -162,8 +197,44 @@ func init() {
...
+		PERMISSION_TEST_SITE_URL.Id,
+		PERMISSION_RELOAD_CONFIG.Id,
+		PERMISSION_INVALIDATE_CACHES.Id,
+		PERMISSION_RECYCLE_DATABASE_CONNECTIONS.Id,
+		PERMISSION_TEST_ELASTICSEARCH.Id,
+		PERMISSION_CREATE_ELASTICSEARCH_POST_INDEXING_JOB.Id,
+		PERMISSION_CREATE_ELASTICSEARCH_POST_AGGREGATION_JOB.Id,
+		PERMISSION_PURGE_ELASTICSEARCH_INDEXES.Id,
+		PERMISSION_TEST_S3.Id,
+		PERMISSION_TEST_EMAIL.Id,
```
hahmadia, line 237: You can remove these.
```suggestion
```
Line 171 in this file already gives the ancillary permissions.

hahmadia (approved):
Looks amazing dude! Great job and well done on environment config. You've handled it well with the function parameter. I love what I see 👍 👍 

Very minor suggestions but non-blocking otherwise.

store/sqlstore/upgrade.go
maxerenberg: OK.

store/sqlstore/role_store.go
```diff
@@ -91,7 +91,6 @@ func newSqlRoleStore(sqlStore *SqlStore) store.RoleStore {
 		table.ColMap("Name").SetMaxSize(64).SetUnique(true)
 		table.ColMap("DisplayName").SetMaxSize(128)
 		table.ColMap("Description").SetMaxSize(1024)
-		table.ColMap("Permissions").SetMaxSize(4096)
```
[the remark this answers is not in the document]
hahmadia: I'm 3/5 to keep it and remove only the SetMaxSize cause I'm not sure if it will have other implications we aren't aware about.

sbishel (approved):

store/sqlstore/upgrade.go
```diff
@@ -968,7 +968,8 @@ func upgradeDatabaseToVersion533(sqlStore *SqlStore) {
 }
 
 func upgradeDatabaseToVersion534(sqlStore *SqlStore) {
-	// if shouldPerformUpgrade(sqlStore, Version5330, Version5340) {
-	// 	saveSchemaVersion(sqlStore, Version5340)
-	// }
+	//if shouldPerformUpgrade(sqlStore, Version5330, Version5340) {
+	sqlStore.AlterColumnTypeIfExists("Roles", "Permissions", "longtext", "text")
+	//	saveSchemaVersion(sqlStore, Version5340)
+	//}
```
BenCookie95, line 974: Should we just remove the commented code?

BenCookie95 (approved):
One small comment, there are also some conflicts that need to be resolved

store/sqlstore/upgrade.go
hahmadia: @BenCookie95 
This commented code is done on purpose. They will uncomment it when server team upgrades servers to 5.34. We won't want to remove it.

sbishel:
Performed QA on this set of PRs. Test write, read and no access for each subsection and ability to perform ancillary actions.

PR #17057, opened 2021-03-03
[MM-33474] Add m1 support for docker-compose

Willyfrog:
#### Summary

Upgrade docker-compose and makefile to enable local development on the new apple M1's architecture

#### Ticket Link
[MM-33474](https://mattermost.atlassian.net/browse/MM-33474)
#### Release Note
```release-note
NONE
```


Makefile
```diff
@@ -109,6 +110,14 @@ PLUGIN_PACKAGES += mattermost-plugin-nps-v1.1.0
 PLUGIN_PACKAGES += mattermost-plugin-welcomebot-v1.2.0
 PLUGIN_PACKAGES += mattermost-plugin-zoom-v1.5.0
 
+ifneq (Darwin,$(PLATFORM))
+	DOCKER_MAKEFILE := ./docker-compose.makefile.yml
+else ifeq (arm64, $(ARCH))
+    DOCKER_MAKEFILE := ./docker-compose.makefile.m1.yml
+else
+	DOCKER_MAKEFILE := ./docker-compose.makefile.yml
+endif
```
Willyfrog, line 119: I'm open for a better way to write this, it is the only way I found to make the tool happy and still get it to work

docker-compose.makefile.m1.yml
```diff
@@ -0,0 +1,65 @@
...
+        file: ./docker-compose.makefile.yml
+        service: inbucket
+  openldap:
+    container_name: mattermost-openldap
+    extends:
+        file: ./docker-compose.makefile.yml
+        service: openldap
+  elasticsearch:
+    container_name: mattermost-elasticsearch
+    image: "docker.elastic.co/elasticsearch/elasticsearch:7.10.2"
```
nronas, line 31: If my understanding is correct, this is the only difference between the `docker-compose.common.yml` right?

Makefile
nronas: 💅 

```shell
DOCKER_MAKEFILE = ./docker-compose.makefile.yml

ifeq (arm64, $(ARCH))
    DOCKER_MAKEFILE = ./docker-compose.makefile.m1.yml	
endif
```

docker-compose.makefile.m1.yml
Willyfrog: this is one, the other is mysql needs to be run as x86_64 as they don't support arm64 for  the time being

Makefile
Willyfrog: there are other arm64 OSes, not sure if it makes sense to group those as the M1

nronas: Ignore me i thought that was a PLATFORM check as well.

docker-compose.makefile.m1.yml
agnivade: I don't think bumping up the required version to 7 is viable for this. We need to test the code against the minimum supported version, and it needs to work for all platforms. A developer running against 7 in darwin/arm64 will miss out on bugs if they don't test against the right version.

We should look into what are the errors we are getting with 6.5.1 and possibly fix them, or wait for the minimum required version to be bumped to 7.

Willyfrog: if there is a way I can help with that, I'm all in.

but as far as I saw, it was the supported version, when I tried the current one it just kept retrying to connect and no errors in the log

Willyfrog: taking a quick look, seems that [official docker](https://hub.docker.com/_/elasticsearch?tab=tags&page=1&ordering=last_updated) images with arm support start from v7.8.0

if someone creates a Mattermost image of a supported elastic search version with arm support, I can try that on my machine.

mkraft:
Regardless of how we proceed with ElasticSearch, I'm really looking forward to getting at least the `platform: linux/x86_64` MySQL part of this PR merged so that I don't have to continue `git stash`ing that change on my branches whenever I want to use MySQL :)

Willyfrog:
> Regardless of how we proceed with ElasticSearch, I'm really looking forward to getting at least the `platform: linux/x86_64` MySQL part of this PR merged so that I don't have to continue `git stash`ing that change on my branches whenever I want to use MySQL.

agree, maybe I should modify so it doesn't try to start the elasticsearch? or would that cause problems in the server? and have a separate issue for the elastic? is it too complicated to create an arm64 image for docker?

mkraft:
> > Regardless of how we proceed with ElasticSearch, I'm really looking forward to getting at least the `platform: linux/x86_64` MySQL part of this PR merged so that I don't have to continue `git stash`ing that change on my branches whenever I want to use MySQL.
> 
> agree, maybe I should modify so it doesn't try to start the elasticsearch? or would that cause problems in the server? and have a separate issue for the elastic? is it too complicated to create an arm64 image for docker?

I don't use the ElasticSearch Docker image with Mattermost in development, so it wouldn't affect me if that were a separate PR dealt with at a later date.

Willyfrog:
@agnivade @mkraft I've removed the customisation for elasticsearch until a proper solution can be reached, but at least people with an M1 can use docker-compose with mysql.

Makefile
agnivade, line 119: Nit: spacing is not uniform.

agnivade (approved):
Looks good to me. I am wondering if it's possible to take the base makefile and apply just the x86 patch on top, and avoid duplication in two places?

Makefile
Willyfrog: weird seems one is a tab and the other are spaces, probably something wrong in my config, will fix it

Willyfrog:
> Looks good to me. I am wondering if it's possible to take the base makefile and apply just the x86 patch on top, and avoid duplication in two places?

I thought about it, but given docker compose and how it works seemed easy to understand (a patch file is always uglier to read) and we can iterate on this approach with other architectures in the future (if there are any new ones)

nronas (approved):
Nice!

Willyfrog:
sorry, I'm not really sure why the tests are failing, any hints? I don't think I've changed anything related to sentry

mkraft:
> sorry, I'm not really sure why the tests are failing, any hints? I don't think I've changed anything related to sentry

I'm getting the same error on a different PR: https://github.com/mattermost/mattermost-server/pull/17115

sapkra:
Any update on this one?

Willyfrog:
I'm considering this change is not needed anymore, right now docker works without any issue, the only problematic one is elasticsearch and it's just a matter of removing the lines that start that container for it to work

PR #17062, opened 2021-03-03
Fix initialism errors (PR-3) 

haardikdharma10:
#### Summary

For #16623 partially (This is PR-3)

#### Ticket Link

#16623  

#### Release Note

```release-note
NONE
```


app/bot.go
```diff
@@ -364,7 +364,7 @@ func (a *App) disableUserBots(userID string) *model.AppError {
 
 		// Get next set of bots if we got the max number of bots
 		if len(userBots) == perPage {
-			options.Page += 1
+			options.Page++
```
agnivade, line 367: While this is good to do, but it does not give much gain in readability. Let's revert these changes.

plugin/client_rpc.go
```diff
@@ -610,25 +610,25 @@ func (s *hooksRPCServer) MessageWillBeUpdated(args *Z_MessageWillBeUpdatedArgs,
 	return nil
 }
 
-type Z_LogDebugArgs struct {
+type ZLogDebugArgs struct {
```
agnivade, line 613: @crspeller - Will there be any issues with changing these exported structs? Are there consumers depending on these?

app/bot.go
haardikdharma10: Reverting it back to original will cause the lint error. Is it fine?

plugin/client_rpc.go
crspeller: I think it should be fine. I think these are only exported for serialization (the reflect limitation). 
But I could be wrong. Someone should test to make sure plugins work on this branch. Break should be super obvious. 

Also a reminder that while the changes to this file will stick. The changes in  `plugin/client_rpc_generated.go` will not as that file is generated by `interface_generator`. That generator will have to be modified.

app/bot.go
agnivade: What lint error are you getting for this?

plugin/client_rpc.go
agnivade: Oh wow, that's a great catch, @haardikdharma10 did you manually modify each and every line of `plugin/client_rpc_generated.go`? I totally missed the fact that `plugin/interface_generator/main.go` is not touched at all :facepalm: 

Let's change that and then generate the file.

haardikdharma10: Yes @agnivade! It was as simple as replacing all `Z_` with `Z`. However, I see that changing [this line](https://github.com/mattermost/mattermost-server/blob/master/plugin/interface_generator/main.go#L473) in `plugin/interface_generator/main.go` is all I have to do here. I have replaced it from `Z_ + name` to `Z + name`. How do i generate the `plugin/client_rpc_generated.go` file now?

crspeller: @haardikdharma10 `make pluginapi` should do it.

haardikdharma10: @crspeller can't figure out what's with the failing check? 🤔

agnivade:
@jupenur - Is this a known failure https://app.circleci.com/pipelines/github/mattermost/mattermost-server/22418/workflows/fb29d59a-df05-49c9-8c86-75ae26f000cb/jobs/138831 ?

jupenur:
@agnivade yeah, see https://community.mattermost.com/core/pl/14pjeddp4jgg8n63kpsbogcj8e

agnivade:
Oh my bad, I thought that was a different one.

/update-branch

isacikgoz (approved):

haardikdharma10:
Is this good enough to be merged? 

cc @agnivade, @isacikgoz

model/client4.go
```diff
@@ -4148,7 +4148,7 @@ func (c *Client4) GetGroups(opts GroupSearchOpts) ([]*Group, *Response) {
 }
 
 // GetGroupsByUserId retrieves Mattermost Groups for a user
-func (c *Client4) GetGroupsByUserId(userId string) ([]*Group, *Response) {
+func (c *Client4) GetGroupsByUserID(userId string) ([]*Group, *Response) {
```
agnivade, line 4151: Please revert this. Anything in the model package is off limits :P

agnivade (changes requested):

model/client4.go
agnivade: You need to change the call sites too.

agnivade (approved):

haardikdharma10:
All checks are passing 🎉 

@agnivade, @isacikgoz

PR #17071, opened 2021-03-04
Turn feature flag on by default for custom statuses

esethna:


crspeller (changes requested):
From my perspective, if the feature is validated on cloud already, the feature flag should be removed.

jespino (approved):
Lgtm

PR #17079, opened 2021-03-05
set CSP unsafe-inline during development

lieut-data:
#### Summary
Extensions in Firefox are hampered by a [long-standing bug](https://bugzilla.mozilla.org/show_bug.cgi?id=1267027) that (incorrectly) applies CSP to content scripts injected by extensions. This precludes the ability to use the React and Redux DevTools in Firefox.

When in dev mode, add `unsafe-inline` to the CSP directives to unlock the use of these extdensions.

#### Ticket Link
None.

#### Release Note
```release-note
NONE
```

### DIFF
--- a/web/handlers.go
+++ b/web/handlers.go
@@ -177,7 +177,13 @@ func (h Handler) ServeHTTP(w http.ResponseWriter, r *http.Request) {
 		// Add unsafe-eval to the content security policy for faster source maps in development mode
 		devCSP := ""
 		if model.BuildNumber == "dev" {
-			devCSP = " 'unsafe-eval'"
+			devCSP += " 'unsafe-eval'"
+		}
+
+		// Add unsafe-inline to unlock extensions like React & Redux DevTools in Firefox
+		// see https://github.com/reduxjs/redux-devtools/issues/380#issuecomment-329085853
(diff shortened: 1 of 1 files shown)


lieut-data:
Skipping QA Review as this only affects developer builds.

jupenur (approved):

web/handlers.go
```diff
@@ -177,7 +177,13 @@ func (h Handler) ServeHTTP(w http.ResponseWriter, r *http.Request) {
 		// Add unsafe-eval to the content security policy for faster source maps in development mode
 		devCSP := ""
 		if model.BuildNumber == "dev" {
-			devCSP = " 'unsafe-eval'"
+			devCSP += " 'unsafe-eval'"
+		}
+
+		// Add unsafe-inline to unlock extensions like React & Redux DevTools in Firefox
+		// see https://github.com/reduxjs/redux-devtools/issues/380#issuecomment-329085853
```
lieut-data, line 184:
```suggestion
 		// see https://github.com/reduxjs/redux-devtools/issues/380
```

hmhealey (approved):
Oh, so that's why the React dev tools never worked for me in Firefox. I didn't realize it was something specific to how Mattermost ran

PR #17084, opened 2021-03-08
Enable custom statuses feature flag by default

jespino:
#### Summary
Enable custom statuses feature flag by default

```release-note
NONE
```

### DIFF
--- a/model/feature_flags.go
+++ b/model/feature_flags.go
@@ -31,7 +31,7 @@ func (f *FeatureFlags) SetDefaults() {
 	f.TestBoolFeature = false
 	f.CloudDelinquentEmailJobsEnabled = false
 	f.CollapsedThreads = false
-	f.CustomUserStatuses = false
+	f.CustomUserStatuses = true
 	f.PluginIncidentManagement = "1.4.0"
 }
 


jespino:
@amyblais this is the change directly to the release-5.33 branch.

amyblais:
Will there also be a PR for cloud branch for March 10th Cloud release?

jespino:
@amyblais that probably is a question for @esethna If you want I can create one, for sure.

esethna:
@amyblais @jespino yes we should enable the feature in cloud on March 10th as well.

esethna (approved):

esethna:
@amyblais @jespino perhaps add different reviewers if this needs to go in today as Eli won't be around till tomorrow

jespino:
@esethna @amyblais I think that is not needed for the cloud, because the cloud is though for enabling/disbling this things on the fly using their system. so... I shouldn't be necessary.

esethna:
@jespino makes sense. @amyblais can we verify that the feature is enabled for the cloud march 10 release?

lieut-data (approved):
I'm not an expert here, but this seems consistent with how we've modified this value for the incident collaboration plugin.

But I had been under the impression, @crspeller, that we were going to /remove/ feature flags before they shipped on-premise, having validated them in cloud. Or has the process changed?

crspeller (approved):
@lieut-data This is actually process breaking. You are correct, the process is to remove the feature flag to ship to on-prem having been validated in cloud first. The feature flag has already been removed from master. This change only effects the 5.33 release branch.

The explanation I have been given is that this is an urgent feature that must be released immediately. Also we are in a state where we are confident enough in the feature to release it to on-prem customers, but not confident enough to release it to cloud without the feature flag. 
I have explained, and the release team understands that feature flags don't work on-prem and the only recourse would be to turn the feature off manually (something cloud customers can also do).

PR #17099, opened 2021-03-10
MM-33708 - Add MentionCountRoot column to ChannelMembers

reflog:
#### Summary

added new column for root-only mentions
added migration to populate that column based on ThreadMembership table
removed redundant API call

#### Ticket Link

https://mattermost.atlassian.net/browse/MM-33708

#### Release Note

```release-note
Added new column in ChannelMembers table called 'MentionCountRoot'. Please not that on installations with large amount of Channels/Users the migration can take up to a few minutes (worst case)
```


reflog:
@streamer45 - I'm tagging you for review, since this can possibly affect performance

store/sqlstore/upgrade.go
```diff
@@ -969,6 +969,8 @@ func upgradeDatabaseToVersion533(sqlStore *SqlStore) {
 
 func upgradeDatabaseToVersion534(sqlStore *SqlStore) {
 	// if shouldPerformUpgrade(sqlStore, Version5330, Version5340) {
+	sqlStore.CreateColumnIfNotExists("ChannelMembers", "MentionCountRoot", "bigint", "bigint", "0")
```
jespino, line 972: Why are you adding the default value?, that can generate some problems with the execution of the migration, adding a NULL column should work fine, adding a column with a default value, I guess it can generate problems. Can we confirm that?

Probably this is also generating the discrepancy between the migration and the fresh schema.

Another thing. Can we use the new migrations system? // @nronas

reflog: @jespino - I don't think having the default value as NULL is a good idea... @streamer45 - can you weigh in on ALTER with default vs ALTER without a default value in terms of performance?

reflog: @jespino - I'll move to NULL be default (will help with lazy migration anyways)

app/channel.go
```diff
@@ -1899,7 +1899,15 @@ func (a *App) GetChannelUnread(channelID, userID string) (*model.ChannelUnread,
...
 	}
-
+	// if we don't have MentionCountRoot yet, calculate it now and retry
+	if channelUnread.MentionCountRoot == nil {
+		err := a.populateMentionCountRoot(channelUnread.TeamId, userID)
+		if err != nil {
+			return nil, model.NewAppError("GetChannelUnread", "app.channel.get_unread.app_error", nil, err.Error(), http.StatusInternalServerError)
+		}
+		// retry
+		return a.GetChannelUnread(channelID, userID)
```
zefhemel, line 1909: This recursive self-call behavior seems potentially dangerous, because it could lead to infinite recursion in case `populateMentionCountRoot` never assigns a non-nil value to MentionCountRoot. I looked at `populateMentionCountRoot` and it seems there's only a few places where a value is assigned, which are nested in some ifs and fors.

Are you absolutely sure that there's no edge case in which `channelUneread.MentionCountRoot` remains `nil` even after an successful populateMentionCountRoot call?

app/channel.go
```diff
@@ -3030,3 +3051,116 @@ func (a *App) getDirectChannel(userID, otherUserID string) (*model.Channel, *mod
 
 	return channel, nil
 }
+
+func (a *App) isUserMentionedInPost(team *model.Team, channel *model.Channel, post *model.Post, user *model.User, channelMemberNotifyPropsMap map[string]model.StringMap) (bool, error) {
```
zefhemel, line 3055: This is method that looks like it's a simple check, but in fact has a whole bunch of side effects (it updates things in passed arguments). Is this an accepted pattern in our codebase? It sounds like potentially confusing (generally I'd hope an `isX` style method would have no side effects).

app/team.go
```diff
@@ -1137,13 +1137,32 @@ func (a *App) GetTeamUnread(teamID, userID string) (*model.TeamUnread, *model.Ap
...
 	var teamUnread = &model.TeamUnread{
-		MsgCount:     0,
-		MentionCount: 0,
-		TeamId:       teamID,
+		MsgCount:         0,
+		MentionCount:     0,
+		MentionCountRoot: 0,
+		TeamId:           teamID,
+	}
+	migrationNeeded := false
```
zefhemel, line 1145: Would it not be cleaner to extract any migration code (which will live forever, and in the future there may be other reasons for migrations to be added) into a separate (private) method?

model/client4.go
```diff
@@ -5885,20 +5885,6 @@ func (c *Client4) DownloadExport(name string, wr io.Writer, offset int64) (int64
 	return n, BuildResponse(r)
 }
 
-func (c *Client4) GetThreadMentionsForUserPerChannel(userId, teamId string) (map[string]int64, *Response) {
```
zefhemel, line 5888: This is effectively our public API right? Can this not break external systems (plugins?) relying on this API call?

reflog:
@jespino @zefhemel FYI I don't like this code. I think the migration method is too heavy to be called lazily and evaluated when the user enters a channel. I feel like this would better be served by a migration task

model/client4.go
reflog: this is an API specifically added for collapsed threads and is not used by anyone else but us @zefhemel

app/channel.go
reflog: @zefhemel - not sure what you mean. This method has no side effects and doesn't update any of the parameters. (unless I missed something)

app/channel.go
reflog: @zefhemel I'll rewrite it as non-recursive

app/team.go
reflog: @zefhemel not sure what you mean. Actual migration is encapsulated in populateMentionCountRoot

app/channel.go
zefhemel: Somehow I thought I saw you were changing state, but now looking at it again you're not. Ignore me.

app/team.go
zefhemel: I mean this line up to 1162 where you track whether migration is required and apply the migration. My suggestion was to extract this code into into a `migrateTeamUnread` method or whatever to keep this method somewhat short. Not a big deal, just an idea.

zefhemel (approved):
I went through this and it looks reasonable. I don't like that we have to litter the code with all this on-the-fly migration code, code that will likely live forever, but... the alternative (a one-time migration) is also not practical. 

My knowledge of the code base is not sufficient to make judgement about code quality, level of testing and if you touched all aspects you need to touch. But assuming that all checks out, LGTM.

reflog:
@streamer45 @jespino from my basic testing, the performance of populateMentionCountRoot on a database with 10M posts is abysmal. No way we can use this in a 'live' lazy migration.
(several things cause this - `GetPostsSince` is not paged, so if a user didn't visit a popular channel in a long while - there can be hundreds of thousands of messages, `isUserMentionedInPost` inside is calling an uncached GetChannelMember... for each post)
So unless I am missing something obvious - this cannot be a live migration

jespino:
@reflog the question is really about "where the time goes and why?" maybe we can improve the performance of that by getting the channel members + the list of posts, or get the list of posts and pre-fetch the channel members for that list of posts, that way you have a paginated approach for that.

reflog:
@jespino - on a large instance it won't matter.
if the number of 'unread' posts is large - there's no way to improve this, we need to scan each post for mentions.

streamer45:
That backup was probably a bit of an extreme case since all posts were concentrated in a few channels. I can help with some additional tests on a more realistic distribution to see if anything improves (and by how much).

streamer45:
Done a couple more tests and it's clear that this approach is not going to work on anything that has significant amount of pre-existing data. `populateMentionCountRoot` is just doing way too much work to be feasible on a live system. 

I am not even sure this could be acceptable as a synchronous migration, as it could take a very long time to complete for large installs. 

I guess an option we are left with would be to have a background (async) migration job although even then we'd have to be careful and tune it properly so that it wont cause too much load on the live system. 

Also, from a correctness perspective, would this logic have to run in a transaction? Posts can still be made while this is running and final counts could be incorrect.

cc @reflog

reflog:
@streamer45 thanks for confirming my suspicion. I'm scrapping the migration bit. There's a possible way around this without actually parsing the posts, so I'm throwing away all of this 'live migration stuff' and writing a query to update the column using existing data from ThreadMembership table.
Should be done today

reflog:
@jespino @streamer45  - please re-review. lazy migration removed

api4/channel_test.go
```diff
@@ -4063,3 +4064,43 @@ func TestMoveChannel(t *testing.T) {
 		require.Equal(t, team2.Id, newChannel.TeamId)
 	}, "Should be able to (force) move private channel by a member that is not member of target team")
 }
+
+func TestRootMentionsMigrations(t *testing.T) {
```
jespino, line 4068: This test name doesn't look like what the test is actually testing. I think this should be something like `TestRootMentionsCount`.

app/channel_test.go
```diff
@@ -1343,15 +1343,25 @@ func TestMarkChannelAsUnreadFromPost(t *testing.T) {
...
+			UserId:    u2.Id,
+			ChannelId: c2.Id,
+			RootId:    p4.Id,
+			CreateAt:  model.GetMillis() + 2000,
+			Message:   "@" + u1.Username,
+		}, c2, false, true)
+
 		response, err := th.App.MarkChannelAsUnreadFromPost(p4.Id, u1.Id)
 		assert.Nil(t, err)
 		assert.Equal(t, int64(1), response.MsgCount)
```
jespino, line 1356: Shouldn't be 2?

reflog: @jespino surprisingly - no :)
MsgCount = TotalMsgCount - UnreadCount = 3 - 2 = 1

store/sqlstore/channel_store.go
```diff
@@ -38,32 +38,34 @@ type SqlChannelStore struct {
...
-		SchemeGuest:  sql.NullBool{Valid: true, Bool: cm.SchemeGuest},
-		SchemeUser:   sql.NullBool{Valid: true, Bool: cm.SchemeUser},
-		SchemeAdmin:  sql.NullBool{Valid: true, Bool: cm.SchemeAdmin},
+		ChannelId:        cm.ChannelId,
+		UserId:           cm.UserId,
+		Roles:            cm.ExplicitRoles,
+		LastViewedAt:     cm.LastViewedAt,
+		MsgCount:         cm.MsgCount,
+		MentionCount:     cm.MentionCount,
+		MentionCountRoot: sql.NullInt64{Valid: true, Int64: cm.MentionCountRoot},
```
jespino, line 63: If this is now done through a migration, it is still needed the `NullInt64`?

store/sqlstore/upgrade.go
```diff
@@ -1000,6 +1000,30 @@ func upgradeDatabaseToVersion533(sqlStore *SqlStore) {
...
 	// if shouldPerformUpgrade(sqlStore, Version5330, Version5340) {
+	sqlStore.CreateColumnIfNotExistsNoDefault("ChannelMembers", "MentionCountRoot", "bigint", "bigint")
+	updateMentionCountRootQuery := `
+	UPDATE ChannelMembers INNER JOIN (
+		SELECT Threads.ChannelId,COALESCE(threadmemberships.UserId,'') AS UserId, COALESCE(SUM(UnreadMentions),0) AS UnreadMentions FROM Channels
+		LEFT JOIN Threads ON Threads.ChannelId = Channels.Id
+		LEFT Join ThreadMemberships ON Threads.PostId = ThreadMemberships.PostId
+		Group BY Threads.ChannelId,ThreadMemberships.UserId
+	) AS q ON q.ChannelId = ChannelMembers.ChannelId AND (q.UserId=ChannelMembers.UserId OR q.UserId='')
+	SET MentionCountRoot = GREATEST(ChannelMembers.MentionCount - q.UnreadMentions, 0)`
```
streamer45, line 1011: I feel that this query is just too heavy to be part of a synchronous migration process. I am yet to see this return on our biggest MySQL dataset. I imagine it could take hours to execute given the amount of data involved in the subquery.

streamer45: Purely for curiosity this has finally returned:

```
mysql> UPDATE ChannelMembers
    -> INNER JOIN
    ->   (SELECT Threads.ChannelId,
    ->           COALESCE(ThreadMemberships.UserId, '') AS UserId,
    ->           COALESCE(SUM(UnreadMentions), 0) AS UnreadMentions
    ->    FROM Channels
    ->    LEFT JOIN Threads ON Threads.ChannelId = Channels.Id
    ->    LEFT JOIN ThreadMemberships ON Threads.PostId = ThreadMemberships.PostId
    ->    GROUP BY Threads.ChannelId,
    ->             ThreadMemberships.UserId) AS q ON q.ChannelId = ChannelMembers.ChannelId
    -> AND (q.UserId=ChannelMembers.UserId
    ->      OR q.UserId='')
    -> SET MentionCountRoot = GREATEST(ChannelMembers.MentionCount - q.UnreadMentions, 0);
Query OK, 3257748 rows affected (3 hours 2 min 16.88 sec)
Rows matched: 3257748  Changed: 3257748  Warnings: 0
```

jespino (approved):
LGTM! Thanks @reflog! Amazing work :)

streamer45 (approved):
Migration code looks good. Similarly to the other PR I'd be expecting an important upgrade note regarding this potentially taking a few minutes on MySQL installations with big `ChannelMembers` tables (in the order of millions of rows).

Also ran a complete load-test comparison on both DB drivers which didn't reveal any issue. 

Approving from a performance perspective :+1:

reflog:
@ogi-m please smoke test around posting/replying

ogi-m (approved):
LGTM

PR #17107, opened 2021-03-10
[MM-29245] Add Billing SubSection

hahmadia:
#### Summary

Add Billing SubSections

#### Ticket Link

https://mattermost.atlassian.net/browse/MM-29245

```release-note
None
```

### **HUGE THANK YOU** to Maria from the Cloud team for helping me out with figuring out these permissions.


ethervoid:
@hahmadia One question, this PR has more changes than the billing ones, for example, bleave. Is that correct?

hahmadia:
> @hahmadia One question, this PR has more changes than the billing ones, for example, bleave. Is that correct?

@ethervoid 

This PR is only focused on billing section.
However we have other PRs merged into our feature branch that focus on those other sections like Bleve.

To see the whole list of PRs
, You can take a look at https://github.com/mattermost/mattermost-server/pull/16917

hahmadia:
**UPDATE**

We had a meeting regarding this particular section and it was decided that we will leave it as it is for the following two reasons:

1. The interdependency of sections makes things a bit more complicated and not something we need to address at this very moment since customers haven't brought it up or asked for it.
2. Maria mentioned that they are still in the early stages of the billing section and have plans to change things which means that our effort for this could potentially go away.

Until required, we will leave Billing as it is and come back in the future once it's in a more stable position.

PR #17119, opened 2021-03-12
MM-33789: Revert fallback to master for GetAllProfilesInChannel

agnivade:
This fixes a regression introduced in https://github.com/mattermost/mattermost-server/pull/16911.

It was causing problems with too many invalidations and overloading the writer instance for big installations.
Reverting this does not affect correctness at all because it was done out of abundance of caution and the
idea at that point was it was to be done for all caches.

https://mattermost.atlassian.net/browse/MM-33789

```release-note
NONE
```

### DIFF
--- a/store/localcachelayer/layer.go
+++ b/store/localcachelayer/layer.go
@@ -292,7 +292,6 @@ func NewLocalCacheLayer(baseStore store.Store, metrics einterfaces.MetricsInterf
 		UserStore:                      baseStore.User(),
 		rootStore:                      &localCacheStore,
 		userProfileByIdsInvalidations:  make(map[string]bool),
-		profilesInChannelInvalidations: make(map[string]bool),
 	}
 
 	// Teams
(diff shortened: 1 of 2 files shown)


agnivade:
@srkgupta - Would you be able to QA this. The steps are essentially the same as https://github.com/mattermost/mattermost-server/pull/16911. But that PR is causing some performance regressions which this PR fixes. We just want to confirm that the issue is still fixed.

streamer45 (approved):
Thanks! Looks good :+1: 

Re-run a load-test to confirm and everything is back to normal:

image

srkgupta (changes requested):
Hi @agnivade 

This code revert is now causing the issue in the ticket MM-30882 to re-appear. Can you please take a look at it and fix it.

srkgupta (approved):
There were some issues with the code deployed on the previous HA instance here. I redeployed the same on a cloud test HA instance and retested and the issue worked fine. Approving the PR. Thanks @agnivade

PR #17124, opened 2021-03-12
[MM-33603] Fix nil dereference panic in (*App).CreatePost()

streamer45:
#### Summary

PR fixes a possible panic due to not handling the error returned by `GetThreadForUser`.

This **doesn't** fix the possible underlying issue of why that might be erroring out which I think might be worth an additional ticket. 

I am wondering if `thread.Participants` can possibly diverge throughout the application lifetime (e.g. a post/user in a thread getting deleted). Another cause for `GetThreadForUser` to return error could be replica lag although it would be rather unlikely given a previous call to `Store.Thread().Get()` succeeded.

#### Ticket

https://mattermost.atlassian.net/browse/MM-33603

#### Release Note

```release-note
Fixed a possible panic on post creation when collapsed threads are enabled.
```


app/notification.go
```diff
@@ -434,7 +434,10 @@ func (a *App) SendNotifications(post *model.Post, team *model.Team, channel *mod
 			}
 			if sendEvent {
 				message := model.NewWebSocketEvent(model.WEBSOCKET_EVENT_THREAD_UPDATED, team.Id, "", uid, nil)
-				userThread, _ := a.Srv().Store.Thread().GetThreadForUser(uid, channel.TeamId, thread.PostId, true)
+				userThread, err := a.Srv().Store.Thread().GetThreadForUser(uid, channel.TeamId, thread.PostId, true)
+				if err != nil {
+					return nil, errors.Wrapf(err, "cannot get thread %q for user %q", thread.PostId, uid)
+				}
 				a.sanitizeProfiles(userThread.Participants, false)
```
agnivade, line 441: There also seems to be a case where it returns `nil, nil`

```
if !thread.Following {
	return nil, nil // in case the thread is not followed anymore - return nil error to be interpreted as 404
}
```

Should we be handling that?

streamer45: Ouch. Most definitely. I implicitly thought it was respecting the common pattern. Nice catch :+1:

streamer45: @reflog Do you want this case handled as an error or should we simply pass through?

reflog: @streamer45 - no, this is an actual bug. If user is not following the thread, we should send nothing. I'll add a fix to your PR if you don't mind

agnivade (approved):
Do we want to create the ticket to investigate the underlying cause?

reflog (approved):

streamer45:
> Do we want to create the ticket to investigate the underlying cause?

Yes, I can make one.

streamer45:
@lindalumitchell  
This case is covered by a unit test so I think it can be safely deferred.

lindalumitchell (approved):

PR #17125, opened 2021-03-12
MM-33756 fix client crashes on sync (post sync order)

wiggin77:
#### Summary
This PR fixes a crash in the client after a sync when edited/deleted posts are included in a certain order.

- don't send the deleted post that is just a copy of an original edited post
- detect true deleted post and use delete api instead of update to get "message deleted" on client
- make sending message for sync a synchronous call so next sync cannot over-take
- add logging levels for logging message traffic

**To be merged with feature branch.**

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-33756

#### Release Note
```release-note
NONE
```

### DIFF
--- a/services/sharedchannel/sync_recv.go
+++ b/services/sharedchannel/sync_recv.go
@@ -24,7 +24,7 @@ func (scs *Service) onReceiveSyncMessage(msg model.RemoteClusterMsg, rc *model.R
 	}
 
 	if scs.server.GetLogger().IsLevelEnabled(mlog.LvlSharedChannelServiceMessagesInbound) {
-		scs.server.GetLogger().Log(mlog.LvlSharedChannelServiceMessagesOutbound, "inbound message",
+		scs.server.GetLogger().Log(mlog.LvlSharedChannelServiceMessagesInbound, "inbound message",
 			mlog.String("remote", rc.DisplayName),
 			mlog.String("msg", string(msg.Payload)),
 		)
--- a/services/sharedchannel/sync_recv.go
+++ b/services/sharedchannel/sync_recv.go
@@ -247,6 +247,13 @@ func (scs *Service) upsertSyncPost(post *model.Post, channel *model.Channel, rc 
 			mlog.String("post_id", post.Id),
 			mlog.String("channel_id", post.ChannelId),
 		)
+	} else if post.DeleteAt > 0 {
+		// delete post
+		rpost, appErr = scs.app.DeletePost(post.Id, post.UserId)
+		scs.server.GetLogger().Log(mlog.LvlSharedChannelServiceDebug, "Deleted sync post",
+			mlog.String("post_id", post.Id),
+			mlog.String("channel_id", post.ChannelId),
+		)
 	} else if post.EditAt > rpost.EditAt || post.Message != rpost.Message {
 		// update post
 		rpost, appErr = scs.app.UpdatePost(post, false)
--- a/services/sharedchannel/sync_recv.go
+++ b/services/sharedchannel/sync_recv.go
@@ -23,6 +23,13 @@ func (scs *Service) onReceiveSyncMessage(msg model.RemoteClusterMsg, rc *model.R
 		return errors.New("empty sync message")
 	}
 
+	if scs.server.GetLogger().IsLevelEnabled(mlog.LvlSharedChannelServiceMessagesInbound) {
+		scs.server.GetLogger().Log(mlog.LvlSharedChannelServiceMessagesOutbound, "inbound message",
+			mlog.String("remote", rc.DisplayName),
+			mlog.String("msg", string(msg.Payload)),
+		)
+	}
+
 	var syncMessages []syncMsg
 
 	if err := json.Unmarshal(msg.Payload, &syncMessages); err != nil {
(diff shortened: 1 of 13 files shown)


store/sqlstore/post_store.go
```diff
@@ -942,14 +942,14 @@ func (s *SqlPostStore) GetPostsSinceForSync(options model.GetPostsSinceForSyncOp
 		Where(sq.GtOrEq{"UpdateAt": options.Since}).
 		Where(sq.Eq{"ChannelId": options.ChannelId}).
 		Limit(uint64(options.Limit)).
-		OrderBy("UpdateAt"+order, "Id")
+		OrderBy("CreateAt"+order, "DeleteAt", "Id")
```
nronas, line 945: I've run explains for both the previous implementation and the new one, and they are identical since there are sufficient indexes to support the new ordering and filtering:

```sql
mattermost_test=# EXPLAIN ANALYSE select * from posts where updateat >= 1614935961180 and channelid = 'jdnjbpjt6jbomgk3n9y9p1n8or' order by updateat, id limit 1000;
                                                                          QUERY PLAN
-------------------------------------------------------------------------------------------------
 Limit  (cost=8.17..8.17 rows=1 width=2906) (actual time=0.086..0.152 rows=2 loops=1)
   ->  Sort  (cost=8.17..8.17 rows=1 width=2906) (actual time=0.074..0.107 rows=2 loops=1)
         Sort Key: updateat, id
         Sort Method: quicksort  Memory: 25kB
         ->  Index Scan using idx_posts_channel_id_delete_at_create_at on posts  (cost=0.14..8.16
               Index Cond: ((channelid)::text = 'jdnjbpjt6jbomgk3n9y9p1n8or'::text)
               Filter: (updateat >= '1614935961180'::bigint)
 Planning time: 0.250 ms
 Execution time: 0.214 ms
(9 rows)

mattermost_test=# EXPLAIN ANALYSE select * from posts where updateat >= 1614935961180 and channelid = 'jdnjbpjt6jbomgk3n9y9p1n8or' order by createat desc, deleteat, id limit 1000;
                                                                          QUERY PLAN
-------------------------------------------------------------------------------------------------
 Limit  (cost=8.17..8.17 rows=1 width=2906) (actual time=0.077..0.136 rows=2 loops=1)
   ->  Sort  (cost=8.17..8.17 rows=1 width=2906) (actual time=0.065..0.092 rows=2 loops=1)
         Sort Key: createat DESC, deleteat, id
         Sort Method: quicksort  Memory: 25kB
         ->  Index Scan using idx_posts_channel_id_delete_at_create_at on posts  (cost=0.14..8.16
               Index Cond: ((channelid)::text = 'jdnjbpjt6jbomgk3n9y9p1n8or'::text)
               Filter: (updateat >= '1614935961180'::bigint)
 Planning time: 0.157 ms
 Execution time: 0.197 ms
(9 rows)
```

services/sharedchannel/sync_send.go
```diff
@@ -329,10 +330,22 @@ func (scs *Service) updateForRemote(task syncTask, rc *model.RemoteCluster) erro
...
+			mlog.String("remote", rc.DisplayName),
+			mlog.Int64("NextSyncAt", scr.NextSyncAt),
+			mlog.String("msg", string(b)),
+		)
+	}
+
 	ctx, cancel := context.WithTimeout(context.Background(), remotecluster.SendTimeout)
 	defer cancel()
 
+	var wg sync.WaitGroup
```
nronas, line 344: So in order for this `wg` to be released the callback function has to be called. So if in `SendMsg` we get a panic, we will `recover()` but we will never call the callback function, essentially deadlocking the caller, right?

Now, the `sendFrameToRemote` looks pretty safe but just highlighting this in case that we are missing anything.

store/sqlstore/post_store.go
streamer45: Nice @nronas :+1:  Out of curiosity, do you know if MySQL would choose similar strategies?

nronas: Yeap, i think so too, assuming that the same indexes are in place, which they are as far as i can see

services/sharedchannel/sync_send.go
wiggin77: Correct, we need the callback called, which is pretty safe IMO.  However I refactored slightly to force the callback even on a panic.  https://github.com/mattermost/mattermost-server/pull/17125/commits/0e96676678c5dd369f44f448a6415c3943dc00f0

nronas (approved):
LGTM

services/sharedchannel/sync_recv.go
```diff
@@ -240,16 +247,23 @@ func (scs *Service) upsertSyncPost(post *model.Post, channel *model.Channel, rc
...
 		)
-	} else if rpost.DeleteAt == 0 {
+	} else if post.DeleteAt > 0 {
+		// delete post
+		rpost, appErr = scs.app.DeletePost(post.Id, post.UserId)
+		scs.server.GetLogger().Log(mlog.LvlSharedChannelServiceDebug, "Deleted sync post",
+			mlog.String("post_id", post.Id),
+			mlog.String("channel_id", post.ChannelId),
+		)
+	} else if post.EditAt > rpost.EditAt || post.Message != rpost.Message {
```
streamer45, line 257: Interesting. Are we being extra cautious or is it actually possible for the content to have changed and not `EditAt`?

streamer45 (approved):
Looks good :+1:

services/sharedchannel/sync_recv.go
wiggin77: Just being cautious.  I hit so many corner cases debugging this stuff that I decided not to trust any contracts.

PR #17134, opened 2021-03-15
Some small improvements in performance

jespino:
#### Summary
Some small improvements in performance extracted from the spike in cockroachdb.

The main changes here is about a better usage of the emojis cache and add the usage of the replyCount denormalized data stored in the new Threads table. This should speed up a bit the performance on posts retrieval.

#### Ticket Link
Not yet

#### Release Note
```release-note
NONE
```


agnivade (commented):
Thank you for the PR Jesus. Can we split this into three separate PRs - emojis, posts and assert->require change? It will be helpful to review and benchmark them independently.

For example, the assert -> require can go in right away and does not need to block the other stuff.

Thanks!

jespino:
@agnivade Sure, let me split that, I see the emojis like a desirable improvement without the need of a loadtest, the asserts->require something quick and simple, but the interesting part is about the denormalization of the ReplyCount.

PR #17142, opened 2021-03-15
MM-33836: Detect and upgrade incorrect HTTP version for websocket handshakes

agnivade:
Our proxy configuration was historically incorrect, due to which
a lot of customers have that in their setups. As a result, strictly
following the websocket RFC results in a breaking change.

For now, we transparently upgrade the version header to 1.1, if we detect 1.0.
If a client was sending 1.0, it wouldn't have worked anyways because persistent
connections were introduced from 1.1 onwards.

https://mattermost.atlassian.net/browse/MM-33836

```release-note
WebSocket handshakes done with HTTP version lower than 1.1 will result in a warning,
and the server will transparently upgrade the version to 1.1 to comply with the
websocket RFC.

This is done to work around incorrect nginx (and other proxy) configs that do not set
the proxy_http_version directive to 1.1.

This facility will be removed in a future Mattermost version and it is strongly recommended
to fix the proxy configuration to correctly use the websocket protocol.
```


agnivade:
Tested locally with nginx and verified that it works correctly as expected.

amyblais:
@agnivade Does this also mean that the breaking change doesn't have to be announced in the blog post / forum (e.g. this will now just be an upgrade note)?

streamer45 (approved):
Awesome :+1:

agnivade:
@amyblais - You are fast :) I was trying to find the PR that changed the release notes.

Yes, we will need to take down the announcement made in the forum, and edit the release notes to remove the bit about the breaking change and change it to what is mentioned in this PR. Thank you for your patience, and sorry for the double work.

amyblais:
@agnivade @streamer45 I've updated the changelog and pushed an update to the upgrade notes in the v5.33-documentation branch. I've removed the forum post and the original breaking change note from the blog post. I'll post an update in the PM:GTM channel tomorrow.

PR #17147, opened 2021-03-15
[MM-33794] Improve password generation during bulk import

streamer45:
#### Summary

PR improves the password generation logic used during the bulk import process.

This is technically a breaking change to a `model` function although I am assuming (and hoping) that's being used internally only. Let me know if you have any concern. I am also happy to move all this logic to the app layer, closer to where it's used.

@srkgupta Feel free to add additional reviewers from your team if you deem it appropriate.

#### Ticket

https://mattermost.atlassian.net/browse/MM-33794

#### Release Note

```release-note
Removed the utility function model.GeneratePassword() for security reasons. An improved version is now being used internally to generate passwords for bulk-imported users.
```


agnivade (commented):
Yes, generating passwords does not, in any way, need to be in model. I'd prefer to have an internal function using crypto/rand in the app layer, and mark model.GeneratePassword as deprecated because it's insecure, and remove it in v6 altogether.

streamer45:
> Yes, generating passwords does not, in any way, need to be in model. I'd prefer to have an internal function using crypto/rand in the app layer, and mark model.GeneratePassword as deprecated because it's insecure, and remove it in v6 altogether.

Makes sense although I am not super in love with keeping the original. It's not really part of our client or anything like that so I am almost inclined to completely removing it. Also when it comes to security I think breaking changes are somewhat allowed :P

agnivade:
I agree with your sentiments. I think this should be well-defined in a policy as to what is allowed and what is not. Right now, we have more or less free reign on removing whatever we want if there's nobody else using it. But in general, I am okay with removing it, along with a note in our release notes.

nronas (approved):
LGTM 🚢 - it

agnivade (approved):

srkgupta (approved):
Thanks @streamer45. The changes LGTM

PR #17148, opened 2021-03-16
MM-33818: Add replica lag metric

agnivade:
We add two new metrics for monitoring replica lag:
- Monitor absolute lag based on binlog distance/transaction queue length.
- Monitor time taken for the replica to catch up.

To achieve this, we add a config setting to run a user defined SQL query
on the database.

We need to specify a separate datasource field as part of the config because
in some databases, querying the replica lag value requires elevated credentials
which are not needed for usual running of the application, and can even be a security risk.

Arguably, a peculiar part of the design is the requirement of the query output to be in a (node, value)
format. But since from the application, the SQL query is a black box and the user can set any query
they want, we cannot, in any way templatize this.

And as an extra note, freno also does it in a similar way.

The last bit is because we need to have a separate datasources, now we consume one extra connection
rather than sharing it with the pool. This is an unfortunate result of the design, and while one extra
connection doesn't make much of a difference in a single-tenant scenario. It does make so, in a multi-tenant scenario.
But in a multi-tenant scenario, the expectation would already be to use a connection pool. So this
is not a big concern.

https://mattermost.atlassian.net/browse/MM-33818

```release-note
Two new gauge metrics were added:
mattermost_db_replica_lag_abs and mattermost_db_replica_lag_time, both
containing a label of "node", signifying which db host is the metric from.

These metrics signify the replica lag in absolute terms and in the time dimension
capturing the whole picture of replica lag.

To use these metrics, a separate config section ReplicaLagSettings was added
under SqlSettings. This is an array of maps which contain three keys: DataSource,
QueryAbsoluteLag, and QueryTimeLag. Each map entry is for a single replica instance.

DataSource contains the DB credentials to connect to the replica instance.

QueryAbsoluteLag is a plain SQL query that must return a single row of which the first column
must be the node value of the Prometheus metric, and the second column must be the value of the lag.

QueryTimeLag is the same as above, but used to measure the time lag.

As an example, for AWS Aurora instances, the QueryAbsoluteLag can be:

select server_id, highest_lsn_rcvd-durable_lsn as bindiff from aurora_global_db_instance_status() where server_id=<>

and QueryTimeLag can be:

select server_id, visibility_lag_in_msec from aurora_global_db_instance_status() where server_id=<>

For MySQL Group Replication, the absolute lag can be measured from the number of pending transactions
in the applier queue:

select member_id, count_transactions_remote_in_applier_queue FROM performance_schema.replication_group_member_stats where member_id=<>

Overall, what query to choose is left to the administrator, and depending on the database and need, an appropriate
query can be chosen.
```


agnivade:
@gabrieljackson - Requesting you to please take a look and let me know if there is any concern from a Cloud perspective.

Basically, this PR adds a metric to monitor replica lag. There are 2 metrics - one for absolute, one for time-based. And both are measured with a SQL query to every replica.

To achieve this, we need to add a config setting for every replica:

```
"ReplicaLagSettings": [
            {
                "DataSource": "postgres://mmuser:mostest@localhost/loadtest?sslmode=disable",
                "QueryAbsoluteLag": "select server_id, highest_lsn_rcvd-durable_lsn as bindiff from aurora_global_db_instance_status() where server_id=<>",
                "QueryTimeLag": "select server_id, visibility_lag_in_msec from aurora_global_db_instance_status() where server_id=<>"
            },
{...},
{...}
        ]
```

Hopefully, this is not too much overhead for the Cloud team. If you have any concerns/suggestions, please let me know. Thanks!

nronas:
👋 @agnivade all the above does make sense, but I was a bit surprised with the usage of a configuration variable for this. 

I am most likely missing context here, but i was expecting a scheduled(every 1-minute lets say) job that would create a connection to the master and calculate the lag from it's binlog positions and report it to metrics(Prometheus), i.e for MySQL would run `SHOW MASTER STATS` and `SHOW REPLICA STATS` and compute the lag for the given node. What's the main need for parameterization here? Like why our customers need to know what SQL query to create in order to make the replica lag calculation work?

agnivade:
@nronas - Yes, great question! The context is that the application treats the lag metric as a black box. We leave it to the user to define whatever way they want to measure the lag. There is no single metric that works - for example for Group Replication, would you track `COUNT_TRANSACTIONS_REMOTE_IN_APPLIER_QUEUE` or `COUNT_TRANSACTIONS_IN_QUEUE`? Both are indicators of lag, and any one can be used.

> What's the main need for parameterization here? Like why our customers need to know what SQL query to create in order to make the replica lag calculation to work?

Different databases in different environments expose the replica lag metric in different ways. For example Aurora Postgres is different from RDS Postgres, Aurora MySQL is different than RDS binlog MySQL, which is different than Group Replication MySQL.

We would need to detect what flavor of a database a customer is running and maintain all of the queries ourselves. So to get rid of all the complexity from within the application, we let the user specify whatever they want and we just capture the value and dump to Prometheus.

This is also the approach taken by Freno. Internally they use pt-heartbeat, but the query is externalized and the user can specify whatever they want. And the tool throttles to minimize that metric.

nronas:
Awesome, i thought for any type of db backend we would only care about the engine that it runs which we already do across the codebase, and then us standard replication base metrics, like how the Datadog agent does it for PG and MySQL. Though if you found that depending on the topology there are better metrics to be used then 👍 to parameterize this.

agnivade:
Yeah .. even within an engine, things are radically different. Entirely different queries to entirely different tables.

For Datadog, I am seeing different metrics even for the same engine:
- https://docs.datadoghq.com/integrations/postgres
- https://docs.datadoghq.com/integrations/amazon_rds/

You at least need to differentiate between on-prem Postgres and RDS Postgres.

nronas:
Yeap totally, i was more thinking us defining a metric that indicates lag and use that across same engines, but i guess we are loosing granularity and/or accuracy so makes sense to differentiate.

nronas (approved):
LGTM

gabrieljackson (approved):
Very nice! Thanks for all the context on this approach.

From the cloud point of view the extra configuration is not an issue for us and can be easily updated. Regarding the extra connection to the replica DB, this also isn't too bad as it's common for our replicas nodes to have much more overhead than the writer.

store/sqlstore/store.go
```diff
@@ -403,6 +428,42 @@ func (ss *SqlStore) TotalMasterDbConnections() int {
...
+			continue
+		}
+		var binDiff int
+		var node string
+		err := ss.replicaLagHandles[i].QueryRow(*item.QueryAbsoluteLag).Scan(&node, &binDiff)
+		if err != nil {
+			mlog.Warn("ReplicaLagAbs Query returned an error", mlog.Err(err))
+			return
+		}
+		ss.metrics.SetReplicaLagAbsolute(node, float64(binDiff))
```
streamer45, line 445: Is `ss.metrics` guaranteed to be non-nil?

store/sqlstore/store.go
```diff
@@ -403,6 +428,42 @@ func (ss *SqlStore) TotalMasterDbConnections() int {
...
+			continue
+		}
+		var timeDiff int
+		var node string
+		err := ss.replicaLagHandles[i].QueryRow(*item.QueryTimeLag).Scan(&node, &timeDiff)
+		if err != nil {
+			mlog.Warn("ReplicaLagTime Query returned an error", mlog.Err(err))
+			return
+		}
+		ss.metrics.SetReplicaLagTime(node, float64(timeDiff))
```
streamer45, line 463: Same.

store/sqlstore/store.go
```diff
@@ -403,6 +428,42 @@ func (ss *SqlStore) TotalMasterDbConnections() int {
...
+		if *item.QueryAbsoluteLag == "" {
+			continue
+		}
+		var binDiff int
+		var node string
+		err := ss.replicaLagHandles[i].QueryRow(*item.QueryAbsoluteLag).Scan(&node, &binDiff)
+		if err != nil {
+			mlog.Warn("ReplicaLagAbs Query returned an error", mlog.Err(err))
+			return
+		}
```
streamer45, line 444: What's the reason for logging the error instead of returning it?

store/sqlstore/store.go
```diff
@@ -403,6 +428,42 @@ func (ss *SqlStore) TotalMasterDbConnections() int {
...
+}
+
+// ReplicaLagAbs queries all the replica databases to get the time-based replica lag value
+// and updates the Prometheus metric with it.
+func (ss *SqlStore) ReplicaLagTime() {
+	for i, item := range ss.settings.ReplicaLagSettings {
+		if *item.QueryTimeLag == "" {
+			continue
+		}
+		var timeDiff int
```
streamer45, line 456: Are we confident that this metric will always fit a int? I am wondering if there are cases where this data could be a floating point number on the db side. In any case maybe we should mention the datatype requirement in the release note.

store/sqlstore/store.go
agnivade: That's a good point. I will check.

store/sqlstore/store.go
agnivade: Hmm .. yeah I thought about it. You mean logging the error in the enterprise metrics handling code? That might work. I just wasn't sure if it was elegant or not. See that PR and let me know if you are okay with it or not?

store/sqlstore/store.go
agnivade: Good point. I will update.

store/sqlstore/store.go
streamer45: I guess the final result would be the same but it just feels better to leave the caller decide what to do with it. Logging on the metrics callback sounds good to me.

agnivade: Sounds good then. I will change.

store/sqlstore/store.go
agnivade: Yes, since this method is called from the metrics store, it will always be. I will add a comment to clarify that.

streamer45 (approved):
Looks great :+1:

agnivade:
@lindalumitchell - Would be good to have this verified by an SDET. The test steps are outlined in the ticket.

lindalumitchell:
@agnivade my apologies for this one falling through the cracks. Getting some help on it as soon as possible.

saturninoabril (approved):
Tested and passed.
Screen Shot 2021-03-29 at 9 20 13 AM

PR #17150, opened 2021-03-16
MM-33746 Add TotalMsgCountRoot and MsgCountRoot columns 

reflog:
#### Summary

added TotalMsgCountRoot and MsgCountRoot columns and migrations
#### Release Note

```release-note
Added TotalMsgCountRoot to Channels table and MsgCountRoot column to ChannelMembers table. Please note that the migration on large MySQL instances can take several minutes to complete.
```


reflog:
@streamer45 - please take a look at the upgrade portion for any performance concerns

zefhemel (approved):
Sooooooo much better than the last implementation. LGTM!

streamer45 (commented):
Took a good look at this. I haven't done any performance test on the application side yet but focused on the migration part.

Adding the columns as explained before can take a while on big MySQL installations but it's a non-locking operation so it would be safe to run asynchronously. These are the results on the data I tested:

```
mysql> ALTER TABLE Channels ADD COLUMN TotalMsgCountRoot bigint;
Query OK, 1184393 rows affected (1 min 41.21 sec)
Records: 1184393  Duplicates: 0  Warnings: 0
```

```
mysql> ALTER TABLE ChannelMembers ADD COLUMN MsgCountRoot bigint;
Query OK, 0 rows affected (1 min 14.32 sec)
```

Overall the first one (on `Channels`) takes longer compared to the number of rows (`ChannelMembers` has ~6M entries). 

Now coming to the updates:

```
mysql> UPDATE Channels INNER JOIN (SELECT Channels.id channelid, COALESCE(COUNT(CASE WHEN Posts.rootid = '' THEN 1 END),0) newcount FROM Channels LEFT JOIN Posts ON Channels.id = Posts.channelid GROUP BY Channels.id) AS q ON q.channelid=Channels.id SET totalmsgcountroot = q.newcount;
Query OK, 1184393 rows affected (3 min 15.72 sec)
```

The equivalent `ChannelMembers` cannot possibly work as it is. We are essentially joining the table over itself. I tried it just out of curiosity but after an hour waiting I just killed it.

```
mysql> EXPLAIN UPDATE ChannelMembers INNER JOIN (SELECT ChannelMembers.channelid, COALESCE(COUNT(CASE WHEN Posts.rootid = '' AND Posts.updateat < ChannelMembers.lastviewedat THEN 1 END),0) newcount FROM ChannelMembers LEFT JOIN Posts ON ChannelMembers.channelid = Posts.channelid GROUP BY ChannelMembers.channelid) AS q ON q.channelid=ChannelMembers.channelid SET msgcountroot = q.newcount\G
           id: 1
  select_type: UPDATE
        table: ChannelMembers
   partitions: NULL
         type: ALL
possible_keys: PRIMARY,idx_channelmembers_channel_id
          key: NULL
      key_len: NULL
          ref: NULL
         rows: 6468649
     filtered: 100.00
        Extra: NULL
*************************** 2. row ***************************
           id: 1
  select_type: PRIMARY
        table: <derived2>
   partitions: NULL
         type: ref
possible_keys: <auto_key0>
          key: <auto_key0>
      key_len: 106
          ref: perconaltdb.ChannelMembers.ChannelId
         rows: 65
     filtered: 100.00
        Extra: NULL
*************************** 3. row ***************************
           id: 2
  select_type: DERIVED
        table: ChannelMembers
   partitions: NULL
         type: index
possible_keys: PRIMARY,idx_channelmembers_channel_id,idx_channelmembers_user_id
          key: PRIMARY
      key_len: 212
          ref: NULL
         rows: 6468649
     filtered: 100.00
        Extra: NULL
*************************** 4. row ***************************
           id: 2
  select_type: DERIVED
        table: Posts
   partitions: NULL
         type: ref
possible_keys: idx_posts_channel_id,idx_posts_channel_id_update_at,idx_posts_channel_id_delete_at_create_at
          key: idx_posts_channel_id
      key_len: 107
          ref: perconaltdb.ChannelMembers.ChannelId
         rows: 65
     filtered: 100.00
        Extra: NULL
4 rows in set (0.00 sec)
```

jespino (approved):
LGTM. Great work @reflog! :confetti_ball:

jespino:
@streamer45 that are problems on MySQL only, right? the PostgreSQL implementation is fast enough, right?

streamer45:
> @streamer45 that are problems on MySQL only, right? the PostgreSQL implementation is fast enough, right?

Adding columns is instant in PostgreSQL so no issues there. I'd expect the actual data migration query to be rather heavy on PostgreSQL as well but haven't tested it yet.

store/sqlstore/upgrade.go
```diff
@@ -1000,6 +1001,99 @@ func upgradeDatabaseToVersion533(sqlStore *SqlStore) {
...
+					break
+				}
+				membership := memberships[membershipIndex]
+				if membership.LastViewedAt < postTimestamp && membership.LastViewedAt > postTimestamps[counter-1] {
+					params := map[string]interface{}{
+						"ChannelId":    channel.Id,
+						"UserId":       membership.UserId,
+						"MsgCountRoot": channel.TotalMsgCountRoot - int64(counter),
+					}
+					if _, err = transaction.Exec("UPDATE ChannelMembers SET MsgCountRoot=:MsgCountRoot WHERE ChannelId=:ChannelId AND UserId=:UserId", params); err != nil {
```
jespino, line 1079: I think it is ok to do it here, but we can squeeze a bit more performance by grouping all the memberships updated with the same count, and updating all with one query.

jespino: Anyway... lets keep it as is now, and if we need that extra performance we can try to squeeze then, because I think the code is going to be a bit more hard to understand.

reflog:
@streamer45 @jespino please re-re-re-review :)
This time the query is significantly cheaper and according to my tests takes only seconds on the largest db I have

store/sqlstore/post_store.go
```diff
@@ -143,6 +144,18 @@ func (s *SqlPostStore) SaveMultiple(posts []*model.Post) ([]*model.Post, int, er
 		}
 
 		if post.RootId == "" {
+			currentChannelCount, ok2 := channelNewRootPosts[post.ChannelId]
+			if !ok2 {
```
zefhemel, line 148: Purely stylistic suggestion (feel free to ignore, but I've been writing some more Go myself recently and started to use this pattern and it avoids the mess of creating a second `ok` variable and the `currentChannelCount` will also be available in the `else` clause):

```
if currentChannelCount, ok := channelNewRootPosts[post.ChannelId]; ok {
   ...
} else {
   ...
}
```

store/sqlstore/upgrade.go
```diff
@@ -1011,6 +1011,49 @@ func upgradeDatabaseToVersion535(sqlStore *SqlStore) {
...
+
+			UPDATE ChannelMembers CM
+			INNER JOIN (` + c2 + `) as q
+			ON q.id=CM.ChannelId and LastViewedAt >= q.lastrootpostat
+			SET MsgCountRoot=TotalMsgCountRoot
+			`
+	}
+	if _, err := sqlStore.GetMaster().Exec(uberQ); err != nil {
+		mlog.Error("Error fetching Channels table", mlog.Err(err))
+	}
```
streamer45, line 1054: This is failing on MySQL with:

```go
{"level":"error","ts":1617104190.3134878,"caller":"sqlstore/upgrade.go:1053","msg":"Error fetching Channels table","error":"Error 1064: You have an error in your SQL syntax; check the manual that corresponds to your MySQL server version for the right syntax to use near 'UPDATE ChannelMembers CM\n\t\t\tINNER JOIN (SELECT TotalMsgCountRoot, Id, LastRootPo' at line 12"}
```

The query looks fine. Maybe it's a formatting issue.

streamer45: On a second look it might actually be cause of the double query. We may need to execute them separately.

store/sqlstore/upgrade.go
```diff
@@ -1011,6 +1011,53 @@ func upgradeDatabaseToVersion535(sqlStore *SqlStore) {
...
+		WITH q as (` + channelsCTE + `)
+		UPDATE ChannelMembers CM SET MsgCountRoot=TotalMsgCountRoot
+		FROM q WHERE q.id=CM.ChannelId AND LastViewedAt >= q.lastrootpostat;
+	`
+	if sqlStore.DriverName() == model.DATABASE_DRIVER_MYSQL {
+		updateChannels = `
+			UPDATE Channels
+			INNER Join (` + totalMsgCountRootCTE + `) as q
+			ON q.channelid=Channels.Id
+			SET TotalMsgCountRoot = q.newcount, LastRootPostAt=q.lastpost;
```
streamer45, line 1045: Tried to run a comparison to test the perf impact of the PR but this query is still taking more than 1 hour to complete on MySQL. Interestingly the subquery (SELECT) only takes ~12 seconds.

streamer45: The main issue seems to be the JOIN on `Posts` which is not using an index and falls back to table-scanning:

```
           id: 1
  select_type: UPDATE
        table: Channels
   partitions: NULL
         type: ALL
possible_keys: PRIMARY
          key: NULL
      key_len: NULL
          ref: NULL
         rows: 175270
     filtered: 100.00
        Extra: NULL
*************************** 2. row ***************************
           id: 1
  select_type: PRIMARY
        table: <derived2>
   partitions: NULL
         type: ref
possible_keys: <auto_key0>
          key: <auto_key0>
      key_len: 106
          ref: claudioltcomp1db.Channels.Id
         rows: 452345
     filtered: 100.00
        Extra: NULL
*************************** 3. row ***************************
           id: 2
  select_type: DERIVED
        table: Channels
   partitions: NULL
         type: index
possible_keys: PRIMARY,Name,idx_channels_team_id,idx_channels_name,idx_channels_update_at,idx_channels_create_at,idx_channels_delete_at,idx_channels_scheme_id,idx_channel_search_txt
          key: idx_channels_update_at
      key_len: 9
          ref: NULL
         rows: 175270
     filtered: 100.00
        Extra: Using index; Using temporary; Using filesort
*************************** 4. row ***************************
           id: 2
  select_type: DERIVED
        table: Posts
   partitions: NULL
         type: ALL
possible_keys: idx_posts_channel_id,idx_posts_root_id,idx_posts_channel_id_update_at,idx_posts_channel_id_delete_at_create_at
          key: NULL
      key_len: NULL
          ref: NULL
         rows: 2714070
     filtered: 16.67
        Extra: Using where; Using join buffer (Block Nested Loop)
4 rows in set (0.00 sec)
```

streamer45: Forcing the index seems to be helping:

#### Before:

```
mysql> UPDATE Channels
    -> INNER JOIN
    ->   (SELECT Channels.Id channelid,
    ->           COALESCE(COUNT(*), 0) newcount,
    ->           COALESCE(MAX(Posts.CreateAt), 0) AS lastpost
    ->    FROM Channels
    ->    LEFT JOIN Posts ON Channels.Id = Posts.ChannelId
    ->    WHERE Posts.RootId = ''
    ->    GROUP BY Channels.Id) AS q ON q.channelid=Channels.Id
    -> SET TotalMsgCountRoot = q.newcount,
    ->     LastRootPostAt=q.lastpost;
Query OK, 4 rows affected (5 hours 9 min 6.70 sec)
Rows matched: 4  Changed: 4  Warnings: 0
```

#### After

```
mysql> UPDATE Channels
    -> INNER JOIN
    ->   (SELECT Channels.Id channelid,
    ->           COALESCE(COUNT(*), 0) newcount,
    ->           COALESCE(MAX(Posts.CreateAt), 0) AS lastpost
    ->    FROM Channels
    ->    LEFT JOIN Posts
    ->    FORCE INDEX(idx_posts_channel_id) ON Channels.Id = Posts.ChannelId
    ->    WHERE Posts.RootId = ''
    ->    GROUP BY Channels.Id) AS q ON q.channelid=Channels.Id
    -> SET TotalMsgCountRoot = q.newcount,
    ->     LastRootPostAt=q.lastpost;
Query OK, 4 rows affected (12.66 sec)
Rows matched: 4  Changed: 4  Warnings: 0
```

store/sqlstore/upgrade.go
```diff
@@ -1019,10 +1019,15 @@ func upgradeDatabaseToVersion535(sqlStore *SqlStore) {
...
 
+	forceIndex := ""
+	if sqlStore.DriverName() == model.DATABASE_DRIVER_MYSQL {
+		forceIndex = "FORCE INDEX(idx_posts_channel_id) ON Channels.Id = Posts.ChannelId"
+	}
 	totalMsgCountRootCTE := `
 		SELECT Channels.Id channelid, COALESCE(COUNT(*),0) newcount, COALESCE(MAX(Posts.CreateAt), 0) as lastpost
 		FROM Channels
 		LEFT JOIN Posts ON Channels.Id = Posts.ChannelId
+		` + forceIndex + `
```
streamer45, line 1030: Needs to go right after the table name, like in my query above.

reflog: @streamer45 oops. fixing

PR #17153, opened 2021-03-16
Use denormalized ReplyCount from Threads table to populate the posts ReplyCount field

jespino:
#### Summary
Use denormalized ReplyCount from Threads table to populate the posts ReplyCount field

#### Ticket Link
Not yet

#### Release Note
```release-note
NONE
```

### DIFF
--- a/store/sqlstore/post_store.go
+++ b/store/sqlstore/post_store.go
@@ -345,7 +345,7 @@ func (s *SqlPostStore) GetFlaggedPosts(userId string, offset int, limit int) (*m
 	pl := model.NewPostList()
 
 	var posts []*model.Post
-	if _, err := s.GetReplica().Select(&posts, "SELECT *, (SELECT count(Posts.Id) FROM Posts WHERE Posts.RootId = (CASE WHEN p.RootId = '' THEN p.Id ELSE p.RootId END) AND Posts.DeleteAt = 0) as ReplyCount FROM Posts p WHERE Id IN (SELECT Name FROM Preferences WHERE UserId = :UserId AND Category = :Category) AND DeleteAt = 0 ORDER BY CreateAt DESC LIMIT :Limit OFFSET :Offset", map[string]interface{}{"UserId": userId, "Category": model.PREFERENCE_CATEGORY_FLAGGED_POST, "Offset": offset, "Limit": limit}); err != nil {
+	if _, err := s.GetReplica().Select(&posts, "SELECT p.*, COALESCE(T.ReplyCount, 0) as ReplyCount FROM Posts p LEFT JOIN Threads as T ON T.PostId = (CASE WHEN p.RootId IS NULL OR p.RootId = '' THEN p.Id ELSE p.RootId END) WHERE p.Id IN (SELECT Name FROM Preferences WHERE UserId = :UserId AND Category = :Category) AND p.DeleteAt = 0 ORDER BY p.CreateAt DESC LIMIT :Limit OFFSET :Offset", map[string]interface{}{"UserId": userId, "Category": model.PREFERENCE_CATEGORY_FLAGGED_POST, "Offset": offset, "Limit": limit}); err != nil {
 		return nil, errors.Wrap(err, "failed to find Posts")
 	}
 
@@ -364,7 +364,7 @@ func (s *SqlPostStore) GetFlaggedPostsForTeam(userId, teamId string, offset int,
 
 	query := `
             SELECT
-                A.*, (SELECT count(Posts.Id) FROM Posts WHERE Posts.RootId = (CASE WHEN A.RootId = '' THEN A.Id ELSE A.RootId END) AND Posts.DeleteAt = 0) as ReplyCount
+                A.*, COALESCE(T.ReplyCount, 0) as ReplyCount
             FROM
                 (SELECT
                     *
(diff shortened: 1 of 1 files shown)


store/sqlstore/post_store.go
```diff
@@ -345,7 +345,7 @@ func (s *SqlPostStore) GetFlaggedPosts(userId string, offset int, limit int) (*m
 	pl := model.NewPostList()
 
 	var posts []*model.Post
-	if _, err := s.GetReplica().Select(&posts, "SELECT *, (SELECT count(Posts.Id) FROM Posts WHERE Posts.RootId = (CASE WHEN p.RootId = '' THEN p.Id ELSE p.RootId END) AND Posts.DeleteAt = 0) as ReplyCount FROM Posts p WHERE Id IN (SELECT Name FROM Preferences WHERE UserId = :UserId AND Category = :Category) AND DeleteAt = 0 ORDER BY CreateAt DESC LIMIT :Limit OFFSET :Offset", map[string]interface{}{"UserId": userId, "Category": model.PREFERENCE_CATEGORY_FLAGGED_POST, "Offset": offset, "Limit": limit}); err != nil {
+	if _, err := s.GetReplica().Select(&posts, "SELECT p.*, COALESCE(T.ReplyCount, 0) as ReplyCount FROM Posts p LEFT JOIN Threads as T ON T.PostId = (CASE WHEN p.RootId IS NULL OR p.RootId = '' THEN p.Id ELSE p.RootId END) WHERE p.Id IN (SELECT Name FROM Preferences WHERE UserId = :UserId AND Category = :Category) AND p.DeleteAt = 0 ORDER BY p.CreateAt DESC LIMIT :Limit OFFSET :Offset", map[string]interface{}{"UserId": userId, "Category": model.PREFERENCE_CATEGORY_FLAGGED_POST, "Offset": offset, "Limit": limit}); err != nil {
```
reflog, line 348: @jespino can RootID be null? I think that condition can be simpler.

store/sqlstore/post_store.go
```diff
@@ -882,9 +879,9 @@ func (s *SqlPostStore) GetPostsSince(options model.GetPostsSinceOptions, allowFr
 		WHERE
 		       UpdateAt > :Time AND ChannelId = :ChannelId
 		       LIMIT 1000)
-		(SELECT *` + replyCountQuery2 + ` FROM cte)
+		(SELECT cte.*, COALESCE(T.ReplyCount, 0) as ReplyCount FROM cte LEFT JOIN Threads as T ON T.PostId = (CASE WHEN cte.RootId IS NULL OR cte.RootId = '' THEN cte.Id ELSE cte.RootId END))
```
reflog, line 882: @jespino this subquery used to run under SkipFetchThreads flag. Now it will run always. Is that what expected? Might cause slight degradation ...

reflog (approved):
looks good, left some questions

store/sqlstore/post_store.go
jespino: Yes from the database perspective, I guess no from the application perspective. I was playing safe there. :)

store/sqlstore/post_store.go
jespino: Good point, I'm not sure the reasoning behind that, I think including the replyCount there is going to be a bit less performant, I can change it.

store/sqlstore/post_store.go
```diff
@@ -837,17 +839,11 @@ func (s *SqlPostStore) GetPostsSince(options model.GetPostsSinceOptions, allowFr
...
-	if options.SkipFetchThreads {
-		replyCountQuery1 = `, (SELECT COUNT(Posts.Id) FROM Posts WHERE Posts.RootId = (CASE WHEN p1.RootId = '' THEN p1.Id ELSE p1.RootId END) AND Posts.DeleteAt = 0) as ReplyCount`
-		replyCountQuery2 = `, (SELECT COUNT(Posts.Id) FROM Posts WHERE Posts.RootId = (CASE WHEN cte.RootId = '' THEN cte.Id ELSE cte.RootId END) AND Posts.DeleteAt = 0) as ReplyCount`
-	}
 	var query string
 
 	// union of IDs and then join to get full posts is faster in mysql
 	if s.DriverName() == model.DATABASE_DRIVER_MYSQL {
-		query = `SELECT *` + replyCountQuery1 + ` FROM Posts p1 JOIN (
+		query = `SELECT p1.*, COALESCE(T.ReplyCount, 0) as ReplyCount FROM Posts p1 JOIN (
```
streamer45, line 846: Did reply counts get migrated somehow or are we going to miss posts from before threads were introduced?

jespino: Good point, I think this would need a migration, but would be interesting to know if this is going to have a important performance implication or not, because the implication in the performance is minimal, probably the migration itself is not going to be interesting.

jespino: Other option is to "fallback" to the old approach or "migrate on the fly" whenever is found a non-existing Thread object for a Thread.

streamer45: Sounds good. I am running a couple of load-tests as we speak. Will report soon.

streamer45:
load-test went fine. Nothing noticeable to report:

image

Yellow is this PR (53a5a66). Green is the base (72089520).

jespino:
@streamer45 I don't see there enough difference in performance to justify this change.

streamer45:
> @streamer45 I don't see there enough difference in performance to justify this change.

Yes, if the primary goal is perf improvement there's isn't much to look at. No significant change between the two runs.

jespino:
Perfect, then we can close this

PR #17155, opened 2021-03-16
MM-33846 disallow cyclic shared channel invites

wiggin77:
#### Summary
This PR ensures that when a cluster has been invited to a shared channel, that cluster cannot then invite other organizations.  This also prevents any cyclic sync scenarios by ensuring only the home server for a channel can send invitations.

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-33846

#### Release Note
```release-note
NONE
```


app/shared_channel.go
```diff
@@ -40,6 +40,23 @@ func (a *App) checkChannelIsShared(channelId string) error {
 	return nil
 }
 
+func (a *App) CheckCanInviteToSharedChannel(channelId string) error {
+	var err error
+	var sc *model.SharedChannel
+	if sc, err = a.GetSharedChannel(channelId); err != nil {
```
agnivade, line 46: I'd suggest breaking up the if-err clause, and remove the `var` declarations above.

agnivade (commented):
Can we also add a test please? :)

app/shared_channel.go
wiggin77: Fixed via https://github.com/mattermost/mattermost-server/pull/17155/commits/3d505410b1785518bda895015e632d5a873c14da  including unit tests.

agnivade (approved):
Hmm .. the new linter unfortunately caught some issues in this branch. I will send a PR to disable it. Looks like it will cause a lot of pain for others.

isacikgoz (approved):
LGTM 👍

PR #17156, opened 2021-03-16
[MM-33076] Add migrations for OAuthAccessData schema

mgdelacroix:
#### Ticket Link
https://mattermost.atlassian.net/browse/MM-33076

#### Release Note
```release-note
* Moves OAuthAccessData gorp driven migrations to schema-based migrations
```


db/migrations/postgres/000012_create_oauth_access_data.up.sql
```diff
@@ -0,0 +1,35 @@
...
+
+ALTER TABLE oauthaccessdata ADD COLUMN IF NOT EXISTS userid VARCHAR(26) DEFAULT '';
+CREATE INDEX IF NOT EXISTS idx_oauthaccessdata_user_id ON oauthaccessdata (userid);
+
+DO $$
+BEGIN
+    IF NOT EXISTS (
+        SELECT conname
+        FROM pg_constraint
+        WHERE conname = 'oauthaccessdata_clientid_userid_key'
```
mgdelacroix, line 21: Not sure if this is the best way to fetch the unique constraint as its not linked to any table, but name is unique and it works. Just in case you know a better way to do it

nronas: It looks like this is part of the gorp create table, so i guess we can added in `CREATE TABLE ...`

PR #17161, opened 2021-03-16
[MM-32044] Reset SAML auth data

maxerenberg:
#### Summary
This PR adds an API endpoint which resets SAML users' AuthData field to their email.

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-32044

#### Release Note
```release-note
Added new API endpoint POST /api/v4/saml/reset_auth_data
```


app/saml.go
```diff
@@ -282,3 +282,16 @@ func (a *App) SetSamlIdpCertificateFromMetadata(data []byte) *model.AppError {
 
 	return nil
 }
+
+func (a *App) ResetSamlAuthDataToEmail(includeDeleted bool, dryRun bool, userIDs []string) (numAffected int64, appErr *model.AppError) {
```
sbishel, line 286: Just use `int`

api4/saml.go
```diff
@@ -247,3 +250,28 @@ func getSamlMetadataFromIdp(c *Context, w http.ResponseWriter, r *http.Request)
 
 	w.Write([]byte(metadata.ToJson()))
 }
+
+func resetAuthDataToEmail(c *Context, w http.ResponseWriter, r *http.Request) {
+	if !c.App.SessionHasPermissionTo(*c.App.Session(), model.PERMISSION_MANAGE_SYSTEM) {
+		c.SetPermissionError(model.PERMISSION_MANAGE_SYSTEM)
+		return
+	}
```
sbishel, line 258: Do we need a license check here?

einterfaces/saml.go
```diff
@@ -12,5 +12,6 @@ type SamlInterface interface {
 	BuildRequest(relayState string) (*model.SamlAuthRequest, *model.AppError)
 	DoLogin(encodedXML string, relayState map[string]string) (*model.User, *model.AppError)
 	GetMetadata() (string, *model.AppError)
+	ResetAuthDataToEmail(includeDeleted bool, dryRun bool, specifiedUserIDs []string) (numAffected int64, err error)
```
sbishel, line 15: just use `int`

model/client4.go
```diff
@@ -3889,6 +3889,28 @@ func (c *Client4) GetSamlMetadataFromIdp(samlMetadataURL string) (*SamlMetadataR
 	return SamlMetadataResponseFromJson(r.Body), BuildResponse(r)
 }
 
+// ResetSamlAuthDataToEmail resets the AuthData field of SAML users to their Email.
+func (c *Client4) ResetSamlAuthDataToEmail(includeDeleted bool, dryRun bool, userIDs []string) (int64, *Response) {
```
sbishel, line 3893: int

sbishel (changes requested):

api4/saml.go
maxerenberg: None of the other SAML functions in the API layer are performing a license check, so I'm guessing we don't need one.

sbishel (approved):

api4/saml.go
```diff
@@ -247,3 +250,28 @@ func getSamlMetadataFromIdp(c *Context, w http.ResponseWriter, r *http.Request)
...
+	type ResetAuthDataParams struct {
+		IncludeDeleted   bool     `json:"include_deleted"`
+		DryRun           bool     `json:"dry_run"`
```
sbishel, line 261: Sorry, I missed this before and noticed when reviewing the documentation.  I gave some bad advice. Dry-run should default to "true", but we don't want to default a boolean to true, so we should turn this flag into something like..."execute" with a default it to false. That way they have to set a flag in order to manipulate the database.

sbishel (changes requested):
Sorry, I missed this before. I noticed when reviewing documentation.

api4/saml.go
maxerenberg: As discussed offline, we will be keeping the API parameters the same as this would be more intuitive.

store/sqlstore/user_store.go
```diff
@@ -307,6 +307,52 @@ func (us SqlUserStore) UpdateAuthData(userId string, service string, authData *s
...
+	if err != nil {
+		return 0, errors.Wrap(err, "update_users_tosql")
+	}
+	result, err := us.GetMaster().Exec(query, args...)
+	if err != nil {
+		return 0, errors.Wrap(err, "failed to update users' AuthData")
+	}
+	numAffected, err := result.RowsAffected()
+	return int(numAffected), err
+}
```
sbishel, line 354: The "where" clause of these queries should be built from common code. If someone updated one section and not the other, the dry-run flag may not do exactly what the execution does.

sbishel (changes requested):
I like this method much better.

store/sqlstore/user_store.go
maxerenberg: Unfortunately sq.SelectBuilder and sq.UpdateBuilder are two different types and I couldn't find a way to apply the `Where` logic in one common place. If anyone has ideas on how to do this, I'm all ears!

sbishel: You could build them into a map, then add them to builder. 
```
	whereClauses := make(map[string]interface{})
	whereClauses["AuthService"] = service
	if len(userIDs) > 0 {
		whereClauses["Id"] = userIDs
	}
	if !includeDeleted {
		whereClauses["DeleteAt"] = 0
	}
```

Then add them to your builder. Looks like they will accept a `map[string]interface{}` so you should be able to add them directly.
```
	builder = builder.Where(whereClauses)
```

sbishel (changes requested):

sbishel (approved):

mkraft (approved):

furqanmlk (approved):
Remove ID Attribute in SAML
image

Reset Auth data by API
image

After running API
image

PR #17163, opened 2021-03-16
MM-33905 -always add FeatureFlags to config

sbishel:
#### Summary
Because FeatureFlags does not contain an `access:` tags. They are only returned for a user with `model.PERMISSION_MANAGE_SYSTEM`, ie. system_admin.

FeatureFlags should be returned for any user that contains access to the SystemConsole. This call is only called for those users with a SYSCONSOLE_READ_* permission.

Returning Feature flags in the Admin Config, is required for using FeatureFlags in `admin_definition.jsx`.

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-33905

#### Release Note
```release-note
NONE
```


hahmadia (approved):
Make sense to me

api4/config.go
```diff
@@ -24,6 +24,7 @@ type filterType string
 
 const filterTypeWrite filterType = "write"
 const filterTypeRead filterType = "read"
+const featureFlagsElementName string = "FeatureFlags"
```
hahmadia, line 27: Super duper Nit but the casing for our CONST in server is now `TitleCase`
You can also probably reduce this to something like
```suggestion
const FilterTypeWrite filterType = "write"
const FilterTypeRead filterType = "read"
const FeatureFlagsElementName string = "FeatureFlags"
```
Here is an example of existing const 
image

hahmadia (approved):
Super nit change

api4/config.go
```diff
@@ -279,7 +282,7 @@ func patchConfig(c *Context, w http.ResponseWriter, r *http.Request) {
 
 func makeFilterConfigByPermission(accessType filterType) func(c *Context, structField reflect.StructField) bool {
 	return func(c *Context, structField reflect.StructField) bool {
-		if structField.Type.Kind() == reflect.Struct {
+		if structField.Type.Kind() == reflect.Struct || structField.Name == FeatureFlagsElementName {
```
mkraft, line 285: Can we use an `access` tag on the `FeatureFlags` struct rather than adding this conditional logic?

sbishel: If we use the `access` tag we would either - 
- need to include all possible system console permissions
or
- only include the one necessary, but it would need updated if we used FF in `admin_definintion`, leaving us vulnerable for this bug again.
or
- introduce the concept of `access:any` 
I just thought of the last one and will explore it.

sbishel: Implemented `access:any`

api4/config.go
```diff
@@ -320,6 +323,16 @@ func makeFilterConfigByPermission(accessType filterType) func(c *Context, struct
 			if tagValue == model.ConfigAccessTagCloudRestrictable {
 				continue
 			}
+			if tagValue == AnyTagValue {
+				checkPermissions := model.SysconsoleReadPermissions
+				if accessType == FilterTypeWrite {
+					checkPermissions = model.SysconsoleWritePermissions
+				}
```
hahmadia, line 330: I think this is not needed. A system admin is given feature flags in the config as long as they have access to the system console. If they have access to the system console, they must at the very least have read permissions. If they have write, they will always have READ+WRITE. So we can just check for sys console read permissions and not worry about the write.

api4/config.go
```diff
@@ -320,6 +323,16 @@ func makeFilterConfigByPermission(accessType filterType) func(c *Context, struct
 			if tagValue == model.ConfigAccessTagCloudRestrictable {
 				continue
 			}
+			if tagValue == AnyTagValue {
```
hahmadia, line 326: Lets add maybe a comment here describing what the any tag is. It took me a while to understand it.
```suggestion
                        // If a config setting has a access tag of any, it means that this partcular config setting will always be return to a system console admin no matter what
	if tagValue == AnyTagValue {
```
Not the best wording but an explanation of what the any tag is and its purpose would be nice for new developers. 
There is a blurb in config.go in model folder describing the permissions so maybe add an explanation there too? 
Off topic but I really think that blurb needs documentation outside of code.
image

model/config.go
```diff
@@ -3098,7 +3098,7 @@ type Config struct {
 	GuestAccountsSettings     GuestAccountsSettings
 	ImageProxySettings        ImageProxySettings
 	CloudSettings             CloudSettings  // telemetry: none
-	FeatureFlags              *FeatureFlags  `json:",omitempty"` // telemetry: none
+	FeatureFlags              *FeatureFlags  `access:"any" json:",omitempty"` // telemetry: none
```
hahmadia, line 3101: Any tag is fine but we definitely need documentation somewhere explaining what it is and how it works. At first my first intuition was that ANYBODY will get it no matter what but then I realized it's any system admin.

api4/config_test.go
```diff
@@ -90,6 +90,10 @@ func TestGetConfigWithAccessTag(t *testing.T) {
 	t.Run("Can read value with permission", func(t *testing.T) {
 		assert.Equal(t, mockVaryByHeader, cfg.RateLimitSettings.VaryByHeader)
 	})
+
+	t.Run("Contains Feature Flags", func(t *testing.T) {
+		assert.NotNil(t, cfg.FeatureFlags)
+	})
```
hahmadia, line 96: I feel that we should add few more tests confirming that we indeed do get the feature flags as long as we are some sort of system console admin.

Test 1 is get config as a system user, make sure we don't get it back.
Test 2 is do it as a read only admin, make sure we have feature flags.

hahmadia (commented):
Nice! Just some suggestions left.

api4/config.go
hahmadia: If you do decide to go the route of only checking read permissions, this could be simplified to
```suggestion
			if tagValue == AnyTagValue && hasPermissionsToAnySysconsoleRead(...) {
			     return true
```

model/config.go
sbishel: All access tags only apply to System admin

api4/config_test.go
sbishel: Test 1 results in a Permissions error, you must have at least one system console read permission, this is already tested in TestGetConfig().
Test 2 is what I added. This is a system user that gets a SYSTEM_CONSOLE_READ permission added to it. So with a single permission we get feature flags.

hahmadia (approved):
Read through responses. Changes look good. Thanks

model/config.go
mkraft: All access tags apply to all system roles.

mkraft: All access tags apply to all roles because they're based on the permissions.

Doesn't each field in `FeatureFlags` also need an access tag?

api4/config.go
```diff
@@ -320,6 +323,11 @@ func makeFilterConfigByPermission(accessType filterType) func(c *Context, struct
 			if tagValue == model.ConfigAccessTagCloudRestrictable {
 				continue
 			}
+			// The 'access' tag 'any' checks for any SYSCONSOLE read permission and grants access if any read permission is allowed.
+			if tagValue == AnyTagValue && c.App.SessionHasPermissionToAny(*c.App.Session(), model.SysconsoleReadPermissions){
+				return true
+			}
+
```
mkraft, line 330: All of the other `access` tags apply to both read and write. Is this `any` value only for reading?

mkraft (approved):

furqanmlk (approved):
image
image

LGTM

PR #17171, opened 2021-03-17
MM-33402: Invalidated the LRU cache for channel members after moderat…

mkraft:
…ion changes.

#### Summary

* Fixed a bug that required a cache purge or server restart for channel moderation changes to be correctly applied. See the internal ticket for further details.
* Adds a test for the case described by the ticket.
* Refactors some code that iterates all of the channel member pages for use in two places.

#### Ticket Link

https://mattermost.atlassian.net/browse/MM-33402

#### Release Note

```release-note
Fixed a bug that required a cache purge or server restart for channel moderation changes to be correctly applied.
```


app/channel_test.go
```diff
@@ -1901,6 +1904,34 @@ func TestPatchChannelModerationsForChannel(t *testing.T) {
...
+		}
+
+		mockSession := model.Session{UserId: user.Id}
+
+		_, err := th.App.PatchChannelModerationsForChannel(channel.DeepCopy(), addCreatePosts)
+		require.Nil(t, err)
+		require.True(t, th.App.SessionHasPermissionToChannel(mockSession, channel.Id, model.PERMISSION_CREATE_POST))
+
+		_, err = th.App.PatchChannelModerationsForChannel(channel.DeepCopy(), removeCreatePosts)
+		require.Nil(t, err)
```
streamer45, line 1932: We now tend to use `NoError` instead of `Nil`.

app/channel.go
```diff
@@ -3007,6 +3009,19 @@ func (a *App) ClearChannelMembersCache(channelID string) {
...
+
+func (a *App) ClearChannelMembersCache(channelID string) {
+	clearSessionCache := func(channelMember model.ChannelMember) error {
+		a.ClearSessionCacheForUser(channelMember.UserId)
+		message := model.NewWebSocketEvent(model.WEBSOCKET_EVENT_CHANNEL_MEMBER_UPDATED, "", "", channelMember.UserId, nil)
+		message.Add("channelMember", channelMember.ToJson())
+		a.Publish(message)
+		return nil
+	}
+	a.forEachChannelMember(channelID, clearSessionCache)
```
streamer45, line 3024: Should we log the returned error (if any)?

streamer45 (approved):
Nice :+1: I really appreciated the functional vibe of the refactor.

app/channel_test.go
mkraft: `(*App).PatchChannelModerationsForChannel` returns a `*model.AppError`, and I think it's like this:

* `error` interfaces should `require.Error` or `require.NoError`, and
* `model.AppError` interfaces should `require.Nil` or `require.NotNil`

Unless I'm wrong?

app/channel.go
mkraft: Nothing in the `clearSessionCache` function returns an error, so it'll never return one in this case.

streamer45: I was thinking about `GetMembers()` failing. Before the refactor we were logging that.

app/channel_test.go
streamer45: Oh interesting. I was convinced `Error/NoError` did actual checks on the interface but instead they merely check for nil equality. Looks good then :+1:

app/channel.go
mkraft: Good point. Added handling for it.

sbishel (approved):

furqanmlk (approved):
image
image

Found different issue while testing. unable to delete post reaction via API. I will log different issue for this,

PR #17172, opened 2021-03-18
MM-34000: Use non-epoll mode for TLS connections

agnivade:
A *crypto/tls.Conn does not expose the underlying TCP connection
or even a File method to get the underlying file descriptor
like the way a *net/TCPConn does. Therefore the netpoll code would
fail to get the file descriptor.

Relevant issue here: https://github.com/mailru/easygo/issues/3

It is indeed possible to use reflect black magic to get the unexported
member, but I have found unexpected errors during writing to the websocket
by getting the file descriptor this way. I do not want to spend time investigating
this especially since this is already released.

Once this is out, we can decide on the right way to fix this, most probably
by proposing to expose the File method or some other way.

https://mattermost.atlassian.net/browse/MM-34000

```release-note
Fix an issue where websockets wouldn't work with TLS connections.
In that case, we just fall back to the way it works for Windows machines,
which is to use a separate goroutine for reader connection.
```

### DIFF
--- a/api4/websocket.go
+++ b/api4/websocket.go
@@ -5,7 +5,6 @@ package api4
 
 import (
 	"net/http"
-	"runtime"
 	"time"
 
 	"github.com/gobwas/ws"
@@ -49,7 +48,7 @@ func connectWebSocket(c *Context, w http.ResponseWriter, r *http.Request) {
 		c.App.HubRegister(wc)
 	}
 
-	if runtime.GOOS == "windows" {
+	if !wc.Epoll() {
 		wc.BlockingPump()
 	} else {
 		go wc.Pump()
--- a/app/web_conn.go
+++ b/app/web_conn.go
@@ -36,6 +36,8 @@ const (
 	webConnMemberCacheTime = 1000 * 60 * 30 // 30 minutes
 )
 
+var errNonEpollConnClose = errors.New("connection closed")
+
 // WebConn represents a single websocket connection to a user.
 // It contains all the necessary state to manage sending/receiving data to/from
 // a websocket.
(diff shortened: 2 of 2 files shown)


streamer45 (approved):
Looks good :+1: Good thing we have kept the other implementation around :)

nronas (approved):
Nice finding. LGTM

agnivade:
@lindalumitchell - I would like this to be tested out by an SDET since this was a regression from a release. The test steps are mentioned in the ticket. Please feel free to assign to anybody.

saturninoabril:
@agnivade Did the steps in the ticket and was able to setup the TLS in local. TLS is verified working both in Chrome and Firefox.
Screen Shot 2021-03-18 at 10 59 52 PM
Screen Shot 2021-03-18 at 11 00 14 PM

But got error on user login and other user activities, like below:
- `ERRO[2021-03-18T23:06:05.624897+08:00] http: panic serving [::1]:60178: could not get file descriptor`
Screen Shot 2021-03-18 at 11 07 43 PM

saturninoabril:
Hmm, for some reason, my config was reset. I didn't know why. But now, I can confirm that my settings are correct (see below).
Screen Shot 2021-03-18 at 11 21 05 PM

With server restart,
1. When TLS is set, the original error on file descriptor is no longer showing up.
2. However, when TLS is set, I'm seeing error log `... websocket.read: error during writing to websocket ...` (See screenshot. Can be seen easily on user logout).
Screen Shot 2021-03-18 at 11 34 53 PM
3. Note that item 2 error didn't show up when TLS is not set.

saturninoabril (approved):
@agnivade Thanks for updating. Tested and passed. No longer seeing the error logs related to websocket, and TLS and websocket connection works as expected.

amyblais:
@agnivade Please also cherry-pick to ``cloud`` in addition to v5.33.

agnivade:
Will do.

agnivade:
Btw, I also cherry-picked another important bug fix which got missed out to be backported to 5.33 (https://github.com/mattermost/mattermost-server/pull/17026). Both of them are part of https://github.com/mattermost/mattermost-server/pull/17179.

The cloud branch already had the other fix.

PR #17174, opened 2021-03-18
MM-34002: Improve AddUserToChannel

agnivade:
When we would add a user to a channel, we would
check whether the user is removed from that team or not.

During LDAP sync, or first login sync of an LDAP user, this check is not required because the
team member would have just been created. Hence, we
pass a boolean flag to bypass the check.

And with that done, we can freely query the replica.

https://mattermost.atlassian.net/browse/MM-34002

```release-note
NONE
```


isacikgoz (approved):
LGTM 👍

agnivade:
ping @mkraft

app/app_iface.go
```diff
@@ -41,12 +41,22 @@ type AppIface interface {
 	ListAutocompleteCommands(teamID string, T i18n.TranslateFunc) ([]*model.Command, *model.AppError)
 	// @openTracingParams teamID, skipSlackParsing
 	CreateCommandPost(post *model.Post, teamID string, response *model.CommandResponse, skipSlackParsing bool) (*model.Post, *model.AppError)
+	// AddChannelMember adds a user to a channel. It is a wrapper over AddUserToChannel.
+	// checkTeamMember is used to indicate whether it will check for the user was deleted from
+	// the team or not. This is useful to avoid in scenarios when we just added the team member,
+	// and thereby know that there is no need to check this.
+	AddChannelMember(userID string, channel *model.Channel, userRequestorId string, postRootId string, checkTeamMember bool) (*model.ChannelMember, *model.AppError)
```
mkraft, line 48: 5 feel like a lot of parameters. I'm thinking this (and `AddUserToChannel`) could take a struct with a boolean field where the default `false` value follows the higher data integrity code path. Maybe something like `unsafeTeamMemberDataIntegrity`. That way if you're using the method you can ignore that property in most cases, and if you don't ignore it you see "unsafe" and don't set it to `true` unless you look into it and know what you're doing.

agnivade: I like it. :+1:

agnivade: I refactored `AddChannelMember`. But kept `AddUserToChannel` as-is. It didn't seem like a huge benefit changing one parameter to a struct. But let me know if you want me to do that too. 0/5.

app/app_iface.go
```diff
@@ -41,12 +41,22 @@ type AppIface interface {
...
 	// The conditional blocks ensure that it sets those cursor IDs immediately as afterPost, beforePost or empty,
 	// and only query to database whenever necessary.
 	AddCursorIdsForPostList(originalList *model.PostList, afterPost, beforePost string, since int64, page, perPage int, collapsedThreads bool)
 	// AddPublicKey will add plugin public key to the config. Overwrites the previous file
 	AddPublicKey(name string, key io.Reader) *model.AppError
+	// AddUserToChannel adds a user to a given channel.
+	// checkTeamMember is used to indicate whether it should be checked that a user has
+	// already been removed from that team or not. This is useful to avoid in scenarios
+	// when we just added the team member, and thereby know that there is no need to check this.
+	AddUserToChannel(user *model.User, channel *model.Channel, doNotCheckTeamMember bool) (*model.ChannelMember, *model.AppError)
```
mkraft, line 59: What do you think about something like this to avoid the double-negative? In the `false` case it'll read like "don't skip team member integrity check" vs. "don't do not check team member":

```suggestion
	AddUserToChannel(user *model.User, channel *model.Channel, skipTeamMemberIntegrityCheck bool) (*model.ChannelMember, *model.AppError)
```

agnivade: Sounds good to me.

mkraft (approved):

furqanmlk (approved):
- Run an LDAP sync (verify that users are added to a channel correctly) ✅ 
- Run a sanity test by joining a channel, inviting other users to a channel. ✅ 

image

image

image

PR #17182, opened 2021-03-18
[MM-33826] add more detailed S3 error messages

maxerenberg:
#### Summary
This PR adds more details error messages when an S3 connection test fails in the system console.

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-33826
https://mattermost.atlassian.net/browse/MM-32679

```release-note
NONE
```


maxerenberg:
Here are some screenshots of the new sysconsole error messages:
**Bucket does not exist**
Screen Shot 2021-03-18 at 1 03 04 PM
**Invalid S3 credentials**
Screen Shot 2021-03-18 at 1 16 45 PM

app/server.go
```diff
@@ -531,7 +531,7 @@ func NewServer(options ...Option) (*Server, error) {
 	} else {
 		nErr := backend.TestConnection()
 		if nErr != nil {
-			if errors.Is(nErr, filestore.ErrNoS3Bucket) {
+			if _, ok := nErr.(*filestore.S3FileBackendNoBucketError); ok {
```
mkraft, line 534: Can we not use `errors.Is` with `filestore.S3FileBackendNoBucketError`?

shared/filestore/s3store.go
```diff
@@ -36,13 +36,18 @@ type S3FileBackend struct {
 	client     *s3.Client
 }
 
+type S3FileBackendAuthError struct {
+	DetailedError string
```
mkraft, line 40: What's the reason this needs a `DetailedError` field?

shared/filestore/s3store.go
```diff
@@ -148,19 +161,19 @@ func (b *S3FileBackend) TestConnection() error {
...
 		exists, err = b.client.BucketExists(context.Background(), b.bucket)
 		if err != nil {
-			return errors.Wrap(err, "unable to check if the s3 bucket exists")
+			return &S3FileBackendAuthError{DetailedError: "unable to check if the S3 bucket exists"}
 		}
 	}
 
 	if !exists {
-		return ErrNoS3Bucket
+		return &S3FileBackendNoBucketError{}
```
mkraft, line 176: Are we not able to define this in one line with `errors.New("no such bucket")`?

app/server.go
maxerenberg: Using `errors.Is` would require `filestore.S3FileBackendNoBucketError` to be an actual variable, not a type. I decided to use a type so that I could easily use a type switch in `connectionTestErrorToAppError` in `app/file.go`.

shared/filestore/s3store.go
maxerenberg: The `DetailedError` field is used to determine which operation caused the call to S3 to fail (see the `TestConnection` function). This is useful information because it tells the site admin exactly which AWS permission they are missing.

shared/filestore/s3store.go
maxerenberg: As mentioned above, I use a type switch in `app/file.go` to determine which error message to display to the user.

sbishel (approved):

mkraft (approved):

furqanmlk (approved):
image


image

PR #17183, opened 2021-03-18
Improve document extraction and including a document extraction command

jespino:
#### Summary
Improve document extraction and including a document extraction command

#### Ticket Link
Not yet

#### Release Note
```release-note
Add a command line document extraction command that allows indexing documents by conent
```


jespino:
@ogi-m again this is strictly part of the file search functionality that is behind a feature flag, this should be QA as part of the frontend PR.

app/file.go
```diff
@@ -773,22 +774,8 @@ func (a *App) UploadFileX(channelID, name string, input io.Reader,
...
-			})
-			if err != nil {
-				mlog.Error("Failed to extract file content", mlog.Err(err))
-				return
-			}
-			if storeErr := a.Srv().Store.FileInfo().SetContent(infoCopy.Id, text); storeErr != nil {
-				mlog.Error("Failed to save the extracted file content", mlog.Err(storeErr))
-			}
+			err := a.ExtractContentFromFileInfo(&infoCopy)
+			mlog.Error("Failed to extract file content", mlog.Err(err), mlog.String("fileInfoId", infoCopy.Id))
```
reflog, line 778: where did the 'if err != nil' go? :)

app/file.go
```diff
@@ -1043,22 +1030,8 @@ func (a *App) DoUploadFileExpectModification(now time.Time, rawTeamId string, ra
...
```
reflog, line 1034: where did the 'if err != nil' go? :)

app/upload.go
```diff
@@ -299,16 +298,8 @@ func (a *App) UploadData(us *model.UploadSession, rd io.Reader) (*model.FileInfo
...
-				ArchiveRecursion: *a.Config().FileSettings.ArchiveRecursion,
-			})
-			if err != nil {
-				mlog.Error("Failed to extract file content", mlog.Err(err))
-				return
-			}
-			if storeErr := a.Srv().Store.FileInfo().SetContent(infoCopy.Id, text); storeErr != nil {
-				mlog.Error("Failed to save the extracted file content", mlog.Err(storeErr))
-			}
+			err := a.ExtractContentFromFileInfo(&infoCopy)
```
reflog, line 301: same

reflog (changes requested):
looks great except for the places where err check is missing

cmd/mattermost/commands/extract_content.go
```diff
@@ -0,0 +1,86 @@
...
+	Short:   "Extracts the documents content",
+	Long:    "Extracts the documents content and stores it in the database for document searche",
+	Example: "extract-documents-content --from=12345",
+	RunE:    extractContentCmdF,
+}
+
+func init() {
+	ExtractContentCmd.Flags().Int64("from", 0, "The timestamp of the earliest file to extract, expressed in seconds since the unix epoch.")
+	ExtractContentCmd.Flags().Int64("to", model.GetMillis(), "The timestamp of the latest file to extract, expressed in seconds since the unix epoch.")
+	RootCmd.AddCommand(ExtractContentCmd)
```
ashishbhate, line 27: How about using `Duration` instead of `int64` for user input? It accepts all values that `time.ParseDuration` can parse so its a lot friendlier for the end user than specifying unix epochs.

jespino: This is very specific one-time-thing that shouldn't be used frequently. And it is consistent with the current interface for other cases. The idea of add a "duration" instead of the `to` makes it a bit weird because you are mixing both, start time and duration instead of start time and end time.

This is consistent with the `export` command that use epoch milliseconds for define from where to export.

app/file.go
jespino: Done

app/upload.go
jespino: Done

app/file.go
jespino: Done

cmd/mattermost/commands/extract_content.go
```diff
@@ -0,0 +1,86 @@
...
+	if err != nil {
+		return errors.New("from flag error")
+	}
+	if startTime < 0 {
+		return errors.New("from must be a positive integer")
+	}
+
+	endTime, err := command.Flags().GetInt64("to")
+	if err != nil {
+		return errors.New("to flag error")
```
reflog, line 51: small nit, maybe quote the word 'to' so that the message is clearer?

reflog (approved):

cmd/mattermost/commands/extract_content.go
jespino: Done!

esethna:
@ashishbhate @ogi-m please help with re-review on this, aiming to merge this as soon as possible.

ashishbhate (approved):

ogi-m (approved):
Will be QA'ed as part of the front end PR

PR #17185, opened 2021-03-18
[MM-33853] remove CSV row limit in compliance monitoring

maxerenberg:
#### Summary
This PR removes the 30,000 row limit in the CSV files for Compliance Monitoring.

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-33853

#### Release Note
```release-note
Compliance Monitoring CSV files are no longer limited to 30,000 rows.
```


hahmadia (commented):
Code looks fine but I'm just curious if introducing offset is gonna slow down especially since this is being done a large dataset. Not sure if this will have a performance impact.
Take a read of this article 
https://use-the-index-luke.com/no-offset

Let me know your thoughts.

maxerenberg:
> Code looks fine but I'm just curious if introducing offset is gonna slow down especially since this is being done a large dataset. Not sure if this will have a performance impact.
> Take a read of this article
> https://use-the-index-luke.com/no-offset
> 
> Let me know your thoughts.

Interesting. I've never thought about that before. I'll give it a shot. We should probably also be doing this for data retention when deleting posts in batches.

maxerenberg:
On second thought, the `ComplianceExport` function is actually returning a UNION of two different tables, so I don't think keyset pagination would be feasible, unless we passed *two* keysets, one for each table.

hahmadia:
> On second thought, the `ComplianceExport` function is actually returning a UNION of two different tables, so I don't think keyset pagination would be feasible, unless we passed _two_ keysets, one for each table.

@maxerenberg 

In your opinion, is that worth doing if it could vastly improve performance? I'm actually not sure but is there anyway you could mock it up where you have your implementation vs passing two id's and compare the performance? 

I'm very curious and wonder if it may be worth it considering it's potential performance improvement.
Watch this video:
https://www.youtube.com/watch?v=WDJRRNCGIRs

It shows how the possible performance could be severalfolds. 
I'd love to see if we can mock it up on a smaller data set in our example and see if it gives us any leverage or not.

hahmadia (commented):
Left a comment

hahmadia:
Before I forget, I'd be in favor of coding both methods up and testing it on a smaller dataset (100,000 posts?) and reporting the results. I'm very interested to see if there is any difference.

sbishel (commented):
I think the code looks good. However, I do think we should run this against our performance tests and get a benchmark or idea of the performance of this job.

hahmadia (commented):
Looks really good!
My only question is that compliance export should only deal with posts and DM's, not GM's (Group Messages), right?

store/sqlstore/compliance_store.go
```diff
@@ -151,17 +149,38 @@ func (s SqlComplianceStore) ComplianceExport(job *model.Compliance) ([]*model.Co
...
+		if len(channelPosts) < limit {
+			cursor.ChannelsQueryCompleted = true
+		} else {
+			cursor.LastChannelsQueryPostCreateAt = channelPosts[len(channelPosts)-1].PostCreateAt
+		}
+	}
+
+	var directMessagePosts []*model.CompliancePost
+	directMessagesQuery := ""
+	if !cursor.DirectMessagesQueryCompleted && len(channelPosts) < limit {
```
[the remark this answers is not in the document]
maxerenberg: The idea behind this clause is to return rows from two tables in one function call.
Let's say there are 3 channel posts, and 3 DMs. The caller is retrieving posts in batches of 2.
The first call will return 2 channel posts; the second call will return 1 channel post and 1 DM; the third call will return 2 DMs.

maxerenberg:
> Looks really good!
> My only question is that compliance export should only deal with posts and DM's, not GM's (Group Messages), right?
> DO you know why that is? I'm a bit curious. Otherwise, I think it's just about doing the performance tests as Scott mentioned.

I'm not actually sure. @sbishel do you have any ideas?

sbishel:
> > Looks really good!
> > My only question is that compliance export should only deal with posts and DM's, not GM's (Group Messages), right?
> > DO you know why that is? I'm a bit curious. Otherwise, I think it's just about doing the performance tests as Scott mentioned.
> 
> I'm not actually sure. @sbishel do you have any ideas?

No I think all messages are included in both compliance export or compliance monitoring. I think System messages are omitted from compliance export.

store/sqlstore/compliance_store.go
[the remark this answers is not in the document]
hahmadia: I know , I just meant because above you do 
```
		if len(channelPosts) < limit {
			cursor.ChannelsQueryCompleted = true
		}
```
then you don't need to recheck the condition and just use the variable which has a name to it and makes it more clean what len(channelPosts) < limit really means. that's what I was suggesting here.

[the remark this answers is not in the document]
maxerenberg: I see. I personally think that `len(channelPosts) < limit` is clearer because it shows that we are trying to "fill up" the limit specified by the caller.

sbishel:
@maxerenberg Have we been able to do any type of performance testing on this?

store/sqlstore/compliance_store.go
```diff
@@ -81,50 +81,48 @@ func (s SqlComplianceStore) Get(id string) (*model.Compliance, error) {
...
-				keywordQuery += "LOWER(Posts.Message) LIKE :Keyword" + strconv.Itoa(index)
-			}
-
-			props["Keyword"+strconv.Itoa(index)] = "%" + keyword + "%"
+			clauses[i] = "LOWER(Posts.Message) LIKE :Keyword" + strconv.Itoa(i)
+			props["Keyword"+strconv.Itoa(i)] = "%" + keyword + "%"
 		}
 
-		keywordQuery += ")"
+		keywordQuery = "AND (" + strings.Join(clauses, " OR ") + ")"
```
sbishel, line 98: Nice update.

sbishel (approved):

maxerenberg:
> @maxerenberg Have we been able to do any type of performance testing on this?

@hahmadia @sbishel 
Claudio gave me a SQL dump with 3M posts, which I used for the tests below. I generated some compliance reports with and without keyset pagination for small, medium and large subsets of the posts. Here are the execution times, in seconds, for each run. The batch size was 30,000 posts.

| # of posts | With keyset pag. | Without keyset pag. |
| ---------: | ----------------: | -------------------: |
| 30,000     | 1.601                    | 1.610                         |
| 300,000  |  16.538                 | 20.754                      |
| 3,000,000 | 165.404             |  934.949                   |

maxerenberg:
The load test also helped me discover a bug. It's possible for two posts to have the exact same timestamp, down to the nearest millisecond. Using `CreateAt` for keyset pagination caused some posts to be skipped. I'll think of a solution.

mkraft:
> The load test also helped me discover a bug. It's possible for two posts to have the exact same timestamp, down to the nearest millisecond. Using `CreateAt` for keyset pagination caused some posts to be skipped. I'll think of a solution.

Yes, to be deterministic is must be sorted on a unique identifier such as `Posts.Id`. It seems like this may encounter the same issues @nronas encountered in https://github.com/mattermost/mattermost-server/pull/16527.

nronas:
> The load test also helped me discover a bug. It's possible for two posts to have the exact same timestamp, down to the nearest millisecond. Using CreateAt for keyset pagination caused some posts to be skipped. I'll think of a solution.

Yeah, basically what @mkraft referenced. Ordering in SQL is not deterministic on collisions. So the best approach is for your ordering to use a compound(multiple columns) uniqueness.

store/sqlstore/compliance_store.go
```diff
@@ -187,24 +212,34 @@ func (s SqlComplianceStore) ComplianceExport(job *model.Compliance) ([]*model.Co
...
+				)
 				AND Posts.CreateAt <= :EndTime
 				` + emailQuery + `
-				` + keywordQuery + `)
-		ORDER BY PostCreateAt
-		LIMIT 30000`
-
-	var cposts []*model.CompliancePost
+				` + keywordQuery + `
+		ORDER BY Posts.CreateAt, Posts.Id
```
mkraft, line 228: @maxerenberg What's the performance like after adding `Posts.Id` to the `ORDER BY` clause?

maxerenberg: @mkraft
Here are the new results:

| # of posts | With keyset pag. | Without keyset pag. |
| ---------: | ----------------: | -------------------: |
| 30,000     | 1.594                    | 1.610                         |
| 300,000  |  16.599                 | 20.754                      |
| 3,000,000 | 336.325             |  934.949                   |

Although adding Posts.Id slowed down the keyset pagination considerably, it's still much faster than using OFFSET + LIMIT.

mkraft: Nice @maxerenberg! Results looks great. We can totally move ahead with this 👍🏻

furqanmlk (approved):
Working as expected

image

amyblais:
@cwarnermm This adds a new config setting ``BatchSize``. Was it confirmed it doesn't need docs?

cwarnermm:
@amyblais - The docs were updated for this PR via https://github.com/mattermost/docs/pull/4543, and the Docs/Done label is applied. Are you seeing something that suggests that this PR is marked as not needing docs?

amyblais:
@cwarnermm I saw the PR, but I don't think the config setting ``BatchSize`` is included in it.

cwarnermm:
Thanks, @amyblais. This work was split between 2 PRs, and the ``BatchSize`` work was completed via https://github.com/mattermost/docs/pull/4497. Sorry about the confusion on this one.

PR #17186, opened 2021-03-18
[MM-31899] Use a custom user-agent when previewing links

elyscape:
#### Summary
Many websites block requests made with Go's default user-agent. We had previous special-cased Twitter links to use a nonstandard user-agent. This makes that behavior apply everywhere and also customizes the user-agent to belong specifically to Mattermost.

**Note**: Ideally, we should also include a link to a Mattermost-hosted webpage describing the bot and its behavior (e.g. use a user-agent string like `Mattermost-Bot/1.1 (+https://about.mattermost.com/default-bot-policy/)`). See for example the pages used (and linked) by [Facebook][facebook] and [Slack][slack]. If you provide me with a URL that I should include, I'll add it to the PR. I'm also open to alternative suggestions for the user-agent.

#### Ticket Link
Fixes https://github.com/mattermost/mattermost-server/issues/16665
https://mattermost.atlassian.net/browse/MM-31899

#### Release Note
```release-note
* Fixed link previews on a number of websites, including Reddit
```

[facebook]: http://www.facebook.com/externalhit_uatext.php
[slack]: https://api.slack.com/robots


hmhealey:
Thanks for the PR, @elyscape! This looks awesome. I'm looking forward to not impersonating someone else to fetch link previews since that always felt hacky.

We actually already use (or at least should be using) a custom user agent already. It's a bit hidden, but all of our external requests go through the HTTP service which uses `mattermost-VERSION` as a user agent (defined [here](https://github.com/mattermost/mattermost-server/blob/master/services/httpservice/client.go) and actually used [here](https://github.com/mattermost/mattermost-server/blob/master/services/httpservice/transport.go#L18)). That said, it obviously doesn't work for sites like Twitter.

Playing around with the curl commands you posted in Jira, it seems like Twitter only responds with OpenGraph metadata when the user agent includes "bot" somewhere in it, so I like the `Mattermost-Bot/VERSION` one you've set up. Having a slash before the version also seems more idiomatic for a user agent anyway.

I'll ask around to see if anyone has any opinions about changing the user agent elsewhere, but I'd be on board with using the new user agent for all requests that go through the HTTP service instead of just being used for fetching link previews.

elyscape:
> We actually already use (or at least should be using) a custom user agent already. It's a bit hidden, but all of our external requests go through the HTTP service which uses `mattermost-VERSION` as a user agent (defined [here](https://github.com/mattermost/mattermost-server/blob/master/services/httpservice/client.go) and actually used [here](https://github.com/mattermost/mattermost-server/blob/master/services/httpservice/transport.go#L18)). That said, it obviously doesn't work for sites like Twitter.

Neato. As an aside, I would recommend against including the server version in the user agent, as that could provide a way for a malicious actor to gain information about private deployments.

hmhealey:
That's a very good point. I checked with our security team, and they agreed with you there. I don't think we've had any vulnerabilities involving injecting stuff through a link preview, but that's definitely something I could see happening in theory, so giving the external web server less info makes sense.

Also I checked,, and no one thinks changing the user agent for all external requests would cause any problems. Could you move the new user agent to the HTTPService then instead of just using it for link previews?

elyscape:
> Could you move the new user agent to the HTTPService then instead of just using it for link previews?

Sure, but I should note that it doesn't seem to actually take effect everywhere. It's currently defined [here][define] and applied [here][apply], but post link previews currently have a user-agent of `Go-http-client/1.1`, despite [seeming to use the custom transport][use]. Any idea why that is?

[define]: https://github.com/mattermost/mattermost-server/blob/v5.33.3/services/httpservice/client.go#L89
[apply]: https://github.com/mattermost/mattermost-server/blob/v5.33.3/services/httpservice/transport.go#L17-L21
[use]: https://github.com/mattermost/mattermost-server/blob/v5.33.3/app/post_metadata.go#L408-L411

elyscape:
Update: it's because of [this line][extraction], which pulls out the underlying `Transport` object. Not doing that should work fine.

[extraction]: https://github.com/mattermost/mattermost-server/blob/v5.33.3/app/post_metadata.go#L411

app/post_metadata.go
```diff
@@ -422,12 +422,7 @@ func (a *App) getLinkMetadata(requestURL string, timestamp int64, isNewPost bool
 
 		client := a.HTTPService().MakeClient(false)
 		client.Timeout = time.Duration(*a.Config().ExperimentalSettings.LinkMetadataTimeoutMilliseconds) * time.Millisecond
-		mmTransport := a.HTTPService().MakeTransport(false)
-		client.Transport = mmTransport.Transport
-
-		if strings.HasPrefix(requestURL, "https://twitter.com/") || strings.HasPrefix(requestURL, "https://mobile.twitter.com/") {
-			request.Header.Add("User-Agent", "facebookexternalhit/1.1 (+http://www.facebook.com/externalhit_uatext.php)")
-		}
+		client.Transport = a.HTTPService().MakeTransport(false)
```
hmhealey, line 425: I don't think this line should be necessary any more. I think `MakeClient` sets `client.Transport` to `MakeTransport()` by default

hmhealey (changes requested):
Thanks for moving that! One more thing to change since we're now just using the default `Transport` set by `MakeClient`

app/post_metadata.go
elyscape: Good catch! Removing.

app/post_metadata.go
```diff
@@ -422,12 +422,6 @@ func (a *App) getLinkMetadata(requestURL string, timestamp int64, isNewPost bool
 
 		client := a.HTTPService().MakeClient(false)
 		client.Timeout = time.Duration(*a.Config().ExperimentalSettings.LinkMetadataTimeoutMilliseconds) * time.Millisecond
-		mmTransport := a.HTTPService().MakeTransport(false)
-		client.Transport = mmTransport.Transport
-
-		if strings.HasPrefix(requestURL, "https://twitter.com/") || strings.HasPrefix(requestURL, "https://mobile.twitter.com/") {
-			request.Header.Add("User-Agent", "facebookexternalhit/1.1 (+http://www.facebook.com/externalhit_uatext.php)")
-		}
```
hanzei, line 430: That code was added as a workaround for twitter links as twitter doesn't send open graph data expect for certain user-agents. Why was it removed?

hanzei (changes requested):
Thanks for the detailed discussion :+1:

app/post_metadata.go
elyscape: Because it's actually the other way around; Twitter sends the appropriate OpenGraph data for most user-agents, specifically blocking a few that they have determined are abusive (including `Go-http-client/1.1`). Furthermore, I have checked and `Mattermost-Bot/1.1` works properly on Twitter:
```shellsession
$ # Golang default user-agent
$ curl -s -H 'User-Agent: Go-http-client/1.1' https://twitter.com/Mattermost/status/1372270097012826114 | grep -o '<meta[^>]*property="og:[^>]*>'
<meta property="og:site_name" content="Twitter" />
$ # Current link preview user-agent
$ curl -s -H 'User-Agent: facebookexternalhit/1.1 (+http://www.facebook.com/externalhit_uatext.php)' https://twitter.com/Mattermost/status/1372270097012826114 | grep -o '<meta[^>]*property="og:[^>]*>'
<meta  property="og:type" content="article">
<meta  property="og:url" content="https://twitter.com/Mattermost/status/1372270097012826114">
<meta  property="og:title" content="Mattermost on Twitter">
<meta  property="og:image" content="https://pbs.twimg.com/media/EwtIWpqWgAEvGS6.jpg:large">
<meta  property="og:image:user_generated" content="true">
<meta  property="og:description" content="“NEW: Mattermost v5.33 is now available - https://t.co/AQICWKhpx0”">
<meta  property="og:site_name" content="Twitter">
$ # New custom user-agent
$ curl -s -H 'User-Agent: Mattermost-Bot/1.0' https://twitter.com/Mattermost/status/1372270097012826114 | grep -o '<meta[^>]*property="og:[^>]*>'
<meta  property="og:type" content="article">
<meta  property="og:url" content="https://twitter.com/Mattermost/status/1372270097012826114">
<meta  property="og:title" content="Mattermost on Twitter">
<meta  property="og:image" content="https://pbs.twimg.com/media/EwtIWpqWgAEvGS6.jpg:large">
<meta  property="og:image:user_generated" content="true">
<meta  property="og:description" content="“NEW: Mattermost v5.33 is now available - https://t.co/AQICWKhpx0”">
<meta  property="og:site_name" content="Twitter">
```
Removing this special case both simplifies the code and removes a behavior that is (IMO) not the best form.

hanzei: I've tested the changes locally and it seem like that twitter links are still shown.

hanzei (approved):
LGTM :+1:

hmhealey (approved):

jgilliam17 (approved):
Thank you @elyscape 
Tested, looks good to merge.
- Verified link previews for various websites, including Reddit posts.
@hmhealey Can you please help merge. Thanks 🙂

hmhealey:
Thanks again for the PR, @elyscape! Hopefully this'll stop new issues with the link previews from popping up in the future.

PR #17188, opened 2021-03-19
MM-34080: Removing sqlite entirely

agnivade:
The initial commit missed removing this blank import.
So the library still remained in our vendor directory.
Removing it for good now.

Bye bye sqlite.

https://mattermost.atlassian.net/browse/MM-34080


nronas (approved):
❤️

isacikgoz (approved):
👋

PR #17190, opened 2021-03-22
Fix for when the 'migrate' command drops the custom default language.…

haalcala:
Fix for when the 'migrate' command drops the custom default language. (Using suggested fix)

P.S. No ticket for this


isacikgoz (commented):
Hey @haalcala, thanks for the contribution!! Could you explain why we need to call `utils.TranslationsPreInit()` before this command? I couldn't understand the reason behinds this change.

haalcala:
> Hey @haalcala, thanks for the contribution!! Could you explain why we need to call `utils.TranslationsPreInit()` before this command? I couldn't understand the reason behinds this change.

Hi, this was the suggested fix, instead of the one I previously provided here https://github.com/mattermost/mattermost-server/pull/15617

mgdelacroix (approved):
Apologies for the late review @haalcala, LGTM! :+1: 

Thanks for the contribution!

isacikgoz (approved):
Thanks @haalcala 👍

PR #17206, opened 2021-03-23
Stop trying to build mattermost-redux in CI

hmhealey:
I forgot the server has to build the webapp, so this needs to be removed with the mattermost-redux migration.

#### Release Note
```release-note
NONE
```

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -28,7 +28,7 @@ jobs:
           export WEBAPP_GIT_COMMIT=$(git rev-parse HEAD)
           echo "$WEBAPP_GIT_COMMIT"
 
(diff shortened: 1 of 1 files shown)


metanerd (approved):

agnivade (approved):
Yay finally!

I remember Jenkins also had to do the same thing. Now that it's decomissioned, it's probably good to check with the private Gitlab and confirm the same change is done there too.

metanerd:
Good point @agnivade , will check with @hmhealey tomorrow!

hmhealey:
Is it using a different configuration than the one here? I remember you showed that to me, @metanerd, but I don't remember having to change it when I updated the webapp build image here recently. If it is though, I'll have to remember to update it when I update the Docker build image to Node 14 later this week.

PR #17207, opened 2021-03-23
[MM-34166] Fix Flaky Test TestGenerateSupportPacket

hahmadia:
#### Summary

Fix Flaky Test because the environment variable is not set which causes problem

#### Ticket Link

https://mattermost.atlassian.net/browse/MM-34166

#### Release Note

```release-note
None
```


wiggin77 (approved):
LGTM :+1:   Also confirmed tests now pass when running local via makefile.

agnivade (approved):
Interesting, so it was the effect of a previous test.

PR #17214, opened 2021-03-24
MM-34158: Handle pings from client

agnivade:
We respond with a pong on a ping message.

And we explicitly decode the message only on a text
frame.

We also make the code more robust and properly
handling fragmented messages.

```release-note
The server now echoes a pong frame in response
to a ping websocket frame from the client.
```

https://mattermost.atlassian.net/browse/MM-34158


agnivade:
Manually tested with:
- Matterhorn
- Load test (with -race)
- Additional manual testing with fragmented messages and very long messages, and compared with 5.32 as well.

PR #17215, opened 2021-03-24
MM-34171: Fix racy test TestSentry

agnivade:
The NewServer call sets the global mlog Info variable.
But before that, if we call UpdateConfig with the functional options,
then the store.Set method will call mlog.Info before
it could be set.

To fix that, we prepare the updated config
and pass that directly to NewServer to avoid
having to call UpdateConfig.

https://mattermost.atlassian.net/browse/MM-34171

```release-note
NONE
```


reflog (approved):

streamer45 (approved):
Nice :+1:

agnivade:
No QA required as this just fixes a racy test.

PR #17220, opened 2021-03-24
[MM-33934] Prevent User.Timezone field from overflowing column capacity

streamer45:
#### Summary

I took a long time to understand how invalid JSON could end up in a user's timezone field especially since we validate the request body on a call to `/user/patch`. I was finally able to reproduce today by sending more data than what would fit in the MySQL `varchar(256)` column. 

The user update works just fine but data gets truncated making it invalid on any future call that fetches the user.
Honestly I thought the db driver would return an error due to possible data loss/inconsistency but apparently that's not the case.

PR adds a check to `User.IsValid()` to prevent this issue going forward.

#### Ticket

https://mattermost.atlassian.net/browse/MM-33934

#### Release Note

```release-note
NONE
```

### DIFF
--- a/api4/user_test.go
+++ b/api4/user_test.go
@@ -1693,6 +1693,16 @@ func TestPatchUser(t *testing.T) {
 	user := th.CreateUser()
 	th.Client.Login(user.Email, user.Password)
 
+	t.Run("Timezone limit error", func(t *testing.T) {
+		patch := &model.UserPatch{}
+		patch.Timezone = model.StringMap{}
+		patch.Timezone["manualTimezone"] = string(make([]byte, model.USER_TIMEZONE_MAX_RUNES))
+		ruser, resp := th.Client.PatchUser(user.Id, patch)
+		CheckBadRequestStatus(t, resp)
+		require.Equal(t, "model.user.is_valid.timezone_limit.app_error", resp.Error.Id)
+		require.Nil(t, ruser)
+	})
+
 	patch := &model.UserPatch{}
 	patch.Password = model.NewString("testpassword")
 	patch.Nickname = model.NewString("Joram Wilander")
--- a/i18n/en.json
+++ b/i18n/en.json
@@ -8399,6 +8399,10 @@
     "translation": "Invalid locale."
   },
   {
+    "id": "model.user.is_valid.marshal.app_error",
+    "translation": "Failed to encode field to JSON"
+  },
+  {
     "id": "model.user.is_valid.nickname.app_error",
     "translation": "Invalid nickname."
   },
--- a/model/user.go
+++ b/model/user.go
@@ -56,6 +56,7 @@ const (
 	USER_NAME_MIN_LENGTH      = 1
 	USER_PASSWORD_MAX_LENGTH  = 72
 	USER_LOCALE_MAX_LENGTH    = 5
+	USER_TIMEZONE_MAX_RUNES   = 256
 )
 
 //msgp:tuple User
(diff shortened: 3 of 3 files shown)


agnivade (commented):
Great find!

It seems to me that this is a generic thing that will happen to any varchar column, and not necessarily just this one.

Investigating a bit more, it looks like this is due to an "amazing" MySQL feature called strict mode:

> If strict SQL mode is not enabled and you assign a value to a CHAR or VARCHAR column that exceeds the column's maximum length, the value is truncated to fit and a warning is generated. For truncation of nonspace characters, you can cause an error to occur (rather than a warning) and suppress insertion of the value by using strict SQL mode. See Section 5.1.10, “Server SQL Modes”.

https://dev.mysql.com/doc/refman/5.7/en/char.html

I am wondering if we can do some testing around setting strict mode from the code itself: https://dev.mysql.com/doc/refman/5.7/en/sql-mode.html#sql-mode-setting and see if that's a viable solution?

That would benefit all code, and the behavior is how you would expect it to be. Rather than us having to maintain two sources of truth for the max varchar column size - one in DB, another in code.

streamer45:
@agnivade 

Thanks for looking further into this. If you agree I think this is worth a dedicated ticket since we've been handling `varchar` fields at the app level since forever. This is possibly also due to supporting different db drivers/versions. For example, if you create a brand new installation, the timezone column datatype will be `text` while on existing ones (like community) it's still a `varchar(256)`. What we tend to do is to default to the minimum supported length/size to avoid any issue but in this case the validation code was missing.

agnivade:
Hmm .. good point on the `text` field. Re: different drivers, I think Postgres already throws an error, so we should be fine on that part.

I still think it might be good to do this, because for varchar fields, we have to duplicate the column size again from the DB layer to the model layer, and we have to maintain two sources of truth. Someone needs to "remember" to correctly update two places every time we change the size limit. Atleast, I don't see any harm in having an extra MySQL command. In case someone adds a new varchar column in future and forgets to add length validation, this might help.

streamer45:
> I still think it might be good to do this, because for varchar fields, we have to duplicate the column size again from the DB layer to the model layer, and we have to maintain two sources of truth. Someone needs to "remember" to correctly update two places every time we change the size limit. 

Couldn't that be simply fixed by having a shared constant at the store level?

> Atleast, I don't see any harm in having an extra MySQL command. In case someone adds a new varchar column in future and forgets to add length validation, this might help.

I am just not sure we'll be able to consistently (and successfully) set the strict mode (does the MM user have permissions?) and we clearly can't trust all administrators to have it set. Needs further investigation of course.

agnivade:
> Couldn't that be simply fixed by having a shared constant at the store level?

But then should all fields have the same length?

> I am just not sure we'll be able to consistently (and successfully) set the strict mode (does the MM user have permissions?) 

True, it needs SUPER permissions. Might be better to just document this, and let admins set it by themselves.

agnivade (approved):

streamer45:
> > Couldn't that be simply fixed by having a shared constant at the store level?
> 
> But then should all fields have the same length?

We would still have a constant per field/column but at least it would live in one place only (store) and both validations and schema creation logic would be using those constants instead of defining yet more values. It's not perfect but at least it should help to guarantee some consistency (between different installations) which we might not have if we rely solely on the db.

> Might be better to just document this, and let admins set it by themselves.

:+1:

agnivade:
>  and both validations and schema creation logic would be using those constants 

Yeah, but not for long though :stuck_out_tongue:  With the schema migrations, and removal of gorp, DB code will move into sql files. Anyways, the permission thing is a no-go. So it's better to just have it as a recommendation.

streamer45:
> > and both validations and schema creation logic would be using those constants
> 
> Yeah, but not for long though  With the schema migrations, and removal of gorp, DB code will move into sql files. Anyways, the permission thing is a no-go. So it's better to just have it as a recommendation.

True. Maybe we could take it a step further and ask the db. On startup we could generate a map of `column -> size` and use that throughout the application's lifetime to do validations. Food for thought. I'll open a ticket to keep track of all that has been discussed.

isacikgoz (approved):
Good finding!

streamer45:
Created [MM-34267](https://mattermost.atlassian.net/browse/MM-34267).

PR #17222, opened 2021-03-24
[MM-7968] restrict creation of direct channels to team members

maxerenberg:
#### Summary
If "Enable users to open Direct Message channels with" is set to "Any member of the team" in the system console, then a user will only be allowed to create direct channels with members who share a common team with the user.

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-7968

#### Release Note
```release-note
Team-restricted direct channel creation is now also applied to the backend. Previously, this was restricted to the frontend.
```


maxerenberg:
If a direct channel already existed beforehand, though, and one of the members leaves the common team, then the two members will still be allowed to send messages to each other.

maxerenberg:
We also need to update the frontend because there are a lot of places where a user could try to create a direct channel and but the API call will fail. For example:
* the /msg command
* a well-crafted URL (`/team_name/messages/other_username`)
* clicking on a user's profile then clicking "Send Message"

hahmadia:
> When a user clicks on a user's profile on the frontend, then click "Send Message" option is still there, but nothing will happen if it is clicked. We should remove that option if the two members are no longer part of a common team.

@maxerenberg Create a Jira ticket for this so we can keep track and not lose this.

app/team.go
```diff
@@ -1049,6 +1049,14 @@ func (a *App) GetTeamMembersByIds(teamID string, userIDs []string, restrictions
 	return teamMembers, nil
 }
 
+func (a *App) GetCommonTeamIDsForUsers(userIDs []string) ([]string, *model.AppError) {
```
hahmadia, line 1052: The naming here can be a little confusing. 
It meant to me for multiple users until I digged further and saw it's actually only meant between two users.
What about renaming it to something like `GetCommonTeamIDsForTwoUsers` or `GetCommonTeamIDsBetween2Users` or `GetCommonTeamIDsForDirectMessages`. Something that more explicitly mentions its for two users rather than having to dig into the the store method to see that if `len() != 2` then it throws error.

store/sqlstore/team_store.go
```diff
@@ -1504,6 +1504,32 @@ func (s SqlTeamStore) GetUserTeamIds(userId string, allowFromCache bool) ([]stri
...
+	const query = `
+	SELECT A.TeamId FROM (
+		SELECT TeamId FROM TeamMembers
+		WHERE UserId = :UserId0 AND DeleteAt = 0
+	) AS A INNER JOIN (
+		SELECT TeamId FROM TeamMembers
+		WHERE UserId = :UserId1 AND DeleteAt = 0
+	) AS B ON A.TeamId = B.TeamId
+	INNER JOIN Teams ON A.TeamId = Teams.Id
+	WHERE Teams.DeleteAt = 0`
```
hahmadia, line 1523: We are trying to avoid these sorts of query writing. I believe server team had initiative to use Squirrel to write queries.  For example, see `GetChannels` in `channel_store.go`. 
```Go
	query := s.getQueryBuilder().
		Select("Channels.*").
		From("Channels, ChannelMembers").
		Where(
			sq.And{
				sq.Expr("Id = ChannelId"),
				sq.Eq{"UserId": userId},
				sq.Or{
					sq.Eq{"TeamId": teamId},
					sq.Eq{"TeamId": ""},
				},
			},
		).
		OrderBy("DisplayName")
```
Try using squirrel to write out the query.
I'm sure there is examples in team_store too you can follow.

store/sqlstore/team_store.go
```diff
@@ -1504,6 +1504,32 @@ func (s SqlTeamStore) GetUserTeamIds(userId string, allowFromCache bool) ([]stri
...
+	props := map[string]interface{}{"UserId0": userIDs[0], "UserId1": userIDs[1]}
+	_, err := s.GetReplica().Select(&teamIDs, query, props)
```
hahmadia, line 1525: @sbishel 
In regards to the master/slave read replica lag issue, could we have a scenario where a user just joined a team that is common with another user but hasn't been replicated in the replica so potentially this query doesn't work? Something we need to be concerned with or nah?

hahmadia (commented):
Great! Just some questions and minor feedback.

app/team.go
maxerenberg: Good idea.

store/sqlstore/team_store.go
maxerenberg: I don't see any examples where squirrel is being used for nested queries. e.g. `getSearchGroupChannelsQuery` in `channel_store.go`. Trying to use squirrel for a nested query would likely require several SQL builders with multiple steps of error checking, which I think would make the code harder to read.

In general, I agree that using a query builder like squirrel can make a SQL query easier to read, especially if the shape of the query depends on certain parameters, but if the query is a constant string, and is formatted nicely, then I think just using that constant string is the most readable solution.

store/sqlstore/team_store.go
sbishel: We don't need to worry about the use case you mentioned because those are separate actions.

Where we need to worry about this is if its all part of the same action (transaction). IOW, if we added the user to a team, then created a direct message as part of the same process or function. Then this could be an issue.

sbishel (changes requested):
The changes look good however, I found a couple issues. This doesn't handle the slash command `/msg`. That path would need this update as well. 
https://github.com/mattermost/mattermost-server/blob/master/app/slashcommands/command_msg.go

I did notice a common function `UserCanSeeOtherUser`, I wonder if this is a better place for this logic. Based on the name, this seems like the place it should go. (I'm not sure what effects, good or bad this would have in other places.)

Also, when calling the API with a user that already has an existing DM, it now fails if no longer in the team. I realize this use case is under review, but wanted to point it out, since this is how the UI works.

store/sqlstore/team_store.go
hahmadia: ooooh true true true its only if its happening right after a write. Good call good call.

maxerenberg:
> The changes look good however, I found a couple issues. This doesn't handle the slash command `/msg`. That path would need this update as well.
> https://github.com/mattermost/mattermost-server/blob/master/app/slashcommands/command_msg.go
> 
> I did notice a common function `UserCanSeeOtherUser`, I wonder if this is a better place for this logic. Based on the name, this seems like the place it should go. (I'm not sure what effects, good or bad this would have in other places.)
> 
> Also, when calling the API with a user that already has an existing DM, it now fails if no longer in the team. I realize this use case is under review, but wanted to point it out, since this is how the UI works.

Good catch. Since the `GetOrCreateDirectChannel` function is being called from multiple endpoints, I think it makes most sense to place the check in the app layer rather than the API layer. 
I don't think `UserCanSeeOtherUser` is the right place to be checking if two users are in the same team, because, as the function name suggests, it only checks if they can see each other, which all members can do by default since they have the VIEW_MEMBERS  permission. I don't think it would be a good idea to change this function either because it's called from a lot of different places. So I just created a new function `GetCommonTeamIDsForTwoUsers`.
Placing this logic in the app layer also solves the problem where two users who have a DM but one user leaves the team, since the call to `GetCommonTeamIDsForTwoUsers` is now happening *after* the call to `getDirectChannel`.

store/sqlstore/team_store.go
hahmadia: Ok so I spent some time and was trying to think of a way to avoid nested sql. idk about others but personally for me I find it hard to read and a little bit hard to follow. I came up with this SQL example to avoid having nested. Is this suitable and does it work? Then we can take this and convert to squirrel. I'll still be approving this as I'm out next week and don't want to block this PR but have a thought about it and see if you can come up with a way to avoid nesting so we can use Squirrel ORM.

```
SELECT t.teamid
FROM   teammembers t
       INNER JOIN teammembers t2
               ON t.teamid = t2.teamid
       INNER JOIN teams t3
               ON t.teamid = t3.id
WHERE  t.userid = 'swrotanigt8gpg5wautz1s8zfo'
       AND t2.userid = '1igeyqqpu38f5mkwkyc7xhxqxy'
       AND t.deleteat = 0
       AND t3.deleteat = 0 
```

hahmadia: I'm just not a big fan of writing out raw SQL queries and I personally always try to avoid nested SQL too cause its just a bit hard to follow when you're trying to understand what the query is really doing. 
I'm 2.5-3/5 on trying to come up with a non nested query that can be converted to squirrel.
If you find it very hard or too time consuming, then I'm okay with the way it is. I don't want it to become a blocker.

hahmadia (approved):
LGTM 👍 
I'm approving as I'll be out next week and don't want to block on this PR's progress.
I've left a suggestion.

store/sqlstore/team_store.go
agnivade: The recommendation is to use squirrel for situations where you are dynamically constructing the query depending on several conditions.

For a simple straight query, anything is fine, since squirrel isn't giving you any advantage there. And for complicated queries, you need to spend more time conjuring the right squirrel commands. Normally people prefer raw SQL in these situations.

maxerenberg: @hahmadia 
😮  I like your query a lot better. And if I'm interpreting the EXPLAIN output correctly, Postgres is smart enough to filter each `TeamMembers` table first before performing the JOIN. Cool!

sbishel (approved):
Nice, I like the move to the app layer.

app/channel.go
```diff
@@ -330,6 +330,17 @@ func (a *App) GetOrCreateDirectChannel(userID, otherUserID string, channelOption
...
+	if *a.Config().TeamSettings.RestrictDirectMessage == model.DIRECT_MESSAGE_TEAM &&
+		!a.SessionHasPermissionTo(*a.Session(), model.PERMISSION_MANAGE_SYSTEM) {
+		commonTeamIDs, err := a.GetCommonTeamIDsForTwoUsers(userID, otherUserID)
+		if err != nil {
+			return nil, err
+		}
+		if len(commonTeamIDs) == 0 {
+			return nil, model.NewAppError("createDirectChannel", "api.channel.create_channel.direct_channel.team_restricted_error", nil, "", http.StatusForbidden)
+		}
+	}
```
mkraft, line 342: I'm not sure why we'd make an exception for the `PERMISSION_MANAGE_SYSTEM` permission here given that this setting is configured in the settings. If we want it to be part of the permissions system we could create a permission `message_non_intersecting_team_members`.

maxerenberg: The idea was that sysadmins could override the direct message policy, i.e. sysadmins can create direct channels between two arbitrary users. Or is this not something we want?

mkraft: The need for that functionality would be a PM question, but in terms of implementation: having the logic partially in the config and partial in the permission system seems like an exception that we should use sparingly.

mkraft: https://mattermost.atlassian.net/browse/MM-7968?focusedCommentId=110111

mkraft (approved):

furqanmlk (approved):
LGTM

PR #17303, opened 2021-03-30
MM-34086: Fix flaky test StartServerTLSOverwriteCipher

agnivade:
This suffered from the same race condition as
https://github.com/mattermost/mattermost-server/pull/17215.

We apply the same approach of using a memstore instead
of a file based config store.

It is likely that the test was also failing because of
the same race. Although I don't obviously see how, because
the race happens with mlog.Info. The test will fail only
if the config wasn't able to set properly. So although
not strictly related, it's somehow related.

If it fails again, we can look into it again. But this
atleast fixes the race.

While here, we also apply some cleanup of the code.

https://mattermost.atlassian.net/browse/MM-34086

```release-note
NONE
```

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -472,7 +472,7 @@ workflows:
           name: test-mysql
           dbdriver: mysql
           dbsource: mmuser:mostest@tcp(mysql:3306)/mattermost_test?charset=utf8mb4,utf8&multiStatements=true
-          racemode: "-race"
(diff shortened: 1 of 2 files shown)


isacikgoz (approved):
LGTM 👍

wiggin77 (approved):
LGTM :+1:    Hopefully I'll get time to remove the global logger soon.

PR #17307, opened 2021-03-30
MM-34487: Add dead queue

agnivade:
Just a circular buffer to store dead messages for now.
Not controlling this via config flag because this does
not have any effect except taking some more memory
per connection

```release-note
NONE
```

https://mattermost.atlassian.net/browse/MM-34487


agnivade:
This is dependent on the previous commits, but sending it out for review to get things moving. Just check the latest commit: https://github.com/mattermost/mattermost-server/pull/17307/commits/3d82452084f6e650a3039988359d80eac3d3f990

agnivade:
Ready for review.

app/web_conn.go
```diff
@@ -47,11 +48,18 @@ type WebConn struct {
...
-	connectionID              atomic.Value
-	endWritePump              chan struct{}
-	pumpFinished              chan struct{}
+	// deadQueue behaves like a queue of a finite size
+	// which is used to store all messages that are sent via the websocket.
+	// It basically acts as the user-space socket buffer, and is used
+	// to resuscitate any messages that might have got lost when the connection is broken.
+	// It is implemented by using a circular buffer to keep it fast.
+	deadQueue        []model.WebSocketMessage
+	deadQueuePointer int // Pointer which indicates the next slot to insert.
```
streamer45, line 57: Did you name it pointer to tease me or was it how you imagined it? :stuck_out_tongue: 
Interesting that you haven't used the more conventional *tail*.

streamer45 (approved):
Code changes looks good :+1: 

The only question/concern I have is on why we are merging this into master instead of having a "feature branch". I understand the previous PR was guarded by feature flag but here we are actually adding live code.

agnivade:
> I understand the previous PR was guarded by feature flag but here we are actually adding live code.

Yeah, this didn't make any functional impact. But you are right, we need to disable this when the full feature is ready. I will add a condition. There are going to be lots of if conditions btw :disappointed:

streamer45:
> > I understand the previous PR was guarded by feature flag but here we are actually adding live code.
> 
> Yeah, this didn't make any functional impact. But you are right, we need to disable this when the full feature is ready. I will add a condition. There are going to be lots of if conditions btw

Yeah, that's reasonable and a small price to pay to avoid unexpected issues I guess. Hopefully they are not going to be there too long.

app/web_conn.go
streamer45: No worries, just being pedantic :) It's clear enough with the comment.

PR #17320, opened 2021-04-01
MM-33746 hotfix

reflog:
#### Summary

Added default value to TotalMsgCountRoot column to avoid potential crash.

#### Release Note

```release-note
NONE
```


enahum (approved):

store/sqlstore/upgrade.go
```diff
@@ -1011,13 +1011,14 @@ func upgradeDatabaseToVersion535(sqlStore *SqlStore) {
 
 	sqlStore.CreateColumnIfNotExists("SidebarCategories", "Collapsed", "tinyint(1)", "boolean", "0")
 
-	sqlStore.CreateColumnIfNotExistsNoDefault("Channels", "TotalMsgCountRoot", "bigint", "bigint")
+	sqlStore.CreateColumnIfNotExists("Channels", "TotalMsgCountRoot", "bigint", "bigint", "0")
```
nronas, line 1014: Adding a column with a default value for PG versions < 11 will obtain an `ACCESS EXCLUSIVE` lock right? For customers that have big channels table the default value will slow down their boot up.

reflog: @nronas - I think that's a possibility. We did however add a release-note saying that migration can take up to a few minutes. cc @streamer45

nronas: Oh superb.

store/sqlstore/upgrade.go
```diff
@@ -1011,13 +1011,14 @@ func upgradeDatabaseToVersion535(sqlStore *SqlStore) {
...
 	sqlStore.CreateColumnIfNotExistsNoDefault("Channels", "LastRootPostAt", "bigint", "bigint")
 	defer sqlStore.RemoveColumnIfExists("Channels", "LastRootPostAt")
 
 	// note: setting default 0 on pre-5.0 tables causes test-db-migration script to fail, so this column will be added to ignore list
 	sqlStore.CreateColumnIfNotExists("ChannelMembers", "MsgCountRoot", "bigint", "bigint", "0")
 	sqlStore.AlterColumnDefaultIfExists("ChannelMembers", "MsgCountRoot", model.NewString("0"), model.NewString("0"))
+	sqlStore.AlterColumnDefaultIfExists("Channels", "TotalMsgCountRoot", model.NewString("0"), model.NewString("0"))
```
jespino, line 1021: Is this going to fix the existing nulls in the database? it is even needed?

reflog: @jespino yes, this is for the existing NULLs. Since this was already deployed on daily I thought this would be the right way to change instead of running queries manually

jespino (approved):

PR #17330, opened 2021-04-02
Remove direct store calls from api4/system_local.go

### LINKED ISSUE #15743: Remove direct Store calls from the Api layer for the file api4/system_local.go
We want to remove any direct access from the API layer into the Store layer, so
any call to the Store has to be done through the App layer.

This ticket is to transform all the direct calls to the store (search for
`Srv().Store....`) in the file api4/system_local.go.

This can imply using an existing `App` method, or creating a whole new `App`
method to access the store in that way.

The Api layer is placed in the `api4` directory and the `app` layer is placed
in the `app` directory.

If you have questions about the ticket or you need any help, feel free contact jesus.espino in https://community.mattermost.com/

msal4:
#### Summary

Remove direct Store calls from the API layer for the file api4/system_local.go

#### Ticket Link

Fixes #15743

#### Release Note

```release-note
NONE
```


agnivade (approved):

jespino (approved):
Looks good to me. Thanks @msal4! :tada:

PR #17339, opened 2021-04-06
Add missing wrapped errors

hanzei:
#### Summary
There are a couple of places where `model.NewAppError` doesn't wrap a returned `error`. This makes debugging harder.

I stumbled across this when debugging an issue in `app/user.go` and decided to fill all other occurrences that I found. The regex I've used is `err != nil \{\n\t+return model.NewAppError\("[a-zA-Z]+", "[a-zA-Z._]+", nil, ""`. Please note that this is surely not an _exhaustive_ fix, but rather a one time afford. 

#### Ticket Link
https://community-daily.mattermost.com/core/pl/3aya19tnmirabcuxjy7jznmejo

#### Release Note
```release-note
NONE
```

### DIFF
--- a/app/brand.go
+++ b/app/brand.go
@@ -62,7 +62,7 @@ func (a *App) SaveBrandImage(imageData *multipart.FileHeader) *model.AppError {
 	a.MoveFile(BrandFilePath+BrandFileName, BrandFilePath+t.Format("2006-01-02T15:04:05")+".png")
 
 	if _, err := a.WriteFile(buf, BrandFilePath+BrandFileName); err != nil {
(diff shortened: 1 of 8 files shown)


streamer45:
Awesome, thank you @hanzei :+1: 

> Please note that this is surely not an exhaustive fix, but rather a one time afford.

I am wondering how feasible it would be to write a vet check for this. In any case it's something to keep in mind when reviewing.

streamer45 (approved):

agnivade (approved):

hanzei:
@streamer45 That is surely feasible, but I've shied away from it due to the mana required.

streamer45:
No worries, that was more of an open question to ourselves (server team). If doable, it might be a good help-wanted candidate.

PR #17359, opened 2021-04-07
MM-34434: Trial license flag

mkraft:
#### Summary

* Added a hard-coded `is_trial` boolean key/value to the JSON payload that is POSTed to the create trial license endpoint.
* Added some tests to the existing create trial license endpoint.
* Fixed a bug in the API client.

Related PR: https://github.com/mattermost/license-generator/pull/15

#### Ticket Link

https://mattermost.atlassian.net/browse/MM-34434

#### Release Note

```release-note
NONE
```


model/client4.go
```diff
@@ -5520,7 +5520,7 @@ func (c *Client4) GetChannelMemberCountsByGroup(channelID string, includeTimezon
 
 // RequestTrialLicense will request a trial license and install it in the server
 func (c *Client4) RequestTrialLicense(users int) (bool, *Response) {
-	b, _ := json.Marshal(map[string]int{"users": users})
+	b, _ := json.Marshal(map[string]interface{}{"users": users, "terms_accepted": true})
```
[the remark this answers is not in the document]
mkraft: Yes this was missing and doesn't work in production. But it doesn't seem to be used yet.

model/license.go
```diff
@@ -51,8 +52,11 @@ type TrialLicenseRequest struct {
 	ReceiveEmailsAccepted bool   `json:"receive_emails_accepted"`
 }
 
-func (tlr *TrialLicenseRequest) ToJson() string {
-	b, _ := json.Marshal(tlr)
+func (lr *TrialLicenseRequest) ToJson() string {
+	b, _ := json.Marshal(struct {
+		*TrialLicenseRequest
+		IsTrial bool `json:"is_trial"`
+	}{lr, true})
```
mkraft, line 59: Decided to hard-code the JSON `"is_trial": true` key/value because there is currently no requirement to request non-trial licenses from Mattermost server.

api4/license_test.go
```diff
@@ -111,3 +111,32 @@ func TestRemoveLicenseFile(t *testing.T) {
...
+		require.False(t, ok)
+	})
+
+	t.Run("trial license user count less than current users", func(t *testing.T) {
+		ok, resp := th.SystemAdminClient.RequestTrialLicense(1)
+		CheckBadRequestStatus(t, resp)
+		require.Equal(t, "api.license.add_license.unique_users.app_error", resp.Error.Id)
+		require.False(t, ok)
+	})
+}
```
mkraft, line 142: These are just tests we didn't have, they don't really test this PR.

model/license_test.go
```diff
@@ -158,6 +159,7 @@ func TestLicenseToFromJson(t *testing.T) {
 			Company: NewId(),
 		},
 		Features: &f,
+		IsTrial:  true,
```
mkraft, line 162: This matches the new field added in https://github.com/mattermost/license-generator/pull/15.

model/license.go
sbishel: Yes. Good call.

sbishel (approved):

cpanato (approved):

model/license.go
marianunez: @mkraft we may not even need this given that in the CWS endpoint for trial I will explicitly set the flag to true to avoid any requests that can override this with a `false` value.

mkraft: Thanks. I didn't realize this was going through CWS, I thought the request was going directly to the license generator lambda via a DNS entry for customers.mattermost.com. I'll remove this `is_trial` JSON key.

mkraft:
@furqanmlk Note that this can be pushed to v5.36 if necessary. The feature discovery banner, etc will still work using license duration.

marianunez (approved):

mkraft:
@furqanmlk As discussed, this story doesn't have anything to test. Neither the APIv4 client nor the `IsTrial` field in the `License` model are used by us. We can safely defer testing these with any future code that uses them.

furqanmlk (approved):

PR #17369, opened 2021-04-09
Fix missing version

metanerd:
#### Release Note

```release-note
```


cpanato (approved):
adding back a version that was removed by mistake

PR #17375, opened 2021-04-09
MM-34712 shared channels: sync user profile image

wiggin77:
#### Summary
This PR adds support for synchronizing user's profile image for shared channels.

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-34712

#### Release Note
```release-note
NONE
```

### DIFF
--- a/services/remotecluster/sendprofileImage.go
+++ b/services/remotecluster/sendprofileImage.go
@@ -1,0 +1,146 @@
+// Copyright (c) 2015-present Mattermost, Inc. All Rights Reserved.
+// See LICENSE.txt for license information.
+
+package remotecluster
+
+import (
+	"bytes"
+	"context"
+	"fmt"
+	"io"
+	"io/ioutil"
+	"mime/multipart"
+	"net/http"
+	"net/url"
+	"path"
+	"time"
+
+	"github.com/mattermost/mattermost-server/v5/model"
+	"github.com/mattermost/mattermost-server/v5/shared/mlog"
+)
+
+type SendProfileImageResultFunc func(userId string, rc *model.RemoteCluster, resp *Response, err error)
+
+type sendProfileImageTask struct {
+	rc       *model.RemoteCluster
+	userID   string
+	provider ProfileImageProvider
+	f        SendProfileImageResultFunc
+}
+
+type ProfileImageProvider interface {
+	GetProfileImage(user *model.User) ([]byte, bool, *model.AppError)
+}
+
+// SendProfileImage asynchronously sends a user's profile image to a remote cluster.
+//
+// `ctx` determines behaviour when the outbound queue is full. A timeout or deadline context will return a
+// BufferFullError if the task cannot be enqueued before the timeout. A background context will block indefinitely.
+//
+// Nil or error return indicates success or failure of task enqueue only.
+//
+// An optional callback can be provided that receives the response from the remote cluster. The `err` provided to the
+// callback is regarding image delivery only. The `resp` contains the decoded bytes returned from the remote.
+// If a callback is provided it should return quickly.
+func (rcs *Service) SendProfileImage(ctx context.Context, userID string, rc *model.RemoteCluster, provider ProfileImageProvider, f SendProfileImageResultFunc) error {
+	task := sendProfileImageTask{
+		rc:       rc,
+		userID:   userID,
+		provider: provider,
+		f:        f,
+	}
+	return rcs.enqueueTask(ctx, rc.RemoteId, task)
+}
+
+// sendProfileImage is called when a sendProfileImageTask is popped from the send channel.
+func (rcs *Service) sendProfileImage(task sendProfileImageTask) {
+	err := rcs.sendProfileImageToRemote(SendTimeout, task)
+	var response Response
+
+	if err != nil {
+		rcs.server.GetLogger().Log(mlog.LvlRemoteClusterServiceError, "Remote Cluster send profile image failed",
+			mlog.String("remote", task.rc.DisplayName),
+			mlog.String("UserId", task.userID),
+			mlog.Err(err),
+		)
+		response.Status = ResponseStatusFail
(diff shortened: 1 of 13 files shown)


Xxoopp20115 (approved):

Xxoopp20115 (commented):
A

Xxoopp20115 (changes requested):
A

services/remotecluster/sendprofileImage.go
```diff
@@ -0,0 +1,157 @@
...
+	u, err := url.Parse(task.rc.SiteURL)
+	if err != nil {
+		return fmt.Errorf("invalid siteURL while sending file to remote %s: %w", task.rc.RemoteId, err)
+	}
+	u.Path = path.Join(u.Path, model.API_URL_SUFFIX, "remotecluster", task.userID, "image")
+
+	body := &bytes.Buffer{}
+	writer := multipart.NewWriter(body)
+
+	part, err := writer.CreateFormFile("image", "profile.png")
```
agnivade, line 118: So this seems like we are reading the entire image, writing it to a byte buffer, and then reading from the buffer again to send it to the remote. Perhaps out of scope for this PR, but I am wondering if we can directly stream the already read image directly to the POST request?

It's another issue though that we don't have streaming read APIs from the app layer.

services/remotecluster/sendprofileImage.go
```diff
@@ -0,0 +1,157 @@
...
+		provider: provider,
+		f:        f,
+	}
+	return rcs.enqueueTask(ctx, rc.RemoteId, task)
+}
+
+// sendProfileImage is called when a sendProfileImageTask is popped from the send channel.
+func (rcs *Service) sendProfileImage(task sendProfileImageTask) {
+	// Ensure a panic from the callback does not exit the goroutine.
+	defer func() {
```
agnivade, line 58: Are we actually panicking from somewhere? If not, then I am wondering why are we doing this only for `sendProfileImage`? It's usually a good practice to let the entire thing crash for a panic.

services/remotecluster/sendprofileImage.go
```diff
@@ -0,0 +1,157 @@
...
+		response.Status = ResponseStatusOK
+	}
+
+	// If callback provided then call it with the results.
+	if task.f != nil {
+		task.f(task.userID, task.rc, &response, err)
+	}
+}
+
+func (rcs *Service) sendProfileImageToRemote(timeout time.Duration, task sendProfileImageTask) error {
```
agnivade, line 93: Would love to have a test for this.

services/remotecluster/sendprofileImage.go
wiggin77: This is done for all three of the sendXXX operations that the remote cluster service supports. It comes from a habit when allowing callbacks from arbitrary places to not trust those callbacks not to crash.  I guess they can be removed.  I could also isolate the panic handler to only guard the callback.  ... ... ...  I think I'll just remove them.

services/remotecluster/sendprofileImage.go
wiggin77: The app layer API to fetch the image returns a []byte of the entire image.  We then need to convert that to a multipart request for the POST.  I could just put the raw binary in the payload and adjust the receiving end to expect that, but that would deviate from how we're doing things with other resource uploads.

If the app layer API provided a reader I could skip the `bytes.NewBuffer` creation.  Profile images are limited to 128x128 so I don't think it is worth it as this stage.  If we fix it we should do it for all resource reading/writing, especially the bigger assets.

services/remotecluster/sendprofileImage.go
wiggin77: Done

services/remotecluster/sendprofileImage.go
wiggin77: Unit test added.

agnivade (approved):

services/remotecluster/sendprofileImage.go
```diff
@@ -0,0 +1,146 @@
...
+
+	if _, err = io.Copy(part, bytes.NewBuffer(img)); err != nil {
+		return err
+	}
+
+	if err = writer.Close(); err != nil {
+		return err
+	}
+
+	req, err := http.NewRequest("POST", u.String(), bytes.NewReader(body.Bytes()))
```
streamer45, line 120:
```suggestion
	req, err := http.NewRequest("POST", u.String(), body)
```

services/remotecluster/sendprofileImage_test.go
```diff
@@ -0,0 +1,187 @@
...
+		}
+	}
+
+	buf := &bytes.Buffer{}
+	png.Encode(buf, img)
+
+	return buf.Bytes(), true, nil
+}
+
+type Flag struct {
```
streamer45, line 172: Does this (and related methods) need to be exported?

services/remotecluster/sendprofileImage_test.go
```diff
@@ -0,0 +1,187 @@
...
+		for y := 0; y < imageHeight; y++ {
+			img.Set(x, y, red)
+		}
+	}
+
+	buf := &bytes.Buffer{}
+	png.Encode(buf, img)
+
+	return buf.Bytes(), true, nil
+}
```
streamer45, line 170: Usually it's just simpler to read some test asset from `tests/` but I like your initiative.

streamer45 (approved):
Looks good :+1:  Only left non-blocking suggestions.

services/remotecluster/sendprofileImage_test.go
wiggin77: No, and fixed.

furqanmlk (approved):
**Name Collision Passed** ✅ 
image

Sync users profile image and Sync custom status passed ✅ 
There are few issues with these feature, will be addressed separately

PR #17388, opened 2021-04-13
MM-34787: Add colored output for non JSON console logs

agnivade:
https://mattermost.atlassian.net/browse/MM-34787

```release-note
A new field EnableColor is added to LogSettings and NotificationLogSettings.
Non-JSON console logs will now be colored if that field is set to true.
```
colors

### DIFF
--- a/model/config.go
+++ b/model/config.go
@@ -1200,6 +1200,7 @@ type LogSettings struct {
 	EnableConsole          *bool   `access:"environment_logging,write_restrictable,cloud_restrictable"`
 	ConsoleLevel           *string `access:"environment_logging,write_restrictable,cloud_restrictable"`
 	ConsoleJson            *bool   `access:"environment_logging,write_restrictable,cloud_restrictable"`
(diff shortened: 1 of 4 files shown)


streamer45 (commented):
Nice addition :+1: I have just a couple of minor concerns with the design:

I am not entirely convinced about passing the whole config just to get `DeveloperMode`.
Do we need the `mlog` package to be aware of the `DeveloperMode`?
I am thinking the decision of using colored output or not can be either left to the caller or even made directly configurable from `model.LogSettings`.

agnivade:
I actually wanted colored output to be by default for all non-JSON output. But there might be some customers running non-JSON in production, in which case they might start to get weird characters in their console. Adding another field to `LogSettings` sounds good to me.

wiggin77 (approved):
LGTM :+1:

streamer45 (approved):
Looks great :+1: 

I am guessing this might need a minor addition to the docs so it probably makes sense to add the label.

PR #17409, opened 2021-04-15
MM-31717: Remove HTTP clustering

agnivade:
https://mattermost.atlassian.net/browse/MM-31717

Webapp PR: https://github.com/mattermost/mattermost-webapp/pull/7924

```release-note
Gossip clustering mode is now GA and is no longer
available as a choice now. All cluster traffic will always
use the gossip protocol.

The config setting UseExperimentalGossip has no effect
and has only been kept for compatibility purposes.

The setting to use gossip has been removed from admin console.

Upgrade note:
If an existing system is not using gossip, then one cannot
upgrade one node in a cluster and keep the other one to the older
version, if an HA upgrade is intended. All nodes
in the cluster must use a single protocol.

Either the customer must use gossip if they want to do an HA
upgrade. Or, shut down all nodes and bring up all of them
after doing an upgrade.
```


grundleborg (approved):

streamer45 (approved):

amyblais:
@agnivade Thanks, v5.35 will be based on Cloud April 21st release and we're already past feature complete. Is this change non-risky?

agnivade:
Oh nvm then, 5.36 it is.

wiersgallak (approved):
Thx!

PR #17414, opened 2021-04-15
[MM-34888] Fix 5.32 migration (cloud branch)

streamer45:
#### Summary

Trying to fix this once and for all. Cloud branch didn't get updated so we ended up doing the migration on MySQL on the newly cut release branch.

#### Ticket

https://mattermost.atlassian.net/browse/MM-34888

#### Release Note

```release-note
NONE
```


ethervoid (approved):
LGTM but the migration applied to old installations is not going to be changed. This will fix it for new installations

streamer45:
> LGTM but the migration applied to old installations is not going to be changed. This will fix it for new installations

That is expected. We are okay with letting PostgreSQL installations "diverge". The important bit is that the MySQL one doesn't run.

PR #17418, opened 2021-04-15
[MM-34919] Use Mattermost-Plugin-ID header to pass ID in inter-plugin requests

hanzei:
#### Summary
Pass the source plugin ID via a `Mattermost-Plugin-ID` header. The exiting mechanism to pass the id -  the `SourcePluginId` in `plugin.Context´ is now deprecated.

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-34919

#### Release Note
```release-note
NONE
```

### DIFF
--- a/app/plugin_requests.go
+++ b/app/plugin_requests.go
@@ -74,6 +74,8 @@ func (a *App) ServeInterPluginRequest(w http.ResponseWriter, r *http.Request, so
 		SourcePluginId: sourcePluginId,
 	}
 
+	r.Header.Set("Mattermost-Plugin-ID", sourcePluginId)
+
 	hooks.ServeHTTP(context, w, r)
 }
 
@@ -134,6 +136,9 @@ func (a *App) servePluginRequest(w http.ResponseWriter, r *http.Request, handler
 	} else {
 		token = r.URL.Query().Get("access_token")
 	}
+
+	// Mattermost-Plugin-ID can only be set by inter-plugin requests
+	r.Header.Del("Mattermost-Plugin-ID")
 
 	r.Header.Del("Mattermost-User-Id")
 	if token != "" {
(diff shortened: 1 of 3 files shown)


levb (approved):

app/plugin_requests.go
```diff
@@ -135,6 +137,9 @@ func (a *App) servePluginRequest(w http.ResponseWriter, r *http.Request, handler
 		token = r.URL.Query().Get("access_token")
 	}
 
+	// Mattermost-Plugin-ID can only be set by inter-plugin requests
+	r.Header.Del("Mattermost-Plugin-ID")
```
srkgupta, line 141: I am not sure if we should follow the CanonicalHeaderKey pattern here, i.e: 
`r.Header.Del("Mattermost-Plugin-Id")`

app/plugin_requests.go
```diff
@@ -74,6 +74,8 @@ func (a *App) ServeInterPluginRequest(w http.ResponseWriter, r *http.Request, so
 		SourcePluginId: sourcePluginId,
 	}
 
+	r.Header.Set("Mattermost-Plugin-ID", sourcePluginId)
```
srkgupta, line 77: I am not sure if we should follow the CanonicalHeaderKey pattern here, i.e:
`r.Header.Set("Mattermost-Plugin-Id", sourcePluginId)`

srkgupta (commented):
Thanks @hanzei. Just a couple of minor observations. But overall it looks good to me.

crspeller (approved):

app/plugin_requests.go
hanzei: `Mattermost-Plugin-ID` is the header name that the GitHub and autolink plugins use. In order for the fix to also apply to the GitHub we need to stick with the same name.

srkgupta (approved):

hanzei:
@srkgupta Would you be open on testing this PR as it requires some code setup?

srkgupta:
QA'd the PR and the changes are working fine. LGTM. Thanks @hanzei

PR #17422, opened 2021-04-16
#MM-34893 Completely disabled Admin Advisor

harshilsharma63:
#### Summary
Completely disabled Admin Advisor

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-34893

#### Release Note

None

#### Additional Pull Requests

1. **Docs** - https://github.com/mattermost/docs/pull/4503

### DIFF
--- a/app/app.go
+++ b/app/app.go
@@ -191,6 +191,7 @@ func (s *Server) getFirstServerRunTimestamp() (int64, *model.AppError) {
 	return value, nil
 }
 
+//nolint:golint,unused,deadcode
 func (s *Server) getLastWarnMetricTimestamp() (int64, *model.AppError) {
 	systemData, err := s.Store.System().GetByName(model.SYSTEM_WARN_METRIC_LAST_RUN_TIMESTAMP_KEY)
 	if err != nil {
@@ -335,6 +336,7 @@ func (a *App) getWarnMetricStatusAndDisplayTextsForId(warnMetricId string, T i18
 	return nil, nil
 }
 
+//nolint:golint,unused,deadcode
 func (a *App) notifyAdminsOfWarnMetricStatus(warnMetricId string, isE0Edition bool) *model.AppError {
 	perPage := 25
 	userOptions := &model.UserGetOptions{
--- a/app/app.go
+++ b/app/app.go
@@ -94,7 +94,6 @@ func (a *App) InitServer() {
 		if a.Srv().runEssentialJobs {
 			a.Srv().Go(func() {
 				runLicenseExpirationCheckJob(a)
-				runCheckWarnMetricStatusJob(a)
 			})
 			a.srv.runJobs()
 		}
--- a/app/bot.go
+++ b/app/bot.go
@@ -88,6 +88,7 @@ func (a *App) CreateBot(bot *model.Bot) (*model.Bot, *model.AppError) {
 	return savedBot, nil
 }
 
+//nolint:golint,unused,deadcode
 func (a *App) getOrCreateWarnMetricsBot(botDef *model.Bot) (*model.Bot, *model.AppError) {
 	botUser, appErr := a.GetUserByUsername(botDef.Username)
 	if appErr != nil {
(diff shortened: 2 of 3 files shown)


sbishel (approved):
@jasonblais What are the expectations of reviving this code? I would prefer to remove the code. I'm afraid of creating a bad pattern. The code can always be recovered from source control.

jasonblais:
@sbishel I'm open to either option, but it is in the near term roadmap. We had originally planned to work on it earlier this year, but lacked the PM/engineer resource for it: https://docs.google.com/document/d/1tzcwEhw2ZyDWjzfBKheo1Wyk8f4wQuvWuHOCka98el0/edit#bookmark=id.nvelcxi2suby

furqanmlk (approved):
Working as expected

app/app.go
```diff
@@ -94,7 +94,6 @@ func (a *App) InitServer() {
 		if a.Srv().runEssentialJobs {
 			a.Srv().Go(func() {
 				runLicenseExpirationCheckJob(a)
-				runCheckWarnMetricStatusJob(a)
```
catalintomai, line 97: Does any documentation need to be updated?

catalintomai:
@sbishel, @jasonblais , is the current code being used by other flows. Some time ago @mgdelacroix was inquiring about using the AdminAdvisor setup for other feature needs.

jasonblais:
@catalintomai not to my knowledge, but maybe @mgdelacroix can confirm?

app/app.go
harshilsharma63: @catalintomai nice catch! I see a few sections dedicated to Admin Advisor in the docs, most notably this one - https://docs.mattermost.com/administration/notices.html?highlight=admin%20advisor#admin-advisor-notices, though there are a few more - https://docs.mattermost.com/search.html?q=admin+advisor&check_keywords=yes&area=default

@jasonblais should we completely remove those from the docs?

jasonblais: Yea, we should update the docs.
- https://docs.mattermost.com/overview/compliance.html#what-information-is-shared-when-i-click-contact-us-on-a-mattermost-admin-advisor-notification can be updated
- https://docs.mattermost.com/administration/notices.html#admin-advisor-notices I would keep and mention they have been paused in v5.35

harshilsharma63: @jasonblais done.

harshilsharma63:
@catalintomai @jasonblais PR for updating docs - https://github.com/mattermost/docs/pull/4503

amyblais:
@mkraft v5.35 will be based on Cloud April 21st release which is coming on Wednesday, please help prioritize review.

mgdelacroix:
@catalintomai thanks for the ping! Yeah, `mmctl` will rely on the `WarnMetricsBot` to notify of certain actions in the channels where they happen, but it only uses it on demand rather than periodically, so this PR doesn't interfere with it. Will update mine accordingly to honor this change if needed

mgdelacroix (approved):

mkraft (approved):

agnivade:
@sbishel / @jasonblais - As Scott predicted, keeping unused code is dangerous and the feature was mistakenly enabled again in an unrelated [PR](https://github.com/mattermost/mattermost-server/pull/16067) while merging with latest master.

We will be removing the code entirely.

jasonblais:
@agnivade As a heads up, the reason it wasn't completely removed was because admin advisors are used for other use cases.

Since then, we have migrated the CLI/mmctl admin messages to use a different bot (see https://github.com/mattermost/mattermost-server/pull/17598).

There is at least one other use case that it's used for - see https://mattermost.atlassian.net/browse/MM-18818. That said, I believe this use case wouldn't work with the admin advisor job disabled so it wouldn't make a difference to disable vs remove it. I would check with Katie + Martin about this use case though.

agnivade:
@jasonblais - I think you perhaps misunderstood my point. This PR disabled some code. But kept that in the codebase, which later introduced a bug by mistakenly enabling the code. MM-18818 uses different code.

We are removing the code here in this PR https://github.com/mattermost/mattermost-server/pull/18054.

jasonblais:
@agnivade sounds good, thank you!

PR #17424, opened 2021-04-16
MM-34758 Collapsed Reply Threads without mobile support

reflog:
#### Summary

added support for legacy clients accessing server with threaded replies via `collapsed_threads_supported` param to viewChannel API and setPostUnread API

#### Ticket Link

https://mattermost.atlassian.net/browse/MM-34758

#### Related Webapp PR
https://github.com/mattermost/mattermost-webapp/pull/7933

#### Release Note

```release-note
added optional `collapsed_threads_supported` parameter to /channels/members/{userId}/view to indicate that the client supports collapsed threads.

added optional `collapsed_threads_supported` parameter to /users/{userId}/posts/{postId}/set_unread to indicate that the client supports collapsed threads.
```


zefhemel (approved):

ashishbhate (approved):

app/channel.go
```diff
@@ -2619,10 +2619,25 @@ func (a *App) MarkChannelsAsViewed(channelIDs []string, userID string, currentSe
...
+		// check if a participant has overridden collapsed threads settings
+		if preference, err := a.Srv().Store.Preference().Get(userID, model.PREFERENCE_CATEGORY_DISPLAY_SETTINGS, model.PREFERENCE_NAME_COLLAPSED_THREADS_ENABLED); err == nil {
+			threadsEnabled = preference.Value == "on"
+		}
+		if threadsEnabled {
+			if err := a.Srv().Store.Thread().MarkAllAsReadInChannels(userID, channelIDs); err != nil {
+				return nil, model.NewAppError("MarkChannelsAsViewed", "app.channel.update_last_viewed_at.app_error", nil, err.Error(), http.StatusInternalServerError)
+			}
+		}
+	}
```
calebroseland, line 2635: If we update the mark-as-unread logic as in the other PR, is this needed? (With CRT:OFF behaving the same as CRT:unsupported?)

reflog: @calebroseland I don't think so. We want this behavior when old mobile clients are viewing the channel. So we can't rely on webapp changes

ogi-m (approved):

app/channel.go
```diff
@@ -2378,17 +2378,100 @@ func (a *App) MarkChannelAsUnreadFromPost(postID string, userID string) (*model.
...
+		// In CRT Supported Client: badge on channel only sums mentions in root posts including and below the post that was marked.
+		// In CRT Unsupported Client: badge on channel sums mentions in all posts (root & replies) including and below the post that was marked unread.
 		if post.RootId == "" {
-			threadId = post.Id
+			channelUnread, aErr := a.Srv().Store.Channel().UpdateLastViewedAtPost(post, userID, unreadMentions, unreadMentionsRoot, false, true)
+			if aErr != nil {
+				return channelUnread, model.NewAppError("MarkChannelAsUnreadFromPost", "app.channel.update_last_viewed_at_post.app_error", nil, aErr.Error(), http.StatusInternalServerError)
+			}
+
+			message := model.NewWebSocketEvent(model.WEBSOCKET_EVENT_POST_UNREAD, channelUnread.TeamId, channelUnread.ChannelId, channelUnread.UserId, nil)
```
zefhemel, line 2401: You know the "way we do things in server" much better than me, but would it make sense to extract the construction/sending of WS messages into a separate method make sense? These functions you're expanding are already getting very, very long.

app/channel.go
```diff
@@ -2378,17 +2378,100 @@ func (a *App) MarkChannelAsUnreadFromPost(postID string, userID string) (*model.
...
+		threadMembership.LastViewed = post.UpdateAt - 1
+		threadMembership.UnreadMentions, err = a.countThreadMentions(user, rootPost, channel.TeamId, post.UpdateAt-1)
+		if err != nil {
+			return nil, err
 		}
+		_, nErr = a.Srv().Store.Thread().UpdateMembership(threadMembership)
+		if nErr != nil {
+			return nil, model.NewAppError("MarkChannelAsUnreadFromPost", "app.channel.update_last_viewed_at_post.app_error", nil, nErr.Error(), http.StatusInternalServerError)
+		}
+		thread, _ := a.Srv().Store.Thread().GetThreadForUser(userID, channel.TeamId, threadId, true)
```
zefhemel, line 2451: Ignoring returned error is ok?

app/channel.go
```diff
@@ -2622,10 +2704,32 @@ func (a *App) MarkChannelsAsViewed(channelIDs []string, userID string, currentSe
 	for _, channelID := range channelsToClearPushNotifications {
 		a.clearPushNotification(currentSessionId, userID, channelID)
 	}
+
+	if !collapsedThreadsSupported {
+		// for compatibility with old clients, when channel is viewed - mark all threads in that channel as read
```
zefhemel, line 2709: I see this "checking if threads are enabled" happen a few time, every time spelled out. Worth extracting into a function? Also seems to happen at line 2456 and perhaps more places>

zefhemel (commented):
First of all thanks for adding a bit more comments on why behaviors are different based on support and enabling of CRT, future readers will thank you. I added a few comments, but not very major. My general sense is that some methods here are getting longer and longer, often due to elaborate specific "if CRT" paths, and I'm wondering if we can factor some of that out more.

app/channel.go
ashishbhate: Do you mean move all the code for `!collapsedThreadsSupported` into a separate functions?

zefhemel: I mean the code to check "is collapsed threads switched on for this user?"

app/channel.go
```diff
@@ -2622,10 +2704,32 @@ func (a *App) MarkChannelsAsViewed(channelIDs []string, userID string, currentSe
 	for _, channelID := range channelsToClearPushNotifications {
 		a.clearPushNotification(currentSessionId, userID, channelID)
 	}
+
+	if !collapsedThreadsSupported {
```
calebroseland, line 2708: For lines 14,15 in the unreads UX spreadsheet.

> When opening a channel from a client that doesn't support CRT or user has CRT turned off, both the channel and the threads in that channel are marked as read on desktop

May be out of scope or done in a follow up PR, but is this block where we should handle that?

ashishbhate: For the users that have CRT turned off?
Supporting CRT turned off will make some of the conditionals here a little more complicated. This PR is already too big, so I'd like to make those changes in a separate PR. 
We have a ticket for the issue:  https://mattermost.atlassian.net/browse/MM-35699

calebroseland: > For the users that have CRT turned off?

Yes, and I agree should be done separately. More-or-less just a question about that. Sounds good on https://mattermost.atlassian.net/browse/MM-35699 👍

jwilander (approved):
LGTM - no immediate concerns

jwilander:
@ashishbhate, not exactly related to this PR, but has the CRT code been load tested?

ashishbhate:
>@ashishbhate, not exactly related to this PR, but has the CRT code been load tested?

I remember there was some discussion about CRT's affect on performance. I found this message in the `developers:performance` channel https://community-daily.mattermost.com/core/pl/nd6d3mq77iga7p3pjprm9c483e (see earlier messages in the channel for full context). It seems CRT has an affect on performance but its not "dramatic". 

Its been a long time since, so I've created a ticket https://mattermost.atlassian.net/browse/MM-36045 to look into updating the loadtest code so that the CRT features are explicitly covered, and to regularly run the tests so that there aren't any unexpected performance regressions.

PR #17431, opened 2021-04-17
MM-34983: Remove check for number of data sources

agnivade:
In case of Aurora databases, a single endpoint
might load balance between multiple replicas.
In that case, this check is incorrect. Therefore,
we remove it

https://mattermost.atlassian.net/browse/MM-34983


agnivade:
@amyblais - This would be good to cherry-pick to 5.35.

streamer45 (approved):

amyblais:
@nronas v5.35 will be based on Cloud April 21st release which is coming on Wednesday, please help prioritize review.

nronas (approved):
LGTM

agnivade:
Oh the branch isn't cut yet.

PR #17444, opened 2021-04-19
[MM-13799] Avoid queries timing out during upgrade

streamer45:
#### Summary

PR updates gorp dependency to fix an issue with `ExecNoTimeout()`.

I've linked this to a rather old ticket because I am thinking it might have been created in response to this bug appearing. 
This is not a general solution, meaning we still need to explicitly call `ExecNoTimeout()` (or one of its wrapping methods) in order to avoid timing out on queries taking longer than `SqlSettings.QueryTimeout` but I felt this was good enough to address the issues at hand also given the migration system will soon change.

#### Related PRs

https://github.com/mattermost/gorp/pull/11

#### Release Note

```release-note
Fixes an issue where queries executed during the upgrade process would preemptively timeout on the application side.
```

#### Ticket

https://mattermost.atlassian.net/browse/MM-13799

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -71,7 +71,7 @@ require (
 	github.com/lib/pq v1.10.0
 	github.com/mailru/easyjson v0.7.7 // indirect
 	github.com/mattermost/go-i18n v1.11.0
-	github.com/mattermost/gorp v1.6.2-0.20210419095934-8fb4e3549b30
+	github.com/mattermost/gorp v1.6.2-0.20210419141818-0904a6a388d3
 	github.com/mattermost/gosaml2 v0.3.3
 	github.com/mattermost/ldap v0.0.0-20201202150706-ee0e6284187d
 	github.com/mattermost/logr v1.0.13
--- a/go.mod
+++ b/go.mod
@@ -71,7 +71,7 @@ require (
 	github.com/lib/pq v1.10.0
 	github.com/mailru/easyjson v0.7.7 // indirect
 	github.com/mattermost/go-i18n v1.11.0
-	github.com/mattermost/gorp v1.6.2-0.20200624165429-2595d5e54111
+	github.com/mattermost/gorp v1.6.2-0.20210419095934-8fb4e3549b30
 	github.com/mattermost/gosaml2 v0.3.3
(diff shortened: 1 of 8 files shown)


go.mod
```diff
@@ -71,7 +71,7 @@ require (
 	github.com/lib/pq v1.10.0
 	github.com/mailru/easyjson v0.7.7 // indirect
 	github.com/mattermost/go-i18n v1.11.0
-	github.com/mattermost/gorp v1.6.2-0.20200624165429-2595d5e54111
+	github.com/mattermost/gorp v1.6.2-0.20210419095934-8fb4e3549b30
```
streamer45, line 74: Will update this once the gorp PR is merged.

store/sqlstore/store_test.go
```diff
@@ -617,3 +617,22 @@ func makeSqlSettings(driver string) *model.SqlSettings {
...
+
+func TestExecNoTimeout(t *testing.T) {
+	StoreTest(t, func(t *testing.T, ss store.Store) {
+		sqlStore := ss.(*SqlStore)
+		var query string
+		timeout := sqlStore.master.QueryTimeout
+		sqlStore.master.QueryTimeout = 1
+		defer func() {
+			sqlStore.master.QueryTimeout = timeout
+		}()
```
streamer45, line 629: I was a bit worried this could be causing a race condition but didn't spot any and given it was just a test I've committed as is. An alternative would be to setup separate stores just to run this. Let me know if you'd prefer that option.

agnivade: Separate stores would be good, but I'm 0/5. It shouldn't be a big deal.

agnivade (approved):

streamer45:
@amyblais 

This should be included in 5.35 to help avoid issues with the upcoming long migration.

amyblais:
@jespino FYI for PR review as 5.35 will be based on cloud April 21st release.

jespino (approved):
LGTM

store/sqlstore/store_test.go
streamer45: Alright, let's see how it goes. It should be trivial to fix in case of failure.

PR #17447, opened 2021-04-19
MM-34871 CRT: Participants of thread include non-replying followers (and past followers)

reflog:
#### Summary

don't include user in participants when just following

#### Ticket Link

https://mattermost.atlassian.net/browse/MM-34871

#### Release Note

```release-note
NONE
```


calebroseland (commented):
> auto-follow after being mentioned also makes you a participant

I'm still seeing this; did you want to split that out into a separate ticket?

reflog:
@calebroseland - Is "auto-follow after being mentioned should not make you a participant" actually in the spec?
cc @matthewbirtch @esethna

zefhemel:
> @calebroseland - Is "auto-follow after being mentioned should not make you a participant" actually in the spec?
> cc @matthewbirtch @esethna

All I can find regarding what constitutes being a participant is "Displays avatars of all unique users who contributed to the thread, including bots" I wouldn't consider being mentioned a contribution. So in my mind it should auto-follow, but not put you in the list of participants (unless you actually reply).

calebroseland:
> > @calebroseland - Is "auto-follow after being mentioned should not make you a participant" actually in the spec?
> > cc @matthewbirtch @esethna
> 
> All I can find regarding what constitutes being a participant is "Displays avatars of all unique users who contributed to the thread, including bots" I wouldn't consider being mentioned a contribution. So in my mind it should auto-follow, but not put you in the list of participants (unless you actually reply).

This was my understanding/thought as well

reflog:
@ogi-m @zefhemel - let's assume that it's the correct interpretation of the spec :)
I've updated the PR and test, now when user is mentioned, they autofollow the thread but are NOT added to the participants list.

matthewbirtch:
> @calebroseland - Is "auto-follow after being mentioned should not make you a participant" actually in the spec?
> cc @matthewbirtch @esethna

I don't think what constitutes being a participant is in the spec. As Zef mentioned, the only thing we've stated there is "Displays avatars of all unique users who contributed to the thread, including bots". So I would agree with the others that that the act of replying or creating the original root post is what makes you a participant. Simply following or being mentioned does not make you a participant

reflog:
thanks for the clarification @matthewbirtch

ashishbhate (approved):

calebroseland (approved):

ogi-m (approved):
Thanks Eli, LGTM 👍🏻

PR #17470, opened 2021-04-20
Fix performance problem on document extraction

jespino:
#### Summary
Fix performance problem on document extraction

#### Release Note
```release-note
NONE
```


services/docextractor/plain.go
```diff
@@ -36,5 +41,6 @@ func (pe *plainExtractor) Extract(filename string, r io.ReadSeeker) (string, err
 		}
 	}
 
+	text, _ := ioutil.ReadAll(r)
```
gabrieljackson, line 44: Don't see a reason to check the error?

gabrieljackson (approved):
👍

mgdelacroix (approved):

PR #17484, opened 2021-04-22
Updates on CI

cpanato:
#### Summary
 - Moving out the lint/vet part from the build step to make more clear when it fails. 
 - some updates for some job in the resource_class, because we pay extra for bigger machines and in some steps we don't need that

#### Ticket Link
N/A

#### Release Note

```release-note
NONE
```

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -325,9 +325,8 @@ ifeq ($(BUILD_ENTERPRISE_READY),true)
 	./scripts/prereq-check-enterprise.sh
 endif
 
-check-style: golangci-lint plugin-checker vet ## Runs golangci against all packages
+check-style: golangci-lint plugin-checker vet ## Runs style/lint checks
 
-check-style-ci: plugin-checker vet ## Runs plugin-checker and vet
 
 do-cover-file: ## Creates the test coverage report file.
 	@echo "mode: count" > cover.out
--- a/Makefile
+++ b/Makefile
@@ -327,6 +327,8 @@ endif
 
 check-style: golangci-lint plugin-checker vet ## Runs golangci against all packages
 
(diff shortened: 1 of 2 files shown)


cpanato:
found some lint errors using golangci-lint latest version, will fix on this PR

agnivade:
Thanks Carlos. Moving out linting to a separate step sounds fine to me.

I am wondering what is the advantage we are getting by trying to use multiple CI systems? Won't that create extra overhead for contributors/staff debugging CI failures?

Secondly, there's not need to update golangci version and spend effort on changing the codebase as part of this PR. We can pin to `v1.33.2` for golang-ci. There's already a contributor working on upgrading to the latest version.

cpanato:
@agnivade I was thinking more in have a small separation and having anything that are very small running on GitHub actions, but I agree with you that we add another ci system. but looks like we already are using github actions for another thing.

but I will refactor and make it an isolated step in circleci.

thanks for the tip for the golangci version and again thanks for your review/feedback

Makefile
```diff
@@ -327,6 +327,8 @@ endif
 
 check-style: golangci-lint plugin-checker vet ## Runs golangci against all packages
 
+check-style-ci: plugin-checker vet ## Runs plugin-checker and vet
```
agnivade, line 330: It's not very clear why there's a `check-style` and `check-style-ci`. I'd probably run `make plugin-checker` and `make vet` serially. 0/5.

agnivade (approved):
Looks great. Just one non-blocking comment.

Makefile
cpanato: sounds reasonable

jespino (approved):
LGTM

PR #17495, opened 2021-04-23
[MM-35077] Add basic support for plugin intra-cluster communication

streamer45:
#### Summary

PR adds initial support for intra-cluster plugin communication.

- Added a new `PublishPluginClusterEvent` API method for the sending side.
- Added a new `OnPluginClusterEvent` hook for the receiving side.

#### Sample usage

```go
package main

import (
	"fmt"
	"net/http"

)

func (p *Plugin) OnPluginClusterEvent(c *plugin.Context, ev model.PluginClusterEvent) {
	p.API.LogInfo("got event!")
	p.API.LogInfo(fmt.Sprintf("%s", ev.Id))
	p.API.LogInfo(fmt.Sprintf("%s", string(ev.Data)))
}

func (p *Plugin) ServeHTTP(c *plugin.Context, w http.ResponseWriter, r *http.Request) {
	if r.URL.Path == "/broadcast" {
		p.API.LogInfo("Broadcasting event to other plugin instances")

		p.API.PublishPluginClusterEvent(model.PluginClusterEvent{
			Id:   "my_custom_event",
			Data: []byte("some data I want to share"),
		}, model.PluginClusterEventSendOptions{
			SendType: model.PluginClusterEventSendTypeReliable,
		})
	}
}
```

#### Ticket Link

https://mattermost.atlassian.net/browse/MM-35077

#### Related PR

https://github.com/mattermost/enterprise/pull/985

#### Release Note

```release-note
Added basic intra-cluster communication support for plugins.
```


app/plugin_api.go
```diff
@@ -1063,3 +1063,22 @@ func (api *PluginAPI) DeleteCommand(commandID string) error {
 
 	return nil
 }
+
+// PublishPluginEvent broadcasts a plugin event to all other running instances of
+// the calling plugin.
+func (api *PluginAPI) PublishPluginEvent(ev model.PluginEvent, opts model.PluginEventSendOptions) error {
+	if api.app.Cluster() == nil {
+		return nil
+	}
```
streamer45, line 1072: We may decide to return an error in this case. Keeping it a no-op for the moment.

model/plugin_event.go
```diff
@@ -0,0 +1,24 @@
...
+
+package model
+
+const (
+	PluginEventSendTypeReliable   = CLUSTER_SEND_RELIABLE
+	PluginEventSendTypeBestEffort = CLUSTER_SEND_BEST_EFFORT
+)
+
+// PluginEvent is used to allow intra-cluster plugin communication.
+type PluginEvent struct {
```
streamer45, line 12: I decided to go with a dedicated structure mainly to try and abstract a bit from the details of our cluster implementation.

streamer45 (commented):
@jespino 

This is ready for initial feedback. I've just implemented the most basic mechanism for plugins to communicate (a broadcast).

As you can see I haven't added a `Subscribe` method mainly because the more I thought about it and the more it felt redundant as there isn't currently a straightforward way to know in advance (before sending a cluster event) if the receiving plugin subscribed to something. Of course we could do the filtering on the receiving side before calling the hook but at that point I don't see any real advantage. Happy to hear your thoughts on this.

As for next steps, I'd like to add the possibility to publish a message to a specific instance. I am thinking of adding a `ReceiverID` to the send options that could match the cluster id of the node the message is targeting.

plugin/hooks.go
```diff
@@ -35,6 +35,7 @@ const (
 	UserHasBeenCreatedID     = 17
 	ReactionHasBeenAddedID   = 18
 	ReactionHasBeenRemovedID = 19
+	OnPluginEventID          = 20
```
jespino, line 38: I think this is misleading, I think should be `OnPluginClusterEventID`

plugin/hooks.go
```diff
@@ -209,4 +210,13 @@ type Hooks interface {
...
 	ReactionHasBeenRemoved(c *Context, reaction *model.Reaction)
+
+	// OnPluginEvent is invoked when an intra-cluster plugin event is received.
+	//
+	// This is used to allow communication between multiple instances of the same plugin
+	// that are running on separate nodes of the same High-Availability cluster.
+	// This hook receives events sent by a call to PublishEventPlugin.
+	//
+	// Minimum server version: 5.36
+	OnPluginEvent(c *Context, ev model.PluginEvent)
```
jespino, line 221: Same here I would use `OnPluginClusterEvent`

plugin/api.go
```diff
@@ -1050,6 +1050,16 @@ type API interface {
...
+
+	// PublishPluginEvent broadcasts a plugin event to all other running instances of
+	// the calling plugin that are present in the cluster.
+	//
+	// This method is used to allow plugin communication in a High-Availability cluster.
+	// The receiving side should implement the OnPluginEvent hook
+	// to receive events sent through this method.
+	//
+	// Minimum server version: 5.36
+	PublishPluginEvent(ev model.PluginEvent, opts model.PluginEventSendOptions) error
```
jespino, line 1062: `PublishPluginClusterEvent`

model/plugin_event.go
```diff
@@ -0,0 +1,24 @@
...
+	// Id is the unique identifier for the event.
+	Id string
+	// Data is the event payload.
+	Data []byte
```
jespino, line 16: I have "confrontend feelings" here, I don't like all this back and forth with the `[]byte`/`string` difference between `PluginEvent` and `ClusterEvent`, but at the same time I think `[]byte` is the right type, but we don't want to break the backward compatibility. I just want to express my concern, but feel free to keep it or change it.

model/plugin_event.go
jespino, line 12: Prabably renaming it to `PluginClusterEvent`

plugin/hooks.go
streamer45: As usual naming was one of the hardest part of the design and I was trying to keep it short but I agree that this way it can read ambiguous/unclear. I'll update.

model/plugin_event.go
streamer45: I share the same feeling although I think we should move in the direction of making the the `ClusterEvent` field a `[]byte` instead. And even though it's in `model` it might be worth considering the modification. It's an enterprise only object which I am not sure has any outside use so I am thinking it could be changed without major side effects.

streamer45: This part will be handled in https://mattermost.atlassian.net/browse/MM-25116.

streamer45:
Added the ability to send plugin cluster events to specific nodes. This was the missing functionality I required and it should cover plenty of cases. At some point we might want to directly expose cluster ids to facilitate the communication but each node can already access its own ID through `GetPluginStatus` (and possibly broadcast it to others) so I am thinking it should be enough for now.

jespino (approved):
LGTM

model/plugin_cluster_event.go
```diff
@@ -0,0 +1,28 @@
...
+	// Data is the event payload.
+	Data []byte
+}
+
+// PluginClusterEventSendOptions defines some properties that apply when sending
+// plugin events across a cluster.
+type PluginClusterEventSendOptions struct {
+	// SendType defines the type of communication channel used to send the event.
+	SendType string
+	// ReceiverId identifies which node the message should be sent to.
```
agnivade, line 24: I feel sender/receiver makes it slightly confusing because "receiver" can mean from both directions. source/destination makes it slightly more clear. 0/5.

app/plugin_api.go
```diff
@@ -1063,3 +1063,34 @@ func (api *PluginAPI) DeleteCommand(commandID string) error {
...
+			"EventID":  ev.Id,
+		},
+		Data: string(ev.Data),
+	}
+
+	// If ReceiverId is not set we broadcast to all other cluster nodes.
+	if opts.ReceiverId == "" {
+		api.app.Cluster().SendClusterMessage(msg)
+	} else {
+		if err := api.app.Cluster().SendClusterMessageToNode(opts.ReceiverId, msg); err != nil {
```
agnivade, line 1090: So how does the plugin know of all the node ids in the cluster?

app/plugin_api.go
```diff
@@ -1063,3 +1063,34 @@ func (api *PluginAPI) DeleteCommand(commandID string) error {
...
+
+	msg := &model.ClusterMessage{
+		Event:            model.CLUSTER_EVENT_PLUGIN_EVENT,
+		SendType:         opts.SendType,
+		WaitForAllToSend: false,
+		Props: map[string]string{
+			"PluginID": api.id,
+			"EventID":  ev.Id,
+		},
+		Data: string(ev.Data),
```
agnivade, line 1083: yay more strings..

agnivade (approved):
Some minor comments. But nothing blocking.

app/plugin_api.go
streamer45: Yeah, although that's due to the original cluster message which, as agreed, I am going to fix soon (see https://github.com/mattermost/mattermost-server/pull/17495#discussion_r619096670).

app/plugin_api.go
streamer45: I've briefly explained my reasoning in https://github.com/mattermost/mattermost-server/pull/17495#issuecomment-826705963. Each node is able to figure out its own cluster id so it can be either broadcast to others or used as part of some key in KVStore. I think that's enough for what we are planning now and we always have the option to export `GetClusterInfos()` (or some variant) to the plugin API to facilitate this even further.

model/plugin_cluster_event.go
streamer45: I see, but this is an object defining sending options so I assumed the receiver could only be the one we are sending the event to. No issue with changing the naming here but I am not sure how this differs from source/destination. They can be both sources/destinations same way they can be both senders/receivers but it's the context that removes the ambiguity.

Maybe we could just name it `TargetId` to get us out of these semantics.

agnivade: Yeah, destination could very well work both ways. It's just how I perceived the meaning. No strong reasons.

streamer45:
@saturninoabril

I had to touch a couple of clustering internals to implement this (see related enterprise PR). I'd like QA to do an high-level smoke test around functionalities in a HA cluster, essentially making sure nothing major broke.

streamer45:
@lieut-data @levb 

Pinging you guys for visibility as I understand you might have interest in this kind of functionality.  Comments or suggestions are most welcome.

levb:
❤️

app/cluster_handlers.go
```diff
@@ -34,6 +35,35 @@ func (a *App) clusterInstallPluginHandler(msg *model.ClusterMessage) {
...
+	}
+	if msg.Props == nil {
+		mlog.Warn("ClusterMessage.Props for plugin event should not be nil")
+		return
+	}
+	pluginID := msg.Props["PluginID"]
+	eventID := msg.Props["EventID"]
+	if pluginID == "" || eventID == "" {
+		mlog.Warn("Invalid ClusterMessage.Props values for plugin event",
+			mlog.String("PluginID", pluginID), mlog.String("EventID", eventID))
```
lieut-data, line 51: nit: for logging purposes, I've found it helpful that existing logs use `plugin_id` and don't require me to search for permutations. Worth logging as snake case as such?

lieut-data (commented):
Really cool stuff!

app/cluster_handlers.go
streamer45: Sounds good to me. I am all for consistency :)

saturninoabril (approved):
Tested and passed. Did some tests around HA cluster (stopping/restarting other machine), plugin uploads, random functionality checks and all looks good.
Screen Shot 2021-04-29 at 1 03 49 AM

PR #17505, opened 2021-04-25
MM-35128 CRT: Global Threads item should not be affected by 'mark as unread' in channel

reflog:
#### Summary

don't trigger mentions for an existing thread membership when unread is done from within channel
#### Ticket Link

https://mattermost.atlassian.net/browse/MM-35128

#### Release Note

```release-note
NONE
```

### DIFF
--- a/app/channel.go
+++ b/app/channel.go
@@ -2389,38 +2389,40 @@ func (a *App) MarkChannelAsUnreadFromPost(postID string, userID string) (*model.
 		}
 
 		threadMembership, _ := a.Srv().Store.Thread().GetMembershipForUser(user.Id, threadId)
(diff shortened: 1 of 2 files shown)


ashishbhate (approved):

ogi-m:
@reflog I was able to repro:

1. Create a root post
2. Reply with a mention for user2
3. As user2 open the thread and make sure it's marked as viewed
4. As user2 mark the root post in the channel as unread

Observed: Thread is marked as unread and Threads area is bolded
Expected: Threads area should not be affected

ogi-m (approved):
Working as expected now 👍🏻

PR #17511, opened 2021-04-26
MM-34419: Fix plugin enable race

agnivade:
A race happens when we try to enable plugins
in an HA environment. There is no loss in functionality
here, but it's a timing bug.

There are two bugs here.

The first one is a case of nested config listeners. a.InitPlugins
is called in InitServer, and as well as a config listener. And that
calls a.SyncPluginsActiveState(). But inside a.InitPlugins,
there is yet another config listener which again calls
a.SyncPluginsActiveState(). The first fix is to simply not call
the method again from the nested listener.

The second bug happens because the config changed message is sent
across the cluster only after saving the config locally
in the store. To fix this, we simply change GetPluginStatus
to not fail and return empty status when other nodes don't have plugins
enabled.

https://mattermost.atlassian.net/browse/MM-34419

```release-note
Fix a race condition where enabling plugins would result
in spurious errors in the logs.
```

### DIFF
--- a/app/plugin.go
+++ b/app/plugin.go
@@ -210,9 +210,13 @@ func (a *App) InitPlugins(pluginDir, webappPluginDir string) {
 	// Sync plugin active state when config changes. Also notify plugins.
 	a.Srv().PluginsLock.Lock()
 	a.RemoveConfigListener(a.Srv().PluginConfigListenerId)
-	a.Srv().PluginConfigListenerId = a.AddConfigListener(func(*model.Config, *model.Config) {
-		a.installFeatureFlagPlugins()
-		a.SyncPluginsActiveState()
+	a.Srv().PluginConfigListenerId = a.AddConfigListener(func(old, new *model.Config) {
+		// If plugin status remains unchanged, only then run this.
+		// Because (*App).InitPlugins is already run as a config change hook.
+		if *old.PluginSettings.Enable == *new.PluginSettings.Enable {
+			a.installFeatureFlagPlugins()
+			a.SyncPluginsActiveState()
+		}
 		if pluginsEnvironment := a.GetPluginsEnvironment(); pluginsEnvironment != nil {
 			pluginsEnvironment.RunMultiPluginHook(func(hooks plugin.Hooks) bool {
 				if err := hooks.OnConfigurationChange(); err != nil {


agnivade:
Continuing the rest of the discussion here in a comment.

So, `a.SyncPluginsActiveState()` eventually calls
`a.GetPluginStatus` which tries to fetch plugin status
from the entire cluster. But since the cluster message
hasn't been broadcasted yet, this fails.

There might be several ways to fix this, none of which are very
pretty.
1. Delay the invoking of the config listeners by passing a
flag in config.Set and calling config.Broadcast separately.
This makes the API brittle and adds more complexity.
2. Broadcast the cluster message from inside config.Set itself.
This increases coupling and makes the config package aware
of the cluster package.
3. Introduce the concept of hooks in the config package. And
pass the cluster broadcast function as a callback to the
"pre_broadcast" hook. This solution isn't too bad, but it feels
like shoe-horning a solution just to solve a one-off issue.
4. Perhaps modify a.SyncPluginsActiveState itself to make it
work irrespective of whether all clusters are in the same state
or not.

I should clarify here that the whole objective is to just fix
a logging issue. There is no data race or any functional bug
here. So there is no need to come up with a design that takes
a week of effort. But nevertheless, I believe that the design
is incorrect and something that we should fix.

I opened a draft PR to solicit feedback. I haven't committed to any design
yet for this second problem. Maybe the config overhaul of @streamer45 
somehow will help in solving this better?

Let me know what you all think.

streamer45:
Very interesting. I was just noticing some serious "flakyness" during plugin activation in a 3 nodes cluster which led to only a subset of the nodes having the plugin running. Wondering if this could be directly related.

agnivade:
It might be, if your plugin relies on `a.notifyPluginEnable` or `a.notifyPluginStatusesChanged` event. These methods are called from `a.SyncPluginsActiveState` which will always fail in the source machine, because the cluster broadcast event hasn't been done yet.

But they should run on the destination machines, because the same will get called again on the other machines, after the cluster broadcast event has reached them.

streamer45:
That's compatible with what I am seeing. Only one instance of three fails to start (I am assuming the source). I'll try to confirm if this is related because it might change how we decide to approach this.

lieut-data:
At first glance, this feels /more complicated/. I wholly agree with the sentiment that we need to revisit the design holistically.

> So, `a.SyncPluginsActiveState()` eventually calls `a.GetPluginStatus` which tries to fetch plugin status from the entire cluster. But since the cluster message hasn't been broadcasted yet, this fails.

Does it suffice to make `GetPluginStatus` not fail and instead return an empty set when a peer doesn't have plugins enabled? It looks like we're just using the output of that to broadcast a state to websocket clients (notably, admins), but it seems perfectly reasonable to broadcast that some nodes don't have a status because they aren't running the plugin yet.

agnivade:
> Does it suffice to make GetPluginStatus not fail and instead return an empty set when a peer doesn't have plugins enabled? It looks like we're just using the output of that to broadcast a state to websocket clients (notably, admins), but it seems perfectly reasonable to broadcast that some nodes don't have a status because they aren't running the plugin yet.

Yes absolutely. I wanted someone with more plugin experience to weigh in on this, but if this sounds good to you, then I can implement this.

@streamer45 - I don't think this approach should conflict with your config overhaul because this is just application logic. But let me know.

streamer45:
> @streamer45 - I don't think this approach should conflict with your config overhaul because this is just application logic. But let me know.

I wouldn't worry too much about conflicting with the overhaul also because I am yet to get deep into that logic and we can always decide to revisit when I get to a PR stage.

What I am unsure about though is this additional `broadcast` flag we are passing to `loadLockedWithOld()`. This appears to be altering the expected behaviour that a config change should trigger an event/callback on configured listeners.

agnivade:
> What I am unsure about though is this additional broadcast flag we are passing to loadLockedWithOld(). This appears to be altering the expected behaviour that a config change should trigger an event/callback on configured listeners.

Oh no no, I meant Jesse's [approach](https://github.com/mattermost/mattermost-server/pull/17511#issuecomment-827558823), not what I did in the PR.

streamer45:
> Oh no no, I meant Jesse's [approach](https://github.com/mattermost/mattermost-server/pull/17511#issuecomment-827558823), not what I did in the PR.

That seems a reasonable (and probably cleaner) solution. No objections from my side.

agnivade:
@streamer45 @lieut-data - This is good to be reviewed. The other part is here https://github.com/mattermost/enterprise/pull/987. Sorry I couldn't add any tests. There is no good way to unit test cluster behavior as of now. We'll have to go with manual tests.

crspeller:
@agnivade I wonder if we should be removing the call to SyncPluginsActiveState from the check on line 173 in InitPlugins? It get's called every config change too. (the condition is a bit confusing, I don't really get why the second part of the || is there as that should never be the case?)

app/plugin.go
```diff
@@ -210,9 +210,13 @@ func (a *App) InitPlugins(pluginDir, webappPluginDir string) {
 	// Sync plugin active state when config changes. Also notify plugins.
 	a.Srv().PluginsLock.Lock()
 	a.RemoveConfigListener(a.Srv().PluginConfigListenerId)
-	a.Srv().PluginConfigListenerId = a.AddConfigListener(func(*model.Config, *model.Config) {
-		a.installFeatureFlagPlugins()
-		a.SyncPluginsActiveState()
+	a.Srv().PluginConfigListenerId = a.AddConfigListener(func(old, new *model.Config) {
+		// If plugin status remains unchanged, only then run this.
+		// Because (*App).InitPlugins is already run as a config change hook.
+		if *old.PluginSettings.Enable == *new.PluginSettings.Enable {
```
lieut-data, line 216: Is this gate still relevant? Wondering how individual plugins being enabled or disabled actually happen without the `a.SyncPluginsActiveState`? (Or is that in the "other" config listener?)

agnivade:
@crspeller - So the `InitPlugins` will only be called from a config change only when plugins are enabled. Otherwise, it just calls `ShutdownPlugins`.

```go
a.AddConfigListener(func(prevCfg, cfg *model.Config) {
	if *cfg.PluginSettings.Enable {
		a.InitPlugins(*cfg.PluginSettings.Directory, *a.Config().PluginSettings.ClientDirectory)
	} else {
		a.srv.ShutDownPlugins()
	}
})
```

And `SyncPluginsActiveState` also calls `pluginsEnvironment.Shutdown()` if plugins are disabled, which is a subset of `a.srv.ShutDownPlugins()`. Overall, yes I think these parts could be refactored. But it's slightly out of scope for this PR.

All this PR does is it fixes the race condition where config update is sent before the cluster message gets sent, and makes `a.GetClusterPluginStatuses()` to just return empty status if plugins are disabled.

app/plugin.go
agnivade: Yes, this is relevant to prevent the same code from being run twice.

> Wondering how individual plugins being enabled or disabled actually happen without the a.SyncPluginsActiveState? (Or is that in the "other" config listener?)

Yes. See my PR message :wink: 

> And that
calls a.SyncPluginsActiveState(). But inside a.InitPlugins,
there is yet another config listener which again calls
a.SyncPluginsActiveState(). The first fix is to simply not call
the method again from the nested listener.

lieut-data (approved):

streamer45 (approved):

crspeller (approved):
Ok. Sounds good.

PR #17518, opened 2021-04-26
MM-35181 Move SharedChannelsUsers.ChannelId column add to 5.36 migration

wiggin77:
#### Summary
See discussion here https://community.mattermost.com/private-core/pl/n6aznzgxfjro9k9m5m6fajigjo

Background:  column was added for a PR that didn’t make it into 5.35, but the migration was added to 5.35.  The column migration was removed from 5.35 and this ticket is to put it into 5.36

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-35181

#### Release Note
```release-note
NONE
```

### DIFF
--- a/store/sqlstore/upgrade.go
+++ b/store/sqlstore/upgrade.go
@@ -20,6 +20,7 @@ import (
 
 const (
 	CurrentSchemaVersion   = Version5350
+	Version5360            = "5.36.0"
 	Version5350            = "5.35.0"
 	Version5340            = "5.34.0"
 	Version5330            = "5.33.0"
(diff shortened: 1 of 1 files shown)


streamer45 (approved):
Looks good, thanks Doug :+1:

cpanato (approved):

cpanato:
this will not break community-daily?

wiggin77:
> this will not break community-daily?

@cpanato I don't see how it would break community but I'm not sure how community gets updated.  If community gets a full refresh from master (instead of cherry picks where some commits are missed) then it should be fine.  It's unfortunate we don't have an easy way to test upgrading from populated databases.

PR #17523, opened 2021-04-27
[MM-35222] Fix DoGuestRolesCreationMigration

streamer45:
#### Summary

This migration was failing during scheme validation since we were setting both channel and team roles.
PR fixes this by only setting the role based on the scope for the scheme.

I am not very experienced with our permissions system so please let me know if you think this should be addressed differently.

#### Ticket Link

https://mattermost.atlassian.net/browse/MM-35222

```release-note
NONE
```

### DIFF
--- a/app/migrations.go
+++ b/app/migrations.go
@@ -206,28 +206,28 @@ func (a *App) DoGuestRolesCreationMigration() {
 				} else {
 					scheme.DefaultTeamGuestRole = savedRole.Name
 				}
-			} else if scheme.Scope == model.SCHEME_SCOPE_CHANNEL {
-				// Channel Guest Role
-				channelGuestRole := &model.Role{
-					Name:          model.NewId(),
-					DisplayName:   fmt.Sprintf("Channel Guest Role for Scheme %s", scheme.Name),
-					Permissions:   roles[model.CHANNEL_GUEST_ROLE_ID].Permissions,
-					SchemeManaged: true,
-				}
-
-				if savedRole, err := a.Srv().Store.Role().Save(channelGuestRole); err != nil {
-					mlog.Critical("Failed to create new guest role for custom scheme.", mlog.Err(err))
-					allSucceeded = false
-				} else {
-					scheme.DefaultChannelGuestRole = savedRole.Name
-				}
 			}
 
-			_, err := a.Srv().Store.Scheme().Save(scheme)
-			if err != nil {
-				mlog.Critical("Failed to update custom scheme.", mlog.Err(err))
+			// Channel Guest Role
+			channelGuestRole := &model.Role{
+				Name:          model.NewId(),
+				DisplayName:   fmt.Sprintf("Channel Guest Role for Scheme %s", scheme.Name),
+				Permissions:   roles[model.CHANNEL_GUEST_ROLE_ID].Permissions,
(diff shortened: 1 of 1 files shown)


hahmadia (approved):
Tagging Scott Bishel as well just to make sure I've not missed anything. Makes sense to me otherwise.

amyblais:
@streamer45 Is this needed for v5.35?

streamer45:
> @streamer45 Is this needed for v5.35?

@amyblais 
I think it would be good to include it. It's not a regression as far as I can see but we can probably prevent some issues by having this fixed sooner rather than later.

app/migrations.go
```diff
@@ -191,34 +191,36 @@ func (a *App) DoGuestRolesCreationMigration() {
...
+					SchemeManaged: true,
+				}
+
+				if savedRole, err := a.Srv().Store.Role().Save(teamGuestRole); err != nil {
+					mlog.Critical("Failed to create new guest role for custom scheme.", mlog.Err(err))
+					allSucceeded = false
+				} else {
+					scheme.DefaultTeamGuestRole = savedRole.Name
+				}
+			} else if scheme.Scope == model.SCHEME_SCOPE_CHANNEL {
```
jespino, line 209: I think this `else if` is wrong. If I remember correctly (and If I'm interpreting correctly the code in `store/sqlstore/scheme_store.go:115`), the `DefaultChannelGuestRole` is going to be part any `SCHEME_SCOPE_TEAM` and `SCHEME_SCOPE_CHANNEL` so... the `DefaultTeamGuestRole` is only for the `SCHEME_SCOPE_TEAM` but the `DefaultChannelGuestRole` is for both.

streamer45: That makes sense. That means we only need the check on the team scope.

sbishel: Yes, `SCHEME_SCOPE_TEAM` should get the Channel roles as well. So the `else if` isn't needed.

app/migrations.go
```diff
@@ -220,12 +222,12 @@ func (a *App) DoGuestRolesCreationMigration() {
...
 				scheme.DefaultChannelGuestRole = savedRole.Name
 			}
+		}
 
-			_, err := a.Srv().Store.Scheme().Save(scheme)
-			if err != nil {
-				mlog.Critical("Failed to update custom scheme.", mlog.Err(err))
-				allSucceeded = false
-			}
+		_, err := a.Srv().Store.Scheme().Save(scheme)
```
jespino, line 227: Now you are saving this for every scheme not only for the ones that needs the migration. It should go inside the if block checking for empty DefaultChannelGuestRole or DefaultTeamGuestRole.

streamer45: Oh dear, I've messed up with the brackets. Thanks.

jespino (approved):
LGTM

app/migrations.go
jespino: yes, gofmt is great until it fixes something that you didn't expect to be fixed :)

lindy65:
@streamer45 - I've asked in QA channel for an SDET to review this PR and test once ticket is resolved as it involves restarting the server and checking `systems` table.

sbishel (approved):

PR #17525, opened 2021-04-27
Fix file info index batch

jespino:
#### Summary
Fix file info index batch

#### Release Note
```release-note
NONE
```


ethervoid (approved):

jespino:
Removing QA because I tested it by myself locally and it still is behind a feature flag.

PR #17534, opened 2021-04-28
MM-35239: Fix racy test HubSessionRevokeRace

agnivade:
registerDummyWebConn already registers a webconn
to the hub. We were registering it again later.

This was incorrect. The test just passed nevertheless,
but there were 2 webconns getting added to the same userID.

https://mattermost.atlassian.net/browse/MM-35239

```release-note
NONE
```


streamer45 (approved):
Nice! :+1:

PR #17537, opened 2021-04-28
MM-35290 CRT: Reply on a unfollowed thread causes re-following

reflog:
#### Summary

avoid re-following thread when a new reply comes in without mention
#### Ticket Link

https://mattermost.atlassian.net/browse/MM-35290
#### Release Note

```release-note
NONE
```

### DIFF
--- a/app/notification.go
+++ b/app/notification.go
@@ -194,10 +194,13 @@ func (a *App) SendNotifications(post *model.Post, team *model.Team, channel *mod
 				// if the user was not explicitly mentioned, check if they explicitly unfollowed the thread
 				if !incrementMentions {
 					membership, err := a.Srv().Store.Thread().GetMembershipForUser(userID, post.RootId)
-					if nfErr := new(store.ErrNotFound); err != nil && !errors.As(err, &nfErr) {
+					var nfErr *store.ErrNotFound
+
+					if err != nil && !errors.As(err, &nfErr) {
 						mac <- model.NewAppError("SendNotifications", "app.channel.autofollow.app_error", nil, err.Error(), http.StatusInternalServerError)
 						return
 					}
+
 					if membership != nil && !membership.Following {
 						return
 					}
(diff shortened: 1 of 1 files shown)


app/notification.go
```diff
@@ -191,6 +191,17 @@ func (a *App) SendNotifications(post *model.Post, team *model.Team, channel *mod
 			go func(userID string) {
 				defer close(mac)
 				_, incrementMentions := mentions.Mentions[userID]
+				// if the user was not explicitly mentioned, check if they explicitly unfollowed the thread
+				if !incrementMentions {
+					membership, err := a.Srv().Store.Thread().GetMembershipForUser(userID, post.RootId)
+					if nfErr := new(store.ErrNotFound); err != nil && !errors.As(err, &nfErr) {
```
jespino, line 197: This like is very confusing, I think it shouldn't be inlined everything.

jespino (changes requested):
A bit of improvement in the readability of this code is needed, but otherwise looks good to me.

app/notification.go
reflog: @jespino this is cleaner?
```
					var nfErr *store.ErrNotFound
		
					if err != nil && !errors.As(err, &nfErr) {
						mac <- model.NewAppError("SendNotifications", "app.channel.autofollow.app_error", nil, err.Error(), http.StatusInternalServerError)
						return
					}
```

jespino: Yes it is.

jespino (approved):
LGTM

ashishbhate (approved):

ogi-m:
@reflog In which scenario was this happening? I couldn't repro on master with the following steps:
1. User-1 posts a root post
2. User-2 replies and the thread shows up in the user's Threads section
3. User-2 explicitly unfollows the thread
4. User-1 posts a new reply in the thread
Expected & Observed: The thread isn't re-followed for User-2

The scenario in which I could reproduce on master but can't on the PR is the one where there's a mention in the root post:
1. User-1 posts a root post with a mention for User-2
2. User-2 explicitly unfollows the thread
3. User-1 posts a new reply in the thread
Expected (and observed on the PR): Thread isn't re-followed for User-2
Observed on master: Thread is re-followed for User-2 and shows up in Threads

If the above was the scenario that was fixed, then it's tested and passed.

reflog:
@ogi-m , yup it's the second scenario. let's approve and merge this bad boy

ogi-m (approved):

PR #17546, opened 2021-04-29
[MM-35316] Fix panic in app.CreateSidebarCategory

streamer45:
#### Summary

PR fixes a potential panic caused by wrapping a `nil` error in  `SqlChannelStore.CreateSidebarCategory()`.

#### Ticket

https://mattermost.atlassian.net/browse/MM-35316

#### Release Note

```release-note
NONE
```

### DIFF
--- a/store/sqlstore/channel_store_categories.go
+++ b/store/sqlstore/channel_store_categories.go
@@ -248,10 +248,8 @@ func (s SqlChannelStore) CreateSidebarCategory(userId, teamId string, newCategor
 	categoriesWithOrder, err := s.getSidebarCategoriesT(transaction, userId, teamId)
 	if err != nil {
 		return nil, err
-	}
-
-	if len(categoriesWithOrder.Categories) < 1 {
-		return nil, errors.Wrap(err, "categories not found")
+	} else if len(categoriesWithOrder.Categories) == 0 {
+		return nil, store.NewErrNotFound("categories not found", fmt.Sprintf("userId=%s,teamId=%s", userId, teamId))
 	}
 
 	newOrder := categoriesWithOrder.Order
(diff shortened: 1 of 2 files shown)


agnivade (approved):

store/sqlstore/channel_store_categories.go
```diff
@@ -478,8 +476,11 @@ func (s SqlChannelStore) getSidebarCategoriesT(db dbSelecter, userId, teamId str
 	}
 
 	if _, err = db.Select(&categories, query, args...); err != nil {
+		return nil, errors.Wrapf(err, "failed to get categories for userId=%s,teamId=%s", userId, teamId)
```
agnivade, line 479: Ahh looks like you have to remove this.

streamer45: I think there's a bit of misusage of not founds errors in the store. Also I am not sure we are very consistent with the semantics. E.g., should an empty slice be considered a not found?

agnivade: Totally. Following the semantics of `database/sql`, for a query that returns a single row, a not found is meant to return an error. If the query returns multiple errors, no rows should return an empty slice.

But I am more for consistency than following the stdlib. We can decide whatever, but we should be consistent.

hmhealey (commented):
What's the actual fix here? It looks like the final change ended up just turning two if statements into an if/else-if

streamer45:
> What's the actual fix here? It looks like the final change ended up just turning two if statements into an if/else-if

The fix is removing `errors.Wrap(err, "categories not found")`. Wrapping a `nil` error returns `nil` so we could have the function returning `nil, nil` which caused a panic on the calling side.

hmhealey (approved):

amyblais:
@streamer45 Is this only on master and how urgent is this?

streamer45:
@amyblais

I'd say not very urgent. It looks like this bug has been there for a while. That said, it's very low risk so it might be worth to include it in 5.35.

PR #17553, opened 2021-05-02
Change displayname of System user to Info

adanielvv:
#### Summary
Change system user display name to "Info". Why:
First impression of a new user to Mattermost: System?! It looks like as if the Terminator is chatting with us! 

#### Release Note
```release-note
Change display name of system user to "Info"
```

### DIFF
--- a/i18n/en.json
+++ b/i18n/en.json
@@ -9068,7 +9068,7 @@
   },
   {
     "id": "system.message.name",
-    "translation": "System"
+    "translation": "Info"
   },
   {
     "id": "web.command_webhook.command.app_error",


adanielvv:
```release-note
Change display name of system user to "Info"
```

adanielvv:
Just a one off PR

And what is this? Change needed here too?
```
"post_info.system": "System",
```
https://github.com/mattermost/mattermost-webapp/blob/master/i18n/en.json#L3793
https://github.com/mattermost/mattermost-mobile/blob/master/assets/base/i18n/en.json#L629

hmhealey:
We're having a bit of a discussion about this on our community server. Sounds like people like the idea of changing the name of the System user, but we need to decide if we're going to do that and if Info is the right name there. We'll also have to figure out where else might need to be updated here since I think the web/mobile apps will also need to be changed, and the docs might also have to change.

That discussion is happening here: https://community.mattermost.com/core/pl/6fauo8n6styp7y6xn1d4ocma6e

hmhealey:
I've bumped this discussion on the Mattermost side. We might need to create a Jira ticket to keep track of it going forward to make sure it's not forgotten about.

aaronrothschild:
cc @esethna for visibility

esethna:
Appreciate the PR! I opened a ticket to think through this here, propose we close this PR in the mean time until we land on a solution that is approved by the design team: https://mattermost.atlassian.net/browse/MM-35937

PR #17570, opened 2021-05-04
Custom status expiry

manojmalik20:
#### Summary
Added expiry support in custom status APIs

#### Related Links
Specs: https://mattermost.atlassian.net/wiki/spaces/CS/pages/983302547/User+Experience+and+Functional+Notes

#### Related pull requests
- https://github.com/mattermost/mattermost-api-reference/pull/636
- https://github.com/mattermost/mattermost-webapp/pull/8059

#### Release Note
```release-note
Added two new fields in CustomStatus struct and modified the APIs to validate and handle them
```


esethna:
@lieut-data @hmhealey I think you guys had helped with getting the dev spec in a good place for this part of the feature (adding expiry times to custom statuses). Can you guys help with reviews and/or hand-offs/mindmelds to whoever is best to move this forward from the dev side?

esethna:
Moved my PM review over to the webapp PR, I'll lean on the devs here for the server PR approvals

model/custom_status.go
```diff
@@ -15,9 +16,21 @@ const (
 	MaxRecentCustomStatuses  = 5
 )
 
+var ValidCustomStatusDuration = map[string]bool{
+	"dont_clear":     true,
```
lieut-data, line 20: Should we handle the empty string as "don't clear" instead of making it explicit? As a user of the REST API, it would seem like duration should be optional.

model/custom_status.go
```diff
@@ -15,9 +16,21 @@ const (
...
+	"this_week":      true,
+	"date_and_time":  true,
+}
+
 type CustomStatus struct {
-	Emoji string `json:"emoji"`
-	Text  string `json:"text"`
+	Emoji     string    `json:"emoji"`
+	Text      string    `json:"text"`
+	Duration  string    `json:"duration"`
```
lieut-data, line 32: Instead of modelling `Duration` and `ExpiresAt`, would we consider modelling `LastSet` and `Duration`? This gives us strictly more information, in always recording the last time a custom status was set, even if no expiration is configured. (We can then compute `ExpiresAt` trivially where needed.)

model/custom_status.go
```diff
@@ -15,9 +16,21 @@ const (
 	MaxRecentCustomStatuses  = 5
 )
 
+var ValidCustomStatusDuration = map[string]bool{
```
lieut-data, line 19: nit: This doesn't feel like something we should export. What if we just inlined a switch statement into the corresponding method instead?

app/slashcommands/command_custom_status.go
```diff
@@ -19,7 +20,8 @@ const (
 	CmdCustomStatus      = app.CmdCustomStatusTrigger
 	CmdCustomStatusClear = "clear"
 
-	DefaultCustomStatusEmoji = "speech_balloon"
+	DefaultCustomStatusEmoji    = "speech_balloon"
+	DefaultCustomStatusDuration = "today"
```
lieut-data, line 24: Why is the default to expire after one day, and not simply stay forever?

app/slashcommands/command_custom_status.go
```diff
@@ -83,3 +94,22 @@ func (*CustomStatusProvider) DoCommand(a *app.App, args *model.CommandArgs, mess
 		}),
 	}
 }
+
+func calculateExpiryTime(userTimezone model.StringMap, timezoneEnabled bool) time.Time {
```
lieut-data, line 98: nit: This method sounds more generic than it is. Could we name it, `calculateEndOfDay` instead?

lieut-data (commented):
Drive by feedback, but take it with a grain of salt since I don't have intimate insight into the product requirements :)

api4/status.go
```diff
@@ -127,7 +127,7 @@ func updateUserCustomStatus(c *Context, w http.ResponseWriter, r *http.Request)
 	}
 
 	customStatus := model.CustomStatusFromJson(r.Body)
-	if customStatus == nil || (customStatus.Text == "" && customStatus.Emoji == "") {
+	if customStatus == nil || customStatus.Emoji == "" || !customStatus.IsDurationValid() || !customStatus.IsExpirationTimeValid() {
```
hmhealey, line 130: Is it intentional that we're now requiring emojis to have an emoji now? Previously, we required at least one of emoji or text.

model/custom_status.go
hmhealey: We could just lower case it to avoid exporting it. Our tendency to capitalize all constants does mean we're exporting a lot.

app/slashcommands/command_custom_status.go
manojmalik20: It was in the spec - https://mattermost.atlassian.net/wiki/spaces/CS/pages/983302547/User+Experience+and+Functional+Notes

model/custom_status.go
manojmalik20: If a user chooses a custom date/time, we would have to calculate duration based on the difference between that time and the current time. Then in the selector, we will have to calculate the expiry time based on that duration and the current time. Considering how much changes would have to be made and how much benefit would that give, I don't think we should change our approach. What do you think?

api4/status.go
manojmalik20: Previously when we required one of emoji and text, it was wrong because not sending emoji in the request body was setting the emoji as an empty string. We can do two things - we can make emoji required or we can set speech_balloon when emoji is not sent in the body. So if you want to go with the latter one, I can make that change. What do you think?

model/custom_status.go
lieut-data: Ah, yeah, if the experience offers directly control of the `ExpiresAt`, this approach makes sense. Thanks for the clarifications :)

app/slashcommands/command_custom_status.go
lieut-data: @esethna, gut feel is that we should default the user interface to expire after one day, but keep the expiration optional and defaulting to never if not explicitly specified in the REST API.

api4/status.go
esethna: It is defined in the spec that if the emoji is not set then it defaults to the speech balloon

app/slashcommands/command_custom_status.go
esethna: @lieut-data I think that is the intention if I'm understanding correctly. The UI selector shows "today" as the default option when the users go set their status (and as such statuses will clear the next day if this is not changed), but users can change it to "Don't clear" if they want it to live forever.

image

model/custom_status.go
manojmalik20: I think we should handle the default duration in the same way for both the API and slash commands. So we can either set default duration to "today" or "don't clear" in both API and slash commands? What do you think?

app/slashcommands/command_custom_status.go
lieut-data: Sounds good, @esethna! @manojmalik20, can we make that distinction, namely to have the UI default as described, but leave the default of the REST API to never expire?

api4/status.go
hmhealey: @esethna Any thoughts about changing this? I'm not sure what the spec says here about the custom status requiring text or an emoji?

model/custom_status.go
lieut-data: My proposal would be to treat the empty string as "don't clear" -- as a user of the REST API, if I forget to specify this field, I'd love to have the behaviour of permanently setting the status. I can then opt-in to the auto-expiry by adding the field and providing it to the REST API.

api4/status.go
esethna: @hmhealey sorry I must not be understanding. We don't want to make any changes to how it works today. Meaning:

- Users can set an emoji with no text
- Users can set text without an emoji (in this case the emoji defaults to the speech balloon)
- Users can set both text and emoji

model/custom_status.go
```diff
@@ -33,6 +56,16 @@ func (cs *CustomStatus) ToJson() string {
...
 }
 
+func (cs *CustomStatus) IsDurationValid() bool {
+	return cs.Duration == "" || validCustomStatusDuration[cs.Duration]
+}
+
+func (cs *CustomStatus) IsExpirationTimeValid() bool {
+	isStatusDurationEmpty := cs.Duration == "" || cs.Duration == "dont_clear"
+	isStatusAlreadyExpired := cs.ExpiresAt.IsZero() || cs.ExpiresAt.Before(time.Now())
+	return isStatusDurationEmpty || !isStatusAlreadyExpired
```
lieut-data, line 66: If I'm reading this correctly, we don't check if `ExpiresAt` is valid when `Duration` is empty. Does that mean I can't just send up `{ExpiresAt: "2021-05-31T00:00:00"}` but /also/ have to specify a `Duration`?

lieut-data (changes requested):
Thanks, @manojmalik20! Just following up on the semantics of `Duration` and `ExpiresAt` below.

model/custom_status.go
manojmalik20: Yes, because not sending a duration means never clearing the status.

manojmalik20: But I'm thinking if `duration` is empty and `expires_at` is not, then we should set `duration` to custom `date_and_time` and set the expiry time. I think this would be better. What do you think?

lieut-data: I wonder if we could make a case for never remembering `Duration`, and just always setting `ExpiresAt`? @esethna, is it important that the client remember the last duration I picked for the next emoji I want to set? I feel like it would be more useful to keep this simple and exclusively model `ExpiresAt` server-side.

esethna: >  @esethna, is it important that the client remember the last duration I picked for the next emoji I want to set

No that's not important in my mind. I'd imagine we reset the expires at time in the UI to the default (tomorrow) for any new custom statuses set

manojmalik20: The field `Duration` is needed for the recent custom statuses that we show in the modal along with the suggestions. Without this field, we will not be able to use the recent statuses with the clear time as defined in the [spec](https://mattermost.atlassian.net/wiki/spaces/CS/pages/983302547/User+Experience+and+Functional+Notes).

lieut-data: Thanks for the extra context, @manojmalik20! Indeed, `Duration` proves useful as you describe, and a `date_and_time` option for when `ExpiresAt` is used suffices.

hmhealey (approved):

lieut-data (approved):
Thanks again, @manojmalik20!

manojmalik20:
@ogi-m Did you have a chance to review/test this?

ogi-m (approved):
@manojmalik20 Let's merge and then test on the webapp PR 👍🏻

manojmalik20:
@ogi-m Can you also add the `EETests` label in this PR?

manojmalik20:
@esethna Can you please merge this PR also?

esethna:
@manojmalik20 a dev has to merge as I don't have permissions. cc// @lieut-data @hmhealey @isacikgoz

PR #17577, opened 2021-05-05
check authdata for different format

sbishel:
#### Summary
The Oid value returned by Office365 when using OpenId is a properly formated GUID. (ie. '0e8fddd4-50d3-4499-9a93-a390ee8cb83d')

However, it appears that when returned from Office365 Oauth, the value did not contain the "-". (ie. '0e8fddd450d344999a93a390ee8cb83d')

There were previous cases of these being padded with zeros as well.
'be95fe607df5dbeb' -> '00000000-0000-0000-be95-fe607df5dbeb' 

This is another attempt to fix. Rather than attempting to look up by the formatted AuthData. When a user with the same email is found, we attempt to compare the non-dashed portions of the AuthData to what is currently stored in the database.

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-35012

#### Release Note
```release-note
Handles condition when the AuthData is formatted differently between Office365 OAuth and Office 365 OpenID
```
#### Related PRs
https://github.com/mattermost/enterprise/pull/990


app/user.go
```diff
@@ -375,7 +375,10 @@ func (a *App) CreateOAuthUser(service string, userData io.Reader, teamID string,
 		if userByEmail.AuthService == "" {
 			return nil, model.NewAppError("CreateOAuthUser", "api.user.create_oauth_user.already_attached.app_error", map[string]interface{}{"Service": service, "Auth": model.USER_AUTH_SERVICE_EMAIL}, "email="+user.Email, http.StatusBadRequest)
 		}
-		return nil, model.NewAppError("CreateOAuthUser", "api.user.create_oauth_user.already_attached.app_error", map[string]interface{}{"Service": service, "Auth": userByEmail.AuthService}, "email="+user.Email, http.StatusBadRequest)
+		if provider.VerifySameUser(userByEmail, user) {
```
mkraft, line 378: Couple of questions:
* What's the benefit of this being in a method vs `if userByEmail.AuthData == user.AuthData`?
* If it's going to be a method can it be renamed such that it's obvious what the truth case is, for example `IsSameUser` or `HasSameAuthDataValue`?
* Is Gitlab the only OAuth provider that needs to implement it?
* Are there any tests that can be updated to reflect this newly found scenario?

sbishel: Yes...I renamed this at least 4 times. I will go with `IsSameUser`, if there are no objections.

mkraft: > Yes...I renamed this at least 4 times. I will go with `IsSameUser`, if there are no objections.

Sounds good to me.

mkraft (approved):

hahmadia (approved):

sbishel:
I added updating the authdata (and added unit test) as we discussed @mkraft 
Please re-review.
Thanks. 
cc @hahmadia

hahmadia (approved):
New changes make sense. Thanks!

app/user.go
```diff
@@ -376,7 +376,14 @@ func (a *App) CreateOAuthUser(c *request.Context, service string, userData io.Re
 		if userByEmail.AuthService == "" {
 			return nil, model.NewAppError("CreateOAuthUser", "api.user.create_oauth_user.already_attached.app_error", map[string]interface{}{"Service": service, "Auth": model.USER_AUTH_SERVICE_EMAIL}, "email="+user.Email, http.StatusBadRequest)
 		}
-		return nil, model.NewAppError("CreateOAuthUser", "api.user.create_oauth_user.already_attached.app_error", map[string]interface{}{"Service": service, "Auth": userByEmail.AuthService}, "email="+user.Email, http.StatusBadRequest)
+		if provider.IsSameUser(userByEmail, user) {
+			if _, err := a.Srv().Store.User().UpdateAuthData(userByEmail.Id, user.AuthService, user.AuthData, "", false); err != nil {
+				// if the user is not updated, write a warning to the log, but don't prevent user login
+				mlog.Warn("Error attempting to update user AuthData", mlog.Err(err))
```
mkraft, line 382: Was this `mlog.Warn` intentionally because this update isn't critical, or was it meant to be `mlog.Error`?

mkraft (approved):

app/user.go
sbishel: `mlog.Warn` was intentional. Yes, the update is more of a "convenience", so I didn't think an error was warranted.

furqanmlk (approved):
1. Login via OAuth Office 365 
2. Remove dashes authdata in database

image

3. Migrate to OpenID connect
4. Login via Office 365
5. authdata has dashes back --> Expected

image

PR #17578, opened 2021-05-06
[MM-32390] Config logic refactor

streamer45:
#### Summary

PR attempts to improve the overall config storing logic and also cleanup the package from unnecessary or unrelated code.

The most significant change is the elimination of `loadLockedWithOld()` config store method and the complete separation of logic between `Set()` and `Load()` methods. 

Another notable change is around the management of the `FeatureFlags` section. The semantic has been slightly changed (from persist to read-only) to make it clearer what's happening behind the hood and possibly allow for a more general solution in the future (e.g. multiple read-only config sections).

Other minor changes include moving stuff out of the `config` package (but otherwise unaltering its behaviour), some tests additions/fixes and some added documentation.

There's one outstanding (and important) change (file watcher logic, now tracked by [MM-35470](https://mattermost.atlassian.net/browse/MM-35470)) that we will merge in the August release. This is due to its breaking nature. We are in the process of notifying customers and users so that we can then proceed with the deprecation and final removal.

#### Ticket

https://mattermost.atlassian.net/browse/MM-32390

#### Release Note

```release-note
Refactored the config storing logic to improve its robustness and performance.
```


agnivade:
I am assuming this also fixes the config.json reset on service restart? We should also link that ticket with the PR.

streamer45:
> I am assuming this also fixes the config.json reset on service restart? We should also link that ticket with the PR.

I am not so sure. I think that was caused by the file watcher functionality which is still there. I couldn't reproduce that one myself and I am waiting for some debug data (file accesses) from Sven to investigate further.

agnivade:
Ah gotcha. All good then :+1:

agnivade:
Hey @streamer45 - sorry I could not review this today. I will be OOO next week. If you want to get this in within next week, please feel free to un-assign me.

streamer45:
> Hey @streamer45 - sorry I could not review this today. I will be OOO next week. If you want to get this in within next week, please feel free to un-assign me.

There's no rush on this whatsoever. As long as it's reviewed before August we should be fine :)

app/feature_flags.go
```diff
@@ -21,7 +20,9 @@ func (s *Server) setupFeatureFlags() {
 	splitConfigured := splitKey != ""
 	syncFeatureFlags := splitConfigured && s.IsLeader()
 
-	s.configStore.PersistFeatures(splitConfigured)
+	if splitConfigured {
+		s.configStore.SetReadOnlyFF(false)
+	}
```
crspeller, line 25:
```suggestion
	s.configStore.SetReadOnlyFF(!splitConfigured)
```
Otherwise this relies on an not intuitive default of true.

config/store.go
```diff
@@ -112,27 +128,14 @@ func NewTestMemoryStore() *Store {
...
-
 // Get fetches the current, cached configuration.
 func (s *Store) Get() *model.Config {
 	s.configLock.RLock()
 	defer s.configLock.RUnlock()
 	return s.config
 }
 
-// Get fetches the current, cached configuration without environment variable overrides.
+// Get fetches the current cached configuration without environment variable overrides.
```
crspeller, line 138:
```suggestion
// GetNoEnv fetches the current cached configuration without environment variable overrides.
```

config/store.go
```diff
@@ -90,14 +89,31 @@ func NewStoreFromBacking(backingStore BackingStore, customDefaults *model.Config
 	return store, nil
 }
 
-func getBackingStore(dsn string, watch bool) (BackingStore, error) {
+// NewStore creates and returns a new config store backed by either a database or file store
```
crspeller, line 92:
```suggestion
// NewStoreFromDSN creates and returns a new config store backed by either a database or file store
```

crspeller (changes requested):

config/store.go
```diff
@@ -188,124 +196,155 @@ func (s *Store) Set(newCfg *model.Config) (*model.Config, error) {
...
 	}
 
-	loadedConfig.SetDefaults()
-
-	s.configNoEnv = loadedConfig.Clone()
-	fixConfig(s.configNoEnv)
+	// We set the SiteURL to empty (if nil) so that the following call to
+	// SetDefaults() will generate missing data. This avoids an additional write
+	// to the backing store.
+	if loadedCfg.ServiceSettings.SiteURL == nil {
```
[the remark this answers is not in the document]
streamer45: Sorry, that was a bit cryptic. There are some fields that get generated on `SetDefaults` (e.g. `SqlSettings.AtRestEncryptKey`) by passing an `isUpdate` boolean that depends on `SiteURL`:

https://github.com/mattermost/mattermost-server/blob/d97daaa1ded834b9c9c339ebfbf12078a54b4810/model/config.go#L3197-L3203

This caused an additional write during config initialization.

crspeller (approved):

app/feature_flags_sync.go
```diff
@@ -1,7 +1,7 @@
 // Copyright (c) 2015-present Mattermost, Inc. All Rights Reserved.
 // See LICENSE.txt for license information.
 
-package config
+package app
```
agnivade, line 4: Would it make sense to have this as a sub-package under app? 0/5.

config/store.go
```diff
@@ -188,124 +196,155 @@ func (s *Store) Set(newCfg *model.Config) (*model.Config, error) {
...
+	}
+
+	s.configNoEnv = newCfgNoEnv
+	s.config = newCfg
+
+	unlockOnce.Do(s.configLock.Unlock)
+
+	if hasChanged {
+		s.invokeConfigListeners(oldCfg, newCfg.Clone())
 	}
```
agnivade, line 244: Can we try to refactor this condition outside the `Set` and `Load` methods and remove the `unlockOnce` contraption? I believe it was added to work around the additional complexity of tracking different return paths, which is much simplified now. Would be good to get rid of this.

config/store.go
```diff
@@ -188,124 +196,155 @@ func (s *Store) Set(newCfg *model.Config) (*model.Config, error) {
...
 
-	if err := s.backingStore.Set(newCfg); err != nil {
+	if err := s.backingStore.Set(newCfgNoEnv); err != nil {
 		return nil, errors.Wrap(err, "failed to persist")
 	}
 
-	if err := s.loadLockedWithOld(oldCfg, &unlockOnce); err != nil {
-		return nil, errors.Wrap(err, "failed to load on save")
+	newCfg = applyEnvironmentMap(newCfgNoEnv, GetEnvironment())
```
agnivade, line 219: Looking deeper into this, `s.GetEnvironmentOverrides` work on the same set as `GetEnvironment` as the previous is just:

```
func (s *Store) GetEnvironmentOverrides() map[string]interface{} {
	return generateEnvironmentMap(GetEnvironment(), nil)
}
```
So is there a benefit in remove env overrides, and then adding the same overrides again? Can't we just reuse `newCfg`?

config/store.go
```diff
@@ -188,124 +196,155 @@ func (s *Store) Set(newCfg *model.Config) (*model.Config, error) {
...
+	fixConfig(newCfg)
+	if err := newCfg.IsValid(); err != nil {
+		return nil, errors.Wrap(err, "new configuration is invalid")
+	}
+
+	hasChanged, err := confsDiff(oldCfg, newCfg)
```
agnivade, line 225: Maybe better renamed as `isDifferent(oldCfg, newCfg)`.

config/store.go
```diff
@@ -175,9 +179,13 @@ func (s *Store) Set(newCfg *model.Config) (*model.Config, error) {
 		return nil, ErrReadOnlyStore
 	}
 
-	oldCfg := s.config.Clone()
+	newCfg = newCfg.Clone()
```
agnivade, line 182: It is not entirely clear to me the need to clone here. Is it out of safety or there is a reason for this?

config/store.go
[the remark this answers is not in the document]
agnivade: Hmm .. this is not super ideal. It makes the config package aware of the application logic. It should not need to know what a `SiteURL` is or why it should be nil. We need an explicit way to denote if a config is initialized or not - perhaps by an unexported variable. Out of scope for this PR, but something to improve on later.

config/store.go
```diff
@@ -188,124 +196,155 @@ func (s *Store) Set(newCfg *model.Config) (*model.Config, error) {
...
+	if loadedCfg.ServiceSettings.SiteURL == nil {
+		loadedCfg.ServiceSettings.SiteURL = model.NewString("")
+	}
 
-	loadedConfig = applyEnvironmentMap(loadedConfig, GetEnvironment())
+	// Setting defaults allows us to accept partial config objects.
+	loadedCfg.SetDefaults()
 
-	fixConfig(loadedConfig)
+	loadedCfgNoEnv := loadedCfg
```
agnivade, line 294: I was wondering why is it not cloned here since it's just a pointer, but then realized that `applyEnvironmentMap` does the cloning internally. Wondering if it might be better to clone it explicitly and let `applyEnvironmentMap` just work on the pointer? 0/5.

Alternatively, just add a comment mentioning what's happening here so that it's clear for the reader.

config/store.go
```diff
@@ -188,124 +196,155 @@ func (s *Store) Set(newCfg *model.Config) (*model.Config, error) {
...
-			err := s.backingStore.Set(s.configNoEnv)
-			s.configNoEnv.FeatureFlags = featureFlags
-			if err != nil && !errors.Is(err, ErrReadOnlyConfiguration) {
-				return errors.Wrap(err, "failed to persist")
-			}
+	// We write back to the backing store only if the store is not read-only
+	// and the config has either changed or is missing.
+	if !s.readOnly && (hasChanged || len(configBytes) == 0) {
+		err := s.backingStore.Set(loadedCfgNoEnv)
+		if err != nil && !errors.Is(err, ErrReadOnlyConfiguration) {
```
agnivade, line 324: Is there a need to check for `!errors.Is(err, ErrReadOnlyConfiguration)` here?

agnivade (commented):
Brilliant work! Just some comments for me to understand better.

app/feature_flags_sync.go
streamer45: Sure, that works as well.

config/store.go
streamer45: I originally had the same sentiment although I can see how it's efficiently and cleanly used to avoid calling `configLock.Unlock()` every time we return early and at the same time allow for an unlocked code-path. Do you find that anti-idiomatic? To refactor this we would either need an additional wrapping method, some defer magic (it would probably get even more verbose) or just be okay with unlocking/locking before/after the listeners invocation.

config/store.go
streamer45: Doing that would be assuming `newCfg` (user input) has the environment overrides already set. There are cases where it might not be like that hence the need to remove/apply them anyway.

config/store.go
streamer45: Yes, as far as I could see I've added it only out of safety and respect to the caller's data.

config/store.go
streamer45: Yeah. At the beginning of this PR I've attempted to remove all the internal clones but it was getting harder to keep track of what was safe and what wasn't so eventually I left it there. Make sense to add a comment.

config/store.go
streamer45: This precedes me so I can only speculate but I believe we need that because internally the file won't get updated in case of a read-only config but we might still want to invoke the listeners and otherwise update the cached version.

config/store.go
agnivade: Ok, perhaps makes sense to add a comment to clarify that? It gets hard to keep all of this in mind.

config/store.go
agnivade: > Do you find that anti-idiomatic?

Yes, slightly. It is a trick which ultimately hurts readability in the long run IMO. I am okay with both wrapping or unlocking/locking.

config/store.go
streamer45: Went with the shorter `equal()` since it has that stdlib vibe. Hope that's acceptable :)

agnivade: All good :+1:

config/store.go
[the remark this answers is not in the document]
streamer45: Totally agree on that. Unfortunately the package has still application logic dependant code (`desanitize()` and `fixConfig()` are some examples). Will open a ticket to track this aspect as well.

agnivade (approved):
Brilliant work! And totally love the sub-package. App is already a monster, so it feels good to see some separations being made.

streamer45:
@saturninoabril This touches the whole core logic of how we handle configs. It would be helpful to see some config related e2e tests passing and maybe a manual smoke-test around config load/update in various scenarios (HA especially).

saturninoabril:
Test on going at https://automation-dashboard.vercel.app/cycles/7b38284d-6f35-4a57-8388-5325fccc7191

saturninoabril (approved):
Tested and passed. Failed tests on Cypress were known issues.

PR #17581, opened 2021-05-06
MM-35332 honour config setting when extending session expiry

wiggin77:
#### Summary
This PR fixes a bug in `(*App)ExtendSessionExpiryIfNeeded` which extended session expiration regardless of `ServiceSettings.ExtendSessionLengthWithActivity` config setting.

Also adds a unit test that tests for the setting being disabled.

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-35332

#### Release Note
```release-note
Fixes bug where session expiration was extended with activity regardless of config setting (ServiceSettings.ExtendSessionLengthWithActivity).
```

### DIFF
--- a/app/session_test.go
+++ b/app/session_test.go
@@ -319,41 +319,51 @@ func TestApp_ExtendExpiryIfNeeded(t *testing.T) {
 		require.False(t, session.IsExpired())
 	})
 
-	var tests = []struct {
-		name    string
-		session *model.Session
-	}{
-		{name: "mobile", session: &model.Session{UserId: model.NewId(), DeviceId: model.NewId(), Token: model.NewId()}},
-		{name: "SSO", session: &model.Session{UserId: model.NewId(), IsOAuth: true, Token: model.NewId()}},
-		{name: "web/LDAP", session: &model.Session{UserId: model.NewId(), Token: model.NewId()}},
+	for i := 0; i <= 1; i++ {
+		enabled := i > 0
+		th.App.UpdateConfig(func(cfg *model.Config) { *cfg.ServiceSettings.ExtendSessionLengthWithActivity = enabled })
+
+		var tests = []struct {
+			name    string
+			session *model.Session
+		}{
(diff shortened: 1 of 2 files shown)


app/session_test.go
```diff
@@ -319,41 +319,51 @@ func TestApp_ExtendExpiryIfNeeded(t *testing.T) {
...
-			require.NoError(t, errGet)
-			require.Equal(t, session.ExpiresAt, cachedSession.ExpiresAt)
-
-			// check database was updated.
-			storedSession, nErr := th.App.Srv().Store.Session().Get(context.Background(), session.Token)
-			require.NoError(t, nErr)
-			require.Equal(t, session.ExpiresAt, storedSession.ExpiresAt)
-		})
-	}
+	for i := 0; i <= 1; i++ {
```
crspeller, line 322: Nit: This confused me. Maybe including this as a parameter in the tests would be more clear?

app/session_test.go
```diff
@@ -319,41 +319,51 @@ func TestApp_ExtendExpiryIfNeeded(t *testing.T) {
...
+				session, err := th.App.CreateSession(test.session)
+				require.Nil(t, err)
+
+				expires := model.GetMillis() + th.App.GetSessionLengthInMillis(session) - hourMillis
+				session.ExpiresAt = expires
 
+				ok := th.App.ExtendSessionExpiryIfNeeded(session)
+
+				if !enabled {
+					require.False(t, ok)
```
crspeller, line 346: Maybe we should verify it was not updated?

crspeller (approved):
Just a couple comments on the test.

isacikgoz (approved):
LGTM 👍

amyblais:
@wiggin77 How risky is this fix? Trying to think if cherry-picking to v5.35 would make sense, but I don't want to introduce a last minute regression given today is code freeze.

saturninoabril:
fyi, Cypress tests added - https://github.com/mattermost/mattermost-webapp/pull/8087

PR #17582, opened 2021-05-06
MM-35430 sanitize user before creation during sync

wiggin77:
#### Summary
This PR fixes a bug where incoming users were not sanitized before being created during shared channel sync.

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-35430

#### Release Note
```release-note
Fixed Shared Channels issue where users were not sanitized during remote synchronization.
```

### DIFF
--- a/services/sharedchannel/sync_recv.go
+++ b/services/sharedchannel/sync_recv.go
@@ -228,6 +228,9 @@ func (scs *Service) insertSyncUser(user *model.User, channel *model.Channel, rc 
 	var userSaved *model.User
 	var suffix string
 
+	// ensure the new user is created with system_user role and random password.
+	user = sanitizeUserForSync(user)
+
 	// save the original username and email in props (if not already done by another remote)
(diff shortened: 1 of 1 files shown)


services/sharedchannel/sync_recv.go
```diff
@@ -290,6 +293,7 @@ func (scs *Service) updateSyncUser(patch *model.UserPatch, user *model.User, cha
 	}
 
 	user.Patch(patch)
+	user = sanitizeUserForSync(user)
```
jupenur, line 296: This looks like it might be unnecessary? `user` is the existing user from the database and `patch` doesn't contain any of the fields that need to be sanitized.

jupenur (approved):

streamer45 (approved):

services/sharedchannel/sync_recv.go
wiggin77: The patch doesn't contain any security related fields, but we also don't want more innocuous stuff like turning on notifications for a fake user.

PR #17591, opened 2021-05-10
Removes JS redirection

anurag6713:
#### Summary
We are unable to redirect with JS when `Content Security Policy` is enabled and restricted with `unsafe-inline`.

Thread:
https://community.mattermost.com/core/pl/33bnmwq3gtfgfeutg8bbg8mhdh

To fix this we are redirecting using the `meta` tag and also added a link in case user redirection is cancelled (on ios, browser shows a prompt whether to redirect or not)

Screenshot 2021-05-10 at 6 09 25 PM

#### Release Note
```release-note
* Replaced JS redirection with html meta tag for OAUTH/SAML.
* Added link to redirect back to the app if redirection fails.
```


enahum (approved):
should we include this in a dot release? cc: @amyblais

hahmadia (approved):

furqanmlk (commented):
**Android**

image

### **Note: I don't know how to fail redirect to see the link**

amyblais:
@anurag6713 Please help take a look at Malik's comment ^

@lindalumitchell Once Malik's comment above has been addressed, we'll likely need your (or another QA's) help to test this PR. I saw in the stand-up channel that Malik is off tomorrow.

anurag6713:
@furqanmlk 

We can use the same CSP header used by `framasoft`. 

in `utils/api.go`

```go
func RenderMobileMessage(w http.ResponseWriter, message string) {
	w.Header().Set("Content-Type", "text/html")
        // ADD THIS LINE
	w.Header().Set("Content-Security-Policy", "script-src 'self' framasoft.org stats.framasoft.org; connect-src 'self' framasoft.org api.gfycat.com wss://framateam.org; style-src 'self' framasoft.org 'unsafe-inline'")
```

After making the above change, without this PR, you will not be redirected back to app.

With this PR, you will be redirected back to app.

furqanmlk:
> @furqanmlk I'm sorry, I should have mentioned this earlier 🙂 
> 
> 
> 
> We can use the same CSP header used by `framasoft`. 
> 
> 
> 
> in `utils/api.go`
> 
> 
> 
> ```go
> 
> func RenderMobileMessage(w http.ResponseWriter, message string) {
> 
> 	w.Header().Set("Content-Type", "text/html")
> 
>         // ADD THIS LINE
> 
> 	w.Header().Set("Content-Security-Policy", "script-src 'self' framasoft.org stats.framasoft.org; connect-src 'self' framasoft.org api.gfycat.com wss://framateam.org; style-src 'self' framasoft.org 'unsafe-inline'")
> 
> ```
> 
> 
> 
> After making the above change, without this PR, you will not be redirected back to app.
> 
> 
> 
> With this PR, you will be redirected back to app.

Thanks @anurag6713 
No worries, Let me re-test :)

furqanmlk (commented):
### **Android**

image


### **IOS**

image

@anurag6713 , thanks for the help. I am getting expected page but there is Green Check Icon, is that expected ?

anurag6713:
@furqanmlk Yes .. As the green tick comes from a CDN, It doesn't load because of the header we have set.

furqanmlk (approved):
Thanks @anurag6713

amyblais:
@anurag6713 Can you help cherry-pick to ``release-5.35`` and ``cloud`` branches?

PR #17602, opened 2021-05-11
[MM-35424] Fix job schedulers server from missing leader change event

streamer45:
#### Summary

PR fixes an issue where the job scheduler server could potentially miss a "leader change" event. The proposed changes are:

- Moved the schedulers server start so that it happens before doing potentially long-running tasks (e.g. plugins initialization)
- Changed the `handleClusterLeaderChange` semantic to guarantee that the latest change is gets delivered.
- Fixed a possible deadlock case in the config change handler.

#### Ticket

https://mattermost.atlassian.net/browse/MM-35424

#### Release Note

```release-note
Fixed an issue where the job scheduler server could miss a "changed leader" cluster event.
```

### DIFF
--- a/jobs/schedulers.go
+++ b/jobs/schedulers.go
@@ -232,7 +232,10 @@ func (schedulers *Schedulers) scheduleJob(cfg *model.Config, scheduler model.Sch
 
 func (schedulers *Schedulers) handleConfigChange(oldConfig, newConfig *model.Config) {
 	mlog.Debug("Schedulers received config change.")
-	schedulers.configChanged <- newConfig
+	select {
+	case schedulers.configChanged <- newConfig:
+	case <-schedulers.stop:
+	}
 }
 
 func (schedulers *Schedulers) handleClusterLeaderChange(isLeader bool) {
--- a/jobs/schedulers.go
+++ b/jobs/schedulers.go
@@ -230,7 +230,7 @@ func (schedulers *Schedulers) scheduleJob(cfg *model.Config, scheduler model.Sch
 	return scheduler.ScheduleJob(cfg, pendingJobs, lastSuccessfulJob)
 }
 
-func (schedulers *Schedulers) handleConfigChange(oldConfig, newConfig *model.Config) {
+func (schedulers *Schedulers) handleConfigChange(_, newConfig *model.Config) {
 	mlog.Debug("Schedulers received config change.")
 	select {
 	case schedulers.configChanged <- newConfig:
--- a/jobs/schedulers.go
+++ b/jobs/schedulers.go
@@ -44,7 +44,7 @@ func (srv *JobServer) InitSchedulers() error {
 		stop:                 make(chan bool),
 		stopped:              make(chan bool),
 		configChanged:        make(chan *model.Config),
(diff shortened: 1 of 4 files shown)


isacikgoz (approved):
Nice 👌 Just some small things

jobs/schedulers.go
```diff
@@ -232,13 +232,26 @@ func (schedulers *Schedulers) scheduleJob(cfg *model.Config, scheduler model.Sch
...
+	}
 }
 
-func (schedulers *Schedulers) HandleClusterLeaderChange(isLeader bool) {
+func (schedulers *Schedulers) handleClusterLeaderChange(isLeader bool) {
 	select {
 	case schedulers.clusterLeaderChanged <- isLeader:
 	default:
-		mlog.Debug("Did not send cluster leader change message to schedulers as no schedulers listening to notification channel.")
+		mlog.Debug("Sending cluster leader change message to schedulers failed.")
```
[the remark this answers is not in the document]
streamer45: We get there if the channel is full. That means that an event already got there (the chan is buffered of length 1) but the goroutine in `Start()` has not yet handled it either because it was not running or doing something else.

By how we moved things around now it would be a rather rare event but still possible so I wouldn't regard it as a warning really. Before the changes we would lose the message completely so it was a problem but now we remove the existing one and enqueue the latest so eventually the latest message will get there (unless the server shuts down before that of course).

jobs/schedulers.go
```diff
@@ -230,15 +230,28 @@ func (schedulers *Schedulers) scheduleJob(cfg *model.Config, scheduler model.Sch
...
+	select {
+	case schedulers.configChanged <- newConfig:
+	case <-schedulers.stop:
+	}
 }
 
-func (schedulers *Schedulers) HandleClusterLeaderChange(isLeader bool) {
+func (schedulers *Schedulers) handleClusterLeaderChange(isLeader bool) {
 	select {
 	case schedulers.clusterLeaderChanged <- isLeader:
```
agnivade, line 243: Wondering what will happen if we make this blocking?

streamer45: I am thinking that could cause a deadlock if the schedulers were not started yet when such event came since all those calls are called from methods of `JobServer` under a lock.

agnivade: Ah yeah .. we should really sort out these initialization steps and start the scheduler before setting up clustering. @isacikgoz 's app changes are a first step towards that. But we have more to do :) Anyways, out of scope for this PR.

agnivade (approved):

streamer45:
@saturninoabril 

I am thinking a general smoke-test around jobs and schedulers properly working in HA should cover this.

saturninoabril (approved):
Tested and passed, especially the ff and seemed to be working as expected:
- with server joining and leaving the cluster, and full HA mode
  - plugin install/uninstall, enable/disable
  - compliance export

Screen Shot 2021-05-19 at 12 17 16 AM
Screen Shot 2021-05-19 at 12 22 02 AM
Screen Shot 2021-05-19 at 12 25 20 AM

PR #17620, opened 2021-05-13
MM-31612 Stream import archive contents to bulk importer

gigawhitlocks:
#### Summary

This PR optimizes the import process by streaming archive contents to the bulk importer instead of unzipping the incoming archive to disk. I believe it should resolve MM-31612 but I have taken a different approach than I suspect was expected, so I'm going to explain the approach I've taken here and then together we can decide the fate of MM-31612 and whether this resolves it fully or whether we feel that there should be some follow-up work for a "cleaner" solution.

First of all, when I first conceived of this solution, I thought it was kind of a hack, but the more I've thought about it, the more I think it's a legitimate solution, so if you start reading this description and think "wow this is a weird approach" please bear with me while I explain the approach and reasoning behind it.

##### Background and More Obvious Approaches
Performing this optimization without leveraging the underlying OS would require taking the `*zip.Reader` and modifying a large portion of the Bulk Import and import_process worker code to be able to access files within the `*zip.Reader` directly instead of doing what it does today, for historical reasons, which is to use one `Reader` pointing to the JSONL file, and then after parsing that, when the importer finds an attached file, the application opens the file in the expected location specified in the JSONL file using `os.Open`, which expects a normal file.

Due to the use of `os.Open`, the way the Importer determines the location of the attached file, and the general need to instead pass a `Reader` for each file to the extraction routine, I determined that it would require a considerable amount of manna to modify the existing extraction routine to take `Reader`s instead of using `os.Open` after parsing the JSONL file. 

Another approach I considered, after discovering how the files are accessed by the Bulk Importer, was to modify the JSONL file format to potentially allow for encoding a S3 URL (seems messy, also very complicated) and to allow the Bulk Importer to access S3 directly, but this would've required deep changes to the Bulk Importer as well as redoing some of the work in the mattermost/cloud and mattermost/awat repos. 

##### Solution: Named Pipes
In order to avoid a massive change to the Bulk Importer, this implementation opens the incoming Zip archive in memory and creates a facsimile of the unzipped contents in the same location where the old code would simply extract the zip file. Instead of containing normal files, however, each file is a named pipe (or FIFO, e.g. https://man7.org/linux/man-pages/man3/mkfifo.3.html ) which is backed by a goroutine ready to stream, decompress, and write the correct file contents as soon as a reader becomes available. 

Let us recall that `Write` called on a FIFO blocks until a reader is ready, so the call to `io.Copy` inside of the goroutine will block until the call to `os.Open` by the bulk import routine is made, at which point the goroutine will write the bits from the Zip file (wherever it is, even in S3) into the pipe and the Bulk Importer will receive those bits from `os.Open` without being aware that these are not normal decompressed files, and that they are never written to the file system.

##### Benefits of this approach:
- It's robust and parallelizes well
- It's self-contained, and a fairly small change
- It leverages the tools the OS gives us to make our lives easier

##### Caveats to this approach:
1) The unpack location used by the bulk importer is set in part via convention and the structure and location of that archive when it is unzipped is assumed in several places. A solution to this issue that didn't use FIFOs may have been able to remove this convention and make it configurable or fully remove the need for an unpack location, but it would be a much more far-reaching change.

2) This approach _only works on Linux_. It's not even possible to compile the Server for Windows or Darwin targets whilst including the call to `syscall.Mkfifo`. However, Go makes it simple for us to keep the old behavior for non-Linux targets. This seems acceptable to me since Linux is our only supported production target. 

#### Ticket Link

Potentially fixes https://mattermost.atlassian.net/browse/MM-31612 and https://github.com/mattermost/mattermost-server/issues/16617

#### Release Note

```release-note
Optimized the bulk import process on Linux to avoid writing files to disk
```

### DIFF
--- a/i18n/en.json
+++ b/i18n/en.json
@@ -7460,7 +7460,7 @@
   },
   {
     "id": "import_process.worker.do_job.file_exists",
-    "translation": "Unable to process import: file does not exists."
+    "translation": "Unable to process import: file does not exist."
   },
   {
     "id": "import_process.worker.do_job.missing_file",
@@ -7481,6 +7481,10 @@
   {
     "id": "import_process.worker.do_job.unzip",
     "translation": "Unable to process import: failed to unzip file."
+  },
+  {
+    "id": "import_process.worker.do_job.write_file",
+    "translation": "Unable to process import: failed to write file."
   },
   {
     "id": "interactive_message.decode_trigger_id.base64_decode_failed",
--- a/jobs/import_process/worker.go
+++ b/jobs/import_process/worker.go
@@ -4,13 +4,10 @@
 package import_process
 
 import (
-	"io"
 	"io/ioutil"
 	"net/http"
 	"os"
 	"path/filepath"
-	"runtime"
-	"strconv"
 
 	"github.com/mattermost/mattermost-server/v5/app"
 	"github.com/mattermost/mattermost-server/v5/app/request"
@@ -18,7 +15,6 @@ import (
 	tjobs "github.com/mattermost/mattermost-server/v5/jobs/interfaces"
 	"github.com/mattermost/mattermost-server/v5/model"
 	"github.com/mattermost/mattermost-server/v5/shared/mlog"
-	"github.com/mattermost/mattermost-server/v5/utils"
 )
 
 func init() {
@@ -125,10 +121,6 @@ func (w *ImportProcessWorker) doJob(job *model.Job) {
 	}
 	defer importFile.Close()
 
-	// TODO (MM-30187): improve this process by eliminating the need to unzip the import
-	// file locally and instead do the whole bulk import process in memory by
-	// streaming the import file.
-
 	// create a temporary dir to extract the zipped import file.
 	dir, err := ioutil.TempDir("", "import")
 	if err != nil {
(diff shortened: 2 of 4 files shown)


streamer45:
Wow this is a weird approach :stuck_out_tongue: 

I can't but appreciate the out-of-the-box thinking here and love how the solution translates into fairly simple code.  
That said, I am bit conflicted with the proposed approach for a couple of reasons:

- As you pointed out, having OS dependent implementations is not great (to say the least :P). 
- It feels a bit like we fixed the strictly needed to unblock some other work and left the mess behind for others to deal with later on.
- With the proposed solution, it looks like we are creating an unbounded number of goroutines (one per file essentially). This is generally not a great idea and I can see this being problematic for huge imports with tens of thousands (or more) files.

I must admit that I haven't thought too deeply at a solution when I first created the ticket but it seemed to me like passing a map of paths to readers to the `app.BulkImport` method (or some variant of it) would have been almost enough.
Of course some internal changes would have been needed inside the import functions but it should have been trivial and that's why I decided to make the ticket as "help-wanted". However, maybe there's more complexity I haven't considered.

gigawhitlocks:
If you're feeling hesitant about this @streamer45 , especially if this feels like "a mess left behind," I'll investigate a more traditional approach more deeply, I don't want to leave behind tech debt, but this approach occurred to me and I was able to write the proof of concept quickly so I thought it was worth proposing

I can keep the branch around in case a more traditional approach fails, in which case we can revisit this and talk about it again

streamer45:
@gigawhitlocks 

That was my (very) raw feeling but I definitely think the solution is valid and certainly good enough to get reviewed and merged. 
Maybe we could enlarge the discussion and include someone else just to get a second opinion on this.

Overall I am just thinking that if the required effort isn't massive (and it doesn't look like it at first glance) then I would much prefer the "once and for all" fix.

cc @mgdelacroix and @isacikgoz to get more feedback.

gigawhitlocks:
Let me take a look again at the traditional fix today and see if I overestimated the effort. It seemed like a much larger change to me but it's very possible that I overestimated. You certainly understand this codebase better than I do

gigawhitlocks:
Regarding the unbounded number of goroutines, do note that these are blocked goroutines and since they are all paused on IO_WAIT until read, it should be possible to have hundreds of thousands of them (or even more, according to that link)

streamer45:
> Regarding the unbounded number of goroutines, do note that these are blocked goroutines since they are all paused on IO_WAIT until read, it should be possible to have [hundreds of thousands of them](https://stackoverflow.com/questions/8509152/max-number-of-goroutines) (or even more, according to that link)

Yes, generally speaking goroutines are cheap enough to be created in the thousands. But there are other factors to consider in this case (e.g. open file descriptors) and from a performance perspective is not ideal to let them increase without control.

gigawhitlocks:
Sure, although it's worth pointing out this solution opens the exact same number of file descriptors as the implementation it's replacing

streamer45:
> Sure, although it's worth pointing out this solution opens the exact same number of file descriptors as the implementation it's replacing

Yes, but not at the same time, unless I am missing something. With this implementation we a spawning N goroutines and opening N files all at once and only closing (exiting) during the actual import process.

gigawhitlocks:
Oh that's true, they're _open_ files in this case. A very large number will eventually get you the `too many open files` error from the kernel, so that's a limitation that does have to be considered, thanks

gigawhitlocks:
Actually I think that might be a serious limitation. Number of open files allowed per process is a tunable kernel parameter but relying on a kernel setting runs against good practice for containerized software since the kernel is determined by the container host

gigawhitlocks:
I'm going to close this at least for now while I do a deeper investigation of a more normal approach

PR #17624, opened 2021-05-14
[MM-35767] Ignore readTimeout param from MySQL datasource

streamer45:
#### Summary

PR fixes an issue where long-running migrations (on MySQL) could still time out. While we are not setting a timeout for migrations anymore at the app layer (since https://github.com/mattermost/mattermost-server/pull/17444) there was still a timeout happening at the network (tcp) layer governed by the `readTimeout` query parameter from the datasource.

PR fixes this by ignoring the provided parameter and resetting it to zero (no timeout). This should be also safe for queries running outside the "migration phase" since those will respect the timeout value defined at the application level through the `SqlSettings.QueryTimeout` setting.

cc @amyblais 

#### Ticket

https://github.com/mattermost/mattermost-server/pull/new/MM-35767

#### Release Note

```release-note
Fixed an issue where long migrations where failing with an "invalid connection" error on MySQL installations.
```

### DIFF
--- a/store/sqlstore/store.go
+++ b/store/sqlstore/store.go
@@ -351,7 +351,17 @@ func (ss *SqlStore) Context() context.Context {
 }
 
 func (ss *SqlStore) initConnection() {
-	ss.master = setupConnection("master", *ss.settings.DataSource, ss.settings)
+	dataSource := *ss.settings.DataSource
+	if ss.DriverName() == model.DATABASE_DRIVER_MYSQL {
+		var err error
+		dataSource, err = ss.resetReadTimeout(*ss.settings.DataSource)
+		if err != nil {
+			mlog.Critical("Failed to reset read timeout from datasource.", mlog.Err(err))
+			os.Exit(ExitGenericFailure)
+		}
+	}
+
+	ss.master = setupConnection("master", dataSource, ss.settings)
 
 	if len(ss.settings.DataSourceReplicas) > 0 {
 		ss.Replicas = make([]*gorp.DbMap, len(ss.settings.DataSourceReplicas))
@@ -1487,6 +1497,15 @@ func (ss *SqlStore) appendMultipleStatementsFlag(dataSource string) (string, err
 	}
 
 	return dataSource, nil
+}
+
+func (ss *SqlStore) resetReadTimeout(dataSource string) (string, error) {
+	config, err := mysql.ParseDSN(dataSource)
+	if err != nil {
+		return "", err
+	}
+	config.ReadTimeout = 0
+	return config.FormatDSN(), nil
 }
 
 type mattermConverter struct{}
--- a/store/sqlstore/store.go
+++ b/store/sqlstore/store.go
@@ -353,8 +353,11 @@ func (ss *SqlStore) Context() context.Context {
 func (ss *SqlStore) initConnection() {
 	dataSource := *ss.settings.DataSource
 	if ss.DriverName() == model.DATABASE_DRIVER_MYSQL {
+		// TODO: We ignore the readTimeout datasource parameter for MySQL since QueryTimeout
+		// covers that already. Ideally we'd like to do this only for the upgrade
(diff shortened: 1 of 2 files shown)


amyblais:
@crspeller @wiggin77 This is a candidate for v5.35 dot release.

wiggin77 (approved):
LGTM :+1:

agnivade (commented):
While I totally agree that setting this globally has no effect because queryTimeout will still take effect, I think this can cause unexpected behaviour from an operator's perspective. The `read_timeout` is still a valid parameter in the MySQL datasource and a sysadmin might want to set different values for read_timeout and query_timeout.

I am 0/5 on whether we want to do this globally or not, but if we do this, I would also suggest to document this more strongly that we do not respect `read_timeout` anymore.

streamer45:
> While I totally agree that setting this globally has no effect because queryTimeout will still take effect, I think this can cause unexpected behaviour from an operator's perspective. The `read_timeout` is still a valid parameter in the MySQL datasource and a sysadmin might want to set different values for read_timeout and query_timeout.
> 
> I am 0/5 on whether we want to do this globally or not, but if we do this, I would also suggest to document this more strongly that we do not respect `read_timeout` anymore.

Yes, if we go with this we would have to update all the existing documentation. Overall I couldn't think of convincing reasons for keeping two timeouts where the lower one will override the other. I think it creates more confusion than anything.

That said, ideally we'd only want to ignore the `readTimeout` during migrations but the way connections are set up doesn't make it possible unless we decide to create brand new connections (without timeouts) just for the upgrade process and close them as soon as that's finished.

agnivade:
The new migration code indeed creates a new connection:

```
conn := setupConnection("migrations", dataSource, ss.settings)
defer conn.Db.Close()
```

It's just that it isn't fully activated yet.

That said, totally up to you. I am 0/5. Just that if we choose to ignore `read_timeout`, we should document it.

streamer45:
> The new migration code indeed creates a new connection:
> 
> ```
> conn := setupConnection("migrations", dataSource, ss.settings)
> defer conn.Db.Close()
> ```
> 
> It's just that it isn't fully activated yet.
> 
> That said, totally up to you. I am 0/5. Just that if we choose to ignore `read_timeout`, we should document it.

Took a brief look and think that's a valid option although it would most probably require extensive code changes to deal with all the calls both from store methods and upgrade functions.
Also, given the new system is coming fairly soon maybe we could just go with this patch to prevent the imminent issues on this release and then review it once we deprecate the old system.

agnivade (approved):

streamer45:
@amyblais  We might want to add an additional note for MySQL installations saying that the `readTimeout` datasource query parameter is now being ignored and the application provided `SqlSettings.QueryTimeout` should be used instead.

amyblais:
@streamer45 Can you also help cherry-pick this to ``cloud``?

What documentation needs to be updated?

streamer45:
> @streamer45 Can you also help cherry-pick this to `cloud`?

Yes, let me do that as well.

> What documentation needs to be updated?

I'd probably stand by on that as we might have to change it again soon anyway and it would also require an additional PR to remove default config values. I think a release note would be good enough for now.

Essentially we would be eliminating the `readTimeout` from the suggested MySQL config datasource (pretty much any reference to it, e.g. https://docs.mattermost.com/search.html?q=readTimeout&check_keywords=yes&area=default)

streamer45:
@amyblais 

Thinking more about it, this won't have any effect on cloud since PR only affects MySQL (cloud being on PostgreSQL). We might just skip this.

amyblais:
@streamer45 Self-managed releases are based off cloud releases.

streamer45:
> @streamer45 Self-managed releases are based off cloud releases.

Thanks for reminding me that, you are 100% right :facepalm: . Will proceed with the cherry-pick.

amyblais:
@streamer45 Please help resolve the Jira ticket with QA test steps.

PR #17636, opened 2021-05-18
Remove redundant additions to cache hit/miss counters

streamer45:
#### Summary

PR fixes a small issue where cache hit/miss metrics counters for user cache were getting incremented/decremented more than needed. `rootStore.doStandardReadCache()` already takes care of that internally. Also, we were using a different cache name which made the resulting metrics more confusing.

#### Release Note

```release-note
NONE
```

### DIFF
--- a/store/localcachelayer/user_layer.go
+++ b/store/localcachelayer/user_layer.go
@@ -136,11 +136,6 @@ func (s *LocalCacheUserStore) GetProfileByIds(ctx context.Context, userIds []str
 		}
 	}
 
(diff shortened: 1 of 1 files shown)


isacikgoz (approved):
LGTM 👍

agnivade (approved):
Are there any other instances like this? Or is this the only one?

streamer45:
> Are there any other instances like this? Or is this the only one?

I had a quick look and couldn't find more.

PR #17651, opened 2021-05-20
Update the about.mattermost.com links to the real ones

spirosoik:
#### Summary
We identified that sometimes the redirection to the right link is broken
so we want to be sure that our users will end up in the right place.

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-35893

#### Release Note
```release-note
NONE
```


hahmadia (approved):
Thanks!

PR #17652, opened 2021-05-20
put timed dnd feature behind feature flag

darkLord19:


darkLord19:
added in https://github.com/mattermost/mattermost-server/pull/17680

PR #17656, opened 2021-05-20
Use Node 14 build image for web app

hmhealey:
Just updating the server to match the client here. I'm not sure if the server is able to successfully build the web app at this point, but if it tries, it'll use the right Docker image at least.

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-34363

#### Related Pull Requests
https://github.com/mattermost/mattermost-webapp/pull/8132

#### Release Note
```release-note
NONE
```

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -17,7 +17,7 @@ jobs:
   setup:
     working_directory: ~/mattermost/mattermost-server
     docker:
-      - image: mattermost/mattermost-build-webapp:20200829_node-10.22
+      - image: mattermost/mattermost-build-webapp:20210518_node-14.16-1
     resource_class: xlarge
     steps:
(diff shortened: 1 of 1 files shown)


metanerd (approved):

agnivade (approved):
I am guessing Gitlab pipeline would also need to be updated. Would you be updating those as well?

hmhealey:
I don't think the GitLab pipeline on the server repo has anything to build the web app, but I'll double check.

hmhealey:
Yeah, there's no reference to the mattermost-build-webapp image in the GitLab pipeline. If something else needs to be updated to Node 14, I guess we'll find out, but hopefully the GitLab pipeline is just relying on prebuilt versions of the webapp.

PR #17660, opened 2021-05-20
Change about and help links back to default redirects

jasonblais:
Originally changed in https://github.com/mattermost/mattermost-server/pull/17651

We want to keep the about link and help link set to a default page that redirects to appropriate page. The reason is that we may, for instance, want to link to a different help link in the future. If we do so, we only have to change the redirect without changes to the codebase.

Did not revert the change to terms of service and privacy policy links, given we want to point these to mattermost.com


spirosoik (approved):

isacikgoz (approved):

agnivade:
_something something semver_ .. :rofl:

spirosoik:
@jasonblais I think we can merge

jasonblais:
@spirosoik I'm not authorized to merge the pull request, would you be open to help?

PR #17667, opened 2021-05-21
[MM-35931] Apply DB migrations regardless of version patch value

streamer45:
#### Summary

Saving dot releases versions in `model/version.go` had the outcome of storing that value in the `Systems` table for new installations.

Unfortunately though `store.shouldPerformUpgrade()` did a simple string equality comparison which would essentially fail for any dot release and prevent the migrations from happening, leaving everything frozen db schema wise.
I've changed the behaviour to actually parse the version and go ahead if both major and minor versions match, hence ignoring the patch version value.

Another approach I think could have been to add mock `upgradeDatabaseToVersion5XX()` methods for all the dot releases affected by this but it didn't look like a great solution.

#### Ticket

https://mattermost.atlassian.net/browse/MM-35931

#### Release Note

```release-note
Fixed an issue where subsequent migrations failed to run after running a dot release on new installations.
```

### DIFF
--- a/store/sqlstore/upgrade_test.go
+++ b/store/sqlstore/upgrade_test.go
@@ -13,6 +13,16 @@ import (
 	"github.com/mattermost/mattermost-server/v5/store"
 )
 
+func TestStoreUpgradeDotRelease(t *testing.T) {
+	StoreTest(t, func(t *testing.T, ss store.Store) {
+		sqlStore := ss.(*SqlStore)
+		saveSchemaVersion(sqlStore, "5.33.1")
+		err := upgradeDatabase(sqlStore, CurrentSchemaVersion)
+		require.NoError(t, err)
+		require.Equal(t, CurrentSchemaVersion, sqlStore.GetCurrentSchemaVersion())
+	})
+}
+
 func TestStoreUpgrade(t *testing.T) {
 	StoreTest(t, func(t *testing.T, ss store.Store) {
 		sqlStore := ss.(*SqlStore)
--- a/store/sqlstore/upgrade.go
+++ b/store/sqlstore/upgrade.go
@@ -223,9 +223,23 @@ func saveSchemaVersion(sqlStore *SqlStore, version string) {
 }
 
 func shouldPerformUpgrade(sqlStore *SqlStore, currentSchemaVersion string, expectedSchemaVersion string) bool {
-	if sqlStore.GetCurrentSchemaVersion() == currentSchemaVersion {
-		mlog.Warn("Attempting to upgrade the database schema version", mlog.String("current_version", currentSchemaVersion), mlog.String("new_version", expectedSchemaVersion))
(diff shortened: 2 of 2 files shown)


streamer45:
@amyblais  This would certainly need some cherry-picking (and possibly backporting).

amyblais:
@streamer45 Do you recommend backporting to v5.33 - v5.35, or just v5.35?

streamer45:
> @streamer45 Do you recommend backporting to v5.33 - v5.35, or just v5.35?

I think given the nature of the issue we should be patching up to ESR (included).

store/sqlstore/upgrade_test.go
```diff
@@ -13,6 +13,16 @@ import (
 	"github.com/mattermost/mattermost-server/v5/store"
 )
 
+func TestStoreUpgradeDotRelease(t *testing.T) {
+	StoreTest(t, func(t *testing.T, ss store.Store) {
+		sqlStore := ss.(*SqlStore)
+		saveSchemaVersion(sqlStore, "5.33.1")
+		err := upgradeDatabase(sqlStore, CurrentSchemaVersion)
```
agnivade, line 20: Would prefer to have a static string here rather than something which changes. 0/5

agnivade (approved):
Oh dear .. can't believe we missed it for so long. Maybe we never did schema changes for dot releases. Can't wait for this entire thing to be overhauled.

Just one last comment - this doesn't affect the cloud branch workflow in any way right?

store/sqlstore/upgrade_test.go
streamer45: I don't think it would work though because the final version gets set to whatever the latest upgrade version is.

streamer45:
> Oh dear .. can't believe we missed it for so long. Maybe we never did schema changes for dot releases. Can't wait for this entire thing to be overhauled.

Yeah, it's getting rather painful lately.

> Just one last comment - this doesn't affect the cloud branch workflow in any way right?

I'd say no. Also I think Cloud is somewhat immune since there we actually check if the schema matches what we expect before running schema upgrades.

cc @ethervoid for visibility.

store/sqlstore/upgrade_test.go
agnivade: Oh yeah my bad, this is the top level function. I thought you were testing `shouldPerformUpgrade` specifically.

hahmadia (approved):

agnivade:
The ee/circleCI 404 thing is a known issue and there is a ticket filed for it. I couldn't really figure that one out when I was working on mattermod. I'll probably surface this up to the release team and see if Carlos has any clue.

amyblais:
@streamer45 Let me know when this is cherry-picked to v5.31 - 5.35. Also, please help resolve the Jira ticket for QA.

PR #17692, opened 2021-05-31
Introduce per platform make package targets

evelikov:
#### Summary

Currently we have per platform `make build-foo` targets yet a single `make package` target.
Introduce separate `make package-foo` targets. Thus people can package only what they need.

As a nice bonus point, this allows macosx, windows and linux packaging to happen in parallel.

#### Ticket Link

N/A

#### Release Note
```release-note
Add per-platform `make package' targets - package-osx, package-linux and package-windows.
```

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,4 +1,4 @@
-.PHONY: build package run stop run-client run-server run-haserver stop-haserver stop-client stop-server restart restart-server restart-client restart-haserver start-docker clean-dist clean nuke check-style check-client-style check-server-style check-unit-tests test dist prepare-enteprise run-client-tests setup-run-client-tests cleanup-run-client-tests test-client build-linux build-osx build-windows internal-test-web-client vet run-server-for-web-client-tests diff-config prepackaged-plugins prepackaged-binaries test-server test-server-ee test-server-quick test-server-race start-docker-check migrations-bindata new-migration migration-prereqs
+.PHONY: build package run stop run-client run-server run-haserver stop-haserver stop-client stop-server restart restart-server restart-client restart-haserver start-docker clean-dist clean nuke check-style check-client-style check-server-style check-unit-tests test dist prepare-enteprise run-client-tests setup-run-client-tests cleanup-run-client-tests test-client build-linux build-osx build-windows package-prep package-linux package-osx package-windows internal-test-web-client vet run-server-for-web-client-tests diff-config prepackaged-plugins prepackaged-binaries test-server test-server-ee test-server-quick test-server-race start-docker-check migrations-bindata new-migration migration-prereqs
 
 ROOT := $(dir $(abspath $(lastword $(MAKEFILE_LIST))))
 
@@ -94,6 +94,9 @@ PLATFORM_FILES="./cmd/mattermost/main.go"
 # Output paths
 DIST_ROOT=dist
 DIST_PATH=$(DIST_ROOT)/mattermost
+DIST_PATH_LIN=$(DIST_ROOT)/linux/mattermost
+DIST_PATH_OSX=$(DIST_ROOT)/osx/mattermost
+DIST_PATH_WIN=$(DIST_ROOT)/windows/mattermost
 
 # Tests
 TESTS=.
--- a/build/release.mk
+++ b/build/release.mk
@@ -63,15 +63,10 @@ build-client:
 
 	cd $(BUILD_WEBAPP_DIR) && $(MAKE) build
 
-package:
+package-prep:
 	@ echo Packaging mattermost
 	@# Remove any old files
 	rm -Rf $(DIST_ROOT)
-
-	@# Create needed directories
-	mkdir -p $(DIST_PATH)/bin
-	mkdir -p $(DIST_PATH)/logs
-	mkdir -p $(DIST_PATH)/prepackaged_plugins
 
 	@# Resource directories
 	mkdir -p $(DIST_PATH)/config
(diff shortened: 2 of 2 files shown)


evelikov:
Force-pushed a fix which should get the windows package back to normal.

agnivade (commented):
0/5 on this. But I'll prefer the dist paths to not change. Rather, in the top-level package rule, we can keep doing the same logic of creating the archive and deleting it.

So `make package` gives the same output as before. And `make package-<platform>` generates the platform specific archive.

evelikov:
@agnivade thanks for having a look.
Can you elaborate a bit on your comment, is there anything I can tweak to make this better?

agnivade:
Sure, let me explain. Right now, for every platform, we do:

```
@# Package
	tar -C dist -czf $(DIST_PATH)-$(BUILD_TYPE_NAME)-osx-amd64.tar.gz mattermost
	@# Cleanup
	rm -f $(DIST_PATH)/bin/mattermost
	rm -f $(DIST_PATH)/bin/platform
	rm -f $(DIST_PATH)/bin/mmctl
```

In your PR, you are copying the binaries into different directories and then packaging them. I would like to keep the `DIST_PATH` same as before. And have each target copy the binaries into the same directory and then create the tarball.

And in between each step, you can run the cleanup.

So this won't break any tool expecting the binaries in that `DIST_PATH`.

Did that help?

evelikov:
Hmm let me try elaborating a bit:

Currently all artefacts - OS specific and "generic" - are fetched and archived in serial fashion. Effectively we download the "generic" artefacts and for each OS - copy the binary artefacts, tarball is created, removed binary artefacts.

This PR - keeps the "generic" artefacts as-is. Then for each OS the binary artefacts are copied in OS specific folder, tarball is created and the OS folder is purged.

So basically the amount of copying is effectively the same, while allowing for parallelism - one for each OS.

One gotcha in the original code is that the linux mattermost binary is _not_ removed from `DIST_PATH`. I kept with this behaviour in this PR - as you can see from the `mv` line and comment.

That said, I cannot see this PR might break any tooling.

```patch
- @# Don't clean up native package so dev machines will have an unzipped package available
- @#rm -f $(DIST_PATH)/bin/mattermost
+ @# Preserve native package so dev machines will have an unzipped package available
+ mv $(DIST_PATH_LIN)/bin $(DIST_PATH)/bin
```

agnivade:
You are right, it does not break any tooling. I was suggesting that we can do all of that without introducing new directory variables in the `Makefile`. Or is there a need to have the platform specific directories?

evelikov:
Yes we do need them, if we want to keep things simple and sane. Although having in `build/release.mk` is an option, if you prefer.

In more detail:
Keeping everything in `DIST_PATH` (as original) and executing "make package" will result in a races when two or all three sub-targets "package-<platform>" are executed concurrently. Be that downloading/managing the same generic artefacts or creating a "linux" tarball which also includes the windows/osx binaries.
 - the former can be handled with the proposed release-prep target
 - for the latter we need a carefully crafted file list to pass to `tar` `zip` - doable but fragile and error prone

evelikov:
With the proposed approach, one could even move the `tmpprepackaged` handling to the respective targets. Thus avoid downloading the packages for platforms they don't need and avoid the copying them around all together.

agnivade (approved):
No issues from my end. I will defer to the build team. Thanks @evelikov !

evelikov:
@agnivade any chance you can poke the build team?

agnivade:
@cpanato

cpanato:
this might impact the internal pipelines
cc @metanerd for review as well

metanerd:
I agree with the sentiment that it would make sense to split the artifacts based on OS and put them into separate folders. The timing needs to be right before a code freeze for this and it will require pipeline code adjustments from our side before merging this. There are more pain points with the Makefile and the packaging process, which I would like to address soon :tm: . I personally would put this PR off until then.

evelikov:
Greetings, are said the internal pipelines working the master branch, or there are additional patches which modify the Makefiles? Reason being, in the former case these changes should not pose any issues, in more detail:
 - no auxiliary scripts are changed - internal tooling can still use them as-is
 - regardless of MR the `$(DIST_ROOT)` is cleaned before and after the packaging is complete - so the internal pipelines cannot rely on any files within `$(DIST_ROOT)`
 - the output/final artefacts are unchanged - filenames, location and contents
 
As @metanerd suggested I'm happy to delay/rebase this for the code freeze, although I cannot find any documentation when that is. Does anyone have a link handy?

metanerd:
Thank you! Since it needs capacity from the release team, I would give you a heads up before this can be merged.

evelikov:
@metanerd with the code-freeze approaching* I've resolved the conflicts and force-pushed the PR.

evelikov:
From a quick look at the failure - it seems to be an issue on https://pr-builds.mattermost.com where the webapp tarball is missing. Perhaps one need explicit ack for that service to generate the file?

metanerd (approved):
@evelikov Thank you for this PR!

metanerd:
PR is good to be merged from my POV despite pipeline failure, as long as there are no new changes.

evelikov:
> PR is good to be merged from my POV despite pipeline failure, as long as there are no new changes.

PR is good as-is. If the team is interested in extra changes, please let me know.

jasonblais:
@agnivade @metanerd what are then the next steps for the PR, is it a QA review prior to merging?

metanerd:
No QA review needed. I am going to merge.

PR #17702, opened 2021-06-01
[MM-29241] Image logic refactor

streamer45:
#### Summary

PR moves most of the image processing logic into a separate `imaging` package under `app`.
Other noticeable changes include:

- Provided implementations for global image decoders/encoders to be used to limit concurrency and peak memory consumption as a result. (to unblock [MM-10166](https://mattermost.atlassian.net/browse/MM-10166))
- Improved memory consumption for transparent PNG uploads. This can be considered a temporary solution until [MM-36111](https://mattermost.atlassian.net/browse/MM-36111) is addressed. I didn't want to commit to such a possibly breaking change so waiting for 6.0 to do it.

Here's a quick test for a sample PNG file with transparency:

imgfill

Annotations indicate the following:

1. Sample image upload (on master).
2. Server restart.
3. Sample image upload (on PR changes).

Now that the filling operation is done in-place we save one full allocation, which for a raw image can be considerable (up to ~196MB for our current resolution limit of 24MP).

#### Benchmark

```
name                                                     old time/op    new time/op    delta
FillImageTransparency/10MPx_fully_transparent_RGBA-8       6.70ms ± 4%    2.77ms ± 4%   -58.64%  (p=0.000 n=10+10)
FillImageTransparency/10MPx_partially_opaque_RGBA-8        6.71ms ± 2%    2.48ms ±10%   -63.13%  (p=0.000 n=10+10)
FillImageTransparency/10MPx_fully_opaque_RGBA-8            6.72ms ± 1%    0.71ms ±17%   -89.38%  (p=0.000 n=8+10)
FillImageTransparency/10MPx_fully_opaque_palette-8         17.4ms ± 3%     0.3ms ± 9%   -98.22%  (p=0.000 n=9+10)
FillImageTransparency/10MPx_fully_transparent_palette-8    19.2ms ± 2%     0.3ms ± 5%   -98.32%  (p=0.000 n=9+10)

name                                                     old alloc/op   new alloc/op   delta
FillImageTransparency/10MPx_fully_transparent_RGBA-8       4.01MB ± 0%    0.00MB ± 8%  -100.00%  (p=0.000 n=9+10)
FillImageTransparency/10MPx_partially_opaque_RGBA-8        4.01MB ± 0%    0.00MB ± 4%  -100.00%  (p=0.000 n=10+10)
FillImageTransparency/10MPx_fully_opaque_RGBA-8            4.01MB ± 0%    0.00MB ± 4%  -100.00%  (p=0.000 n=10+10)
FillImageTransparency/10MPx_fully_opaque_palette-8         4.01MB ± 0%    0.00MB ±42%  -100.00%  (p=0.000 n=10+9)
FillImageTransparency/10MPx_fully_transparent_palette-8    4.01MB ± 0%    0.00MB ±38%  -100.00%  (p=0.000 n=9+10)

name                                                     old allocs/op  new allocs/op  delta
FillImageTransparency/10MPx_fully_transparent_RGBA-8         4.00 ± 0%      1.00 ± 0%   -75.00%  (p=0.000 n=10+10)
FillImageTransparency/10MPx_partially_opaque_RGBA-8          4.00 ± 0%      1.00 ± 0%   -75.00%  (p=0.000 n=10+10)
FillImageTransparency/10MPx_fully_opaque_RGBA-8              4.00 ± 0%      1.00 ± 0%   -75.00%  (p=0.000 n=10+10)
FillImageTransparency/10MPx_fully_opaque_palette-8           4.00 ± 0%      0.00       -100.00%  (p=0.000 n=10+10)
FillImageTransparency/10MPx_fully_transparent_palette-8      4.00 ± 0%      0.00       -100.00%  (p=0.000 n=10+10)
```

#### Ticket

https://mattermost.atlassian.net/browse/MM-29241

#### Release Note

```release-note
Improved memory performance for large image uploads, particularly PNGs with transparency.
```


agnivade:
tkmm

app/imaging/orientation.go
```diff
@@ -0,0 +1,76 @@
...
+	  88          88  888888  888888
+	*/
+	Upright            = 1
+	UprightMirrored    = 2
+	UpsideDown         = 3
+	UpsideDownMirrored = 4
+	RotatedCWMirrored  = 5
+	RotatedCCW         = 6
+	RotatedCCWMirrored = 7
+	RotatedCW          = 8
```
agnivade, line 32: Let's improve it a bit to use auto-increment.
```
const (
 dummy = iota
 Upright
 UprightMirrored
 ...
)
```

app/imaging/decode.go
```diff
@@ -0,0 +1,146 @@
...
+		return nil, "", fmt.Errorf("imaging: failed to decode image: %w", err)
+	}
+
+	return img, format, nil
+}
+
+// DecodeMemBounded works similarly to Decode but also returns a release function that
+// must be called when access to the raw image is not needed anymore.
+// This sets the raw image data pointer to nil in an attempt to help the GC to re-use the underlying data as soon as possible.
+func (d *Decoder) DecodeMemBounded(rd io.Reader) (img image.Image, format string, releaseFunc func(), err error) {
```
agnivade, line 73: I am wondering if instead of a release function, doing an `runtime.GC` makes any difference? Or is the runtime unable to reclaim the memory until you explicitly mark them as nil?

app/server.go
```diff
@@ -245,6 +249,18 @@ func NewServer(options ...Option) (*Server, error) {
...
 	}
 
+	var imgErr error
+	s.imgDecoder, imgErr = imaging.NewDecoder(imaging.DecoderOpts{
+		ConcurrencyLevel: runtime.NumCPU(),
+	})
+	if imgErr != nil {
+		return nil, errors.Wrap(imgErr, "failed to create image decoder")
+	}
+	s.imgEncoder, imgErr = imaging.NewEncoder(imaging.EncoderOpts{})
```
agnivade, line 259: Just curious, are we not specifying a concurrency limit for encoding until we see a need for it?

app/imaging/utils_bench_test.go
```diff
@@ -0,0 +1,103 @@
...
+		"10MPx fully opaque palette",
+		fullyOpaquePaletteGen(1000, 1000),
+	},
+	{
+		"10MPx fully transparent palette",
+		fullyTransparentPaletteGen(1000, 1000),
+	},
+}
+
+func BenchmarkFillImageTransparency(b *testing.B) {
```
agnivade, line 79: Would love to see the results of this in the PR message.

app/file.go
```diff
@@ -1070,113 +1039,72 @@ func (a *App) HandleImages(previewPathList []string, thumbnailPathList []string,
...
+		release()
 	}
-	wg.Wait()
 }
 
-func prepareImage(fileData []byte) (image.Image, int, int) {
+func prepareImage(imgDecoder *imaging.Decoder, fileData []byte) (image.Image, func(), error) {
 	// Decode image bytes into Image object
-	img, imgType, err := image.Decode(bytes.NewReader(fileData))
+	img, imgType, release, err := imgDecoder.DecodeMemBounded(bytes.NewReader(fileData))
```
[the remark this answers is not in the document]
streamer45: Yes, there's definitely some room for improvement there. I'll have a look :+1:

app/server.go
streamer45: Good point. I think we can probably default to NumCPU as well. The way I wrote the PR I just forgot to add it.

app/imaging/utils_bench_test.go
streamer45: Sure thing :+1:

app/imaging/decode.go
streamer45: I am not sure I wanted to go as far as forcing the GC call which may have other performance implications in itself and doesn't look like something recommended. The idea here was just to do our best to help with re-using memory by setting its reference to nil. I believe GC behaviour is overall non-deterministic so that's not a guarantee of memory release but it will help since the caller might be doing a lot more before the image reference goes out of scope.

model/file_info.go
```diff
@@ -162,19 +156,6 @@ func NewInfo(name string) *FileInfo {
 	return info
 }
 
-func GenerateMiniPreviewImage(img image.Image) *[]byte {
```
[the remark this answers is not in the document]
streamer45: Honestly if it's not used by either plugins or client4.go I think it's fair game. Also a bit of common sense tells me that method wasn't getting used elsewhere.

app/imaging/decode.go
streamer45: Agreed, ideally we'd like something like a shared buffer pool to have more control on this front but we need some help from the API side to achieve that. Might be an idea for a possible upstream contribution who knows :p

agnivade (approved):
Great work!

app/file.go
```diff
@@ -1070,113 +1039,75 @@ func (a *App) HandleImages(previewPathList []string, thumbnailPathList []string,
...
+		}(img, previewPathList[i])
+
+		wg.Wait()
+		release()
 	}
-	wg.Wait()
 }
 
-func prepareImage(fileData []byte) (image.Image, int, int) {
+func prepareImage(imgDecoder *imaging.Decoder, imgData io.ReadSeeker) (img image.Image, release func(), err error) {
```
isacikgoz, line 1063: I have a question here, would it make sense to convert this function to a `imaging.Decoder` method? Seems like we are doing an imaging operation here and would it make sense to give this responsibility to `imaging` package?

app/imaging/decode.go
```diff
@@ -0,0 +1,146 @@
...
+)
+
+// DecoderOpts holds configuration options for an image decoder.
+type DecoderOpts struct {
+	// The level of concurrency for the decoder. This defines a limit on the
+	// number of concurrently running encoding goroutines.
+	ConcurrencyLevel int
+}
+
+func (o *DecoderOpts) Validate() error {
```
isacikgoz, line 28: I wonder if this should remain unexported to reduce API surface fir this package?

app/imaging/decode.go
```diff
@@ -0,0 +1,146 @@
...
+	return cfg.Width, cfg.Height, err
+}
+
+// This is only needed to try and simplify GC work.
+func releaseImageData(img image.Image) {
+	switch raw := img.(type) {
+	case *image.Alpha:
+		raw.Pix = nil
+	case *image.Alpha16:
+		raw.Pix = nil
```
isacikgoz, line 128: Too bad that `fallthrough` statement doesn't work on type switches :p

app/imaging/encode.go
```diff
@@ -0,0 +1,86 @@
...
+)
+
+// EncoderOpts holds configuration options for an image encoder.
+type EncoderOpts struct {
+	// The level of concurrency for the encoder. This defines a limit on the
+	// number of concurrently running encoding goroutines.
+	ConcurrencyLevel int
+}
+
+func (o *EncoderOpts) Validate() error {
```
isacikgoz, line 23: Same concern for this one.

app/imaging/preview.go
```diff
@@ -0,0 +1,66 @@
...
+	"bytes"
+	"fmt"
+	"image"
+	"image/jpeg"
+
+	"github.com/disintegration/imaging"
+)
+
+// GenPreview generates the preview for the given image.
+func GenPreview(img image.Image, width int) image.Image {
```
isacikgoz, line 16: Not a fan of shortening function names for exported functions but 0/5

app/imaging/utils.go
```diff
@@ -0,0 +1,149 @@
...
+)
+
+type rawImg interface {
+	Set(x, y int, c color.Color)
+	Opaque() bool
+}
+
+func isFullyTransparent(c color.Color) bool {
+	var a uint32
+	switch c.(type) {
```
isacikgoz, line 20: Why this type switch is needed?

isacikgoz (approved):
Amazing piece of work! 🎉 I have some small things below.

app/file.go
streamer45: I've thought about that but eventually felt this was a bit too app specific to fit in the package which at some point might even go under `shared` as a standalone service/library.

app/imaging/utils.go
streamer45: As is not at much. The idea there was to possibly optimize further by knowing the type of color and extract only what's needed (the alpha value). Now it's just defaulting to the sub-optimal case.

app/imaging/preview.go
streamer45: It just saved some typing. I am okay either way as well.

app/imaging/utils.go
streamer45: Simplified and added a TODO.

app/file.go
isacikgoz: I see. Thanks for the explanation 👍

streamer45:
@saturninoabril 

For this I am thinking a smoke-test specifically around image uploads (profile picture, post attachments, brand image) and making sure thumbnail/previews work as expected.

saturninoabril (approved):
Tested and passed - https://automation-dashboard.vercel.app/cycles/d0e2c5c7-a3cd-4e3d-bfac-1cd8b83158c1

PR #17718, opened 2021-06-04
Fixup: app_test: remove unused search engine initialization

jayaddison-collabora:
#### Summary
While running `mattermost-server` unit tests locally by using `make test-server-quick`, I'd run into the following Go panic:

```
panic(0x21ea6e0, 0x3987f00)
    /usr/lib/go-1.15/src/runtime/panic.go:969 +0x1b9
github.com/mattermost/mattermost-server/v5/services/searchengine.(*Broker).RegisterBleveEngine(...)
    /home/jka/Documents/mattermost-server/services/searchengine/searchengine.go:23
github.com/mattermost/mattermost-server/v5/app.TestUnitUpdateConfig(0xc0026c9500)
    /home/jka/Documents/mattermost-server/app/app_test.go:46 +0x120
```

(also reported using the Mattermost community developers channel [here](https://community.mattermost.com/core/pl/x1u7rhckx3rt8j4cc8ph9sbqow))

The root cause appears to be that the `app_test` tries to register a search engine against a [nil/empty `SearchEngine` member variable](https://github.com/mattermost/mattermost-server/blob/09f99e242699c5aaf380f1076d3aa043a88853b0/app/app_test.go#L46), leading to a nil pointer dereference in [this function](https://github.com/mattermost/mattermost-server/blob/09f99e242699c5aaf380f1076d3aa043a88853b0/services/searchengine/searchengine.go#L22-L24).

The fix suggested here essentially copies a pattern found elsewhere in the codebase, in the [`app/server` module here](https://github.com/mattermost/mattermost-server/blob/09f99e242699c5aaf380f1076d3aa043a88853b0/app/server.go#L298-L304).

#### Ticket Link
N/A

#### Release Note
```release-note
NONE
```


app/app_test.go
```diff
@@ -41,9 +41,11 @@ func init() {
 func TestUnitUpdateConfig(t *testing.T) {
 	th := SetupWithStoreMock(t)
 	defer th.TearDown()
+	searchEngine := searchengine.NewBroker(th.App.Config(), th.App.Srv().Jobs)
```
mgdelacroix, line 44: You need to import `searchengine` above

mgdelacroix (changes requested):
Thanks for the PR @jayaddison-collabora! :rocket: The approach looks valid, although I wonder if we should not disable this test when the `short` flag is enabled, as the searchengine is not being initialised because of the store itself not being initialised

//cc @jespino

jayaddison-collabora:
@mgdelacroix @jespino FWIW - during [configuration of custom GitLab CI](https://gitlab.collabora.com/sysadmin/mattermost/mattermost-server/-/merge_requests/4), I've found that -- on the `release-5.35` branch, albeit currently a few (unrelated) commits behind at 1a0f7331fd29abf9f5b2950cd9c3927b679f5712, the following tests had issues during short tests:

- [`migrate_test`](https://github.com/mattermost/mattermost-server/blob/1a0f7331fd29abf9f5b2950cd9c3927b679f5712/config/migrate_test.go#L21) as noted here by @mgdelacroix 
- [`version_test`](https://github.com/mattermost/mattermost-server/blob/1a0f7331fd29abf9f5b2950cd9c3927b679f5712/model/version_test.go)
- Tests within the `app` module (worked around by using [this non-ideal commit](https://gitlab.collabora.com/sysadmin/mattermost/mattermost-server/-/merge_requests/4/diffs?commit_id=13b7725205d27c3cf2c6bb6b56bcd1d09f942c1a))

These issues only seem to appear within a GitLab CI runner environment (and are reproducible with `gitlab-runner` which can be handy for local development testing).

There are quite a few moving parts here and I don't feel like I have a full understanding of how they interact, but I'll try to chip away at one or two small details and provide patches where they seem correct and safe.

jespino (changes requested):
The real problem here is, "why the searchEngine is even needed here?" actually is not, if you revmoe the search engine initialization entirely, the tests passes perfectly fine.

jespino:
@jayaddison-collabora about the migration test, it isn't using one of the autodetected setup functions for not-short test runs, instead, we have to manually exclude that test using:
```go
if testing.Short() {
       t.SkipNow()
}
```

jespino (approved):
LGTM

mgdelacroix (approved):
LGTM, thanks @jayaddison-collabora!!! :+1:

jayaddison-collabora:
(CircleCI seemed not to initialize for f526823; merged latest changes to bump the build)

jayaddison-collabora:
@mgdelacroix Could you try re-applying the `EETests` label when you have a moment?  I think/hope that should complete the test suite here.

mgdelacroix:
I've just reenabled EETests and merged `master` in to try and relaunch the tests, hope the report makes it back this time. Apologies for the waiting @jayaddison-collabora

jayaddison-collabora:
@mgdelacroix No problem at all, and thank you!

PR #17719, opened 2021-06-04
[MM-36198] - Activating a user in System Console does not respect cloud user limits

AGMETEOR:
#### Summary
For cloud instances, when we're at the cloud limit, we should not allow activation of users from the system console

#### Ticket Link
[MM-36198](https://mattermost.atlassian.net/browse/MM-36198?atlOrigin=eyJpIjoiMDMzMDA3ZDM1MGQ2NGI3ZDgxOWNiZDk0NTk4ZWM3OGEiLCJwIjoiaiJ9)

#### Release Note
```release-note
NONE
```

### DIFF
--- a/app/cloud.go
+++ b/app/cloud.go
@@ -79,6 +79,52 @@ func (a *App) SendAdminUpgradeRequestEmail(username string, subscription *model.
 	return nil
 }
 
+func (a *App) GetSubscriptionStats() (*model.SubscriptionStats, *model.AppError) {
+	if a.Srv().License() == nil || !*a.Srv().License().Features.Cloud {
+		return nil, model.NewAppError("app.GetSubscriptionStats", "api.cloud.license_error", nil, "", http.StatusInternalServerError)
+	}
+
+	subscription, appErr := a.Cloud().GetSubscription("")
+	if appErr != nil {
+		return nil, model.NewAppError("app.GetSubscriptionStats", "api.cloud.request_error", nil, appErr.Error(), http.StatusInternalServerError)
+	}
+
+	count, err := a.Srv().Store.User().Count(model.UserCountOptions{})
+	if err != nil {
+		return nil, model.NewAppError("app.GetSubscriptionStats", "app.user.get_total_users_count.app_error", nil, err.Error(), http.StatusInternalServerError)
+	}
+	cloudUserLimit := *a.Config().ExperimentalSettings.CloudUserLimit
+
+	s := cloudUserLimit - count
+
+	return &model.SubscriptionStats{
+		RemainingSeats: int(s),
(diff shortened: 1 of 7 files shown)


AGMETEOR:
@stevemudie it's possible to test now with the instance above but we need to also verify that we receive the "at limit warning email". I think a post merge with an instance set up from the test portal is better. For an instance set up in the test portal all you have to do is set `cws-is-paid-tier` metadata on the subscription to false

stevemudie (approved):
Worked as expected!

Thanks for getting that set up for me @AGMETEOR 

My only concern (it's a small one) is the text reads "Unable to activate more users as the cloud account is over capacity." when - at that point - you're not technically "over" capacity, you're directly at it with 10 users.

Not sure if that's a great concern at this point.

app/cloud.go
```diff
@@ -79,6 +79,52 @@ func (a *App) SendAdminUpgradeRequestEmail(username string, subscription *model.
...
+		return false, err
+	}
+
+	if (stats.IsPaidTier == "false") && (stats.RemainingSeats < 1) {
+		return true, nil
+	}
+
+	if (stats.IsPaidTier == "false") && (stats.RemainingSeats > 1) {
+		return false, nil
+	}
```
marianunez, line 123: We could simplify this checks and return early if `stats.isPaidTier=="true"`
```suggestion
	if (stats.IsPaidTier == "true"){
		return false, nil
	}
	if (stats.RemainingSeats < 1) {
		return true, nil
	}

	if (stats.RemainingSeats > 1) {
		return false, nil
	}
```

marianunez (approved):
LGTM, could we add a test to cover these new cases here? https://github.com/mattermost/mattermost-server/blob/45550d4d2ab3a18752d2aa09a8e11f77848277ba/api4/user_test.go#L2106

marianunez:
> My only concern (it's a small one) is the text reads "Unable to activate more users as the cloud account is over capacity." when - at that point - you're not technically "over" capacity, you're directly at it with 10 users.
> 
> Not sure if that's a great concern at this point.

@stevemudie can we also do regression testing on activating/deactivating users in regular self-managed servers? I think we can change the tag here to the other `Setup Test Server`

AGMETEOR:
> LGTM, could we add a test to cover these new cases here?
> 
> https://github.com/mattermost/mattermost-server/blob/45550d4d2ab3a18752d2aa09a8e11f77848277ba/api4/user_test.go#L2106

@marianunez added a test case

nickmisasi (approved):

marianunez:
> @marianunez added a test case

Looks great, thanks @AGMETEOR! After merge we need to cherry-pick to `cloud` given that Fast-Forward was completed this morning.

PR #17722, opened 2021-06-04
[MM-36217] Improve column default utilities

streamer45:
#### Summary

While reviewing some migrations I've figured something odd was happening with column defaults. 
Passing an empty string didn't result in an empty string default but would instead cause the default to get dropped.

PR fixes this issue and also splits the behaviour into two separate methods (alter and remove) to avoid more unwanted surprises. I've also fixed the naming which was rather ambiguous (the exists refers to the column and not the default).

Pre-existing migration code was made semantically equivalent.

#### Ticket

https://mattermost.atlassian.net/browse/MM-36217

#### Release Note

```release-note
NONE
```

### DIFF
--- a/store/sqlstore/store.go
+++ b/store/sqlstore/store.go
@@ -909,7 +909,23 @@ func (ss *SqlStore) AlterColumnTypeIfExists(tableName string, columnName string,
 	return true
 }
 
-func (ss *SqlStore) AlterColumnDefaultIfExists(tableName string, columnName string, mySqlColDefault *string, postgresColDefault *string) bool {
+func (ss *SqlStore) RemoveDefaultIfColumnExists(tableName, columnName string) bool {
+	if !ss.DoesColumnExist(tableName, columnName) {
+		return false
+	}
+
+	_, err := ss.GetMaster().ExecNoTimeout("ALTER TABLE " + tableName + " ALTER COLUMN " + columnName + " DROP DEFAULT")
+	if err != nil {
+		mlog.Critical("Failed to drop column default", mlog.String("table", tableName), mlog.String("column", columnName), mlog.Err(err))
+		time.Sleep(time.Second)
+		os.Exit(ExitGenericFailure)
+		return false
+	}
+
+	return true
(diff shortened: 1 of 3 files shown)


store/sqlstore/store.go
```diff
@@ -909,7 +909,23 @@ func (ss *SqlStore) AlterColumnTypeIfExists(tableName string, columnName string,
...
-func (ss *SqlStore) AlterColumnDefaultIfExists(tableName string, columnName string, mySqlColDefault *string, postgresColDefault *string) bool {
+func (ss *SqlStore) RemoveDefaultIfColumnExists(tableName, columnName string) bool {
+	if !ss.DoesColumnExist(tableName, columnName) {
+		return false
+	}
+
+	_, err := ss.GetMaster().ExecNoTimeout("ALTER TABLE " + tableName + " ALTER COLUMN " + columnName + " DROP DEFAULT")
+	if err != nil {
+		mlog.Critical("Failed to drop column default", mlog.String("table", tableName), mlog.String("column", columnName), mlog.Err(err))
+		time.Sleep(time.Second)
```
agnivade, line 920: I've always wondered why do we sleep for a second here? :P

agnivade (approved):
Nasty one indeed.

store/sqlstore/store.go
streamer45: Likewise :) And what's worse I keep copying it for some inner sense of consistency but fully knowing it's most probably totally unnecessary :/

hmhealey: I wouldn't be surprised if it's to allow the log file to be flushed or something like that. The server didn't used to shut down as gracefully before as it does now.

hmhealey (approved):

hmhealey:
Looking back at https://github.com/mattermost/mattermost-server/pull/16030/commits/71b3b3ea0ccaf698354116002bfedaae40ba1a64#r512756520, Agniva and I were also confused by this function's behaviour last year 😛

PR #17752, opened 2021-06-09
[MM-36268] Fix replication lag error on post reply

streamer45:
#### Summary

During recent load-tests, several errors in response to creating posts were spotted, specifically on replies.

After a deeper investigation it became clear that replication lag could be involved as `ThreadStore.Get()` was failing with a "not found" error. A thread entry gets inserted only after the first reply to a given post comes in so it's possible, especially under some load, that the thread doesn't exist on all replicas at the time `ThreadStore.MaintainMembership()` gets called.

PR fixes this using the common approach of passing a context and querying the writer for such case.

#### Ticket

https://mattermost.atlassian.net/browse/MM-36268

```release-note
NONE
```


store/sqlstore/thread_store.go
```diff
@@ -517,24 +517,24 @@ func (s *SqlThreadStore) DeleteMembershipForUser(userId string, postId string) e
 	return nil
 }
 
-func (s *SqlThreadStore) MaintainMembership(userId, postId string, following, incrementMentions, updateFollowing, updateViewedTimestamp, updateParticipants bool) (*model.ThreadMembership, error) {
+func (s *SqlThreadStore) MaintainMembership(ctx context.Context, userId, postId string, opts store.ThreadMembershipOpts) (*model.ThreadMembership, error) {
```
agnivade, line 520: Love the refactoring work :heart:

agnivade (approved):

store/sqlstore/post_store.go
```diff
@@ -2252,7 +2252,7 @@ func (s *SqlPostStore) cleanupThreads(postId, rootId string, permanent bool) err
 		return nil
 	}
 	if rootId != "" {
-		thread, err := s.Thread().Get(rootId)
+		thread, err := s.Thread().Get(WithMaster(context.Background()), rootId)
 		if err != nil {
 			var nfErr *store.ErrNotFound
 			if !errors.As(err, &nfErr) {
```
agnivade, line 2258: GH won't let me comment on the lines below but this is about the Get and Update queries combined. It seems like we are fetching the thread object by its ID, setting replyCount to `-1` and then writing it back to the DB again. Any reason we cannot do a simple `UPDATE threads set replyCount = -1 where id = <>`?

agnivade (commented):
Sorry, taking a slightly deeper look now. Can you point to me the read-after-write flow here? You mention that the thread is being created only when we reply to a post. I am finding it a bit hard to find out at which point exactly is the thread being created.

streamer45:
> Sorry, taking a slightly deeper look now. Can you point to me the read-after-write flow here? You mention that the thread is being created only when we reply to a post. I am finding it a bit hard to find out at which point exactly is the thread being created.

As far as I could see the thread is created in `PostStore.updateThreadsFromPosts()` called by `PostStore.SaveMultiple`:

https://github.com/mattermost/mattermost-server/blob/364ea5ed63e3ccb8c953a849b4c52773bc9ffa08/store/sqlstore/post_store.go#L2311-L2317

https://github.com/mattermost/mattermost-server/blob/364ea5ed63e3ccb8c953a849b4c52773bc9ffa08/store/sqlstore/post_store.go#L193-L195

Then the read happens after the post (and thread) have been created, during the notification logic (`app.SendNotifications()`).

store/sqlstore/post_store.go
streamer45: I believe we need to decrement the count by 1 there. Either way you are correct, it can probably be done with a single query:

```sql
UPDATE Threads 
SET ReplyCount = ReplyCount - 1
WHERE PostId = ?
AND ReplyCount > 0
```

agnivade:
Thanks! I was running out of ways to grep :P

Okay, so this looks all looks like it's being done in a single unit of work. The thread gets created in poststore.savemultiple, but then it is being fetched again from app.SendNotifications.

Wondering how hard would it be to return the created thread from `poststore.SaveMultiple` and pass it all the way through to `app.SendNotifications`? I see that the thread.Get call is conditional on `updateParticipants` being `true`. We can even take that logic out of that method to directly call the `Update` method if we have the thread object, else just call from replica. 0/5.

Any thoughts?

streamer45:
> Okay, so this looks all looks like it's being done in a single unit of work. The thread gets created in poststore.savemultiple, but then it is being fetched again from app.SendNotifications.
> 
> Wondering how hard would it be to return the created thread from `poststore.SaveMultiple` and pass it all the way through to `app.SendNotifications`? I see that the thread.Get call is conditional on `updateParticipants` being `true`. We can even take that logic out of that method to directly call the `Update` method if we have the thread object, else just call from replica. 0/5.
> 
> Any thoughts?

I am thinking that should work. Not super in love with returning threads from a method saving posts in order to save a db read but I'll take a look nonetheless :)

agnivade:
Fully understand that. I think in extreme cases earlier, like [here](https://github.com/mattermost/mattermost-server/pull/17176) where we create roles while creating schemes, it becomes very weird to return a slice of roles when saving a scheme. In this case, a thread and a post seems like related entities.

I think the "create post" flow is so complicated that it is super hard to do everything in a single transaction, although ideally it should be. Keeping the code complexity aside, the next best thing would be to pass the created objects.

streamer45:
> Fully understand that. I think in extreme cases earlier, like [here](https://github.com/mattermost/mattermost-server/pull/17176) where we create roles while creating schemes, it becomes very weird to return a slice of roles when saving a scheme. In this case, a thread and a post seems like related entities.
> 
> I think the "create post" flow is so complicated that it is super hard to do everything in a single transaction, although ideally it should be. Keeping the code complexity aside, the next best thing would be to pass the created objects.

True, but with the proposed approach we are keeping the door open (even more than now) for possible inconsistencies as well. The returned thread could get stale if multiple replies come in at the same time and the final update could end-up messing up the overall state. A better approach would be to at least run `MaintainMembership()` as a single transaction.

agnivade:
That is actually a good point, but I don't believe theoretically it increases the chances any more than now (multiple replies could come in between `s.Get` and `s.Update` as well).  Fully agree on making `MaintainMembership` to be a single transaction. :+1: 

I think we have to restart the discussion of executing long running transactions again at some point. :see_no_evil: Executing against master buys us some time, but is not a long-term solution. Either we reshuffle the code to do the thread creation+updation in a single place, or do it in a super-long transaction. Anything else has tradeoffs.

This part about transactions is something I missed during my replica lag investigation. We were simply doing a Create() followed by Get() + Update() in separate queries. So the approach of Create() + Update() didn't seem to make any difference consistency wise. But I should have brought this up at that time itself.

I like how this entire discussion exposed the deep underlying issues we still have. :)

app/channel.go
```diff
@@ -2408,8 +2408,15 @@ func (a *App) MarkChannelAsUnreadFromPost(postID string, userID string, collapse
...
-			threadData, _ := a.Srv().Store.Thread().Get(threadId)
+			opts := store.ThreadMembershipOpts{
+				Following:             true,
+				IncrementMentions:     true,
+				UpdateFollowing:       true,
+				UpdateViewedTimestamp: true,
+				UpdateParticipants:    false,
+			}
+			threadMembership, _ = a.Srv().Store.Thread().MaintainMembership(context.Background(), user.Id, threadId, opts)
+			threadData, _ := a.Srv().Store.Thread().Get(context.Background(), threadId)
```
mkraft, line 2419: What was the reason these errors are assigned the blank identifier (I realize it was there before your change)?

store/store.go
```diff
@@ -296,7 +296,7 @@ type ThreadStore interface {
 	GetMembershipsForUser(userId, teamID string) ([]*model.ThreadMembership, error)
 	GetMembershipForUser(userId, postID string) (*model.ThreadMembership, error)
 	DeleteMembershipForUser(userId, postID string) error
-	MaintainMembership(userID, postID string, following, incrementMentions, updateFollowing, updateViewedTimestamp, updateParticipants bool) (*model.ThreadMembership, error)
+	MaintainMembership(ctx context.Context, userID, postID string, opts ThreadMembershipOpts) (*model.ThreadMembership, error)
```
mkraft, line 299: It would be great to have a comment on the interface or implementation for `MaintainMembership` to explain it's purpose (I know it's pre-existing).

streamer45: I remember asking to add a comment inside because the logic was getting a bit complex. I'll try to see if we can expand a bit on that although the team owning this feature may have more context.

app/channel.go
streamer45: I can't really answer that and I believe the author is no longer working with us so I can only speculate. To me it looks like we are trying to avoid handling the errors and simply check if they return non-nil to proceed.

streamer45: I believe the author is not working with us anymore so I can only speculate. It looks to me like we are just avoiding error handling and only proceed if the returned object are not `nil`.

store/sqlstore/post_store.go
agnivade: lol yeah .. that would have been a funny bug

streamer45:
@agnivade 

I went with the transaction approach. Let me know what you think.

agnivade (approved):
I like it :+1:

store/sqlstore/post_store.go
streamer45: Created [MM-36366](https://mattermost.atlassian.net/browse/MM-36366)

mkraft (approved):

streamer45:
@furqanmlk 

For this one it would be good to do a general smoke-test around the new collapsed thread feature, specifically on thread auto follow, mostly making sure the expected functionality is unaltered by the changes.

furqanmlk (approved):
New collapsed thread feature has been tested, specifically following scenarios,
All Tests Passed ✅ 
1. If the original parent post was created by you, the thread is automatically followed ✅ 

2. If you reply to a thread, it is automatically followed ✅ 

3. If you are mentioned in a thread, that thread is automatically followed ✅ 

4. A single parent message is not considered a ‘thread' and is not automatically ‘followed’ until the thread has replies.✅ 

5. If you Unfollow a thread, but then are mentioned, the thread is now automatically followed again ✅ 

6. Marking as unread on a post within a thread will automatically follow that thread if it’s not already followed ✅ 

7. Parent posts with no replies that have mentions are automatically followed and display as ```Following``` ✅ 

8. Marking as unread on a post within a thread will automatically follow that thread if it’s not already followed ✅

PR #17754, opened 2021-06-09
MM-35396 Refactor GetThreadForUser store func to take membership as argument to prevent replica lag issues and reduce joins in query

jwilander:
#### Summary
Refactor GetThreadForUser store func to take membership as argument to prevent replica lag issues and reduce joins in query.

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-35396

#### Release Note

```release-note
NONE
```


esethna:
Lol, it was a one line fix 🤣 Nice work Joram!

agnivade:
Thanks Joram!

I want to dig a bit deeper to really understand the read-after-write flow here. Can you point me to the code flow that creates the thread, and then how does query the replica instance to encounter this bug?

Typically, in a read-after-write situation, we pass the created object itself down the stack, rather than querying for the object again. (More info here: https://community-daily.mattermost.com/core/pl/nphtmkowcjd8ic76tbqtapx6nc)

And in cases where that's not possible, we pass a context to the store method to avoid querying master unconditionally. (See (https://github.com/mattermost/mattermost-server/pull/17176, https://github.com/mattermost/mattermost-server/pull/16170 for  examples)

jwilander:
@agnivade, my bad I meant to include that info in the summary but got distracted. It comes from here:

https://github.com/mattermost/mattermost-server/blob/master/app/user.go#L2393

What's happening in the previous lines is we mark the ThreadMembership for a user as read by updating the `LastViewedAt` time to the timestamp. We also count the number of mentions the user has in thread too before going and updating the ThreadMembership via `UpdateMembership` and `MarkAsRead`.

Immediately after doing that we call `GetThreadForUser` which ends up doing a SQL query to pull a bunch of data. The data that's causing the problem here is the `UnreadReplies` field. The part of the query that pulls that data relies on `ThreadMembership.LastViewedAt`. If there's replica lag, though, that might be out of date and contain the old timestamp, resulting in `UnreadReplies` having old data.

agnivade:
This is great!

Ok so the issue seems to be a classic read-after-write case:
1. We are modifying `ThreadMemberships.LastViewedAt` for a single row
2. Reading `ThreadMemberships` again assuming that it is latest.

Looking deeper into `GetThreadForUser`, I see that although we are joining with some other tables, the filter conditions essentially remain the same as the earlier update query, and it just returns a single row.

This leads me to believe that we can just return the updated membership object from `a.Srv().Store.Thread().MarkAsRead` and pass that to `a.GetThreadForUser`. 

For example, for the unreadReplies query here:

```
SELECT COUNT(Posts.Id) From Posts Where Posts.RootId=ThreadMemberships.PostId AND Posts.CreateAt >= ThreadMemberships.LastViewed AND Posts.DeleteAt=0 AND ...
```
can just be 

```
SELECT COUNT(Posts.Id) From Posts Where Posts.CreateAt >= <LastViewed> AND Posts.DeleteAt=0 AND ...
```

Same for the main query also. We are just re-fetching `ThreadMemberships.LastViewed as LastViewedAt, ThreadMemberships.UnreadMentions as UnreadMentions, ThreadMemberships.Following` which we _already_ have with us just in the previous query.

I don't think we need to join with the `ThreadMemberships` table at all.

jwilander:
Awesome, thanks @agnivade! That sounds like a fairly elegant solution. I'll add a commit that tries it out.

store/sqlstore/thread_store.go
```diff
@@ -334,38 +338,46 @@ func (s *SqlThreadStore) GetThreadForUser(userId, teamId, threadId string, exten
...
 		Column(sq.Alias(sq.Expr(unreadRepliesQuery), "UnreadReplies")).
 		LeftJoin("Posts ON Posts.Id = Threads.PostId").
 		LeftJoin("Channels ON Posts.ChannelId = Channels.Id").
-		LeftJoin("ThreadMemberships ON ThreadMemberships.PostId = Threads.PostId").
 		Where(fetchConditions).ToSql()
-	err := s.GetReplica().SelectOne(&thread, query, args...)
 
+	args := append(unreadRepliesArgs, threadArgs...)
+
+	err := s.GetMaster().SelectOne(&thread, query, args...)
```
agnivade, line 366: Do we still need to query master here? I thought that was the fix .. :thinking:

model/feature_flags.go
```diff
@@ -41,7 +41,7 @@ func (f *FeatureFlags) SetDefaults() {
 	f.TestFeature = "off"
 	f.TestBoolFeature = false
 	f.CloudDelinquentEmailJobsEnabled = false
-	f.CollapsedThreads = false
+	f.CollapsedThreads = true
```
agnivade, line 44: Unrelated?

app/channel.go
```diff
@@ -2525,7 +2525,7 @@ func (a *App) markChannelAsUnreadFromPostCRTUnsupported(postID string, userID st
 	if nErr != nil {
```
agnivade, line 2525: Ooo .. we automatically get some added benefits here, as that's how the query flow seems to happen. Nice!

Just a small thing - I'd suggest to use the object returned on L2524, the one ignored by "_". Seems more natural as that's the latest updated one returned from the store.

agnivade (commented):
Looks like a great improvement overall. Now we are passing objects rather than IDs and re-querying the same objects again. Much more well-defined flows via DI and less DB load.

I actually wanted `a.Srv().Store.Thread().MarkAsRead` to return the `membership` object as we discussed. But looking at your solution now, I like this too :+1:

model/feature_flags.go
jwilander: Whoops, yes unrelated

store/sqlstore/thread_store.go
jwilander: Ah yes, my mistake forgot to revert that part.

jwilander:
>I actually wanted a.Srv().Store.Thread().MarkAsRead to return the membership object as we discussed. But looking at your solution now, I like this too 👍

Yeah, I went a slightly different route since `GetUserFromThread` was called in a number of different places separate from `MarkAsRead` and in the place causing the issue we already had the ThreadMembership.

agnivade (approved):
Good stuff :+1:

store/sqlstore/thread_store.go
```diff
@@ -334,38 +338,46 @@ func (s *SqlThreadStore) GetThreadForUser(userId, teamId, threadId string, exten
...
+
+	err := s.GetReplica().SelectOne(&thread, query, args...)
 	if err != nil {
 		return nil, err
 	}
 
-	if !thread.Following {
-		return nil, nil // in case the thread is not followed anymore - return nil error to be interpreted as 404
-	}
+	thread.Following = threadMembership.Following
```
ashishbhate, line 371: We can drop `Following` from `JoinedThread` struct above and remove this assignment. The conditional that used it was moved to the top of the function.

ashishbhate (approved):
LGTM!

jwilander:
@jgilliam17 this can be tested on https://joram01.test.mattermost.cloud which has a replica DB set up

jgilliam17 (approved):
Thanks @jwilander 
Tested, looks good to merge.
- Verified that opening an unread thread marks it as read and clears the unread dot.

PR #17755, opened 2021-06-09
MM-35699: CRT off should behave same as CRT unsupported

ashishbhate:
#### Summary

- CRT for a user that has turned it off should behave the same as though the client does not support CRT

#### Ticket Link

- https://mattermost.atlassian.net/browse/MM-35699

#### Release Note

```release-note
NONE
```


jwilander (approved):
LGTM

isacikgoz (approved):
LGTM 👍

esethna:
@ogi-m can we please prioritize the PR reviews for any server related fixes for CRT? I think the server changes are at the biggest risk given our tight timeline.

esethna:
@ogi-m or @jgilliam17 please help with priority reviews so we can get this on daily for the testing session tomorrow

ogi-m (approved):
One issue I'm seeing is this:
1. With CRT off, mark a reply message as unread - channel is bolded
2. Switch to a different channel and enable CRT- channel is unbolded, Threads area is bolded instead
3. Open the channel where the thread is - note the blue dot in the thread footer indicating unread replies
4. Switch to a different channel again and disable CRT
 
**Observed**: The channel with the unread reply remains unbolded, and opening the channel doesn't show the new messages line above the reply that was marked as unread
**Expected**: The channel with the unread reply should be bolded again and when opened should show the new messages line above the post.

Otherwise everything looks good. @ashishbhate I don't know if it's related to this PR but any case we can merge and open a separate ticket for it.

ashishbhate:
>Otherwise everything looks good. @ashishbhate I don't know if it's related to this PR but any case we can merge and open a separate ticket for it.

I've opened a ticket to investigate: https://mattermost.atlassian.net/browse/MM-36533

PR #17762, opened 2021-06-10
Give local dev HA cluster a name

ashishbhate:
#### Summary

- The servers in a HA cluster won't communicate with each other unless they have the same name. An empty name doesn't work, so we set a generic name. Even with an incorrect configuration the cluster appeared to work because the nginx proxy was distributing http API requests to each individual server, but cluster messaging wouldn't work, because of which websocket events weren't firing correctly.
 
 
#### Ticket Link

- Discovered while working on https://mattermost.atlassian.net/browse/MM-35396
#### Release Note

```release-note
NONE
```

### DIFF
--- a/docker-compose.yaml
+++ b/docker-compose.yaml
@@ -107,6 +107,7 @@ services:
       - "MM_NO_DOCKER=true"
       - "RUN_SERVER_IN_BACKGROUND=false"
       - "MM_CLUSTERSETTINGS_ENABLE=true"
+      - "MM_CLUSTERSETTINGS_CLUSTERNAME=mm_dev_cluster"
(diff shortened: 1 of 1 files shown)


agnivade (approved):
Good stuff. I don't use the Makefile much so I never knew.

ashishbhate:
The makefile HA server doesn't set `DataSourceReplicas` either, so its just one DB used by multiple servers. So replica lag issues don't surface. 
I'll try to fix this tomorrow during OSF, if I don't get to it I'll open a jira ticket.

jwilander (approved):
Nice fix 👍

PR #17764, opened 2021-06-11
MM-35030: Hoist GetDBVersion before creating store

agnivade:
The creation of the store included running the migrations
which means that SQL statements would be executed before
we could make the check for DB version.

We make this a package-level function to allow checking
the DB version even before creating a store.

https://mattermost.atlassian.net/browse/MM-35030

```release-note
NONE
```

### DIFF
--- a/app/server.go
+++ b/app/server.go
@@ -20,7 +20,6 @@ import (
 	"os/exec"
 	"path"
 	"runtime"
-	"strconv"
 	"strings"
 	"sync"
 	"sync/atomic"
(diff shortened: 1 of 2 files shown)


isacikgoz (approved):
LGTM 👍

streamer45 (commented):
PR looks good. 

I'd just like to make a couple of related considerations:

- It looks a bit weird to open and close a connection just to check the version and do that again (in most cases) on the next call. I am thinking the root issue here is that `sqlStore.New()` does way more than what a normal constructor should be doing. Maybe that's something that we could address at some point but it needs some thinking.
- Another option, that wouldn't require much of a change, would be to move this logic inside `sqlstore.New()` and make that return an error in case of invalid version. An object/parameter containing the minimum required versions could be passed by the caller. A separate method to return the cached version would still be available.

agnivade:
Nice, I like option 2. The `MinimumRequiredPostgresVersion` is a package level constant, so I think it would be okay to have a `New` function of that package respect that constant, rather than give the client an ability to control that.

And I don't really want to go to the extent of caching the returned version for the separate method. Let's just let that be, because there are two options - string/numeric. One would have to cache both. And in any case, the telemetry requests starts to taper off after a few times. So it should be fine.

Thanks for looking into it. I went into OSF mode and didn't really think too deeply, sorry about that.

streamer45 (approved):
Looks good :+1: 

I've created a ticket for the failing test. ([MM-36411](https://mattermost.atlassian.net/browse/MM-36411))

PR #17766, opened 2021-06-11
[MM-36318] Avoid redundant bot user update

streamer45:
#### Summary

A sudden spike in `getChannelMember` API calls was noticed during the (now hourly) community server updates (full context [here](https://community-daily.mattermost.com/private-core/pl/gen5i1kcu7n68xxs4n5zyufnrr)). Moreover, most of these calls were failing with 404s. After an initial investigation we were able to verify that such errors were all related  to bot users (plugin created) which are usually not part of any channel (hence the not founds).

Most of these plugins do a call to `Helpers.EnsureBot()` during activation. This caused the bot user to get updated and a websocket event to get sent. In response to such event each client would make the API call, which was tracked down to:

https://github.com/mattermost/mattermost-webapp/blob/0e8a3f375306d18eaf52aa191e07d8d5556fbea9/actions/websocket_actions.jsx#L1002

PR adds an additional check to verify if the bot update is needed, and if not return early. This avoids the ws event to get sent and should prevent the spike of calls coming in response.

#### Ticket

https://mattermost.atlassian.net/browse/MM-36318

#### Release Note

```release-note
Fixed an issue where a redundant user_update websocket event was generated for bot users.
```

### DIFF
--- a/app/bot.go
+++ b/app/bot.go
@@ -216,6 +216,10 @@ func (a *App) PatchBot(botUserId string, botPatch *model.BotPatch) (*model.Bot, 
 		return nil, err
 	}
 
+	if !bot.WouldPatch(botPatch) {
+		return bot, nil
+	}
+
 	bot.Patch(botPatch)
 
 	user, nErr := a.Srv().Store.User().Get(context.Background(), botUserId)
--- a/model/bot.go
+++ b/model/bot.go
@@ -129,6 +129,8 @@ func BotFromJson(data io.Reader) *Bot {
 }
 
 // Patch modifies an existing bot with optional fields from the given patch.
+// TODO 6.0: consider returning a boolean to indicate whether or not the patch
+// applied any changes.
 func (b *Bot) Patch(patch *BotPatch) {
 	if patch.Username != nil {
 		b.Username = *patch.Username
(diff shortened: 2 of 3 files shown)


agnivade (approved):
Brilliant stuff as usual!

I guess this is mainly to do with our unique deployment. Since you have other users connected to the cluster, even if you restart one server, the websocket broadcast happens across all nodes.

streamer45:
> I guess this is mainly to do with our unique deployment. Since you have other users connected to the cluster, even if you restart one server, the websocket broadcast happens across all nodes.

Yes, our setup doesn't help as ws connections are active during the upgrade happens. But I could see something similar (of lesser proportions) still happen on any plugin activation, regardless of whether the server has been restarted.

hmhealey (approved):

hmhealey:
Looks good. I wonder if this could be related to some of the performance issues we've been battling with for a customer. I think they saw some of these messages in their logs, so maybe there's a rogue plugin causing their client to make a bunch of requests

streamer45:
> Looks good. I wonder if this could be related to some of the performance issues we've been battling with for a customer. I think they saw some of these messages in their logs, so maybe there's a rogue plugin causing their client to make a bunch of requests

That's possible although I'd expect the performance to be impacted in short bursts (a couple of minutes tops). Anything longer would suggest some other issue.

furqanmlk (approved):
**Before PR**
image

**After PR**
image

PR #17775, opened 2021-06-12
[MM-11143] Strip markdown formatting characters from push notifications

ahmadkarlam:
#### Summary
Strip markdown formatting characters from push notifications

#### Ticket Link
Fixes https://github.com/mattermost/mattermost-server/issues/17616

#### Release Note

```release-note
Markdown formatting is now stripped from push notifications.
```


streamer45:
Hi @ahmadkarlam . Thanks for your contribution :+1: .
Some checks and tests are failing, can you have a look? Let us know if you need help.

ahmadkarlam:
> Hi @ahmadkarlam . Thanks for your contribution 👍 .
> Some checks and tests are failing, can you have a look? Let us know if you need help.

Still have some check pending

ahmadkarlam:
@streamer45 `ee/circleci` still fail, can you tell me what's wrong ?

streamer45:
> @streamer45 `ee/circleci` still fail, can you tell me what's wrong ?

I think it's unrelated, we may just have to merge latest changes from master.

agnivade:
@ahmadkarlam - Thank you for your contribution!

I am wondering if we can use a markdown parser here to go through the text and strip out any formatting? Using so many regexes seems a bit brittle here especially since a markdown parser can do a single pass and split out nodes as it encounters them. What do you think?

ahmadkarlam:
> @ahmadkarlam - Thank you for your contribution!
> 
> I am wondering if we can use a markdown parser here to go through the text and strip out any formatting? Using so many regexes seems a bit brittle here especially since a markdown parser can do a single pass and split out nodes as it encounters them. What do you think?

At first, I thought using regex would use a lot of resources, markdown parser didn't think of that before, it seems like a good idea
I found this package https://github.com/gomarkdown/markdown, what do you think?

agnivade:
I am leaning towards https://github.com/yuin/goldmark because it has fuzz tests. Markdown parsing is notorious for causing crashes.

ahmadkarlam:
> I am leaning towards https://github.com/yuin/goldmark because it has fuzz tests. Markdown parsing is notorious for causing crashes.

okay, I will try it

ahmadkarlam:
> I am leaning towards https://github.com/yuin/goldmark because it has fuzz tests. Markdown parsing is notorious for causing crashes.

Hi, I'm making a custom renderer with goldmark, is that what you want? Or goldmark has a ready made renderer without the need to make a custom one? Just to make sure.

agnivade:
Yeah, so basically you need to create a `NodeRenderer` and just register listener methods for all kinds of nodes. And in that method, just dump the node without doing anything. Here's a quick example which just dumps heading nodes. You probably need to dig a bit deeper into how to display the `ast.Node` properly, but this should give you a starting point:

```go
package main

import (
	"bytes"
	"fmt"

	"github.com/yuin/goldmark"
	"github.com/yuin/goldmark/ast"
	"github.com/yuin/goldmark/extension"
	"github.com/yuin/goldmark/parser"
	"github.com/yuin/goldmark/renderer"
	"github.com/yuin/goldmark/util"
)

func main() {
	md := goldmark.New(
		goldmark.WithExtensions(extension.GFM),
		goldmark.WithParserOptions(
			parser.WithAutoHeadingID(),
		),
		goldmark.WithRenderer(
			renderer.NewRenderer(renderer.WithNodeRenderers(util.Prioritized(new(), 1000))),
		),
	)

	var buf bytes.Buffer
	if err := md.Convert([]byte(`hifd
## hello

		dkj toy`), &buf); err != nil {
		panic(err)
	}

	fmt.Println(buf.String())
}

type myRenderer struct {
}

func new() *myRenderer {
	return &myRenderer{}
}

func (r *myRenderer) RegisterFuncs(reg renderer.NodeRendererFuncRegisterer) {
	reg.Register(ast.KindHeading, r.renderHeading)
        // You need to fill out these methods as well
        // reg.Register(ast.KindDocument, r.renderDocument)
	// reg.Register(ast.KindBlockquote, r.renderBlockquote)
	// reg.Register(ast.KindCodeBlock, r.renderCodeBlock)
	// reg.Register(ast.KindFencedCodeBlock, r.renderFencedCodeBlock)
	// reg.Register(ast.KindHTMLBlock, r.renderHTMLBlock)
	// reg.Register(ast.KindList, r.renderList)
	// reg.Register(ast.KindListItem, r.renderListItem)
	// reg.Register(ast.KindParagraph, r.renderParagraph)
	// reg.Register(ast.KindTextBlock, r.renderTextBlock)
	// reg.Register(ast.KindThematicBreak, r.renderThematicBreak)
}

func (r *myRenderer) renderHeading(w util.BufWriter, source []byte, node ast.Node, entering bool) (ast.WalkStatus, error) {
	n := node.(*ast.Heading)
	if entering {
                // This is probably not the best way to dump node contents. Please re-check.
		for _, attr := range n.Attributes() {
			w.Write(attr.Value.([]byte))
		}
	}
	return ast.WalkContinue, nil
}
```

ahmadkarlam:
I have some problem when using extension from goldmark.
Screenshot 2021-06-23 at 13 43 33

After `go mod vendor`, folder extension doesn't appear on vendor folder. @streamer45 can you help me with this ? I never use `go mod vendor` so I don't know whats the root cause.

I already open discussion on https://github.com/yuin/goldmark/discussions/226

agnivade:
I think perhaps it's because you are using the 1.3.5 release. Try using the 1.3.8 release and run "go mod vendor" again.

ahmadkarlam:
> I think perhaps it's because you are using the 1.3.5 release. Try using the 1.3.8 release and run "go mod vendor" again.

thanks, I can use extension now. I've updated the code.

ahmadkarlam:
I already updated the code, can you check it ? @streamer45 @agnivade

agnivade:
Nice work @ahmadkarlam , I will take a look soon. Claudio is on PTO, so I have swapped him with @harshilsharma63.

app/notification_push.go
```diff
@@ -136,6 +137,7 @@ func (a *App) sendPushNotification(notification *PostNotification, user *model.U
 	cfg := a.Config()
 	channel := notification.Channel
 	post := notification.Post
+	post.Message = utils.StripMarkdown(post.Message)
```
agnivade, line 140: We should do the processing in the asynchronous queue for push notification. Let's move this inside the `*App.sendPushNotificationSync` method instead.

utils/markdown.go
```diff
@@ -0,0 +1,179 @@
...
+	md := goldmark.New(
+		goldmark.WithExtensions(extension.GFM),
+		goldmark.WithRenderer(
+			renderer.NewRenderer(renderer.WithNodeRenderers(
+				util.Prioritized(newNotificationRenderer(), 500),
+			)),
+		),
+	)
+
+	var buf bytes.Buffer
```
agnivade, line 28: Let's use `var buf strings.Builder` since the end result we want is always a string.

utils/markdown.go
```diff
@@ -0,0 +1,179 @@
...
+	if err := md.Convert([]byte(markdown), &buf); err != nil {
+		return ""
```
agnivade, line 30: Let's return the error to the caller. And in the call-site, let's do an `mlog.Warn` of the error and proceed with the original string instead. We don't want to send a blank push notification if the parsing fails.

utils/markdown.go
```diff
@@ -0,0 +1,179 @@
...
+	}
+
+	return strings.TrimSpace(buf.String())
+}
+
+type notificationRenderer struct {
```
agnivade, line 36: Perhaps `markdownStripper` is a better name? [0/5](https://handbook.mattermost.com/company/about-mattermost/list-of-terms#0-5-1-5-2-5-3-5-4-5-5-5).

utils/markdown.go
```diff
@@ -0,0 +1,179 @@
...
+
+// renderDefault renderer function to renderDefault without changes
+func (r *notificationRenderer) renderDefault(w util.BufWriter, source []byte, node ast.Node, entering bool) (ast.WalkStatus, error) {
+	return ast.WalkContinue, nil
+}
+
+func (r *notificationRenderer) renderHeading(w util.BufWriter, source []byte, node ast.Node, entering bool) (ast.WalkStatus, error) {
+	if !entering {
+		if node.NextSibling() != nil {
+			_, _ = w.WriteString(" ")
```
agnivade, line 82: I see `w.WriteString(" ")` mixed with `w.WriteByte(' ')`. Let's choose one and stick with it to remain consistent.

utils/markdown.go
```diff
@@ -0,0 +1,179 @@
...
+}
+
+func (r *notificationRenderer) renderListItem(w util.BufWriter, source []byte, node ast.Node, entering bool) (ast.WalkStatus, error) {
+	if entering {
+		fc := node.FirstChild()
+		if fc != nil {
+			if _, ok := fc.(*ast.TextBlock); !ok {
+				_ = w.WriteByte(' ')
+			}
+		}
```
agnivade, line 95: Can you explain why this is needed? Removing this does not fail any test case. In that case, `renderHeading`, `renderListItem` and `renderParagraph` can all become the same method.

utils/markdown.go
```diff
@@ -0,0 +1,179 @@
...
+		if node.NextSibling() != nil {
+			_, _ = w.WriteString(" ")
+		}
+	}
+	return ast.WalkContinue, nil
+}
+
+func (r *notificationRenderer) renderCodeBlock(w util.BufWriter, source []byte, node ast.Node, entering bool) (ast.WalkStatus, error) {
+	n := node.(*ast.CodeBlock)
+	source = bytes.ReplaceAll(source, []byte("\n"), []byte(" "))
```
agnivade, line 115: There is no need to replace `\n`. Mobile can display it correctly.

utils/markdown.go
```diff
@@ -0,0 +1,179 @@
...
+	if entering {
+		r.writeLines(w, source, n)
+	}
+
+	return ast.WalkContinue, nil
+}
+
+func (r *notificationRenderer) renderFencedCodeBlock(w util.BufWriter, source []byte, node ast.Node, entering bool) (ast.WalkStatus, error) {
+	n := node.(*ast.FencedCodeBlock)
+	source = bytes.ReplaceAll(source, []byte("\n"), []byte(" "))
```
agnivade, line 125: Same as above.

utils/markdown.go
```diff
@@ -0,0 +1,179 @@
...
+}
+
+func (r *notificationRenderer) renderText(w util.BufWriter, source []byte, node ast.Node, entering bool) (ast.WalkStatus, error) {
+	if !entering {
+		return ast.WalkContinue, nil
+	}
+	n := node.(*ast.Text)
+	segment := n.Segment
+	if n.IsRaw() {
+		_, _ = w.Write(segment.Value(source))
```
agnivade, line 140: Move this line outside the `if` condition.

utils/markdown.go
```diff
@@ -0,0 +1,179 @@
...
+	_, _ = w.Write(n.Value)
+	return ast.WalkContinue, nil
+}
+
+func (r *notificationRenderer) renderTable(w util.BufWriter, source []byte, node ast.Node, entering bool) (ast.WalkStatus, error) {
+	return ast.WalkSkipChildren, nil
+}
+
+func (r *notificationRenderer) writeLines(w util.BufWriter, source []byte, n ast.Node) {
+	l := n.Lines().Len()
```
agnivade, line 173: Move the value inline. [0/5](https://handbook.mattermost.com/company/about-mattermost/list-of-terms#0-5-1-5-2-5-3-5-4-5-5-5).

utils/markdown.go
```diff
@@ -0,0 +1,179 @@
...
+		return ast.WalkContinue, nil
+	}
+	n := node.(*ast.String)
+	_, _ = w.Write(n.Value)
+	return ast.WalkContinue, nil
+}
+
+func (r *notificationRenderer) renderTable(w util.BufWriter, source []byte, node ast.Node, entering bool) (ast.WalkStatus, error) {
+	return ast.WalkSkipChildren, nil
+}
```
agnivade, line 170: Any reason why tables are not rendered properly? It is okay to just let them be markdown and revisit it in a next PR. But this just skips the rendering entirely.

utils/markdown_test.go
```diff
@@ -0,0 +1,269 @@
...
+			name: "text: empty string",
+			args: "",
+			want: "",
+		},
+	}
+	for _, tt := range tests {
+		t.Run(tt.name, func(t *testing.T) {
+			if got := StripMarkdown(tt.args); got != tt.want {
+				t.Errorf("StripMarkdown() = %v, want %v", got, tt.want)
+			}
```
agnivade, line 266: We use the testify library. This should be `assert.Equal(t, tt.want, got)`.

utils/markdown.go
```diff
@@ -0,0 +1,179 @@
...
+
+	// table
+	reg.Register(astExt.KindTable, r.renderTable)
+
+	// strikethrough
+	reg.Register(astExt.KindStrikethrough, r.renderDefault)
+}
+
+// renderDefault renderer function to renderDefault without changes
+func (r *notificationRenderer) renderDefault(w util.BufWriter, source []byte, node ast.Node, entering bool) (ast.WalkStatus, error) {
```
agnivade, line 75: Let's replace the variables with `_` to indicate that they aren't used.

app/notification_push.go
```diff
@@ -11,6 +11,7 @@ import (
 	"strings"
 	"sync"
 
+	"github.com/mattermost/mattermost-server/v5/utils"
```
agnivade, line 14: Let's move this to the import group below.

agnivade (commented):
Nice work there @ahmadkarlam ! I have left some minor comments.

utils/markdown_test.go
```diff
@@ -0,0 +1,269 @@
...
+			want: "Italics with asterisks or underscores.",
+		},
+		{
+			name: "del: Strikethrough ~~strike this.~~",
+			args: "Strikethrough ~~strike this.~~",
+			want: "Strikethrough strike this.",
+		},
+		{
+			name: "links: [inline-style link](http://localhost:8065)",
+			args: "[inline-style link](http://localhost:8065)",
```
harshilsharma63, line 193: I'd rather think the URL is more important than the label. I'd prefer if we converted -

```markdown
[inline-style link](http://localhost:8065)
```

to

```markdown
inline-style link (http://localhost:8065)
```

I can see the URL in the notification and make sense of it.

For example, consider this markdown-

> Hey, check out [this](https://example.com) page!

Your logic will convert this to-

> Hey, check out this page!

And my proposal will do this-

> Hey, check out this (https://example.com) page!

utils/markdown_test.go
```diff
@@ -0,0 +1,269 @@
...
+			want: "inline-style link",
+		},
+		{
+			name: "image: ![image link](http://localhost:8065/image)",
+			args: "![image link](http://localhost:8065/image)",
+			want: "image link",
```
harshilsharma63, line 199: Same for the image. The label provides absolutely no value to the reader. The URL, however, does. I'd prefer if we retained the URL in the converted string just like I proposed for the URL.

utils/markdown_test.go
```diff
@@ -0,0 +1,269 @@
...
+		},
+		{
+			name: "text: plain",
+			args: "This is plain text.",
+			want: "This is plain text.",
+		},
+		{
+			name: "text: multiline",
+			args: "This is multiline text.\nHere is the next line.\n",
+			want: "This is multiline text. Here is the next line.",
```
harshilsharma63, line 209: Multiline notifications work fine. We can retain them.

harshilsharma63 (changes requested):
Awesome work!

utils/markdown.go
ahmadkarlam: I will convert all `w.WriteString()` to `w.WriteByte()`

utils/markdown.go
ahmadkarlam: Actually I just copy-paste from html renderer, yeah those three function can be wrap into only one function, thanks.

utils/markdown.go
ahmadkarlam: That means, in the code block there are characters `\n` that will be handled by mobile ? If yes I will update the unit test, https://github.com/mattermost/mattermost-server/pull/17775/files#diff-fd01f7b59eaca4eb783d3811157920648fb0fc7915e5cf0b4ab6202aa8075b6bR54-R58

utils/markdown_test.go
ahmadkarlam: I just implemented based on this task [MM-11143](https://mattermost.atlassian.net/browse/MM-11143)
It is okay change the test case ?

utils/markdown.go
ahmadkarlam: I think that's good idea

utils/markdown.go
ahmadkarlam: Screenshot_1624691285

Maybe this still acceptable for user read it.

utils/markdown.go
ahmadkarlam: I follow other renderer, they use postfix renderer and not all of markdown will be stripped in this function.

utils/markdown.go
agnivade: Yes, they will be.

utils/markdown.go
agnivade: I think this is a good start. Tables are not that common. We can come back to it later.

utils/markdown.go
```diff
@@ -0,0 +1,151 @@
...
+				util.Prioritized(newNotificationRenderer(), 500),
+			)),
+		),
+	)
+
+	var buf strings.Builder
+	if err := md.Convert([]byte(markdown), &buf); err != nil {
+		mlog.Warn("failed parse to markdown")
+		return markdown
+	}
```
agnivade, line 33: The idea is to return the error to the caller. Like `StripMarkdown(string) (string, error)`. And then in `notification_push.go`, check the error, and log it, and continue with the original string.

This is a low level function, and should not be deciding how to respond to an error.

agnivade (commented):
Excellent work @ahmadkarlam !

Just one last comment to take care of.

utils/markdown.go
ahmadkarlam: I thought you want me to just logged it, sorry I will update.

utils/markdown.go
ahmadkarlam: okay, I will return original string for tables.

utils/markdown.go
```diff
@@ -0,0 +1,142 @@
...
+		goldmark.WithRenderer(
+			renderer.NewRenderer(renderer.WithNodeRenderers(
+				util.Prioritized(newNotificationRenderer(), 500),
+			)),
+		),
+	)
+
+	var buf strings.Builder
+	if err := md.Convert([]byte(markdown), &buf); err != nil {
+		return markdown, err
```
agnivade, line 30: This is not correct. We should return `"", err` here. Only at the call-site, should the decision of using the original string in case of an error should be taken.

app/notification_push.go
```diff
@@ -55,6 +56,11 @@ type PushNotification struct {
 func (a *App) sendPushNotificationSync(post *model.Post, user *model.User, channel *model.Channel, channelName string, senderName string,
 	explicitMention bool, channelWideMention bool, replyToThreadType string) *model.AppError {
 	cfg := a.Config()
+	message, errMarkdown := utils.StripMarkdown(post.Message)
+	if errMarkdown != nil {
+		mlog.Warn("failed parse to markdown", mlog.Err(errMarkdown))
+	}
+	post.Message = message
```
agnivade, line 63:
```suggestion
	message, err := utils.StripMarkdown(post.Message)
	if err != nil {
		mlog.Warn("Failed parse to markdown", mlog.String("post_id", post.Id), mlog.Err(err))
	} else {
	        post.Message = message
	}
```

And rename `err` to `appErr` below.

ahmadkarlam: I'm sorry, I miss understanding this one 🙏

agnivade (approved):
Thank you @ahmadkarlam ! Brilliant stuff.

utils/markdown_test.go
ahmadkarlam: @harshilsharma63 @agnivade how about this case ? Will it be implemented or not?

harshilsharma63: It that was the requirement in the ticket then I'm fine with this PR.

We can discuss these suggestions with Eric Sethna independent of this PR.

harshilsharma63 (approved):
LGTM

PR #17777, opened 2021-06-14
[MM-36411] Fix flaky TestUpdateConfigDiffInAuditRecord

streamer45:
#### Summary

PR attempts to fix flaky `TestUpdateConfigDiffInAuditRecord`. I wasn't able to reproduce this locally but I have a feeling it may be caused by the logs not getting flushed in time. To force a flush we are now shutting down the server before attempting to read the log file.

#### Ticket

https://mattermost.atlassian.net/browse/MM-36411

#### Release Note

```release-note
NONE
```

### DIFF
--- a/api4/config_test.go
+++ b/api4/config_test.go
@@ -456,7 +456,12 @@ func TestUpdateConfigDiffInAuditRecord(t *testing.T) {
 		},
 	}
 	th := SetupWithServerOptions(t, options)
-	defer th.TearDown()
+	shouldShutdown := true
+	defer func() {
+		if shouldShutdown {
+			th.TearDown()
+		}
+	}()
 
 	cfg, resp := th.SystemAdminClient.GetConfig()
 	CheckNoError(t, resp)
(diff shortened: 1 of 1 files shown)


api4/config_test.go
```diff
@@ -470,6 +475,10 @@ func TestUpdateConfigDiffInAuditRecord(t *testing.T) {
 	})
 	require.Equal(t, timeoutVal+1, *cfg.ServiceSettings.ReadTimeout)
 
+	// Shutting down the server to make sure all logged records get flushed.
+	th.TearDown()
```
wiggin77, line 479: Here you could just call `th.Server.Log.Flush()`.  I suggest doing that anyway in case the "flush not called on shutdown" issue is still in this version of server (I suspect it is).

Shutdown does make attempts to drain any queues but is not as complete as flush.

streamer45: Awesome, will update :+1:

agnivade (approved):

wiggin77 (approved):
LGTM :+1:

PR #17789, opened 2021-06-16
Prevent JS Execution on Pre-login OAUTH screen 

anurag6713:
#### Summary
Prevents `javascript:` protocol in pre login OAUTH screens. 

We are using the user passed query param value in the HTML output so that in case of misconfiguration, it will still let the user to **go back** to their respective mobile app. 

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-36250

#### Release Note
```release-note
NONE
```

### DIFF
--- a/web/oauth_test.go
+++ b/web/oauth_test.go
@@ -390,19 +390,29 @@ func TestMobileLoginWithOAuth(t *testing.T) {
 
 	responseWriter := httptest.NewRecorder()
 
-	// Incase of valid but not supported scheme, we should throw error and include the invalid scheme in the response
-	request, _ := http.NewRequest(http.MethodGet, th.App.GetSiteURL()+"/oauth/gitlab/mobile_login?redirect_to="+url.QueryEscape("randomScheme://"), nil)
+	t.Run("Should include redirect URL in the output when valid URL Scheme is passed", func(t *testing.T) {
+		request, _ := http.NewRequest(http.MethodGet, th.App.GetSiteURL()+"/oauth/gitlab/mobile_login?redirect_to="+url.QueryEscape("randomScheme://"), nil)
 
-	mobileLoginWithOAuth(c, responseWriter, request)
-	assert.Contains(t, responseWriter.Body.String(), "randomScheme://")
-	assert.NotContains(t, responseWriter.Body.String(), siteURL)
+		mobileLoginWithOAuth(c, responseWriter, request)
+		assert.Contains(t, responseWriter.Body.String(), "randomScheme://")
+		assert.NotContains(t, responseWriter.Body.String(), siteURL)
+	})
 
-	// Incase of invalid scheme or JS Injection, redirectURL should be replaced by SiteURL
-	request2, _ := http.NewRequest(http.MethodGet, th.App.GetSiteURL()+"/oauth/gitlab/mobile_login?redirect_to="+url.QueryEscape("javascript:alert('hello')"), nil)
+	t.Run("Should not include the redirect URL consisting of javascript protocol", func(t *testing.T) {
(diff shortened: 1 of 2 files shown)


srkgupta (dismissed):
Hi @anurag6713

As generated by the CodeQL alert here, we should also add checks for data and vbscript.

https://github.com/mattermost/mattermost-server/security/code-scanning/86

Its better if we use a whitelisted approach and only allow valid protocols like http(s). For everything else you can simply default to the siteUrl

anurag6713:
> Its better if we use a whitelisted approach and only allow valid protocols like http(s). For everything else you can simply default to the browser's back button behavior.

Redirect URL here is for mobile app custom schemes. To login, we allow only the whitelisted ones like `mmauth://` and `mmauthbeta://`.

When we get an invalid ones because of misconfiguration, we want users to go back to their respective App. That is the only reason why we are rendering the user input in the html output.

anurag6713:
> I'd suggest adding a test with a mixed-case JaVaScriPt: URI, otherwise LGTM.

Thank you @jupenur. I did test with the mixed-case, as URL parser would return the scheme in lower case, I haven't written the test case around that but covering it in the test case now.

sbishel (approved):

web/oauth_test.go
```diff
@@ -367,6 +367,54 @@ func TestOAuthAccessToken(t *testing.T) {
...
+	th.App.UpdateConfig(func(cfg *model.Config) { *cfg.ServiceSettings.SiteURL = siteURL })
+
+	translationFunc := i18n.GetUserTranslations("en")
+	c.AppContext.SetT(translationFunc)
+	buffer := &bytes.Buffer{}
+	c.Logger = mlog.NewTestingLogger(t, buffer)
+	provider := &MattermostTestProvider{}
+	einterfaces.RegisterOauthProvider(model.SERVICE_GITLAB, provider)
+
+	responseWriter := httptest.NewRecorder()
```
ashishbhate, line 391: A new recorder should be made inside each individual test (i.e. a new `responseWriter := httptest.NewRecorder()` inside each `t.Run`) call. By using a single recorder we're appending every test's response to the recorder, which may cause spurious test fails in later tests if the earlier tests generate a response the later tests don't like.

web/oauth_test.go
```diff
@@ -367,6 +367,54 @@ func TestOAuthAccessToken(t *testing.T) {
...
+		mobileLoginWithOAuth(c, responseWriter, request)
+		assert.NotContains(t, responseWriter.Body.String(), "javascript:alert('hello')")
+		assert.Contains(t, responseWriter.Body.String(), siteURL)
+	})
+
+	t.Run("Should not include the redirect URL consisting of javascript protocol in mixed case", func(t *testing.T) {
+		request, _ := http.NewRequest(http.MethodGet, th.App.GetSiteURL()+"/oauth/gitlab/mobile_login?redirect_to="+url.QueryEscape("JaVasCript:alert('hello')"), nil)
+
+		mobileLoginWithOAuth(c, responseWriter, request)
+		assert.NotContains(t, responseWriter.Body.String(), "JaVasCript:alert('hello')")
```
ashishbhate, line 413: Should we be checking the lower case version of this scheme is not present? 

`assert.NotContains(t, responseWriter.Body.String(), "javascript:alert('hello')")`

anurag6713: We are not converting it to lowercase in the output. It is rendered as it is.

web/oauth_test.go
anurag6713: Thanks @ashishbhate  🙂 Updated it.

ashishbhate (approved):

jupenur:
@anurag6713 Rohitesh is OoO so feel free to merge this without his approval.

PR #17796, opened 2021-06-17
MM-22845: Added support for permalink previews.

mkraft:
#### Summary

* Adds metadata about the referenced post when a post is created or updated via the API or App where the first link in the message is a permalink.
* Caches the data into the post metadata cache.
* Only embeds the data if the current user has permissions to view the post.
* Adds a prop in the post record which saves the ID of the referenced post.
* Adds an instance-wide config setting to disable the embedding of permalink metadata.
* Fixed a few linting warnings unrelated to the ticket.

#### Ticket Link

https://mattermost.atlassian.net/browse/MM-22845

#### Related PRs

* https://github.com/mattermost/enterprise/pull/1005
* https://github.com/mattermost/mattermost-webapp/pull/8256

#### Release Note

```release-note
Embedded information about referenced post for first links that are permalinks.
```


srkgupta:
Adding myself for review since some CodeQL alerts were generated on this PR.

model/client4.go
```diff
@@ -2955,6 +2959,19 @@ func (c *Client4) GetPost(postId string, etag string) (*Post, *Response) {
 	return PostFromJson(r.Body), BuildResponse(r)
 }
 
+// GetPost gets a single post.
```
hanzei, line 2962: Typo

mkraft:
> Adding myself for review since some CodeQL alerts were generated on this PR. Will be reviewing the PR once the changes are completed.

The indentation of that LOC changed and it seems to have triggered an alert. What's the fix?

app/post_metadata.go
```diff
@@ -383,74 +417,154 @@ func (a *App) getImagesInMessageAttachments(post *model.Post) []string {
...
+
+			var referencedTeam *model.Team
+			if referencedChannel != nil {
+				referencedTeam, appErr = a.GetTeam(referencedChannel.TeamId)
+				if appErr != nil {
+					return nil, nil, nil, appErr
+				}
+			}
+
+			permalink.PreviewPost = model.PreviewPostFromPost(referencedPost, referencedTeam, referencedChannel)
```
hmhealey, line 483: 0/5, but perhaps we could make a store method to get a PreviewPost in a single DB query

app/post_metadata.go
```diff
@@ -383,74 +417,154 @@ func (a *App) getImagesInMessageAttachments(post *model.Post) []string {
...
 	// Check the database if this isn't a new post. If it is a new post and the data is cached, it should be in memory.
 	if !isNewPost {
-		og, image, ok = a.getLinkMetadataFromDatabase(requestURL, timestamp)
+		og, image, permalink, ok = a.getLinkMetadataFromDatabase(requestURL, timestamp)
+
+		if !*a.Config().ServiceSettings.EnablePermalinkPreviews {
+			permalink = nil
+		}
+
+		if permalink != nil && permalink.PreviewPost == nil {
```
hmhealey, line 456: Since we're not persisting any information to the DB, should we maybe just skip this step? This branch also duplicates most of the code from below.

app/post_metadata.go
```diff
@@ -164,22 +167,31 @@ func (a *App) getEmojisAndReactionsForPost(post *model.Post) ([]*model.Emoji, []
...
 		}, nil
 	}
 
-	if firstLink == "" || !*a.Config().ServiceSettings.EnableLinkPreviews {
+	if firstLink == "" {
 		return nil, nil
 	}
 
-	og, image, err := a.getLinkMetadata(firstLink, post.CreateAt, isNewPost)
+	// Permalink previews are not toggled via the ServiceSettings.EnableLinkPreviews config setting.
```
hmhealey, line 181: Should permalink previews also be behind a feature flag?

hmhealey (changes requested):

srkgupta:
Checked the CodeQL alert and found that it was a False positive and have dismissed the same. Removing myself from list of reviewers.

api4/post.go
```diff
@@ -125,7 +126,7 @@ func createEphemeralPost(c *Context, w http.ResponseWriter, r *http.Request) {
 
 	w.WriteHeader(http.StatusCreated)
 	rp = model.AddPostActionCookies(rp, c.App.PostActionCookieSecret())
-	rp = c.App.PreparePostForClient(rp, true, false)
+	rp = c.App.PreparePostForClient(rp, true, false, c.AppContext.Session().UserId)
```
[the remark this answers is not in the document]
mkraft: Good question, there are at least 3 user ids in scope:
* `c.AppContext.Session().UserId`
* `ephRequest.Post.UserId`
* `ephRequest.UserID`
I think the session id is the correct one to use because it's not given as a user input; we don't want users to be able to provide a user ID and then have access to posts that they wouldn't normally be able to see.

app/post_metadata.go
esethna: @hmhealey yes good call. cc// @mkraft

app/post_metadata.go
```diff
@@ -196,6 +208,28 @@ func (a *App) getEmbedForPost(post *model.Post, firstLink string, isNewPost bool
...
+		}
+
+		if permalink.PreviewPost != nil {
+			referencedChannel, err := a.GetChannel(permalink.PreviewPost.Post.ChannelId)
+			if err != nil && err.StatusCode != http.StatusNotFound {
+				return nil, err
+			}
+
+			// Only embed the referenced post into the Data field if the containing channel wasn't deleted and the requesting user has access to the post.
+			if referencedChannel != nil && a.HasPermissionToChannel(askingUserID, referencedChannel.Id, model.PERMISSION_READ_CHANNEL) || (referencedChannel.Type == model.CHANNEL_OPEN && a.HasPermissionToTeam(askingUserID, referencedChannel.TeamId, model.PERMISSION_READ_PUBLIC_CHANNEL)) {
```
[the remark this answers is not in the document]
mkraft: Good question. I copied this from the main place that checks this permissions but I haven't questioned whether just the first part is sufficient.

app/post_metadata.go
mkraft: Thanks for bringing this up. I was on the fence about adding a feature flag but I've added it because it's low effort and gives us the choice to disable or phase the rollout.

app/post.go
```diff
@@ -351,7 +351,12 @@ func (a *App) CreatePost(c *request.Context, post *model.Post, channel *model.Ch
 
 	// Normally, we would let the API layer call PreparePostForClient, but we do it here since it also needs
 	// to be done when we send the post over the websocket in handlePostEvents
-	rpost = a.PreparePostForClient(rpost, true, false)
+	rpost = a.PreparePostForClient(rpost, true, false, c.Session().UserId)
```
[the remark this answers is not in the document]
mkraft: You're right. Thanks for flagging this. I've made a change that customizes the WS message based on authz to the post in the case that the post is a permalink preview post.

app/post_metadata.go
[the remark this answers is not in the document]
mkraft: `PERMISSION_READ_PUBLIC_CHANNEL` is a team-scoped permission, so it has to be checked on the team, whereas `PERMISSION_READ_CHANNEL` is channel-scoped. Long story is this is how it needs to be done for now and how it's done elsewhere. However, I did add a helper method to make it a bit less wordy.

app/post_metadata.go
mkraft: That's a great idea that I'd like to defer to a future improvement for the sake of getting this out an iterating on it. I've added [a Jira story](https://mattermost.atlassian.net/browse/MM-37081) that links to this comment.

app/post_metadata.go
```diff
@@ -383,74 +417,154 @@ func (a *App) getImagesInMessageAttachments(post *model.Post) []string {
 	return images
 }
 
-func (a *App) getLinkMetadata(requestURL string, timestamp int64, isNewPost bool) (*opengraph.OpenGraph, *model.PostImage, error) {
+func looksLikeAPermalink(url, siteURL string) bool {
+	expression := fmt.Sprintf(`^(%s).*(/pl/)[a-z0-9]{26}$`, siteURL)
```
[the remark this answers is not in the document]
mkraft: We could maybe support relative links, but that may be an edge case for our community/community-daily case. Not sure how common that is. Where are you 0-5 that we should support relative links out of the gate?

What's the sneaky regex thing you're talking about? I didn't follow.

app/post_metadata.go
mkraft: This populates the permalink that eventually is returned as part of the JSON payload, I'm not sure how this could be skipped. Maybe I'm not seeing something?

esethna:
Nice! @mkraft is there a way to test this front-end and backend together?

mkraft:
> Nice! @mkraft is there a way to test this front-end and backend together?

Yes, @BenCookie95's all over it for you. We'll DM you details.

app/post_metadata.go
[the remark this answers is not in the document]
hmhealey: I'm 0/5 on whether or not we ship this without support for community, but relative link support is 100% necessary to have this work on community since it uses the relative links to work across servers in the cluster. But that also needs https://mattermost.atlassian.net/browse/MM-29224 anyway because you need to use a Markdown link to do a relative link.

As for the sneaky bit, I was trying to figure out if there was any way to trick the regex and have it return true for something that wasn't a valid permalink, but I couldn't come up with any way to actually abuse that, so it's probably not an issue.

api4/post.go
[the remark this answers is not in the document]
hmhealey: Isn't `ephRequest.UserID` the one who would actually be receiving the post though? If I understand the code correctly, the first two IDs are the one making the ephemeral post and the third is the one receiving it, so I'd expect it to be done based on the recipient's user ID.

app/post_metadata.go
hmhealey: Maybe I'm misreading this, but the logic looks somewhat like
```
if the metadata is in the database:
    if the url is a permalink:
        load the post and return it

load the post and return it
```
so the information stored in the database isn't being used for permalinks because we're not actually retrieving anything about the preview post with `getLinkMetadataFromDatabase`. That's what's confusing me a bit about this code. I don't think it'll cause any problems, but the logic seems weird to me.

app/notification.go
```diff
@@ -458,7 +458,39 @@ func (a *App) SendNotifications(post *model.Post, team *model.Team, channel *mod
...
+		previewedPostID = val
+	}
+	var previewedChannel *model.Channel
+	if model.IsValidId(previewedPostID) {
+		previewedPost, err := a.GetSinglePost(previewedPostID)
+		if err != nil {
+			return nil, err
+		}
+
+		channelMembers, err := a.GetChannelMembersPage(channel.Id, 0, 10000000)
```
hmhealey, line 472: ~~Is this okay or should we be paging through this properly?~~

Never mind. We do this in a few other places. It looks kind of odd, but I guess we'll never have a channel with 10 million users. If we do end up with that, we definitely have bigger concerns than this breaking though.

app/notification.go
```diff
@@ -458,7 +458,39 @@ func (a *App) SendNotifications(post *model.Post, team *model.Team, channel *mod
...
+			return nil, err
+		}
+
+		for _, cm := range *channelMembers {
+			postForIndividualUser := post.Clone()
+			if !a.HasPermissionToReadChannel(cm.UserId, previewedChannel) {
+				postForIndividualUser.Metadata.Embeds[0].Data = nil
+			}
+			message.Broadcast.UserId = cm.UserId
+			message.Add("post", postForIndividualUser.ToJson())
```
hmhealey, line 488: Since these are mutating the message, do we need to copy the message in each of these loops? I'm thinking both because future iterations of the loop might want the non-cleared metadata, but also because we would potentially be getting some racy behaviour here assuming `a.Publish` does something asynchronous.

app/notification.go
```diff
@@ -458,7 +458,39 @@ func (a *App) SendNotifications(post *model.Post, team *model.Team, channel *mod
...
+		if err != nil {
+			return nil, err
+		}
+
+		for _, cm := range *channelMembers {
+			postForIndividualUser := post.Clone()
+			if !a.HasPermissionToReadChannel(cm.UserId, previewedChannel) {
+				postForIndividualUser.Metadata.Embeds[0].Data = nil
+			}
+			message.Broadcast.UserId = cm.UserId
```
hmhealey, line 487: Not for this PR, but I feel like it would be really nice if this had a `UserIds` so we'd only have to potentially publish two messages here, one for users with permission and one for those without.

app/post_metadata.go
```diff
@@ -211,18 +211,14 @@ func (a *App) getEmbedForPost(post *model.Post, firstLink string, isNewPost bool
...
-
 		if permalink.PreviewPost != nil {
 			referencedChannel, err := a.GetChannel(permalink.PreviewPost.Post.ChannelId)
 			if err != nil && err.StatusCode != http.StatusNotFound {
 				return nil, err
 			}
 
 			// Only embed the referenced post into the Data field if the containing channel wasn't deleted and the requesting user has access to the post.
-			if referencedChannel != nil && a.HasPermissionToChannel(askingUserID, referencedChannel.Id, model.PERMISSION_READ_CHANNEL) || (referencedChannel.Type == model.CHANNEL_OPEN && a.HasPermissionToTeam(askingUserID, referencedChannel.TeamId, model.PERMISSION_READ_PUBLIC_CHANNEL)) {
+			if askingUserID == "" || (referencedChannel != nil && a.HasPermissionToReadChannel(askingUserID, referencedChannel)) {
```
hmhealey, line 221: Why pass the empty string for new posts and updates? I think it might be safer to keep passing that since I think it gives us an extra layer of security by preventing us from potentially accidentally leaking some data if a user without permission to see a post is posting a permalink to it.

I don't expect us to hopefully leak stuff here, but I'm trying to make the default case that we don't send anything just in case something gets broken in the future.

hmhealey (changes requested):

api4/post.go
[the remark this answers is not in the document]
mkraft: Maybe... I'm a bit confused by the difference between `PostEphemeral.UserID` and `PostEphemeral.Post.UserId` (which is set to `c.AppContext.Session().UserId`).

app/post_metadata.go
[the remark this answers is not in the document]
mkraft: > relative link support is 100% necessary to have this work on community

Yes, totally agree. There was some talk about disabling that plugin, but I'll take a look at what's involved in supporting it out-of-the-gate and I'll either add it to this PR and we can review the changes or I'll add it as a separate ticket for afterwards.

app/notification.go
mkraft: Yeah, only has a small performance for some peace of mind.

app/post_metadata.go
mkraft: Good question. There are absolutely other ways to achieve the same thing. It stems from the edge case that you might post a link to a post that you yourself can't access but others in the channel can.

This method puts the referenced post in the metatdata and *then* `nil`s it if for those who don't have authz. Alternatively, we could *not* put the referenced post in the metadata and then *add* it for those who have authz.

I totally see your point, you're concerned that someone might come along and use the app method with a blank `askingUserID` and expect to have the results authorized to the current user?

BenCookie95 (changes requested):

api4/post.go
```diff
@@ -372,6 +373,57 @@ func getPost(c *Context, w http.ResponseWriter, r *http.Request) {
 	w.Write([]byte(post.ToJson()))
 }
 
+func getPosts(c *Context, w http.ResponseWriter, r *http.Request) {
+	var postIDs []string
```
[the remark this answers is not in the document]
mkraft: Perfect, I'll remove all of this code then.

app/post_metadata.go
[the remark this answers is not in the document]
mkraft: Recapping what we talked about offline: supporting the relative links used by community servers depends upon first [supporting markdown links](https://mattermost.atlassian.net/browse/MM-29224). When that's complete we can make a quick change to the regex so that relative permalinks embed the referenced post info.

api4/post.go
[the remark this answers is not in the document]
mkraft: Recapping what we discussed: it looks like we have the correct user ID here.

app/post_metadata.go
mkraft: I've completely removed the knowledge of the "asking user" from this method. It now embeds the referenced posts' metadata (when appropriate) regardless of authz rules. A later "sanitization" method can optionally be invoked to invoke the authz rules for the given user.

michaelgamble (dismissed):
Three minor UX issues:
1. post__preview-footer >> has double opacity on it, should just be the one (56%)
2. showMoreButton >> should be font-size 12 instead of 14
3. On hover of the post preview >> the dropshadow is supposed to change from "Elevation 1" to "Elevation 2" shadow style
 
To clarify point 3, only when hovering the preview card itself, not the whole post row.. so similar to how hovering over the preview link causes it to underline.. hovering over the card will cause a slight shift in the dropshadow.

/* Elevation 1 */
box-shadow: 0px 2px 3px rgba(0, 0, 0, 0.08);

/* Elevation 2 */
box-shadow: 0px 4px 6px rgba(0, 0, 0, 0.12);

esethna:
@BenCookie95 I think these were meant to be posted on the webapp PR ^

app/post.go
```diff
@@ -647,16 +679,47 @@ func (a *App) UpdatePost(c *request.Context, post *model.Post, safeUpdate bool)
...
-	message := model.NewWebSocketEvent(model.WEBSOCKET_EVENT_POST_EDITED, "", rpost.ChannelId, "", nil)
-	message.Add("post", rpost.ToJson())
-	a.Publish(message)
+	rpost, previewPost, nErr := a.addPostPreviewProp(rpost)
+	if nErr != nil {
+		return nil, model.NewAppError("UpdatePost", "app.post.update.app_error", nil, nErr.Error(), http.StatusInternalServerError)
+	}
+
+	if previewPost != nil {
+		channelMembers, err := a.GetChannelMembersPage(channel.Id, 0, 10000000)
```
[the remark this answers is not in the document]
mkraft: You're absolutely right, that's a good point. I did some refactoring to remove the duplication. Take a look and let me know your thoughts.

stylianosrigas:
Removing the test server as it is failing to be created with the following error: Treatments: you passed PermalinkPreviews that does not exist in this environment, please double check what Splits exist in the web console

model/post_list.go
```diff
@@ -25,6 +25,15 @@ func NewPostList() *PostList {
 	}
 }
 
+func (o *PostList) Clone() *PostList {
+	return &PostList{
+		Order:      o.Order,
```
hmhealey, line 30: I think we also need to copy Order and Posts here. I'm pretty sure they're stored as something like pointers that aren't copied.

hmhealey (changes requested):

app/post.go
```diff
@@ -664,15 +685,69 @@ func (a *App) UpdatePost(c *request.Context, post *model.Post, safeUpdate bool)
...
+	} else {
+		return false, nil
+	}
+
+	if !model.IsValidId(previewedPostID) {
+		mlog.Warn("invalid post prop value", mlog.String("prop_key", model.POST_PROPS_PREVIEWED_POST), mlog.String("prop_value", previewedPostID))
+		return false, nil
+	}
+
+	previewedPost, err := a.GetSinglePost(previewedPostID)
```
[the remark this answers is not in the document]
mkraft: I think you're right that the WS message for the post shouldn't be interrupted simply because the referenced post (or channel) were deleted. I've update `publishWebsocketEventForPermalinkPost` to warn but not return the error if the referenced post or channel is 404.

hmhealey (approved):

BenCookie95 (approved):
LGTM, spent some time testing today!

esethna:
@mkraft just checking, who are the required reviews for this PR? Can you help remove anyone whose review wouldn't block this from going in? I'd like to aim to merge this for v6.0 deadline

@jgilliam17 for QA review ^

cc// @amyblais

mkraft:
> @mkraft just checking, who are the required reviews for this PR? Can you help remove anyone whose review wouldn't block this from going in? I'd like to aim to merge this for v6.0 deadline if possible (Aug 16)
> 
> @jgilliam17 for QA review ^
> 
> cc// @amyblais

I think @hanzei had just identified a typo but wasn't planning to be a reviewer and I think @michaelgamble had meant to be on the webapp pr.

esethna (approved):
Approving the server PR, webapp I think might need some pending changes as per: https://github.com/mattermost/mattermost-webapp/pull/8403#issuecomment-888482579

furqanmlk (commented):
Thanks @BenCookie95 
Test cases passed except few, I mentioned below with red icon

<meta charset="utf-8"><b style="font-weight:normal;" id="docs-internal-guid-fe8da6cd-7fff-ef59-c9b5-5b7aa1a62562"><div dir="ltr" style="margin-left:-40.5pt;" align="left">

Test Case
--
Permalink Preview - Short Text ✅  
Permalink Preview - Short Text - Hover Post  ✅

</div></b>
<meta charset="utf-8"><b style="font-weight:normal;" id="docs-internal-guid-f919ef58-7fff-e413-754a-a0336a325fa2"><div dir="ltr" style="margin-left:-40.5pt;" align="left">

Permalink Preview - Short Text - Hover Preview  ✅
Permalink Preview - Truncated ✅

</div></b>

<meta charset="utf-8"><b style="font-weight:normal;" id="docs-internal-guid-24a8e543-7fff-507b-0f8a-dc997c0ed91c"><div dir="ltr" style="margin-left:-40.5pt;" align="left">

Nested Permalink ✅ 
Permalink Preview - image🔴  **( FAILED )** --> **No Images in Preview**
Permalink Preview - Multiple Attachments🔴  **( FAILED )** **No Files in Preview**

</div></b>
<meta charset="utf-8"><b style="font-weight:normal;" id="docs-internal-guid-2e34ecb3-7fff-9402-8d7d-c23599403f80"><div dir="ltr" style="margin-left:-40.5pt;" align="left">

Permalink Preview - DM’s  ✅
Permalink Preview - Private channels ✅
Permalink Preview - Guest 🔴  **( FAILED )** --> **Guest can see and open Permalink Preview**

</div></b>
<meta charset="utf-8"><b style="font-weight:normal;" id="docs-internal-guid-b4854db6-7fff-cbfd-68d2-a4711939fa9a"><div dir="ltr" style="margin-left:-40.5pt;" align="left">

Permalink Preview - Public into Private Channel  ✅
Permalink Preview - Edited ✅
Permalink Preview -Deleted ✅

</div></b>

mkraft:
> Permalink Preview - image🔴 **( FAILED )** --> **No Images in Preview**

That is coming in a future iteration.

> Permalink Preview - Multiple Attachments🔴 **( FAILED )** **No Files in Preview**

Coming in a future iteration.

> Permalink Preview - Guest 🔴 **( FAILED )** --> **Guest can see and open Permalink Preview**

My understanding is that guests *should* see permalink previews if they have the appropriate authorization to the referenced post.

cc @esethna @BenCookie95

esethna:
> My understanding is that guests should see permalink previews if they have the appropriate authorization to the referenced post.

That's correct @mkraft @furqanmlk

esethna:
@furqanmlk to clarify your testing results: For the DM and private channel case, can you confirm that the observed behavior was that previews were shown to members that have access to the DM/GM/private channel?

furqanmlk:
> @furqanmlk to clarify your testing results: For the DM and private channel case, can you confirm that the observed behavior was that previews are shown to members that have access to the DM/GM/private channel?

Yes @esethna , that's correct, `Permalink Preview is shown to the member who have access to the DM/GM/Private Channels.`
[Test plan](https://docs.google.com/document/d/1HvfQKgABCKO9sXtGxcvykI4_ZZVT7IYFlvwTAvQ3d0M/edit#)

esethna:
@furqanmlk thanks, does this pass QA in that case? Please add your approval on all the PRs if so, thanks!

furqanmlk:
Hi @esethna 
All tests passed except one that I posted in other PR

https://github.com/mattermost/enterprise/pull/1005

Martin is looking at this. I will approve as soon as I hear from him.

mkraft:
> Hi @esethna
> All tests passed except one that I posted in other PR
> 
> [mattermost/enterprise#1005](https://github.com/mattermost/enterprise/pull/1005)
> 
> Martin is looking at this. I will approve as soon as I hear from him.

We're going to address the 2nd post created as a separate bug.

furqanmlk (approved):

PR #17799, opened 2021-06-18
MM-35297 Add the ability to control if thread is followed when marking post as unread and don't autofollow from API

jwilander:
#### Summary
The API endpoint for setting posts unread was only being used by the client for posts marked unread from the center channel, in which case the thread should not be followed. However, if a thread membership didn't exist we were setting it to followed automatically so in some instances a user marking a post unread from the center channel would seemingly follow and unfollowed thread. I added the option to specify in `MarkChannelAsUnreadFromPost` what the behaviour should be and set the API endpoint to not follow the threads.

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-35297

#### Release Note

```release-note
NONE
```

### DIFF
--- a/api4/post_test.go
+++ b/api4/post_test.go
@@ -2547,33 +2547,6 @@ func TestSetChannelUnread(t *testing.T) {
 	})
 }
 
-func TestMarkUnreadCausesAutofollow(t *testing.T) {
-	os.Setenv("MM_FEATUREFLAGS_COLLAPSEDTHREADS", "true")
-	defer os.Unsetenv("MM_FEATUREFLAGS_COLLAPSEDTHREADS")
-	th := Setup(t).InitBasic()
-	defer th.TearDown()
-	th.App.UpdateConfig(func(cfg *model.Config) {
-		*cfg.ServiceSettings.ThreadAutoFollow = true
-		*cfg.ServiceSettings.CollapsedThreads = model.COLLAPSED_THREADS_DEFAULT_ON
-	})
-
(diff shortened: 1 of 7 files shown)


api4/post_test.go
```diff
@@ -2547,33 +2547,6 @@ func TestSetChannelUnread(t *testing.T) {
 	})
 }
 
-func TestMarkUnreadCausesAutofollow(t *testing.T) {
```
jwilander, line 2550: Moved this since it wasn't testing the api, only the app

app/channel.go
```diff
@@ -2541,6 +2541,15 @@ func (a *App) MarkChannelAsUnreadFromPost(postID string, userID string, collapse
...
 				a.Publish(message)
 			}
+		} else if !threadMembership.Following && followThread {
+			opts := store.ThreadMembershipOpts{
+				Following:       true,
+				UpdateFollowing: true,
+			}
+			_, storeErr = a.Srv().Store.Thread().MaintainMembership(user.Id, threadId, opts)
+			if storeErr != nil && !errors.As(storeErr, &nfErr) {
+				return nil, model.NewAppError("MarkChannelAsUnreadFromPost", "app.channel.update_last_viewed_at_post.app_error", nil, storeErr.Error(), http.StatusInternalServerError)
```
streamer45, line 2551: Slightly unrelated but the error id used in this method doesn't look right. Maybe it got moved around.

streamer45 (approved):
Looks good :+1:

app/channel.go
jwilander: You're right, id doesn't look right but the text it translate to works for this scenario so it's probably ok. Looking forward to us removing a lot of these translations

ashishbhate (approved):

PR #17800, opened 2021-06-18
MM-36430 Follow threads without memberships if marked as unread

jwilander:
#### Summary
If the thread wasn't participated in or followed before (aka no thread membership existed) we were erroring out when the thread was marked as unread. This fixes it to create the thread membership if needed in that case.

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-36430

#### Release Note

```release-note
NONE
```


koox00 (approved):

ashishbhate (approved):

jwilander:
Same for this one @jgilliam17

PR #17803, opened 2021-06-20
[MM-35345][MM-35494] fixes for incorrect mentions and unreads for threads and channels

ashishbhate:
#### Summary

The CRT backend was first released in version 5.29.0. Since then, the behaviour of the CRT feature has been refined, several bugs have been fixed, and a few sql columns have been added. Before these various fixes went in, the threads and channel membership tables have accumulated incorrect mention and unreads data. This PR fixes some of this bad historical data.
Summary of fixes:
-  Marks threads as read for users where the last reply time of the thread is earlier than the time the user viewed the channel.  Marking a thread means setting the mention count to zero and setting the last viewed at time of the the thread as the last viewed at time of the channel. This is done through a "sql migration"
- Fix channel counts, i.e. the total message count, total root message count, mention count, and mention count in root messages for users who have viewed the channel after the last post in the channel. This is done as a "sql migration"

#### Ticket Link

Parts of
- https://mattermost.atlassian.net/browse/MM-35494
- https://mattermost.atlassian.net/browse/MM-35345
 
#### Release Note

```release-note
 - Fixes some of the incorrect mention counts and unreads around threads and channels since the introduction of CRT. This fix is done through a SQL migration and it may take several minutes to complete for large databases.
 - Add to important upgrade notes: https://github.com/mattermost/mattermost-server/pull/17803#issuecomment-879007420
```


store/sqlstore/thread_store.go
```diff
@@ -52,7 +52,7 @@ func threadToSlice(thread *model.Thread) []interface{} {
 		thread.ChannelId,
 		thread.LastReplyAt,
 		thread.ReplyCount,
-		thread.Participants,
+		model.ArrayToJson(thread.Participants),
```
ashishbhate, line 55: `Thread.Save` was broken because `Participants` (model.StringArray) wouldn't serialize properly.
We didn't actually use `Thread.Save` anywhere so we never noticed.
We use the `Exec` method to hack around gorp not supporting inserting multiple rows in a single transaction,
but `Exec` method doesn't call the `TypeConverter`, so here we explicitly convert it to the right format.

store/sqlstore/upgrade.go
```diff
@@ -1211,3 +1212,178 @@ func upgradeDatabaseToVersion537(sqlStore *SqlStore) {
...
+			FROM cte3 WHERE cte3.UserId = ChannelMembers.UserId AND ChannelMembers.ChannelId = cte3.ChannelId and '' != ALL (cte3.Typ)
+		`
+		if _, err := sqlStore.GetMaster().ExecNoTimeout(systemMessagesFixQuery); err != nil {
+			mlog.Error("Error updating unreads caused by join leave messages", mlog.Err(err))
+		}
+		return
+	}
+
+	if sqlStore.DriverName() == model.DATABASE_DRIVER_MYSQL {
+		trans, err := sqlStore.GetMaster().Begin()
```
ashishbhate, line 1335: Temporary tables only stick around for a session (connection). We use a transaction to ensure everything happens in the same session.
We could probably use joins instead of temporary tables, but I think this is much neater and easier to understand.

jwilander:
@streamer45 @jespino if possible could you guys review this today?

store/sqlstore/upgrade.go
```diff
@@ -1208,6 +1209,201 @@ func upgradeDatabaseToVersion537(sqlStore *SqlStore) {
...
+// 2. Marks threads as read for users where the last reply time of the thread
+//    is earlier than the time the user viewed the channel.
+//    Marking a thread means setting the mention count to zero and
+//    setting the last viewed at time of the the thread as the last viewed at time
+//    of the channel
+func fixCRTThreadCountsAndUnreads(sqlStore *SqlStore) {
+	countsAndLastReplyCTE := `
+		SELECT RootId, COUNT(Id) AS FixedReplyCount, max(Posts.CreateAt) as FixedLastReply
+		FROM Posts where Posts.DeleteAt = 0
+		GROUP BY RootId
```
jwilander, line 1231: That's one massive read, right? Counting all rows of posts for every thread? If there's millions of posts in the DB this might take a long time. Can we make it so we only count the replies for root ids that we care about instead of every single root id?

store/sqlstore/upgrade.go
```diff
@@ -1208,6 +1209,201 @@ func upgradeDatabaseToVersion537(sqlStore *SqlStore) {
...
+			ON Threads.PostId = q.RootId
+			SET ReplyCount = q.FixedReplyCount, LastReplyAt = q.FixedLastReply
+		`
+	}
+
+	participantsCTE := `
+		SELECT COALESCE(NULLIF(RootId,''), Id) AS ThreadId, ARRAY_AGG(DISTINCT UserId) FixedParticipants
+		FROM Posts
+		WHERE Posts.DeleteAt = 0
+		GROUP BY ThreadId
```
jwilander, line 1254: Same here, looks like we're grabbing participants for all threads instead of just the ones we care about?

store/sqlstore/upgrade.go
```diff
@@ -1208,6 +1209,201 @@ func upgradeDatabaseToVersion537(sqlStore *SqlStore) {
...
+	now := fmt.Sprintf("%d", model.GetMillis())
+
+	if sqlStore.DriverName() == model.DATABASE_DRIVER_POSTGRES {
+		systemMessagesFixQuery := `
+			WITH
+			cte1 AS (SELECT ChannelId, UserId, LastViewedAt FROM ChannelMembers INNER JOIN Channels ON ChannelMembers.ChannelId = Channels.Id WHERE ChannelMembers.MsgCountRoot != Channels.TotalMsgCountRoot),
+			cte2 AS (SELECT ARRAY_AGG(DISTINCT Posts.Type) AS Typ, cte1.UserId, cte1.ChannelId FROM cte1 INNER JOIN Posts ON Posts.Channelid = cte1.ChannelId WHERE Posts.Createat > cte1.LastViewedAt GROUP BY cte1.UserId, cte1.ChannelId),
+			cte3 AS (SELECT cte2.*, TotalMsgCountRoot, TotalMsgCount FROM cte2 INNER JOIN Channels on cte2.ChannelId = Channels.Id)
+			UPDATE ChannelMembers SET MsgCountRoot = cte3.TotalMsgCountRoot, MsgCount = cte3.TotalMsgCount, LastUpdateAt = ` + now + `
+			FROM cte3 WHERE cte3.UserId = ChannelMembers.UserId AND ChannelMembers.ChannelId = cte3.ChannelId and '' != ALL (cte3.Typ)
```
jwilander, line 1342: Can you write a comment for this one that describes what it's doing? It's a little hard to parse

jwilander:
> I'm not sure, but I think @streamer45 is on PTO. I'll do my best to review it today.

Ah, didn't know that. @agnivade are you able to give this a look instead?

agnivade:
This is marked with `Perf Review`. Does that mean it is intended to run through a load-test? Then I'm afraid I won't be able to finish it by today. Otherwise, doing the code review now.

store/sqlstore/upgrade.go
agnivade: Agree. Not even considering the query time, it has to scan the _entire_ posts table without any index and group it. This has the potential to cause an OOM.

I would recommend to process in batches. Use a cursor for the post id, and move the cursor every batch. That's much more efficient. It uses the PK index, and processes small batches in memory.

The `LastReplyAt` is a max field, so it shouldn't matter if a root_id group falls in 2 batches. The `ReplyCount` is a bit tricky. One way might be to set the field to zero initially, and then increment it in every batch.

store/sqlstore/upgrade.go
```diff
@@ -1208,6 +1209,201 @@ func upgradeDatabaseToVersion537(sqlStore *SqlStore) {
...
+	if sqlStore.DriverName() == model.DATABASE_DRIVER_MYSQL {
+		updateThreadMembershipQuery = `
+			UPDATE ThreadMemberships
+			INNER JOIN (` + threadMembershipsCTE + `) as q
+			ON ThreadMemberships.Postid = q.PostId AND ThreadMemberships.UserId = q.UserId
+			SET LastViewed = q.CM_LastViewedAt, UnreadMentions = 0, LastUpdated =` + now + `
+		`
+	}
+
+	if _, err := sqlStore.GetMaster().ExecNoTimeout(updateReplyCountsQuery); err != nil {
```
agnivade, line 1292: Suggest to move the exec code to just below the SQL queries to keep it readable. 0/5.

store/sqlstore/upgrade.go
agnivade: Yep, same. Need to batch.

store/sqlstore/upgrade.go
```diff
@@ -1208,6 +1209,201 @@ func upgradeDatabaseToVersion537(sqlStore *SqlStore) {
...
+	}
+}
+
+// fixCRTChannelMembershipCounts fixes the channel counts, i.e. the total message count,
+// total root message count, mention count, and mention count in root messages for users
+// who have viewed the channel after the last post in the channel
+func fixCRTChannelMembershipCounts(sqlStore *SqlStore) {
+	now := fmt.Sprintf("%d", model.GetMillis())
+
+	channelMembershipsCountsAndMentions := `
```
agnivade, line 1309: We need to parameterize this and the query below to avoid SQL injections. I know the input is in our control, but it's a good practice and keeps the code clean.

store/sqlstore/upgrade.go
```diff
@@ -1208,6 +1209,201 @@ func upgradeDatabaseToVersion537(sqlStore *SqlStore) {
...
+		if _, err := sqlStore.GetMaster().ExecNoTimeout(systemMessagesFixQuery); err != nil {
+			mlog.Error("Error updating unreads caused by join leave messages", mlog.Err(err))
+		}
+		return
+	}
+
+	if sqlStore.DriverName() == model.DATABASE_DRIVER_MYSQL {
+		trans, err := sqlStore.GetMaster().Begin()
+		if err != nil {
+			mlog.Critical("Error starting transaction", mlog.Err(err))
```
agnivade, line 1353: We need to return from here. Also use the `defer finalizeTransaction(trans)` helper to properly rollback a transaction if it caused an error. See other examples in the store.

store/sqlstore/upgrade.go
```diff
@@ -1208,6 +1209,201 @@ func upgradeDatabaseToVersion537(sqlStore *SqlStore) {
...
+			INNER JOIN cte3 ON cte3.UserId = ChannelMembers.UserId AND ChannelMembers.ChannelId = cte3.ChannelId
+			SET MsgCountRoot = cte3.TotalMsgCountRoot, MsgCount = cte3.TotalMsgCount, LastUpdateAt = ` + now + `
+			WHERE FIND_IN_SET("", cte3.typ) = 0 && cte3.typ != '';
+		`
+		if _, err = trans.ExecNoTimeout(systemMessagesFixQuery); err != nil {
+			mlog.Error("Error updating unreads caused by join leave messages", mlog.Err(err))
+		}
+		return
+	}
+	mlog.Error("Error unknown DB type")
```
agnivade, line 1408: No need for this. The config validation only support MySQL and Postgres DB types.

store/sqlstore/upgrade.go
```diff
@@ -1208,6 +1209,201 @@ func upgradeDatabaseToVersion537(sqlStore *SqlStore) {
...
+
+	if _, err := sqlStore.GetMaster().ExecNoTimeout(channelMembershipsCountsAndMentions); err != nil {
+		mlog.Error("Error updating counts and unreads for channelmemberships", mlog.Err(err))
+	}
+}
+
+// fixCRTChannelUnreadsForJoinLeaveMessages marks channels as read for users
+// if there are only system posts after the last time the user viewed the channel
+func fixCRTChannelUnreadsForJoinLeaveMessages(sqlStore *SqlStore) {
+	now := fmt.Sprintf("%d", model.GetMillis())
```
agnivade, line 1333: Same as before, we need to parameterize.

store/sqlstore/upgrade.go
```diff
@@ -1208,6 +1209,201 @@ func upgradeDatabaseToVersion537(sqlStore *SqlStore) {
...
+
+	if sqlStore.DriverName() == model.DATABASE_DRIVER_POSTGRES {
+		systemMessagesFixQuery := `
+			WITH
+			cte1 AS (SELECT ChannelId, UserId, LastViewedAt FROM ChannelMembers INNER JOIN Channels ON ChannelMembers.ChannelId = Channels.Id WHERE ChannelMembers.MsgCountRoot != Channels.TotalMsgCountRoot),
+			cte2 AS (SELECT ARRAY_AGG(DISTINCT Posts.Type) AS Typ, cte1.UserId, cte1.ChannelId FROM cte1 INNER JOIN Posts ON Posts.Channelid = cte1.ChannelId WHERE Posts.Createat > cte1.LastViewedAt GROUP BY cte1.UserId, cte1.ChannelId),
```
agnivade, line 1339: Unfortunately this is joining the posts table with the previous CTE. It doesn't look like very risky, but we need to test this one.

agnivade (commented):
I have left some suggestions, but we need to test with a data set of 10M posts and threads to get a representative estimate of the impact of this.

esethna:
FYI for those reviewing this: We discussed with the CRT team that we're going to prioritize shipping the SQL migrations that Ashish listed in the PR description and hold off on the more risky and performance heavy app migration piece.

image

Please continue to code review this PR and then Ashish will split out the above migrations in a new PR for the beta release in v5.37. 

cc// @amyblais @agnivade @jwilander @ashishbhate @jespino

app/migrations.go
```diff
@@ -323,8 +326,85 @@ func (s *Server) doContentExtractionConfigDefaultTrueMigration() {
...
+	var appErr *model.AppError
+	var ok bool
+	var prevThreadId string
+	for _, t := range newerThreadsInfo {
+		// We could potentially have hundreds of thousands of threads
+		// GetAllThreadsNewerThanChannelLastViewedAt orders by ThreadId
+		// This way we don't have to fetch the root post for every iteration
+		// and we don't have to maintain a huge map of known root posts
+		if t.ThreadId != prevThreadId {
+			rootPost, appErr = a.GetSinglePost(t.ThreadId)
```
jespino, line 354: I think it still is going to generate a ton of queries

jwilander:
For anyone following this PR see https://community.mattermost.com/core/pl/5cfw75qthbbc9ehwyo1khmq6cw for latest update

ashishbhate:
@jwilander @agnivade @jespino 

After a lot testing it turns out that the SQL queries for some of the fixes are far too heavy to run. There wasn't much room for optimization either. I've updated this PR to only include the quicker fixes. The other heavier fixes will be implemented as an async job in a separate PR.

agnivade (approved):

jwilander (approved):
Good work 👍

esethna:
@ashishbhate what's next steps to get this merged and how should we QA?

cc// @jgilliam17 @ogi-m

ashishbhate:
@esethna 
This PR is good to go. I'd like to wait a day to give @jespino a chance to take a look. Ideally this PR should be merged after https://mattermost.atlassian.net/browse/MM-36737 is done, but its not a blocker. AFAIK, "daily" is deployed multiple times a day, and until release the migrations will run on every deployment to daily. 

There isn't much to QA here. It doesn't affect new threads. Its a one-time fix for older threads. To test this we'll need to set-up a server on an older version of the app with bad data in the DB, and then upgrade to this version. The upgrade process will fix the bad data. The PR has tests which essentially does what I've described above, so we should be good to merge this.

store/sqlstore/upgrade.go
```diff
@@ -1211,3 +1213,69 @@ func upgradeDatabaseToVersion537(sqlStore *SqlStore) {
 		saveSchemaVersion(sqlStore, Version5370)
 	}
 }
+
+func upgradeDatabaseToVersion538(sqlStore *SqlStore) {
+	// TODO: uncomment when the time arrive to upgrade the DB for 5.38
+	// if shouldPerformUpgrade(sqlStore, Version5370, Version5380) {
+	fixCRTChannelMembershipCounts(sqlStore)
+	fixCRTChannelMembershipCounts(sqlStore)
```
streamer45, line 1221: Do we need to call this twice?

streamer45 (commented):
Finally getting to read this and only have one outstanding question: do we have a sense of how long this migration will take on big datasets? (e.g. time per million rows). 
I understand load-tests were run so just wondering if anyone has any performance data to share.

store/sqlstore/upgrade.go
ashishbhate: >Do we need to call this twice?

Definitely not. Fixed.

ashishbhate:
> I understand load-tests were run so just wondering if anyone has any performance data to share.

I can definitely share:

1.
Using a DB I got from the performance team: 
The fixCRTChannelMembershipCounts fix takes 1 minute and 20 seconds for a DB containing approximately 4 million channel memberships and about 130,000 channels. This is on mysql 5.6.51

2.
The DB I got from the performance team didn't have thread memberships so I ran a load test for a few days to generate a large-ish DB with thread memberships.

The fixCRTThreadCountsAndUnreads fix takes about 3 minutes and 30 seconds for a DB containing 56367 threads, 124587 thread memberships, and 220801 channel memberships. This is also on mysql 5.6.51.

streamer45:
> Using a DB I got from the performance team:
> The fixCRTChannelMembershipCounts fix takes 1 minute and 20 seconds for a DB containing approximately 4 million channel memberships and about 130,000 channels. This is on mysql 5.6.51
> 
> 
> The DB I got from the performance team didn't have thread memberships so I ran a load test for a few days to generate a large-ish DB with thread memberships.
> 
> The fixCRTThreadCountsAndUnreads fix takes about 3 minutes and 30 seconds for a DB containing 56367 threads, 124587 thread memberships, and 220801 channel memberships. This is also on mysql 5.6.51.

Awesome, thanks. It would be good to mention this in important upgrade notes for the 5.38 release. cc @amyblais

jgilliam17:
Thanks @ashishbhate QA can test after merge. 
I already have a bit of "bad data" on daily that this PR should clear up.
Screen Shot 2021-07-13 at 8 55 48 AM

amyblais:
> > Using a DB I got from the performance team:
> > The fixCRTChannelMembershipCounts fix takes 1 minute and 20 seconds for a DB containing approximately 4 million channel memberships and about 130,000 channels. This is on mysql 5.6.51
> > The DB I got from the performance team didn't have thread memberships so I ran a load test for a few days to generate a large-ish DB with thread memberships.
> > The fixCRTThreadCountsAndUnreads fix takes about 3 minutes and 30 seconds for a DB containing 56367 threads, 124587 thread memberships, and 220801 channel memberships. This is also on mysql 5.6.51.
> 
> Awesome, thanks. It would be good to mention this in important upgrade notes for the 5.38 release. cc @amyblais

I added a note in the Release Note section in the PR description.

PR #17804, opened 2021-06-20
[MM-18496] Private channels should appear in channel link autocomplete

arvinDarmawan:
#### Summary
This code will fix issue private channel not showing in channel link autocomplete

#### Jira Card
https://mattermost.atlassian.net/browse/MM-18496

#### Ticket Link
Fixes https://github.com/mattermost/mattermost-server/issues/16245

```release-note
Release Note
- Add param userId  (current user id in session) on function AutocompleteChannels and AutocompleteInTeam
- Fix testing scenario that related with changes on function AutocompleteChannels and AutocompleteInTeam
```


store/storetest/channel_store.go
```diff
@@ -5173,8 +5173,8 @@ func testChannelStoreSearchInTeam(t *testing.T, ss store.Store) {
 	}
 
 	for name, search := range map[string]func(teamId string, term string, includeDeleted bool) (*model.ChannelList, error){
-		"AutocompleteInTeam": ss.Channel().AutocompleteInTeam,
-		"SearchInTeam":       ss.Channel().SearchInTeam,
+		// "AutocompleteInTeam": ss.Channel().AutocompleteInTeam,
+		"SearchInTeam": ss.Channel().SearchInTeam,
```
arvinDarmawan, line 5177: Hi, for SearchInTeam, did we need to add userId too ?

api4/channel.go
```diff
@@ -916,7 +916,7 @@ func autocompleteChannelsForTeam(c *Context, w http.ResponseWriter, r *http.Requ
 
 	name := r.URL.Query().Get("name")
 
-	channels, err := c.App.AutocompleteChannels(c.Params.TeamId, name)
+	channels, err := c.App.AutocompleteChannels(c.Params.TeamId, c.AppContext.Session().UserId, name)
```
arvinDarmawan, line 919: Hi, I add currentUserId in this function.

store/sqlstore/channel_store.go
```diff
@@ -2425,48 +2425,64 @@ func (s SqlChannelStore) GetMembersForUserWithPagination(teamId, userId string,
 	return dbMembers.ToModel(), nil
 }
 
-func (s SqlChannelStore) AutocompleteInTeam(teamId string, term string, includeDeleted bool) (*model.ChannelList, error) {
-	deleteFilter := "AND Channels.DeleteAt = 0"
+func (s SqlChannelStore) AutocompleteInTeam(teamId string, userId string, term string, includeDeleted bool) (*model.ChannelList, error) {
```
arvinDarmawan, line 2428: hi, in this function, I parse userId as parameter.
Is it the correct way @agnivade  ?

arvinDarmawan: because previous code didn;t handle private channel (only public channel)

store/storetest/channel_store.go
arvinDarmawan: cc: @agnivade

store/sqlstore/channel_store.go
agnivade: Yes, this is correct.

agnivade:
Can you check the `(s SqlChannelStore) SearchForUserInTeam` method? It looks like it is already doing what you are trying to achieve here.

arvinDarmawan:
> Can you check the `(s SqlChannelStore) SearchForUserInTeam` method? It looks like it is already doing what you are trying to achieve here.
Yes @agnivade , it have parameter user id too. But this function use inner join with public channel, so I think it will not display private channel.

Screen Shot 2021-06-23 at 11 33 23

agnivade:
Ah yes, that's correct.

store/storetest/channel_store.go
agnivade: I don't believe you need to add in `SearchInTeam`. You can use the same function for `AutocompleteInTeam`, just modify the test case to include private channels too.

arvinDarmawan:
> > Can you check the `(s SqlChannelStore) SearchForUserInTeam` method? It looks like it is already doing what you are trying to achieve here.
> > Yes @agnivade , it have parameter user id too. But this function use inner join with public channel, so I think it will not display private channel.
> 
> Screen Shot 2021-06-23 at 11 33 23

Did I need to change this code too, @agnivade ? or it's enough on AutocompleteInTeam function and its related code ?

arvinDarmawan:
Hi @agnivade, I have a question.

In my local mattermost, my code already working as expected.
Screen Shot 2021-06-30 at 20 44 54

But, I'am confused because it didn't worked as expected in my test case (searchChannelStoreTests).

If I test it using  `make test-server`, it will return several errors.
This is one of the error : 
Screen Shot 2021-06-30 at 09 15 46

From this error, we can see that list A have private channel Id declared but list B (from function autocompletion) didn't have private channel id. 
But if I remove private channel Id from list A and rerun the test case again, it  will throw same error with different reason.
List B (from function autocompletion)  will have private channel id and list A didn't have it (since I removed private channel id before).


The number of errors from test-mysql-1 and test-postgres-1  seems different too (please ignore error from TestSendAutoResponseIfNecessary).

Do you have any suggestion how to debug this test case ?
And since run `make test-server` will run every test case in mattermost-server and it take quite a lot of time,
 can  we run single test case file in mattermost-server (only searchChannelStoreTests ) ? 


Thank youu.

agnivade:
You can run a single test manually in this way: `go test -v -run=TestSearchChannelStore ./store/sqlstore/`. You can also use the [dlv](https://github.com/go-delve/delve/) debugger to debug a test.

arvinDarmawan:
> You can run a single test manually in this way: `go test -v -run=TestSearchChannelStore ./store/sqlstore/`. You can also use the [dlv](https://github.com/go-delve/delve/) debugger to debug a test.

Hi @agnivade , thank you for the suggestion. 
I already test it on one scenario "TestSearchChannelStore/MySQL/Should_be_able_to_autocomplete_a_channel_by_name".
I think I found the issue.

So in function SetupBasicFixtures(), we define this code to create private channel with name "channel-private".
```
channelPrivate, err := th.createChannel(team.Id, "channel-private", "ChannelPrivate", "", model.CHANNEL_PRIVATE, false)
	if err != nil {
		return err
	}
```

In scenario "TestSearchChannelStore/MySQL/Should_be_able_to_autocomplete_a_channel_by_name", we try to call function AutocompleteInTeam with term "channel-a".

Using bleve engine, in PostgreSQL if we search term "channel-a",  word "channel-private" will be shown in query but in
MySQL,  it didn't. This condition will  make an error on test case for  MySQL, but in PostgreSQL it will be success.

This is the result if I use default code using name "channel-private" in function SetupBasicFixtures():
Screen Shot 2021-06-30 at 22 20 23



This is the result if I change the private channel name become "**channel-aprivate**" in function SetupBasicFixtures() and run the test again:
Screen Shot 2021-06-30 at 22 23 15

No error in PostgreSQL and MySQL.


So, can I change the default **private channel** name / description in function SetupBasicFixtures() or do you have any suggestion?

agnivade:
Can you look into why it returns `channel-private` when you search for `channel-a`? Only channels starting with the prefix `channel-a` should be returned right?

arvinDarmawan:
> Can you look into why it returns `channel-private` when you search for `channel-a`? Only channels starting with the prefix `channel-a` should be returned right?

ok, let me check it first

arvinDarmawan:
Hi @agnivade ,  in PostgreSQL when i search using term `channel-a`, it will return `channel private`.
It is because the current implementation of full text search in PostgreSQL.

When we call function buildFulltextClause, it will split term `channel-a`  and rejoin it to be `Channel:* & A:*`
After that, this new term `Channel:* & A:*` will be used to create new full text clause PostgreSQL like below:
`fulltextClause = fmt.Sprintf("((to_tsvector('english', %s)) @@ to_tsquery('english', :FulltextTerm))", convertMySQLFullTextColumnsToPostgres(searchColumns))`

But if I set the full text clause to be like code below, it will return correct result (search using term `channel-a`,  `channel private` will not returned):
`fulltextClause = fmt.Sprintf("((to_tsvector('english', %s)) @@ to_tsquery('simple', :FulltextTerm))", convertMySQLFullTextColumnsToPostgres(searchColumns))`

What do you think ?

agnivade:
Great, thanks for digging into it. The search code is a bit sensitive and we've had issues tinkering with the filters. Let's just add a small check for the db type and change our expectations accordingly in the test code.

arvinDarmawan:
> Great, thanks for digging into it. The search code is a bit sensitive and we've had issues tinkering with the filters. Let's just add a small check for the db type and change our expectations accordingly in the test code.

hi @agnivade , so i didn't need to change the code when do full text search using PostgreSQL, right ?

So, I just need to update the error test case for not testing it using PostgreSQL  like code below?
`Tags: []string{EngineMySql, EngineElasticSearch, EngineBleve},`

agnivade:
Yes, we don't need to change the code for full text search. But we _do_ want to test the PostgreSQL case by modifying the test expectations to the correct ones for Postgres.

You can add an if condition to check if search engine is postgres, then expect an additional result item.

arvinDarmawan:
> Yes, we don't need to change the code for full text search. But we _do_ want to test the PostgreSQL case by modifying the test expectations to the correct ones for Postgres.
> 
> You can add an if condition to check if search engine is postgres, then expect an additional result item.

ok, I will try it

arvinDarmawan:
Hi @agnivade , I code it like this : 
```
func testAutocompleteChannelByName(t *testing.T, th *SearchTestHelper) {
	alternate, err := th.createChannel(th.Team.Id, "channel-alternate", "Channel Alternate", "Channel Alternate", model.CHANNEL_OPEN, false)
	require.NoError(t, err)
	defer th.deleteChannel(alternate)
	res, err := th.Store.Channel().AutocompleteInTeam(th.Team.Id, th.User.Id, "channel-a", false)
	require.NoError(t, err)

	driverName := th.Store.GetDatabaseDriverName()
	fmt.Println("driverName", driverName)
	if driverName == model.DATABASE_DRIVER_POSTGRES {
		th.checkChannelIdsMatch(t, []string{th.ChannelBasic.Id, th.ChannelPrivate.Id, alternate.Id}, res)
	} else {
		th.checkChannelIdsMatch(t, []string{th.ChannelBasic.Id, alternate.Id}, res)
	}
}
```

So I create new function GetDatabaseDriverName in store/sqlstore.go

```
func (ss *SqlStore) GetDatabaseDriverName() string {
	return ss.DriverName()
}

```

The test result is good. Do you have any input or review ?
If it is already ok, I will try to implement it on another test case.

arvinDarmawan:
Hi @agnivade , I already fix  all the test case.
I already test it too using `go test -v -run=TestSearchChannelStore ./store/sqlstore/`

Screen Shot 2021-07-02 at 21 33 29

But somehow the pipeline still showing  error after I push it to github.
Do you have any suggestion ?

agnivade:
I believe it's a different set of tests that are failing - `TestBleveEngineTestSuite/TestBleveSearchStoreTests`. Perhaps you'll need to make some modifications there too.

arvinDarmawan:
Hi @agnivade , I already fix the issue failed test case blevesearch. It is because when create  channel index in blevesearch, it is only indexing the public channel only.   You can see my changes in file store/searchlayer/channel_layer.go (line 39).

But this makes me wonder.
This github issue need to solve the issue  for private channel  not appear, right ?
The current PR already handle postgre and mysql database, but not for searchengine.

But, for search engine like blevesearch and elasticsearch, do we need to add user id when reindex channel to searchengine ?
So, we can send additional parameter current user id to search engine to check if current user has access to private channel (I need to check it first if it's possible to get public and private channel (according to current user id) from blevesearch.
This is to handle conditions where admin activate experimental feature Bleve in mattermost admin console.

agnivade:
Yes, I think we need to do that. Ahh I just realized that the elasticsearch code is in the enterprise repo.

Ok, we can go ahead with doing this only for MySQL and Postgres for now. You can add support for Bleve if you want. Or do it in a later PR if you prefer. You'll need to modify the `func (b *BleveEngine) IndexChannel` method and the `BLVChannelFromChannel` function inside it to also have a `UserId` field in the `BLVChannel` struct.

For elasticsearch functionality, a staff member has to do it later.

/cc FYI on this @esethna

arvinDarmawan:
> Yes, I think we need to do that. Ahh I just realized that the elasticsearch code is in the enterprise repo.
> 
> Ok, we can go ahead with doing this only for MySQL and Postgres for now. You can add support for Bleve if you want. Or do it in a later PR if you prefer. You'll need to modify the `func (b *BleveEngine) IndexChannel` method and the `BLVChannelFromChannel` function inside it to also have a `UserId` field in the `BLVChannel` struct.
> 
> For elasticsearch functionality, a staff member has to do it later.
> 
> /cc FYI on this @esethna

Hi @agnivade , it's fine. I will add support for Bleve in this PR too. I will check it tonight.

arvinDarmawan:
> Yes, I think we need to do that. Ahh I just realized that the elasticsearch code is in the enterprise repo.
> 
> Ok, we can go ahead with doing this only for MySQL and Postgres for now. You can add support for Bleve if you want. Or do it in a later PR if you prefer. You'll need to modify the `func (b *BleveEngine) IndexChannel` method and the `BLVChannelFromChannel` function inside it to also have a `UserId` field in the `BLVChannel` struct.
> 
> For elasticsearch functionality, a staff member has to do it later.
> 
> /cc FYI on this @esethna

Hi @agnivade , correct me if i'm wrong.
So, when indexing channel, for every public channel we didn't need to insert UserIds, but for every private channel, we need to insert UserIds (array with more than one channel member).

```
type BLVChannel struct {
	Id          string
	TeamId      []string
         UserIds  []string
	NameSuggest []string
}
```

And then when query using bleve for private channel, i need to check if current user id is exists in property UserIds.
Is this implementation good or do you have any suggestion ?

agnivade:
Oops yeah, my bad there. Yes, we need a slice of userids.

> To get UserIds, I will call function GetMembers(channelID string, offset, limit int) from channel_store.go (maybe using limit 100) or maybe I will create new function GetAllMembers(channelID) which will return all channel members by channel id.

You mean to freshly index all the data right? I would lean on advice from @mgdelacroix on this one. Miguel - what would be the right way to populate a new field in the bleve index for channels.

> Hmm, but if we set UserIds on BLVChannelFromChannel, every time we add / update member to channel, we not only need to call indexUserFromID (to index the user data to bleve) but also need to call indexChannel.

That is absolutely correct. We will have a two-way inverted index. User index containing channels, and a Channel index containing users. Search indexes are usually denormalized, so I think it should be fine.

@mgdelacroix - Any guidance would be appreciated here. We are trying to index the userIDs to a channel because we need to search for private channels. Does the above approach look correct to you?

esethna:
@agnivade can you help open a ticket for the elasticsearch work and route it to the appropriate team sp it doesn't get lost?

agnivade:
Yep - https://mattermost.atlassian.net/browse/MM-36867

mgdelacroix:
What I would suggest is to add both the `UserIds` slice and the channel `Type` as well, and then modify the search query to either look for Open channels and match by name, team, etc. or look for Private channels with the user in the slice and then match by the same conditions.

To add a field to the index, check the `services/searchengine/bleveengine/bleve.go` file; you'll have to add the field to the index itself and then modify the conversion methods and the structs they reference (`BLVChannelFromChannel`) as you already mentioned.

arvinDarmawan:
> What I would suggest is to add both the `UserIds` slice and the channel `Type` as well, and then modify the search query to either look for Open channels and match by name, team, etc. or look for Private channels with the user in the slice and then match by the same conditions.
> 
> To add a field to the index, check the `services/searchengine/bleveengine/bleve.go` file; you'll have to add the field to the index itself and then modify the conversion methods and the structs they reference (`BLVChannelFromChannel`) as you already mentioned.

So, we will do it using this way, right , @mgdelacroix @agnivade  ?

agnivade:
Yes.

esethna:
@arvinDarmawan how's this coming? Anything we can help with? Really looking forward to this feature!

agnivade:
@arvinDarmawan - Just checking up on this. Are you still planning to work on this? 

We have a productivity sprint coming and I would be happy to take over this and wrap this up along with the Elasticsearch changes.

arvinDarmawan:
Hi @agnivade sorry for my late response. Looks like I don't have time to working on blevesearch since lately I have quite a lot of work to do. But I think the current code already handle MySQL and Postgres for now.

agnivade:
@arvinDarmawan - Yep it does support MySQL and Postgres. However I think it can be improved slightly: the SQL query can be done in a single query and there are some merge conflicts to resolve.

I will be sure to credit your work in the commit so that it doesn't get lost. Thank you for your time @arvinDarmawan ! I will go ahead and close this one out.

PR #17806, opened 2021-06-21
MM-36590 Re-add old emoji aliases

hmhealey:
Server-side version of https://github.com/mattermost/mattermost-webapp/pull/8287

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-36590

#### Related Pull Requests
https://github.com/mattermost/mattermost-webapp/pull/8287

#### Release Note
```release-note
NONE
```


Willyfrog (approved):

hmhealey:
I fixed the unit test that was failing. I'm slightly worried that test might be flaky since I'm pretty sure we're using a for loop to iterate over the map stored in emoji_data which is non-deterministic, but I ran the tests a bunch of times locally, and they seemed to pass.

jwilander (approved):

hmhealey:
For testing this, the only things the server uses the old emoji names for are preventing the user from creating custom emojis with the same name and for translating unicode emojis used with the /custom_status command into the text form. I don't know if the latter is changed by this, but the custom emojis code can certainly be tested.

Also, I double checked and the test server on this PR is using the correct branch from the webapp. 🎉

jwilander:
@hmhealey if you're feeling confident in it I'm ok to merge this and defer QA testing until it's on daily. Assuming @jgilliam17 is ok with that. Thoughts?

Same goes for the webapp one

hmhealey:
Yeah, I'd be comfortable with that. It's only a data change, so it's less risky than even changing code.

jwilander:
Cool, feel free to merge :)

PR #17821, opened 2021-06-22
MM-32421: Extend channel sidebar category APIs for Plugins

madhavhugar:
#### Summary

Add CreateChannelSidebarCategory, GetChannelSidebarCategories and UpdateChannelSidebarCategories to Plugin API

#### Ticket Link

JIRA: https://mattermost.atlassian.net/browse/MM-32421

#### Release Note

```release-note
Added CreateChannelSidebarCategory, GetChannelSidebarCategories and UpdateChannelSidebarCategories to Plugin API
```


madhavhugar:
@agarciamontoro created this PR to support new Plugin APIs for [categorizing incident channel](https://github.com/mattermost/mattermost-plugin-incident-collaboration/pull/598). Please review.

I see that `test-mysql-1` pipeline fails, the error seems unrelated to these changes, let me know what can be done to resolve it

agarciamontoro:
Thanks for your contribution, @madhavhugar! This is fantastic :)

The failing test seems like a flaky test. I've clicked re-run to see if it passes.

plugin/api.go
```diff
@@ -450,6 +450,30 @@ type API interface {
 	// Minimum server version: 5.6
 	SearchChannels(teamID string, term string) ([]*model.Channel, *model.AppError)
 
+	// CreateChannelSidebarCategory creates a new sidebar category for a set of channels.
+	//
+	// @tag Channel
+	// @tag User
+	// @tag Team
```
agarciamontoro, line 457: What if we create a new tag, `ChannelSidebar`, instead of adding these three? (same comment for the other two declarations)

madhavhugar: Sounds good to me, added the new tag ✅ 

Should we explicitly create a new tag or just adding them is sufficient? Also, how are these tags used/parsed?

agarciamontoro: Just adding them is enough. The tags are parsed when building the plugin documentation, which outputs the page in https://developers.mattermost.com/extend/plugins/server/reference/. The code parsing the tags is in https://github.com/mattermost/mattermost-developer-documentation/blob/master/cmd/plugin-godocs/plugin-godocs.go#L73-L81

agarciamontoro (approved):
This is great, @madhavhugar, thank you very much :pray:

plugin/api.go
madhavhugar: ah I see thank you

agarciamontoro:
@hmhealey, we usually test it from the PR that motivated the change (in this case, https://github.com/mattermost/mattermost-plugin-incident-collaboration/pull/598): worst case scenario, the new functionality added here needs to be fixed somehow, but it will not break anything else. That said, merging this one! :)

agarciamontoro:
All tests passing, thank you again, @madhavhugar!

PR #17822, opened 2021-06-23
Bump focalboard prepackaged plugin version to v0.7.0

mgdelacroix:
```release-note
Updated focalboard builtin plugin version to v0.7.0
```


jespino (approved):

chenilim (commented):
Just double checking that the format of the name changed from `focalboard-v0.6.7-plugin` to `focalboard-v0.7.0`. What does `focalboard-v0.7.0` correspond to @mgdelacroix? No need to hold up the merge for this, but want to check if this affects any dependencies. Thanks.

mgdelacroix:
@chenilim it's correct. The name is generated from the tag that is used on the release, and we've changed from using `v0.6.7-plugin` to include everything on the same release, using `v0.7.0` without the suffix.

PR #14047, opened 2020-03-12
Adding correctness in the ReplyCount generation

jespino:
#### Summary
Adding correctness in the ReplyCount generation. This way all the posts has the ReplyCount of the thread it belongs to.

This is part of the skip threads optimization.

#### Ticket Link
[MM-17468](https://mattermost.atlassian.net/browse/MM-17468)

PR #17845, opened 2021-06-28
[MM-36711] nested sql query for reply_count in getSingle post method of post_sto…

source-punk:
#### Summary
Returns reply_count from the GET post api

#### Ticket Link
  Fixes https://mattermost.atlassian.net/browse/MM-36711
  Related Issue https://github.com/mattermost/mattermost-server/issues/17834

#### Release Note
```release-note
NONE
```


agnivade:
I am not really familiar with the CRT feature. Would prefer someone else to take a look.

ashishbhate:
@source-punk 

The ReplyCount field of the post object has been around from before the new collapsed reply threads stuff. 

I looked through the git history and source code comments, but I couldn't find anything definitive why this store method doesn't return the post counts, but other methods do (e.g `GetPostsByIds`) . The api layer populates the returned post with reactions and other things, so I don't see a reason we shouldn't fill in the reply count as well.

Although this is likely a long hidden bug, I'm curious to find out if there is a historical reason why we don't calculate reply counts in this store method

@jespino any ideas? While digging through the git history I noticed you implemented a related PR a long time ago - #14047

source-punk:
@ashishbhate We only found this issue because we were trying to use the normal post's GET api to fetch reply counts. And they were always being returned 0. Other than that, i'm pretty new to mattermost server and I dont really have any idea of why we shouldn't be computing the reply count in the store method.

agnivade:
I think @hmhealey knows the reason for this. I remember this being discussed sometime back.

hmhealey:
It looks like that was added a little over a year ago when we changed it so that a post could be fetched without fetching every single post in its thread. That was mainly impacting things like loading posts in the web app's post list, so maybe we didn't bother calculating the ReplyCount for API routes that weren't used by that part of the web app. I'm not actually sure if the `GetSingle` method is used for anything related to the web app to begin with.

Here's the PR that added it for anyone who's curious: https://github.com/mattermost/mattermost-server/pull/13653

ashishbhate:
@source-punk the PR looks good in principle. Could you add tests covering some common cases for this change (no replies, deleted posts etc.)?

source-punk:
@ashishbhate sure, I'll have to learn how to do that in mattermost, so might take a day or two to push some tests.

ashishbhate:
https://github.com/mattermost/mattermost-server/blob/master/store/storetest/post_store.go has a test for this store method. You can just add a few sub-tests. Let me know if I can help with anything.

source-punk:
Sure, and thanks for pointing me to the file that has tests.

source-punk:
Sorry, i've been busy with my day job and couldn't get the time to look at writing tests, i'll finish that and push the commit over the weekend.

source-punk:
@ashishbhate when going through the storetest/post_store.go i found this
`t.Run("should return the last posts created in a channel and the threads and the reply count must be 0", func(t *testing.T) {`

this is run as part of the test  `	t.Run("GetPosts", func(t *testing.T) { testPostStoreGetPosts(t, ss) })`

the above test creates thread posts with root ids and asserts that the fetched root posts also show their reply count as 0. I'm not sure what is going on with that test.

source-punk:
I have added tests to the GetSingle test in `storetest/post_store.go` @ashishbhate

ashishbhate:
>the above test creates thread posts with root ids and asserts that the fetched root posts also show their reply count as 0. I'm not sure what is going on with that test.

Based on the `SkipFetchThreads` option the GetPosts method may not return the reply count.

store/storetest/post_store.go
```diff
@@ -471,12 +471,34 @@ func testPostStoreGetSingle(t *testing.T, ss store.Store) {
...
+	o3 := &model.Post{}
+	o3.ChannelId = o1.ChannelId
+	o3.UserId = o1.UserId
+	o3.Message = "zz" + model.NewId() + "d"
+	o3.RootId = o1.Id
+
+	o4 := &model.Post{}
+	o4.ChannelId = o1.ChannelId
+	o4.UserId = o1.UserId
+	o4.Message = "zz" + model.NewId() + "e"
```
ashishbhate, line 483: We should use the `model.NewRandomString` method here and above (line 477)

agnivade: It's better to use the `NewTestId` function for this. That was specifically made to remove this "zz" hack.

ashishbhate: Here we just need a random string for the post message body, so I thought `NewRandomString` would be good enough.

agnivade: Ahh sure, that's fine. :+1:

source-punk: I just followed the pattern of other objects o1 and o2  for message body, but sure, I'll change it to NewRandomString()

agnivade (approved):
Looks good after the changes suggested by @ashishbhate are done. Thanks @source-punk !

source-punk:
Done. Thanks for awesome feedback on the PR!

ashishbhate (approved):
Great job @source-punk !

ashishbhate:
@jgilliam17 This PR changes some data returned from a post API that appears to be unused by the webapp. I checked on mobile and we aren't using the changed field anywhere in mobile either. 

Some smoke testing around posts and thread reply counts (old style threads) should be good enough.

jgilliam17 (approved):
Thank you @source-punk 
Tested, looks good to merge.
- Verified posts and old style thread reply counts on webapp and mobile - no issues, as expected.

PR #17846, opened 2021-06-28
[MM-36481] Fix emoji regex to support thumbsup

Willyfrog:
#### Summary

Add + as a valid emoji character

#### Ticket Link

[MM-36481](https://mattermost.atlassian.net/browse/MM-36481)

#### Release Note

```release-note
NONE
```


jwilander (approved):
LGTM

app/post_metadata.go
```diff
@@ -250,7 +250,7 @@ func (a *App) getImagesForPost(post *model.Post, imageURLs []string, isNewPost b
 }
 
 func getEmojiNamesForString(s string) []string {
-	names := model.EMOJI_PATTERN.FindAllString(s, -1)
```
hmhealey, line 253: Instead of switching this to ALL_EMOJI_PATTERN, could you just get rid of ALL_EMOJI_PATTERN and make EMOJI_PATTERN include the plus sign? We don't need to keep both of them around any more, and we can also remove [this TODO comment](https://github.com/mattermost/mattermost-server/blob/master/model/emoji.go#L24) at the same time

hmhealey (changes requested):

hmhealey (approved):

Willyfrog:
seems the test are failing not regarding this PR but by some previous one, is it ok to merge? or should we wait until it gets fixed?

PR #17852, opened 2021-06-30
Use forked version of gziphandler

agnivade:
```release-note
NONE
```


agnivade:
cc @isacikgoz

ashishbhate (approved):
LGTM

One minor point: the NOTICE.TXT in this repo still mentions `NYTimes/gziphandler`. 
Maybe it would be more correct to move it to the NOTICE.TXT file in the `mattermost/gziphandler` repo, since the mattermost fork is where we actually create the derivative work.

agnivade:
Yeah I saw that. @amyblais takes care of updating the NOTICE.txt file every release. I think it would be changed at that time.

PR #17680, opened 2021-05-26
[MM-8497] Ability to set Do Not Disturb for a specified period of time

darkLord19:
This reverts commit ff383990f8c8920798c559fdbd91150812c8b363.

PR #17853, opened 2021-06-30
move migration to future release as 536 is already done

darkLord19:
Related to PR #17680


darkLord19:
@agnivade

jespino (approved):
LGTM

agnivade:
@darkLord19 - I am trying to look back into the timeline of events here. It seems like the revert commit added the migration rather than removing it. Was that intentional?

And did the revert happen _after_ 5.36 was released or before?

darkLord19:
@agnivade 
original pr was reverted on 21st may, so there you can see migration being removed
then i had reverted the above revert to basically bring back the feature which was merged around 14-15th June and it's not shipped in 5.36.1
So, i have raised this pr to move migrations to 5.37 release

agnivade:
Right, so then technically, when we un-reverted it, the migration should have moved to the 5.37 function right? Just confirming the timeline of events here.

darkLord19:
Yes you are correct. I was under the impression that it'll go in 5.36 but today I got the chance to take look at release branch and didn't find the pr there so raised separate pr to move the migration

agnivade (approved):

PR #17879, opened 2021-07-08
Consider empty body before unmarshal on do post action

larkox:
#### Summary
Consider an empty body in the doPostAction function before doing the unmarshalling.

If we don't do this, an empty response (which is common) will make this function error out after creating the trigger ID and sending it to the integration.
A subsequent call to openDialog will not fail because the trigger ID exist in the server, but the trigger ID has not been passed over to the client (because the function error out). Therefore, openDialog says that everything is fine, but no dialog is shown in the client.

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-36757

#### Release Note
```release-note
Fix bug when opening a dialog from an interactive message when returning an empty response
```

### DIFF
--- a/app/integration_action.go
+++ b/app/integration_action.go
@@ -23,7 +23,6 @@ import (
 	"encoding/json"
 	"errors"
 	"fmt"
-	"io"
 	"io/ioutil"
 	"net/http"
 	"net/url"
@@ -250,7 +249,7 @@ func (a *App) DoPostActionWithCookie(c *request.Context, postID, actionId, userI
 	defer resp.Body.Close()
 
 	var response model.PostActionIntegrationResponse
-	respBytes, err := io.ReadAll(resp.Body)
+	respBytes, err := ioutil.ReadAll(resp.Body)
 	if err != nil {
 		return "", model.NewAppError("DoPostActionWithCookie", "api.post.do_action.action_integration.app_error", nil, "err="+err.Error(), http.StatusBadRequest)
 	}
--- a/app/integration_action.go
+++ b/app/integration_action.go
@@ -23,6 +23,7 @@ import (
 	"encoding/json"
 	"errors"
 	"fmt"
+	"io"
 	"io/ioutil"
 	"net/http"
 	"net/url"
@@ -249,8 +250,15 @@ func (a *App) DoPostActionWithCookie(c *request.Context, postID, actionId, userI
 	defer resp.Body.Close()
 
 	var response model.PostActionIntegrationResponse
-	if err := json.NewDecoder(resp.Body).Decode(&response); err != nil {
+	respBytes, err := io.ReadAll(resp.Body)
+	if err != nil {
 		return "", model.NewAppError("DoPostActionWithCookie", "api.post.do_action.action_integration.app_error", nil, "err="+err.Error(), http.StatusBadRequest)
+	}
+
+	if len(respBytes) > 0 {
+		if err = json.Unmarshal(respBytes, &response); err != nil {
+			return "", model.NewAppError("DoPostActionWithCookie", "api.post.do_action.action_integration.app_error", nil, "err="+err.Error(), http.StatusBadRequest)
+		}
 	}
 
 	if response.Update != nil {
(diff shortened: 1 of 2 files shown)


hanzei (changes requested):
Can we add a test for this?

app/integration_action.go
```diff
@@ -249,10 +249,17 @@ func (a *App) DoPostActionWithCookie(c *request.Context, postID, actionId, userI
 	defer resp.Body.Close()
 
 	var response model.PostActionIntegrationResponse
-	if err := json.NewDecoder(resp.Body).Decode(&response); err != nil {
+	respBytes, err := ioutil.ReadAll(resp.Body)
```
hanzei, line 252: Nit: `ioutil` is deprecated 
```suggestion
	respBytes, err := io.ReadAll(resp.Body)
```

hanzei (approved):
Thanks for adding a test :+1:

app/integration_action.go
larkox: Govet was complaining because I used io. Not sure if we must do some update to govet.

amyblais:
@larkox From your final investigation, can you help confirm if this is a regression? (I know you mentioned it seems that the code hasn't been touched for 3 years, just wanted to double-check)

larkox:
@amyblais Correct. I don't think it is a regression.

DHaussermann:
@larkox or @hanzei Can you provide some test guidance or steps for validating this change with functional testing?
I'm not sure the test-app would capture this case.

larkox:
@DHaussermann it is out of the scope of the test app, but I can make some adjustments to have this ready testable with it.

wiggin77 (approved):
LGTM  :+1:

larkox:
@DHaussermann text app master has a command to test this now. It is on `/test other`.

DHaussermann:
@larkox thanks for adding the text app commands. 

Some of these don't seem to be working as expected. `empty response`, `ephemeral response` and `update response` seem like they should open a modal but they do not. Is that correct?  
image

The test app shows the following `empty response`

```
2021/07/16 10:30:19 http: panic serving [::1]:56625: interface conversion: interface {} is nil, not string
goroutine 156 [running]:
net/http.(*conn).serve.func1(0xc00046a1e0)
	/usr/local/go/src/net/http/server.go:1824 +0x153
panic(0x1894140, 0xc000442ae0)
	/usr/local/go/src/runtime/panic.go:971 +0x499
github.com/mattermost/mattermost-app-test/routers/mattermost.postOpenDialogTestEmptyResponse(0x1aa4ea0, 0xc0001728c0, 0xc000292d00)
	/Users/dylanhaussermann/go/src/github.com/larkox/mattermost-app-test/routers/mattermost/fOther.go:128 +0x33f
net/http.HandlerFunc.ServeHTTP(0x19dfdc8, 0x1aa4ea0, 0xc0001728c0, 0xc000292d00)
	/usr/local/go/src/net/http/server.go:2069 +0x44
github.com/gorilla/mux.(*Router).ServeHTTP(0xc000338300, 0x1aa4ea0, 0xc0001728c0, 0xc000292b00)
	/Users/dylanhaussermann/go/pkg/mod/github.com/gorilla/mux@v1.8.0/mux.go:210 +0xd3
net/http.(*ServeMux).ServeHTTP(0x1f973a0, 0x1aa4ea0, 0xc0001728c0, 0xc000292b00)
	/usr/local/go/src/net/http/server.go:2448 +0x1ad
net/http.serverHandler.ServeHTTP(0xc0003282a0, 0x1aa4ea0, 0xc0001728c0, 0xc000292b00)
	/usr/local/go/src/net/http/server.go:2887 +0xa3
net/http.(*conn).serve(0xc00046a1e0, 0x1aa65e0, 0xc00028c640)
	/usr/local/go/src/net/http/server.go:1952 +0x8cd
created by net/http.(*Server).Serve
	/usr/local/go/src/net/http/server.go:3013 +0x39b
````

Please advise on if there are futher changes needed to the PR or the Test App.

larkox:
@DHaussermann Fixed the test app. Master should behave properly now.

DHaussermann:
Thanks @larkox the changes made to the test-app now work to validate this change.

DHaussermann (approved):
Tested and passed
- All 5 actions are working as expected
- Tested on Webapp and Mobile
- No issues found

LGTM!

I will add a TM4J test case to cover this.

PR #17881, opened 2021-07-08
MM-36894: Racy test: TestExecuteCommandInDirectMessageChannel

agnivade:
We were using the pointer to the same post object
in multiple push notifications because we were sending
it in a loop of mentioned user list.

This wasn't an issue so far, but after we started to parse
markdown in the push notification flow, we started to edit
the post struct as well.

To fix this, we move the markdown parsing deep into the push notification
construction such that we don't need to write back to the same post pointer
again.

https://mattermost.atlassian.net/browse/MM-36894

```release-note
NONE
```

### DIFF
--- a/app/notification.go
+++ b/app/notification.go
@@ -371,7 +371,6 @@ func (a *App) SendNotifications(post *model.Post, team *model.Team, channel *mod
 					replyToThreadType = model.COMMENTS_NOTIFY_ROOT
 				}
 
-				notification.Post = notification.Post.Clone()
 				a.sendPushNotification(
 					notification,
 					profileMap[id],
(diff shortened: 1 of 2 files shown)


agnivade:
The post is already cloned once when the notification struct is created. But I did it again out of abundance of caution, in case the same struct is used elsewhere in future. Let me know if you prefer it to be removed.

agnivade:
Actually, thinking a bit more, we can avoid this by doing the markdown stuff before sending it to the push notification hub. I thought of doing the parsing off the main code path. But it seems like to avoid races, we need to do a clone.

So the tradeoff is between doing a clone vs markdown parsing. :thinking:

streamer45 (commented):
If the side effect isn't need elsewhere, I am wondering if this could be fixed by just moving the markdown parsing even deeper in the code, only in `buildFullPushNotificationMessage` at which point we would only care about the returned message.

agnivade:
Nice one, I like it. :+1:

streamer45 (approved):
Nice :+1:

PR #17887, opened 2021-07-08
Default CollapsedThreads feature flag to on

jwilander:
#### Summary
For beta we want to default the feature flag to on. The feature itself has a config that defaults it to off. This just allows admins to turn on the beta without manually setting the feature flag env var to on.

cc @amyblais @esethna we need this for 5.37

#### Ticket Link
N/A

#### Release Note

```release-note
Default make collapsed reply threads beta feature accessible to turn on from the system console
```

### DIFF
--- a/model/feature_flags.go
+++ b/model/feature_flags.go
@@ -44,7 +44,7 @@ func (f *FeatureFlags) SetDefaults() {
 	f.TestFeature = "off"
 	f.TestBoolFeature = false
 	f.CloudDelinquentEmailJobsEnabled = false
-	f.CollapsedThreads = false
+	f.CollapsedThreads = true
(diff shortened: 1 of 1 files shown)


koox00 (approved):

ashishbhate (approved):

amyblais:
@jwilander @esethna This is needed for the next cloud release too, I think?

esethna:
@amyblais I think this only affects on-prem as the cloud release already has the feature flag on afaik, @jwilander?

jwilander:
That's correct @esethna, it's only needed for on-prem. cc @amyblais

jwilander:
Cherry pick complete

PR #17888, opened 2021-07-08
MM-36779 - Add ChimeraOAuthProxyURL config option

Szymongib:
#### Summary

This PR adds a new configuration option `ChimeraOAuthProxyUrl` which allows specifying [Chimera](https://github.com/mattermost/chimera) URL that can be used by plugins to connect with pre-created OAuth applications.

#### Ticket Link

https://mattermost.atlassian.net/browse/MM-36779

#### Release Note

```release-note
Added a new config setting PluginSettings.ChimeraOAuthProxyURL
```

### DIFF
--- a/model/config.go
+++ b/model/config.go
@@ -2778,6 +2778,7 @@ type PluginSettings struct {
 	RequirePluginSignature      *bool                             `access:"plugins,write_restrictable,cloud_restrictable"`
 	MarketplaceUrl              *string                           `access:"plugins,write_restrictable,cloud_restrictable"`
 	SignaturePublicKeyFiles     []string                          `access:"plugins,write_restrictable,cloud_restrictable"`
+	ChimeraOAuthProxyUrl        *string                           `access:"plugins,write_restrictable,cloud_restrictable"`
 }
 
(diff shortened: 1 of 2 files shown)


jwilander:
Throwing @larkox on review so that someone more familiar with integrations/plugins takes a look

larkox:
All plugins will rely on one single chimera? There is any usecase where I want some plugins connect through the general chimera, and some others through a custom chimera?

Szymongib:
> All plugins will rely on one single chimera? There is any usecase where I want some plugins connect through the general chimera, and some others through a custom chimera?

@larkox Yes, the plan is that they will rely on a single Chimera. We will run a single instance and while it would be possible that user decides to run their own Chimera I do not really think that would make much sense (there will still be an option to just connection plugin through your own app instead of Chimera, therefore I do not see a reason to run your own).

Having this config here will also make it easier for us to configure it for the cloud. We currently do not have a way to change plugins configuration and we probably do not want users to mess with the URL in the cloud offering.

gabrieljackson (approved):
👍 No issues from me on this, but I will defer to any concerns @larkox has.

larkox (approved):

PR #17890, opened 2021-07-08
Reliable Websockets: default to true

agnivade:
```release-note
The config setting EnableReliableWebSockets now defaults to true.
```


isacikgoz (approved):
Yay 🎉

PR #17901, opened 2021-07-12
MM-36947: Fix TestHeaderProviderDoCommand

agnivade:
Setting AutomaticPrepackagedPlugins to true was making the test
to download the plugin from the marketplace and initialize it.

While this wasn't wrong in itself but that's not what the test does
and just part of the app initialization and it unnecessarily slowed
down the test by downloading the plugin zip file.

And initializing a plugin from inside a test has some issues
to work properly. While those could be solved, a cleaner way
is to just not download the plugin at all.

https://mattermost.atlassian.net/browse/MM-36947

```release-note
NONE
```

```


isacikgoz (approved):
👍

mgdelacroix (approved):

PR #17910, opened 2021-07-13
MM-21333: Remove deprecated session method

agnivade:
```release-note
Removed (*Session).SetExpireInDays method.
```

https://mattermost.atlassian.net/browse/MM-21333


agnivade:
@hmhealey - Would you be able to check and confirm that the remove event types aren't used in the web/mobile client any more?

model/session.go
```diff
@@ -125,17 +125,6 @@ func (s *Session) IsExpired() bool {
 	return false
 }
 
-// Deprecated: SetExpireInDays is deprecated and should not be used.
```
agnivade, line 128: /cc @wiggin77

agnivade:
Ah no, the webapp and mobile still reference them. I will send other PRs separately.

store/sqlstore/channel_store.go
```diff
@@ -2179,8 +2179,6 @@ func (s SqlChannelStore) UpdateLastViewedAt(channelIds []string, userId string,
 func (s SqlChannelStore) CountPostsAfter(channelId string, timestamp int64, userId string) (int, int, error) {
 	joinLeavePostTypes := []string{
 		// These types correspond to the ones checked by Post.IsJoinLeaveMessage
-		model.PostTypeJoinLeave,
```
agnivade, line 2182: Ahh bummer .. this is needed for the SQL query to work with old posts. So it seems like we need that in server too. @hmhealey. I'll revert the changes.

wiggin77 (approved):
Good to get rid of that.  :+1:

hmhealey:
Perhaps as a middleground, we can remove the exported constant and keep it internal? That'll at least keep the public API cleaner even if we have some gremlins of our own caused by the old post type.

agnivade:
Can you expand on what do you mean by keeping it internal? We have to export the constant anyways.

model/session_test.go
```diff
@@ -41,26 +39,6 @@ func TestSessionDeepCopy(t *testing.T) {
...
-
-	require.False(t, session.IsExpired(), "Shouldn't expire")
-
-	session.ExpiresAt = GetMillis()
-	time.Sleep(10 * time.Millisecond)
-
-	require.True(t, session.IsExpired(), "Should expire")
-
-	session.SetExpireInDays(10)
-}
```
streamer45, line 62: How is this test related? Apart from removing the `SetExpireInDays()` call I am guessing this can be left as-is.

streamer45 (approved):
Nice :+1:

model/session_test.go
agnivade: Yeah, I guess so. Will change.

hmhealey:
Oh right, I forgot how private fields work in Go since even other packages can't access them. I guess we could have some sort of `model/internal` or just move the constants out of models and into the code that uses it, but that might be complicating things too much. 0/5

cwarnermm:
@agnivade - Can you please confirm that this PR has no implications to our documented Sessions config settings, starting with https://docs.mattermost.com/configure/configuration-settings.html#session-lengths and ending with https://docs.mattermost.com/configure/configuration-settings.html#session-idle-timeout-minutes? Thank you so much for your help with this!

agnivade:
Yes, there is no functional impact.

PR #17913, opened 2021-07-13
[MM-35470] Disable config watching logic

streamer45:
#### Summary

PR disables the file watching logic for the main config file. This should permanently fix various related issues that were reported by users (e.g. config resets on restart).

Ideally we'd like to remove this logic completely but I see the watcher is also used by the advanced logging system. @wiggin77 let me know if you have any objection or foresee problems in doing that.

#### Ticket

https://mattermost.atlassian.net/browse/MM-35470

#### Release Note

```release-note
Disabled config file watching logic.
```

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -451,7 +451,7 @@ run-server: prepackaged-binaries validate-go-version start-docker ## Starts the 
 	@echo Running mattermost for development
 
 	mkdir -p $(BUILD_WEBAPP_DIR)/dist/files
-	$(GO) run $(GOFLAGS) -ldflags '$(LDFLAGS)' $(PLATFORM_FILES) --disableconfigwatch 2>&1 | \
+	$(GO) run $(GOFLAGS) -ldflags '$(LDFLAGS)' $(PLATFORM_FILES) 2>&1 | \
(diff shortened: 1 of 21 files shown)


agnivade (approved):
Looks great!

wiggin77:
> Ideally we'd like to remove this logic completely but I see the watcher is also used by the advanced logging system

Ok to remove IMO.  The logging config supports hot reload only because the main config did.

wiggin77 (approved):
LGMT :+1:   Nice reduction in complexity.

streamer45:
@amyblais 
Just letting you know that this was scheduled for 5.38 as announced a couple of months back so it would be good to include it there. Thanks :+1:

amyblais:
> @amyblais
> Just letting you know that this was scheduled for 5.38 as announced a couple of months back so it would be good to include it there. Thanks 👍

We're past feature complete, so we can re-discuss about if we want to cherry-pick this once it has been tested.

streamer45:
> We're past feature complete, so we can re-discuss about if we want to cherry-pick this once it has been tested.

Technically this is a bugfix but makes sense, let's wait for QA first. cc @wiersgallak

amyblais:
@furqanmlk Kind reminder for PR review.

furqanmlk (approved):
getting/setting config values are working as expected

PR #17919, opened 2021-07-14
MM-37034: Bump the MySQL min version to 5.7

agnivade:
https://mattermost.atlassian.net/browse/MM-37034
https://mattermost.atlassian.net/browse/MM-36449

```release-note
Minimum required MySQL version is now 5.7.12
```


streamer45:
I am assuming this is addressing https://mattermost.atlassian.net/browse/MM-36449. In that case we need to make sure the minimum supported version is 5.7.8 so that we can support native JSON fields.

agnivade:
Ahh thanks for that. Will update.

agnivade:
@wiersgallak @amyblais - We did some testing and found a bug in 5.7.8, which got fixed in 5.7.9. After some more discussion, we have decided to bump the minimum patch version to 12. This is version is also available on AWS Aurora, AWS RDS which are our benchmarks.

Kindly update the v6 announcements to reflect that the minimum MySQL version required is now 5.7.12, not 5.7.8. Thank you :)

store/sqlstore/store.go
```diff
@@ -179,24 +181,62 @@ func New(settings model.SqlSettings, metrics einterfaces.MetricsInterface) *SqlS
 
 	store.initConnection()
 
-	if *settings.DriverName == model.DatabaseDriverPostgres {
-		ver, err := store.GetDbVersion(true)
-		if err != nil {
-			mlog.Critical("Cannot get DB version.", mlog.Err(err))
+	ver, err := store.GetDbVersion(true)
```
streamer45, line 184: Would it be feasible for the version check to become a separate function that either returns an error or simple boolean? It may look better overall and offer some simplifications (e.g. early returns, less logging calls).

amyblais:
> @wiersgallak @amyblais - We did some testing and found a bug in 5.7.8, which got fixed in 5.7.9. After some more discussion, we have decided to bump the minimum patch version to 12. This is version is also available on AWS Aurora, AWS RDS which are our benchmarks.
> 
> Kindly update the v6 announcements to reflect that the minimum MySQL version required is now 5.7.12, not 5.7.8. Thank you :)

I'll update the changelog.

wiersgallak:
> Kindly update the v6 announcements to reflect that the minimum MySQL version required is now 5.7.12, not 5.7.8. Thank you :)

Blog has been updated as well.

store/sqlstore/store.go
```diff
@@ -1678,12 +1669,74 @@ func IsDuplicate(err error) bool {
...
 }
 
-// VersionString converts an integer representation of a DB version
+// ensureMinimumDBVersion gets the DB version and ensures it is
+// above the required minimum version requirements.
+func (ss *SqlStore) ensureMinimumDBVersion() (bool, error) {
+	ver, err := ss.GetDbVersion(true)
+	if err != nil {
+		return false, err
+	}
```
streamer45, line 1678: I am thinking that if we decouple this from getting the current db version, we can then more easily test this method in isolation since it's doing a bit of parsing as well which would be good to get covered.

agnivade: Sure.

store/sqlstore/store.go
```diff
@@ -1678,12 +1675,70 @@ func IsDuplicate(err error) bool {
...
+		if err2 != nil {
+			return false, fmt.Errorf("cannot parse DB version: %v", err2)
+		}
+		if intVer < minimumRequiredPostgresVersion {
+			return false, fmt.Errorf("minimum Postgres version requirements not met. Found: %s, Wanted: %s", versionString(intVer, *ss.settings.DriverName), versionString(minimumRequiredPostgresVersion, *ss.settings.DriverName))
+		}
+	case model.DatabaseDriverMysql:
+		// Usually a version string is of the form 5.6.49-log, 10.4.5-MariaDB etc.
+		if strings.Contains(strings.ToLower(ver), "maria") {
+			mlog.Debug("MariaDB detected. Skipping version check.")
```
streamer45, line 1693: We could either break or return here to simplify the else block.

streamer45 (approved):
Awesome :+1:

mkraft:
@agnivade Was it decided not to go with the latest 5.7.35 because there was some missing support in AWS?

agnivade:
That is correct. AWS Aurora only supports till 5.7.12.

cwarnermm:
Docs updated via https://github.com/mattermost/docs/pull/4971

PR #17923, opened 2021-07-14
[MM-29217] Remove Posts.ParentId

streamer45:
#### Summary

Removing the deprecated `Posts.ParentId`.

#### Ticket

https://mattermost.atlassian.net/browse/MM-29217

#### Release Notes

```release-note
Removed deprecated Posts.ParentId in favor of the semantically equivalent Posts.RootId. Also removed CommandWebhook.ParentId and CompliancePost.ParentId for the same reason.
```

#### Related PRs

https://github.com/mattermost/enterprise/pull/1019
https://github.com/mattermost/mattermost-api-reference/pull/648
https://github.com/mattermost/mattermost-webapp/pull/8401
https://github.com/mattermost/mattermost-mobile/pull/5541


agnivade (approved):
Nice!

store/sqlstore/upgrade.go
```diff
@@ -1211,3 +1212,10 @@ func upgradeDatabaseToVersion537(sqlStore *SqlStore) {
 	// saveSchemaVersion(sqlStore, Version5370)
 	// }
 }
+
+func upgradeDatabaseToVersion600(sqlStore *SqlStore) {
+	sqlStore.GetMaster().ExecNoTimeout("UPDATE Posts SET RootId = ParentId WHERE RootId = '' AND RootId != ParentId")
+	sqlStore.RemoveColumnIfExists("Posts", "ParentId")
+	sqlStore.GetMaster().ExecNoTimeout("UPDATE CommandWebhooks SET RootId = ParentId WHERE RootId = '' AND RootId != ParentId")
+	sqlStore.RemoveColumnIfExists("CommandWebhooks", "ParentId")
```
streamer45, line 1220: Honestly I don't know if the updates are really needed but added them to be on the safe side.

isacikgoz (approved):
Less code! I like this 🥳

streamer45:
@furqanmlk 
This should be a very safe change since we've not been using `ParentId` in a very long time. I guess either running e2e tests or smoke-test around post (with replies) should be sufficient to cover QA for this.

furqanmlk (approved):
97% E2E Passed ✅ 

https://automation-dashboard.vercel.app/cycles/84d02be4-23a5-414e-9689-34f0b1b87948

streamer45:
This is another schema change on `Posts`. Similarly to [the previous one](https://github.com/mattermost/mattermost-server/pull/17930) we should be expecting an extended migration time during which no updates to the table can be made.

/cc @spirosoik

spirosoik:
@streamer45 when do we plan to merge this? 

We need to mention this on the release notes because a second migration on posts will affect on-prem customers, of course based on the size of their data and database. cc @stylianosrigas

streamer45:
> @streamer45 when do we plan to merge this?

Does Saturday morning sound good? :P This is ready to merge whenever you guys feel it's appropriate/convenient.

> We need to mention this on the release notes because a second migration on posts will affect on-prem customers, of course based on the size of their data and database.

Sure. That's planned already as with every release making such changes. I usually prepare extensive release notes.

store/sqlstore/upgrade.go
spirosoik: could we cross-check if these updates are really needed.

My only worry here is mostly for MySQL. MySQL (as far I remember for 5.6 and above) locks the table for alter. There is an attribute `LOCK=NONE` for `InnoDB` which allows read and writes during the operation I think.

Could we give a try to see what happens and see if the read/write still works?

agnivade: It's a bit difficult to say whether the updates are really needed or not because it really depends on the customer's DB state. But more importantly, I think the update should not take more than a few seconds since we are searching by a single index. We can easily confirm that. Given the benefit it gives (a cleaner DB state), that seems like a small cost to pay.

> There is an attribute LOCK=NONE for InnoDB which allows read and writes during the operation I think.

That is only for `ALTER TABLE` statements, not `UPDATE TABLE`. And even for those scenarios, it won't work for our case because we use more than two full-text indexes in the Posts table.

streamer45: Those were only added to prevent data loss but I agree that most probably they are not needed. I've checked our community db and we actually have cases (I found 55) where both `RootId` and `ParentId` are set but with different, non-empty values so that case wouldn't be covered by that. If I recall we were experimenting with subthreading at some point so that may be related.

Of course I've checked the update query on a big dataset (25M posts) and it took a couple of minutes which is nothing compared to the main schema change so eventually decided to leave it.

streamer45:
@stylianosrigas Wondering if we could schedule this to be merged sometime soon?

agnivade:
Stelios is off. Pinging @spirosoik @angeloskyratzakos @stafot

spirosoik:
@agnivade @streamer45 we will discuss with team to merge this early morning on Monday

spirosoik:
@streamer45 please rebase and resolve the conflicts, we plan to merge this early tomorrow morning (GMT+3 timezone) so we can have less impact in the community.

spirosoik:
@streamer45 seems tests are failing

cwarnermm:
@streamer45 - Is any documentation outstanding for this PR?

PR #17926, opened 2021-07-14
[MM-35925] Follow thread when posting in one, even if previously unfollowed

ashishbhate:
#### Summary

- When ThreadAutoFollow is on, replying to a thread that was previously unfollowed should cause the thread to be followed again.

#### Ticket Link

- https://mattermost.atlassian.net/browse/MM-35925
- 
#### Release Note

```release-note
NONE
```

### DIFF
--- a/app/post.go
+++ b/app/post.go
@@ -353,6 +353,16 @@ func (a *App) CreatePost(c *request.Context, post *model.Post, channel *model.Ch
 	// to be done when we send the post over the websocket in handlePostEvents
 	rpost = a.PreparePostForClient(rpost, true, false)
 
+	// Make sure poster is following the thread
+	if *a.Config().ServiceSettings.ThreadAutoFollow && rpost.RootId != "" {
+		_, err := a.Srv().Store.Thread().MaintainMembership(user.Id, rpost.RootId, store.ThreadMembershipOpts{
+			Following: true,
+		})
+		if err != nil {
+			mlog.Warn("Failed to update thread membership", mlog.Err(err))
+		}
+	}
+
 	if err := a.handlePostEvents(c, rpost, user, channel, triggerWebhooks, parentPostList, setOnline); err != nil {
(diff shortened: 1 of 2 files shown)


app/post.go
```diff
@@ -353,6 +353,16 @@ func (a *App) CreatePost(c *request.Context, post *model.Post, channel *model.Ch
...
+	// Make sure poster is following the thread
+	if *a.Config().ServiceSettings.ThreadAutoFollow && rpost.RootId != "" {
+		_, err := a.Srv().Store.Thread().MaintainMembership(user.Id, rpost.RootId, store.ThreadMembershipOpts{
+			Following: true,
+		})
+		if err != nil {
+			mlog.Warn("Failed to update thread membership", mlog.Err(err))
+		}
+	}
+
```
ashishbhate, line 365: Existing convention would suggest updating the code in  `SendNotifications` in `app/notifications.go`. I want to clean up that function (eventually), so the first step is to not add more stuff to it.

hmhealey (approved):
Looks good to me, but I'm surprised there wasn't a block of code like this before. Does the existing logic only follow the thread if the user hasn't manually unfollowed it before?

ashishbhate:
>Looks good to me, but I'm surprised there wasn't a block of code like this before. Does the existing logic only follow the thread if the user hasn't manually unfollowed it before?

The existing logic follows threads if its the user's first time replying (i.e. the user doesn't have a thread membership) or if the user was mentioned in the thread. All of that code is in `app/notifications.go:SendNotifications`, its a bit of a mess.

hmhealey:
Yeah, I'm not surprised. I've worked with that file before in the past, and it already had a lot going on then

jwilander (approved):

jgilliam17 (approved):
Thanks @ashishbhate 
Tested, looks good to merge.
- Verified when replying in a thread you previously unfollowed, thread is followed again.

PR #17927, opened 2021-07-15
MM-36411- Skip flaky TestUpdateConfigDiffInAuditRecord

agnivade:
The fix in https://github.com/mattermost/mattermost-server/pull/17777
doesn't seem to have worked fully as this was seen again.

Skipping it for now until we revisit this.

https://mattermost.atlassian.net/browse/MM-36411

```release-note
NONE
```

### DIFF
--- a/api4/config_test.go
+++ b/api4/config_test.go
@@ -441,7 +441,6 @@ func TestUpdateConfigRestrictSystemAdmin(t *testing.T) {
 }
 
 func TestUpdateConfigDiffInAuditRecord(t *testing.T) {
-	t.Skip("MM-36411")
 	logFile, err := ioutil.TempFile("", "adv.log")
 	require.NoError(t, err)
 	defer os.Remove(logFile.Name())
@@ -471,6 +470,8 @@ func TestUpdateConfigDiffInAuditRecord(t *testing.T) {
 		cfg.ServiceSettings.ReadTimeout = model.NewInt(timeoutVal)
 	})
 	require.Equal(t, timeoutVal+1, *cfg.ServiceSettings.ReadTimeout)
+
+	require.NoError(t, logFile.Sync())
 
 	// Forcing a flush before attempting to read log's content.
 	err = th.Server.Log.Flush(context.Background())
--- a/api4/config_test.go
+++ b/api4/config_test.go
@@ -441,6 +441,7 @@ func TestUpdateConfigRestrictSystemAdmin(t *testing.T) {
 }
 
 func TestUpdateConfigDiffInAuditRecord(t *testing.T) {
+	t.Skip("MM-36411")
 	logFile, err := ioutil.TempFile("", "adv.log")
 	require.NoError(t, err)
 	defer os.Remove(logFile.Name())
--- a/api4/config_test.go
+++ b/api4/config_test.go
@@ -471,11 +471,11 @@ func TestUpdateConfigDiffInAuditRecord(t *testing.T) {
 	})
 	require.Equal(t, timeoutVal+1, *cfg.ServiceSettings.ReadTimeout)
 
-	require.NoError(t, logFile.Sync())
-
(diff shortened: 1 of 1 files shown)


streamer45:
As explained in https://mattermost.atlassian.net/browse/MM-36411?focusedCommentId=112849 I think it may be worth it to attempt adding a `logFile.Sync()` call before `io.ReadAll()` and see if it fixes this for good.

agnivade:
Ah thanks, I missed that :+1:

agnivade:
Hmm .. it's a bit tricky. The `File` struct takes an `io.WriteCloser`: https://github.com/mattermost/logr/blob/3d8eb9f15fcba10bcaf3c998ebd7ca5536923731/targets/file.go#L47-L49

which actually comes from `lumberjack.Logger` itself. Unless we modify that, it might be difficult to call sync on the file.

@wiggin77 - thoughts ?

streamer45:
> Hmm .. it's a bit tricky. The `File` struct takes an `io.WriteCloser`: https://github.com/mattermost/logr/blob/3d8eb9f15fcba10bcaf3c998ebd7ca5536923731/targets/file.go#L47-L49

For the purpose of the test we have direct access to the log file so I am thinking calling `Sync` on that should be sufficient?

agnivade:
Ah sure, from the comments in the ticket, I thought the recommendation was to change the logging library. Let me do that :+1:

streamer45:
> Ah sure, from the comments in the ticket, I thought the recommendation was to change the logging library. Let me do that

Yeah that would be the ideal solution but I understand it may take some time (and thinking) so it makes sense to just fix the test failure for now.

streamer45 (approved):
Awesome :+1:

api4/config_test.go
```diff
@@ -471,6 +471,8 @@ func TestUpdateConfigDiffInAuditRecord(t *testing.T) {
 	})
 	require.Equal(t, timeoutVal+1, *cfg.ServiceSettings.ReadTimeout)
 
+	require.NoError(t, logFile.Sync())
```
wiggin77, line 474: Shouldn't this happen after the flush?

agnivade: I checked that. Flush actually closes the file descriptor. So it should be synced before it.

agnivade: Ah my bad, I was checking the wrong thing. Yes, this should happen after the flush.

api4/config_test.go
```diff
@@ -471,6 +471,8 @@ func TestUpdateConfigDiffInAuditRecord(t *testing.T) {
...
+
 	// Forcing a flush before attempting to read log's content.
 	err = th.Server.Log.Flush(context.Background())
```
wiggin77, line 475: I think the issue here is that flush doesn't close the file like shutdown does.  There is a bug in the version of Logr used here that doesn't do an explicit flush during shutdown, meaning you should flush then shutdown.  This is fixed in Focalboard and I'll be bringing that to server next sprint.

agnivade: Hmm .. so do you want me to hold off from making this change?

wiggin77: No, the change is good.  I just don't know what file.Sync is going to do on a open file.  I'm suggesting you call shutdown after flush, then sync.  That should cover all bases and will continue to work in the future.

agnivade: But shutdown closes the file descriptor. So the kernel can reclaim it for other uses.

[Sync](https://pkg.go.dev/os#File.Sync) will just flush the contents to disk. Internally it just calls `fsync`  on the `fd`. So calling sync on an open file should be fine I believe.

And since `th.Teardown` already calls Shutdown, I think that should be fine. As long as the contents are synced to disk, the test should pass.

wiggin77: Ok, sounds good.

wiggin77 (approved):
LGTM

PR #17938, opened 2021-07-16
[MM-35345]: fix unreads in threads and migration

ashishbhate:
#### Summary

- Fix unreads query for threads to check deleted option
- Normalize unreads query to look for threads after given time
- Fix migration to only run once, and increment timestamp by one so that threads are definitely marked as read.

#### Ticket Link

- Fixes issues with https://mattermost.atlassian.net/browse/MM-35345
- And possibly some issues with thread unreads. 

#### Release Note

```release-note
NONE
```

### DIFF
--- a/store/sqlstore/thread_store.go
+++ b/store/sqlstore/thread_store.go
@@ -130,7 +130,6 @@ func (s *SqlThreadStore) GetThreadsForUser(userId, teamId string, opts model.Get
 		model.Post
 	}
 
-	unreadRepliesQuery := "SELECT COUNT(Posts.Id) From Posts Where Posts.RootId=ThreadMemberships.PostId AND Posts.CreateAt >= ThreadMemberships.LastViewed"
 	fetchConditions := sq.And{
 		sq.Or{sq.Eq{"Channels.TeamId": teamId}, sq.Eq{"Channels.TeamId": ""}},
 		sq.Eq{"ThreadMemberships.UserId": userId},
@@ -159,7 +158,7 @@ func (s *SqlThreadStore) GetThreadsForUser(userId, teamId string, opts model.Get
 			LeftJoin("ThreadMemberships ON Posts.RootId = ThreadMemberships.PostId").
 			LeftJoin("Channels ON Posts.ChannelId = Channels.Id").
 			Where(fetchConditions).
-			Where("Posts.CreateAt >= ThreadMemberships.LastViewed").ToSql()
+			Where("Posts.CreateAt > ThreadMemberships.LastViewed").ToSql()
 
 		totalUnreadThreads, err := s.GetMaster().SelectInt(repliesQuery, repliesQueryArgs...)
 		totalUnreadThreadsChan <- store.StoreResult{Data: totalUnreadThreads, NErr: errors.Wrapf(err, "failed to get count unread on threads for user id=%s", userId)}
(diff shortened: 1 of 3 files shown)


store/sqlstore/thread_store.go
```diff
@@ -159,7 +158,7 @@ func (s *SqlThreadStore) GetThreadsForUser(userId, teamId string, opts model.Get
 			LeftJoin("ThreadMemberships ON Posts.RootId = ThreadMemberships.PostId").
 			LeftJoin("Channels ON Posts.ChannelId = Channels.Id").
 			Where(fetchConditions).
-			Where("Posts.CreateAt >= ThreadMemberships.LastViewed").ToSql()
+			Where("Posts.CreateAt > ThreadMemberships.LastViewed").ToSql()
```
ashishbhate, line 161: We normalize all unread queries to return threads (posts) that are fully after the LastViewed time.

store/sqlstore/thread_store.go
```diff
@@ -218,6 +217,24 @@ func (s *SqlThreadStore) GetThreadsForUser(userId, teamId string, opts model.Get
...
+				sq.Expr("Posts.DeleteAt = 0"),
+			}
+		}
+
+		unreadRepliesQuery, _ := sq.
+			Select("COUNT(Posts.Id)").
+			From("Posts").
+			Where(unreadRepliesFetchConditions).
+			MustSql()
+
```
ashishbhate, line 237: Unread count should respect `deleted` option

store/sqlstore/upgrade.go
```diff
@@ -1230,6 +1230,11 @@ func upgradeDatabaseToVersion538(sqlStore *SqlStore) {
 // last viewed at time of the the thread as the last viewed at time
 // of the channel
 func fixCRTThreadCountsAndUnreads(sqlStore *SqlStore) {
+	var system model.System
+	if err := sqlStore.GetMaster().SelectOne(&system, "SELECT * FROM Systems WHERE Name = 'CRTThreadCountsAndUnreadsMigrationComplete'"); err == nil {
+		return
+	}
```
ashishbhate, line 1236: If CRT is enabled, viewing a channel should NOT mark a thread as read. This migration will mark some threads in channels as read, every time it is run. This is a problem on daily. We leverage the systems table to ensure this migration is done only once. 

We might have to toggle this once after https://mattermost.atlassian.net/browse/MM-36737 is done.

store/sqlstore/upgrade.go
```diff
@@ -1238,7 +1243,7 @@ func fixCRTThreadCountsAndUnreads(sqlStore *SqlStore) {
 	`
 	updateThreadMembershipQuery := `
 		WITH q as (` + threadMembershipsCTE + `)
-		UPDATE ThreadMemberships set LastViewed = q.CM_LastViewedAt, UnreadMentions = 0, LastUpdated = :Now
+		UPDATE ThreadMemberships set LastViewed = q.CM_LastViewedAt + 1, UnreadMentions = 0, LastUpdated = :Now
```
ashishbhate, line 1246: Bump LastViewed count by 1 so that threads don't marked as unread. This is not strictly necessary after the fixes in the thread_store, but just in case 🤞

jwilander (approved):
LGTM

PR #17939, opened 2021-07-16
[MM-32604] Remove hard-coded override of MaxNotificationsPerChannel

streamer45:
#### Summary

Another mystery to add to the collection. This was Introduced in e74c66d7f6a1f515d2380d0f163b2958f03c2bc7 . I couldn't exactly figure out why but maybe the config setting was mainly thought for enterprise customers (to limit push notifications?) and a higher default was deemed appropriate for team edition instead. Anyhow, removing this now as it should be left to the configuration.

#### Ticket

https://mattermost.atlassian.net/browse/MM-32604

#### Release Notes

```release-note
Removed hard-coded override of TeamSettings.MaxNotificationsPerChannel on unlicensed servers (e.g team edition).
```

### DIFF
--- a/app/server.go
+++ b/app/server.go
@@ -72,8 +72,6 @@ import (
 	"github.com/mattermost/mattermost-server/v5/utils"
 )
 
-var MaxNotificationsPerChannelDefault int64 = 1000000
-
 // declaring this as var to allow overriding in tests
 var SentryDSN = "placeholder_sentry_dsn"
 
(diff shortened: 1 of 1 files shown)


agnivade (approved):
Weird, the mode/config.go default is 1000. Anyways, is this necessary only for 6.0? I think this can go in normally. Existing installations shouldn't be affected as the config would be already saved.

streamer45:
> Weird, the mode/config.go default is 1000. Anyways, is this necessary only for 6.0? I think this can go in normally. Existing installations shouldn't be affected as the config would be already saved.

It's not necessary for 6.0 as it's not breaking the API. I guess I kept it here since I was worried about the possible impact on team editions. I am thinking most of the admins may not be aware of this hard-coded limit and going back to the default (or whatever is in the config) may cause issues on large installations (notifications not getting sent).

agnivade:
Yeah, makes sense.

cwarnermm:
@streamer45 - Any outstanding documentation needed for this PR from your perspective?

streamer45:
> @streamer45 - Any outstanding documentation needed for this PR from your perspective?

Nope :+1:

PR #17941, opened 2021-07-16
[MM-10166] Add MaxImageResolution config setting

streamer45:
#### Summary

PR adds a new `FileSettings.MaxImageResolution` setting to allow admins to set a custom limit to the dimension of uploaded images. We also raise the default limit to 33MP (8K resolution) which seems like a good compromise that should cover most cases.

#### Ticket

https://mattermost.atlassian.net/browse/MM-10166

#### Release Note

```release-note
Added a new FileSettings.MaxImageResolution config setting to control the maximum dimension (in pixels) of image uploads.
Removed MB constant from model package.
```

### DIFF
--- a/model/utils.go
+++ b/model/utils.go
@@ -31,7 +31,6 @@ const (
 	UppercaseLetters = "ABCDEFGHIJKLMNOPQRSTUVWXYZ"
 	NUMBERS          = "0123456789"
 	SYMBOLS          = " !\"\\#$%&'()*+,-./:;<=>?@[]^_`|~"
-	MB               = 1 << 20
 )
 
 type StringInterface map[string]interface{}
--- a/api4/file_test.go
+++ b/api4/file_test.go
@@ -542,6 +542,20 @@ func TestUploadFiles(t *testing.T) {
 				}
 			},
 		},
+
+		{
+			title:                 "Error image too large",
+			names:                 []string{"test.png"},
+			skipSuccessValidation: true,
+			checkResponse:         CheckBadRequestStatus,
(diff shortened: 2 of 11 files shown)


model/utils.go
```diff
@@ -31,7 +31,6 @@ const (
 	UppercaseLetters = "ABCDEFGHIJKLMNOPQRSTUVWXYZ"
 	NUMBERS          = "0123456789"
 	SYMBOLS          = " !\"\\#$%&'()*+,-./:;<=>?@[]^_`|~"
-	MB               = 1 << 20
```
agnivade, line 34: Good to add this in the release notes for completeness.

agnivade (approved):

isacikgoz (approved):
Looks good 💪 I have a question that is not related to this PR implementation but still I wonder why the resolution limit is needed? The file size limit sounds to be an efficient check.

streamer45:
> Looks good  I have a question that is not related to this PR implementation but still I wonder why the resolution limit is needed? The file size limit sounds to be an efficient check.

Mainly because, due to encoding, there isn't much correlation between the two values. We can have a 4KB PNG file that's 30MPx. Once decoded (and we need to in order to generate previews) that can be worth hundreds of MBs of memory.

furqanmlk (approved):
`FileSettings.MaxImageResolution=10000001`

### 1.99 MegaPix (1881x1058)
image

### 2 MegaPix (1821x1098)
image

### 0.99 MegaPix (1282x773)
image


### 0.5 MegaPix (708x708)
image

PR #17945, opened 2021-07-18
Cleanup model/utils.go

agnivade:
- Removed some functions which were unsued. (The unused linter was somehow failing to catch this).
- Moved some functions inside other packages.
- Removed two duplicate instances of the same function - RemoveDuplicateStrings and UniqueStrings.
- Used a more efficient in-place version of RemoveDuplicateStrings.
- Moved some regexes to global vars to prevent compilation every time.

```release-note
Removed the following functions:
- UniqueStrings
- IsValidUserNotifyLevel
- IsValidTrueOrFalseString
- IsValidPushStatusNotifyLevel
- IsValidCommentsNotifyLevel
- IsValidEmailBatchingInterval
- IsSamlFile
- IsSafeLink
- IsDomainName
- AsStringBoolMap
- IsValidAlphaNum
- StringToJson
- StringFromJson
- IsLower
- IsValidTurnOrStunServer
- IsValidWebsocketUrl
- IsValidNumberString
```


model/utils.go
```diff
@@ -575,73 +512,25 @@ func IsValidId(value string) bool {
...
-	out := []string{}
-	seen := make(map[string]bool, len(in))
-
-	for _, item := range in {
-		if !seen[item] {
-			out = append(out, item)
-
-			seen[item] = true
+	// In-place de-dup.
+	// Copied from https://github.com/golang/go/wiki/SliceTricks#in-place-deduplicate-comparable
```
streamer45, line 517: I am all for optimizations but unless we start dealing with thousands of strings at a time does it really matter? Aside from the added complexity (which is not that bad anyway) what slightly worries me is the change in input/output access. Before we were treating the input as immutable and returning a new copy of the data. I think that's usually a good trade-off since it's safer as the caller doesn't have to be particularly careful.

That said, I am okay with keeping the new logic, but maybe I'd add some documentation explicitly saying the operation is done in-place and as a result the input data can be altered. And maybe consider not returning anything so that's made even clearer.

streamer45 (commented):
Very nice, tidying up things :+1: Left some minor comments/suggestions.

api4/post_test.go
```diff
@@ -881,6 +881,7 @@ func TestPatchPost(t *testing.T) {
 		rpost, resp = Client.PatchPost(post.Id, patch)
 		CheckNoError(t, resp)
 
+		sort.Strings(fileIds[0:2])
```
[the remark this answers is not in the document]
agnivade: This is because the `RemoveDuplicateStrings` sorts the strings. Hence the output is sorted. I don't believe the API makes any sort of guarantees that the output will be in a fixed order. So the in-place de-dup should be fine.

> Wondering if we could move it after the initial for loop and sort the whole thing?

Yeah sure. :+1:

model/utils.go
agnivade: Funnily enough, we were doing `slice = model.RemoveDuplicateStrings(slice)` in a lot of places :laughing:

model/utils.go
```diff
@@ -575,73 +512,24 @@ func IsValidId(value string) bool {
...
+	for i := 1; i < len(in); i++ {
+		if in[j] == in[i] {
+			continue
 		}
+		j++
+		in[j] = in[i]
 	}
-
-	return out
+	in = in[:j+1] //nolint:staticcheck
```
streamer45, line 532: Not sure this is going work actually :/ I think we cannot propagate the change in length since we are technically handling a copy of the given slice (although sharing the same backing data).

agnivade: Ahh you are right actually. So basically we will need to document the changes. And just mention that the returned slice is reused.

streamer45: Indeed.

streamer45 (approved):
:+1:

PR #17948, opened 2021-07-19
[MM-27312] Allow uploading 10 files per post

streamer45:
#### Summary

We can finally go ahead with this. Bumping from 5 to 10 the limit of file attachments per post.

#### Ticket

https://mattermost.atlassian.net/browse/MM-27312

#### Release Note

```release-note
Increased the limit of uploaded file attachments per post from 5 to 10.
```

#### Related PRs

https://github.com/mattermost/mattermost-webapp/pull/8419


store/sqlstore/upgrade.go
```diff
@@ -1297,3 +1298,13 @@ func fixCRTChannelMembershipCounts(sqlStore *SqlStore) {
 		mlog.Error("Error marking migration as done", mlog.Err(err))
 	}
 }
+
+func upgradeDatabaseToVersion600(sqlStore *SqlStore) {
+	// if shouldPerformUpgrade(sqlStore, Version5380, Version600) {
+
+	// allow 10 files per post
+	sqlStore.AlterColumnTypeIfExists("Posts", "FileIds", "text", "varchar(300)")
```
agnivade, line 1306: Do you think it's worth to diverge the schema to use native array types for Postgres? We can query the data more efficiently in those cases.

The querying side might have to change, so there will be a cost I think.

streamer45: Technically we could have converted to JSON this one as well since that's what we store in it but I opted not to since we have plans to remove the column ([MM-32754](https://mattermost.atlassian.net/browse/MM-32754)), although I am not 100% convinced yet and currently thinking of deferring it to a later time.

agnivade (approved):

hmhealey:
It doesn't affect this PR specifically, but I forgot that I wanted to suggest deprecating Post.FileIds as something that's stored in the database and sent back to users with 6.0. We might be able to stop using this field entirely since we already store the relation the other way around with FileInfo.PostId.

Prior to adding post metadata, we needed the Post object to include the number of file attachments on the post so that we could leave space for the components in the post list, and I did that by storing the FileIds on the post since it seemed helpful to have the IDs for other things. But since adding post metadata, we now send all attached FileInfo objects to the client anyway, so Post.FileIds is redundant.

We still need the client to send up Post.FileIds when creating a post so we can set FileInfo.PostId, but we wouldn't need to store it in the DB or send it back to the client in response any more.

streamer45:
@hmhealey 
You didn't forget, I am tracking it at https://mattermost.atlassian.net/browse/MM-32754. I am just not confident enough to pull the trigger on this especially since we are most probably leaving behind other things (`Posts.Filenames`).

streamer45:
This has been open for so long I literally forgot about it. This is an additional schema change on the `Posts` table. It won't affect community though cause it's been done already in the past (and then the PR was reverted). Just so you are aware @spirosoik .

I'll make sure to include this in the migration analysis as well and also consider combining it with the other two statements affecting this table so to minimize the execution time to a single full-table copy.

spirosoik:
@streamer45 yeah seems safe to me

streamer45:
@amyblais This will need cherry picking as well (and related webapp PR).

PR #17965, opened 2021-07-20
[MM-37118] Don't mark channel as read for reply posts with CRT on

ashishbhate:
#### Summary

- With CRT on, posting a reply to a thread should NOT mark the channel containing the the thread as read

#### Ticket Link

- https://mattermost.atlassian.net/browse/MM-37118
 
#### Release Note

```release-note
NONE
```


jwilander:
>So this is going to change the behaviour of using the RHS with CRT enabled as well since we don't know whether or not the user is using the threads view to do this. Is that going to be a problem at all?

I don't think it's a problem and is by design. Basically replying to a thread in the RHS for thread view should never mark a channel as read with CRT on.

app/post.go
```diff
@@ -87,7 +87,7 @@ func (a *App) CreatePostAsUser(c *request.Context, post *model.Post, currentSess
 	// or if it does not have from_bot set (e.g. from discovering the user is a bot within CreatePost).
```
jwilander, line 87: Could you update the comment to explain the CRT related part?

ashishbhate:
>So this is going to change the behaviour of using the RHS with CRT enabled as well since we don't know whether or not the user is using the threads view to do this. Is that going to be a problem at all?

As Joram said, this is by design. I can share a spreadsheet with you that defines unread and badging behaviour. In short - threads and channels unreads and badges are tracked separately. Replies in threads (unreads and badging) only affects threads (the sidebar entry, and the global threads screen) and not the channel.

jwilander (approved):

jgilliam17:
Thanks @ashishbhate 
Replying from a thread on Global Threads no longer marks the channel as unread, i - looks good.
Seeing few other issues:
1. When a new post is made by other users, there is a delay on the test server before that channel is bolded. Tested few times and it ranged from few seconds to 2 minutes. This may be test server issue, noting here just in case it's related to your changes
Screen Shot 2021-07-22 at 10 35 46 AM

2. There is an error on the console. Happened few more times not sure of exact steps
Screen Shot 2021-07-22 at 10 41 07 AM

3. After new post becomes a thread and current user is mentioned on that thread (auto-following) and replies to it on global threads - channel is not marked as read on the LHS. Steps:

user1: belongs to both channel1 and channel2, viewing channel1 (have all channels as read, unbolded)
user2: post to channel2 
user1: observes channel2 is marked as unread/bolded 
user2: reply to the same post in channel2 and mention user1
user1: observes LHS Threads item is bolded, opens the global threads and replies to post they were mentioned in
user1: observe channel2 remains bolded/unread even though they have seen and replied to the new post

ashishbhate:
>When a new post is made by other users, there is a delay on the test server before that channel is bolded. Tested few times and it ranged from few seconds to 2 minutes. This may be test server issue, noting here just in case it's related to your changes

I can't repro on the test server, or on my local machine

>There is an error on the console. Happened few more times not sure of exact steps

I don't think this PR is causing this issue, but I'm looking into it.

>After new post becomes a thread and current user is mentioned on that thread (auto-following) and replies to it on global threads - channel is not marked as read on the LHS. Steps:
>user1: belongs to both channel1 and channel2, viewing channel1 (have all channels as read, unbolded)
>user2: post to channel2
>user1: observes channel2 is marked as unread/bolded
>user2: reply to the same post in channel2 and mention user1
>user1: observes LHS Threads item is bolded, opens the global threads and replies to post they were mentioned in
>user1: observe channel2 remains bolded/unread even though they have seen and replied to the new post

Isn't this expected behaviour - a reply post (with CRT enabled) should not mark a channel as unread.

jgilliam17:
@ashishbhate Additional info for the 3rd point - the channel remains unread even though user saw the new post in global threads and replied.
Here is a short video; The issue I described happens at about 00:26 point.
I expected the channel to be read/unbolded as sysadmin opens the thread with the mention in global threads and basically reads all the new posts from this channel.

https://user-images.githubusercontent.com/52937121/126799579-58c9bca6-5110-4654-b780-6de9a9d7a6ec.mov

jgilliam17:
Thanks @ashishbhate - after checking with Eric, we should fix the missing new message line on the channel that remains bolded on the LHS

ashishbhate:
@jgilliam17 I'm not able to repro the "400 bad request" error you're seeing in the console. I checked the code paths that could return the error, but I couldn't find anything obvious that could be causing the issue.

ashishbhate:
@jgilliam17 the missing new line issue is fixed in a separate PR in the webapp repo https://github.com/mattermost/mattermost-webapp/pull/8489

jgilliam17 (approved):
Thanks @ashishbhate 
Tested, looks good to merge.
- Verified channel is not marked as read when replying to post on global threads. 
The errors are no longer reproducing and new line will be fixed in a separate PR.

ashishbhate:
@hmhealey gentle bump for review

hmhealey (approved):
Sorry about that. I thought I had actually approved this

PR #17973, opened 2021-07-21
MM-36271: Bump major version to 6.0

agnivade:
https://mattermost.atlassian.net/browse/MM-36271

```release-note
We bump the major version to 6.0
```


streamer45:
Were we planning of bumping model and schema version as well or keeping those in a separate PR?

agnivade:
Ah good point. Let me do that.

agnivade:
Done.

streamer45 (approved):
Awesome, thanks :+1:

agnivade:
@streamer45 - the store schema version upgrade has a problem because there is a unit test which upgrades the DB schema and verifies it matches with `CurrentSchemaVersion`. Which essentially means we need to have the line `if shouldPerformUpgrade(sqlStore, Version5380, Version6000) {`.

Given that we don't even have the line for 5380 yet, I would defer it from being bumped until we actually cut v6.

Thoughts?

streamer45:
> @streamer45 - the store schema version upgrade has a problem because there is a unit test which upgrades the DB schema and verifies it matches with `CurrentSchemaVersion`. Which essentially means we need to have the line `if shouldPerformUpgrade(sqlStore, Version5380, Version6000) {`.
> 
> Given that we don't even have the line for 5380 yet, I would defer it from being bumped until we actually cut v6.
> 
> Thoughts?

I think we should remove the if blocks from 5.38 since that will be based of Cloud which has been merged already and then proceed normally. Also, we just merged 6.0 branch into master so make sure to update the base branch for this.

metanerd:
Changed base branch to master, otherwise I am not able to delete release-6.0 branch.

PR #17989, opened 2021-07-22
MM-36545 Fix replica lag issue causing missed websocket events to update following

jwilander:
#### Summary
When user A commented on another user B's root post that had no previous replies, user B should receive a websocket event updating the thread, particularly the following status, for that root post. The thread membership for user B is created as part of the notification code. Further down in that same notification code that thread membership was attempted to be read from the database again. If read replicas are configured this is a classic case for replica lag to cause the thread membership to not seem to exist, which leads to user B not getting the websocket event.

The fix was to maintain references to the thread memberships for use later in the SendNotifications function instead of re-querying the database for them.

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-36545

#### Release Note

```release-note
Fix root posts not being shown as followed for the post creator after receiving the first reply that affected servers with collapsed reply threads turned on and database read replicas configured.
```


app/notification.go
```diff
@@ -203,6 +204,7 @@ func (a *App) SendNotifications(post *model.Post, team *model.Team, channel *mod
 					}
 
 					if membership != nil && !membership.Following {
+						participantMemberships[userID] = membership
```
agnivade, line 207: This is a data race - we are writing to this map from multiple goroutines. Please guard this with a mutex.

app/notification.go
```diff
@@ -474,9 +477,10 @@ func (a *App) SendNotifications(post *model.Post, team *model.Team, channel *mod
 			}
 			if sendEvent {
 				message := model.NewWebSocketEvent(model.WebsocketEventThreadUpdated, team.Id, "", uid, nil)
-				threadMembership, err := a.Srv().Store.Thread().GetMembershipForUser(uid, post.RootId)
-				if err != nil {
-					return nil, errors.Wrapf(err, "cannot get thread membership %q for user %q", post.RootId, uid)
+				threadMembership := participantMemberships[uid]
+				if threadMembership == nil {
```
agnivade, line 481: Is it expected that `threadMembership` can be nil? If so, I'd suggest to tune it down to debug rather than warn. If not, then warn is fine.

agnivade (commented):
As an aside, I just realized that we are spawning as many goroutines as the number of `threadParticipants` in line 191. And each goroutine can make up to 2 DB calls. In a scenario with 5000 participants, this code will make 10000 DB calls simultaneously and bring down the DB.

This is the first item in our style guide: https://developers.mattermost.com/contribute/server/style-guide/#default-to-sync-instead-of-async. Do not use goroutines until you understand their complete behavior :P Anyways, I have digressed. But this is a ticking time-bomb which we should fix.

ashishbhate:
>. Do not use goroutines until you understand their complete behavior 

We do understand the behaviour - its very bad ;) 

The `SendNotifications` function is on my radar for cleaning up, hopefully after the migrations stuff is done.

app/notification.go
jwilander: Right, will do

app/notification.go
jwilander: It should never be nil.

jwilander:
As part of that I'm not convinced we need to send the `WebsocketEventThreadUpdated`. I'm pretty sure any client receiving the new post websocket event should have enough information to construct the thread. That would be a fairly large refactor and require some investigation but it might be a way to reduce the complexity here and the number of websocket broadcasts.

@ashishbhate can you make a ticket for the goroutine issue @agnivade raised? We should tackle that before GA of CRT.

ashishbhate:
@jwilander I made this ticket https://mattermost.atlassian.net/browse/MM-37368 a little while earlier. Do you want me to make a separate ticket? I was going to bring that ticket up during the next planning meeting for triage.

jwilander:
That's good enough since it calls out the goroutine issue but probably better to split the two items in the ticket off into their own tickets so we can track separately

agnivade (approved):

app/notification.go
```diff
@@ -474,9 +483,10 @@ func (a *App) SendNotifications(post *model.Post, team *model.Team, channel *mod
...
+					a.Log().Warn("Missing thread membership for participant in notifications.", mlog.String("user_id", uid), mlog.String("thread_id", post.RootId))
+					continue
```
ashishbhate, line 489: `participantMemberships` will only contain members who have been mentioned in the post. The loop here notifies all followers of the thread. A person who manually followed the thread wont be mentioned (or be a participant), so the won't get notified here. I think for `threadMembership == nil` we need to query the DB for that membership instead of just logging the error.

ashishbhate (approved):

jwilander:
@amyblais I'd actually like this one to soak on community for a bit first before going to cloud. I can cherry-pick afterwards.

PR #17995, opened 2021-07-23
MM-36448: Removes legacy CLI commands.

mkraft:
#### Summary

Removed the following deprecated CLI commands:

* `channel`
* `command`
* `config`
* `extract`
* `group`
* `import`
* `integrity`
* `jobserver`
* `ldap`
* `license`
* `logs`
* `permissions`
* `plugin`
* `reset`
* `roles`
* `sampledata`
* `team`
* `user`
* `webhook`

The remaining commands are:

* `db`
* `export`
* `server`
* `version`

#### Ticket Link

https://mattermost.atlassian.net/browse/MM-36448

#### Release Note

```release-note
Removed the following deprecated CLI commands:
* channel
* command
* config
* extract
* group
* import
* integrity
* jobserver
* ldap
* license
* logs
* permissions
* plugin
* reset
* roles
* sampledata
* team
* user
* webhook
```


wiersgallak (approved):
LGTM

streamer45:
@mkraft This is going to break a couple of scripts mainly due to removing the "version" command. We should consider implementing a solution before merging.

BenCookie95:
> @mkraft This is going to break a couple of scripts mainly due to removing the "version" command. We should consider implementing a solution before merging.

Also might be worth broadcasting in a channel that these are being deleted just in case they're being used somewhere we're unaware of

mkraft:
> @mkraft This is going to break a couple of scripts mainly due to removing the "version" command. We should consider implementing a solution before merging.

I've re-added `version` command for now. It's being used in some scripts as a way to migrate the database, with the output of version is being discarded. One option would be to move that into a `db migrate` command that doesn't even output the version. Thoughts?

mkraft:
> > > @mkraft This is going to break a couple of scripts mainly due to removing the "version" command. We should consider implementing a solution before merging.
> > 
> > 
> > I've re-added `version` command for now. It's being used in some scripts as a way to migrate the database, with the output of version is being discarded. One option would be to move that into a `db migrate` command that doesn't even output the version. Thoughts?
> 
> Yes, that's what we were talking about, relying on implicit side effects. Big +1 on adding a dedicated command that does what it says.

Something like https://github.com/mattermost/mattermost-server/compare/MM-36448_remove-legacy-cli-commands...MM-36448_remove-legacy-cli-commands-with-migrate.

streamer45:
> Something like [MM-36448_remove-legacy-cli-commands...MM-36448_remove-legacy-cli-commands-with-migrate](https://github.com/mattermost/mattermost-server/compare/MM-36448_remove-legacy-cli-commands...MM-36448_remove-legacy-cli-commands-with-migrate).

Yeah that would work. Actually I don't think we need to start up the server at all. We just need to create the store, similarly to what `db init` does.

BenCookie95 (approved):
Is there a list of the commands removed? Maybe we should add that to the release notes

mkraft:
> Is there a list of the commands removed? Maybe we should add that to the release notes

Added them, thanks.

mkraft:
> > Something like [MM-36448_remove-legacy-cli-commands...MM-36448_remove-legacy-cli-commands-with-migrate](https://github.com/mattermost/mattermost-server/compare/MM-36448_remove-legacy-cli-commands...MM-36448_remove-legacy-cli-commands-with-migrate).
> 
> Yeah that would work. Actually I don't think we need to start up the server at all. We just need to create the store, similarly to what `db init` does.

Looks like it'll have to be in a separate PR as we're out of time for v6.

streamer45 (approved):

furqanmlk (approved):
E2E Passed (90% passed, 10% is expected failures in Prod)
https://automation-dashboard.vercel.app/cycles/76a8f17c-8cb3-4182-8c71-3f540ff2508f

streamer45:
@mkraft 

I've reintroduced the `jobserver` command in https://github.com/mattermost/mattermost-server/pull/18220 since we still need to support dedicated job servers. We should either cherry-pick this to cloud with that included or add the linked PR on top of it. Let me know what you prefer and if I can help with it.

amyblais:
> @mkraft
> 
> I've reintroduced the `jobserver` command in #18220 since we still need to support dedicated job servers. We should either cherry-pick this to cloud with that included or add the linked PR on top of it. Let me know what you prefer and if I can help with it.

A cherry-pick PR is open here: https://github.com/mattermost/mattermost-server/pull/18216

PR #17998, opened 2021-07-24
MM-25516: Changed to byte slice instead of string for cluster messages

agnivade:
https://mattermost.atlassian.net/browse/MM-25116

Testing:
Manually tested.
Load-tested with Cluster Controller.

I looked into changing the serialization method to use msgpack,
but the ClusterMessage struct was mainly used for only 3 fields
which didn't lead to much of a CPU time improvement, whereas
actually led to more allocations using msgpack. Hence, I chose
to remain with JSON.

```
name              old time/op    new time/op    delta
ClusterMarshal-8    3.51µs ± 1%    3.10µs ± 2%  -11.59%  (p=0.000 n=9+10)

name              old alloc/op   new alloc/op   delta
ClusterMarshal-8      776B ± 0%     1000B ± 0%  +28.87%  (p=0.000 n=10+10)

name              old allocs/op  new allocs/op  delta
ClusterMarshal-8      12.0 ± 0%      13.0 ± 0%   +8.33%  (p=0.000 n=10+10)
```

```release-note
Changed the field type of Data in model.ClusterMessage to []byte from string.
```


app/busy.go
```diff
@@ -109,11 +110,12 @@ func (b *Busy) notifyServerBusyChange(sbs *model.ServerBusyState) {
 	if b.cluster == nil {
 		return
 	}
+	buf, _ := json.Marshal(sbs)
```
agnivade, line 113: I didn't want to check the error just yet. Kept the scope of the PR just to changing the field type.

app/plugin_event.go
```diff
@@ -4,16 +4,19 @@
 package app
 
 import (
+	"encoding/json"
+
 	"github.com/mattermost/mattermost-server/v6/model"
 )
 
 func (s *Server) notifyClusterPluginEvent(event model.ClusterEvent, data model.PluginEventData) {
+	buf, _ := json.Marshal(data)
```
agnivade, line 13: Same

streamer45 (approved):
Looks good :+1:

app/busy.go
streamer45: Let's just make sure to track this.

agnivade: https://mattermost.atlassian.net/browse/MM-37409

agnivade:
@amyblais - Just a quite note, this PR requires an upgrade note for v6.

PR #17999, opened 2021-07-26
MM-22051: Remove ToJson methods from network writes

agnivade:
We replace the double conversion of
[]byte to string, with a direct write
to http.ResponseWriter.

https://mattermost.atlassian.net/browse/MM-22051

Tried using gofmt -r, but it only accepts Go
expressions. So had to resort to an ugly sed replace

sed -E -i 's/w.Write\(\[\]byte\((.*).ToJson\(\)\)\)/if err := json.NewEncoder\(w\).Encode\(\1\); err != nil { mlog.Warn\("Error while writing response", mlog.Err\(err\)\)}/g' *.go

```release-note
NONE
```

### DIFF
--- a/api4/bot.go
+++ b/api4/bot.go
@@ -13,6 +13,7 @@ import (
 
 	"github.com/mattermost/mattermost-server/v6/audit"
 	"github.com/mattermost/mattermost-server/v6/model"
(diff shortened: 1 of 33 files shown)


streamer45 (approved):
One liners FTW :muscle:

PR #18008, opened 2021-07-27
[MM-36318] Set bot profile image and icon only if they have changed

streamer45:
#### Summary

Spikes in `getChannelMember` calls have not reduced during community servers restart. Had a second look and it appears the `user_updated` event are also triggered when setting bot profile image and icon which happen as part of the `EnsureBot` API call. PR addresses this by adding an equality check on the data before going ahead with the update.

#### Ticket

https://mattermost.atlassian.net/browse/MM-36318

#### Release Note

```release-note
NONE
```


app/plugin_api.go
```diff
@@ -661,11 +661,16 @@ func (api *PluginAPI) GetProfileImage(userID string) ([]byte, *model.AppError) {
...
 
 func (api *PluginAPI) SetProfileImage(userID string, data []byte) *model.AppError {
-	_, err := api.app.GetUser(userID)
-	if err != nil {
+	if _, err := api.app.GetUser(userID); err != nil {
 		return err
 	}
 
+	if storedData, err := api.app.ReadFile("users/" + userID + "/profile.png"); err == nil {
+		if buf, err := api.app.AdjustImage(bytes.NewReader(data)); err == nil && bytes.Equal(storedData, buf.Bytes()) {
```
agnivade, line 669: `SetProfileImageFromFile` already calls `AdjustImage`. Should we move the logic into that? Otherwise this just calls `AdjustImage` twice which looks like a costly operation.

streamer45: We can do that. I was slightly worried of extending the check to the underlying function cause that would affect everything, not only bots initializations.

agnivade: Should be fine IMO. A lot of our store code has similar logic for return-if-same.

streamer45: Well we wouldn't know until we do it and break something anyway :p Will update.

isacikgoz (approved):
LGTM 👍

agnivade (approved):

furqanmlk (commented):
@streamer45 I am getting error in `bot.got` file

image

agnivade:
@furqanmlk - You would have to upgrade to use Go 1.16.

@streamer45 - Wondering if we should just use `ioutil.ReadAll` for now? I'd prefer to fix all `ReadAll`s at once, and also add a vet check to urge developers to use the new function. And we would also need to bump the go.mod version if we want to use 1.16's features, otherwise developers will keep getting these errors. Might be better to just keep using 1.15 for now and do this in a separate PR. 0/5.

furqanmlk (approved):
Thanks @streamer45  
Working as Expected.

PR #18011, opened 2021-07-27
[MM-37202] Use native upsert query in StatusStore.SaveOrUpdate()

streamer45:
#### Summary

Replacing the two queries (select + insert) which were causing some errors (seen a lot of these during recent load-tests) with a single upsert query.

#### Ticket

https://mattermost.atlassian.net/browse/MM-37202

#### Release Note
```release-note
NONE
```


agnivade (approved):
Ah you are doing our team's work now :)

streamer45:
> Ah you are doing our team's work now :)

Old habits, or maybe planning for the future :roll_eyes:

isacikgoz (approved):
LGTM 👍

streamer45:
@furqanmlk 

A smoke-test around changing user statuses should be sufficient here.

srkgupta:
Creating Spinwick for my testing purpose. Not related to this PR.

furqanmlk (approved):
User Status E2E Passed ✅ 

image

image

image

image

image

image

image

store/sqlstore/status_store.go
```diff
@@ -39,18 +38,29 @@ func (s SqlStatusStore) createIndexesIfNotExists() {
...
-			return errors.Wrap(err, "failed to update Status")
-		}
+func (s SqlStatusStore) SaveOrUpdate(st *model.Status) error {
+	query := s.getQueryBuilder().
+		Insert("Status").
+		Columns("UserId", "Status", "Manual", "LastActivityAt", "DNDEndTime", "PrevStatus").
+		Values(st.UserId, st.Status, st.Manual, st.LastActivityAt, st.DNDEndTime, st.PrevStatus)
+
+	if s.DriverName() == model.DatabaseDriverMysql {
+		query = query.SuffixExpr(sq.Expr("ON DUPLICATE KEY UPDATE UserId = ?, Status = ?, Manual = ?, LastActivityAt = ?, DNDEndTime = ?, PrevStatus = ?",
```
agnivade, line 48: Sorry, I missed this previously, but do we need to update the userId here and the below? Seems redundant as that's the primary key here.

streamer45: Yes that's redundant given it will be the same value so it can be safely removed.

PR #18016, opened 2021-07-27
MM-37299 - add feature flag for invite members ab testing

pablovelezvidal:
#### Summary
This PR adds the feature flag for determine which type of invite members button will be shown in the UI and allow to perform the A/B testing of this feature.

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-37299

#### Release Note
```release-note
NONE
```


model/feature_flags.go
```diff
@@ -41,6 +41,9 @@ type FeatureFlags struct {
 
 	// Enable the Global Header
 	GlobalHeader bool
+
+	// Enable the Invite Members Call To Action button on the left panel
+	InviteMembersCTAButton string
```
nickmisasi, line 46: If this is only going to have 2 states (on or off) it should probably just be a boolean `true` for enabled `false` for disabled

nickmisasi (changes requested):

model/feature_flags.go
pablovelezvidal: This is going to be a string value that could have four values = "none", "sticky", "lhs_button", "user_icon".

marianunez: First, would suggest dropping `CTA` for the name given that it is redundant with `Button` so I would rename to `InviteMembersButton`

I was going to suggest to define the values here and have this be of that type but a) I think Split.io implementation here may not support something else than `string` b) there seems no value in having these constants here given that they are only used in the front end - there we should have them defined as constants.

model/feature_flags.go
```diff
@@ -41,6 +41,9 @@ type FeatureFlags struct {
 
 	// Enable the Global Header
 	GlobalHeader bool
+
+	// Enable the Invite Members button on the left panel, possible values = ("none", "sticky", "lhs_button", "user_icon")
```
marianunez, line 45:
```suggestion
	// Enable A/B test for the Invite Members button on the left panel, possible values = ("none", "sticky", "lhs_button", "user_icon")
```

marianunez (approved):
LGTM 🚢 it!

nickmisasi (approved):

PR #18020, opened 2021-07-28
[MM-36544][MM-37439] Don't re-follow on reply to unfollowed thread

ashishbhate:
#### Summary

- If a user has unfollowed a thread, another user's reply in the thread should not cause the first user to re-follow the thread. The first user will be re-followed if they are mentioned in the thread.
- Simplify and make flexible the logic surrounding following and remove some duplicate code.
   
#### Ticket Link

- https://mattermost.atlassian.net/browse/MM-36544
- https://mattermost.atlassian.net/browse/MM-37439
 
#### Release Note

```release-note
Fix issue with users re-following previously unfollowed thread when other users reply to the thread
```


app/notification.go
```diff
@@ -212,14 +212,16 @@ func (a *App) SendNotifications(post *model.Post, team *model.Team, channel *mod
...
 				}
+
+				updateFollowing := *a.Config().ServiceSettings.ThreadAutoFollow
 				if mentionType == ThreadMention || mentionType == CommentMention {
 					incrementMentions = false
+					updateFollowing = false
 				}
-
 				opts := store.ThreadMembershipOpts{
 					Following:             true,
```
jwilander, line 222: Not blocking but it feels a little weird to me that `Following` is true here even though it won't get updated to true since `UpdateFollowing` is false. Maybe it would make sense to change `Following` to a pointer and have it do the update any time it's not `nil` instead of having a separate `UpdateFollowing` field. Not pushing for any changes here, just something to think about for the future.

jwilander (approved):

app/notification.go
streamer45: That would remove a variable from the equation but wouldn't simplify the logic since it would result in encapsulating an additional state (being nil) into `Following`. I am thinking the additional `UpdateFollowing` field probably gives a bit more explicitness on the calling side and visibility on the underlying store logic. 

That said, I see how it can get confusing. I tend to think at `Following` as the state we are aiming to set and `UpdateFollowing` as the permission needed to do it so it makes sense to have the former `true` and the latter `false`.

streamer45 (approved):
Nice work navigating all corner cases :+1:

app/notification.go
jwilander: That's fair about the explicitness. It's just a little weird to me that if we just set `Following` it's a no-op if `UpdateFollowing` isn't set - so what's even the point of setting `Following`. That said, the pointer might be too implicit and as long as the whole thing is documented maybe it's safer to just keep it more explicit.

streamer45: I guess the way the code is currently written it allows for `UpdateFollowing` to be dynamic. Alternatively we could fill the `opts` object with only the needed fields for each case.

ashishbhate: `UpdateFollowing` is required when we don't want to change the following state . We can't rely on the zero value of  `ThreadMembershipOpts.Following` because it could set the following state to `false` (the zero value).

We can completely drop `UpdateFollowing` by doing a DB read `GetThreadMembershipForUser` and set the `Following` value based on the returned membership, but this would be inefficient.

To make the code simpler we could invert `UpdateFollowing` to `SkipUpdateFollowing`. This has two benefits
1. Makes the option even more explicit
2. Currently every time we want to change the following state we have to set two options `Following` and `UpdateFollowing`. By inverting `UpdateFollowing` to `SkipUpdateFollowing`, devs will only have to set the `Following` value if they want to change following state. In the rarer case of not needing a `Following` change, the dev can set `SkipUpdateFollowing`. On the other hand this might be "gotcha", because it sets the following state to `false` if both `Following` and `SkipUpdateFollowing` are omitted.

ogi-m (approved):
LGTM 👍🏻

app/notification.go
streamer45: Personally I am okay with the current approach. We can revisit this during a later refactor if needed.

amyblais:
@ashishbhate The cherry-pick PR is failing, maybe this needs a manual cherry-pick to ``release-5.39``?

ashishbhate:
@amyblais the changes in this PR will need to be backported to v5.39. I'll do it by EoD tomorrow.

amyblais:
@ashishbhate If the cherry-pick is too complex, we can just have this for v6.0. It's not urgent for v5.39, just a potential cherry-pick candidate. Just let me know what you think once you start to investigate the cherry-pick!

ashishbhate:
Cherry-pick for v5.39 done in https://github.com/mattermost/mattermost-server/pull/18325

Issue #18004, opened 2021-07-27
Fix staticcheck errors from _test.go files


agnivade:
#### Summary

Remove the exclude section from .golangci.yml file for staticcheck

```
- linters:
      - staticcheck
      path: ".*_test.go|store/storetest"
```

Run `make golangci-lint`

Fix all the issues.

Please feel free to ping here for any questions.

yedamao:
I'd like to work on this issue.

haardikdharma10:
@yedamao, sure, please go ahead. Let us know if you have any questions 🙂

PR #18033, opened 2021-07-30
[GH-18004] Fix staticcheck errors from _test.go files

yedamao:
#### Summary

Fix staticcheck errors from _test.go files

#### Ticket Link

Fixes https://github.com/mattermost/mattermost-server/issues/18004
#### Release Note

```release-note
None
```


api4/config_test.go
```diff
@@ -63,8 +63,8 @@ func TestGetConfigWithAccessTag(t *testing.T) {
 	th := Setup(t)
 	defer th.TearDown()
 
-	varyByHeader := *&th.App.Config().RateLimitSettings.VaryByHeader // environment perm.
-	supportEmail := *&th.App.Config().SupportSettings.SupportEmail   // site perm.
+	varyByHeader := th.App.Config().RateLimitSettings.VaryByHeader // environment perm.
+	supportEmail := th.App.Config().SupportSettings.SupportEmail   // site perm.
```
agnivade, line 67: This is incorrect. `SupportEmail` is a pointer. We should dereference that, and use the pointer to the variable in line 70. 

Actually the entire `defer th.App.UpdateConfig` part is unnecessary. We can remove it.

api4/user_test.go
```diff
@@ -5502,11 +5502,13 @@ func TestGetThreadsForUser(t *testing.T) {
 
 		var rootIds []*model.Post
 		for i := 0; i < 30; i++ {
+			// nolint:staticcheck
 			time.Sleep(1)
```
agnivade, line 5506: Can we please remove all of these `time.Sleep(1)` calls? I think all of these are redundant. If it surfaces any underlying issues in the test, we would want to fix that rather than using these sleeps.

api4/system_test.go
```diff
@@ -683,11 +683,13 @@ func TestGetServerBusyExpires(t *testing.T) {
 	th.App.Srv().Busy.Set(time.Second * 30)
 
 	t.Run("as system user", func(t *testing.T) {
+		// nolint:staticcheck
```
agnivade, line 686: What was the warning here?

model/bot_test.go
```diff
@@ -246,7 +246,15 @@ func TestBotPreSave(t *testing.T) {
 		DeleteAt:       0,
 	}
 
-	originalBot := &*bot
+	originalBot := &Bot{
```
agnivade, line 249: Nice, I like this approach. Very readable.

agnivade (commented):
Nice work! Please have a look at the comments.

api4/config_test.go
yedamao: I do't understand the `*&`, I made some test. 
image

image

looks like the result is not difference.

agnivade: I meant to dereference the pointer, and just store the value in a variable. And then later, store the pointer to that in the config field.

But it doesn't matter anyways since the defer clause is unecessary. So let's remove that.

yedamao: aha, I got it.

api4/system_test.go
yedamao: This is for the deprecated GetServerBusyExpires method, I will Use GetServerBusy.

api4/user_test.go
yedamao: I remove all of these time.Sleep(1), the test was passed in local.

api4/system_test.go
agnivade: Ah okay, let's keep the `nolint` comment. I'll send a separate PR to remove the method.

agnivade:
@yedamao - I have removed the deprecated method. Please update your branch with the latest master at your convenience. Thanks!

yedamao:
@agnivade I have rebased with upstream master.

yedamao:
ping

agnivade:
Apologies @yedamao. We also need to fix the errors for our EE repo. I have sent out a PR for it. When that's merged, we can go ahead with this.

isacikgoz (approved):
Nice 🎉

agnivade (approved):
Great work!

PR #18035, opened 2021-07-30
[MM-36444] Manage config deprecations for 6.0

BenCookie95:
#### Summary
Removed all of the following configs and cleaned up any code around them:
```
- EnableOnlyAdminIntegrations
- RestrictCustomEmojiCreation
- RestrictPostDelete
- AllowEditPost
- ImageProxyType
- ImageProxyURL
- ImageProxyOptions
- UseExperimentalGossip
- EnableTeamCreation 
- RestrictTeamInvite
- RestrictPublicChannelManagement
- RestrictPrivateChannelManagement
- RestrictPublicChannelCreation
- RestrictPrivateChannelCreation
- RestrictPublicChannelDeletion
- RestrictPrivateChannelDeletion
- RestrictPrivateChannelManageMembers
- DisableLegacyMFA
- ExperimentalTownSquareIsReadOnly
- ExperimentalHideTownSquareinLHS
- EnableXToLeaveChannelsFromLHS
- CloseUnusedDirectMessages
- ExperimentalChannelOrganization
- ExperimentalChannelSidebarOrganization
- EnableLegacySidebar
```

Removed the legacy MFA endpoint.
Removed `utils/authorization.go` and moved any permissions to the `MakeDefaultRoles()` function

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-36444

#### Release Note

```release-note
Removed all of the following configs and cleaned up any code around them:
- EnableOnlyAdminIntegrations
- RestrictCustomEmojiCreation
- RestrictPostDelete
- AllowEditPost
- ImageProxyType
- ImageProxyURL
- ImageProxyOptions
- UseExperimentalGossip
- EnableTeamCreation 
- RestrictTeamInvite
- RestrictPublicChannelManagement
- RestrictPrivateChannelManagement
- RestrictPublicChannelCreation
- RestrictPrivateChannelCreation
- RestrictPublicChannelDeletion
- RestrictPrivateChannelDeletion
- RestrictPrivateChannelManageMembers
- DisableLegacyMFAEndpoint
- ExperimentalTownSquareIsReadOnly
- ExperimentalHideTownSquareinLHS
- EnableXToLeaveChannelsFromLHS
- CloseUnusedDirectMessages
- ExperimentalChannelOrganization
- ExperimentalChannelSidebarOrganization
- EnableLegacySidebar

Removed the legacy MFA endpoint.
Removed `utils/authorization.go` and moved any permissions to the MakeDefaultRoles() function
```


app/app_test.go
```diff
@@ -470,7 +229,7 @@ func TestDoEmojisPermissionsMigration(t *testing.T) {
 	systemAdmin2, systemAdminErr2 := th.App.GetRoleByName(context.Background(), model.SystemAdminRoleId)
 	assert.Nil(t, systemAdminErr2)
 	sort.Strings(systemAdmin2.Permissions)
-	assert.Equal(t, expectedSystemAdmin, systemAdmin2.Permissions, fmt.Sprintf("'%v' did not have expected permissions", model.SystemAdminRoleId))
+	// assert.Equal(t, expectedSystemAdmin, systemAdmin2.Permissions, fmt.Sprintf("'%v' did not have expected permissions", model.SYSTEM_ADMIN_ROLE_ID))
```
BenCookie95, line 232: Need to check back why I commented this out

model/role.go
```diff
@@ -851,6 +864,7 @@ func MakeDefaultRoles() map[string]*Role {
 			PermissionCreateDirectChannel.Id,
 			PermissionCreateGroupChannel.Id,
 			PermissionViewMembers.Id,
+			PermissionCreateTeam.Id,
 		},
 		SchemeManaged: true,
 		BuiltIn:       true,
```
mkraft, line 870: What was the reason for the default roles getting added permissions?

BenCookie95: The old configs were used to add more permissions to each role on app setup (utils/authorization.go). Those configs are now gone so we don't have any way to add those permissions on startup. I added the permissions to the appropriate roles based on their old config default value.

mkraft: So if I'm understanding correctly the permissions that are added here to the defaults are from the `else` cases in https://github.com/mattermost/mattermost-server/pull/18035/files#diff-f8055deb0d618974840639d595f837567ad315388c8cbdefe56b19c00b6ec22f if no less restricting config setting was found?

BenCookie95: Yes they're from the `else` cases

mkraft (approved):

streamer45:
@hanzei I believe this is what you were referring to.

app/migrations.go
```diff
@@ -31,7 +30,6 @@ func (s *Server) doAdvancedPermissionsMigration() {
 
 	mlog.Info("Migrating roles to database.")
 	roles := model.MakeDefaultRoles()
-	roles = utils.SetRolePermissionsFromConfig(roles, s.Config(), s.License() != nil)
```
streamer45, line 34: Slightly unrelated but wondering if we should be keeping 3+ years old migration code around. I am thinking this could be an opportunity to just drop it (in a separate PR of course). 

@mkraft Thoughts?

streamer45 (approved):
Nicely done :+1:

app/migrations.go
mkraft: > Slightly unrelated but wondering if we should be keeping 3+ years old migration code around. I am thinking this could be an opportunity to just drop it (in a separate PR of course).
> 
> @mkraft Thoughts?

Generally sounds like a good idea. Do you want to create a ticket @streamer45 and part of the ticket can be to investigate the details of whether it's feasible?

streamer45: Sure :+1: [(MM-37842)](https://mattermost.atlassian.net/browse/MM-37842)

esethna:
@BenCookie95 if this needs QA can you help assign the appropriate person and add the label when ready?

BenCookie95:
> @BenCookie95 if this needs QA can you help assign the appropriate person and add the label when ready?

@esethna It will need QA but I'm preparing the frontend PR right now, it's hard to test the server in isolation. Will update it when webapp gets approved

cpanato:
@BenCookie95 can you fix your git config? looks like you have mixed the commit with two configs/users setup, also rebase and squash some duplicate commits

hanzei:
@BenCookie95 Is there a corresponding webapp PR? Does this PR impact the mobile app?

hanzei:
@BenCookie95 Does this PR impact the mobile app?

mkraft:
> @BenCookie95 Does this PR impact the mobile app?

Yes, here's the related mobile PR: https://github.com/mattermost/mattermost-mobile/pull/5627

streamer45:
Build is green :tada:

furqanmlk (approved):
Will run full E2E on WebApp PR after it gets merged

BenCookie95:
> Will run full E2E on WebApp PR after it gets merged

@furqanmlk I don't think we can merge this before webapp/mobile. They would need to merge at the same time because webapp/mobile will break

hanzei:
What the submodule `gosaml2` added intentionally?

BenCookie95:
> What the submodule `gosaml2` added intentionally?

Good catch, not sure how that got in there tbh

furqanmlk (changes requested):
@BenCookie95 
All settings are removed except **ImageProxyType**

furqanmlk (approved):
All settings ware tested

lieut-data:
@BenCookie95, it looks like we need to rework this PR in the same spirit as https://github.com/mattermost/mattermost-webapp/pull/8886. For example, in the very first diff for `api4/post.go` here, removing `ExperimentalTownSquareIsReadOnly` from the conditions just makes Town Square read-only for all but system admins.

I've been fixing this from the perspective of our e2e tests in playbooks, but wondering if you can own reverting or patching the changes instead?

mkraft:
> @BenCookie95, it looks like we need to rework this PR in the same spirit as [mattermost/mattermost-webapp#8886](https://github.com/mattermost/mattermost-webapp/pull/8886). For example, in the very first diff for `api4/post.go` here, removing `ExperimentalTownSquareIsReadOnly` from the conditions just makes Town Square read-only for all but system admins.
> 
> I've been fixing this from the perspective of our e2e tests in playbooks, but wondering if you can own reverting or patching the changes instead?

Did you try using the channel moderation permissions to give users and/or guests the permissions you're looking for?

Screen Shot 2021-09-23 at 12 30 38 PM

lieut-data:
@mkraft, we're testing against an enterprise edition, with no special permissions. The problem (similar to the webapp PR linked) is that Town Square is simply permanently read-only for non-admins. Simply removing the `ExperimentalTownSquare` check left the surrounding code intact, e.g.:

https://github.com/mattermost/mattermost-server/blob/9858ba9320f9a825eb99997739ee9b544424e776/api4/post.go#L727-L732

I didn't realize this locally for a while given that I often run as an admin there.

mkraft:
Ah, I see. The UX looks right but the server API request returns unauthorized. We probably missed that in our dev and testing. That block of code can be removed. It's not needed.

lieut-data:
Thanks, @mkraft & @BenCookie95! Just to be precise, there are lots of other blocks like that that should be reviewed, e.g. reactions, webhooks, etc.

mkraft:
> Thanks, @mkraft & @BenCookie95! Just to be precise, there are lots of other blocks like that that should be reviewed, e.g. reactions, webhooks, etc.

@BenCookie95 Would you be able to open a new PR with the required fixes?

BenCookie95:
> > Thanks, @mkraft & @BenCookie95! Just to be precise, there are lots of other blocks like that that should be reviewed, e.g. reactions, webhooks, etc.
> 
> @BenCookie95 Would you be able to open a new PR with the required fixes?

Yeah I'll take a look through those today!

PR #18041, opened 2021-08-02
Print a helpful message for JSON column update failures.

agnivade:
Refer: https://github.com/mattermost/mattermost-server/issues/18038

```release-note
NONE
```

### DIFF
--- a/store/sqlstore/store.go
+++ b/store/sqlstore/store.go
@@ -920,7 +920,11 @@ func (ss *SqlStore) AlterColumnTypeIfExists(tableName string, columnName string,
 	}
 
 	if err != nil {
-		mlog.Critical("Failed to alter column type", mlog.Err(err))
+		msg := "Failed to alter column type."
+		if mySqlColType == "JSON" && postgresColType == "jsonb" {
+			msg += " It is likely you might have invalid JSON values in the column. Please fix the values manually and run the migration again."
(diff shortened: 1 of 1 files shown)


store/sqlstore/store.go
```diff
@@ -920,7 +920,11 @@ func (ss *SqlStore) AlterColumnTypeIfExists(tableName string, columnName string,
 	}
 
 	if err != nil {
-		mlog.Critical("Failed to alter column type", mlog.Err(err))
+		msg := "Failed to alter column type."
+		if mySqlColType == "JSON" && postgresColType == "jsonb" {
+			msg += " It is likely you might have invalid JSON values in the column. Please fix the values manually and run the migration again."
```
agnivade, line 925: 0/5 on a better error message here. Let me know if you want something changed.

streamer45: I'd remove the "might" since it's already a conditional but other than that sounds good to me.

streamer45 (approved):
Thanks, this will probably avoid a few community issues down the line.

store/sqlstore/store.go
agnivade:
```suggestion
			msg += " It is likely you have invalid JSON values in the column. Please fix the values manually and run the migration again."
```

PR #18043, opened 2021-08-03
[MM-24714]  Render markdown in email notifications

Zeezee1210:
#### Summary
This PR adds a function in the utils/markdown.go to take a string containing markdown and then return the string converted to HTML tags. It's then called when creating the body of notification emails to convert the post message markdown to HTML.

#### Ticket Link
https://github.com/mattermost/mattermost-server/issues/17613

#### Release Note

```release-note
Added rendering for posts containing markdown in email notifications
```


hmhealey (commented):
Thanks for the PR, @Zeezee1210! This is something we've wanted for a long time, but I honestly didn't realize the had a proper Markdown library for use on the server now.

This looks good to me, but I'm going to defer my review to @agnivade since he's worked with the server's Markdown code much more recently than I have.

agnivade (commented):
Thank you @Zeezee1210 for the PR!

app/notification_email_test.go
```diff
@@ -825,3 +825,96 @@ func TestLandingLinkPermalink(t *testing.T) {
 	require.NoError(t, err)
 	require.Contains(t, body, teamURL+"/pl/"+post.Id, fmt.Sprintf("Expected email text '%s'. Got %s", teamURL, body))
 }
+
+func TestMarkdownLinksRender(t *testing.T) {
```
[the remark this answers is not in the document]
Zeezee1210: Hi @agnivade,
Thanks a lot for the great review and appreciate your feedback.
I've switched this to table-driven test (which I've hopefully done correctly)...when it comes to adding more tests I popped in a few based on looking at the tests used in the markdown_test.go but if there's any more specific ones you'd want adding just say.

app/notification_email_test.go
```diff
@@ -825,3 +825,96 @@ func TestLandingLinkPermalink(t *testing.T) {
...
+
+	storeMock := th.App.Srv().Store.(*mocks.Store)
+	teamStoreMock := mocks.TeamStore{}
+	teamStoreMock.On("GetByName", "testteam").Return(&model.Team{Name: "testteam"}, nil)
+	storeMock.On("Team").Return(&teamStoreMock)
+
+	body, err := th.App.getNotificationEmailBody(recipient, post, channel, channelName, senderName, teamName, teamURL, emailNotificationContentsType, true, translateFunc, "user-avatar.png")
+	expect := "This is <a href=\"https://mattermost.com\">Mattermost</a>"
+	require.NoError(t, err)
+	require.Contains(t, body, expect, fmt.Sprintf("Expected email text '%s'. Got %s", expect, body))
```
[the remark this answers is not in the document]
Zeezee1210: Yes, I based the tests on the previous ones. I've removed the string as I don't see a need when testify is good enough.

app/notification_email_test.go
```diff
@@ -825,3 +825,96 @@ func TestLandingLinkPermalink(t *testing.T) {
...
+	channel := &model.Channel{
+		DisplayName: "ChannelName",
+		Type:        model.ChannelTypeOpen,
+	}
+	channelName := "ChannelName"
+	senderName := "sender"
+	teamName := "testteam"
+	teamURL := "http://localhost:8065/landing#/testteam"
+	emailNotificationContentsType := model.EmailNotificationContentsFull
+	translateFunc := i18n.GetUserTranslations("en")
```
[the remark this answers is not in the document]
Zeezee1210: Agreed, cleaner just passing directly for a test so I've changed that.

mkraft (approved):

mkraft:
@Zeezee1210 This is going to be really awesome!

app/notification_email_test.go
```diff
@@ -825,3 +825,99 @@ func TestLandingLinkPermalink(t *testing.T) {
...
+	for _, tt := range tests {
+		t.Run(tt.name, func(t *testing.T) {
+			post := &model.Post{
+				Id:      "Test_id",
+				Message: tt.args,
+			}
+			got, err := th.App.getNotificationEmailBody(recipient, post, channel, "ChannelName", "sender", "testteam", "http://localhost:8065/landing#/testteam", model.EmailNotificationContentsFull, true, i18n.GetUserTranslations("en"), "user-avatar.png")
+			if err != nil {
+				t.Fatalf("error: %v", err)
+			}
```
agnivade, line 919: Change this to `require.NoError(t, err)`

app/notification_email_test.go
```diff
@@ -825,3 +825,99 @@ func TestLandingLinkPermalink(t *testing.T) {
...
+			require.Contains(t, got, tt.want)
```
agnivade, line 920: Should it be `Contains` or `Equal`? I'd assume it should be equal.

agnivade (commented):
Nice work! Just a couple of finishing touches needed.

app/notification_email_test.go
Zeezee1210: haha damn! I did have this but for some reason went with the t.Fatalf.

app/notification_email_test.go
Zeezee1210: I believe this should be `Contains` as it is testing against a render of the whole email HTML and we're only looking for the message portion.

agnivade: Ah yes, I missed that the entire email contains lots of other stuff.

agnivade (approved):
Thank you @Zeezee1210 ! 

I think I somehow missed that we already have a custom markdown renderer in our codebase. I am wondering how much of a parity we have with that. Would it be possible for you to try the `shared/markdown.RenderHTML` function? 

I'd also want to move these two functions from inside utils to the `shared/markdown` package.

Zeezee1210:
@agnivade, I did spot that when I first started working on it and gave it a go. Unfortunately it only worked for things like links, I initially switched to the current package to support emphasis. I suspect it won't have supported strikethrough/tables either.

agnivade:
Sounds good, thanks for the thorough investigation! @hmhealey - I think we should unify the two modules at some point. If the goldmark parser is a superset, we should be able to replace some of the home-grown code.

hmhealey:
Yeah, I agree that it would make sense to combine them. Apart from additions like at-mentions, we do have a few places where we deviate from proper Markdown that we'd probably like to address, but we aren't 100% consistent even between mobile and web yet, so those would probably be acceptable for the time being.

For context around the `shared/markdown` package, that was added fairly early on to do some URL rewriting back when the image proxy was first added, and I think it's also used more recently for mention parsing and pulling out URLs for file previews. It only has the ability to parse things that will affect those (essentially just links and code blocks).

agnivade:
The goldmark parser is fully extendable and allows for callbacks when walking the entire markdown tree. So extracting URLs, mention parsing should all be doable.

furqanmlk (commented):
Thanks @Zeezee1210 

I have verified following formate, all looks good except "**Block Quotes**"
**Text Style** ✅ 
**Links** ✅ 
**Headings** ✅ 
**Lists** ✅ 
**Code Block** ✅ 
**Block quotes** 🔴 
**Tables** ✅ 


image

Zeezee1210:
Thanks for the testing @furqanmlk, I took a look at the blockquote issue you spotted. It appears there was a call to html.EscapeString just prior to the markdown conversion and therefore the > was escaped and didn't get recognised as start of blockquote markdown. 

I've removed that call as I don't believe it was needed, and also added a test for blockquotes in the table tests. Looks to have sorted it on my local dev build.

hmhealey (changes requested):
That sounds like it should be fine to remove the `html.EscapeString` call since I assume the Markdown library does its own escaping, but could you add a test to confirm that? Something as simple as the input being `<b>not bold</b>` would be fine

Zeezee1210:
@hmhealey good point. I might have to re-think it. 
Adding that test to current code shows it converts `<b>not bold</bold>` to `<!-- raw HTML omitted -->not bold<!-- raw HTML omitted -->`

furqanmlk:
> Thanks for the testing @furqanmlk, I took a look at the blockquote issue you spotted. It appears there was a call to html.EscapeString just prior to the markdown conversion and therefore the > was escaped and didn't get recognised as start of blockquote markdown.
> 
> I've removed that call as I don't believe it was needed, and also added a test for blockquotes in the table tests. Looks to have sorted it on my local dev build.

Thanks @Zeezee1210 
I re-tested, I don't see quote in correct quote format, is that expected. ?

image

Zeezee1210:
> Thanks @Zeezee1210 
> 
> I re-tested, I don't see quote in correct quote format, is that expected. ?
> 
> 
> 
> image
> 
> 

Hi @furqanmlk,
I would expect a markdown blockquote to render like this:
image

With a bar on the left and the text indented. Further reference here:
https://www.w3schools.io/file/markdown-blockquotes/

furqanmlk:
> > Thanks @Zeezee1210
> > I re-tested, I don't see quote in correct quote format, is that expected. ?
> > image
> 
> Hi @furqanmlk,
> I would expect a markdown blockquote to render like this:
> image
> 
> With a bar on the left and the text indented. Further reference here:
> https://www.w3schools.io/file/markdown-blockquotes/

Yeah, I don't see that vertical  bar in my email

image

Zeezee1210:
> > > Thanks @Zeezee1210
> 
> > > I re-tested, I don't see quote in correct quote format, is that expected. ?
> 
> > > image
> 
> > 
> 
> > Hi @furqanmlk,
> 
> > I would expect a markdown blockquote to render like this:
> 
> > image
> 
> > 
> 
> > With a bar on the left and the text indented. Further reference here:
> 
> > https://www.w3schools.io/file/markdown-blockquotes/
> 
> 
> 
> Yeah, I don't see that vertical  bar in my email
> 
> 
> 
> image
> 
> 

Hmm odd. Does the raw html of the email still have the correct html tags? Something probably like this:
```
<blockquote>
<p>block quotes</p>
</blockquote>
```

furqanmlk:
> > > > Thanks @Zeezee1210
> > 
> > 
> > > > I re-tested, I don't see quote in correct quote format, is that expected. ?
> > 
> > 
> > > > image
> > 
> > 
> > > 
> > 
> > 
> > > Hi @furqanmlk,
> > 
> > 
> > > I would expect a markdown blockquote to render like this:
> > 
> > 
> > > image
> > 
> > 
> > > 
> > 
> > 
> > > With a bar on the left and the text indented. Further reference here:
> > 
> > 
> > > https://www.w3schools.io/file/markdown-blockquotes/
> > 
> > 
> > Yeah, I don't see that vertical  bar in my email
> > image
> 
> Hmm odd. Does the raw html of the email still have the correct html tags? Something probably like this:
> 
> ```
> <blockquote>
> <p>block quotes</p>
> </blockquote>
> ```

Looks like, it's there

image

Zeezee1210:
> > > > > Thanks @Zeezee1210
> 
> > > 
> 
> > > 
> 
> > > > > I re-tested, I don't see quote in correct quote format, is that expected. ?
> 
> > > 
> 
> > > 
> 
> > > > > image
> 
> > > 
> 
> > > 
> 
> > > > 
> 
> > > 
> 
> > > 
> 
> > > > Hi @furqanmlk,
> 
> > > 
> 
> > > 
> 
> > > > I would expect a markdown blockquote to render like this:
> 
> > > 
> 
> > > 
> 
> > > > image
> 
> > > 
> 
> > > 
> 
> > > > 
> 
> > > 
> 
> > > 
> 
> > > > With a bar on the left and the text indented. Further reference here:
> 
> > > 
> 
> > > 
> 
> > > > https://www.w3schools.io/file/markdown-blockquotes/
> 
> > > 
> 
> > > 
> 
> > > Yeah, I don't see that vertical  bar in my email
> 
> > > image
> 
> > 
> 
> > Hmm odd. Does the raw html of the email still have the correct html tags? Something probably like this:
> 
> > 
> 
> > ```
> 
> > <blockquote>
> 
> > <p>block quotes</p>
> 
> > </blockquote>
> 
> > ```
> 
> 
> 
> Looks like, it's there
> 
> 
> 
> image
> 
> 

That looks like a difference in mail client html rendering then. I have a test email in ProtonMail client which shows the bar, but on Apple iOS mail app it chooses not to show the bar. Appears to be the correct html tagging though.

Zeezee1210:
> That sounds like it should be fine to remove the `html.EscapeString` call since I assume the Markdown library does its own escaping, but could you add a test to confirm that? Something as simple as the input being `<b>not bold</b>` would be fine

Hi @hmhealey,
It would be great if you could take a look at the latest commit to see what you think.  The issue I found was using html.EscapeString on the entire postmessage first meant that the markdown parser missed the blockquote angle bracket. Then when I removed that step, you rightly showed that it would then affect html tags in the post message. 
So I added a cleaning step in the markdown function to escape strings in matching pairs of html tags and leave the others. Anything not caught should be protected by the markdown package that doesn't allow raw html by replacing it with ` <!-- raw HTML omitted -->`

utils/markdown.go
```diff
@@ -35,12 +37,19 @@ func StripMarkdown(markdown string) (string, error) {
 
 // MarkdownToHTML takes a string containing Markdown and returns a string with HTML tagged version
 func MarkdownToHTML(markdown string) (string, error) {
+	re := regexp.MustCompile(`<(.*)>.*?|<(.*) />`)
```
hmhealey, line 40: This regex isn't going to work unfortunately. Obviously we can't use `html.EscapeString` on the whole thing because of the issue with block quotes, but I'd be worried we'd miss something with anything like this that's much more complicated than just replacing the unsafe characters with HTML entities. Generally, trying to use regexes to do much with HTML is a bad idea.

hmhealey (changes requested):
That's unfortunate that there's no built-in option in Goldmark to just escape raw HTML and render it as plain text. Is that maybe something we could override by messing with the renderer?

Zeezee1210:
> That's unfortunate that there's no built-in option in Goldmark to just escape raw HTML and render it as plain text. Is that maybe something we could override by messing with the renderer?

From what I have read and understand there is a renderer option to allow raw html (called unsafe) which would then pass back a message post with `<b>bold</b>` but then obviously as that gets put in an email it's rendered as html because it hasn't encoded it.

The reason I put the regexp in was it needs to be just good enough to not encode the blockquote angle bracket but capture most html tagging and encode it. If it doesn't catch it, then the default behaviour of goldmark is to not allow unsafe html and strips it entirely. Which makes the regexp safer because there's the second sweep looking for html and removes it.

If you think that's still not a good idea, I'll keep having a think about it...would also appreciate any suggestions ;)

Zeezee1210:
> > That's unfortunate that there's no built-in option in Goldmark to just escape raw HTML and render it as plain text. Is that maybe something we could override by messing with the renderer?
> 
> From what I have read and understand there is a renderer option to allow raw html (called unsafe) which would then pass back a message post with `<b>bold</b>` but then obviously as that gets put in an email it's rendered as html because it hasn't encoded it.
> 
> The reason I put the regexp in was it needs to be just good enough to not encode the blockquote angle bracket but capture most html tagging and encode it. If it doesn't catch it, then the default behaviour of goldmark is to not allow unsafe html and strips it entirely. Which makes the regexp safer because there's the second sweep looking for html and removes it.
> 
> If you think that's still not a good idea, I'll keep having a think about it...would also appreciate any suggestions ;)

@hmhealey, having re-read this and just in case I didn't explain it clearly. The regexp is acting as an "allowlist" and anything it captures will then be encoded and show safely in a post message...everything else will be removed by the Goldmark renderer.

utils/markdown.go
Zeezee1210: Hi @hmhealey, how about doing the opposite? 
I've re-added the html.EscapeString step but within utils.MarkdownToHTML I now search for > at the beginning of a line to then unescape just that for blockquote parsing?

hmhealey:
Apologies for the delay. I ended up being out all of last week, and I forgot to check in on this PR.

I think I understand what you were trying to do before, but I was worried there might have been ways to either get around the regex or cause it to have performance problems (which have come up a lot in the web app's Markdown parser in the past).

That said, I actually really like the new way of doing this where we start with everything escaped and then unescape what we need for block quotes. Starting with it escaped makes me comfortable that we shouldn't let anything through by mistake, and the regex is simple enough that it can't cause any performance problems there. I wasn't sure how we'd be able to fix that without doing something more complicated like forking the Markdown library. Nice job!

hmhealey (approved):

esethna (approved):

furqanmlk:
@Zeezee1210 Are we ok without `Quote` format ?
I still don't vertical bar with quote in the email.

### Mattermost
image

### Email
image

Zeezee1210:
> @Zeezee1210 Are we ok without `Quote` format ?
> 
> I still don't vertical bar with quote in the email.
> 
> 
> 
> ### Mattermost
> 
> image
> 
> 
> 
> ### Email
> 
> image
> 
> 

Hi @furqanmlk,
I'm ok with that as the HTML is correct and the rendering is client specific/dependent so can't control it.

furqanmlk (approved):

PR #18052, opened 2021-08-04
MM-37573: Ignore nil fields in attachments

agnivade:
We extend the same logic done in https://github.com/mattermost/mattermost-server/pull/16556
to Field.

Also handled the case elsewhere in the code with a light grep
for extra safety.

https://mattermost.atlassian.net/browse/MM-37573

```release-note
NONE
```


isacikgoz (approved):
Of course the field could be `nil` as well 🤦 Nice catch again!

agnivade:
Thank Sentry, not me :P

PR #18053, opened 2021-08-04
MM-35754: Fail fast if templates directory is not found

agnivade:
We used to log an error and move forward, which led to
panics later during server execution, because the templates
don't exist.

While technically speaking the server will run fine
because it's just one HTTP handler goroutine which panics,
but properly handling the error requires much more adjustments
in the code to check for nil templates.

It is better to fail earlier and force the application
to start in a clean slate.

https://mattermost.atlassian.net/browse/MM-37574

```release-note
NONE
```

### DIFF
--- a/app/server.go
+++ b/app/server.go
@@ -389,19 +389,18 @@ func NewServer(options ...Option) (*Server, error) {
 
 	templatesDir, ok := templates.GetTemplateDirectory()
 	if !ok {
-		mlog.Error("Failed find server templates", mlog.String("directory", "templates"))
-	} else {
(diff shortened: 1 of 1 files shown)


isacikgoz (approved):
Nice catch, this must escaped from me during the log refactor.

jespino (approved):
Good point, looks good to me.

jespino:
@agnivade I don't know if this needs QA

agnivade:
@jespino - We are back to usual "server platform" rules. So race tests fixes/flaky test fixes/sentry crashes don't need QA. :)

PR #18054, opened 2021-08-04
[MM-37496] Temporarily disable admin advisor notifications

streamer45:
#### Summary

PR temporarily disables (again) the job triggering admin advisor notifications. I've opted to return an error early (essentially making it a no-op) instead of removing the call to facilitate unit testing and make sure this doesn't get erroneously re-enabled.

#### Ticket

https://mattermost.atlassian.net/browse/MM-37496

#### Release Note

```release-note
Admin advisor notifications were temporarily disabled.
```


harshilsharma63 (approved):

agnivade:
Hey all, I ran into this by accident. I would recommend to remove the code entirely rather than adding `//nolint` comments, or returning early. Keeping unused code lying around is not a good practice.

Since it's already part of the git commit history, all we need is a git revert to re-enable this feature.

streamer45:
> Hey all, I ran into this by accident. I would recommend to remove the code entirely rather than adding `//nolint` comments, or returning early. Keeping unused code lying around is not a good practice.
> 
> Since it's already part of the git commit history, all we need is a git revert to re-enable this feature.

No objections from my side but that inevitably means removing the unit test for it.

cc @mkraft

mkraft:
> > Hey all, I ran into this by accident. I would recommend to remove the code entirely rather than adding `//nolint` comments, or returning early. Keeping unused code lying around is not a good practice.
> > Since it's already part of the git commit history, all we need is a git revert to re-enable this feature.
> 
> No objections from my side but that inevitably means removing the unit test for it.
> 
> cc @mkraft

Sounds good to me.

furqanmlk (approved):

PR #18072, opened 2021-08-05
[MM-37306] Replace db count query in user signup process

streamer45:
#### Summary

Performance insights on recent load-tests highlighted a count query on the `Users` table as one of the most expensive being made. This was used to check whether the user was the first in the system in which case it would get promoted to admin.
To improve performance we are replacing this with a simple check on whether the table is empty which runs in constant time.
A quick load-test showed than on average the `createUser` API call time improved by nearly 40%,

#### Ticket

https://mattermost.atlassian.net/browse/MM-37306

#### Release Note

```release-note
NONE
```

### DIFF
--- a/app/import_functions.go
+++ b/app/import_functions.go
@@ -493,8 +493,8 @@ func (a *App) importUser(data *UserImportData, dryRun bool) *model.AppError {
 				return appErr
 			case errors.Is(err, users.AcceptedDomainError):
 				return model.NewAppError("importUser", "api.user.create_user.accepted_domain.app_error", nil, "", http.StatusBadRequest)
-			case errors.Is(err, users.UserCountError):
-				return model.NewAppError("importUser", "app.user.get_total_users_count.app_error", nil, nErr.Error(), http.StatusInternalServerError)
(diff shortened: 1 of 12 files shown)


agnivade (approved):
Awesome. There are a couple of other places like this too. But I haven't touched them because I wasn't sure if they were costly or not. Maybe we can take another look sometime.

streamer45:
> Awesome. There are a couple of other places like this too. But I haven't touched them because I wasn't sure if they were costly or not. Maybe we can take another look sometime.

Absolutely. This mainly affected load-tests given we signup lots of users in a small amount of time. Still, no need to do a full count.

isacikgoz (approved):
Awesome 🚀

streamer45:
@srkgupta 

This should be safe but since it involves making a user sysadmin I opted for a second look from your side of things.

srkgupta (approved):
Thanks @streamer45. The changes LGTM from security perspective.

furqanmlk (approved):
**No Users**
image

**ADMIN**
image
**MEMBER**
image

PR #18077, opened 2021-08-06
MM-37491: Fix flaky test on file content extraction

jespino:
#### Summary
Fix a flaky test of trying to read the extracted content from a recently
uploaded file (extracted in a different goroutine).

Now waits for the content extraction and also clears the cache properly.

#### Ticket Link
[MM-37491](https://mattermost.atlassian.net/browse/MM-37491)

#### Release Note
```release-note
NONE
```


app/file.go
```diff
@@ -1356,6 +1356,12 @@ func (a *App) ExtractContentFromFileInfo(fileInfo *model.FileInfo) error {
 		if storeErr := a.Srv().Store.FileInfo().SetContent(fileInfo.Id, text); storeErr != nil {
 			return errors.Wrap(storeErr, "failed to save the extracted file content")
 		}
+		reloadFileInfo, storeErr := a.Srv().Store.FileInfo().Get(fileInfo.Id)
+		if storeErr != nil {
+			mlog.Warn("Failed to invalidate the fileInfo cache.", mlog.Err(storeErr), mlog.String("file_info_id", fileInfo.Id))
```
agnivade, line 1361: I would suggest to return the error rather than just log as a warning. Cache invalidation can lead to subtle errors. It might be good to make this more visible.

app/post_metadata_test.go
```diff
@@ -189,7 +190,13 @@ func TestPreparePostForClient(t *testing.T) {
 
 		fileInfo.PostId = post.Id
 
-		clientPost := th.App.PreparePostForClient(post, false, false)
+		var clientPost *model.Post
+		assert.Eventually(t, func() bool {
+			p, nErr := th.App.Srv().Store.Post().Get(context.Background(), post.Id, false, false, false, "")
+			require.NoError(t, nErr)
```
agnivade, line 196: Wondering why do we need to do this. Isn't calling `PreparePostForClient` repeatedly sufficient?

app/post_metadata_test.go
```diff
@@ -189,7 +190,13 @@ func TestPreparePostForClient(t *testing.T) {
...
+			clientPost = th.App.PreparePostForClient(p.Posts[post.Id], false, false)
+			return assert.ObjectsAreEqual([]*model.FileInfo{fileInfo}, clientPost.Metadata.Files)
+		}, time.Second, 100*time.Millisecond)
 
 		assert.Equal(t, []*model.FileInfo{fileInfo}, clientPost.Metadata.Files, "should've populated Files")
```
agnivade, line 201: Do we need to keep this? I thought the above assertion was sufficient.

app/file.go
jespino: The point is, returning an error there give you the sensation that the error has happened, but that's not totally true, because the `SetContent` what is the main part, is already executed. This shouldn't fail ever, and in any case, is an improvement over the before lack of cache invalidation before.

app/post_metadata_test.go
jespino: Yes, probably we don't need it, good catch.

app/post_metadata_test.go
jespino: Yes, the idea is to provide a better output in that case, the previous error wasn't providing a nice output.

app/post_metadata_test.go
jespino: Done

agnivade (approved):
Good stuff!

isacikgoz (approved):
LGTM

PR #18079, opened 2021-08-06
Fix possible endless loop

streamer45:
#### Summary

Fixing a possible endless loop in case more of 25 admins are present in the system when `notifyAdminsOfWarnMetricStatus` runs.

#### Release Note

```release-note
NONE
```


agnivade (approved):
Would it be possible to write a unit test for this?

streamer45:
> Would it be possible to write a unit test for this?

I tried my best.

isacikgoz (approved):
LGTM 👍

PR #18081, opened 2021-08-06
mmctl plugin add: introduce default-false 'replace' flag

### LINKED ISSUE #18080: mmctl: provide an option to allow existing plugins to be replaced during 'mmctl plugin add'
#### Summary
The `mmctl plugin add` command does not currently support replacement of existing plugins, something that is possible when uploading plugins using the Mattermost System Console.

As a workaround, it's possible to use `mmctl plugin delete` to remove an existing plugin prior to the `add` operation.

It could be convenient to have an (opt-in) ability to replace existing plugins.

#### Steps to reproduce
1. Upload a plugin (either using the System Console, or using `mmctl`)
2. Attempt to upload the same plugin again using `mmctl plugin add`

The following error appears:

```
Unable to add plugin: dist/example-plugin-1.0.0.tgz. Error: : Unable to install plugin. A plugin with the same ID is already installed.
```

#### Possible fixes
The `replace` parameter to `InstallPlugin` is currently [fixed to `false`](https://github.com/mattermost/mattermost-server/blob/bf1f60d309998429b4cd17ea329215b0a0080c40/cmd/mattermost/commands/plugin.go#L123); that could be adjusted to accept a flag from the command-line.

jayaddison-collabora:
…allows existing plugins to be replaced
#### Summary
Adds a `--replace` flag to the `mmctl plugin add` command, allowing replacement of an existing plugin with the same ID.

#### Ticket Link
Resolves #18080.

#### Release Note
```release-note
Introduce support for replacement of existing plugins to the 'mmctl plugin add' command
```

### DIFF
--- a/cmd/mattermost/commands/plugin.go
+++ b/cmd/mattermost/commands/plugin.go
@@ -91,6 +91,7 @@ func init() {
 		PluginAddPublicKeyCmd,
 		PluginDeletePublicKeyCmd,
 	)
+	PluginAddCmd.Flags().Bool("replace", false, "Allow replacement of existing plugin with the same ID.")
 	PluginCmd.AddCommand(
 		PluginAddCmd,
 		PluginDeleteCmd,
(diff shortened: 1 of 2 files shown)


jayaddison-collabora:
Ah, I see from #17995 that the `plugin` command is legacy and has been removed from the master branch.  Closing this, and will investigate alternatives.

PR #18082, opened 2021-08-06
Pre-package Incident Collaboration v1.16.0

cpoile:
#### Summary
Pre-package Incident Collaboration v1.16.0

#### Ticket Link
None.

#### Release Note
```release-note
Pre-package Incident Collaboration v1.16.0
```


calebroseland (approved):

model/feature_flags.go
```diff
@@ -50,7 +50,7 @@ func (f *FeatureFlags) SetDefaults() {
 	f.CollapsedThreads = true
 	f.EnableRemoteClusterService = false
 	f.AppsEnabled = false
-	f.PluginIncidentManagement = "1.14.3"
+	f.PluginIncidentManagement = "1.16.0"
```
lieut-data, line 53: Let's set this to "blank" so that split.io doesn't get involved at all:
```suggestion
	f.PluginIncidentManagement = ""
```

lieut-data (changes requested):

cpoile:
@lieut-data Sorry, I'm not sure what's happening with the prepackaged build step -- is there something we've missed?

lieut-data:
@cpoile, I think the error here is that the plugin signature doesn't match the plugin bundle. I wonder if this was a result of running concurrent `/mb cutplugin` builds? We may have to cut a brand new `v1.16.0` from the same commit to get the desired output.

cpoile:
@lieut-data Seems to be failing for the same reason again. It looks like it can't find the osx package? 
```
cp: cannot stat 'tmpprepackaged/mattermost-plugin-incident-collaboration-v1.16.0-osx-amd64.tar.gz': No such file or directory
cp: cannot stat 'tmpprepackaged/mattermost-plugin-incident-collaboration-v1.16.0-osx-amd64.tar.gz.sig': No such file or directory
tar: dist/osx/mattermost/prepackaged_plugins/mattermost-plugin-incident-collaboration-v1.16.0-osx-amd64.tar.gz: Cannot open: No such file or directory
tar: Error is not recoverable: exiting now
Contains  in mattermost-plugin-incident-collaboration-v1.16.0-osx-amd64.tar.gz but needs dist/plugin-darwin-amd64
```

Okay if I try making a 1.16.1 and see if that fixes it? Like you said, I might have messed it up by running two in quick succession.

lieut-data:
Hmm, weird. I'm confused as well! Does the v1.16.0 contain the expected `osx` build? Feel free to blow away the builds altogether and we can just cut a fresh one from scratch.

I also found this note in the feature flags docs which makes me re-think my approach for clearing out the value earlier:

> Yes. If you leave the default "” then nothing will happen for self-managed installations. You can ask the Cloud team to set split.io/environment to a specific version.

Your original changes may have been more correct than my suggestion.

Makefile
```diff
@@ -111,7 +111,7 @@ PLUGIN_PACKAGES += mattermost-plugin-channel-export-v0.2.2
 PLUGIN_PACKAGES += mattermost-plugin-custom-attributes-v1.3.0
 PLUGIN_PACKAGES += mattermost-plugin-github-v2.0.1
 PLUGIN_PACKAGES += mattermost-plugin-gitlab-v1.3.0
-PLUGIN_PACKAGES += mattermost-plugin-incident-collaboration-v1.16.2
```
lieut-data: Oh goodness, is that all it ended up being? :p

lieut-data (approved):

Makefile
cpoile: Yep. 🤕

amyblais:
@cpoile Will this also need a cherry-pick to ``cloud``?

PR #18087, opened 2021-08-09
[MM-37716] Drop support for LHS specific bot icons

hanzei:
#### Summary
Before the new sidebar was added, bots could set LHS specific icons to allow for easier identification. Unfortunately, this feature was never implemented on mobile. With the new sidebar user icons are shown anyway on the LHS rendering this feature unnecessary. Hence, this PR removes it.

Please note that existing bot icons are kept in the file store. Given the low number of files I didn't consider removing them important.

Please also note that the related plugin API methods also get removed as part of this PR. I didn't found a single plugin that makes use of them and removing them seemed more clear to me. For testing purposed I setup a plugin that uses it. When running it against this branch I could confirm that it can still be enabled. There is a warning about a non-existing RPC endpoint in the logs.

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-37716

#### Release Note
```release-note
Drop support for LHS specific bot icons
```


larkox (approved):

streamer45 (approved):
Looks good :+1:

DHaussermann (approved):
Tested and passed
- Deployed Server and Webapp PRs
- SVG use in sidebar is no longer functional
- Even if already set - Upgrading to this PR will use profile image now instead
- Tested legacy sidebar as well to ensure that LHS icon will revert to generic bot icon 
No issues found

LGTM!

PR #18088, opened 2021-08-09
CRT: desktop thread notifications

koox00:
#### Summary

Adds thread followers to mentions depending on CRT and desktop_threads notify prop setting.
This is done to be able to handle notifications for CRT users.

#### Ticket Link

https://mattermost.atlassian.net/browse/MM-36234

#### Release Note

```release-note
NONE
```


app/notification.go
```diff
@@ -286,6 +329,28 @@ func (a *App) SendNotifications(post *model.Post, team *model.Team, channel *mod
...
+
+			senderProfileImage, _, err := a.GetProfileImage(sender)
+			if err != nil {
+				a.Log().Warn("Unable to get the sender user profile image.", mlog.String("user_id", sender.Id), mlog.Err(err))
+			}
+			if err := a.sendNotificationEmail(notification, profileMap[id], team, senderProfileImage); err != nil {
+				mlog.Warn("Unable to send notification email.", mlog.Err(err))
+			}
+		}
+	}
```
jwilander, line 352: This whole block is very similar to the previous block. What do you think about combining them a bit?

The primary difference seems to be that for the followers you don't need to check if the user allows emails since it's already been done. I'm thinking something along the lines of creating a new array from `mentionedUsersList` and `notifyCRTEmail`, looping through them all and then just changing the check to be:

```go
if notifyCRT.Email.Contains(id) || a.userAllowsEmail(profileMap[id], channelMemberNotifyPropsMap[id], post) {
```

app/notification.go
```diff
@@ -255,6 +281,23 @@ func (a *App) SendNotifications(post *model.Post, team *model.Team, channel *mod
 		updateMentionChans = append(updateMentionChans, umc)
 	}
 
+	notifyCRT := &CRTNotifiers{}
```
jwilander, line 284: nit: `notifyCRT` seems a little ambiguous to me. What about something like `notificationsForCRT`?

jwilander (commented):
Good work! The notification code is a bit of a beast but you seemed to have navigated pretty successfully. Just have a couple suggestions.

hmhealey (approved):

app/notification.go
```diff
@@ -274,7 +320,7 @@ func (a *App) SendNotifications(post *model.Post, team *model.Team, channel *mod
 				continue
 			}
 
-			if a.userAllowsEmail(profileMap[id], channelMemberNotifyPropsMap[id], post) {
+			if notificationsForCRT.Email.Contains(id) || a.userAllowsEmail(profileMap[id], channelMemberNotifyPropsMap[id], post) {
```
ashishbhate, line 323: On line 309 we add `notificationsForCRT.Email` to `emailReceipients`, and here we're looping over `emailReceipients` so this `Contains` check isn't required.

koox00: It's because `channelMemberNotifyPropsMap` may not allow the email, but `email_threads` notify prop may allow it.
hm, I'll add the check in `a.userAllowsEmail` and remove the `Contains` from here.

app/notification.go
```diff
@@ -433,6 +479,28 @@ func (a *App) SendNotifications(post *model.Post, team *model.Team, channel *mod
...
+				status = &model.Status{UserId: id, Status: model.StatusOffline, Manual: false, LastActivityAt: 0, ActiveChannel: ""}
+			}
+
+			if DoesStatusAllowPushNotification(profileMap[id].NotifyProps, status, post.ChannelId) {
+				a.sendPushNotification(
+					notification,
+					profileMap[id],
+					false,
+					false,
+					model.CommentsNotifyAny,
```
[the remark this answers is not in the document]
koox00: oh, you're correct, I'll change that to "commented on a thread you are following"

app/notification.go
```diff
@@ -433,6 +479,28 @@ func (a *App) SendNotifications(post *model.Post, team *model.Team, channel *mod
...
+				continue
+			}
+
+			var status *model.Status
+			var err *model.AppError
+			if status, err = a.GetStatus(id); err != nil {
+				status = &model.Status{UserId: id, Status: model.StatusOffline, Manual: false, LastActivityAt: 0, ActiveChannel: ""}
+			}
+
+			if DoesStatusAllowPushNotification(profileMap[id].NotifyProps, status, post.ChannelId) {
```
ashishbhate, line 494: In other places we DEBUG log that a notification isn't sent, if a notification shouldn't be sent. We should do that here as well.

app/notification.go
```diff
@@ -465,8 +533,19 @@ func (a *App) SendNotifications(post *model.Post, team *model.Team, channel *mod
...
+		usersToReceiveNotification = append(usersToReceiveNotification, mentionedUsersList...)
+	}
+
+	if len(notificationsForCRT.Desktop) != 0 {
+		usersToReceiveNotification = append(usersToReceiveNotification, notificationsForCRT.Desktop...)
+	}
+
+	if len(usersToReceiveNotification) != 0 {
+		usersToReceiveNotification = model.RemoveDuplicateStrings(usersToReceiveNotification)
+		message.Add("mentions", model.ArrayToJson(usersToReceiveNotification))
```
ashishbhate, line 548: I'm not sure about this side-effects of this change. `usersToReceiveNotification` is mentionedUsers + followers. Adding followers to the `mentions` field of the websocket message doesn't seem correct.

koox00: This change was done so that we won't have to call the thread in the front-end to see if the user is following or not.
But we could just add another field called `followers` so that we don't mix.

app/notification.go
```diff
@@ -230,6 +252,10 @@ func (a *App) SendNotifications(post *model.Post, team *model.Team, channel *mod
 					mac <- model.NewAppError("SendNotifications", "app.channel.autofollow.app_error", nil, err.Error(), http.StatusInternalServerError)
 					return
 				}
+				followersMutex.Lock()
+				followers = append(followers, userID)
+				followersMutex.Unlock()
+
```
ashishbhate, line 258: With this addition a user who has participated in a thread and then unfollowed the thread will get notifications. Is this intended?

app/notification.go
```diff
@@ -478,28 +557,23 @@ func (a *App) SendNotifications(post *model.Post, team *model.Team, channel *mod
...
-			return nil, errors.Wrapf(err, "cannot get thread %q followers", post.RootId)
-		}
+	if isCRTAllowed && post.RootId != "" {
 		for _, uid := range followers {
-			sendEvent := *a.Config().ServiceSettings.CollapsedThreads == model.CollapsedThreadsDefaultOn
-			// check if a participant has overridden collapsed threads settings
-			if preference, prefErr := a.Srv().Store.Preference().Get(uid, model.PreferenceCategoryDisplaySettings, model.PreferenceNameCollapsedThreadsEnabled); prefErr == nil {
-				sendEvent = preference.Value == "on"
+			if profileMap[uid] == nil {
+				continue
```
ashishbhate, line 563: Adding to our previous offline discussion - I think we should add a comment here (or maybe make a ticket as a known issue). `profileMap` contains all profiles in a channel. A user who follows a thread and then leaves the channel containing the thread wont get a notification. The thread will still appear in the Global Threads Screen, so a user might expect to be notified.

app/notification.go
koox00: @ashishbhate this was made to add a newly mentioned user to the followers array.
I think we are exiting early is there was an existing membership and the user is not following
https://github.com/mattermost/mattermost-server/blob/e90d9970466de01d3daed6f6835e2c409c498a3a/app/notification.go#L230-L235
at least that was my impression.

ashishbhate (approved):
Awesome work @koox00! 💯

jwilander (approved):
Great work on this @koox00 ! Quite the difficult bit of code to dive into for one of your first forays into the server 😛

jgilliam17:
Thanks @koox00 Settings changes look good
few issues with notifications
1. Notification is received for unfollowed thread replies (no mention)  when my account setting for desktop notifications is set to "For all activity"
2. Notification is not received for replies in threads I am following when the setting is "Only for mentions .." and "Notify me about threads ..." is checked

koox00:
> Thanks @koox00 Settings changes look good
> few issues with notifications
> 
>     1. Notification is received for unfollowed thread replies (no mention)  when my account setting for desktop notifications is set to "For all activity"
> 
>     2. Notification is not received for replies in threads I am following when the setting is "Only for mentions .." and "Notify me about threads ..." is checked

I think that the last changes were not deployed on the test server, let me create it again

koox00:
@jgilliam17 It's been deployed correctly now, and should be working as expected :)

jgilliam17:
Thanks @koox00 
2. is fixed, but 1. is still reproducing for me on the test server
```Notification is received for unfollowed thread replies (no mention)  when my account setting for desktop notifications is set to "For all activity"```

File: store/sqlstore/thread_store.go [hunk 9]
```diff
@@ -359,12 +359,25 @@ func (s *SqlThreadStore) GetThreadsForUser(userId, teamId string, opts model.Get
 	return result, nil
 }
 
-func (s *SqlThreadStore) GetThreadFollowers(threadID string) ([]string, error) {
+func (s *SqlThreadStore) GetThreadFollowers(threadID string, fetchOnlyActive bool) ([]string, error) {
 	var users []string
+
+	fetchConditions := sq.And{
```
jwilander, line 365: I don't think you need the `sq.And` here since it's just the one condition

jwilander (approved):
Looks good - just one comment

store/sqlstore/thread_store.go
koox00: @jwilander I get `Incompatible assign` in the re-assignment below if I don't add the `sq.And` when declaring.

ashishbhate (approved):

spirosoik:
@koox00 FYI test server fails and cannot connect to database at all. could you please rebase and retry one more the test server?

jgilliam17:
Thanks @koox00 
Seeing one issues
- Setting: All activity, get a mention in an `unfollowed` thread 
Observed: No notification on first mention, 2nd mention or subsequent replies yes, but that is expected as now the thread is `followed`

esethna:
Did some brief testing, this looks good. Nice work @koox00!

esethna:
@jgilliam17 for any non critical issues, perhaps we can file follow-up tickets and approve these PRs so we can get this and the webapp PR merged?

koox00:
> Thanks @koox00
> Seeing one issues
> 
>     * Setting: All activity, get a mention in an `unfollowed` thread
>       Observed: No notification on first mention, 2nd mention or subsequent replies yes, but that is expected as now the thread is `followed`
>       Edited: Added a separate issue for this to address later https://mattermost.atlassian.net/browse/MM-38002

I think this is because on the first message the thread was not a thread yet. 
I mean if you mention on the third post and have no mention on the second post it will notify.
So it's not first mention, but a mention in the first reply that won't notify the user.

jgilliam17 (approved):
Thanks @koox00 
Tested, looks good to merge.
- Verified thread notifications. 
Lets keep the ticket I created as a follow up for now, we can close after Thu. testing or after release testing if it's no longer needed.

esethna:
@amyblais please cherry pick for v6.0

amyblais:
> @amyblais please cherry pick for v6.0

@esethna Ok, the dev can just use ``/cherry-pick cloud``.

PR #18101, opened 2021-08-10
[MM-37557] Move error out of client4 response

hanzei:
#### Summary
The PR moves the `Error` field out of `model.Response` into a separate return value. This allows for more idiomatic error checking and makes the client easier to use. Also the returned value is `error` instead of `*model.AppError` now.

This change only effects the go driver client4 and all tests that use it. While the PR touches a lot of code, it doesn't impact the production code.

There are a couple of minor additional changes in here:
- There `CreateWebSocketClient*` methods also return an `error` now

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-37557

#### Release Note
```release-note
Move error out of client4 response
```


model/client4.go
```diff
@@ -3962,287 +3943,287 @@ func (c *Client4) UploadLicenseFile(data []byte) (bool, *Response) {
...
-func (c *Client4) GetSamlMetadata() (string, *Response) {
+func (c *Client4) GetSamlMetadata() (string, *Response, error) {
 	r, err := c.DoApiGet(c.GetSamlRoute()+"/metadata", "")
 	if err != nil {
-		return "", BuildErrorResponse(r, err)
+		return "", BuildResponse(r), err
 	}
 	defer closeBody(r)
 	buf := new(bytes.Buffer)
 	_, _ = buf.ReadFrom(r.Body)
```
agnivade, line 4225: Unrelated, but will be good to handle the error here.

model/client4.go
```diff
@@ -4267,94 +4248,94 @@ func fileToMultipart(data []byte, filename string) ([]byte, *multipart.Writer, e
...
 	body, writer, err := fileToMultipart(data, filename)
 	if err != nil {
-		return false, &Response{Error: NewAppError("UploadSamlIdpCertificate", "model.client.upload_saml_cert.app_error", nil, err.Error(), http.StatusBadRequest)}
+		return false, nil, NewAppError("UploadSamlIdpCertificate", "model.client.upload_saml_cert.app_error", nil, err.Error(), http.StatusBadRequest)
 	}
 
-	_, resp := c.DoUploadFile(c.GetSamlRoute()+"/certificate/idp", body, writer.FormDataContentType())
-	return resp.Error == nil, resp
+	_, resp, err := c.DoUploadFile(c.GetSamlRoute()+"/certificate/idp", body, writer.FormDataContentType())
+	return true, resp, err //TODO
```
agnivade, line 4258: Unhandled TODO? We need to check for `err != nil` here to determine the bool return value. As an improvement, I'd suggest to remove the bool param altogether, since it is derivable from the error variable which we are already returning.

model/client4.go
```diff
@@ -4363,60 +4344,59 @@ func (c *Client4) ResetSamlAuthDataToEmail(includeDeleted bool, dryRun bool, use
...
+	return ComplianceFromJson(r.Body), BuildResponse(r), nil
 }
 
 // DownloadComplianceReport returns a full compliance report as a file.
-func (c *Client4) DownloadComplianceReport(reportId string) ([]byte, *Response) {
+func (c *Client4) DownloadComplianceReport(reportId string) ([]byte, *Response, error) {
 	rq, err := http.NewRequest("GET", c.ApiUrl+c.GetComplianceReportDownloadRoute(reportId), nil)
 	if err != nil {
-		return nil, &Response{Error: NewAppError("DownloadComplianceReport", "model.client.connecting.app_error", nil, err.Error(), http.StatusBadRequest)}
+		return nil, nil, nil
```
agnivade, line 4399: missing error

model/client4.go
```diff
@@ -4425,69 +4405,69 @@ func (c *Client4) DownloadComplianceReport(reportId string) ([]byte, *Response)
 
 	rp, err := c.HttpClient.Do(rq)
 	if err != nil || rp == nil {
-		return nil, &Response{Error: NewAppError("DownloadComplianceReport", "model.client.connecting.app_error", nil, err.Error(), http.StatusBadRequest)}
+		return nil, nil, nil
```
agnivade, line 4408: Same

model/client4.go
```diff
@@ -4633,170 +4609,170 @@ func (c *Client4) GetGroups(opts GroupSearchOpts) ([]*Group, *Response) {
...
 	body, writer, err := fileToMultipart(data, LdapPublicCertificateName)
 	if err != nil {
-		return false, &Response{Error: NewAppError("UploadLdapPublicCertificate", "model.client.upload_ldap_cert.app_error", nil, err.Error(), http.StatusBadRequest)}
+		return false, nil, NewAppError("UploadLdapPublicCertificate", "model.client.upload_ldap_cert.app_error", nil, err.Error(), http.StatusBadRequest)
 	}
 
-	_, resp := c.DoUploadFile(c.GetLdapRoute()+"/certificate/public", body, writer.FormDataContentType())
-	return resp.Error == nil, resp
+	_, resp, err := c.DoUploadFile(c.GetLdapRoute()+"/certificate/public", body, writer.FormDataContentType())
+	return true, resp, err //TODO
```
agnivade, line 4671: Same as above.

model/client4.go
```diff
@@ -4974,923 +4950,923 @@ func (c *Client4) GetOAuthAccessToken(data url.Values) (*AccessResponse, *Respon
...
-func (c *Client4) uploadPlugin(file io.Reader, force bool) (*Manifest, *Response) {
+func (c *Client4) uploadPlugin(file io.Reader, force bool) (*Manifest, *Response, error) {
 	body := new(bytes.Buffer)
 	writer := multipart.NewWriter(body)
 
 	if force {
 		err := writer.WriteField("force", c.boolString(true))
 		if err != nil {
-			return nil, &Response{Error: NewAppError("UploadPlugin", "model.client.writer.app_error", nil, err.Error(), 0)}
+			return nil, nil, nil
```
agnivade, line 5850: Same problem.

model/client4.go
```diff
@@ -5900,817 +5876,816 @@ func (c *Client4) uploadPlugin(file io.Reader, force bool) (*Manifest, *Response
...
 // GetMarketplacePlugins will return a list of plugins that an admin can install.
-func (c *Client4) GetMarketplacePlugins(filter *MarketplacePluginFilter) ([]*MarketplacePlugin, *Response) {
+func (c *Client4) GetMarketplacePlugins(filter *MarketplacePluginFilter) ([]*MarketplacePlugin, *Response, error) {
 	route := c.GetPluginsRoute() + "/marketplace"
-	u, parseErr := url.Parse(route)
-	if parseErr != nil {
-		return nil, &Response{Error: NewAppError("GetMarketplacePlugins", "model.client.parse_plugins.app_error", nil, parseErr.Error(), http.StatusBadRequest)}
+	u, err := url.Parse(route)
+	if err != nil {
+		return nil, nil, nil
```
agnivade, line 5982: Here too

agnivade (commented):
Great work here. I have just reviewed the model package, because I assume the rest of the changes is just to support those.

Looks good in general. Some parts of the PR are incomplete and needs to be fixed.

But other than that, my main feedback is to make the *Response return consistent for all error cases.

model/client4.go
hanzei: Good suggestion. Added in 3ac313456.

hanzei:
Thanks for the review @agnivade :+1:  Based on your feedback I've submitted the following changes:
- In 6f51cde all methods in client4 will not return a `Response` once a request was made to the server
- In 41afad1 methods will not longer use  `bool` to indicate it's success. Check `err` instead.
- 2c3c5e5 is a cleanup of `model.AppError` in `web`
- In 9d7e3b7 the `client4.Must` methods gets removed forcing more idiomatic error checking

model/client4.go
hanzei: Good suggestion. Added that in 41afad1.

streamer45 (approved):
Awesome :+1:

PR #18102, opened 2021-08-10
[MM-37760] Reduce export surface of client4

hanzei:
#### Summary
To simplify the API exposed from `client4` this PR un-exports the `Get*Route` methods.

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-37760

#### Release Note
```release-note
Reduce export surface of client4
```


api4/bot_test.go
```diff
@@ -465,7 +465,7 @@ func TestPatchBot(t *testing.T) {
 		CheckCreatedStatus(t, resp)
 		defer th.App.PermanentDeleteBot(createdBot.UserId)
 
-		r, err := th.Client.DoApiPut(th.Client.GetBotRoute(createdBot.UserId), `{"creator_id":"`+th.BasicUser2.Id+`"}`)
+		r, err := th.Client.DoApiPut("/bots/"+createdBot.UserId, `{"creator_id":"`+th.BasicUser2.Id+`"}`)
```
agnivade, line 468: Can we use the actual method calls themselves here rather than us constructing the URLs? That seems to be a better solution here.

Same for the rest of the _test.go files.

agnivade (commented):
Also would be good to remove the "get" prefix from the methods to make them more idiomatic Go. 0/5.

api4/bot_test.go
streamer45: I think the problem is that we are unexporting them so they can't be called from the `api4` package.

agnivade: I meant use the high level methods like - `client.PatchBot`, `client.UpdateConfig` etc. Instead of manually calling Put/Post with the URL.

hanzei: Good suggestion @agnivade :+1: Given that this PR needs to be merged soon I'm incident to deffer these changes to another (HW)-ticket. Does that sound reasonable to you?

hanzei:
I like removing the `get` prefix. Added that in 4c88ed516.

api4/bot_test.go
hanzei: https://mattermost.atlassian.net/browse/MM-37822

streamer45 (approved):
Nice cleanup :+1:

api4/bot_test.go
yedamao: @hanzei @agnivade These cases seems like can't use the high level methods provided by client. 

In the above case of patch bot, that pass the `creator_id` to the patch api and client.PatchBot can't do this.

agnivade: I would need to check. Are all cases like that?

yedamao: Yes

hanzei: Thanks for confirming that. In that case we can close the ticket.

PR #18123, opened 2021-08-12
Improve session cleanup

agnivade:
- We move logging statements to the upper layer.
Store functions are low-level methods and should return error
upwards rather than logging.
- Used IN instead of any (array ()) which is equivalent.
- Made the delay to be of type time.Duration and un-exported it.
- Unexported the batch size constant.

```release-note
NONE
```

### DIFF
--- a/app/server.go
+++ b/app/server.go
@@ -1511,11 +1511,15 @@ func doCommandWebhookCleanup(s *Server) {
 }
 
 const (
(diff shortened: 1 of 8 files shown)


jwilander (approved):
LGTM

isacikgoz (approved):
Nice 👍

PR #18133, opened 2021-08-13
Prepare support for sqlx

agnivade:
Just setting up some basic skeleton to start
replacing gorp queries with sqlx.

There is still an issue with efficiently writing
cross-platform named queries, because :FieldName
needs to be :fieldname for Postgres.

The positional params can be rebound depending on
driver name. But named queries can't from my
investigation. Will look into this.

```release-note
NONE
```


jespino:
Two small concerns that I'm having now are:

1. We are missing with this change the query timeout (I guess we are going to address that with something like QueryXContext, making that explicit to every query, or creating a `queryX` function in the store to address that (but is important to not forget about it).
2. We are missing also the tracing, that is not supported in sqlx, and I don't know how we are going to address that (there is a fork from DataDog that does exactly that, but I don't know how up-to-date it is).

Other than that, the PR looks good to me.

agnivade:
Excellent points @jespino ! I will note them and appropriately come up with a solution. Thank you!

jespino (approved):
LGTM

store/sqlstore/store.go
```diff
@@ -137,11 +138,18 @@ type SqlStoreStores struct {
...
+	srCounter int64
+
+	master  *gorp.DbMap
+	masterX *sqlx.DB
+
+	Replicas  []*gorp.DbMap
+	ReplicaXs []*sqlx.DB
+
+	searchReplicas  []*gorp.DbMap
+	searchReplicaXs []*sqlx.DB
```
streamer45, line 151: Is the plan to keep both (gorp and sqlx) for a while or just during development?

streamer45 (approved):

store/sqlstore/store.go
agnivade: Yes, the plan is to keep them both during the transition period. And then remove them, and change all (Master|Replica)X methods to (Master|Replica) ones.

streamer45: Sounds good :+1:

agnivade:
Gentle ping @isacikgoz / @mgdelacroix. I have the migration PRs waiting to be sent out. And there's the other [PR](https://github.com/mattermost/mattermost-server/pull/18148) in draft stage.

isacikgoz (approved):
LGTM 👍

mgdelacroix (approved):
LGTM!! :+1:

PR #18138, opened 2021-08-13
[MM-22051] Remove To/From JSON methods from model

streamer45:
#### Summary

Removed To/From JSON for several structures (see commits).

#### Ticket

https://mattermost.atlassian.net/browse/MM-22051

#### Release Note

```release-note
NONE
```


app/group.go
```diff
@@ -122,7 +128,11 @@ func (a *App) DeleteGroup(groupID string) (*model.Group, *model.AppError) {
 
 	if err == nil {
 		messageWs := model.NewWebSocketEvent(model.WebsocketEventReceivedGroup, "", "", "", nil)
-		messageWs.Add("group", deletedGroup.ToJson())
+		groupJSON, jsonErr := json.Marshal(deletedGroup)
+		if jsonErr != nil {
+			mlog.Warn("Failed to encode group to JSON")
+		}
+		messageWs.Add("group", groupJSON)
```
agnivade, line 135: The string conversion is missing.

utils/license.go
```diff
@@ -119,7 +123,13 @@ func GetAndValidateLicenseFileFromDisk(location string) (*model.License, []byte)
 		mlog.Error("Found license key at %v but it appears to be invalid.", mlog.String("filename", fileName))
 		return nil, nil
 	}
-	return model.LicenseFromJson(strings.NewReader(licenseStr)), licenseBytes
+
+	var license model.License
+	if jsonErr := json.Unmarshal([]byte(licenseStr), &license); jsonErr != nil {
```
agnivade, line 128: Shouldn't we do an early return here?

cmd/mattermost/commands/webhook.go
```diff
@@ -374,7 +375,9 @@ func modifyOutgoingWebhookCmdF(command *cobra.Command, args []string) error {
 		return fmt.Errorf("unable to find webhook '%s'", webhookArg)
 	}
 
-	updatedHook := model.OutgoingWebhookFromJson(strings.NewReader(oldHook.ToJson()))
+	var updatedHook model.OutgoingWebhook
+	js, _ := json.Marshal(oldHook)
```
agnivade, line 379: This seems like a place where we should handle the error?

streamer45: All these commands are getting removed so I opted not to make an effort.

utils/license.go
streamer45: Sure, I was keeping the semantic unchanged but it makes sense to throw an error.

api4/config.go
```diff
@@ -167,15 +172,20 @@ func updateConfig(c *Context, w http.ResponseWriter, r *http.Request) {
 		},
 	})
 	if mergeErr != nil {
-		c.Err = model.NewAppError("getConfig", "api.config.update_config.restricted_merge.app_error", nil, err.Error(), http.StatusInternalServerError)
+		c.Err = model.NewAppError("updateConfig", "api.config.update_config.restricted_merge.app_error", nil, err.Error(), http.StatusInternalServerError)
```
hanzei, line 175: Good catch :+1:

api4/config.go
```diff
@@ -291,12 +301,17 @@ func patchConfig(c *Context, w http.ResponseWriter, r *http.Request) {
...
 	})
 	if mergeErr != nil {
-		c.Err = model.NewAppError("getConfig", "api.config.patch_config.restricted_merge.app_error", nil, err.Error(), http.StatusInternalServerError)
+		c.Err = model.NewAppError("patchConfig", "api.config.patch_config.restricted_merge.app_error", nil, err.Error(), http.StatusInternalServerError)
 	}
 
 	w.Header().Set("Cache-Control", "no-cache, no-store, must-revalidate")
 	if c.App.Srv().License() != nil && *c.App.Srv().License().Features.Cloud {
-		w.Write([]byte(cfg.ToJsonFiltered(model.ConfigAccessTagType, model.ConfigAccessTagCloudRestrictable)))
+		js, jsonErr := cfg.ToJSONFiltered(model.ConfigAccessTagType, model.ConfigAccessTagCloudRestrictable)
```
hanzei, line 309: Maybe out of scope of this PR: Should `cjf.ToJSONFiltered` return a `[]byte` for easier output to a writer?

api4/channel.go
```diff
@@ -1467,8 +1467,8 @@ func updateChannelMemberSchemeRoles(c *Context, w http.ResponseWriter, r *http.R
 		return
 	}
 
-	schemeRoles := model.SchemeRolesFromJson(r.Body)
-	if schemeRoles == nil {
+	var schemeRoles model.SchemeRoles
+	if jsonErr := json.NewDecoder(r.Body).Decode(&schemeRoles); jsonErr != nil {
 		c.SetInvalidParam("scheme_roles")
```
hanzei, line 1472: Out of scope of this PR: It would be immensely helpful for API users if the decode errors would also return the json error. Separate ticket?

api4/emoji.go
```diff
@@ -83,15 +82,14 @@ func createEmoji(c *Context, w http.ResponseWriter, r *http.Request) {
 		return
 	}
 
-	emoji := model.EmojiFromJson(strings.NewReader(props["emoji"][0]))
-	if emoji == nil {
+	var emoji model.Emoji
+	if jsonErr := json.Unmarshal([]byte(props["emoji"][0]), &emoji); jsonErr != nil {
 		c.SetInvalidParam("emoji")
-		return
```
hanzei, line 89: Why doesn't the code return here any longer?

api4/job.go
```diff
@@ -167,7 +167,12 @@ func getJobs(c *Context, w http.ResponseWriter, r *http.Request) {
 		return
 	}
 
-	w.Write([]byte(model.JobsToJson(jobs)))
+	js, jsonErr := json.Marshal(jobs)
```
hanzei, line 170: What does `js` stand for? It's not obvious to me.

api4/job.go
```diff
@@ -167,7 +167,12 @@ func getJobs(c *Context, w http.ResponseWriter, r *http.Request) {
...
+	if jsonErr != nil {
+		c.Err = model.NewAppError("getJobs", "api.marshal_error", nil, jsonErr.Error(), http.StatusInternalServerError)
+		return
+	}
+	w.Write(js)
```
hanzei, line 175: What was the reasoning behind using this approach vs. `json.NewDecoder(w)`?

api4/post_test.go
```diff
@@ -355,7 +355,9 @@ func testCreatePostWithOutgoingHook(
 			ResponseType: respPostType,
 		}
 
-		fmt.Fprint(w, outGoingHookResponse.ToJson())
+		hookJSON, jsonErr := json.Marshal(outGoingHookResponse)
+		require.NoError(t, jsonErr)
+		fmt.Fprint(w, string(hookJSON))
```
hanzei, line 360: Nit pick: Is `w.Write` simpler here?

api4/team.go
```diff
@@ -1581,11 +1581,12 @@ func updateTeamScheme(c *Context, w http.ResponseWriter, r *http.Request) {
...
 	}
 
-	schemeID := model.SchemeIDFromJson(r.Body)
-	if schemeID == nil || (!model.IsValidId(*schemeID) && *schemeID != "") {
+	var p model.SchemeIDPatch
+	if jsonErr := json.NewDecoder(r.Body).Decode(&p); jsonErr != nil || p.SchemeID == nil || (!model.IsValidId(*p.SchemeID) && *p.SchemeID != "") {
 		c.SetInvalidParam("scheme_id")
 		return
 	}
+	schemeID := p.SchemeID
```
hanzei, line 1589: I find checking the json err and the semantics in the same statement a bit hard to read. Is the following easier to follow?
```suggestion
	var p model.SchemeIDPatch
	if jsonErr := json.NewDecoder(r.Body).Decode(&p); jsonErr != nil {
		c.SetInvalidParam("scheme_id")
		return
	}
	schemeID := p.SchemeID
	
	if schemeID == nil || (!model.IsValidId(*schemeID) && *shemeID != "") { 
		c.SetInvalidParam("scheme_id")
		return
	}
```

app/busy.go
```diff
@@ -136,7 +137,7 @@ func (b *Busy) ClusterEventChanged(sbs *model.ServerBusyState) {
 	}
 }
 
-func (b *Busy) ToJson() string {
+func (b *Busy) ToJSON() (string, error) {
```
hanzei, line 140: Should this return a `[]byte`?

app/cluster_handlers.go
```diff
@@ -120,7 +121,11 @@ func (s *Server) clusterClearSessionCacheForAllUsersHandler(msg *model.ClusterMe
 }
 
 func (s *Server) clusterBusyStateChgHandler(msg *model.ClusterMessage) {
-	s.serverBusyStateChanged(model.ServerBusyStateFromJson(bytes.NewReader(msg.Data)))
+	var sbs model.ServerBusyState
+	if jsonErr := json.Unmarshal(msg.Data, &sbs); jsonErr != nil {
+		mlog.Warn("Failed to decode server busy state from JSON", mlog.Err(jsonErr))
```
hanzei, line 126: Should the code `return` here?(There are a lot of code pieces like this, so only asking this once)

app/role.go
```diff
@@ -240,7 +242,11 @@ func (a *App) CheckRolesExist(roleNames []string) *model.AppError {
 
 func (a *App) sendUpdatedRoleEvent(role *model.Role) {
 	message := model.NewWebSocketEvent(model.WebsocketEventRoleUpdated, "", "", "", nil)
-	message.Add("role", role.ToJson())
+	roleJSON, jsonErr := json.Marshal(role)
+	if jsonErr != nil {
+		mlog.Warn("Failed to encode role to JSON")
```
hanzei, line 247: Nit pick: Should `jsonErr` also get logged?

app/webhook.go
```diff
@@ -96,7 +97,11 @@ func (a *App) TriggerWebhook(c *request.Context, payload *model.OutgoingWebhookP
 	var body io.Reader
 	var contentType string
 	if hook.ContentType == "application/json" {
-		body = strings.NewReader(payload.ToJSON())
+		js, jsonErr := json.Marshal(payload)
+		if jsonErr != nil {
+			mlog.Warn("Failed to encode to JSON", mlog.Err(jsonErr))
+		}
+		body = strings.NewReader(string(js))
```
hanzei, line 104: Why `strings.Reader` over `bytes.Reader`?

hanzei (changes requested):
Great work @streamer45 :+1: 

While I don't want to block the PR to long, I'm wondering if the `ToJSON` methods should all return a `[]byte` instead of `string`.

api4/channel.go
streamer45: Agreed. Created [MM-37920](https://mattermost.atlassian.net/browse/MM-37920).

api4/emoji.go
streamer45: Thanks, that looks like a mistake.

api4/job.go
streamer45: It stands for json, given the full name is taken by the package.

api4/job.go
streamer45: No strong reason, just wanted to separate the error type and make sure that if we fail with that code is due to marshaling and not some other error writing the response. We can change it later on to make it consistent over all endpoints.

api4/post_test.go
streamer45: Sure :+1:

app/cluster_handlers.go
streamer45: I've decided to keep the existing semantic and just logging to avoid unexpected results. We can revisit all these in the near future.

app/role.go
streamer45: Ideally yes, I'll try to update these.

app/webhook.go
streamer45: It was the quickest change that left the semantic unchanged. Had hit a couple of places where using bytes wasn't working as expected. This one seems safe though so I can update.

api4/config.go
hanzei: Is that ready the case for `ToJSONFiltered`? It's only used in `api74/config.go`.

api4/job.go
hanzei: 0/5 maybe use `out` or `jsonOut` as it's more obvious.

agnivade (approved):

api4/job.go
streamer45: I am going to defer this as it's a minor detail and proper replacement could take a bit.

api4/config.go
streamer45: Updated.

api4/team.go
streamer45: It's a bit dense I admit. Updated.

app/busy.go
streamer45: Done.

hanzei (approved):
Awesome :tada:

PR #18148, opened 2021-08-17
Added a sqlXPassThrough layer

agnivade:
This layer embeds the original sqlx object and wraps some
key methods to avoid some boilerplate.

- Add timeout to all methods from the timeout in settings.
- Added trace functionality.
- Added auto-rebind capabilities so that each store method
doesn't need to call rebind every time.
- Added auto-lowering of query string to support NamedExec
for both MySQL and Postgres

All methods are not wrapped as all are likely not to be called
from our code. As and when they are called, support will be added.

```release-note
NONE
```


agnivade:
This is meant to go after https://github.com/mattermost/mattermost-server/pull/18133. Just sending a draft PR for now until that is merged. But please feel free to review.

store/sqlstore/sqlx_wrapper.go
```diff
@@ -0,0 +1,251 @@
...
+	if w.trace {
+		printArgs(query, args)
+	}
+
+	return w.DB.GetContext(ctx, dest, query, args...)
+}
+
+func (w *sqlxDBWrapper) NamedExec(query string, arg interface{}) (sql.Result, error) {
+	if w.DB.DriverName() == model.DatabaseDriverPostgres {
+		query = strings.ToLower(query)
```
jespino, line 56: I have doubts about this, I can see somebody passing something directly in the query (for example a case sensitive value) and if that happen is going to be really really hard to debug.

agnivade: This was a simple solution to what could turn out slightly more complicated. From my understanding, only quoted identifiers can run into this problem. Postgres automatically converts any unquoted identifiers to [lower case](https://www.postgresql.org/docs/12/sql-syntax-lexical.html#SQL-SYNTAX-IDENTIFIERS).

So far, I don't believe we use quoted identifiers anywhere in our queries. Any query like `UPDATE "Posts" SET "Id" = "xyz";` will automatically fail because the table "Posts" doesn't exist.

As long as we don't use quoted strings while adding new tables during schema changes, I don't believe this should be a problem.

It's also possible that I might have missed something. Wondering if you can show an example where a valid Postgres query will fail with this logic in our existing SQL schema.

jespino: I was thinking something more in the line of `UPDATE Channels Set Type = "P" WHERE Id = :ChannelId`, you are passing the parameter there, and it is ok, but in the query string you are passing the a value too, in this case the `Type = "P"` is going to be converted to `Type = "p"` that doesn't mean the same thing. I'm not saying that the code has something like that anywhere, but is something that can happen and is hard to track.

agnivade: Indeed, you are right, I will look into it. Thanks @jespino!

agnivade: Improved slightly with regex. It's still not 100% fool-proof, but I think should handle all common cases. If we see those corner cases, then we'll have to add an escape hatch and directly pass separate queries.

jespino: I'm not strongly against this approach, but it still can leads to obscure bugs, actually now, is harder to generate the bug, but at the same time the bug is more obscure. for example `UPDATE Whatever SET value = "Whatever"` is going to work, but for example something like `"UPDATE Whatever SET composedValue = \"" + team.Id + ":" + channel.Id + "\""`, would generate a super weird bug.

Said that, I think that is going to be very unlikely, so I'm ok following this path, but we need to know that there is a possible pitfall.

agnivade: Yes, it will still fail, and I have mentioned that in the comment. I did take a look at the parsing logic in sqlx, but it was failing for some scenarios. I will take another look to see if I can incorporate it.

agnivade: Ah okay, so I just verified, it fails properly with both sqlx and gorp. I tried with something like

`INSERT INTO oauthapps
	(id, creatorid)
VALUES
(:id, 'ji:random')
`

and passed a struct of `struct{ Id string }{Id: "sddsd"}`. Expecting it to silently insert the row. But it didn't.

Sqlx failed with `could not find name random in struct { Id string }{Id:"sddsd"}`.

And gorp failed with `pq: syntax error at or near ":"`.

So given that it's hard to deliberately introduce any bug, I don't think there is any need to specially handle anything. We can just keep this current logic.

jespino: This sounds wrong, how in the hell a string insertion with the text `:random` inside is getting replaced with a SQL param?, I guess there the problem is the `'` instead of `"` or something like that. But string contents shouldn't be replaced by SQL params.

agnivade: @jespino - You are most welcome to try it yourself if you don't believe my evidence :)

But it makes sense if you think about it. Libraries will just do a basic string replace of strings starting with `:`. Otherwise they need to manually parse the SQL queries which is not feasible. So a named query with constant strings containing `:` is just not allowed. Users need to use manual exec for that.

And regarding `'` and `"`, in Postgres `'` is used for [string constants](https://www.postgresql.org/docs/current/sql-syntax-lexical.html#SQL-SYNTAX-STRINGS) and `"` is used to table and [column names](https://www.postgresql.org/docs/current/sql-syntax-lexical.html#SQL-SYNTAX-IDENTIFIERS). So it's not possible to write `INSERT INTO oauthapps (id, creatorid) VALUES (:id, "randomstring")`. That's an invalid syntax.

agnivade: Ok, I double checked. Sqlx definitely fails always. Gorp fails sometimes and passes in some cases. So it's even more inconsistent in that sense . Atleast sqlx clearly fails always.

store/sqlstore/sqlx_wrapper.go
```diff
@@ -0,0 +1,259 @@
...
+	case 0:
+		mlog.Debug(query)
+	default:
+		fields := make([]mlog.Field, 0, len(args))
+		for i, arg := range args {
+			fields = append(fields, mlog.Any("arg"+strconv.Itoa(i), arg))
+		}
+		mlog.Debug(query, fields...)
+	}
+}
```
streamer45, line 259: Gorp tracing had a nice addition of the query's execution time. It would be nice to retain that.

store/sqlstore/sqlx_wrapper.go
```diff
@@ -0,0 +1,259 @@
...
+type sqlxDBWrapper struct {
+	*sqlx.DB
+	queryTimeout time.Duration
+	trace        bool
+}
+
+func newSqlxDBWrapper(db *sqlx.DB, timeoutSecs int, trace bool) *sqlxDBWrapper {
+	return &sqlxDBWrapper{
+		DB:           db,
+		queryTimeout: time.Duration(timeoutSecs) * time.Second,
```
streamer45, line 36: Should we make the input parameter to be a `time.Duration` from the start?

agnivade: Can I say the word .. ? :rofl: I am ___. :)

PR #18152, opened 2021-08-17
[MM-23279] Idiomatic naming (IP)

hanzei:
#### Summary
This is a the third PR that changes identifies to use more idiomatic naming.

I intentional didn't change the field names of structs that are used to create DB columns as `gorp` doesn't like it when you change field names.

In the future we might be able to enforce this via golangci-lint.

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-23279

#### Release Note
```release-note
Renamed plugin.Context.IpAddress to plugin.Context.IPAddress
Renamed fields in the model package to have more idiomatic names
```

### DIFF
--- a/model/cluster_info.go
+++ b/model/cluster_info.go
@@ -12,7 +12,7 @@ type ClusterInfo struct {
 	Id         string `json:"id"`
 	Version    string `json:"version"`
 	ConfigHash string `json:"config_hash"`
-	IpAddress  string `json:"ipaddress"`
+	IPAddress  string `json:"ipaddress"`
 	Hostname   string `json:"hostname"`
 }
 
--- a/app/command_autocomplete.go
+++ b/app/command_autocomplete.go
@@ -259,7 +259,7 @@ func (a *App) getDynamicListArgument(c *request.Context, commandArgs *model.Comm
 	pluginContext := pluginContext(c)
 	params.Add("request_id", pluginContext.RequestId)
 	params.Add("session_id", pluginContext.SessionId)
-	params.Add("ip_address", pluginContext.IpAddress)
+	params.Add("ip_address", pluginContext.IPAddress)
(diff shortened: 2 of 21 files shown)


agnivade:
The code freeze has started. Do we still want to get this in?

hanzei:
@agnivade Yes, I would like to get this in. It's mostly a naming change which will help with getting the naming on the code base consistently.

streamer45 (approved):

agnivade:
Ok, just coordinate with @streamer45 to avoid merge conflicts as much as possible :)

model/cluster_info.go
```diff
@@ -12,7 +12,7 @@ type ClusterInfo struct {
 	Id         string `json:"id"`
 	Version    string `json:"version"`
 	ConfigHash string `json:"config_hash"`
-	IpAddress  string `json:"ipaddress"`
+	IPAddress  string `json:"ipaddress"`
```
agnivade, line 15: Please ensure that there is no issue running this on a cluster with older nodes, so that we can give SRE team prior heads up if needed.

I don't think there will be, since the `json` attribute is the same. But good to check.

agnivade (approved):
Thanks Ben.

agnivade:
Also, will be good to update the release notes with the package API changes made.

model/cluster_info.go
hanzei: I can confirm that the JSON payload send over the wire is the same with and without this change. I didn't test a cluster with older nodes as I didn't find an easy way to set this up. Does my testing cover your concerns?

agnivade: Yes, I can see that the JSON payload will not change. This was more of a sanity check.

You can just spin up docker containers and run separate versions. But it's okay to leave it if that is taking time. Thanks Ben.

PR #18189, opened 2021-08-18
MM-38004 fix racy unit test

wiggin77:
#### Summary
This PR fixes a racy unit test.  Error logging was resolving all `error` interfaces to a string during formatting which happens asynchronously in a log writer thread.  In this case, due to a recent change for AppErrors, the AppError's fields are modified after the log call thus causing a race.  The fix is to resolve the error to a string in the calling thread and not storing the original `error`.

Other structs and interfaces are already resolved to strings by the caller - `error` was a special case because it has a specific `mlog.Err` generating the log field.

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-38004

#### Release Note
```release-note
NONE
```

### DIFF
--- a/shared/mlog/mlog.go
+++ b/shared/mlog/mlog.go
@@ -103,18 +103,25 @@ var String = logr.String
 
 // Stringer constructs a field containing a key and a fmt.Stringer value.
 // The fmt.Stringer's `String` method is called lazily.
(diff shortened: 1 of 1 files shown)


agnivade (approved):
Thanks!

PR #18207, opened 2021-08-18
MM-38016 Fix racy unit test (TestService_AddTopicListener)

wiggin77:
#### Summary
While monitoring CI builds after merging mlog refactor I noticed a build fail with a race condition due to the new mlog.  https://app.circleci.com/pipelines/github/mattermost/mattermost-server/27034/workflows/8750c72a-693b-4cf1-9d41-e9e6eabf3da9/jobs/187950/tests#failed-test-1

~~The issue is a new test helper within mlog that creates a logger using `testing.T` to write the logs.  Unfortunately `testing.T`  is not thread safe and cannot be used in this way.  The fix is to remove this utility (CreateTestLogger) and modify the unit tests to use a different logger.~~

The actual issue is a missed `shutdown` call for a mockServer.  Thanks @agnivade .

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-38016

#### Release Note
```release-note
NONE
```

### DIFF
--- a/services/remotecluster/mocks_test.go
+++ b/services/remotecluster/mocks_test.go
@@ -5,7 +5,6 @@ package remotecluster
 
 import (
 	"context"
-	"testing"
 
 	"github.com/mattermost/mattermost-server/v6/einterfaces"
 	"github.com/mattermost/mattermost-server/v6/model"
@@ -21,8 +20,8 @@ type mockServer struct {
 	user    *model.User
 }
 
-func newMockServer(t *testing.T, remotes []*model.RemoteCluster) *mockServer {
-	testLogger := mlog.CreateTestLogger(t, nil, mlog.StdAll...)
+func newMockServer(remotes []*model.RemoteCluster) *mockServer {
+	testLogger := mlog.CreateConsoleTestLogger(true, mlog.LvlDebug)
 
 	return &mockServer{
 		remotes: remotes,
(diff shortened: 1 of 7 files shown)


agnivade:
The race isn't coming from `TestService_AddTopicListener`. It's coming from `TestService_sendProfileImageToRemote`. If you look at the logs, you will see that the race is original from the previous test. 

>  Unfortunately testing.T is not thread safe and cannot be used in this way. 

`testing.T` is indeed thread safe.

https://pkg.go.dev/testing#T

> The other reporting methods, such as the variations of Log and Error, may be called simultaneously from multiple goroutines.

What is happening here is after `TestService_sendProfileImageToRemote` finishes, it still logs in some goroutine. The current approach is okay, but it masks the fact that the logging still happens after a test is over.

I believe adding `defer mockServer.Shutdown()` after `mockServer := newMockServer(t, makeRemoteClusters(NumRemotes, ts.URL))` in `TestService_sendProfileImageToRemote` should fix the race.

wiggin77:
> I believe adding `defer mockServer.Shutdown()`

Thanks @agnivade. You're right, this was due to a missed mockerServer shutdown.  I checked and there are no other missing shutdowns.

I going to leave the `testing.T` logger deleted.  It was only used in two places and we should stick to only one way for logging during unit tests.

agnivade (approved):
Thanks Doug. Looks great.

streamer45 (approved):

PR #18210, opened 2021-08-18
Combine alterations on Posts table for 6.0 schema migration

streamer45:
#### Summary

Given most of the time is spent copying the `Posts` table we combine the three `ALTER` statements on it. This more or less reduces the total time taken by a factor of 3.

```
{"timestamp":"2021-08-18 17:03:39.973 Z","level":"debug","msg":"sql-trace: ALTER TABLE Posts MODIFY COLUMN FileIds text, MODIFY COLUMN Props JSON, DROP COLUMN ParentId [] (26m9.43801248s)","caller":"sqlstore/adapters.go:65"}
```

#### Release Note

```release-note
NONE
```


wiggin77 (approved):
LGTM :+1:  Too bad MySQL and Postgres have such different alter syntax.

streamer45:
> LGTM  Too bad MySQL and Postgres have such different alter syntax.

Yeah, also in this particular case there isn't much to gain in doing the same on Postgres. Two out of three statements are more or less instant.

store/sqlstore/upgrade.go
```diff
@@ -1301,23 +1301,37 @@ func fixCRTChannelMembershipCounts(sqlStore *SqlStore) {
 func upgradeDatabaseToVersion600(sqlStore *SqlStore) {
 	// if shouldPerformUpgrade(sqlStore, Version5380, Version600) {
 
+	if sqlStore.DriverName() == model.DatabaseDriverMysql {
+		if sqlStore.DoesColumnExist("Posts", "ParentId") {
+			sqlStore.GetMaster().ExecNoTimeout("ALTER TABLE Posts MODIFY COLUMN FileIds text, MODIFY COLUMN Props JSON, DROP COLUMN ParentId")
```
agnivade, line 1306: This excludes the column exist check for FileIds and Props. Although it is highly unlikely a posts table would exist without those columns.

streamer45: Yes I am assuming those exist otherwise the instance is broken to start with.

streamer45:
> And just to confirm, these changes will be a noop if the column already exists right? Just ensuring this will not cause any issues with Community.

If `ParentId` exists it gets dropped. If it doesn't then we won't attempt dropping it. Altering the column type is idempotent and if the type is already migrated it will return instantly.

PR #18226, opened 2021-08-19
Fixes race condition in notification.go

koox00:
#### Release Note

```release-note
NONE
```


app/notification.go
```diff
@@ -258,7 +260,7 @@ func (a *App) SendNotifications(post *model.Post, team *model.Team, channel *mod
 				// add new followers to existing followers
 				if threadMembership.Following && !followers.Contains(userID) {
```
agnivade, line 261: Please expand the mutex critical section to include the entire if condition. That should fix the race. I don't believe there is a need to introduce an additional map.

agnivade (approved):

agnivade (approved):
Thanks!

jwilander (approved):

amyblais:
@koox00 Is this for v6.0?

koox00:
> @koox00 Is this for v6.0?

Yes

PR #18229, opened 2021-08-19
Fix for test-data make target

mkraft:
#### Summary

Fixes `make test-data` to work with mmctl now that the cli has removed the `config` and `sampledata` commands.

#### Ticket Link

n/a

#### Release Note

```release-note
NONE
```

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -426,8 +426,9 @@ cover: ## Runs the golang coverage tool. You must run the unit tests first.
 	$(GO) tool cover -html=ecover.out
 
 test-data: run-server ## Add test data to the local instance.
- 	# TODO: Replace sleep with something better.
-	sleep 10
+	@if ! ./scripts/wait-for-system-start.sh; then \
+		make stop; \
+	fi
 
 	@echo ServiceSettings.EnableLocalMode must be set to true.
 
--- a/Makefile
+++ b/Makefile
@@ -425,9 +425,9 @@ cover: ## Runs the golang coverage tool. You must run the unit tests first.
 	$(GO) tool cover -html=cover.out
 	$(GO) tool cover -html=ecover.out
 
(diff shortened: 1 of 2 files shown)


cpoile (approved):
Thank you! I hope the sleep works, I'm a little concerned about that.

mkraft:
> Thank you! I hope the sleep works, I'm a little concerned about that.

Yeah agreed. 

> Thank you! I hope the sleep works, I'm a little concerned about that.

You're right. I've improved it now, please have a re-review.

cpoile (approved):
Awesome!

Makefile
```diff
@@ -425,9 +425,15 @@ cover: ## Runs the golang coverage tool. You must run the unit tests first.
...
 
-test-data: start-docker ## Add test data to the local instance.
-	$(GO) run $(GOFLAGS) -ldflags '$(LDFLAGS)' $(PLATFORM_FILES) config set TeamSettings.MaxUsersPerTeam 100
-	$(GO) run $(GOFLAGS) -ldflags '$(LDFLAGS)' $(PLATFORM_FILES) sampledata -w 4 -u 60
+test-data: run-server ## Add test data to the local instance.
+	@if ! ./scripts/wait-for-system-start.sh; then \
+		make stop; \
+	fi
+
+	@echo ServiceSettings.EnableLocalMode must be set to true.
```
streamer45, line 433: Wondering if we should automate this using `jq` on `config.json` and force a server restart instead to make sure it gets applied. After all, this is a dev environment we are talking about.

mkraft: I was debating that and thought it seemed like a bit of an overreach and was preferring to iterate on that addition if we think it's necessary in the future. Thoughts?

streamer45: Fair enough :+1:

streamer45 (approved):

PR #18232, opened 2021-08-20
[MM-36872] add threading headers to notification emails

engineereng:
#### Summary
This PR adds email headers to notification emails so the emails can be threaded by mail clients and so that emails are not marked as spam:

- `Message-ID` - the unique identifier for every email in the format `<unique_id@domain.com>` (a.k.a. "email id")
- `In-Reply-To` - the email id of the "parent" email of this email
- `References` - the email id of the "parent" email (as well as the ids of grandparents, great-grandparents, and so on).

These headers are defined by [RFC 5322](https://datatracker.ietf.org/doc/html/rfc5322).

The mail functions' signatures have been changed to add three parameters, each of which correspond to one of these headers. Further PRs should utilize this change to add headers for all of the different types of emails the Mattermost client sends.

For this PR, `Message-ID` uses the format "<post_id@mattermost.com>", where `post_id` is the id of the post the notification email is about. Email IDs should be unique for every email, and `post_id` _should_ be unique enough in the mattermost.com domain, but I'm not 100% sure. RFC 5322 suggests incorporating a timestamp into email IDs to make absolutely certain email IDs are unique, but perhaps a future PR could tackle that task.

`In-Reply-To` is the id of the root post, and `References` is identical to `In-Reply-To`. Normally, `References` contains a list of parents, however I saw that GitHub's emails (which I used as a reference) were threaded correctly with just the direct parent.

Additionally, I added some tests for adding these headers to `mail_test.go`.

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-36872

https://github.com/mattermost/mattermost-server/issues/17862

#### Release Note
```release-note
Added email headers to notification emails so they can be threaded by email clients
```


shared/mail/mail.go
```diff
@@ -234,7 +237,7 @@ func TestConnection(config *SMTPConfig) error {
 	return nil
 }
 
-func SendMailWithEmbeddedFilesUsingConfig(to, subject, htmlBody string, embeddedFiles map[string]io.Reader, config *SMTPConfig, enableComplianceFeatures bool, ccMail string) error {
+func SendMailWithEmbeddedFilesUsingConfig(to, subject, htmlBody string, embeddedFiles map[string]io.Reader, config *SMTPConfig, enableComplianceFeatures bool, messageID string, inReplyTo string, references string, ccMail string) error {
```
BenCookie95, line 240: I just have a small nit, why did we put the new params before `ccMail` in the parameter list? Also, do we think the parameter list is getting a bit long? We could create a type for them, just a suggestion

engineereng: There was no particular reason why I put the new parameters before ccMail.

I'm not sure how creating a type would work, but if there's a way of shortening the parameter list, I'm all for it.

mkraft (approved):

BenCookie95 (approved):

furqanmlk (commented):
Hi @mkraft , I tested this PR on following clients including `gmail`, `apple mail` ,`outlook` and `Thunderbird client` I don't see any problem in any of these clients

**Gmail Client Mail**
image

**Apple Client Mail**
image

**Outlook Client Mail**
image

**Thunderbird client**
image

mkraft:
@furqanmlk Sounds good. So no issues. And we get the much requested message-id header. And how about threading? Is it an improvement in some of those clients?

furqanmlk:
> @furqanmlk Sounds good. So no issues. And we get the much requested message-id header. And how about threading? Is it an improvement in some of those clients?

@mkraft 
Yes threading is working fine, I attached the images of all the clients, including gmail.

PR #18235, opened 2021-08-20
MM-35017: order and inclusion of participants

koox00:
#### Summary

The ordering of participants was done by first reply on the thread.
Later replies of same users never updated the participants order.
Also root poster was included in the participants list.

From now on root poster is not considered a thread participant,
only repliers are participants.

Ordering: Another reply to the thread from the same user should place
the user last at the participants list.

#### Ticket Link

https://mattermost.atlassian.net/browse/MM-35017

#### Release Note

```release-note
NONE
```


agnivade:
Hey @koox00, I don't think I am the right person to review this. I can give a quick look, but I'm afraid I don't much about the CRT logic.

koox00:
Hey @agnivade, please give it a quick look if you can, or re-assign if you like, I'll also ask a review from @jwilander :)

store/sqlstore/user_store.go
```diff
@@ -2040,3 +2040,30 @@ func (us SqlUserStore) IsEmpty(excludeBots bool) (bool, error) {
...
+		}
+
+		if _, err := us.SqlStore.DBFromContext(ctx).Select(&users, queryString, args...); err != nil {
+			return nil, errors.Wrap(err, "failed to find Users")
+		}
+	} else {
+		for _, userId := range userIds {
+			users = append(users, &model.User{Id: userId})
+		}
+	}
```
agnivade, line 2066: Trying to understand the reasoning behind adding this flag and doing this logic in the store layer. All this is doing is creating a clone of the slice only with the id field populated. Any reason we cannot do that in the app layer? That should remove the flag and make this method more straighforward.

store/sqlstore/user_store.go
```diff
@@ -2040,3 +2040,30 @@ func (us SqlUserStore) IsEmpty(excludeBots bool) (bool, error) {
 	}
 	return !hasRows, nil
 }
+
+func (us SqlUserStore) GetParticipantProfilesByIds(ctx context.Context, userIds []string, extended bool, allowFromCache bool) ([]*model.User, error) {
```
agnivade, line 2044: Please add tests for any new store method. See storetest/user_store.go

store/sqlstore/post_store.go
```diff
@@ -2440,9 +2432,10 @@ func (s *SqlPostStore) updateThreadsFromPosts(transaction *gorp.Transaction, pos
 			// metadata exists, update it
 			for _, post := range posts {
 				thread.ReplyCount += 1
-				if !thread.Participants.Contains(post.UserId) {
-					thread.Participants = append(thread.Participants, post.UserId)
+				if thread.Participants.Contains(post.UserId) {
+					thread.Participants = thread.Participants.Remove(post.UserId)
 				}
+				thread.Participants = append(thread.Participants, post.UserId)
```
agnivade, line 2438: Shouldn't this be in an else clause? If we are always unconditionally adding the userid, then it makes the removing part moot.

agnivade (approved):
Please also add a test which verifies this last-created order in the thread rather than first-created.

store/sqlstore/post_store.go
koox00: @agnivade we keep de-duplicated user ids in the participants so when a user replies again the id gets added at the end and the previous value is deleted.

store/sqlstore/user_store.go
koox00: @agnivade, you are correct, also on a second thought this method is unnecessary, I'll refactor and use the `GetProfileByIds`.

koox00:
> Please also add a test which verifies this last-created order in the thread rather than first-created.

Just a note, the order is still first-created like before, the only change is the reordering when a participant replies again.
I'll add a test to verify that.

The other changes are fixes to keep the order of the thread participants as it is in the thread.Participants 
because `GetProfileByIds` orders by username.

agnivade:
> Just a note, the order is still first-created like before, the only change is the reordering when a participant replies again.
I'll add a test to verify that.

Thanks for the clarification Kyriakos!

store/sqlstore/post_store.go
agnivade: Ahh I see now. The participants slice is now order-sensitive.

I don't have much knowledge of CRT, but in general, I would suggest to do it like we are doing it for posts. i.e. keep the raw data and the order as separate items. I'll leave it to your better judgement.

koox00: That seems pretty reasonable, but then we would need to scan the posts data each time we fetch the thread(s).

agnivade: No worries from my end. Once you add a test, it's good to go from my side. :)

jwilander (approved):

agnivade (approved):

ashishbhate (approved):

koox00:
@jgilliam17 I recreated the test server and It seems to be OK now

jgilliam17 (approved):
Thanks @koox00 
Tested, looks good to merge!
- Verified avatar sorting order and inclusion on the thread footer; when viewed in the channel and on thread item in global threads - per ticket description.
(The difference in the order between 2 areas tested did throw me off a bit, but it is as listed in ticket description 🙂 )

PR #18238, opened 2021-08-23
Added a sqlXPassThrough layer

agnivade:
This layer embeds the original sqlx object and wraps some
key methods to avoid some boilerplate.

- Add timeout to all methods from the timeout in settings.
- Added trace functionality.
- Added auto-rebind capabilities so that each store method
doesn't need to call rebind every time.
- Added auto-lowering of query string to support NamedExec
for both MySQL and Postgres

All methods are not wrapped as all are likely not to be called
from our code. As and when they are called, support will be added.

```release-note
NONE
```

### DIFF
--- a/store/sqlstore/sqlx_wrapper.go
+++ b/store/sqlstore/sqlx_wrapper.go
@@ -60,7 +60,7 @@ func (w *sqlxDBWrapper) Get(dest interface{}, query string, args ...interface{})
 	err := w.DB.GetContext(ctx, dest, query, args...)
 
 	if w.trace {
-		printArgs(query, time.Now().Sub(then), args)
+		printArgs(query, time.Since(then), args)
 	}
 
 	return err
@@ -81,7 +81,7 @@ func (w *sqlxDBWrapper) NamedExec(query string, arg interface{}) (sql.Result, er
 	res, err := w.DB.NamedExecContext(ctx, query, arg)
 
 	if w.trace {
-		printArgs(query, time.Now().Sub(then), arg)
+		printArgs(query, time.Since(then), arg)
 	}
 
 	return res, err
@@ -102,7 +102,7 @@ func (w *sqlxDBWrapper) NamedQuery(query string, arg interface{}) (*sqlx.Rows, e
 	rows, err := w.DB.NamedQueryContext(ctx, query, arg)
 
 	if w.trace {
-		printArgs(query, time.Now().Sub(then), arg)
+		printArgs(query, time.Since(then), arg)
 	}
 
 	return rows, err
@@ -121,7 +121,7 @@ func (w *sqlxDBWrapper) QueryRowX(query string, args ...interface{}) *sqlx.Row {
 	row := w.DB.QueryRowxContext(ctx, query, args...)
 
 	if w.trace {
-		printArgs(query, time.Now().Sub(then), args)
+		printArgs(query, time.Since(then), args)
 	}
 
 	return row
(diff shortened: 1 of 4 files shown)


store/sqlstore/sqlx_wrapper.go
```diff
@@ -0,0 +1,342 @@
...
+	var then time.Time
+	if w.trace {
+		then = time.Now()
+	}
+
+	err := w.DB.GetContext(ctx, dest, query, args...)
+
+	if w.trace {
+		printArgs(query, time.Since(then), args)
+	}
```
streamer45, line 64: I'd like to propose a minor syntactic change here. What if do everything trace related in single block by using defer?
Something like:

```go
if w.trace {
   defer func(then time.Time) {
      printArgs(query, time.Since(then), args)
   }(t.Now())
}
```

This avoids the second block and we can also directly return the wrapped method.

store/sqlstore/sqlx_wrapper.go
```diff
@@ -0,0 +1,342 @@
...
+	// but we ignore it for now until someone complains.
+	if unicode.IsSpace(r) && r != ' ' {
+		return -1
+	}
+	return r
+}
+
+func printArgs(query string, dur time.Duration, args ...interface{}) {
+	query = strings.Map(removeSpace, query)
+	switch len(args) {
```
streamer45, line 331: Do we really need a `switch` here? Looks like we are only interested in the `len > 0` case. I am thinking that even the duration can be appended to the mlog fields regardless of the case we are in.

store/sqlstore/sqlx_wrapper_test.go
```diff
@@ -0,0 +1,83 @@
...
+
+			var query string
+			if store.DriverName() == model.DatabaseDriverMysql {
+				query = `SELECT SLEEP(:Timeout);`
+			} else if store.DriverName() == model.DatabaseDriverPostgres {
+				query = `SELECT pg_sleep(:timeout);`
+			}
+			arg := struct{ Timeout int }{Timeout: 2}
+			_, err = tx.NamedQuery(query, arg)
+			require.True(t, err == context.DeadlineExceeded)
```
streamer45, line 47: Could we use `require.ErrorIs()`?

store/sqlstore/sqlx_wrapper.go
```diff
@@ -0,0 +1,342 @@
...
+import (
+	"context"
+	"database/sql"
+	"regexp"
+	"strconv"
+	"strings"
+	"time"
+	"unicode"
+
+	"github.com/jmoiron/sqlx"
```
streamer45, line 15: This could go in its own block.

streamer45 (approved):
Looks good :+1: Left purely non-blocking suggestions.

store/sqlstore/sqlx_wrapper.go
agnivade: Ooh brilliant ..

store/sqlstore/sqlx_wrapper_test.go
agnivade: This is actually specified in the documentation of `database/sql`. Don't believe there is a need to unwrap as of now.

agnivade:
Gentle ping on this @jespino and @mgdelacroix :)

store/sqlstore/sqlx_wrapper_test.go
agnivade: Yep, could be. :+1:

agnivade:
@jespino - I am going to merge this now. But please feel free to leave any comments, and I can send a separate PR.

PR #18240, opened 2021-08-23
Jobs: Remove status as a filter condition

agnivade:
The id is the primary key for the jobs table. There
is no need to filter by old status as well because
it's redundant.

https://community-daily.mattermost.com/boards/workspace/zyoahc9uapdn3xdptac6jb69ic/285b80a3-257d-41f6-8cf4-ed80ca9d92e5/495cdb4d-c13a-4992-8eb9-80cfee2819a4?c=65f70258-4e63-4999-8ab5-395db4cf7023

```release-note
NONE
```

### DIFF
--- a/api4/job_test.go
+++ b/api4/job_test.go
@@ -251,7 +251,7 @@ func TestDownloadJob(t *testing.T) {
 	CheckBadRequestStatus(t, resp)
 
 	job.Data["is_downloadable"] = "true"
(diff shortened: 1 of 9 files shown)


streamer45 (commented):
Changes looks okay, but aren't we altering the semantics slightly here? The previous behaviour was to update a job only if both id and status matched. I am not sure how/if that was used in practice but it made it possible to update a job only if it had the given/expected status.

agnivade:
Discussed offline. Closing.

PR #18244, opened 2021-08-23
Add index on Jobs table

agnivade:
From Grafana charts, GetCountbyStatusandType and
GetNewestJobByStatusesAndType were the two top queries.

Overall, a through look into all job methods leads to
the conclusion of 2 indexes - one on CreateAt, another
a compound index of Status+Type. I have just gone ahead
with the compound index for now. Once the job cleaner
is implemented, I want to take a second look to decide
whether to add the second index or not.

Here is the before-after of the queries:

Query 1:
```
explain analyze select count(*) from jobs where status='error' and type='migrations';
                                                         QUERY PLAN
-----------------------------------------------------------------------------------------------------------------------------
 Aggregate  (cost=103.42..103.43 rows=1 width=8) (actual time=0.032..0.034 rows=1 loops=1)
   ->  Bitmap Heap Scan on jobs  (cost=4.54..103.42 rows=1 width=0) (actual time=0.027..0.028 rows=0 loops=1)
         Recheck Cond: ((type)::text = 'migrations'::text)
         Filter: ((status)::text = 'error'::text)
         ->  Bitmap Index Scan on idx_jobs_type  (cost=0.00..4.54 rows=34 width=0) (actual time=0.018..0.019 rows=0 loops=1)
               Index Cond: ((type)::text = 'migrations'::text)

explain analyze select count(*) from jobs where status='error' and type='migrations';
                                                         QUERY PLAN
----------------------------------------------------------------------------------------------------------------------------
Aggregate  (cost=8.31..8.32 rows=1 width=8) (actual time=0.079..0.080 rows=1 loops=1)
   ->  Index Only Scan using jobs_multi on jobs  (cost=0.29..8.30 rows=1 width=0) (actual time=0.072..0.073 rows=0 loops=1)
         Index Cond: ((status = 'error'::text) AND (type = 'migrations'::text))
         Heap Fetches: 0

explain analyze select * from jobs where status='error' and type='migrations' order by createat desc limit 1;
```

Query 2:
```
                                                            QUERY PLAN
----------------------------------------------------------------------------------------------------------------------------------
 Limit  (cost=8.31..8.32 rows=1 width=187) (actual time=0.037..0.039 rows=0 loops=1)
   ->  Sort  (cost=8.31..8.32 rows=1 width=187) (actual time=0.035..0.036 rows=0 loops=1)
         Sort Key: createat DESC
         Sort Method: quicksort  Memory: 25kB
         ->  Index Scan using idx_jobs_type on jobs  (cost=0.29..8.30 rows=1 width=187) (actual time=0.027..0.027 rows=0 loops=1)
               Index Cond: ((type)::text = 'migrations'::text)
               Filter: ((status)::text = 'error'::text)

explain analyze select * from jobs where status='error' and type='migrations' order by createat desc limit 1;
                                                          QUERY PLAN
-------------------------------------------------------------------------------------------------------------------------------
 Limit  (cost=8.31..8.32 rows=1 width=187) (actual time=0.065..0.067 rows=0 loops=1)
   ->  Sort  (cost=8.31..8.32 rows=1 width=187) (actual time=0.063..0.064 rows=0 loops=1)
         Sort Key: createat DESC
         Sort Method: quicksort  Memory: 25kB
         ->  Index Scan using jobs_multi on jobs  (cost=0.29..8.30 rows=1 width=187) (actual time=0.021..0.022 rows=0 loops=1)
               Index Cond: (((status)::text = 'error'::text) AND ((type)::text = 'migrations'::text))

```

```release-note
NONE
```

https://community-daily.mattermost.com/boards/workspace/zyoahc9uapdn3xdptac6jb69ic/285b80a3-257d-41f6-8cf4-ed80ca9d92e5/495cdb4d-c13a-4992-8eb9-80cfee2819a4?c=65f70258-4e63-4999-8ab5-395db4cf7023
#### Release Note

```release-note

```


store/sqlstore/upgrade.go
```diff
@@ -1349,6 +1349,8 @@ func upgradeDatabaseToVersion600(sqlStore *SqlStore) {
 		sqlStore.AlterColumnTypeIfExists("Posts", "FileIds", "text", "varchar(300)")
 	}
 
+	sqlStore.CreateCompositeIndexIfNotExists("idx_jobs_status_type", "Jobs", []string{"Status", "Type"})
```
streamer45, line 1352: Is this planned for 6.0? If not, I'd suggest creating the 6.1 migration function.

streamer45 (commented):
I am wondering if this is any better than extending the existing index on Type to also cover Status?

Also, I agree on waiting for more data before adding a third index. As discussed internally I think the Jobs table needs some maintenance job (no pun intended) as there is no real need for letting it grow indefinitely.

store/sqlstore/upgrade.go
agnivade: It's not. That's why I added the wait tag. I am waiting for 6.0 comment to be removed. But yeah, I can do a 6.1 as well.

agnivade:
> I am wondering if this is any better than extending the existing index on Type to also cover Status?

Good point. There were also some queries filtering by status. I'll check those, and if it's not that important, I'll just expand the existing index.

isacikgoz (approved):
Looks good 🦶🌔

agnivade:
@streamer45 - Good to go for 6.1 now. Wondering if we want to mention the SQL queries to pre-run these? Doesn't feel necessary, but let me know.

Also cc @spirosoik / @stylianosrigas - This migration adds an index on the Jobs table. It should not take much time, but we will let you know if this needs to go via the SRE route.

streamer45:
> @streamer45 - Good to go for 6.1 now. Wondering if we want to mention the SQL queries to pre-run these? Doesn't feel necessary, but let me know.

I think it wouldn't hurt to mention something in the upgrade notes. I wouldn't be surprised if some big customers have millions of rows in the table.

streamer45 (approved):

agnivade:
@spirosoik - Community has 400k rows, and I checked that with around 465k rows, the index addition takes 2.27 seconds. Is that acceptable time to let this PR be merged by itself?

PR #18253, opened 2021-08-23
MM-37787: ThreadAutoFollow must be true to enable CollapsedThreads

ashishbhate:
#### Summary

- [ThreadAutoFollow](https://docs.mattermost.com/configure/configuration-settings.html#automatically-follow-threads) must be set to true to enable CollapsedThreads. 

#### Ticket Link

- https://mattermost.atlassian.net/browse/MM-37787

#### Release Note

```release-note
The ThreadAutoFollow setting must be set to true to enable CollapsedThreads. Previously, this was stated the Mattermost documentation but not enforced - https://docs.mattermost.com/configure/configuration-settings.html#automatically-follow-threads
```

### DIFF
--- a/model/config_test.go
+++ b/model/config_test.go
@@ -1472,13 +1472,15 @@ func TestConfigServiceSettingsIsValid(t *testing.T) {
 	cfg := Config{}
 	cfg.SetDefaults()
 
-	err := cfg.ExportSettings.isValid()
+	err := cfg.ServiceSettings.isValid()
 	require.Nil(t, err)
 
 	*cfg.ServiceSettings.CollapsedThreads = CollapsedThreadsDisabled
+	err = cfg.ServiceSettings.isValid()
 	require.Nil(t, err)
 
 	*cfg.ServiceSettings.ThreadAutoFollow = false
+	err = cfg.ServiceSettings.isValid()
 	require.Nil(t, err)
 
 	*cfg.ServiceSettings.CollapsedThreads = CollapsedThreadsDefaultOff
--- a/model/config_test.go
+++ b/model/config_test.go
@@ -1467,3 +1467,27 @@ func TestConfigExportSettingsIsValid(t *testing.T) {
 	require.NotNil(t, err)
 	require.Equal(t, "model.config.is_valid.export.retention_days_too_low.app_error", err.Id)
 }
+
+func TestConfigServiceSettingsIsValid(t *testing.T) {
+	cfg := Config{}
+	cfg.SetDefaults()
+
+	err := cfg.ExportSettings.isValid()
+	require.Nil(t, err)
+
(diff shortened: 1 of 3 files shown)


agnivade (approved):

model/config_test.go
```diff
@@ -1467,3 +1467,27 @@ func TestConfigExportSettingsIsValid(t *testing.T) {
...
+	cfg.SetDefaults()
+
+	err := cfg.ExportSettings.isValid()
+	require.Nil(t, err)
+
+	*cfg.ServiceSettings.CollapsedThreads = CollapsedThreadsDisabled
+	require.Nil(t, err)
+
+	*cfg.ServiceSettings.ThreadAutoFollow = false
+	require.Nil(t, err)
```
streamer45, line 1482: Was the plan to call `isValid()` here? Otherwise no need to check the error again.

streamer45 (commented):
Wondering what's the purpose of `ThreadAutoFollow` if it's always required to be `true` for the feature to work. Do we actually need two settings?

ashishbhate:
>Wondering what's the purpose of ThreadAutoFollow if it's always required to be true for the feature to work. Do we actually need two settings?

ThreadAutoFollow config value was added in an earlier release to start populating the thread related tables. Later the UI and feature changes were added, along with the `CollapsedThreads` settings. This was done so that when the feature was enabled by a user, the user would already have a corpus of CRT threads.

Currently there are two reasons why there are two settings:

1. If a user disables CRT, his thread memberships following status will be kept updated. In the future, when the user enables CRT again his thread memberships will automatically be up to date even though he hasn't enabled the UI side of the feature. 

2. As you correctly inferred, we can eliminate the `ThreadAutoFollow` setting by updating a user's thread memberships regardless of the user's CRT setting. There are some performance, code  and DB points that need to be resolved first. `ThreadAutoFollow` allows an admin to opt out of the thread memberships updates if he feels that the performance hit, or other issues, are unacceptable. I'd be more comfortable with removing the `ThreadAutoFollow` option once we fix these issues. Some of these points have been highlighted in our documentation - https://docs.mattermost.com/configure/configuration-settings.html#automatically-follow-threads

streamer45 (approved):
Looks good. Thank you @ashishbhate for the additional context :+1:

jgilliam17:
Thanks @ashishbhate 
Tested on the PR test server where ThreadAutoFollow in set to true, Sysadmin has an option to enable CRT - as expected.
Note: Additional testing when ThreadAutoFollow in set to false will be done tomorrow on an custom AWS server. (thanks for setting that up Ashish 🙂 )

amyblais:
@ashishbhate Please help cherry-pick this manually to the ``release-5.39`` branch.

PR #18260, opened 2021-08-24
[MM-36792] limit number of threads returned from SQL store

ashishbhate:
#### Summary

- Limit the number of threads returned in a single SQL store call by using the per_page query param instead of pageSize. Our param handling code automatically limits the number of records that can be requested. To support older mobile clients we continue to support the pageSize param until version 6.0 of the server is the minimum supported server version on mobile. 

Related PRs: 
- https://github.com/mattermost/mattermost-webapp/pull/8700 
- https://github.com/mattermost/mattermost-mobile/pull/5643 
#### Ticket Link

- https://mattermost.atlassian.net/browse/MM-36792

#### Release Note

```release-note
Update API to use `per_page` query parameter instead of `pageSize`. This makes the threads API consistent with other endpoints, and automatically limits the number of requested threads with our param handling code. The `pageSize` query parameter will still be supported till version 6.0 of the server becomes the minimum version required by the mobile client. 
```

### DIFF
--- a/web/params.go
+++ b/web/params.go
@@ -361,18 +361,19 @@ func ParamsFromRequest(r *http.Request) *Params {
 // getPerPageFromQuery returns the PerPage value from the given query.
 // This function should be removed and the support for `pageSize`
 // should be dropped after v1.46 of the mobile app is no longer supported
+// https://mattermost.atlassian.net/browse/MM-38131
 func getPerPageFromQuery(query url.Values) int {
 	val, err := strconv.Atoi(query.Get("per_page"))
 	if err != nil {
 		val, err = strconv.Atoi(query.Get("pageSize"))
-		if err != nil || val < 0 {
-			return PerPageDefault
-		} else if val > PerPageMaximum {
-			return PerPageMaximum
-		}
-		return val
+		// if err != nil || val < 0 {
+		// 	return PerPageDefault
+		// } else if val > PerPageMaximum {
+		// 	return PerPageMaximum
+		// }
+		// return val
 	}
-	if val < 0 {
+	if err != nil || val < 0 {
 		return PerPageDefault
 	} else if val > PerPageMaximum {
 		return PerPageMaximum
--- a/web/params.go
+++ b/web/params.go
@@ -255,8 +255,7 @@ func ParamsFromRequest(r *http.Request) *Params {
 		params.Permanent = val
 	}
 
-	// this should be removed when mobile v1.46 is no longer supported
-	if val, err := strconv.Atoi(query.Get("pageSize")); err != nil || val < 0 {
+	if val, err := strconv.Atoi(query.Get("per_page")); err != nil || val < 0 {
 		params.PerPage = PerPageDefault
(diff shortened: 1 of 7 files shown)


streamer45 (commented):
I think the underlying issue here is that we are not using the commonly used name for the parameter which is currently `per_page` (vs `pageSize`). That is limited by the API to 200 (see `web/params.go`). I am thinking that's where the limitation should be and not at the store level.

ashishbhate:
>I think the underlying issue here is that we are not using the commonly used name for the parameter which is currently per_page (vs pageSize)

IIRC we moved to cursor based paging at the request of the webapp in : https://github.com/mattermost/mattermost-server/pull/16748 (admittedly even before that we should have been using `per_page` instead of `pageSize`)

There is no `page` parameter, so `per_page` doesn't seem like the right semantics, even though it is consistent with params for other requests. I'm 0/5, so I'll defer to @jwilander and you. We'll need to make changes to the webapp and mobile, but that shouldn't be a problem.

jwilander:
I'd lean towards `per_page` so that we're consistent with the other API endpoints, even if the semantics are a little bit different. For cursor paging I don't think `pageSize` is any more or less accurate than `per_page` and I think its better to be consistent for APIs even when it's being consistently bad (not that this is a bad case).

ashishbhate:
Sounds good to me. 

We'll still need to support `pageSize` here till the current mobile version is no longer supported.

streamer45:
We could add a case in `web/params.go` for that so that it relies on the same default. If we do, let's just add a comment and create a ticket to track the future removal.

ashishbhate:
https://mattermost.atlassian.net/browse/MM-38131 to track the removal of the changes here.

web/params.go
```diff
@@ -363,3 +357,25 @@ func ParamsFromRequest(r *http.Request) *Params {
 
 	return params
 }
+
+// getPerPageFromQuery returns the PerPage value from the given query.
+// This function should be removed and the support for `pageSize`
+// should be dropped after v1.46 of the mobile app is no longer supported
```
streamer45, line 363: I'd suggest to either add a `TODO` or the ticket number so that we make this easily discoverable later on.

web/params.go
```diff
@@ -363,3 +357,25 @@ func ParamsFromRequest(r *http.Request) *Params {
...
+func getPerPageFromQuery(query url.Values) int {
+	val, err := strconv.Atoi(query.Get("per_page"))
+	if err != nil {
+		val, err = strconv.Atoi(query.Get("pageSize"))
+		if err != nil || val < 0 {
+			return PerPageDefault
+		} else if val > PerPageMaximum {
+			return PerPageMaximum
+		}
+		return val
```
streamer45, line 373: I think we can remove this and rely on the check below. Just need to check for the error as well.

streamer45 (approved):
Looks good :+1:  Left a couple of non-blocking suggestions.

jwilander (approved):

PR #18266, opened 2021-08-25
Update to bleve v2 and to use the scorch index

mgdelacroix:
#### Summary
This PR updates bleve and makes it use the scorch index on the newly created indexes, which is not only going to be the new default but much more efficient regarding the size of the stored data.

As the new index type is only used on new indexes (old ones get open normally) this should not be a breaking change for existing customers, but they should trigger a reindex to benefit from the improvements of this PR.

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-29916

#### Release Note
```release-note
Updated bleve to v2 and use the scorch index type
```


wiggin77 (approved):
LGTM :+1:   The recommendation to customers to reindex should be in the release note IMO.

lindy65 (approved):
Hey @mgdelacroix - I have smoke-tested around indexing/purging and then did a general test on searches, filters and auto-completion and did not find any issues.  Malik will add his review separately 👍

furqanmlk (approved):
Working as expected.
Performed following steps as Miguel de la Cruz suggested.

1. Started with an old server (master ), trigger the index
2. Performed Smoke test around searches and autocompletion to acknowledge that the index works
3. Without purging the indexes, changed to the PR branch
4. Started the server and performed smoke test, verified  no errors happen in the logs when the new code opens and uses old indexes, and that everything works fine
5. Noted down the size of the index folder to compare to the new indexes later
still with the new code and knowing that the old indexes work, purge and reindex
check that the new indexes work as well as the old ones did.
6. Performed another round of smoke testing around searches and autocompletion
compared the size of the new indexes to the noted size of the old ones. 
7. New ones should be way smaller

Same Steps cause **1.3 MB** size in master
Using this PR, folder size is **520 KB**

jespino (approved):
LGTM

PR #18271, opened 2021-08-25
Unfurl Focalboard Server PR

hahmadia:
# Changes needed for https://github.com/mattermost/focalboard/pull/1081

#### Summary

Changes needed for Unfurl Focalboard 

#### Ticket Link

https://community-daily.mattermost.com/plugins/focalboard/workspace/qgsck6cts3fwpqwyjiupjm5cde/47aa9bb4-6967-4a96-83c7-11bd6b20f1eb/baa742b3-f3f1-406e-811a-00f18d723781?c=c96abe7a-e318-459d-877b-5752cbf9888e

#### Release Note

```release-note
None
```

image

### DIFF
--- a/model/feature_flags.go
+++ b/model/feature_flags.go
@@ -69,7 +69,7 @@ func (f *FeatureFlags) SetDefaults() {
 	f.AddChannelButton = "by_team_name"
 	f.PrewrittenMessages = "none"
 	f.DownloadAppsCTA = "none"
-	f.BoardsUnfurl = false
+	f.BoardsUnfurl = true
 }
 
 func (f *FeatureFlags) Plugins() map[string]string {


sbishel (approved):

jespino (approved):
LGTM

jespino:
Added @hmhealey here because I think he has better knowledge about the Post Embed/Metadata, and I want to know his opinion about not "cleaning up" the Metadata always (It is used from focalboard to pass the Embed object directly in the post interception as part of the plugin).

app/post_metadata.go
```diff
@@ -102,11 +102,14 @@ func (a *App) PreparePostForClient(originalPost *model.Post, isNewPost bool, isE
 
 	a.OverrideIconURLIfEmoji(post)
 
-	post.Metadata = &model.PostMetadata{}
+	if post.Metadata == nil {
+		post.Metadata = &model.PostMetadata{}
+	}
```
hmhealey, line 107: What's this change for? Are we calling this method twice on a single post, or are we providing metadata from somewhere else?

I'd be worried about changing this since it means that calling this method twice on a single post won't result in the same object. If you do that, it'll have a a doubled embeds array

app/post_metadata.go
```diff
@@ -209,6 +214,13 @@ func (a *App) getEmbedForPost(post *model.Post, firstLink string, isNewPost bool
 		}, nil
 	}
 
+	if _, ok := post.GetProps()["boards"]; ok {
+		return &model.PostEmbed{
```
hmhealey, line 218: For message attachment embeds, we don't attach any data to the embed itself because the client knows to look at the props to get that data. Would it make sense to do that with boards embeds as well to avoid duplicating that data?

PR #18297, opened 2021-08-30
MM-37575: Handle nil value in override_username

agnivade:
MM-37575: Handle nil value in override_username (#18057)


isacikgoz (approved):
LGTM 👍

PR #18298, opened 2021-08-30
MM-37576: Fix reference to incorrect error variable (#18066)

agnivade:
We use the correct error variable returned from
the store method prior to it.

https://mattermost.atlassian.net/browse/MM-37576

```release-note
NONE
```


isacikgoz (approved):
Nice one, we should definitely write a checker for this 🤦

PR #18302, opened 2021-08-30
Migrating OAuthStore to sqlx

agnivade:
https://community-daily.mattermost.com/boards/workspace/zyoahc9uapdn3xdptac6jb69ic/285b80a3-257d-41f6-8cf4-ed80ca9d92e5/495cdb4d-c13a-4992-8eb9-80cfee2819a4?c=71efa7c4-53be-4732-87cc-d53726d2cd53

```release-note
NONE
```


agnivade:
Blocked on https://github.com/mattermost/mattermost-server/pull/18300.

agnivade:
Gentle ping @streamer45 @isacikgoz

store/sqlstore/oauth_store.go
```diff
@@ -69,7 +68,10 @@ func (as SqlOAuthStore) SaveApp(app *model.OAuthApp) (*model.OAuthApp, error) {
 		return nil, err
 	}
 
-	if err := as.GetMaster().Insert(app); err != nil {
+	if _, err := as.GetMasterX().NamedExec(`INSERT INTO OAuthApps
+		(Id, CreatorId, CreateAt, UpdateAt, ClientSecret, Name, Description, IconURL, CallbackUrls, Homepage, IsTrusted)
+		VALUES
+		(:Id, :CreatorId, :CreateAt, :UpdateAt, :ClientSecret, :Name, :Description, :IconURL, :CallbackUrls, :Homepage, :IsTrusted)`, app); err != nil {
```
streamer45, line 74: I can see this (and updates) becoming rather verbose for bigger structs. Is there any way we could generalize it a bit without becoming gorp again? :P

agnivade: Nope, I don't think so. The only alternative is to use reflection and therefore become gorp. :P

A slightly better way might be to use squirrel for big queries, just to keep it neat and tidy. I can suggest it more broadly when I demo it.

streamer45: I see. What if we generated a file with a big map of `[struct name]-->[db fields list]`? It's a bit more involved initially but it would avoid reflection and still support future schema alterations. Of course I am just brainstorming here, not asking for any immediate change.

streamer45 (approved):

store/sqlstore/oauth_store.go
agnivade: Definitely possible, but involves some manual work to keep them in sync. Personally, I am not a big fan of creating any abstractions over SQL queries. All of them eventually burn out in the long run, except maybe [sqlc](https://github.com/kyleconroy/sqlc), but it only supports MySQL for now.

I am a bit hesitant to start looking into solutions until we actually see this to be a problem. Writing raw SQL queries are just okay. Let's keep an eye on this for now.

isacikgoz (approved):
LGTM 👍

PR #18313, opened 2021-08-31
Update schemas to include new indexes

streamer45:
#### Summary

New indexes were missing from the previously generated schemas in https://github.com/mattermost/mattermost-server/pull/18223

Not sure how the merge didn't cause a failure though. Maybe we are not running the schema check step over there.

#### Release Note

```release-note
NONE
```

### DIFF
--- a/scripts/mattermost-mysql-6.0.sql
+++ b/scripts/mattermost-mysql-6.0.sql
@@ -1,14 +1,18 @@
--- MySQL dump 10.13  Distrib 5.7.35, for Linux (x86_64)
+-- MariaDB dump 10.19  Distrib 10.5.10-MariaDB, for Linux (x86_64)
 --
--- Host: claudiolt-mysql-instance-1.c0gzspkyf2r5.us-east-1.rds.amazonaws.com    Database: mattermost
+-- Host: 127.0.0.1    Database: mattermost_test
 -- ------------------------------------------------------
--- Server version       5.7.12
+-- Server version	5.7.12
 /*!40103 SET @OLD_TIME_ZONE=@@TIME_ZONE */;
 /*!40103 SET TIME_ZONE='+00:00' */;
 /*!40014 SET @OLD_UNIQUE_CHECKS=@@UNIQUE_CHECKS, UNIQUE_CHECKS=0 */;
 /*!40014 SET @OLD_FOREIGN_KEY_CHECKS=@@FOREIGN_KEY_CHECKS, FOREIGN_KEY_CHECKS=0 */;
 /*!40101 SET @OLD_SQL_MODE=@@SQL_MODE, SQL_MODE='NO_AUTO_VALUE_ON_ZERO' */;
 /*!40111 SET @OLD_SQL_NOTES=@@SQL_NOTES, SQL_NOTES=0 */;
+
+--
+-- Table structure for table `Audits`
+--
 
 /*!40101 SET @saved_cs_client     = @@character_set_client */;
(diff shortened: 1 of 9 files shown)


metanerd:
Sorry, I don't quite understand. What do you mean by over there? Do you mean this? https://github.com/mattermost/mattermost-server/blob/ac252f0771de25d056220b562297a04e1e40eeb0/.gitlab-ci.yml#L128

streamer45:
> Sorry, I don't quite understand. What do you mean by over there? Do you mean this?
> 
> https://github.com/mattermost/mattermost-server/blob/ac252f0771de25d056220b562297a04e1e40eeb0/.gitlab-ci.yml#L128

I was expecting both the original PR (https://github.com/mattermost/mattermost-server/pull/18223) and the subsequent merge into master (https://github.com/mattermost/mattermost-server/commit/ac252f0771de25d056220b562297a04e1e40eeb0) to trigger a failure given the schema is missing things.

metanerd:
OK. The schema tests ran in both CircleCI and GitLab and it succeeded in both with "Both schemas are same", not with some failure and pipeline still succeeds. I don't know enough to check what's causing it.

streamer45:
> OK. The schema tests ran in both CircleCI and GitLab and it succeeded in both with "Both schemas are same", not with some failure and pipeline still succeeds. I don't know enough to check what's causing it.

Need to verify, maybe there's some more difference in the checks between CI and local. The `make test-db-migration` is currently failing on master if you try locally.

scripts/mattermost-mysql-6.0.sql
```diff
@@ -1,15 +1,19 @@
--- MySQL dump 10.13  Distrib 5.7.35, for Linux (x86_64)
+-- MariaDB dump 10.19  Distrib 10.5.10-MariaDB, for Linux (x86_64)
```
agnivade, line 1: Hmm .. why MariaDB though? Does it match with CI?

agnivade (approved):
Looks good, but we should look into why CI acts differently.

scripts/mattermost-mysql-6.0.sql
streamer45: That's just my local client, it's a comment though.

streamer45:
> Looks good, but we should look into why CI acts differently.

Yes, me and @metanerd are having a look.

scripts/mattermost-mysql-6.0.sql
agnivade: Ah nvm, I thought we were matching the final output with this text file. Now I see that we just use this file for importing and then dump it again.

streamer45: Yeah, it's just an input "template" to initialize the migrated db.

streamer45:
By the way, something I've found out while working on this that may be of note is that the schema checking script doesn't really test migrations cause they never get to run. In fact, the schema version (the one saved in the Systems table) is missing on the newly created database so we set it to whatever the server version is and no upgrade executes due to: https://github.com/mattermost/mattermost-server/blob/b6564fec6d62c4f9b3a742748cd2c2364f049646/store/sqlstore/upgrade.go#L126-L135

I am going to assume this is the intended behaviour since it's been  like this for a very long time but I'd also consider adding an additional step that runs the migrations and compares the final output. Of course this may be all unnecessary coming the new system.

metanerd (approved):
Sorry, forgot to approve!

PR #18327, opened 2021-09-01
Fix test schema docker-compose exec exit problem 

metanerd:
#### Summary
 It seems like docker-compose exec behaves differently regarding exit codes compared to docker exec. I guess, I could figure it out at some stage, but instead I can just switch to docker exec instead and it behaves like we expect it to. If approved, I will remove the showcase gitlab ci modification. 

#### Release Note

```release-note

```


streamer45 (approved):
Brilliant :100:

.gitlab-ci.yml
```diff
@@ -129,6 +129,7 @@ test-schema-mysql:
     - k8s-memory
   rules:
     - if: '$CI_COMMIT_REF_NAME == "master"'
+    - if: '$CI_COMMIT_REF_NAME == "fix-test-schema-exit-1"'
```
metanerd, line 132:
```suggestion
```

.gitlab-ci.yml
```diff
@@ -156,5 +157,6 @@ test-schema-postgres:
...
```
metanerd, line 159:
```suggestion
```

PR #18328, opened 2021-09-01
[MM-37757] Fix racy test TestPreparePostForClient

streamer45:
#### Summary

PR fixes a potential race that was causing `TestPreparePostForClient/no_metadata_for_deleted_posts` to fail.
We do this by avoiding passing the same post object that `App.DeletePost()` also returns to the goroutine that deletes post attachments. Since we are there we also unexport some methods and only create the goroutine if it's actually needed.

#### Ticket Link

https://mattermost.atlassian.net/browse/MM-37757

#### Release Note

```release-note
NONE
```


agnivade (approved):
Brilliant stuff

mkraft (approved):

PR #18329, opened 2021-09-02
Use new mattermost-govet

agnivade:
```release-note
NONE
```

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -618,10 +618,7 @@ update-dependencies: ## Uses go get -u to update all the dependencies while hold
 	$(GO) mod tidy
 
 vet: ## Run mattermost go vet specific checks
-	@if ! [ -x "$$(command -v $(GOBIN)/mattermost-govet)" ]; then \
-		echo "mattermost-govet is not installed. Please install it executing \"GO111MODULE=off GOBIN=$(PWD)/bin go get -u github.com/mattermost/mattermost-govet\""; \
-		exit 1; \
-	fi;
+	$(GO) install github.com/mattermost/mattermost-govet/v2@new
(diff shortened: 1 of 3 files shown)


Makefile
```diff
@@ -618,10 +618,7 @@ update-dependencies: ## Uses go get -u to update all the dependencies while hold
 	$(GO) mod tidy
 
 vet: ## Run mattermost go vet specific checks
-	@if ! [ -x "$$(command -v $(GOBIN)/mattermost-govet)" ]; then \
-		echo "mattermost-govet is not installed. Please install it executing \"GO111MODULE=off GOBIN=$(PWD)/bin go get -u github.com/mattermost/mattermost-govet\""; \
-		exit 1; \
-	fi;
+	$(GO) get -modfile=go.tools.mod github.com/mattermost/mattermost-govet/v2@new
```
streamer45, line 621: Was the bump to v2 needed after all?

agnivade: Semantically I'd say so. If it'd work on v5 and v6 both, then I'd be okay, but it takes only a single package path. So effectively, it doesn't work on old codebases.

streamer45 (commented):
It's probably a go 1.17 thing but I am getting a warning running this:

```
go get: installing executables with 'go get' in module mode is deprecated.
	To adjust and download dependencies of the current module, use 'go get -d'.
	To install using requirements of the current module, use 'go install'.
	To install ignoring the current module, use 'go install' with a version,
	like 'go install example.com/cmd@latest'.
	For more information, see https://golang.org/doc/go-get-install-deprecation
	or run 'go help get' or 'go help install'.
```

Wondering if it's something we could easily fix now or better to defer it.

agnivade:
Interesting, it looks like we don't need the go.tools.mod pattern at all. Since `go install` with a version directly ignores the current module. I will update this PR for now, but will send another PR later to fix all other tool installs.

metanerd:
Nice find @streamer45 !

PR #18330, opened 2021-09-02
[MM-36909] Fix flaky TestCreateDefaultMemberships

streamer45:
#### Summary

We ensure `timeAfterLeaving` is actually after the member's `LeaveTime` otherwise the delete won't happen.

#### Ticket Link

https://mattermost.atlassian.net/browse/MM-36909

#### Release Note

```release-note
NONE
```

### DIFF
--- a/app/syncables_test.go
+++ b/app/syncables_test.go
@@ -13,7 +13,6 @@ import (
 )
 
 func TestCreateDefaultMemberships(t *testing.T) {
-	t.Skip("MM-36909")
 	th := Setup(t).InitBasic()
 	defer th.TearDown()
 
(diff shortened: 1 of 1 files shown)


app/syncables_test.go
```diff
@@ -297,14 +296,15 @@ func TestCreateDefaultMemberships(t *testing.T) {
 		t.Errorf("unable to add user to channel: %s", err.Error())
 	}
 
-	timeAfterLeaving := model.GetMillis()
+	timeAfterLeaving := model.GetMillis() + 1
```
agnivade, line 299: Would it make sense to do `timeAfterLeaving := timeBeforeLeaving + 1`? Or some such. Just to make the dependency clear.

streamer45: I don't think that would fix the issue though cause we need to wait for `LeaveChannel` to have executed.

agnivade: Hmm .. but isn't it sufficient to have `timeAfterLeaving` greater than `timeBeforeLeaving`? Anyways, the current way is also fine.

agnivade (approved):

app/syncables_test.go
streamer45: I don't think so because what we end up checking on is the `ChannelMemberHistory.LeaveTime` field is set in `ChannelMemberHistoryStore.LogLeaveEvent()` which is eventually called by `App.LeaveChannel()` so we can't really control what that value is going to be.

agnivade: Got it. Thanks for clearing that up.

mkraft (approved):

PR #18333, opened 2021-09-02
[MM-38321] Fix possible panics during license validation

streamer45:
#### Summary

PR fixes two possible panics that could occur during license validation if the input data was particularly "bad".

1. We were enforcing a minimum length but then proceeded to remove some data (null terminators). Then we assumed the data was long enough. This is what triggered the reported panic.
2. I've also noticed the for loop that took care of removing null bytes didn't have an exit condition so it could panic in case the input data was full of them.

Although this should fix the potential crashes, I am still not sure what could have caused it given the reported panic happened on a license trial request so I am thinking whatever data was returned by `https://customers.mattermost.com/api/v1/trials` triggered this. /cc @marianunez 

#### Ticket

https://mattermost.atlassian.net/browse/MM-38321

#### Release Note

```release-note
Fixed a possible panic during license validation.
```

### DIFF
--- a/app/license.go
+++ b/app/license.go
@@ -6,6 +6,7 @@ package app
 import (
 	"bytes"
 	"encoding/json"
+	"fmt"
 	"net/http"
 	"os"
 	"time"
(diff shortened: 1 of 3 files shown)


jupenur (approved):

isacikgoz (approved):
LGTM 👍

agnivade (approved):
Nice!

marianunez:
> Although this should fix the potential crashes, I am still not sure what could have caused it given the reported panic happened on a license trial request so I am thinking whatever data was returned by `https://customers.mattermost.com/api/v1/trials` triggered this. /cc @marianunez

Thanks @streamer45 for catching this. Would it make sense to also add a check for status code on the actual http request here: https://github.com/mattermost/mattermost-server/blob/22b33c625b1b2bef90a9edf83c2e5548b155f245/app/license.go#L270-L273

This way we can log the issue and won't even need to hit the point of trying to validate the response of something different than a 200

marianunez:
> Sounds like a great idea. Let me add an early return for non 2xx codes.

Thanks @streamer45! LGTM 🚀

amyblais:
If this is a good cherry-pick candidate, it would need manual cherry-picking to ``release-5.39``.

streamer45:
> If this is a good cherry-pick candidate, it would need manual cherry-picking to `release-5.39`.

Yes since it fixes a potential crash it makes sense to include it. I'll take care of that :+1:

PR #18337, opened 2021-09-03
Fix stopping mattermost server

streamer45:
#### Summary

After the `disableconfigwatch` flag [went away](https://github.com/mattermost/mattermost-server/pull/17913/files) we were left with a rather generic grep match that would result into killing more than expected (any unrelated process running through `go run`). We fix this by adding an additional filtering.

#### Release Note

```release-note
NONE
```


agnivade (approved):

isacikgoz (approved):
LGTM 👍

streamer45:
@amyblais I think cherry-pick for this can be skipped. It's not an issue with the application but only with the developers tooling so it's fine to leave it in master.

PR #18338, opened 2021-09-03
Fix trap

metanerd:
#### Release Note

```release-note

```

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -29,10 +29,14 @@ jobs:
           export WEBAPP_GIT_COMMIT=$(git rev-parse HEAD)
           echo "$WEBAPP_GIT_COMMIT"
 
-          trap 'export WEBAPP_GIT_COMMIT=$(git rev-parse HEAD^); echo "previous $WEBAPP_GIT_COMMIT"; curl -f -o ./dist.tar.gz https://pr-builds.mattermost.com/mattermost-webapp/commit/${WEBAPP_GIT_COMMIT}/mattermost-webapp.tar.gz; mkdir ./dist && tar -xvf ./dist.tar.gz -C ./dist --strip-components=1;' ERR
-          curl -f -o ./dist.tar.gz https://pr-builds.mattermost.com/mattermost-webapp/commit/${WEBAPP_GIT_COMMIT}/mattermost-webapp.tar.gz
-          mkdir ./dist && tar -xvf ./dist.tar.gz -C ./dist --strip-components=1
-          trap - ERR
+          FILE_DIST=dist.tar.gz
+          until curl --max-time 5 -f -o $FILE_DIST https://pr-builds.mattermost.com/mattermost-webapp/commit/${WEBAPP_GIT_COMMIT}/mattermost-webapp.tar.gz; do echo waiting for webapp build; sleep 110; done;
+          if [ ! -f $FILE_DIST ]; then
+            npm ci && make build
+          else
+            mkdir dist && tar -xvf $FILE_DIST -C dist --strip-components=1
+          fi
+          cd -
       - persist_to_workspace:
           root: ~/mattermost
           paths:
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -29,7 +29,7 @@ jobs:
           export WEBAPP_GIT_COMMIT=$(git rev-parse HEAD)
           echo "$WEBAPP_GIT_COMMIT"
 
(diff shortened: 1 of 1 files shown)


.circleci/config.yml
```diff
@@ -29,7 +29,7 @@ jobs:
           export WEBAPP_GIT_COMMIT=$(git rev-parse HEAD)
           echo "$WEBAPP_GIT_COMMIT"
 
-          trap 'npm ci && make build' ERR
+          trap 'export WEBAPP_GIT_COMMIT=$(git rev-parse HEAD^); echo "previous $WEBAPP_GIT_COMMIT"; curl -f -o ./dist.tar.gz https://pr-builds.mattermost.com/mattermost-webapp/commit/${WEBAPP_GIT_COMMIT}/mattermost-webapp.tar.gz; mkdir ./dist && tar -xvf ./dist.tar.gz -C ./dist --strip-components=1;' ERR
```
agnivade, line 32: I wonder if this is the right fix. Building with the previous commit will result in incorrect builds where server is a commit x and webapp is at commit x-1.

IMO, I don't think the issue is with `npm ci && make build`. The problem is with the `trap` itself. When the error happens, the trap is executed and `make build` actually happens - as you can see [here](https://app.circleci.com/pipelines/github/mattermost/mattermost-server/27466/workflows/01fbc8ff-aea4-4f6c-ac80-0dc08ac3d0a8/jobs/192415).

But eventually, at the end it errors out with:
```
Exited with code exit status 22
CircleCI received exit code 22
```

Perhaps we should add a loop to check existence of the latest build for 5 times, and if not, then move forward with the manual build.

metanerd: So strange. OK, updated.

agnivade:
It's getting late here, and I am off next week. Looks good overall, but would suggest to ask someone else to take a second look.

.circleci/config.yml
```diff
@@ -29,10 +29,14 @@ jobs:
           export WEBAPP_GIT_COMMIT=$(git rev-parse HEAD)
           echo "$WEBAPP_GIT_COMMIT"
 
-          trap 'export WEBAPP_GIT_COMMIT=$(git rev-parse HEAD^); echo "previous $WEBAPP_GIT_COMMIT"; curl -f -o ./dist.tar.gz https://pr-builds.mattermost.com/mattermost-webapp/commit/${WEBAPP_GIT_COMMIT}/mattermost-webapp.tar.gz; mkdir ./dist && tar -xvf ./dist.tar.gz -C ./dist --strip-components=1;' ERR
-          curl -f -o ./dist.tar.gz https://pr-builds.mattermost.com/mattermost-webapp/commit/${WEBAPP_GIT_COMMIT}/mattermost-webapp.tar.gz
-          mkdir ./dist && tar -xvf ./dist.tar.gz -C ./dist --strip-components=1
-          trap - ERR
+          FILE_DIST=dist.tar.gz
+          until curl --max-time 5 -f -o $FILE_DIST https://pr-builds.mattermost.com/mattermost-webapp/commit/${WEBAPP_GIT_COMMIT}/mattermost-webapp.tar.gz; do echo waiting for webapp build; sleep 110; done;
```
streamer45, line 33: Why do we need to sleep for 110 seconds?

metanerd: https://app.circleci.com/pipelines/github/mattermost/mattermost-webapp/42646/workflows/8e2227e4-bb96-4d9f-bfc5-64e84e321450

I wanted to check 5 times during a full cycle until it is uploaded. But turns out it would be 361seconds until s3 upload. So ~370/5=~75seconds . 
Or nope? Just sleep 5?

streamer45: I am wondering if we could use a timeout approach instead (e.g. https://jeromebelleman.gitlab.io/posts/devops/until/) and attempt multiple times without waiting for  such a long time. We can still add an initial sleep if we know it won't be ready for a while but at least we get to poll and fail/succeed faster afterwards.

metanerd: OK. What about
```suggestion
          until curl --max-time 30 -f -o $FILE_DIST https://pr-builds.mattermost.com/mattermost-webapp/commit/${WEBAPP_GIT_COMMIT}/mattermost-webapp.tar.gz; do echo waiting for webapp build; sleep 5; done;
```
for now?

streamer45: Sure :+1:

streamer45 (approved):

PR #18340, opened 2021-09-03
[WIP][MM-37013] Async job to fix CRT channel unreads

ashishbhate:
#### Release Note

```release-note

```

### DIFF
--- a/jobs/fix_crt_channel_unreads/worker.go
+++ b/jobs/fix_crt_channel_unreads/worker.go
@@ -4,6 +4,7 @@
 package fix_crt_channel_unreads
 
 import (
+	"database/sql"
 	"strconv"
 
 	"github.com/mattermost/mattermost-server/v6/app"
@@ -11,6 +12,10 @@ import (
 	tjobs "github.com/mattermost/mattermost-server/v6/jobs/interfaces"
 	"github.com/mattermost/mattermost-server/v6/model"
 	"github.com/mattermost/mattermost-server/v6/shared/mlog"
+)
+
+const (
+	JobName = "FixCRTChannelUnreads"
 )
 
 type FixCRTChannelUnreadsWorker struct {
@@ -36,7 +41,7 @@ func init() {
 func (i *FixCRTChannelUnreadsJobInterfaceImpl) MakeWorker() model.Worker {
 
 	return &FixCRTChannelUnreadsWorker{
-		name:        "FixCRTChannelUnreads",
+		name:        JobName,
 		stopChan:    make(chan struct{}),
 		stoppedChan: make(chan struct{}),
 		jobsChan:    make(chan model.Job),
@@ -85,20 +90,35 @@ func (w *FixCRTChannelUnreadsWorker) doJob(job *model.Job) {
 	} else if !claimed {
 		return
 	}
+	if _, err := w.app.Srv().Store.System().GetByName(model.MigrationKeyFixCRTChannelUnreads); err == nil {
+		mlog.Info("Worker: migration already done", mlog.String("worker", w.name), mlog.String("job_id", job.Id))
+		w.setJobSuccess(job)
+		return
+	}
 
 	userID := ""
 	channelID := ""
-	fixedInThisRun := 0
+	updatedInThisRun := 0
 	fixedBadCM := 0
+	migrationDone := false
 	for {
-		cms, err := w.app.Srv().Store.Channel().GetCRTUnfixedChannelMembershipsAfter(userID, channelID, 100)
-		if err != nil {
+		cms, sErr := w.app.Srv().Store.Channel().GetCRTUnfixedChannelMembershipsAfter(userID, channelID, 100)
+		if sErr != nil {
+			if sErr == sql.ErrNoRows {
+				migrationDone = true
+				break
+			}
 			mlog.Warn("Failed to get bad channel memberships",
 				mlog.String("worker", w.name),
 				mlog.String("job_id", job.Id),
-				mlog.String("error", err.Error()))
+				mlog.String("error", sErr.Error()))
 			continue
 		}
+		lastCM := cms[len(cms)-1]
+		userID = lastCM.UserId
+		channelID = lastCM.ChannelId
+
+		markAsFixed := make([]model.ChannelMember, 0, len(cms))
 
 		for _, cm := range cms {
 			isUnread, err := w.app.Srv().Store.Channel().IsChannelMemberUnread(cm, true)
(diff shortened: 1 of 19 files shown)


store/sqlstore/upgrade.go
```diff
@@ -1310,7 +1310,34 @@ func upgradeDatabaseToVersion600(sqlStore *SqlStore) {
...
 	sqlStore.AlterColumnTypeIfExists("Users", "NotifyProps", "JSON", "jsonb")
 	sqlStore.AlterColumnTypeIfExists("Users", "Timezone", "JSON", "jsonb")
-
+	fixCRTChannelCountsAndUnreadsUpdateColumn(sqlStore)
 	// saveSchemaVersion(sqlStore, Version600)
 	// }
 }
+
+func fixCRTChannelCountsAndUnreadsUpdateColumn(sqlStore *SqlStore) {
+	created := sqlStore.CreateColumnIfNotExists("ChannelMembers", "CRTFixDone", "tinyint(1)", "boolean", "1")
```
ashishbhate, line 1319: Newly created channel memberships don't have this problem so the default is `true`.

ashishbhate: On my test DB (docker mysql 5.7.12) containing 4.2 million channel members, this takes about 3 minutes.

store/sqlstore/upgrade.go
```diff
@@ -1310,7 +1310,34 @@ func upgradeDatabaseToVersion600(sqlStore *SqlStore) {
...
+	`
+
+	if sqlStore.DriverName() == model.DatabaseDriverMysql {
+		setFalseForUnreadChannels = `
+			UPDATE ChannelMembers
+			INNER JOIN Channels on Channels.Id = ChannelMembers.ChannelId
+			SET CRTFixDone=false
+			WHERE Channels.TotalMsgCountRoot > ChannelMembers.MsgCountRoot;
+		`
+	}
```
ashishbhate, line 1338: We set unread channelmembers as not fixed. Then in the job we check each channelmember and mark as read and/or fixed. On my test DB (docker mysql 5.7.12) containing 4.2 million channel members, this takes about 7.5 minutes, 1.5M of the ChannelMembers were unread.

ashishbhate:
@isacikgoz @streamer45 this PR still needs some polish and work, but its ready for the first round of reviews. Requesting the review early for early feedback.

Feel free to comment if even the slightest thing is unclear. There are some bits that may require CRT specific context. 

Also, please share your thoughts on shipping this in 6.0. Even with the fix happening as a job, its still quite heavy.

ashishbhate:
@streamer45  I'm still trying to figure out why the schema check is failing.  Look at the check and the git history, it seems I should just be skipping the check for the new column addition?

streamer45:
> @streamer45 I'm still trying to figure out why the schema check is failing. Look at the check and the git history, it seems I should just be skipping the check for the new column addition?

If we are adding a column we should also update the related store to make sure it gets created on initial schema creation (new installation case).

streamer45:
Haven't gone into too much detail yet so I'll start from the general approach to this: I am not entirely convinced by adding the `CRTFixDone` column to the schema, especially of a frequently large table like `ChannelMembers` for the sole purpose of a migration. I am wondering if we could think of a different approach that doesn't require that, something cursor based maybe?

ashishbhate:
>I am not entirely convinced by adding the CRTFixDone column to the schema, especially of a frequently large table like ChannelMembers for the sole purpose of a migration. I am wondering if we could think of a different approach that doesn't require that, something cursor based maybe?

It considerably reduces the number of channels to check in the async job by marking read channels as already fixed in the sql upgrade part of the migration. Doing so makes the async job take much less time (because it has fewer channel memberships to process) thus making CRT usable quicker for the end user. The time cost of this is a few minutes added upfront during the sql upgrade.

An additional benefit is that we can stop and restart the job and it will continue from where it left off. Alternatively, we could save progress state in the systems or jobs table, but I thought this way would be easier and cleaner to do. 

Could you elaborate on why you're not convinced that adding a column is the right way to go about this?

streamer45:
> It considerably reduces the number of channels to check in the async job by marking read channels as already fixed in the sql upgrade part of the migration. Doing so makes the async job take much less time (because it has fewer channel memberships to process) thus making CRT usable quicker for the end user. The time cost of this is a few minutes added upfront during the sql upgrade.

But isn't that just a check on `Channels.TotalMsgCountRoot > ChannelMembers.MsgCountRoot`? I am thinking that could be done at runtime instead (during the async job).

> An additional benefit is that we can stop and restart the job and it will continue from where it left off. Alternatively, we could save progress state in the systems or jobs table, but I thought this way would be easier and cleaner to do.

That's what I was thinking. Why not saving the last `ChannelId, UserId` pair that we processed. That should keep track of the progress nicely since we are sorting data by it.

> Could you elaborate on why you're not convinced that adding a column is the right way to go about this?

Because it doesn't look like a clean solution as it's not a temporary thing but we'll be stuck with the added column for a possibly long time. Also it really has nothing to do with the schema/model but it's purely a convenience thing. To that I think I'd personally prefer an entire new table.

ashishbhate:
@streamer45 thanks for expanding on your original comment. I think you've raised some valid concerns. I've updated the PR to not use an added column.

jobs/fix_crt_channel_unreads/worker.go
```diff
@@ -0,0 +1,260 @@
...
+					mlog.String("error", err.Error()))
+				if prevErr {
+					w.updateProgress(job, cm.UserId, cm.ChannelId)
+					w.setJobError(job, model.NewAppError(w.name, "fix_crt_channel_unreads.worker.do_job.get_post_types", nil, err.Error(), http.StatusInternalServerError))
+					return
+				}
+				prevErr = true
+				continue
+			}
+			if !containsNormalPost(postTypes) {
```
streamer45, line 177: I am clearly missing some context. Why do we need to act here if no "normal" posts is found?

jobs/fix_crt_channel_unreads/worker.go
```diff
@@ -0,0 +1,260 @@
...
+		select {
+		case <-w.stopChan:
+			shouldStop = true
+		default:
+			shouldStop = false
+		}
+		if shouldStop {
+			break
+		}
+		cms, sErr := w.app.Srv().Store.Channel().GetCRTUnfixedChannelMembershipsAfter(userID, channelID, 100)
```
streamer45, line 135: You mentioned that on your test environment this job took around 3 minutes. Were you able to monitor the impact of this on DB CPU? I am just wondering if we should artificially limit the throughput to avoid stressing the server too much given this is meant to be run on live systems.

streamer45 (commented):
@ashishbhate 

Thank you for addressing my previous concerns. It's looking great.  
I've left a couple of questions.

jobs/fix_crt_channel_unreads/worker.go
ashishbhate: >you mentioned that on your test environment this job took around 3 minutes
The previous column addition takes three minutes.  This whole job takes a very long time (scales with the number of bad channel memberships). 

>Were you able to monitor the impact of this on DB CPU?

There is indeed a noticeable spike in the DB CPU usage. I haven't noticed any slow-down in my test server usage, but I don't have any solid numbers to measure user-level impact.


>I am just wondering if we should artificially limit the throughput to avoid stressing the server too much given this is meant to be run on live systems.

Yep, I was thinking about this as well. What would be the best way to achieve this? Will a simple sleep to do it? I wasn't sure how we could get some sort of back pressure from the connection pool or DB.

jobs/fix_crt_channel_unreads/worker.go
ashishbhate: When we added the `TotalMsgCountRoot` (to Channels) and `MsgCountRoot` (to channelMembers) columns, system messages were included in the count (they normally aren't). This lead to many channels being marked as unread for users. These channels only had join/leave messages, and no user created posts. 

This migration fixes that issue by marking a channel as read if there are no user posted messages. At the moment this is a bit heavy handed - it ignores all system messages, not just join/leave messages (i.e. containsNormalPost is `true` only if `post.Type == ""` ). I'm looking into refining this further, but the check will be similar.

jobs/fix_crt_channel_unreads/worker.go
streamer45: > There is indeed a noticeable spike in the DB CPU usage. I haven't noticed any slow-down in my test server as a consequence, but I don't have any solid numbers to measure user-level impact.

Okay, @isacikgoz and I can probably assist on this front. It would be good to setup a new test cluster with some good amount of data (e.g. 10M posts) and fake some bad memberships. As a real datapoint, we have around 100k in community db right now.

> Yep, I was thinking about this as well. What would be the best way to achieve this? Will a simple sleep to do it? I wasn't sure how we could get some sort of back pressure from the connection pool or DB.

I think simple sleeping could work but will need some testing to tune it properly. To improve things further I think we should also consider a couple other things:

- Amount of returned memberships. It's currently set to 100 but increasing this a bit could help minimizing read operations. Need to be tested of course.
- Batching write operations. `ChannelStore.UpdateLastViewedAt()` is one of the most complex store calls. We should first consider if we actually need it as it does a lot more than what it says. If the answer is yes, meaning we need all the logic attached to it then I'd suggest trying to call it the least possible number of times by passing multiple channel ids at once.

isacikgoz (approved):
LGTM, we need to communicate with the SRE team and customers that this migration could take some time to complete. The performance impact seems to be minimal though. Great work on this Ashish! And thanks for your patience.

PR #18361, opened 2021-09-09
MM-37721: Add prop for post preview before saving record.

mkraft:
#### Summary

Adds `previewed_post` prop _before_ saving the record so that two records aren't created for a new post.

#### Ticket Link

https://mattermost.atlassian.net/browse/MM-37721

#### Release Note

```release-note
NONE
```


BenCookie95:
Is there a unit test we can write for this fix?

mkraft:
> Is there a unit test we can write for this fix?

Yes, I've added one. Good call.

BenCookie95 (approved):

engineereng (approved):
Looks good to me!

mkraft:
@furqanmlk I think you already tested this? The only change I made since your testing is I added a unit test.

furqanmlk (approved):
Working as expected 

image

image

PR #18363, opened 2021-09-10
MM-37896: thread recency when updating a post

koox00:
#### Summary

When a post is edited we should not update the thread recency
(LastReplyAt).

#### Ticket Link

https://mattermost.atlassian.net/browse/MM-37896

#### Release Note

```release-note
NONE
```


ashishbhate (approved):
LGTM. We should add a test. It will ensure this behaviour won't change in the future without us finding out about it.

jwilander (approved):

PR #18366, opened 2021-09-10
MM-36743: adds last_root_post_at in channels table

koox00:
#### Summary

Channel recency for CRT users should not count replies,
this commit solves that issue by adding a new column to the channels
table: LastRootPostAt.
With that new info CRT users can have recent channels to work as
expected.

#### Ticket Link

https://mattermost.atlassian.net/browse/MM-36743

#### Release Note

```release-note
NONE
```

### DIFF
--- a/store/sqlstore/upgrade.go
+++ b/store/sqlstore/upgrade.go
@@ -1350,6 +1350,43 @@ func upgradeDatabaseToVersion600(sqlStore *SqlStore) {
 		sqlStore.AlterColumnTypeIfExists("Posts", "FileIds", "text", "varchar(300)")
 	}
 
+	forceIndex := ""
+	if sqlStore.DriverName() == model.DatabaseDriverMysql {
+		forceIndex = "FORCE INDEX(idx_posts_channel_id)"
+	}
+
+	lastRootPostAtExists := sqlStore.DoesColumnExist("Channels", "LastRootPostAt")
+	sqlStore.CreateColumnIfNotExists("Channels", "LastRootPostAt", "bigint", "bigint", "0")
+
+	lastRootPostAtCTE := `
+		SELECT Channels.Id channelid, COALESCE(MAX(Posts.CreateAt), 0) as lastrootpost
+		FROM Channels
+		LEFT JOIN Posts ` + forceIndex + ` ON Channels.Id = Posts.ChannelId
+		WHERE Posts.RootId = ''
+		GROUP BY Channels.Id
+	`
+
+	updateChannels := `
+		WITH q AS (` + lastRootPostAtCTE + `)
+		UPDATE Channels SET LastRootPostAt=q.lastrootpost
+		FROM q where q.channelid=Channels.Id;
+	`
+
+	if sqlStore.DriverName() == model.DatabaseDriverMysql {
+		updateChannels = `
+			UPDATE Channels
+			INNER Join (` + lastRootPostAtCTE + `) as q
+			ON q.channelid=Channels.Id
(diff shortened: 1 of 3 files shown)


koox00:
Hi @streamer45 I need to add a new column (LastRootPostAt) to the channels table,
do I need to add the new column at [here](https://github.com/mattermost/mattermost-server/blob/9534f9fb62f4c2e43d9ee0568aba94d2c33d065b/scripts/mattermost-postgresql-6.0.sql) as well?

streamer45:
> Hi @streamer45 I need to add a new column (LastRootPostAt) to the channels table,
> do I need to add the new column at [here](https://github.com/mattermost/mattermost-server/blob/9534f9fb62f4c2e43d9ee0568aba94d2c33d065b/scripts/mattermost-postgresql-6.0.sql) as well?

Yes, and the `mysql` one as well.

You also need to add the new column mapping in:

https://github.com/mattermost/mattermost-server/blob/10d81d09f393cfe30ef9ff63f9afdd093f778dd9/store/sqlstore/channel_store.go#L388

koox00:
> You also need to add the new column mapping in:
> 
> https://github.com/mattermost/mattermost-server/blob/10d81d09f393cfe30ef9ff63f9afdd093f778dd9/store/sqlstore/channel_store.go#L388

`LastPostAt` is not added, should I add it as well? (This PR adds `LastRootPostAt`).

streamer45:
> > You also need to add the new column mapping in:
> > https://github.com/mattermost/mattermost-server/blob/10d81d09f393cfe30ef9ff63f9afdd093f778dd9/store/sqlstore/channel_store.go#L388
> 
> `LastPostAt` is not added, should I add it as well? (This PR adds `LastRootPostAt`).

Apologies, just realized you're adding a integer. No need to change that file then. Just make sure to update the sql schema files.

store/sqlstore/upgrade.go
```diff
@@ -1350,6 +1350,43 @@ func upgradeDatabaseToVersion600(sqlStore *SqlStore) {
 		sqlStore.AlterColumnTypeIfExists("Posts", "FileIds", "text", "varchar(300)")
 	}
 
+	forceIndex := ""
```
streamer45, line 1353: The other thing is that I am guessing this is meant to go in 6.1 since we are past code complete so we should put the migration in a new `upgradeDatabaseToVersion610` function. You can check https://github.com/mattermost/mattermost-server/pull/18390/files#diff-c7561d15f1d9c1a92942b9ce5d2b1d979c89ffd525a8975f190983220cd6f2af for an example.

store/sqlstore/upgrade.go
```diff
@@ -1353,3 +1355,46 @@ func upgradeDatabaseToVersion600(sqlStore *SqlStore) {
 	// saveSchemaVersion(sqlStore, Version600)
 	// }
 }
+
+func upgradeDatabaseToVersion610(sqlStore *SqlStore) {
+	// if shouldPerformUpgrade(sqlStore, Version600, Version610) {
+
+	forceIndex := ""
+	if sqlStore.DriverName() == model.DatabaseDriverMysql {
+		forceIndex = "FORCE INDEX(idx_posts_channel_id)"
```
streamer45, line 1364: Starting in 5.37 we don't have this index any longer as it was redundant. We should use `idx_posts_channel_id_update_at` for now.

streamer45 (approved):
Looks good :+1:  Will review this further from a migration performance perspective during 6.1 analysis.

ashishbhate (approved):

PR #18368, opened 2021-09-11
MM-38342 Fix fields output when logging for plugins

wiggin77:
#### Summary
This PR fixes an issue where context fields logged by server on behalf of plugins didn't contain correct name/value pairs.

* bump Logr version to add key/value methods to sugar logger
* expose factory args when configuring logging with custom target types (needed for FocalBoard to create log target adapter that converts typed fields into slices of interface{} as per plugin logging API)

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-38342

#### Release Note
```release-note
NONE
```


agnivade:
Just got back from vacation, and probably a bit too late since the logr [PR](https://github.com/mattermost/logr/commit/b544af1828eeab671734955b530d663085b01aad) is already merged, but I am wondering whether it would be possible to do this without introducing any new method.

Could we have done a type assertion for `Field`  and then took the old path, and if the param was not a field, then used the new path?

Anyways, this is also okay, but the method name could be slightly improved. Maybe `...KV`? 0/5.

wiggin77:
> Could we have done a type assertion for `Field` and then took the old path

Thanks for having a look @agnivade.  If I'm understanding your suggestion correctly, the issue is that the logging method call via RPC is done with a variadic set of `interface{}` and the underlying type is lost in the RPC call. 

On a related note, this coming week I was going to ask whoever was leading the plugin API development if I could add a method to the plugin API that accepts a `logr.LogRecord` (or mlog type alias) and bypass all the KV conversion and gain proper caller info, which is lost in the RPC calls currently.  This would also provide the option of using the same mlog APIs server uses within the plugin.  I was planning this improvement as a separate PR.

As for naming the APIs `DebugKV` , etc,  I agree at makes sense.  I named them like `Debugw` in keeping with how they are named in Zap.

agnivade:
> Thanks for having a look @agnivade. If I'm understanding your suggestion correctly, the issue is that the logging method call via RPC is done with a variadic set of interface{} and the underlying type is lost in the RPC call.

Hmm .. that's interesting. IIRC, when I was writing the DB RPC API, that too worked with `...interface{}` but it was able to type assert the underlying types.

> On a related note, this coming week I was going to ask whoever was leading the plugin API development if I could add a method to the plugin API that accepts a logr.LogRecord (or mlog type alias) and bypass all the KV conversion and gain proper caller info, which is lost in the RPC calls currently. This would also provide the option of using the same mlog APIs server uses within the plugin. I was planning this improvement as a separate PR.

Yeah, that would be great.

agnivade (approved):
Looking forward to the future improvements!

levb (approved):

PR #18370, opened 2021-09-13
[MM-25018] Bump Post.Props size limits

streamer45:
#### Summary

Bumping the limit now that the column datatype supports it. I've simply multiplied existing values by 100 to give plenty of new space but still have some sort of protection against post bombs. Let me know if you had other numbers in mind.

#### Ticket

https://mattermost.atlassian.net/browse/MM-25018

#### Release Notes
```release-note
Increased Post.Props size limit to 800,000 characters.
```


agnivade (approved):

hanzei (approved):
That should be enough :+1: Thanks, Claudio!

streamer45:
@amyblais

This would be a good 6.0 cherry-pick candidate as we have everything else related already in. This is totally safe of course as it doesn't affect existing functionality. Without this in customers would have to wait until 6.1 to make use of the extended limits.

PR #18380, opened 2021-09-14
Only consider flagged posts that are on channels you are member of

larkox:
#### Summary
Previously, we used to send all flagged posts. This created some weird problems with Sysadmins, which had the permissions to see the channels, but they were not member of them (which implied certain values not being available in redux).

To solve this, we are only returning the flagged posts for channels where the user is a member of.

We also have refactored the getFlaggedPosts functions to unify the query string, in a similar fashion to what is being done in the search function.

#### Ticket Link
None

#### Release Note
```release-note
Get flagged posts endpoint will return now only flagged posts for channels the user is member of.
```

### DIFF
--- a/store/storetest/post_store.go
+++ b/store/storetest/post_store.go
@@ -1961,6 +1961,13 @@ func testPostStoreGetFlaggedPostsForTeam(t *testing.T, ss store.Store, s SqlStor
 	require.NoError(t, err)
 	time.Sleep(2 * time.Millisecond)
 
+	m0 := &model.ChannelMember{}
+	m0.ChannelId = c1.Id
+	m0.UserId = o1.UserId
+	m0.NotifyProps = model.GetDefaultChannelNotifyProps()
+	_, err = ss.Channel().SaveMember(m0)
+	require.NoError(t, err)
+
 	o4 := &model.Post{}
 	o4.ChannelId = model.NewId()
 	o4.UserId = model.NewId()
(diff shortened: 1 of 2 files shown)


larkox:
@agnivade Not sure who else should I add as a reviewer. Do you mind adding whoever you find fit to review this?

agnivade:
Added @isacikgoz .

store/storetest/post_store.go
```diff
@@ -1995,6 +2002,15 @@ func testPostStoreGetFlaggedPostsForTeam(t *testing.T, ss store.Store, s SqlStor
...
 	time.Sleep(2 * time.Millisecond)
 
+	// Post on channel where user is not a member
+	o6 := &model.Post{}
+	o6.ChannelId = model.NewId()
+	o6.UserId = m2.UserId
+	o6.Message = NewTestId()
+	o6, err = ss.Post().Save(o6)
+	require.NoError(t, err)
+	time.Sleep(2 * time.Millisecond)
```
agnivade, line 2012: I realize these lines were already there from before. But do you know of the reason? It's not good to be indefinitely adding sleeps.

agnivade (approved):
Looks good! Nothing blocking.

store/storetest/post_store.go
larkox: My wild guess is to deal with read replica situations, but it is just a guess.

agnivade: There are no read replicas in tests. I'd be willing to remove the sleep and then see what happens, and fix that. But anyways, outside the scope of this PR.

PR #18383, opened 2021-09-14
Set BuildDate compile flag to dev during development

jespino:
#### Summary
The `BuildDate` flag was set at compile time as the current datetime, what makes the go compiler cache useless for the model package.

This change should save some compilation time.

#### Release Note
```release-note
NONE
```

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -24,6 +24,7 @@ BUILD_DATE = $(shell date -u)
 BUILD_HASH = $(shell git rev-parse HEAD)
 # If we don't set the build number it defaults to dev
 ifeq ($(BUILD_NUMBER),)
+	BUILD_DATE := dev
 	BUILD_NUMBER := dev
 endif
 BUILD_ENTERPRISE_DIR ?= ../enterprise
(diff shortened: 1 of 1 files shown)


Makefile
```diff
@@ -24,6 +24,7 @@ BUILD_DATE = $(shell date -u)
 BUILD_HASH = $(shell git rev-parse HEAD)
 # If we don't set the build number it defaults to dev
 ifeq ($(BUILD_NUMBER),)
+	BUILD_DATE := dev
```
agnivade, line 27: 0/5, I think `n/a` feels better. `dev` does not feel like a date. :P

streamer45: Clearly not a blocker but do we need to define this in the first place? Could it just be left blank in this case?

streamer45 (approved):

agnivade (approved):

Makefile
jespino: I think it is not needed, but it is interesting to have it exposed like "Hey, there is this thing, it doesn't apply on develoment environment, but it is there".

streamer45: Fair enough :+1:

agnivade:
Hey @jespino - I understand this was a small change, but let's always try to ensure that CI always passes before merging a change. :slightly_smiling_face:  Unless it's an urgent customer issue, it's always good to have full CI pass. If there are CI issues, then build team is always there to help us.

PR #18385, opened 2021-09-15
[MM-36529] - Add the config option for EnableReadReceipts

nevyangelova:
#### Summary
This PR adds the new config key for EnableReadReceipts as part of https://mattermost.atlassian.net/browse/MM-36529 so that the option can be turned off on admin level 

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-36529

#### Release Note

```release-note
Added a new config setting TeamSettings.EnableReadReceipts. Added telemetry. 
```

### DIFF
--- a/model/config.go
+++ b/model/config.go
@@ -1872,6 +1872,7 @@ type TeamSettings struct {
 	EnableUserDeactivation              *bool    `access:"experimental_features"`
 	RestrictCreationToDomains           *string  `access:"authentication_signup"` // telemetry: none
 	EnableCustomUserStatuses            *bool    `access:"site_users_and_teams"`
+	EnableReadReceipts                  *bool    `access:"site_users_and_teams"`
 	EnableCustomBrand                   *bool    `access:"site_customization"`
 	CustomBrandText                     *string  `access:"site_customization"`
 	CustomDescriptionText               *string  `access:"site_customization"`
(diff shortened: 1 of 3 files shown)


model/config.go
```diff
@@ -1914,6 +1915,10 @@ func (s *TeamSettings) SetDefaults() {
 		s.EnableCustomUserStatuses = NewBool(true)
 	}
 
+	if s.EnableReadReceipts == nil {
+		s.EnableReadReceipts = NewBool(true)
```
larkox, line 1919: Do we want to enable this by default?

larkox (approved):
LGTM, just one small comment.

model/config.go
nevyangelova: yes per ticket description

hmhealey (approved):
0/5 but perhaps it would make sense to start with this behind a feature flag before enabling it for everyone. I know it's not behind the feature flag in the spec but since our biggest reason against this sort of feature in the past was performance impacts, it might be good to roll it out more slowly.

nevyangelova:
> 0/5 but perhaps it would make sense to start with this behind a feature flag before enabling it for everyone. I know it's not behind the feature flag in the spec but since our biggest reason against this sort of feature in the past was performance impacts, it might be good to roll it out more slowly.

@esethna can advise but since we wouldn't be adding this for channels, only DM's to start with I think it wouldn't affect the performance at this point

esethna:
@hmhealey @nevyangelova I agree this should be behind a feature flag. It actually doesn't matter if we're concerned about perf, feature flag guidelines are that any new feature that changes UI should be behind a feature flag: https://developers.mattermost.com/contribute/server/feature-flags/#when-to-use

PR #18391, opened 2021-09-15
Making private unnecessary exposed App methods for oauth

jespino:
#### Summary
Making private unnecessary exposed App methods for oauth

#### Release Note
```release-note
NONE
```


jespino:
This is solved in the #18395

PR #18392, opened 2021-09-15
MM-38122 Update to use new plugin ID for playbooks.

crspeller:
#### Summary
Updates instances of the playbook plugin ID. 
Should be merged after https://github.com/mattermost/mattermost-plugin-playbooks/pull/740 

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-38122

#### Release Note
```release-note
NONE
```


lieut-data (approved):
Should we pre-package a version with the `playbooks` id here too before merging?

cpoile (approved):

crspeller:
@lieut-data It's not necessary to wait to merge I don't think since the in-between versions only effect our servers. I left in telemetry in for both.

PR #18400, opened 2021-09-16
MM-38081: Fix for disappearing permalink previews

mkraft:
#### Summary

Fixes disappearing permalink previews when the feature is toggled on/off by the config setting or feature flag.

#### Ticket Link

https://mattermost.atlassian.net/browse/MM-38081

#### Release Note

No release note because feature was not yet released.

```release-note
NONE
```


api4/system_test.go
```diff
@@ -775,7 +775,6 @@ func TestPushNotificationAck(t *testing.T) {
 
 		handler.ServeHTTP(resp, req)
 		assert.Equal(t, http.StatusForbidden, resp.Code)
-		fmt.Printf("DEBUG/resp.Body: %+v\n", resp.Body)
```
mkraft, line 778: Unrelated to PR.

hmhealey (approved):
Code looks good.

If I understand correctly, this makes it so that the permalink previews remain visible even if the setting is disabled as long as it was enabled when the link was posted?

mkraft:
> Code looks good.
> 
> If I understand correctly, this makes it so that the permalink previews remain visible even if the setting is disabled as long as it was enabled when the link was posted?

Exactly right. 

Later if we think it's worth it we can add a migration that goes through all posts and adds the `previewed_post` prop to each post that looks like a permalink. That would enable permalink previews for all older posts.

BenCookie95 (approved):

furqanmlk:
Hi @mkraft 

Initially all steps passed, 

image


but when I posted the Post with Website, it went on `loading```` page.

image


You can have a look 

https://mattermost-server-pr-18400.test.mattermost.cloud/pr18400/channels/town-square

furqanmlk:
Thanks @mkraft 
Everything working as expected, except the cache issue that we discussed over zoom
In summary , toggling the feature does not apply until we post a new message. This is happening for other feature too.
I believe, that is existing issue.

furqanmlk (approved):

PR #18402, opened 2021-09-16
[MM-25966] - WIP - Remove restrictions for adding members to group message

nevyangelova:
#### Summary
As part of https://mattermost.atlassian.net/browse/MM-25966 we need to remove some restrictions in order to be able to add 
users to group messages as well as leave the group message. 

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-25966

#### Release Note
-->
```release-note
NONE
```

### DIFF
--- a/api4/channel.go
+++ b/api4/channel.go
@@ -1517,7 +1517,7 @@ func addChannelMember(c *Context, w http.ResponseWriter, r *http.Request) {
 	defer c.LogAuditRec(auditRec)
 	auditRec.AddMeta("channel", channel)
 
-	if channel.Type == model.ChannelTypeDirect || channel.Type == model.ChannelTypeGroup {
+	if channel.Type == model.ChannelTypeDirect {
 		c.Err = model.NewAppError("addUserToChannel", "api.channel.add_user_to_channel.type.app_error", nil, "", http.StatusBadRequest)
 		return
 	}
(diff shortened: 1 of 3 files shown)


hmhealey:
Yeah, I think this needs a review from at least one of the feature teams since, but of the feature, not just of the code. There's a lot of cases I'd be worried about here like:
1. What if you have 3 people in a GM and one leaves? Do you end up with a GM with only 2 people in it which is essentially a weird DM?
2. What if you have two GMs, one with 3 users and another with those same 3 plus a fourth, and you add the fourth user to the first GM? Do you end up with two GMs that are identical?
3. There's a few things around GMs that assume the members of the GM are in its name which is hardcoded. I don't know if changing the users will cause problems there since you can already get the GM to have the "wrong" name by having a user change their name, but it's still a potential concern in things like the quick switcher or search.

Now that I've read more of the ticket, this seems like it's also a huge change in how GMs are treated. GMs and DMs are currently similar in that: users can't be added, users can't leave, and you can only have a single GM with a given group of people. All of those are being changed by this, and that's likely going to have a lot of wide reaching effects on other features. This feels opposed to other tickets that make GMs work more like DMs (like the one to make all messages in GMs be treated as mentions) since they're both pulling GMs in opposite ways.

nevyangelova:
Really good points @hmhealey I can answer the following:
1. I figured this will be a 'weird' GM at this point but still curious about @esethna 's opinion.
2. At the moment the idea is to not duplicate the GM's, whenever you end up with the same people in a GM then it should redirect you to that message.
3. From what I saw when I was testing this the GM name gets updated when you add or remove users. Didn't get to detailed testing on the quick switcher though.

I guess the goal of this ticket is to move GMs closer to a private channel and the only difference will be the limit of people you can add. However if there are other tickets pulling in the opposite direction then that might call for a bigger discussion.

nevyangelova:
Closing this in favour of a more detailed approach

PR #18411, opened 2021-09-17
[GH-15744] Remove direct token store access in api4/user module

dizlv:
#### Summary
Removed last direct `Store` access from `api4/user.go` module to the `app` method.

#### Ticket Link
Fixes https://github.com/mattermost/mattermost-server/issues/15744

#### Release Note
* Added new `App` interface `GetCreateUserToken`
* Added `GetCreateUserToken` Interface implementation
* Replaced direct `Store.Token()` access in `api4/user.go` module with new `GetCreateUserToken` method

#### Release Note
```release-note
NONE
```


api4/user.go
```diff
@@ -123,16 +123,9 @@ func createUser(c *Context, w http.ResponseWriter, r *http.Request) {
...
-		if nErr != nil {
-			var status int
-			switch nErr.(type) {
-			case *store.ErrNotFound:
-				status = http.StatusNotFound
-			default:
-				status = http.StatusInternalServerError
-			}
-			c.Err = model.NewAppError("CreateUserWithToken", "api.user.create_user.signup_link_invalid.app_error", nil, nErr.Error(), status)
+		token, appErr := c.App.GetCreateUserToken(tokenId)
```
agnivade, line 126: This looks like it is just getting the token. Let's name it `GetTokenByUserId`

dizlv: Hey @agnivade, I'm bit out of domain knowledge, shouldn't we better name it `GetTokenByTokenId` then?

agnivade: My bad there. It's not the userID. `GetTokenById` should be fine. Thanks @dizlv.

agnivade (approved):

PR #18417, opened 2021-09-17
MM-38611 getSharedChannels: only return channels user is member of

wiggin77:
#### Summary
getSharedChannels rest API should only return channels the user is a member of, unless the user has "manage_shared_channels" permission.

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-38611

#### Release Note
```release-note
NONE
```

### DIFF
--- a/store/sqlstore/shared_channel_store.go
+++ b/store/sqlstore/shared_channel_store.go
@@ -222,10 +222,7 @@ func (s SqlSharedChannelStore) getSharedChannelsQuery(opts model.SharedChannelFi
 	}
 
 	if opts.TeamId != "" {
-		query = query.Where(sq.Or{
-			sq.Eq{"sc.TeamId": opts.TeamId},
-			sq.Eq{"sc.TeamId": ""},
-		})
+		query = query.Where(sq.Eq{"sc.TeamId": opts.TeamId})
 	}
 
 	if opts.CreatorId != "" {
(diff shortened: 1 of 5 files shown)


jupenur (approved):

store/sqlstore/shared_channel_store.go
```diff
@@ -216,8 +216,16 @@ func (s SqlSharedChannelStore) getSharedChannelsQuery(opts model.SharedChannelFi
...
+	if opts.MemberId != "" {
+		query = query.Join("ChannelMembers AS cm ON cm.ChannelId = sc.ChannelId").
+			Where(sq.Eq{"cm.UserId": opts.MemberId})
+	}
+
 	if opts.TeamId != "" {
-		query = query.Where(sq.Eq{"sc.TeamId": opts.TeamId})
+		query = query.Where(sq.Or{
+			sq.Eq{"sc.TeamId": opts.TeamId},
+			sq.Eq{"sc.TeamId": ""},
```
streamer45, line 227: Why was the empty TeamId condition added here? Asking because on first looks it doesn't seem directly related to the issue the PR is addressing.

wiggin77: While I was looking at how best to check membership in the sqlstore, I saw this code that lead me to believe teamId="" is a valid case:
https://github.com/mattermost/mattermost-server/blob/8a7af3bef29d7acedc31851794d406bf46371b0e/store/sqlstore/channel_store.go#L960

I agree it is unrelated.

streamer45: DMs/GMs have empty team ids. But here we are inside a check on `opts.TeamId != ""` so it's a bit weird to then include empty fields. Anyway, not a blocker just something I've noticed.

streamer45 (approved):

store/sqlstore/shared_channel_store.go
wiggin77: Good point.  I'll remove it.

PR #18418, opened 2021-09-17
[MM-24461] Show last active time of users when not online

BenCookie95:
#### Summary
Show when a user was last active

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-24461

#### Related Pull Requests

- https://github.com/mattermost/mattermost-webapp/pull/8889

#### Release Note

```release-note
Show when a user was last online
```


engineereng (approved):
Looks good to me!

mkraft (commented):
Other user preferences go in the `Preferences` table. Does it make sense to put this one there too?

BenCookie95:
> Other user preferences go in the `Preferences` table. Does it make sense to put this one there too?

@mkraft Other users need to see this preference in order to determine to display last active status or not. From what I saw, a user cannot view another users preferences

BenCookie95:
> I think we've started putting those settings that other people need (like the user's timezone and custom status) in the `User.Props` to avoid adding new columns for everything. Not to keep pulling everyone else into this, but Agniva or someone on Server Platform might have an opinion on that

I'll take a look at that, I feel like a new column is overkill right now

BenCookie95:
> @BenCookie95 how's this going in OSF? Keen to get this merged, so let me know if I can help unblock things at all.

@esethna This is no longer needed, only need webapp code here https://github.com/mattermost/mattermost-webapp/pull/8889

PR #18438, opened 2021-09-22
MM-38712: Flush logs while exiting from store upgrade

agnivade:
Our store upgrade methods used mlog.Critical and then slept
for a second and triggered os.Exit.

First of all, with the new mlog refactor Critical wasn't a standard
log level so it was never to be logged.

And secondly, this completely missed the fact that there already
existed an API method Fatal which did exactly this.

We mark the Critical API as deprecated, and use Fatal for all
occurences.

The choice should be between either to use Error or Fatal.

While here, we also remove the non-standard exit codes which
just confused things further while debugging.

https://mattermost.atlassian.net/browse/MM-38712

```release-note
NONE
```

### DIFF
--- a/shared/mlog/global.go
+++ b/shared/mlog/global.go
@@ -112,6 +112,7 @@ func Error(msg string, fields ...Field) {
 }
 
 // Convenience method equivalent to calling `Log` with the `Critical` level.
+// DEPRECATED: Either use Error or Fatal.
 func Critical(msg string, fields ...Field) {
 	logger := getGlobalLogger()
 	if logger == nil {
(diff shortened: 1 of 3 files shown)


agnivade:
cc @amyblais for cherry-pick to v6.0.

streamer45 (approved):
Great finding and thanks for getting rid of those exit codes as I literally had a moment of embarrassment when faced with `116 ESTALE Stale NFS file handle` the other day.

wiggin77 (approved):
LGTM :+1:

PR #18440, opened 2021-09-22
MM-38657 - Sentry crash: Fix nil dereference when post not found

agnivade:
We were ignoring post not found errors, but moving forward nevertheless. This led to nil dereference panics.

To fix this, we return for any error but exclude from logging the error if it was a not found error.

And we also add a unit test and revert some of the old code.

https://mattermost.atlassian.net/browse/MM-38657

```release-note
NONE
```


mkraft (approved):

mickmister (approved):

streamer45 (approved):

amyblais:
@agnivade Is this needed for v6.0?

agnivade:
This is an improvement over @mickmister's fix which is already backported. No need to make more backports.

PR #18443, opened 2021-09-22
Add OAuth Scopes

larkox:
#### Summary
Approach to OAuth Scopes as defined in https://docs.google.com/document/d/1W1uOVmPIcCg2QkoIrFY1FCs3h5wNcsthljRmggox3JI/edit#

Right now we are only adding a small set of scopes.

Summary of changes (order by appearance in the diff):
- Updated all relevant API endpoints to deny any scoped session.
- Added the relevant scopes to the API endpoints we want open to OAuth Apps
- As example, handle the "getClientConfig" open API endpoint based on scopes. Similar approach may have to be taken for other open API endpoints (all have been marked by a "TODO OAUTH" comment).
- Updated the api handlers to consider OAuth input and set the web handler scope
- Updated the OAuth creation and update endpoints at API level to normalize the scopes (remove duplicates).
- Updated the OAuth update endpoints at APP level to revoke sessions when the new scopes are not included in the old scopes.
- On OAuth endpoints at APP level, pass the whole app instead of only the name to the session creation functions.
- Add to the OAuth sessions the OAuth App ID and the Apps Framework App ID
- Scope plugin requests to scopes like "plugin:plugin_id"
- Add the Scopes to the request context
- Add App and store endpoints to revoke all sessions for an OAuth App
- Update the model of OAuth Apps to allow Scopes and AppsFrameworkAppsIDs
- Add Scope validation to OAuth app validation
- Add the Scope model
- Update the web handler to filter requests based on scopes

Missing pieces
- Add endpoints to revoke all sessions for an Apps Framework App (can be deferred to a different PR)
- Consider how should the Apps Framework create new OAuth apps to avoid the creation of a "fake Apps Framework OAuth App" by using the API (can be deferred to a different PR)
- Add OAuth Store cache
- Testing

#### Release Note
```release-note
OAuth Apps can now be scoped to the desired functionality.
```


web/handlers.go
```diff
@@ -343,6 +352,34 @@ func (h Handler) ServeHTTP(w http.ResponseWriter, r *http.Request) {
 	}
 }
 
+func (h *Handler) checkScopes(c *Context, appID string) model.Scope {
+	oAuthApp, err := c.App.GetOAuthApp(appID)
```
catalintomai, line 356: @larkox - is it always guaranteed that `oAuthApp` is non-nil (or, in general, any obj. that comes from the DB), or is it always the case that `oAuthApp` nil means that `err` is non-nil? We've seen issues like : https://github.com/mattermost/mattermost-server/pull/18437, trying to prevent similar issue here.

model/oauth.go
```diff
@@ -18,17 +18,19 @@ const (
...
+	CreateAt           int64       `json:"create_at"`
+	UpdateAt           int64       `json:"update_at"`
+	ClientSecret       string      `json:"client_secret"`
+	Name               string      `json:"name"`
+	Description        string      `json:"description"`
+	IconURL            string      `json:"icon_url"`
+	CallbackUrls       StringArray `json:"callback_urls"`
+	Homepage           string      `json:"homepage"`
+	IsTrusted          bool        `json:"is_trusted"`
+	Scopes             Scope       `json:"scopes"`
```
agnivade, line 32: We should get the terminology consistent. If we are referring to the slice of strings as a singular scope, then the field name should also be `Scope`. But it seems like each permission attribute is already called a scope. So then we should just use `Scopes []Scope`.

model/scope.go
```diff
@@ -0,0 +1,168 @@
+package model
+
+import (
+	"encoding/json"
+	"strings"
+)
+
+type Scope []string
```
agnivade, line 8: It would be better to have typed strings as a scope. And then have a slice of scopes.

```
type Scope string
const (
	ScopeChannelJoin   Scope = "channels:join"
	ScopeMessageRead Scope = "message:read"
)
```

And then use either `Scope` for a single scope, or `[]Scope` for multiple scopes. This also gives you extra type safety against random strings getting into scopes.

model/scope.go
```diff
@@ -0,0 +1,168 @@
...
+	return false
+}
+
+func (s Scope) IsPluginInScope(pluginID string) bool {
+	if s == nil {
+		return false
+	}
+
+	for _, allowed := range s {
+		if strings.HasPrefix(allowed, "plugin:") &&
```
agnivade, line 51: `HasPrefix` is redundant here. If prefix isn't there, the `TrimPrefix` comparison would anyways fail.

model/scope.go
```diff
@@ -0,0 +1,168 @@
...
+	return out
+}
+
+func (s Scope) ToJSON() string {
+	bytes, err := json.Marshal(s)
+	if err != nil {
+		return ""
+	}
+	return string(bytes)
+}
```
agnivade, line 85: I keep postponing the announcement, but we have got rid of all `ToJSON` methods in the codebase. Just call json.Marshal directly and _always_ handle the error.

model/scope.go
```diff
@@ -0,0 +1,168 @@
...
+
+func ScopeFromJSON(in string) (Scope, error) {
```
agnivade, line 87: Same as earlier, just call `json.Unmarshal` directly.

model/scope.go
```diff
@@ -0,0 +1,168 @@
...
+OUTER:
+	for _, inScope := range s {
+		for _, outScope := range out {
+			if inScope == outScope {
+				continue OUTER
+			}
+		}
+
+		out = append(out, inScope)
+	}
```
agnivade, line 165: can you explain what is it that you are trying to achieve here?

agnivade (commented):
Given that `model.ScopeDeny()` is the most common scope, I am wondering if we should create separate API helpers to reduce the amount of code changes.

Something like:
`APISessionRequiredWithDenyScope(handler)`
`APISessionRequiredWithCustomScope(handler, scope)`

model/scope.go
larkox: Not necessarily. Imagine the scope "message:read". It doesn't have the prefix, and trimPrefix will return the string untouched, like "message:read". Now imagine that for whatever reason, someone decided to have its plugin id as "message:read". Now the plugin would be considered allowed.

It is uncommon and not very exploitable? Yes. But still, I prefer to cover that case.

model/scope.go
larkox: Remove repetitions. If there is any scope that we already added to out, we ignore it.

larkox: I will add some comments to explain that.

web/handlers.go
larkox: In theory, if there is a entry in the database, it will return the entry (with all the information), and if not, it will return a "not found" error. That being said, that is also the case for what is being solved in the other PR.

It will probably not hurt to add the check. @agnivade any thought on this?

larkox:
@agnivade Addressed the feedback and added tests for the scope file (and some fixes).

model/scope.go
```diff
@@ -0,0 +1,208 @@
+package model
+
+import (
+	"strings"
+)
+
+type Scopes []Scope
```
agnivade, line 7: Nit: this feels unnecessary. We can just use `[]Scope` directly.

model/scope.go
```diff
@@ -0,0 +1,208 @@
...
+
+// Normalize removes all repeated scopes from a list of Scopes.
+func (ss Scopes) Normalize() Scopes {
+	if ss == nil {
+		return nil
+	}
+
+	out := Scopes{}
+
+OUTER:
```
agnivade, line 196: There's a cool zero-alloc algorithm I suggest for this use-case: https://github.com/golang/go/wiki/slicetricks#in-place-deduplicate-comparable. 0/5 on whether you want to use it, but if you do, please add a comment explaining what this code is doing.

agnivade (commented):
Looks like a great start Daniel.

model/scope.go
larkox: If we do that, Go doesn't allow us to do the beautiful
`func (s []Scope) DoSomething()`
since it is not a named type.

If we were to go that direction, all the functions declared here would need to be written as:
`func DoSomething(s []Scope)`
which would polute a lot the model package and I feel is less elegant.

web/handlers.go
agnivade: Yes, it is guaranteed that if err is nil, then oAuthApp is not nil. It is a well-establlished convention and I don't think there is a need to again check for `oAuthApp == nil`.

#18437 was _not_ that case. It happened because we were ignoring not found errors. I have removed all those redundant checks now - https://github.com/mattermost/mattermost-server/pull/18440.

catalintomai: Thanks @agnivade , that makes sense, it was my understanding of the "theory" as well = ). What is less clear (to me) is the "guaranteed" aspect of the "well-established convention" and how do we make sure that the convention is well-enforced throughout the codebase and for new PRs. In the  #18437 case, fwis, the code was structured in such a way that the convention was not respected (the `appErr` could be nil  and `referencedPost` could've been nil as well). In this case, having the extra checks, while reducing the readability, etc. , would've prob. avoided the havoc it created. Bottom line, it looks like the q is: either we rely on the developers to be aware of AND enforce the convention (or find other - automatic? - ways to ensure it is observed) or we'd need to add defensive code with the assumption that the convention _could_ not be respected (with the corresponding drawbacks).  Not sure where the answer lies, but the potential for future mishap has not been eliminated yet imo.

agnivade: Thanks Catalin, I totally get your point. For sure, there are automated [checks](https://github.com/golang/tools/tree/master/go/analysis/passes/nilness) we already run to catch these cases, but they are not always 100% effective.

> Bottom line, it looks like the q is: either we rely on the developers to be aware of AND enforce the convention (or find other - automatic? - ways to ensure it is observed)

IMO, adding defensive coding is also a human factor. How can you "guarantee" that every developer always does defensive coding? The answer should be IMO, to write tests to lock the behavior. But as we all know well enough, that doesn't always happen due to "reasons". I'd suggest for PR reviewers to continue the good practice of asking authors to add unit tests to guarantee these API contracts.

We already have tests which guarantee this behavior for all store methods. But some cases like these, as we can see, the error handling is not done properly. We need to be more careful in PR reviews and add more tests.

Tests and automated checks are the best possible tools against these problems. We just need to spend the time to write them.

catalintomai: Thanks @agnivade , that makes sense, it was my understanding of the "theory" as well = ). Going beyond the scope of this particular change, what is less clear (to me) is the "guaranteed" aspect of the "well-established convention" and how do we make sure that the convention is well-enforced (aka guaranteed) throughout the codebase and for new PRs. You are right, #18437 does not illustrate the case (though one could argue that it illustrates a failure to fully understand the other "side" of the "convention": when appErr is not-nil, the referencedPost is nil) , but the point imo still stands -  we still rely on the developers to be aware of AND enforce the convention (e.g. in cases other than "get obj from store"). With the assumption  that the convention _could_ not be observed (or interpreted incorrectly as above) , one would need to add defensive code - with the corresponding drawbacks - or fix the code - which might be too late.  Not sure where the answer lies, but the potential for future mishap is still there imo.

app/oauth.go
```diff
@@ -367,12 +373,14 @@ func (a *App) GetOAuthAccessTokenForCodeFlow(clientId, grantType, redirectURI, c
...
-func (a *App) newSession(appName string, user *model.User) (*model.Session, *model.AppError) {
+func (a *App) newSession(app *model.OAuthApp, user *model.User) (*model.Session, *model.AppError) {
 	// Set new token an session
 	session := &model.Session{UserId: user.Id, Roles: user.Roles, IsOAuth: true}
 	session.GenerateCSRF()
 	a.srv.userService.SetSessionExpireInDays(session, *a.Config().ServiceSettings.SessionLengthSSOInDays)
-	session.AddProp(model.SessionPropPlatform, appName)
+	session.AddProp(model.SessionPropPlatform, app.Name)
+	session.AddProp(model.SessionPropOAuthAppID, app.Id)
+	session.AddProp(model.SessionPropAppsFrameworkAppID, app.AppsFrameworkAppID) // TODO OAUTH use this value to remove all sessions belonging to an app when uninstalled.
```
hanzei, line 383: Why does the server need to remove the sessions? Can the apps plugin do that?

app/plugin_requests.go
```diff
@@ -194,9 +198,32 @@ func (s *Server) servePluginRequest(w http.ResponseWriter, r *http.Request, hand
...
+					return
+				}
+
+				b, err := json.Marshal(oAuthApp.Scopes)
+				if err != nil {
+					// TODO OAUTH write error
+					w.WriteHeader(http.StatusInternalServerError)
+					return
+				}
+				r.Header.Set("Mattermost-Scopes", string(b))
```
hanzei, line 225: Should this header be deleted if the session _isn't_ an OAuth one?

app/users/session.go
```diff
@@ -153,6 +153,21 @@ func (us *UserService) RevokeSessionsForDeviceId(userID string, deviceID string,
 	return nil
 }
 
+func (us *UserService) RevokeSessionsForOAuthApp(appID string) error {
+	// TODO OAUTH this is really underperformant.
```
hanzei, line 157: Ticket?

model/scope.go
```diff
@@ -0,0 +1,258 @@
...
+			return false
+		}
+	}
+
+	return true
+}
+
+// Normalize removes all repeated scopes from a list of Scopes.
+func (ss Scopes) Normalize() Scopes {
+	if ss == nil {
```
hanzei, line 197: Should this check against `ScopeDeny()` for better readability?

model/scope.go
```diff
@@ -0,0 +1,258 @@
...
+			}
+		}
+
+		out = append(out, inScope)
+	}
+
+	return out
+}
+
+func (ss Scopes) Value() (driver.Value, error) {
```
hanzei, line 217: 1/5 add a comment why this methods is needed

store/localcachelayer/layer.go
```diff
@@ -115,6 +115,9 @@ func NewLocalCacheLayer(baseStore store.Store, metrics einterfaces.MetricsInterf
 		cluster: cluster,
 		metrics: metrics,
 	}
+
+	// TODO OAUTH consider add the OAuth Store cache.
```
hanzei, line 119: Ticket?

hanzei (changes requested):
Great work, Daniel :+1: 

I would love to see a ticket for the left over API scope discussions just to ensure they don't get forgotten.

The other main feedback is that I'm not sure that `AppsFrameworkAppID` is really required. Please see my comments below.

larkox:
@saturninoabril would it be possible to run the e2e tests on this PR to make sure nothing breaks?

saturninoabril:
@larkox Sure, see E2E test at https://automation-dashboard.vercel.app/cycles/795989d2-d450-4ca6-a3dd-351f3ff47b96

store/sqlstore/upgrade.go
```diff
@@ -1403,6 +1403,10 @@ func upgradeDatabaseToVersion610(sqlStore *SqlStore) {
 		}
 	}
 
+	// TODO OAUTH not sure if needed
+	sqlStore.CreateColumnIfNotExistsNoDefault("OAuthApps", "Scopes", "Text", "VARCHAR(1024)")
+	sqlStore.CreateColumnIfNotExists("OAuthApps", "AppsFrameworkAppID", "VARCHAR(26)", "VARCHAR(26)", "")
+
```
streamer45, line 1409: Two comments/questions:

- This probably needs to be moved to 6.2 migration. 
- What's the reason for not setting the default for `Scopes` but doing it for `AppsFrameworkAppID`?

model/scope.go
streamer45: Is this check needed? If it's `nil` the length would be 0 anyway.

model/scope.go
```diff
@@ -0,0 +1,266 @@
...
+		return true
+	}
+
+	if ss2 == nil {
+		return false
+	}
+
+	if len(ss2) > len(ss) {
+		return false
+	}
```
streamer45, line 184: Are we assuming the "sets" have been deduplicated here? Otherwise this could potentially return wrong results.
Slightly tangential question but did you consider using a map to implement scopes? It seems like a closer data structure to what you need.

streamer45 (commented):
Looks good overall :+1: Left a couple of comments.
By the way, is a security review scheduled for this PR?

saturninoabril:
@larkox I haven't look into the details yet but you might want to check the failed tests related to MFA - https://mattermost-cypress-report.s3.amazonaws.com/53982-e50cc27-ee-ent-server-pr-18443-mm-ee-test:e50cc27/mochawesome.html

Screen Shot 2021-11-09 at 6 34 48 AM

model/scope.go
larkox: I did not consider it, no. Usually, they are represented as a list of strings. Not sure if it is worth now making the change to a map.

store/sqlstore/upgrade.go
larkox: I want to keep nil on the existing register. The idea of that is that nil has a special meaning (legacy oauth apps), and want to preserve it.

larkox:
@saturninoabril Are those related to this PR? I do not see how any of my changes may have broken that flow, but if you can verify, I will double check.

model/scope.go
streamer45: Of course not, that was mostly a curiosity of mine. The only real problem with using arrays (slices) is dealing with deduplication really so as long as you do that properly it should be fine.

saturninoabril:
@larkox You're right. I did verify and those were not related to this PR and were timing issue with Cypress.

larkox:
@levb I added the changes we talked about. Please take a look.

model/scope.go
```diff
@@ -0,0 +1,263 @@
...
+type Scope string
+
+const (
+	ScopeChannelJoin  Scope = "channels:join"
+	ScopeMessageRead  Scope = "message:read"
+	ScopeMessageWrite Scope = "message:write"
+	ScopeFilesRead    Scope = "files:read"
+	ScopeFilesWrite   Scope = "files:write"
+	ScopeSearchPosts  Scope = "search:post"
+	ScopeNoScope      Scope = ""
```
levb, line 25: 2/5 rename `ScopeNoScope` to `AnyScope`, more intention-revealing and less typographically redundant.

model/scope.go
```diff
@@ -0,0 +1,263 @@
...
+
+	PluginScopePrefix         = "plugin:"
+	PluginSpecificScopePrefix = "pluginscope:"
+)
+
+func ScopeAny(scopes ...Scope) Scopes {
+	return scopes
+}
+
+func ScopeAllow() Scopes {
```
levb, line 35: 2/5 I still find the representation of Allow as `{}` and Deny as `nil` to be confusing. Is there a strong argument against explicitly codifying `ScopeAllow` as `{""}`?

0/5 I'd add a  `type APIScope Scope` for which `AnyScope = ""` is valid, in addition to all of normal `Scope` values. `nil` for DenyAll seems appropriate, intention-revealing, and the default.

api4/handlers.go
```diff
@@ -8,14 +8,16 @@ import (
...
 	"github.com/mattermost/mattermost-server/v6/web"
 )
 
 type Context = web.Context
 
 // APIHandler provides a handler for API endpoints which do not require the user to be logged in order for access to be
 // granted.
-func (api *API) APIHandler(h func(*Context, http.ResponseWriter, *http.Request)) http.Handler {
+func (api *API) APIHandler(h func(*Context, http.ResponseWriter, *http.Request), scopes ...model.Scope) http.Handler {
+	handlerScopes := scopeFromArgs(scopes)
```
levb, line 20: See my comment in `scope.go`; 1/5 with my proposed change you just do `Scopes(scopes)`, no need for different representations. You can still use the custom `Scopes` type for all the method bindings, but no need to change the representation?

api4/plugin.go
```diff
@@ -37,11 +37,11 @@ func (api *API) InitPlugin() {
 	api.BaseRoutes.Plugin.Handle("/enable", api.APISessionRequired(enablePlugin)).Methods("POST")
 	api.BaseRoutes.Plugin.Handle("/disable", api.APISessionRequired(disablePlugin)).Methods("POST")
 
-	api.BaseRoutes.Plugins.Handle("/webapp", api.APIHandler(getWebappPlugins)).Methods("GET")
+	api.BaseRoutes.Plugins.Handle("/webapp", api.APIHandler(getWebappPlugins, model.ScopeNoScope)).Methods("GET")
 
 	api.BaseRoutes.Plugins.Handle("/marketplace", api.APISessionRequired(getMarketplacePlugins)).Methods("GET")
 
-	api.BaseRoutes.Plugins.Handle("/marketplace/first_admin_visit", api.APIHandler(setFirstAdminVisitMarketplaceStatus)).Methods("POST")
+	api.BaseRoutes.Plugins.Handle("/marketplace/first_admin_visit", api.APISessionRequired(setFirstAdminVisitMarketplaceStatus)).Methods("POST")
```
levb, line 44: why did this change from `APIHandler` to `APISessionRequired`?

api4/saml.go
```diff
@@ -28,7 +28,7 @@ func (api *API) InitSaml() {
 
 	api.BaseRoutes.SAML.Handle("/certificate/status", api.APISessionRequired(getSamlCertificateStatus)).Methods("GET")
 
-	api.BaseRoutes.SAML.Handle("/metadatafromidp", api.APIHandler(getSamlMetadataFromIdp)).Methods("POST")
+	api.BaseRoutes.SAML.Handle("/metadatafromidp", api.APISessionRequired(getSamlMetadataFromIdp)).Methods("POST")
```
levb, line 31: why did this change from `APIHandler` to `APISessionRequired`?

api4/system.go
```diff
@@ -62,8 +62,8 @@ func (api *API) InitSystem() {
 	api.BaseRoutes.APIRoot.Handle("/upgrade_to_enterprise/status", api.APISessionRequired(upgradeToEnterpriseStatus)).Methods("GET")
 	api.BaseRoutes.APIRoot.Handle("/restart", api.APISessionRequired(restart)).Methods("POST")
 	api.BaseRoutes.APIRoot.Handle("/warn_metrics/status", api.APISessionRequired(getWarnMetricsStatus)).Methods("GET")
-	api.BaseRoutes.APIRoot.Handle("/warn_metrics/ack/{warn_metric_id:[A-Za-z0-9-_]+}", api.APIHandler(sendWarnMetricAckEmail)).Methods("POST")
-	api.BaseRoutes.APIRoot.Handle("/warn_metrics/trial-license-ack/{warn_metric_id:[A-Za-z0-9-_]+}", api.APIHandler(requestTrialLicenseAndAckWarnMetric)).Methods("POST")
+	api.BaseRoutes.APIRoot.Handle("/warn_metrics/ack/{warn_metric_id:[A-Za-z0-9-_]+}", api.APISessionRequired(sendWarnMetricAckEmail)).Methods("POST")
```
levb, line 65: why did this change from `APIHandler` to `APISessionRequired`?

api4/system.go
```diff
@@ -339,7 +339,7 @@ func postLog(c *Context, w http.ResponseWriter, r *http.Request) {
 	forceToDebug := false
 
 	if !*c.App.Config().ServiceSettings.EnableDeveloper {
-		if c.AppContext.Session().UserId == "" {
+		if c.AppContext.Session().UserId == "" || c.AppContext.Scopes() != nil {
```
levb, line 342: Can you please elaborate on this change, @larkox ?

model/scope.go
levb: Since `Scopes` is already a type, can be a map in go, serialized as an array in JSON

levb (changes requested):

app/oauth.go
```diff
@@ -101,6 +101,12 @@ func (a *App) UpdateOAuthApp(oldApp, updatedApp *model.OAuthApp) (*model.OAuthAp
 		}
 	}
 
+	if !oldApp.Scopes.IsSuperset(updatedApp.Scopes) {
```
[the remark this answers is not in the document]
larkox: IsSuperset means "has all the scopes from the other app and more". The logic here is "delete all the sessions if I need more scopes than what I have already been granted". Therefore, an old app with scopes `user:read` and `user:write` and a new app with `user:read` won't trigger the deletion of sessions (which will if we apply your change). The other way around, an old app with scopes `user:read` and a new app with scopes `user:read` and `user:write` should trigger the deletion of sessions (which will not if we apply your change).

model/scope.go
larkox: My problem with this renaming would be that we would have in the same namespace the constant "ScopeAny" (to define the any scope on API level), and the function "ScopeAny" (to create a Scope[] from a list of scopes).

I am open to change both names, but not sure which name to choose.

api4/plugin.go
larkox: Piggybacked semantical fix. This function was checking a session was available, but the API handler was not doing it. There are no functional problems, but I guess I should change it.

api4/saml.go
larkox: Same as in previous comment.

api4/system.go
larkox: Sames as previous comment.

api4/system.go
larkox: `nil` in this context means "it is a legacy app, or no app". If the session is not coming from an app or the app is legacy, act as a normal user.

If not, go safe, and act as if you had no permission for this endpoint.

model/scope.go
larkox: > 2/5 I still find the representation of Allow as {} and Deny as nil to be confusing. Is there a strong argument against explicitly codifying ScopeAllow as {""}?

It makes things slightly more complicated in terms of code. The current behavior is simple codewise:
- Nil = deny
- Empty = allow
- With something = scoped

If we make your proposed change, we have way more things to handle:
- Nil = deny
- With something:
  - Has some "" element? If so, it must be of length 1, and means it is allowed. Otherwise is not valid
  - Has no "" element? scoped
- Empty = invalid?

This also may add complexity to other functions, but I haven't gone through it yet.

api4/handlers.go
larkox: If we go with Scopes(scopes), we would be changing the whole meaning to:
- Empty = deny
- Includes "" = allow
- Anything else = scoped

We could go that direction, but not sure what is more confusing as a whole, doing this big change here, or keep it as is.

model/scope.go
levb: Then how about `NoScope`? Seems unique and self-explanatory?

model/scope.go
levb: Empty and nil should be equivalent, IMO.

levb (approved):
I still feel unconvinced about the `{}` vs `nil` semantic difference, but I think this is something we can fix in the code, separately and later if we find it useful. Otherwise great work, @larkox; thank you for taking it on and your patience.

model/scope.go
hanzei: From reading the syntax, scopes seems like to form a tree. `files:read` and `files:write` are both leaves it in. Are there also branch nodes? Will there be a scope `file`?

If that is the case, then there is a root leave, it could be named `""`, `":"` or `"."`. Then the metal model stays the same for other scoped cases and for the root: Use the scope to determinate, if the request is valid. If the scope is `root`, it's allowed.

larkox: The plan is that no, there are no trees. The tree-like names are so they are clearer, and easier to visualize  / organize. We could create some "alias" like `file:all`, but I would shy away from that idea (at least for now).

I am OK using the empty string as any, but just as a "clearer way to do things", not so much as root/leaves mental model.

store/sqlstore/upgrade.go
```diff
@@ -1438,6 +1438,9 @@ func upgradeDatabaseToVersion640(sqlStore *SqlStore) {
 		sqlStore.GetMaster().ExecNoTimeout("ALTER TABLE Users DROP COLUMN AcceptedTermsOfServiceId")
 	}
 
+	sqlStore.CreateColumnIfNotExistsNoDefault("OAuthApps", "Scopes", "Text", "VARCHAR(1024)")
```
levb, line 1441: @isacikgoz @larkox `CreateColumnIfNotExistsNoDefault` was removed as unused in https://github.com/mattermost/mattermost-server/pull/19354. Any recommendation on how to resolve the conflict?

larkox: The new migration system has to be used, so these have to be removed and handled in the files in `/db/migrations/`. There should be some documentation about how to handle this, but I don't remember where.

hanzei: https://github.com/mattermost/mattermost-server/pull/19214 can be used an an example

levb:
Replaced by https://github.com/mattermost/mattermost-server/pull/20941

PR #18451, opened 2021-09-23
MM-36862: should remove user from participants

koox00:
#### Summary

When deleting a reply in a thread we should also delete the participant
from the participants array. This should happen if they have no other
replies in that thread.

This commit fixes that.

#### Ticket Link

https://mattermost.atlassian.net/browse/MM-36862

#### Release Note

```release-note
NONE
```

### DIFF
--- a/store/sqlstore/post_store.go
+++ b/store/sqlstore/post_store.go
@@ -673,8 +673,9 @@ func (s *SqlPostStore) Delete(postID string, time int64, deleteByID string) erro
 		return errors.Wrap(err, "failed to update Posts")
 	}
 
+	ids := postIds{}
 	// TODO: change this to later delete thread directly from postID
-	rootID, err := s.GetReplica().SelectStr("SELECT RootId FROM Posts WHERE Id = :Id", map[string]interface{}{"Id": postID})
+	err = s.GetReplica().SelectOne(&ids, "SELECT RootId, UserId FROM Posts WHERE Id = :Id", map[string]interface{}{"Id": postID})
 	if err != nil {
 		if err == sql.ErrNoRows {
 			return store.NewErrNotFound("Post", postID)
@@ -683,7 +684,7 @@ func (s *SqlPostStore) Delete(postID string, time int64, deleteByID string) erro
 		return errors.Wrapf(err, "failed to delete Post with id=%s", postID)
 	}
 
-	return s.cleanupThreads(postID, rootID, false)
+	return s.cleanupThreads(postID, ids.RootId, false, ids.UserId)
 }
 
 func (s *SqlPostStore) permanentDelete(postId string) error {
@@ -692,7 +693,7 @@ func (s *SqlPostStore) permanentDelete(postId string) error {
 	if err != nil && err != sql.ErrNoRows {
 		return errors.Wrapf(err, "failed to get Post with id=%s", postId)
 	}
-	if err = s.cleanupThreads(post.Id, post.RootId, true); err != nil {
+	if err = s.cleanupThreads(post.Id, post.RootId, true, post.UserId); err != nil {
 		return errors.Wrapf(err, "failed to cleanup threads for Post with id=%s", postId)
 	}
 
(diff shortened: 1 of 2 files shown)


store/sqlstore/post_store.go
```diff
@@ -2364,7 +2365,40 @@ func (s *SqlPostStore) cleanupThreads(postId, rootId string, permanent bool) err
...
+
+		if err != nil {
+			mlog.Warn("Error getting thread participants", mlog.Err(err))
+		}
+
+		if count == 0 && participants.Contains(userId) {
+			participants = participants.Remove(userId)
+		}
+
+		_, err = s.GetMaster().Exec(`UPDATE Threads SET ReplyCount = ReplyCount - 1, Participants = :Participants WHERE PostId = :Id AND ReplyCount > 0`, map[string]interface{}{"Id": rootId, "Participants": participants})
```
streamer45, line 2401: A couple of things:

- We should probably aim at fetching (and setting of course) participants only when we actually have to update the list (count is zero). I think it should be possible using the squirrel query builder without causing too much code duplication.
- Since we made the `Threads.Participants` column to be of native JSON type we may want to explore a way to do the search and update on the DB side. This can be deferred to a later PR but something to keep in mind. /cc @agnivade

koox00: > Since we made the `Threads.Participants` column to be of native JSON type we may want to explore a way to do the search and update on the DB side. This can be deferred to a later PR but something to keep in mind. /cc @agnivade

If you ever get to that please ping me, I would like to see how it's going to be implemented :)

koox00: > We should probably aim at fetching (and setting of course) participants only when we actually have to update the list (count is zero). I think it should be possible using the squirrel query builder without causing too much code duplication.

Just a question, could the following be reduced in one line?

```go

			queryString, args, _ = s.getQueryBuilder().
				Select("Participants").
				From("Threads").
				Where(sq.Eq{"PostId": rootId}).
				ToSql()

			var participants model.StringArray
			err = s.GetReplica().SelectOne(&participants, queryString, args...)
```

streamer45: I don't think we ever made use of `RunWith()` but I guess it should work just fine.

store/sqlstore/post_store.go
```diff
@@ -2364,7 +2365,54 @@ func (s *SqlPostStore) cleanupThreads(postId, rootId string, permanent bool) err
...
+				sq.Eq{"UserId": userId},
+				sq.Eq{"DeleteAt": 0},
+			}).
+			ToSql()
+
+		count, err := s.GetReplica().SelectInt(queryString, args...)
+
+		if err != nil {
+			mlog.Warn("Error counting user's posts in thread", mlog.Err(err))
+		}
```
ashishbhate, line 2382: If this fails, then `count` will be `0` and the user will be removed from the participants array. Maybe we should skip removal if this fails. I'm 0/5.

ashishbhate (approved):

store/sqlstore/post_store.go
koox00: made the change :)

streamer45 (approved):

store/sqlstore/post_store.go
agnivade: @koox00 - the code is already here:

https://github.com/mattermost/mattermost-server/blob/68517a77195664c15ee24e7b12677b0f66a94d5c/store/sqlstore/thread_store.go#L685-L703


We just need to remove from the array and invert the membership check. We added native JSON support to explicitly avoid double querying like this. I have been stuck with some other tasks and therefore have not been able to demo these new features.

But I'd love to see us use these features. Would you be open to creating an HW ticket to convert this to native JSON queries? Seems like an interesting ticket to have.

koox00: Yep sure, added a ticket as a HW candidate https://mattermost.atlassian.net/browse/MM-39157

jgilliam17:
> If the thread is open I still have the avatar with "message delete", but the main channel view is ok (number of replies and avatar). Is it normal? 

@alauregaillard I think that is ok. The deleted message is ephemeral, and will be gone once user closes/opens RHS again or refreshes the page.

PR #18454, opened 2021-09-24
Exclude sending file metadata to plugin hooks

agnivade:
A FileInfo object contained a MiniPreview which is
a slice of bytes. This can be particularly costly
while marshalling to plugin hooks.

We avoid this by refactoring the Embeds and Images
population to a separate method and calling that
to prevent posts from getting updated.

https://community-daily.mattermost.com/boards/workspace/zyoahc9uapdn3xdptac6jb69ic/285b80a3-257d-41f6-8cf4-ed80ca9d92e5/495cdb4d-c13a-4992-8eb9-80cfee2819a4?c=9c0b5413-5401-4ef2-83d5-b9f756585bbc

```release-note
NONE
```

### DIFF
--- a/api4/post.go
+++ b/api4/post.go
@@ -128,7 +128,7 @@ func createEphemeralPost(c *Context, w http.ResponseWriter, r *http.Request) {
 
 	w.WriteHeader(http.StatusCreated)
 	rp = model.AddPostActionCookies(rp, c.App.PostActionCookieSecret())
-	rp = c.App.PreparePostForClient(rp, true, false)
+	rp = c.App.PreparePostForClient(rp, true, false, true)
 	rp, err := c.App.SanitizePostMetadataForUser(rp, c.AppContext.Session().UserId)
 	if err != nil {
 		c.Err = err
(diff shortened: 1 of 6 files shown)


streamer45 (approved):
Nice :+1:

lieut-data (approved):
Thank you, @agnivade!

I didn't scrutinize each new Boolean parameter's value in all the cases. (It might be interesting to add methods like `PrepareNewPostForClientWithEmbeds` to make this really explicit.)

Do we understand the source of the overhead? Is the payload particularly big, or is there something about encoding byte slices specifically?

agnivade:
I'll update the method names.

> Do we understand the source of the overhead? Is the payload particularly big, or is there something about encoding byte slices specifically?

It's about encoding the mini previews of images. The resolution of the image is 16x16, but it does not control the size of the image in bytes, and is dependent on the image. And moreover, the total underlying size of the slice can be more than the size of image in bytes. And then it gets marshalled `n` number of times for each plugin hook, magnifying the total allocation.

agnivade:
@mkraft - I am going to merge this to unblock community releases. But let me know if you have any comments, and I can send another PR.

agnivade:
@amyblais - Request for permission to cherry-pick to 6.0 and cloud.

PR #18464, opened 2021-09-24
[MM-38579] Removing some town square permissions stuff

BenCookie95:
#### Summary
I removed all the mentions of the read only error and now you can post, react and pin just like any other channel. That can be disabled with the permissions in system console.

Town square still has some exceptions to the regular permissions though where you can't leave the channel, change it's type or archive it.

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-38579

#### Release Note

```release-note
NONE
```

### DIFF
--- a/api4/post.go
+++ b/api4/post.go
@@ -713,32 +713,12 @@ func saveIsPinnedPost(c *Context, w http.ResponseWriter, isPinned bool) {
 		return
 	}
 
-	// Restrict pinning if the experimental read-only-town-square setting is on.
-	user, err := c.App.GetUser(c.AppContext.Session().UserId)
-	if err != nil {
-		c.Err = err
-		return
-	}
-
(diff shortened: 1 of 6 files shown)


lieut-data (approved):
Sweet! Thank you, @BenCookie95 :)

mkraft (approved):

lieut-data:
@amyblais / @BenCookie95, should this go to `cloud` as well?

amyblais:
It could go to Oct 12th cloud release, unless this is high priority for tomorrow's release? @lieut-data

lieut-data:
@amyblais, I believe the implications of this bug is that Town Square will be unusable for anyone in cloud that's not a system administrator. We're seeing it continue to fail on our E2E tests that run against the current `cloud` branch.

PR #18468, opened 2021-09-27
Cross team search+autocomplete

agnivade:
#### Summary

This is the sequel to https://github.com/mattermost/mattermost-server/pull/18374 which contains cross-team search capabilities along with private channels.

New endpoints have been added to return all joined channels from all teams to help reduce the number of API requests (as per request from front-end)

#### Ticket Link

https://mattermost.atlassian.net/browse/MM-26721
https://mattermost.atlassian.net/browse/MM-18496

#### Webapp PR
https://github.com/mattermost/mattermost-webapp/pull/8888

#### Release Note

```release-note
Channel results in channel switcher will include channels from other teams as well as private channels.
For users using Bleve or Elasticsearch for autocomplete/search would have to reindex their data to get the new results. Since this can take a long time, it is suggested to disable autocomplete and run indexing in the background. And then when it's complete, re-enable autocomplete.
```

cc @AshishDhama


agnivade:
CircleCI outage. Not to worry.

agnivade:
@streamer45 , @mgdelacroix - this is ready for review.

store/sqlstore/channel_store.go
```diff
@@ -998,6 +998,83 @@ func (s SqlChannelStore) GetChannels(teamId string, userId string, includeDelete
...
+	}
+
+	if len(channels) == 0 {
+		return nil, store.NewErrNotFound("Channel", "userId="+userId)
+	}
+
+	return channels, nil
+}
+
+func (s SqlChannelStore) GetAllChannelMembersById(id string) ([]string, error) {
```
streamer45, line 1053: Could we rename the input parameter to `ChannelID`? Just to make it explicit since `ChannelMembers` are identified by two ids.

store/sqlstore/channel_store.go
```diff
@@ -998,6 +998,83 @@ func (s SqlChannelStore) GetChannels(teamId string, userId string, includeDelete
...
+		for _, member := range channelMembers {
+			res = append(res, member.UserId)
+		}
+
+		if len(channelMembers) < perPage {
+			break
+		}
+
+		page++
+	}
```
streamer45, line 1073: Is there an actual benefit/need to page through here? Can't we simply run a query to return them all at once?

store/sqlstore/channel_store.go
```diff
@@ -2396,6 +2473,34 @@ func (s SqlChannelStore) GetChannelsByIds(channelIds []string, includeDeleted bo
...
+	if !includeDeleted {
+		query = `SELECT c.*, t.DisplayName AS TeamDisplayName, t.Name AS TeamName, t.UpdateAt AS TeamUpdateAt
+			 FROM Channels c, Teams t
+			 WHERE
+				c.TeamId=t.Id
+				AND
+				c.DeleteAt = 0
+				AND
+			  c.Id IN ` + keys + ` ORDER BY c.Name`
+	}
```
streamer45, line 2493: Could we use squirrel here? Maybe the `!includeDeleted` case would become much simpler.

store/sqlstore/channel_store.go
```diff
@@ -2465,48 +2570,84 @@ func (s SqlChannelStore) GetMembersForUserWithPagination(teamId, userId string,
 	return dbMembers.ToModel(), nil
 }
 
-func (s SqlChannelStore) AutocompleteInTeam(teamId string, term string, includeDeleted bool) (model.ChannelList, error) {
-	deleteFilter := "AND Channels.DeleteAt = 0"
-	if includeDeleted {
-		deleteFilter = ""
+func (s SqlChannelStore) GetTeamMembersForChannel(channelID string) ([]string, error) {
+	teamMemberIDs := []string{}
+	if err := s.GetReplicaX().Select(&teamMemberIDs, `SELECT tm.UserId
```
streamer45, line 2575: Do we mean using `GetReplicaX()`?

streamer45 (commented):
Looks good overall :+1:  

I've left some comments here and there but my main concern, as usual, would be around performance. Once client-side changes are finalized they should be added to the load-test side as well and properly tested.

agnivade:
> I've left some comments here and there but my main concern, as usual, would be around performance. Once client-side changes are finalized they should be added to the load-test side as well and properly tested.

Aye aye. @AshishDhama - could you please clarify if my understanding is correct: at page load, we get the channels from all teams using the `/channels` API, and then for autocomplete from channel switcher, we use the `/autocomplete` API.

model/client4.go
```diff
@@ -3100,6 +3100,24 @@ func (c *Client4) GetChannelsForTeamAndUserWithLastDeleteAt(teamId, userId strin
 	return ch, BuildResponse(r), nil
 }
 
+// GetChannelsForUserWithLastDeleteAt returns a list channels for a user, additionally filtered with lastDeleteAt.
+func (c *Client4) GetChannelsForUserWithLastDeleteAt(userID string, lastDeleteAt int) ([]*Channel, *Response, error) {
```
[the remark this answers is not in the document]
agnivade: I was trying to match with the other existing methods. But 0/5, let me know.

store/searchlayer/channel_layer.go
```diff
@@ -128,18 +156,52 @@ func (c *SearchChannelStore) SaveDirectChannel(directchannel *model.Channel, mem
 	if err != nil {
 		c.rootStore.indexUserFromID(member1.UserId)
 		c.rootStore.indexUserFromID(member2.UserId)
+		c.indexChannel(channel)
```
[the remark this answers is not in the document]
agnivade: Good question. I must say I am just following what was already done :see_no_evil: Why were we indexing users on failure? 

@mgdelacroix any ideas? Should I remove this?

store/searchlayer/channel_layer.go
```diff
@@ -165,32 +227,56 @@ func (c *SearchChannelStore) AutocompleteInTeam(teamId string, term string, incl
...
 
 func (c *SearchChannelStore) PermanentDeleteMembersByUser(userId string) error {
+	channels, errGetChannels := c.ChannelStore.GetChannelsByUser(userId, false, 0, -1, "")
+	if errGetChannels != nil {
+		mlog.Warn("Encountered error indexing channel after removing user", mlog.String("user_id", userId), mlog.Err(errGetChannels))
+	}
+
 	err := c.ChannelStore.PermanentDeleteMembersByUser(userId)
-	if err == nil {
+	if err == nil && errGetChannels == nil {
```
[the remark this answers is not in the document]
agnivade: Ah nice one. Will update. :+1:

store/sqlstore/channel_store.go
agnivade: Good point. What I was thinking is that, purely from a theoretical perspective, the total amount of required DB memory would be slightly larger than just the number of rows, since that query has to join 5 tables. OTOH, the server just has to handle purely the results. And moreover, since this is only called from a job, it's okay to spread the load on the DB slightly.

I don't have 100% experimental proof though. Just using the force. Okay to fetch full results if you think it'll be better.

store/sqlstore/channel_store.go
agnivade: Didn't get this. Can you clarify?

api4/channel.go
```diff
@@ -899,6 +900,79 @@ func getChannelsForTeamForUser(c *Context, w http.ResponseWriter, r *http.Reques
...
+	}
+	if lastDeleteAt < 0 {
+		c.SetInvalidURLParam("last_delete_at")
+		return
+	}
+
+	pageSize := 100
+	fromChannelID := ""
+	// We have to write `[` and `]` separately because we want to stream the response.
+	// The internal API is paginated, but the client always needs to get the full data.
```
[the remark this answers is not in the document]
AshishDhama: @agnivade @streamer45 As we need data of channels across team we had two options either load channels when user opens Channel switcher or Load all the channels on app load. So to make UX better [as discussed here](https://community-daily.mattermost.com/core/pl/16otwhqhpfn58ym9h3tjfoiiso) we have taken middle path of loading all the channels user is part of across teams on app load. As generally user is looking to switch channel instead of joining.

store/sqlstore/channel_store.go
streamer45: Did you mean calling the sqlx method instead of the gorp one?

api4/channel.go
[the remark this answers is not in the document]
streamer45: Okay cool, this only returns channels the user is part of (has joined). Was slightly worried for a second :)

store/sqlstore/channel_store.go
agnivade: Yes, I meant the sqlx method.

store/sqlstore/channel_store.go
streamer45: I think your theory would work but only if the limiting of data was done in the JOINs. Looking at the query we are just limiting the final set of rows so from a DB perspective there wouldn't be any performance benefit of re-running the query more than once as it's the same work again no matter the page size.

agnivade: Of course, you are right.

esethna:
@agnivade just to understand, does this also cover the backend changes needed for https://mattermost.atlassian.net/browse/MM-6482?

agnivade:
@esethna - Yes. @AshishDhama will be working on that once he's done with the channel switcher.

store/searchlayer/channel_layer.go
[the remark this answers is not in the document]
mgdelacroix: Eeeer nope, I think we can safely remove this

store/searchtest/helper.go
```diff
@@ -251,6 +251,13 @@ func (th *SearchTestHelper) createChannel(teamID, name, displayName, purpose str
 		return nil, err
 	}
 
+	if channelType == model.ChannelTypePrivate && user != nil {
```
mgdelacroix, line 254: Why adding the user only if the channel is private?

store/sqlstore/channel_store.go
```diff
@@ -2465,48 +2570,84 @@ func (s SqlChannelStore) GetMembersForUserWithPagination(teamId, userId string,
...
+		FROM Channels c, Teams t, TeamMembers tm
```
mgdelacroix, line 2576: Can we try and use squirrel in this and the rest of the queries as well? Happy to create some hacktoberfest tickets for them when this gets merged if preferred :)

mgdelacroix (approved):
Looking great!! :+1: 

Left some minor comments, nothing really blocking

store/sqlstore/channel_store.go
agnivade: I think this is simple enough as a raw query. Unless there are various if conditions while constructing the query, I prefer raw SQL queries. :)

model/client4.go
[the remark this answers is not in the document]
streamer45: It feels like `lastDeleteAt` is really a filtering option so from an API design perspective I think it would be wiser to have it inside an object that can be extended without breaking things nor requiring adding yet another method. That said, if you think this method is not going to need any addition in the foreseeable future (v6 lifespan) then it's totally fine to leave it as is.

api4/channel.go
```diff
@@ -899,6 +900,79 @@ func getChannelsForTeamForUser(c *Context, w http.ResponseWriter, r *http.Reques
...
+	w.Write([]byte(`[`))
+	enc := json.NewEncoder(w)
+	for {
+		channels, err := c.App.GetChannelsForUser(c.Params.UserId, c.Params.IncludeDeleted, lastDeleteAt, pageSize, fromChannelID)
+		if err != nil {
+			// If the page size was a perfect multiple of the total number of results,
+			// then the last query will always return zero results.
+			if fromChannelID != "" && err.Id == "app.channel.get_channels.not_found.app_error" {
+				break
+			}
```
streamer45, line 942: Is this case currently covered by tests? I am just concerned that if the error ID string were to change nothing would apparently break.

model/client4.go
```diff
@@ -3160,6 +3178,29 @@ func (c *Client4) SearchAllChannels(search *ChannelSearch) (ChannelListWithTeamD
...
 }
 
+// SearchAllChannelsForUser search in all the channels for a regular user.
+func (c *Client4) SearchAllChannelsForUser(term string) (ChannelListWithTeamData, *Response, error) {
+	search := &ChannelSearch{
+		Term: term,
+	}
+	searchJSON, jsonErr := json.Marshal(search)
+	if jsonErr != nil {
+		return nil, nil, NewAppError("SearchAllChannels", "api.marshal_error", nil, jsonErr.Error(), http.StatusInternalServerError)
```
streamer45, line 3188:
```suggestion
		return nil, nil, NewAppError("SearchAllChannelsForUser", "api.marshal_error", nil, jsonErr.Error(), http.StatusInternalServerError)
```

model/client4.go
```diff
@@ -3160,6 +3178,29 @@ func (c *Client4) SearchAllChannels(search *ChannelSearch) (ChannelListWithTeamD
...
+	r, err := c.DoAPIPost(c.channelsRoute()+"/search?system_console=false", string(searchJSON))
+	if err != nil {
+		return nil, BuildResponse(r), err
+	}
+	defer closeBody(r)
+
+	var ch ChannelListWithTeamData
+	err = json.NewDecoder(r.Body).Decode(&ch)
+	if err != nil {
+		return nil, BuildResponse(r), NewAppError("SearchAllChannels", "api.marshal_error", nil, err.Error(), http.StatusInternalServerError)
```
streamer45, line 3199:
```suggestion
		return nil, BuildResponse(r), NewAppError("SearchAllChannelsForUser", "api.marshal_error", nil, err.Error(), http.StatusInternalServerError)
```

store/sqlstore/channel_store.go
```diff
@@ -998,6 +998,73 @@ func (s SqlChannelStore) GetChannels(teamId string, userId string, includeDelete
...
+}
+
+func (s SqlChannelStore) GetAllChannelMembersById(channelID string) ([]string, error) {
+	var dbMembers channelMemberWithSchemeRolesList
+	_, err := s.GetReplica().Select(&dbMembers, ChannelMembersWithSchemeSelectQuery+"WHERE ChannelId = :ChannelId", map[string]interface{}{"ChannelId": channelID})
+	if err != nil {
+		return nil, errors.Wrapf(err, "failed to get ChannelMembers with channelID=%s", channelID)
+	}
+
+	res := make([]string, 0, len(dbMembers.ToModel()))
```
streamer45, line 1060: Do we need to do the actual conversion to get the length? I am thinking we could either do it once to be re-used in the loop below or just use the length of the original slice as I don't believe it changes.

streamer45 (approved):
Left a couple more non-blocking comments but otherwise looks great :+1:

esethna:
@agnivade can this be merged without the associated front-end changes? ie will anything break if we merge this earlier? 

If it's safe should we remove the blocking label?

agnivade:
No we cannot merge this. Community will immediately break.

I am keeping an eye on it. Dhama has some more work to be done on the front-end side. And then after proper testing, we need to do some steps with the SRE team and then we can merge it. I will do all those.

The label was for the API PR as well.

hmhealey:
All of the channel members for the current user, so instead of having an API that just returns an array of channels, we would add one that returns an object with two arrays, one with the channels and another with the corresponding channel members for that user.

PR #18477, opened 2021-09-27
Pre-package Playbooks 1.20.0

lieut-data:
#### Summary
Pre-package Playbooks 1.20.0

#### Ticket Link
Fixes: https://mattermost.atlassian.net/browse/MM-37938

#### Release Note
```release-note
Pre-package Playbooks 1.20.0
```


model/feature_flags.go
```diff
@@ -59,7 +59,7 @@ func (f *FeatureFlags) SetDefaults() {
 	f.CollapsedThreads = true
 	f.EnableRemoteClusterService = false
 	f.AppsEnabled = false
-	f.PluginIncidentManagement = "1.16.1"
+	f.PluginIncidentManagement = "1.20.0"
```
crspeller, line 62: Wrong one. Needs to be `PluginPlaybooks`. You can delete this line and the old feature flag if we don't want to control the old versions with split.io.

crspeller (changes requested):

crspeller (approved):

PR #18491, opened 2021-09-30
build: add targets for linux/darwin arm64 and include those in the package

cpanato:
#### Summary
Add targets to build linux and darwin arm64 and include that in the package as well

TODO and not in this PR:
- build the plugins ARM64 targets
- build mmctl ARM64 targets
- update the package here to include those
- build the container image for linux arm64 as well

#### Ticket Link
JIRA: https://mattermost.atlassian.net/browse/DOPS-633

#### Release Note

```release-note
build: add targets for linux/darwin arm64 and include those in the package
```

### DIFF
--- a/build/release.mk
+++ b/build/release.mk
@@ -8,6 +8,13 @@ else
 	mkdir -p $(GOBIN)/linux_amd64
 	env GOOS=linux GOARCH=amd64 $(GO) build -o $(GOBIN)/linux_amd64 $(GOFLAGS) -trimpath -ldflags '$(LDFLAGS)' ./...
 endif
+	@echo Build Linux arm64
+ifeq ($(BUILDER_GOOS_GOARCH),"linux_arm64")
+	env GOOS=linux GOARCH=arm64 $(GO) build -o $(GOBIN) $(GOFLAGS) -trimpath -ldflags '$(LDFLAGS)' ./...
+else
+	mkdir -p $(GOBIN)/linux_arm64
+	env GOOS=linux GOARCH=arm64 $(GO) build -o $(GOBIN)/linux_arm64 $(GOFLAGS) -trimpath -ldflags '$(LDFLAGS)' ./...
+endif
 
 build-osx:
 	@echo Build OSX amd64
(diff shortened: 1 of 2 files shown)


cpanato:
@stylianosrigas @spirosoik @jwilander first step in towards to have ARM nodes in the cloud servers to reduce costs :)

cpanato:
sample screenshot running on AWS ubuntu ARM64

Screenshot 2021-09-30 at 11 20 49

spirosoik:
@cpanato great work Carlos!!!

cpanato:
there are some refactors we can do in the makefile, but the focus of this PR is not to do this kind of things for now

build/release.mk
```diff
@@ -117,114 +145,186 @@ endif
...
+			if [ "$$HAS_ARCH" != "dist/plugin-linux-arm64" ]; then \
+				echo "Contains $$HAS_ARCH in $$plugin_package-$$ARCH.tar.gz but needs dist/plugin-linux-arm64"; \
+				exit 1; \
+			fi; \
+			gpg --verify $(DIST_PATH_LIN_ARM64)/prepackaged_plugins/$$plugin_package-$$ARCH.tar.gz.sig $(DIST_PATH_LIN_ARM64)/prepackaged_plugins/$$plugin_package-$$ARCH.tar.gz; \
+			if [ $$? -ne 0 ]; then \
+				echo "Failed to verify $$plugin_package-$$ARCH.tar.gz|$$plugin_package-$$ARCH.tar.gz.sig"; \
+				exit 1; \
+			fi; \
+		fi; \
```
agnivade, line 285: Seems like this is repeated in 3 places now. Wondering if it would be possible to refactor this to a common block?

build/release.mk
```diff
@@ -117,114 +145,186 @@ endif
...
+			exit 1; \
+		fi; \
+	done
+	@# Package
+	cd $(DIST_PATH_WIN)/.. && zip -9 -r -q -l ../mattermost-$(BUILD_TYPE_NAME)-windows-amd64.zip mattermost ../mattermost && cd ../..
+	@# Cleanup
+	rm -rf $(DIST_ROOT)/windows
 
-package: package-osx package-windows package-linux
+package: package-osx-amd64 package-osx-arm64 package-linux-amd64 package-linux-arm64 package-windows
```
agnivade, line 328: 0/5, but would prefer to keep these targets unchanged, and call these individual targets from under the os level target.

build/release.mk
```diff
@@ -17,6 +24,13 @@ else
 	mkdir -p $(GOBIN)/darwin_amd64
 	env GOOS=darwin GOARCH=amd64 $(GO) build -o $(GOBIN)/darwin_amd64 $(GOFLAGS) -trimpath -ldflags '$(LDFLAGS)' ./...
 endif
+	@echo Build OSX arm64
+ifeq ($(BUILDER_GOOS_GOARCH),"darwin_arm64")
```
agnivade, line 28: Since these will be part of the official releases, I am wondering of the confidence we have in darwin/arm64?

Totally fine with linux/arm64, but maybe we want to add somewhere that linux/amd64 is our primary target and fully tested. Otherwise it might start giving incorrect expectations.

cc @jwilander

build/release.mk
agnivade: Apologies, I just saw https://github.com/mattermost/mattermost-server/pull/18491#issuecomment-931260977. Any reason we cannot do these refactors with the PR? I'd prefer to keep the Makefile clean if possible.

build/release.mk
cpanato: we currently don't test the windows and mac 

but we also starting to have some developers using the Mac m1 so might be good to have this

build/release.mk
cpanato: the idea is not to refactor anything now, we are planning to make a better refactor on this soon, and this might disappear, but I need to validate some things before, so I would prefer to keep all the duplication for now

jwilander (approved):

puerco (approved):

PR #18492, opened 2021-09-30
MM-36589: provide previous values for unreads

koox00:
#### Summary

To successfully figure out the new counts of mentions or unread replies
for CRT we need to provide previous values alongside with the new.
This is needed so we'll know how many to subtract from the total.

This commit provides those numbers upon publishing `WebsocketEventThreadUpdated`
and `WebsocketThreadReadChanged` events.

#### Ticket Link

https://mattermost.atlassian.net/browse/MM-36589

#### Release Note

```release-note
NONE
```


Willyfrog:
@koox00 what's blocking this PR?

app/channel.go
```diff
@@ -2564,6 +2564,8 @@ func (a *App) MarkChannelAsUnreadFromPost(postID string, userID string, collapse
 				}
 				message := model.NewWebSocketEvent(model.WebsocketEventThreadUpdated, channel.TeamId, "", userID, nil)
 				message.Add("thread", string(payload))
+				message.Add("previous_unread_mentions", 0)
+				message.Add("previous_unread_replies", 0)
```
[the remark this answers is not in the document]
koox00: We can skip sending those to denote that it's a newly followed thread, although we don't make use of this info in the client.
Also the client will set those to 0 if they are not provided.

koox00:
> @koox00 what's blocking this PR?

I had forgot about that... thanks @Willyfrog.
I've updated the PR and re-request reviews

app/notification.go
```diff
@@ -613,6 +613,16 @@ func (a *App) SendNotifications(post *model.Post, team *model.Team, channel *mod
...
 
+					previousUnreadMentions := userThread.UnreadMentions
+					previousUnreadReplies := userThread.UnreadReplies - 1
+
+					if mentions.isUserMentioned(uid) {
+						previousUnreadMentions = userThread.UnreadMentions - 1
+					}
+
+					message.Add("previous_unread_mentions", previousUnreadMentions)
+					message.Add("previous_unread_replies", previousUnreadReplies)
```
agnivade, line 624: Can we add an app layer test for this?

agnivade (commented):
@koox00 - I read through the JIRA ticket and having a bit of trouble understanding the core problem here.

From what I understand, "baseline" unreads are at a team-level and not mapped to any thread. BUT if a ws event arrives with the new values, and there are no threads in store, then the increment/decrement logic gets messed up?

My first question here is - since unreads are at a team-level, why is there a dependency on threads here? Isn't that just a team-level counter which gets updated?
Secondly, now the server is just providing a before and after, and all the client does is subtract. So why not send the subtracted value itself?

koox00:
> My first question here is - since unreads are at a team-level, why is there a dependency on threads here? Isn't that just a team-level counter which gets updated?

@agnivade unread mentions of a thread contribute to the team level counter as do channel member counts.
the difference is that we have all channel member values for a user upon load, but that is not the case for threads.
You have to load the threads to get those values.

So the scenario is:
You have x number of unread mentions (at team level),
some of those belong to threads and those threads you don't know yet. 
You view one of those threads in another tab/client, 
so now you've seen y mentions.
In the other tab you get that thread viewed via ws, and it has 0 unread mentions. 
Team level counter should show x - y mentions, but you don't know y.
 
>  Secondly, now the server is just providing a before and after, and all the client does is subtract. So why not send the subtracted value itself?

we need the actual number to display a badge besides the thread item. We still'll need to return both actual and new
so something `unread_mentions: 0`, `new_unread_mentions: -3`?
That will require more changes in webapp code, (it was build with `previous_unread_mentions` even if the server didn't provide them)

app/notification.go
```diff
@@ -613,6 +613,16 @@ func (a *App) SendNotifications(post *model.Post, team *model.Team, channel *mod
 					}
 					message.Add("thread", string(payload))
 
+					previousUnreadMentions := userThread.UnreadMentions
+					previousUnreadReplies := userThread.UnreadReplies - 1
```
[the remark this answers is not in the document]
koox00: At first I thought this would always be >= 1, since we are receiving a new message, 
but that's not the case of poster sending the message AKA `userThread.UnreadReplies == 0`, 
the client handles it, but still the server should never send negative values.

store/sqlstore/thread_store.go
```diff
@@ -853,3 +853,23 @@ func (s *SqlThreadStore) DeleteOrphanedRows(limit int) (deleted int64, err error
 	deleted = rpcDeleted + rptDeleted
 	return
 }
+
+// return number of unread replies and number of unread mentions for a single thread
```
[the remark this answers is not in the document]
koox00: oh, sorry forgot to edit, at first I had created a method that returned both

app/notification.go
koox00: @agnivade can you spy on a method call in go? I'd appreciate some pointers on how to proceed testing this

api4/user_test.go
```diff
@@ -5882,6 +5889,78 @@ func TestThreadSocketEvents(t *testing.T) {
...
+						caught = true
+						data := ev.GetData()
+						var thread model.ThreadResponse
+						jsonErr := json.Unmarshal([]byte(data["thread"].(string)), &thread)
+						require.NoError(t, jsonErr)
+
+						require.Equal(t, float64(0), data["previous_unread_replies"])
+						require.Equal(t, float64(0), data["previous_unread_mentions"])
+						require.Equal(t, int64(0), thread.UnreadReplies)
+						require.Equal(t, int64(0), thread.UnreadMentions)
```
koox00, line 5953: @agnivade, @ashishbhate I'm cannot test this properly. I post a new reply as BasicUser2 and should get `unread_replies == 1`, but it seems it is always changing the `thread.LastViewedAt` for `BasicUser` to a value slightly larger than `reply.CreateAt`, so it never has `unreads`.
Could we be changing the `LastViewedAt` of the thread upon CreatePost? 
OR any other ideas on why this is happening?

agnivade: > Could we be changing the LastViewedAt of the thread upon CreatePost?

Yes, that seems to be what's happening. This line apparently does that in app/notification.go

```go
opts := store.ThreadMembershipOpts{
		Following:             true,
		IncrementMentions:     incrementMentions,
		UpdateFollowing:       updateFollowing,
		UpdateViewedTimestamp: userID == post.UserId, // for the OP, the last viewed timestamp gets updated
		UpdateParticipants:    userID == post.UserId,
	}
	threadMembership, err := a.Srv().Store.Thread().MaintainMembership(userID, post.RootId, opts)
```

I love these tests and the fact that this is catching things like this!

agnivade: I also verified it manually, and the db entry does get updated whenever I reply to a thread.

koox00: oh, totally missed that line!
OK it seems that I can't use `client.CreatePost` to create a post as another user...

koox00: @agnivade, @ashishbhate  I've updated the tests and added a broken test case with some comments to explain the issues.

https://github.com/mattermost/mattermost-server/blob/9aad9602f81e90b8c5bb1a3f5319477fe62a7a2e/api4/user_test.go#L5935-L5946

agnivade (approved):
Nice work!

api4/user_test.go
```diff
@@ -5882,6 +5889,94 @@ func TestThreadSocketEvents(t *testing.T) {
...
+			{&model.Post{ChannelId: th.BasicChannel.Id, Message: "mention reply 2 @" + th.BasicUser.Username, UserId: th.BasicUser2.Id, RootId: rpost.Id}, 2, 1, 3, 2},
+			// the following case is a broken
+			// there are a couple of issues,
+			// the results that passes the test are 0, 2, 0, 2
+			// when they should be 3, 2, 0, 0
+			// which means 1. we are zeroing unread but not mentions for some reason.
+			// and 2. previous unread are zeroed when they shouldn't
+			// not an actual problem, (2.) since to post you have to view the thread first
+			// could be a problem if we have quick replies in the app though (reply from a notification)
+			// {&model.Post{ChannelId: th.BasicChannel.Id, Message: "self reply 2", UserId: th.BasicUser.Id, RootId: rpost.Id}, 3, 2, 0, 0},
```
ashishbhate, line 5946: Shouldn't the post here be created by BasicUser**2**?
i.e.
```suggestion
			// {&model.Post{ChannelId: th.BasicChannel.Id, Message: "self reply 2", UserId: th.BasicUser2.Id, RootId: rpost.Id}, 3, 2, 0, 0},
```

Then the expected result will be (3, 2, 4, 2). 
I don't understand how we're supposed to get (3, 2, 0, 0) in either case.

koox00: since it's the current user posting the unread should not increment. 
so it's either 3, 2, 0, 0 or 3, 2, 3, 2(?)

api4/user_test.go
koox00: I tried to fix these 2 issues, the tests pass but can you re-review @agnivade, @ashishbhate?

store/storetest/thread_store.go
```diff
@@ -470,6 +470,35 @@ func testThreadStorePopulation(t *testing.T, ss store.Store) {
...
+		require.NoError(t, err1)
+
+		unreads, err := ss.Thread().GetThreadUnreadReplyCount(m)
+		require.NoError(t, err)
+		require.Equal(t, int64(0), unreads)
+
+		err = ss.Thread().MarkAsRead(newPosts[0].UserId, newPosts[0].Id, newPosts[0].CreateAt)
+		require.NoError(t, err)
+		m, err = ss.Thread().GetMembershipForUser(newPosts[0].UserId, newPosts[0].Id)
+		require.NoError(t, err)
```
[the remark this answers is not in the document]
koox00: yes, to the get the new counts.

agnivade:
Seems like tests are failing. Feel free to ping me once they pass.

koox00:
oof, finally tests pass... can you please re-review @ashishbhate, @agnivade?

app/channel.go
```diff
@@ -2538,36 +2538,44 @@ func (a *App) MarkChannelAsUnreadFromPost(postID string, userID string, collapse
...
-				}
+			threadMembership.UnreadMentions, err = a.countThreadMentions(user, post, channel.TeamId, post.UpdateAt-1)
+			if err != nil {
+				return nil, err
+			}
+			_, nErr = a.Srv().Store.Thread().UpdateMembership(threadMembership)
+			if nErr != nil {
+				return nil, model.NewAppError("MarkChannelAsUnreadFromPost", "app.channel.update_last_viewed_at_post.app_error", nil, nErr.Error(), http.StatusInternalServerError)
+			}
+			if opts.UpdateFollowing {
```
koox00, line 2578: It seemed to me that we didn't update unread mentions when we mark a thread as unread from a post
(we did that only when there was no thread membership).
At least unit tests showed that much. I don't know at which point we fix this when returning the thread to the user,
because it seems that the user receives the correct amount of mentions...
So I'm re-calculating mentions and updating the `ThreadMembership` regardless of follow state,
but send a WS socket event only when the thread is to be followed.

Anyway I just want to draw your attention upon this change.

agnivade:
@jupenur / @srkgupta - Drawing your attention to some more unrelated security alerts. I didn't want to ignore them by myself. Please take a look. Thanks :)

srkgupta:
@agnivade will discuss this offline. Thanks

api4/user_test.go
```diff
@@ -5882,6 +5919,58 @@ func TestThreadSocketEvents(t *testing.T) {
...
+	// read the thread
+	_, resp, err = th.Client.UpdateThreadReadForUser(th.BasicUser.Id, th.BasicTeam.Id, rpost.Id, replyPost.CreateAt+1)
+	require.NoError(t, err)
+	CheckOKStatus(t, resp)
+
+	t.Run("Listen for thread updated event after create post", func(t *testing.T) {
+		testCases := []struct {
+			post        *model.Post
+			preReplies  float64
+			preMentions float64
```
agnivade, line 5931: Why this is in float and the others are in int64?

api4/user_test.go
```diff
@@ -5882,6 +5919,58 @@ func TestThreadSocketEvents(t *testing.T) {
...
+			var caught bool
+			func() {
+				for {
+					select {
+					case ev := <-userWSClient.EventChannel:
+						if ev.EventType() == model.WebsocketEventThreadUpdated {
+							caught = true
+							data := ev.GetData()
+							var thread model.ThreadResponse
+							jsonErr := json.Unmarshal([]byte(data["thread"].(string)), &thread)
```
agnivade, line 5957: Is there a need for this double-conversion? Could we directly cast to `[]byte`?

api4/user_test.go
```diff
@@ -6164,27 +6261,27 @@ func TestMaintainUnreadMentionsInThread(t *testing.T) {
...
 	// test DM
 	dm := th.CreateDmChannel(th.SystemAdminUser)
 	dm_root_post, _ := postAndCheck(t, client, &model.Post{ChannelId: dm.Id, Message: "hi @" + th.SystemAdminUser.Username})
 
 	// no changes
-	checkThreadList(th.Client, th.BasicUser.Id, 1, 1)
+	checkThreadList(th.Client, th.BasicUser.Id, 0, 1)
 
-	// post reply by the same user
+	// post reply by the BasicUser
```
agnivade, line 6274: If it's a different user, shouldn't the count increase? Same for below as well.

app/channel.go
agnivade: It looks like the only place where this is called from is `setPostUnread` in api4/post.go, and it always sets `followThread` to be false. Which means `else if !threadMembership.Following && followThread` never evaluates to true.

So in that case, we will be calling `MaintainMembership` with default opts rather than not calling it at all, in the original scenario. Is that correct?

app/notification.go
```diff
@@ -598,6 +598,32 @@ func (a *App) SendNotifications(post *model.Post, team *model.Team, channel *mod
...
+						}
+						// should set unread mentions, and unread replies to 0
+						um, err := a.Srv().Store.Thread().MaintainMembership(uid, post.RootId, opts)
+
+						if err != nil {
+							return nil, errors.Wrapf(err, "cannot maintain thread membership %q for user %q", post.RootId, uid)
+						}
+
+						userThread.UnreadMentions = um.UnreadMentions                                        // should be 0
+						userThread.UnreadReplies, err = a.Srv().Store.Thread().GetThreadUnreadReplyCount(um) // should be 0
```
agnivade, line 621: If they should be 0, why are we making the store call? Can we not set to 0, and write a test to lock the behavior?

app/notification.go
```diff
@@ -249,7 +249,7 @@ func (a *App) SendNotifications(post *model.Post, team *model.Team, channel *mod
 					Following:             true,
 					IncrementMentions:     incrementMentions,
 					UpdateFollowing:       updateFollowing,
-					UpdateViewedTimestamp: userID == post.UserId,
+					UpdateViewedTimestamp: false,
```
agnivade, line 252: It seems like we are setting it to false here, and then again making another call later if `uid == post.UserId`. Can you clarify why is that?

agnivade (commented):
Left some comments, but I am a bit concerned that the overall logic is getting hard to understand from all flows. Especially the app/notification.go file needs more refactoring. 
Otherwise, you will have to keep fixing bugs. I'd suggest to set aside some time next year to do a proper refactor and add more tests to clearly lay out everything.

api4/user_test.go
koox00: because `Json.Unmarshal` returns `float64`, I can cast them to `int64` so it won't be confusing

api4/user_test.go
koox00: no sorry, `BasicUser` is the same user. I'll add that to the comment

app/notification.go
koox00: I was thinking about that, and wasn't sure if it will be prone to error or not, I'll try to break a test to see if we are safe and then change UnreadReplies to 0 instead of the store call

app/notification.go
koox00: Having this made previous unread replies and mentions for the commenter be always 0, so I moved the "marking the thread as viewed" after getting previous unread counts.

api4/user_test.go
koox00: It asks for type assertion, `cannot convert data["thread"] (type interface {}) to type []byte: need type assertion`.

koox00:
> It looks like the only place where this is called from is `setPostUnread` in api4/post.go, and it always sets `followThread` to be false. Which means `else if !threadMembership.Following && followThread` never evaluates to true.
> 
> So in that case, we will be calling `MaintainMembership` with default opts rather than not calling it at all, in the original scenario. Is that correct?

@agnivade oh, you're right, what threw me off was that in tests we were calling it with `followThread == true`.
OK I refactored a bit so it won't update the membership when we are not going to follow the thread but also support `followThread` correctly.

app/notification.go
koox00: done, `TestThreadSocketEvents` cover this behavior, and will break if something changes

koox00:
> Left some comments, but I am a bit concerned that the overall logic is getting hard to understand from all flows. Especially the app/notification.go file needs more refactoring.
> Otherwise, you will have to keep fixing bugs. I'd suggest to set aside some time next year to do a proper refactor and add more tests to clearly lay out everything.

I completely agree with you on this. Apollo team already has a ticket for refactoring `app/notification.go`

api4/user_test.go
agnivade: You are right, the underlying value is a string. My bad there.

agnivade:
Great stuff! Looks good to me. The PR already has my approval.

ashishbhate (approved):

jgilliam17:
Thanks @koox00 and sorry about delay
Tested manually, there is an issue with replies not posting. This is not related to your PR, I've seen this one few other and asked Carlos if he knows how to correct this. Seems to be a server configuration issue.
Here is the [E2E report ](https://mattermost-cypress-report.s3.amazonaws.com/105156-1d0c164-onprem-ent-server-pr-18492-mm-ee-test:1d0c164/mochawesome.html), failures do not appear to be related to your changes 
These are the tests that failed/differ from the master report for visibility
- focuses user email input by default
- onboarding (1) and onboarding sysadmin (6)

jgilliam17:
@koox00 
I created a test server here, since the one created from build server for PR label still has reply post issue.
https://pr18492.test.mattermost.cloud
Seeing one issue on refresh (see video). 
Threads count goes from 2 to 4 after refresh, and drops back to 2 when user opens global threads, this is all happening without opening to read individual threads. 

https://user-images.githubusercontent.com/52937121/150590852-5e423cae-fe31-48a5-8105-c787a56f0987.mov

jgilliam17:
@koox00 Something is still off with the mention count
Steps: 
Receive few mentions on threads in Town Square, open global threads but do not read the unreads, switch team and back
Switch to Town Square, observe the number on global threads,  - number changes, increases 
See video

https://user-images.githubusercontent.com/52937121/152390570-a5f0386c-c38c-4518-87a2-d2fa2a57a4e7.mov

koox00:
@jgilliam17 I've found the issue for this... Currently we don't have any means to differentiate a newly created thread arriving via websocket or an updated existing one... In this context we treat both as new threads adding to the global counts.
Which creates this inconsistency. I'll open another ticket to solve it and then we can re-test this one.

koox00:
@jgilliam17 can you re-test please?

jgilliam17 (approved):
Thanks @koox00 
Tested, looks good to merge.
- Verified when there are unreads and/or mentions, correct unread counts are loaded after refresh.
- [E2E report](https://mattermost-cypress-report.s3.amazonaws.com/148695-316a7ac-onprem-ent-server-pr-18492-mattermost/mm-ee-test316a7ac/mochawesome.html) looks good. There were few(5) failures that differ from master - retested manually and all are passing: should switch channels when clicking on a channel in the sidebar, MM-T1466, MM-T613, MM-T87, MM-T1100

PR #18496, opened 2021-09-30
MM-38624 Migrate ChannelMemberHistoryStore to sqlx

pikami:
#### Summary
Migrate ChannelMemberHistoryStore to sqlx

#### Ticket Link
Fixes https://github.com/mattermost/mattermost-server/issues/18408

#### Release Note
```release-note
NONE
```

### DIFF
--- a/store/sqlstore/channel_member_history_store.go
+++ b/store/sqlstore/channel_member_history_store.go
@@ -120,7 +120,7 @@ func (s SqlChannelMemberHistoryStore) hasDataAtOrBefore(time int64) (bool, error
 
 func (s SqlChannelMemberHistoryStore) getFromChannelMemberHistoryTable(startTime int64, endTime int64, channelId string) ([]*model.ChannelMemberHistoryResult, error) {
 	query, args, err := s.getQueryBuilder().
-		Select("cmh.*, u.Email, u.Username, Bots.UserId IS NOT NULL AS IsBot, u.DeleteAt AS UserDeleteAt").
+		Select(`cmh.*, u.Email AS "Email", u.Username, Bots.UserId IS NOT NULL AS IsBot, u.DeleteAt AS UserDeleteAt`).
 		From("ChannelMemberHistory cmh").
 		Join("Users u ON cmh.UserId = u.Id").
 		LeftJoin("Bots ON Bots.UserId = u.Id").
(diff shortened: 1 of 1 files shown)


agnivade:
It is failing with `missing destination name email in *[]*model.ChannelMemberHistoryResult` only for Postgres. We do have the `db` struct tag 

```
UserEmail    string `db:"Email"`
```

in the `ChannelMemberHistoryResult` struct.

@pikami - Could you investigate further into it?

pikami:
Thanks for the info @agnivade I'll check it out tomorrow

pikami:
> It is failing with `missing destination name email in *[]*model.ChannelMemberHistoryResult` only for Postgres. We do have the `db` struct tag
> 
> ```
> UserEmail    string `db:"Email"`
> ```
> 
> in the `ChannelMemberHistoryResult` struct.
> 
> @pikami - Could you investigate further into it?

Looks like the email column that is being returned from PostgreSQL is in lowercase ('email') and in MySQL the first letter is uppercase ("Email")
I've fixed this by adding `AS "Email"` statement so it would return the same column name for both DB implementations.

store/sqlstore/channel_member_history_store.go
```diff
@@ -154,7 +157,7 @@ func (s SqlChannelMemberHistoryStore) getFromChannelMembersTable(startTime int64
 	}
 
 	var histories []*model.ChannelMemberHistoryResult
```
agnivade, line 159: This needs to be `histories := []*model.ChannelMemberHistoryResult{}`

store/sqlstore/channel_member_history_store.go
```diff
@@ -134,7 +137,7 @@ func (s SqlChannelMemberHistoryStore) getFromChannelMemberHistoryTable(startTime
 		return nil, errors.Wrap(err, "channel_member_history_to_sql")
 	}
 	var histories []*model.ChannelMemberHistoryResult
```
agnivade, line 139: Same as below.

agnivade (commented):
Looking good @pikami ! Just some minor things to take care of.

agnivade:
Hey @pikami - checking to see how is it going with this. Just some small things to take care of and it should be good to go.

PR #18529, opened 2021-10-04
MM-37934: supports channel override for CRT notifications

koox00:
#### Summary

Users with CRT 'ON' are enabled to override global CRT notification
settings per channel basis.
#### Ticket Link

https://mattermost.atlassian.net/browse/MM-37934

#### Release Note

```release-note
NONE
```

### DIFF
--- a/app/notification.go
+++ b/app/notification.go
@@ -329,7 +329,7 @@ func (a *App) SendNotifications(post *model.Post, team *model.Team, channel *mod
 			}
 
 			// add user id to notificationsForCRT depending on threads notify props
-			notificationsForCRT.addUserToNotify(profile, mentions)
+			notificationsForCRT.addFollowerToNotify(profile, mentions, channelMemberNotifyPropsMap[profile.Id], channel)
 		}
 	}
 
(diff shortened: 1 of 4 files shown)


model/channel_member.go
```diff
@@ -156,6 +156,27 @@ func IsIgnoreChannelMentionsValid(ignoreChannelMentions string) bool {
 	return ignoreChannelMentions == IgnoreChannelMentionsOn || ignoreChannelMentions == IgnoreChannelMentionsOff || ignoreChannelMentions == IgnoreChannelMentionsDefault
 }
 
+func ShouldChannelMemberNotifyCRT(notifyProps StringMap, isMentioned bool) (bool, bool) {
```
agnivade, line 159: Let's make this a private function inside app/notification.go. There's no need to expose this in model.

Also, let's use named returns `notifyDesktop, notifyPush bool` for readability.

model/user.go
```diff
@@ -802,6 +802,34 @@ func (u *UserPatch) SetField(fieldName string, fieldValue string) {
 	}
 }
 
+func (u *User) ShouldNotifyCRT(isMentioned bool) (bool, bool, bool) {
```
agnivade, line 805: Same as above. Let's move this inside app/notification.go and make it a private function.

app/notification.go
```diff
@@ -1338,28 +1338,31 @@ type CRTNotifiers struct {
...
-	desktopThreads := user.NotifyProps[model.DesktopThreadsNotifyProp]
-	emailThreads := user.NotifyProps[model.EmailThreadsNotifyProp]
-	pushThreads := user.NotifyProps[model.PushThreadsNotifyProp]
+	notifyDesktop, notifyPush, notifyEmail := user.ShouldNotifyCRT(userWasMentioned)
 
-	_, userWasMentioned := mentions.Mentions[user.Id]
+	notifyChannelDesktop, notifyChannelPush := model.ShouldChannelMemberNotifyCRT(channelMemberNotificationProps, userWasMentioned)
 
-	if desktop != model.UserNotifyNone && (userWasMentioned || desktopThreads == model.UserNotifyAll || desktop == model.UserNotifyAll) {
+	if channelMemberNotificationProps[model.DesktopNotifyProp] == model.ChannelNotifyDefault {
```
agnivade, line 1349: I think the 3 slice populations can be simplified further. Firstly, the action in both conditions are same, so they can be shortened to:

```go
if (channelMemberNotificationProps[model.DesktopNotifyProp] == model.ChannelNotifyDefault && notifyDesktop) || notifyChannelDesktop {
  c.Desktop = append(c.Desktop, user.Id)
}
```

Secondly, this entire logic can be moved inside `shouldChannelMemberNotifyCRT`. This method can just append `c.Desktop`, `c.Email` and `c.Push`.

Also, let's write some comments to explain the code in `ShouldChannelMemberNotifyCRT` once the code is moved inside that. I am finding it hard to understand so many if conditions.

koox00: > I think the 3 slice populations can be simplified further. Firstly, the action in both conditions are same, so they can be shortened to:

It's not the same, the rule is if the channel notify prop is set to default respect **only** the user global setting,
in any other case respect the channel member setting.

koox00: > Secondly, this entire logic can be moved inside `shouldChannelMemberNotifyCRT`. This method can just append `c.Desktop`, `c.Email` and `c.Push`.

Since the logic is somewhat complicated having 2 different checks for notify props, and one that selects between the both was easier for me to reason about.

agnivade: > It's not the same, the rule is if the channel notify prop is set to default respect only the user global setting,
in any other case respect the channel member setting.

Ah my bad. Then let's just merge the first 2 conditions.

koox00: > Ah my bad. Then let's just merge the first 2 conditions.

you mean these 2?

```go
channelMemberNotificationProps[model.DesktopNotifyProp] == model.ChannelNotifyDefault && notifyDesktop
```

agnivade: Yep.

koox00: done

agnivade (approved):
Looks very readable now. Thanks @koox00 !

app/notification.go
agnivade: I see that you actually merged all the conditions. Did something change?

jgilliam17:
Thanks @koox00 
I am having some issues with disabling thread notifications on the channel.
for example:
Logged in as user-1 on the test server
When global notification setting is set to send notifications for "threads I'm following"
 
Screen Shot 2021-10-19 at 4 05 42 PM

Off-topic notification preference for the channel is set on only mentions and dms, and threads unchecked

Screen Shot 2021-10-19 at 4 05 26 PM

The notifications are still sent for the thread replies in this channel.

PR #18540, opened 2021-10-05
Remove searchEngine from App

agnivade:
```release-note
NONE
```


agnivade:
I was trying to actually fix this, then found that there is actually nothing to fix. Something must have changed last time I was looking at this.

streamer45 (approved):

PR #18525, opened 2021-10-04
MM-35847 inline style for email templates

koox00:
#### Summary

Some email clients don't support the style element,
this commit changes `mjml-style` to generate inline styles instead of a
internal style-sheet.

#### Ticket Link

https://mattermost.atlassian.net/browse/MM-35847

#### Release Note

```release-note
NONE
```

PR #18541, opened 2021-10-05
[MM-39072] - Run make build-templates on latest master

AGMETEOR:
#### Summary
Some code from the style.mjml partial was not built into some generated templates when `make build-templates` is run. This means there's a diff when `make build-templates` is run on master, which shouldn't be the case.

From the diff files, only the invitation template has changes that are not just CSS classes. However, it still renders as before

Screenshot 2021-10-05 at 12 06 18

The guest email template was failing as a result of this change. This was expected due to a change in the past that updated the card.mjml partial for messages notification like mentions and batch email but didn't consider the fact that guest invitation templates also use it.

#### Ticket Link
[MM-39072](https://mattermost.atlassian.net/browse/MM-39072)

#### Release Note
```release-note
NONE

```

### DIFF
--- a/app/email/email.go
+++ b/app/email/email.go
@@ -515,11 +515,6 @@ func (es *Service) SendGuestInviteEmails(team *model.Team, channels []*model.Cha
 			data.Props["Title"] = i18n.T("api.templates.invite_body.title", map[string]interface{}{"SenderName": senderName, "TeamDisplayName": team.DisplayName})
 			data.Props["SubTitle"] = i18n.T("api.templates.invite_body_guest.subTitle")
 			data.Props["Button"] = i18n.T("api.templates.invite_body.button")
-			data.Props["SenderName"] = senderName
-			data.Props["Message"] = ""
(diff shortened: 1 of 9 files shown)


koox00 (approved):

AGMETEOR:
Hi @koox00 so maybe I can merge this first and then this https://github.com/mattermost/mattermost-server/pull/18525 goes in after

koox00:
> Hi @koox00 so maybe I can merge this first and then this #18525 goes in after

yes!

AGMETEOR:
Hi @stevemudie . Requesting a QA review because there was some changes made to the invite template. Both members and guest invitations should be QA tested

stevemudie (approved):
Looks good @AGMETEOR!

Both Guest and Member invite emails land looking as they should.

marianunez (approved):

agnivade:
Would you like it if I add a CI check for this so that it gets caught much earlier?

PR #18571, opened 2021-10-06
MM-38497: Fix Sentry crash in PostAction.Equals

agnivade:
We check for nil pointer before moving ahead.

https://mattermost.atlassian.net/browse/MM-38497

```release-note
NONE
```


streamer45 (approved):
Looks good :+1:

PR #18623, opened 2021-10-09
Move Channels into App

agnivade:
In this PR, we make Channels as part of App
instead of Server. This is part of the transition period
of moving fields from Server to Channels.

For now, Channels contains Server. So the hierarchy is

App -> Channels -> Server.

And as a first step, we also move httpService to Channels.


agnivade:
To be merged after https://github.com/mattermost/mattermost-server/pull/18611.

PR #15824, opened 2020-10-06
[MM-28694] Add MM User information in the CWS request headers

ethervoid:
#### Summary

This PR includes changes made [here]() to refactor the code and remove App dependency and change from AppError to Error as suggested by @jespino 

#### Related PRs

https://github.com/mattermost/enterprise/pull/794

#### Ticket Link

https://mattermost.atlassian.net/browse/MM-28694

```release-note
NONE
```

PR #18668, opened 2021-10-13
Move cloud init back to initEnterprise

agnivade:
```release-note
NONE
```

### DIFF
--- a/app/server.go
+++ b/app/server.go
@@ -464,14 +464,6 @@ func NewServer(options ...Option) (*Server, error) {
 
 	})
 
-	// This enterprise init should happen after the store is set
-	// but we don't want to move the s.initEnterprise() call because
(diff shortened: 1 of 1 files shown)


marianunez:
> Could you suggest some tests around this to validate it?
> 
> This was changed in #15824 by Mario. But I am failing to see the issue here.

I don't have the context on why the change from App to Server there TBH. However, regarding testing there are unit tests here https://github.com/mattermost/enterprise/blob/8a6c9eee545318b86be836166bfe7723f84bc0c1/cloud/cloud_test.go and also, for manual test, you can setup the label `Setup Cloud + CWS` and smoke test that the System Console  > Billing area is showing properly (no white screens). That's the area that uses the Enterprise cloud endpoints

PR #18676, opened 2021-10-13
[MM-35863] Add IsEnterpriseReady Plugin API

Zeezee1210:
#### Summary
This PR exposes the model.BuildEnterpriseReady value to Plugins through an API. 
Plugins can now call IsEnterpriseReady()

#### Ticket Link
Fixes https://github.com/mattermost/mattermost-server/issues/17904

#### Release Note

```release-note
Adds new Plugin API IsEnterpriseReady()
```

### DIFF
--- a/app/plugin_api_test.go
+++ b/app/plugin_api_test.go
@@ -1924,10 +1924,13 @@ func TestPluginAPIUpdateCommand(t *testing.T) {
 }
 
 func TestPluginAPIIsEnterpriseReady(t *testing.T) {
-	model.BuildEnterpriseReady = "false"
+	oldValue := model.BuildEnterpriseReady
+	defer func() { model.BuildEnterpriseReady = oldValue }()
+
+	model.BuildEnterpriseReady = "true"
 	th := Setup(t)
 	defer th.TearDown()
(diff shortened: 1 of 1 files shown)


Zeezee1210:
> Awesome, thank you @Zeezee1210! Could I trouble you to add a unit test to https://github.com/mattermost/mattermost-server/blob/master/app/plugin_api_test.go that exercises this case? It think it can be as simple as verifying that calling this API returns the known value of `model.BuildEnterpriseReady` -- it's probably not possible to test all cases directly in a const value like this.

Hey @lieut-data,
I hope the test I added is what you meant, I couldn't see a way to edit the model other that to set it directly but if I missed this or there's a better way then please let me know.

lieut-data:
@agarciamontoro, I vaguely recall your needing this API for /something/ we built on Playbooks. If you recall the specifics, would you be willing to give this PR a whirl with that plugin and confirm the intended behaviour?

emilyacook:
Thanks for participating in Hacktoberfest! You can claim your sticker set here: https://get.printfection.com/hacktober21/4144583267 @Zeezee1210

PR #18734, opened 2021-10-18
Fix race condition in telemetry

larkox:
#### Summary
Use `InsertIfExist` (the naming is misleading) to fix race conditions on the setting of the telemetry ID.

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-39343

#### Release Note
```release-note
Fix race condition in telemetry IDs on High Availability scenarios.
```

### DIFF
--- a/api4/cors_test.go
+++ b/api4/cors_test.go
@@ -127,11 +127,8 @@ func TestCORSRequestHandling(t *testing.T) {
 				*cfg.ServiceSettings.CorsAllowCredentials = testcase.CorsAllowCredentials
 			})
 			defer th.TearDown()
-			systemStore := mocks.SystemStore{}
-			systemStore.On("Get").Return(make(model.StringMap), nil)
 			licenseStore := mocks.LicenseStore{}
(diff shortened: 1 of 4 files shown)


amyblais:
> @amyblais this is good. But we should be back porting this fix all the way to the latest ESR and every version in between.
> 
> This won't be used until mobile v2 is out, so back porting is not urgent and we should determine to which servers is this fix needed.
> 
> On the other hand this could be causing lots of noise for telemetry in current HA environments so we should take that into consideration.

@enahum Do you know who would make the decision about what versions are needed, and to look into the telemetry noice?

jasonblais:
My two cents: We should fix and backport it, but I wouldn't hold back the v6.0.1 release for this as we have a few customers waiting for that release.

Plus, we'll likely need more consideration to determine what versions we want this to be backported to, including the note from Nelson above.

amyblais:
> @amyblais This is merged now, but no milestone assigned to it yet, so not sure if and where should I be cherrypicking it. Also, there is the open discussion on to what versions should we apply this.

@larkox This can be backported to cloud, 5.37, 5.38, 5.39, and 6.0.

PR #18755, opened 2021-10-19
ThreadStore: Remove quadratic complexity for participants population

agnivade:
While populating the participants, we would traverse the global list
twice to get the users for that thread. For threads with large number
of users, this can take considerable time. We replace that with a map
to change it to O(n).

While here, we improve memory usage of this code by pre-allocating
slices and maps. And also add some comments to help readers.

https://community-daily.mattermost.com/boards/workspace/zyoahc9uapdn3xdptac6jb69ic/285b80a3-257d-41f6-8cf4-ed80ca9d92e5/495cdb4d-c13a-4992-8eb9-80cfee2819a4?c=c9aa17f3-b918-4a78-86aa-01562e418634

```release-note
NONE
```


agnivade:
@ashishbhate - gentle ping :)

PR #18756, opened 2021-10-19
Fix migration scripts

agnivade:
Set the DB schema version in DB prior to running migration

https://community-daily.mattermost.com/boards/workspace/zyoahc9uapdn3xdptac6jb69ic/285b80a3-257d-41f6-8cf4-ed80ca9d92e5/495cdb4d-c13a-4992-8eb9-80cfee2819a4?c=c2525dbb-23eb-4fe7-8fb4-e9a5f503dc8c

```release-note
NONE
```

### DIFF
--- a/scripts/mattermost-postgresql-6.0.0.sql
+++ b/scripts/mattermost-postgresql-6.0.0.sql
@@ -112,7 +112,7 @@ CREATE TABLE public.channels (
     groupconstrained boolean,
     shared boolean,
     totalmsgcountroot bigint,
-    lastrootpostat bigint DEFAULT '0'::bigint,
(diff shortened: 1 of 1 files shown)


agnivade:
While testing the script, I can see ordering issues with the schema dump (in both MySQL and Postgres), inspite of both dump commands remaining same. I think we are seeing this now because the script is actually working for the first time.

diff
Would you be able to take a look?

streamer45:
@agnivade  This should be ready now. I am not entirely convinced by the default on `Channels.LastRootPostAt` given we don't really use defaults elsewhere but that's a separate issue.

agnivade:
Looks good. You'll have to approve yourself :)

PR #18794, opened 2021-10-21
[MM-39448] Implement auditUserPatch

streamer45:
#### Summary

Adding conversion for `model.UserPatch` as well that was completely missing.

#### Ticket Link

https://mattermost.atlassian.net/browse/MM-39448

#### Release Note

```release-note
NONE
```


streamer45:
> This will start to raise maintenance problems in the future unless we start auto-generating this code. Or do something else. I don't fully understand yet what the code does.

Absolutely, this is just a hot-fix. We'll need to plan a proper fix accordingly cause it's so easy to miss new potentially sensitive objects this way. 

From my understanding the code is just a convenient function that converts (and sanitizes) our model structs to make them more easily loggable.

PR #18795, opened 2021-10-21
MM-36862: removes participant from thread

koox00:
#### Summary

Removing a participant upon last reply deleted from a thread didn't work
reliably, a suspect on this is replica lagging... Since we are first
deleting the post and then counting non deleted posts of the participant
to decide on whether to delete or not.

The findings that led to this conclusion is that the reply count gets
updated but the participant is not removed (participant removal depends
on the number of replies this participant has in the thread.)

This commit fixes that by consistently first deleting the post and then
removing the participant, and makes this whole operation transactional.

Furthermore it fixes cleaning up threads when we delete posts by user.
#### Ticket Link

https://mattermost.atlassian.net/browse/MM-36862
https://mattermost.atlassian.net/browse/MM-40116

#### Release Note

```release-note
NONE
```


stylianosrigas:
FYI The spinwick fails with the following error. Removing it and feel free to add it again:
```
{"timestamp":"2021-11-08 12:04:59.106 Z","level":"fatal","msg":"Failed to apply database migrations.","error":"try lock failed in line 0: SELECT pg_advisory_lock($1) (details: pq: unnamed prepared statement does not exist)","stacktrace":[{"Function":"github.com/mattermost/mattermost-server/v6/store/sqlstore.New","File":"github.com/mattermost/mattermost-server/v6/store/sqlstore/store.go","Line":162},{"Function":"github.com/mattermost/mattermost-server/v6/app.NewServer.func2","File":"github.com/mattermost/mattermost-server/v6/app/server.go","Line":369},{"Function":"github.com/mattermost/mattermost-server/v6/app.NewServer","File":"github.com/mattermost/mattermost-server/v6/app/server.go","Line":418},{"Function":"github.com/mattermost/mattermost-server/v6/cmd/mattermost/commands.runServer","File":"github.com/mattermost/mattermost-server/v6/cmd/mattermost/commands/server.go","Line":74},{"Function":"github.com/mattermost/mattermost-server/v6/cmd/mattermost/commands.serverCmdF","File":"github.com/mattermost/mattermost-server/v6/cmd/mattermost/commands/server.go","Line":58},{"Function":"github.com/spf13/cobra.(*Command).execute","File":"github.com/spf13/cobra@v1.2.1/command.go","Line":856},{"Function":"github.com/spf13/cobra.(*Command).ExecuteC","File":"github.com/spf13/cobra@v1.2.1/command.go","Line":974},{"Function":"github.com/spf13/cobra.(*Command).Execute","File":"github.com/spf13/cobra@v1.2.1/command.go","Line":902},{"Function":"github.com/mattermost/mattermost-server/v6/cmd/mattermost/commands.Run","File":"github.com/mattermost/mattermost-server/v6/cmd/mattermost/commands/root.go","Line":14},{"Function":"main.main","File":"github.com/mattermost/mattermost-server/v6/cmd/mattermost/main.go","Line":31},{"Function":"runtime.main","File":"runtime/proc.go","Line":225},{"Function":"runtime.goexit","File":"runtime/asm_amd64.s","Line":1371}]}
```

koox00:
> EDIT: Actually, it looks like the permanentDelete methods were already broken since they were calling cleanUpThreads before, and then deleting the post. So participants would never get removed. In any case, I think the non-transactional problem still remains and would be better to have a proper fix.

`cleanupThreads` when called from permanent delete actions just deleted threads, and thread memberships, and did nothing when called on comments and not root posts.

I tried to fix that by some minor refactoring, just mentioning to also give attention to that when reviewing, thank you!

store/sqlstore/post_store.go
```diff
@@ -2411,12 +2464,10 @@ func (s *SqlPostStore) cleanupThreads(postId, rootId string, permanent bool, use
 			if s.DriverName() == model.DatabaseDriverPostgres {
 				updateQuery = updateQuery.Set("Participants", sq.Expr("Participants - ?", userId))
 			} else {
-				// The .Where is because JSON_REMOVE returns null if the element to remove wasn't present
 				updateQuery = updateQuery.
 					Set("Participants", sq.Expr(
-						`JSON_REMOVE(Participants, JSON_UNQUOTE(JSON_SEARCH(Participants, 'one', ?)))`, userId,
-					)).
-					Where(sq.Expr(`JSON_CONTAINS(Participants, ?)`, strconv.Quote(userId)))
```
koox00, line 2419: I had to remove the `WHERE` clause to support calling the cleanup from a loop in the same transaction (the case of `deleteAllCommentsByUser`.

Case:

- User has made 2 comments in a thread.
- We permanent delete all posts by User.
- `deleteAllCommentByUser` deletes at first all comments and then
- `cleanupThreads` is called in a loop for each of those 2 comments.
- The first time it will remove User from participants and set ReplyCount -1.
- The second time the Where clause would have failed and the Thread would not update
- resulting in a ReplyCount that's not correct.

By just removing the Where clause in mysql Participants were set to `null` which resulted in a string to null conversion error.
Adding `IFNULL` fixes that.

store/sqlstore/post_store.go
```diff
@@ -2411,12 +2464,10 @@ func (s *SqlPostStore) cleanupThreads(postId, rootId string, permanent bool, use
 			if s.DriverName() == model.DatabaseDriverPostgres {
```
agnivade, line 2464: Wanted to comment in L2461, but GitHub won't let me.

It would be good to add a comment in the lines of "Updating replyCount and reducing participants if this was the last post in the thread for the user" to explain what is going on here.

store/sqlstore/post_store.go
agnivade: Excellent catch, I missed this in the original code review.

store/sqlstore/post_store.go
```diff
@@ -734,21 +759,33 @@ type postIds struct {
...
 
+	_, err = transaction.Exec("DELETE FROM Posts WHERE UserId = :UserId AND RootId != ''", map[string]interface{}{"UserId": userId})
+
+	if err != nil {
+		return errors.Wrapf(err, "failed to delete Posts with userId=%s", userId)
+	}
+
 	for _, ids := range results {
-		if err = s.cleanupThreads(ids.Id, ids.RootId, true, userId); err != nil {
+		if err = s.cleanupThreads(transaction, ids.Id, ids.RootId, true, userId); err != nil {
```
agnivade, line 780: So now, it seems like we have a proper separation between `cleanupThreads` and `permanentDeleteThreads`. And in this case, it looks like we are always querying for `RootId != ''` which means the `permanent && rootId == ""` condition will always evaluate to false.

Wondering if we can totally remove the inner condition from `cleanupThreads` and rename it to `cleanupThreadComments`? That way we can get rid of the `permanent` parameter and always either call `cleanupThreadComments` or `permanentDeleteThreads`. Because this seems to be the only place where `cleanupThreads` is called with permanent as true.

Would that make sense?

agnivade (commented):
Great work Kyriakos. Left some comments.

store/sqlstore/post_store.go
koox00: I've made those changes.
So to clarify on functionality.

- Deleting a post deletes the post and then cleans the threads if it's a comment, and does nothing for root posts. (there is another ticket about this)
- Permanent deletion of user's posts deletes all comments by user and cleans up those threads and then deletes all root posts by user and deletes threads and thread memberships for those root posts
- Permanent deletion of posts in channel, deletes all channel posts, comments and root posts, and then deletes all threads and thread memberships (no thread cleanup here, since we are deleting threads and thread memberships)

agnivade: Great stuff. Would be even better if you want to add this as a comment somewhere - maybe above `cleanupThreadComments` or `permanentDeleteThreads`. 0/5.

agnivade (approved):
Really thorough and fantastic job there @koox00 ! Appreciate that you took the time and effort :+1:

agnivade:
@koox00 - Did you mean to ask for review again? :)

Btw, one thing that I forgot to mention earlier is that now the channelDelete and userDelete is running in transactions now. And the problem there is that a loop runs through each post cleaning up the thread comments which makes things tremendously slow if you have 10000 posts in a channel. It was already like this, but now with transactions, those rows will be locked. If we are deleteing the user/channel, that shouldn't make much of an impact, but we will see if someone complains :)

Overall, we need to do that in batches rather than one-by-one. Would you mind creating a HW ticket for this?

koox00:
> @koox00 - Did you mean to ask for review again? :)

oh, I didn't realize, it was by mistake 😅 

> Overall, we need to do that in batches rather than one-by-one. Would you mind creating a HW ticket for this?

will do yes!

ashishbhate (approved):

jgilliam17 (approved):
Thank you @koox00 
Tested, looks good to merge.
- Verified participant's avatars are removed from the thread footer when their post is deleted, reply count updates as expected.
Issue with updating reply at time and removing root post without replies (all deleted) from global threads will be filed separately.

koox00:
> Overall, we need to do that in batches rather than one-by-one. Would you mind creating a HW ticket for this?

Done, https://mattermost.atlassian.net/browse/MM-40483

PR #18801, opened 2021-10-22
GH-18288: Allow configuring unsafe-eval and unsafe-inline

SharathHuddar:
#### Summary
Added a new config setting - ServiceSettings.DeveloperFlags. This is a list of strings and each string is a flag which is used to set the content security policy for the webapp. Each flag has the following format -> `flag=true`. The `flag` is added to the CSP only when it is set to `true`. 

#### Ticket Link

Fixes https://github.com/mattermost/mattermost-server/issues/18288

#### Release Note

```release-note

Added a new config setting ServiceSettings.DeveloperFlags.

```


amyblais:
> hey @amyblais . Can you please take a look at this PR?

Not sure why this hasn't been assigned yet. I've assigned a few reviewers now.

model/config.go
```diff
@@ -416,6 +417,10 @@ func (s *ServiceSettings) SetDefaults(isUpdate bool) {
 		s.EnableDeveloper = NewBool(false)
 	}
 
+	if !*s.EnableDeveloper || s.DeveloperFlags == nil {
+		s.DeveloperFlags = []string{}
```
lieut-data, line 421: @hmhealey might be able to correct me, but I think we want this to just be a `string` in the configuration vs. having the enduser actually configure this as an array.

model/config.go
```diff
@@ -302,6 +302,7 @@ type ServiceSettings struct {
 	RestrictLinkPreviews                              *string  `access:"site_posts"`
 	EnableTesting                                     *bool    `access:"environment_developer,write_restrictable,cloud_restrictable"`
 	EnableDeveloper                                   *bool    `access:"environment_developer,write_restrictable,cloud_restrictable"`
+	DeveloperFlags                                    []string `access:"environment_developer"`
```
lieut-data, line 305: Oh, cool! I didn't know `environment_developer` was a thing!

web/handlers.go
```diff
@@ -176,16 +176,13 @@ func (h Handler) ServeHTTP(w http.ResponseWriter, r *http.Request) {
...
 		devCSP := ""
-		if model.BuildNumber == "dev" {
-			devCSP += " 'unsafe-eval'"
-		}
-
-		// Add unsafe-inline to unlock extensions like React & Redux DevTools in Firefox
-		// see https://github.com/reduxjs/redux-devtools/issues/380
-		if model.BuildNumber == "dev" {
-			devCSP += " 'unsafe-inline'"
+		// Add flags like unsafe-eval and unsafe-inline for debugging during development
```
lieut-data, line 180: @agnivade, is there a recommended Go library we should use for parsing the desired flags as a string vs. an array? I seem to recall your mentioning we can copy what Go does in a case like this.

lieut-data (changes requested):
Thanks so much @SharathHuddar! Your implementation helped me realize there were some other things to think about, so I've looped in @agnivade in addition to @hmhealey for some perspective.

Appreciate your work here :)

model/config.go
agnivade: Not Harrison, but yes, this should be a string.

web/handlers.go
agnivade: Yes, I'll need to check the code. Although I don't think it's available as a library. We just need to copy over the `GODEBUG` parsing implementation.

web/handlers.go
```diff
@@ -176,16 +176,13 @@ func (h Handler) ServeHTTP(w http.ResponseWriter, r *http.Request) {
 		// Instruct the browser not to display us in an iframe unless is the same origin for anti-clickjacking
 		w.Header().Set("X-Frame-Options", "SAMEORIGIN")
 
-		// Add unsafe-eval to the content security policy for faster source maps in development mode
 		devCSP := ""
-		if model.BuildNumber == "dev" {
-			devCSP += " 'unsafe-eval'"
```
hmhealey, line 182: I think we need to leave the old defaults when developing locally or else source maps will stop working. I don't think people should need to configure the new setting to get the old behaviour during development

hmhealey (changes requested):

emilyacook:
Thanks for participating in Hacktoberfest! You can claim your sticker set here: https://get.printfection.com/hacktober21/4144583267 @SharathHuddar

web/handlers.go
SharathHuddar: @agnivade I'm very new to golang. Can you please tell me when I can find the `GODEBUG` parsing implementation.

model/config.go
SharathHuddar: If we're going with a string, is there a preferred format for this string? As this string would still contain a list of values like `unsafe-eval` , `unsafe-inline` etc

web/handlers.go
agnivade: It's nothing too complicated. You just have to run a loop on `strings.Split(os.Getenv("GODEBUG"), ",")`. And then split again on "=", and check for the appropriate key.

SharathHuddar:
@lieut-data @agnivade @hmhealey have resolved all your comments. Please take a look

lieut-data (approved):
Thanks for the second stab, @SharathHuddar! I realized I've left a bunch of details unclear in my original proposal, and took it upon myself to clarify a few things by building on your work and pushing directly to your branch. I'll let @hmhealey and @agnivade finish the review, but a few extra comments below.

hmhealey (approved):

services/telemetry/telemetry.go
```diff
@@ -373,6 +373,7 @@ func (ts *TelemetryService) trackConfig() {
 		"experimental_enable_authentication_transfer":             *cfg.ServiceSettings.ExperimentalEnableAuthenticationTransfer,
 		"enable_testing":                                          cfg.ServiceSettings.EnableTesting,
 		"enable_developer":                                        *cfg.ServiceSettings.EnableDeveloper,
+		"developer_flags":                                         isDefault(*cfg.ServiceSettings.DeveloperFlags, model.ServiceSettingsDefaultDeveloperFlags),
```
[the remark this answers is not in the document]
agnivade: None :)

web/handlers.go
```diff
@@ -177,15 +178,52 @@ func (h Handler) ServeHTTP(w http.ResponseWriter, r *http.Request) {
...
+				c.Logger.Warn("Unable to parse developer flag", mlog.String("developer_flag", devFlagKeyValueStr))
+				continue
+			}
+			devFlagKey := devFlagKeyValueSplit[0]
+			devFlagValue := devFlagKeyValueSplit[1]
+
+			// Ignore disabled keys
+			if devFlagValue != "true" {
+				continue
+			}
```
agnivade, line 205: This is not correct. They keys can have any value. What values can a key have is decided by the key.

In the switch-case, check for the value, and if it's true, then set the map entry. You might also want to handle quotes. But it's not necessary for this PR. If you want to leave it, then just add a comment that quotes are not handled.

web/handlers.go
```diff
@@ -177,15 +178,52 @@ func (h Handler) ServeHTTP(w http.ResponseWriter, r *http.Request) {
 		w.Header().Set("X-Frame-Options", "SAMEORIGIN")
 
 		// Add unsafe-eval to the content security policy for faster source maps in development mode
-		devCSP := ""
+		devCSPMap := make(map[string]bool)
```
agnivade, line 181: This whole logic is getting too big to be contained inside `ServeHTTP`. Let's refactor this outside to a separate function which takes the `*c.App.Config().ServiceSettings.DeveloperFlags` setting and returns the devCSP value. That would also make it easier for you to test this function.

web/handlers.go
```diff
@@ -177,15 +178,52 @@ func (h Handler) ServeHTTP(w http.ResponseWriter, r *http.Request) {
...
 		// Add unsafe-inline to unlock extensions like React & Redux DevTools in Firefox
 		// see https://github.com/reduxjs/redux-devtools/issues/380
 		if model.BuildNumber == "dev" {
-			devCSP += " 'unsafe-inline'"
+			devCSPMap["unsafe-inline"] = true
+		}
+
+		// Add supported flags for debugging during development, even if not on a dev build.
+		for _, devFlagKeyValueStr := range strings.Split(*c.App.Config().ServiceSettings.DeveloperFlags, ",") {
+			devFlagKeyValueSplit := strings.SplitN(devFlagKeyValueStr, "=", 2)
```
agnivade, line 194: Nit: too big of a variable name. Might be good to reduce it to `kvSplit`. Always prefer short names rather than long: https://github.com/golang/go/wiki/CodeReviewComments#variable-names.

web/handlers.go
```diff
@@ -177,15 +178,52 @@ func (h Handler) ServeHTTP(w http.ResponseWriter, r *http.Request) {
...
+				c.Logger.Warn("Unrecognized developer flag", mlog.String("developer_flag", devFlagKeyValueStr))
+			}
+		}
+
+		// Build a slice for sorting
+		var devCSPKeys []string
+		for key := range devCSPMap {
+			devCSPKeys = append(devCSPKeys, key)
+		}
+		sort.Strings(devCSPKeys)
```
agnivade, line 221: Wondering why do we need to sort the keys? I don't believe they are order sensitive.

agnivade (commented):
Thanks @SharathHuddar. Coming along nicely, just some small improvements.

web/handlers.go
SharathHuddar: @agnivade you mean we should support flag other than `unsafe-eval` and `unsafe-inline` ?

web/handlers.go
lieut-data: The order here is necessary to ensure a stable CSP header directive. This is necessary at least for tests, though I'm not sure it has an impact elsewhere.

web/handlers.go
lieut-data: I take the fault for the complexity add-on here :)

web/handlers.go
lieut-data: We could definitely support more developer flags in other places, at which point we could refactor the necessary code to parse thereafter.

But in the context of `web/handlers.go`, I think there's only `unsafe-eval` and `unsafe-inline`, and we shouldn't pre-emptively add more complexity here. Unless we also want to support a dynamic type of CSP directive?

web/handlers.go
SharathHuddar: I've refactored this into a separate function and have added tests for the same. However, I'm running into the following error while using `c.Logger.Warn` :

`
--- FAIL: TestGenerateDevCSP (0.53s)
    --- FAIL: TestGenerateDevCSP/dev_mode (0.53s)
panic: runtime error: invalid memory address or nil pointer dereference [recovered]
[signal SIGSEGV: segmentation violation code=0x1 addr=0x0 pc=0x28e3035]

goroutine 8236 [running]:
testing.tRunner.func1.2({0x2b04bc0, 0x4525800})
        /usr/local/Cellar/go/1.17.2/libexec/src/testing/testing.go:1209 +0x24e
testing.tRunner.func1()
        /usr/local/Cellar/go/1.17.2/libexec/src/testing/testing.go:1212 +0x218
panic({0x2b04bc0, 0x4525800})
        /usr/local/Cellar/go/1.17.2/libexec/src/runtime/panic.go:1038 +0x215
github.com/mattermost/mattermost-server/v6/shared/mlog.(*Logger).Warn(...)
        /Users/spotdraft/personal_workspace/mattermost-server/shared/mlog/mlog.go:302
command-line-arguments.(*Handler).generateDevCSP(0x326e110, {{0x32b0410, 0xc00231c358}, 0xc0037b6400, 0x0, 0x0, 0x0, {0x0, 0x0}})
        /Users/spotdraft/personal_workspace/mattermost-server/web/handlers.go:107 +0x7f5
command-line-arguments.TestGenerateDevCSP.func1(0x0)
        /Users/spotdraft/personal_workspace/mattermost-server/web/handlers_test.go:412 +0x1b5
testing.tRunner(0xc000fc76c0, 0x2eca960)
        /usr/local/Cellar/go/1.17.2/libexec/src/testing/testing.go:1259 +0x102
created by testing.(*T).Run
        /usr/local/Cellar/go/1.17.2/libexec/src/testing/testing.go:1306 +0x35a
FAIL    command-line-arguments  3.914s
FAIL

`

I'm not able to figure out what the issue is here. Can one of you please help me out here?

web/handlers.go
agnivade: Indeed. In that case, I'd just suggest to check for key existence in the map and then append that in the string.

web/handlers.go
agnivade: Got it, I don't think we would add any more complexity than a few more lines. We can leave quote handling out of this. But let's move the parsing logic to a separate function.

web/handlers.go
agnivade: The error says: "invalid memory address or nil pointer dereference". So that means it is trying to dereference a nil pointer. If you look at the code, you can see it's coming from the line `c.Logger.Warn`. And you can see that Logger is not set for the context when you are initializing the test.

Also, there's no need to make it a method of `*Handler`, just a normal function is fine. We can just pass the context, no need to dereference that, and get the string.

web/handlers.go
```diff
@@ -86,6 +86,53 @@ type Handler struct {
 	cspShaDirective string
 }
 
+func GenerateDevCSP(c Context) string {
```
agnivade, line 89: Let's unexport this. There's no need to export an internal function.

web/handlers.go
```diff
@@ -86,6 +86,53 @@ type Handler struct {
...
+		case "unsafe-eval", "unsafe-inline":
+			devCSPMap[devFlagKey] = true
+		default:
+			c.Logger.Warn("Unrecognized developer flag", mlog.String("developer_flag", devFlagKVStr))
+		}
+	}
+	var devCSP string
+	supportedCSPFlags := []string{"unsafe-eval", "unsafe-inline"}
+	for _, devCSPFlag := range supportedCSPFlags {
+		if _, ok := devCSPMap[devCSPFlag]; ok {
```
agnivade, line 128: There is no need for `_, ok`. We can directly write `if devCSPMap[devCSPFlag] {`. If the key doesn't exist, it will return false.

agnivade (commented):
Looking great now. Just some minor adjustments suggested.

agnivade (approved):

agnivade:
Great work @SharathHuddar ! Thank you for your patience during the review cycle.

@saturninoabril - I assigned it to you, but please feel free to re-assign appropriately.

saturninoabril (approved):
@SharathHuddar Tested and worked as expected. Great work and thanks!

PR #18855, opened 2021-10-27
[MM-39597] Implement ConfigDiffs.Sanitize()

streamer45:
#### Summary

They say _never two without three_  so we implement a new `Sanitize()` function that we call before logging config diffs.

This is currently achieved by having a prepared map of sensitive config paths that we compare.  This can be later improved by using something more dynamic, maybe struct tags or something else. Also to be reviewed is the sanitization of specific config settings that can contain a mixture of sensitive and non-sensitive portions like DSNs and plugin configs. Right now we replace those as a whole.

#### Ticket

https://mattermost.atlassian.net/browse/MM-39597

#### Release Note

```release-note
NONE
```

### DIFF
--- a/api4/config.go
+++ b/api4/config.go
@@ -400,8 +400,6 @@ func migrateConfig(c *Context, w http.ResponseWriter, r *http.Request) {
 	}
 
 	auditRec := c.MakeAuditRecord("migrateConfig", audit.Fail)
-	auditRec.AddMeta("from", from)
-	auditRec.AddMeta("to", to)
 	defer c.LogAuditRec(auditRec)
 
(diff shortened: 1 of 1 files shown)


amyblais:
@streamer45 Is this needed for 6.0.x?

streamer45:
> @streamer45 Is this needed for 6.0.x?

I am guessing so but I've asked Security to confirm.

streamer45:
@amyblais Confirmed this needs backporting up to (and including) ESR.

streamer45:
@amyblais I need to do those manually as well. Will do it in the morning.

PR #18886, opened 2021-10-28
Reuse database connection if running on PostgreSQL instead of creating a new one

mgdelacroix:
#### Summary
This PR reuses the server's database connection for migrations when running on PostgreSQL as there is no need to create a new one.

```release-note
NONE
```


agnivade:
Feel free to request a review whenever it's ready. :)

PR #18927, opened 2021-11-03
Fix PluginHealthCheckJob

agnivade:
We were incorrectly calling to start PluginHealthCheckJob
without initializing the plugins environment.

To fix this, we move the code inside initPlugins right after
the environment is initialized.

To respond to config changes, we call it again from pluginsEnvironment.Shutdown
which gets called from ConfigListener when plugins are disabled. And initPlugins
is anyways called again from ConfigListener which plugins are enabled,
so we can avoid checking for PluginSettings.Enable again in the call.

We also rename the method to better indicate its nature.

During this, we also uncover and fix another bug where disabling
plugins would not shut down plugins at all because we were calling
s.GetPluginsEnvironment() directly which returns nil if plugins
were disabled. The approach we follow is to manually acquire the lock
whenever we need access to the struct ignoring config setting.
We fix that as well.

https://community-daily.mattermost.com/boards/workspace/zyoahc9uapdn3xdptac6jb69ic/285b80a3-257d-41f6-8cf4-ed80ca9d92e5/495cdb4d-c13a-4992-8eb9-80cfee2819a4?c=6ef6178c-3512-4e57-8edd-1d2b66a09c9e
```release-note
NONE
```

### DIFF
--- a/app/plugin.go
+++ b/app/plugin.go
@@ -208,6 +208,8 @@ func (s *Server) initPlugins(c *request.Context, pluginDir, webappPluginDir stri
 	s.PluginsEnvironment = env
 	s.PluginsLock.Unlock()
 
+	s.PluginsEnvironment.TogglePluginHealthCheckJob(*s.Config().PluginSettings.EnableHealthCheck)
+
 	if err := s.syncPlugins(); err != nil {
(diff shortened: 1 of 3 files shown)


mickmister (approved):
LGTM, code is cleaner too now :+1:

PR #18930, opened 2021-11-03
Revert "DOPS-602: Change how mmctl gets packaged and locally fetched. "

cpanato:
Reverts mattermost/mattermost-server#18335

it is breaking the internal build
reverting and will think how to redo that in another way


cpanato:
i'm going to merge to unblock master branch

PR #18936, opened 2021-11-04
Add JobSettings.CleanupJobsThresholdDays to telemetry

agnivade:
```release-note
NONE
```

### DIFF
--- a/model/config.go
+++ b/model/config.go
@@ -2615,8 +2615,8 @@ func (s *DataRetentionSettings) SetDefaults() {
 }
 
 type JobSettings struct {
-	RunJobs                  *bool `access:"write_restrictable,cloud_restrictable"` // telemetry:none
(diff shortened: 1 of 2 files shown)


isacikgoz (approved):
I think we might want to disable `// telemetry: none` in https://github.com/mattermost/mattermost-server/blob/c88c9b6625252ec14fd359ec853d31479e1f2a19/model/config.go#L3028 As whenever we add new setting to `JobSettings`, it can be overlooked again. Instead, we can just ignore these two:
https://github.com/mattermost/mattermost-server/blob/c88c9b6625252ec14fd359ec853d31479e1f2a19/model/config.go#L2618
https://github.com/mattermost/mattermost-server/blob/c88c9b6625252ec14fd359ec853d31479e1f2a19/model/config.go#L2619

I'm 0/5 on this, hence I'm approving and leaving it to you :)

agnivade:
Oh yes, good point. Let me do that.

PR #18942, opened 2021-11-04
[MM-38239 & MM-39788] Recent files causing crash

BenCookie95:
#### Summary
The crash was caused by missing posts in the redux state. I added this endpoint so we can load in the missing posts.

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-38239
https://mattermost.atlassian.net/browse/MM-39788

#### Related PRs
- https://github.com/mattermost/mattermost-webapp/pull/9323
- https://github.com/mattermost/mattermost-api-reference/pull/663

#### Release Note

```release-note
NONE
```


api4/post.go
```diff
@@ -396,6 +397,54 @@ func getPost(c *Context, w http.ResponseWriter, r *http.Request) {
 	}
 }
 
+func getPostsByIds(c *Context, w http.ResponseWriter, r *http.Request) {
+	var postIDs []string
+
+	json.NewDecoder(r.Body).Decode(&postIDs)
```
hmhealey, line 403: Along those same lines, we might want to wrap `r.Body` in an `io.LimitReader` to avoid causing problems if someone sends an arbitrarily long array of IDs to the server. I can't remember if we have to do that every time we read arbitrary JSON from the request body, but I do remember doing that other places where we were using `ReadAll` to read image data and such

api4/post.go
```diff
@@ -396,6 +397,54 @@ func getPost(c *Context, w http.ResponseWriter, r *http.Request) {
...
+	if postIDs == nil {
+		c.SetInvalidParam("post_ids")
+		return
+	}
+
+	var posts = []*model.Post{}
+
+	for _, postID := range postIDs {
+		post, err := c.App.GetSinglePost(postID)
```
hmhealey, line 412: Should we maybe get all the posts from the DB in a single query? The `getUsersByIds` one works like that

api4/post.go
```diff
@@ -396,6 +397,54 @@ func getPost(c *Context, w http.ResponseWriter, r *http.Request) {
 	}
 }
 
+func getPostsByIds(c *Context, w http.ResponseWriter, r *http.Request) {
+	var postIDs []string
+
+	json.NewDecoder(r.Body).Decode(&postIDs)
+	if postIDs == nil {
```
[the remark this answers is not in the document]
BenCookie95: It would probably be a good idea to enforce a limit, the users one doesn't but I think it would be nice. I can probably set it to 60 which is generally the post fetch limit anyway.

hmhealey (approved):
I didn't even realize that we already had a store method to get the posts by IDs. That's convenient 😄

api4/post.go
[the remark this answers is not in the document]
mkraft: I'm not sure of the history, but I do see [a limit of 1000 posts for getting posts by channel](https://github.com/mattermost/mattermost-server/blob/master/store/sqlstore/post_store.go#L990). I think a limit like that is totally more than reasonable.

api4/post.go
mkraft: 1000 IDs shouldn't be more than about 35 KB. 

I do see `LimitReader` used in a few place and a lot of places that are using `ReadAll` unlimited, deferring to limits imposed by the underlying infrastructure. Another option is [`MaxBytesReader`](https://pkg.go.dev/net/http#MaxBytesReader). I'm 3/5 that this should be handle consistently across the API. 

Perhaps the Server Platform team (cc @agnivade @isacikgoz) have related plans? Perhaps a campaign is in order?

agnivade: This is already there: https://github.com/mattermost/mattermost-server/blob/df3ed307df0924f57363ca990703862d66193633/web/handlers.go#L151-L157.

api4/post.go
```diff
@@ -396,6 +397,54 @@ func getPost(c *Context, w http.ResponseWriter, r *http.Request) {
...
 
+func getPostsByIds(c *Context, w http.ResponseWriter, r *http.Request) {
+	postIDs := model.ArrayFromJSON(r.Body)
+
+	if len(postIDs) == 0 {
+		c.SetInvalidParam("post_ids")
+		return
+	}
+
+	postsList, err := c.App.GetPostsByIds(postIDs)
```
mkraft, line 408: Were you planning to limit `postIDs` to 1000 or did you decide against that?

mkraft (approved):

furqanmlk (approved):
Working as expected 
Details has been added in Webapp PR

store/sqlstore/post_store.go
```diff
@@ -1970,7 +1970,7 @@ func (s *SqlPostStore) GetPostsCreatedAt(channelId string, time int64) ([]*model
 func (s *SqlPostStore) GetPostsByIds(postIds []string) ([]*model.Post, error) {
 	keys, params := MapStringsToQueryParams(postIds, "Post")
 
-	query := `SELECT p.*, (SELECT count(Posts.Id) FROM Posts WHERE Posts.RootId = (CASE WHEN p.RootId = '' THEN p.Id ELSE p.RootId END) AND Posts.DeleteAt = 0) as ReplyCount FROM Posts p WHERE p.Id IN ` + keys + ` ORDER BY CreateAt DESC`
+	query := `SELECT p.*, (SELECT count(Posts.Id) FROM Posts WHERE Posts.RootId = (CASE WHEN p.RootId = '' THEN p.Id ELSE p.RootId END) AND Posts.DeleteAt = 0) as ReplyCount FROM Posts p WHERE p.Id IN ` + keys + ` ORDER BY CreateAt DESC LIMIT 1000`
```
agnivade, line 1973: We should not hardcode limits like this in the SQL query. It breaks other callers of this API. You might query for 1001 IDs, but get only 1000 posts.

Instead, we should enforce the limitation on the API layer. In the API layer, please check for the number of post ids, and if they are more than 1000, throw an error to enforce proper usage rather than silently truncating posts.

app/post.go
```diff
@@ -1668,3 +1668,18 @@ func (a *App) GetPostIfAuthorized(postID string, session *model.Session) (*model
 
 	return post, nil
 }
+
+func (a *App) GetPostsByIds(postIDs []string) ([]*model.Post, *model.AppError) {
+	posts, err := a.Srv().Store.Post().GetPostsByIds(postIDs)
+	if err != nil {
+		var nfErr *store.ErrNotFound
+		switch {
+		case errors.As(err, &nfErr):
```
agnivade, line 1677: The store API never returns a `store.ErrNotFound` error. But it's a good idea if you want to change the store method to return this error. It's only called in searchlayer/post_layer.go, so it should be fine.

go.tools.mod
```diff
@@ -7,7 +7,7 @@ require (
 	github.com/golang-migrate/migrate/v4 v4.14.1 // indirect
 	github.com/jstemmer/go-junit-report v0.9.1 // indirect
 	github.com/jteeuwen/go-bindata v3.0.7+incompatible // indirect
-	github.com/mattermost/mattermost-utilities/mmgotool v0.0.0-20210721133912-8b250bf4d0f6 // indirect
+	github.com/mattermost/mattermost-utilities/mmgotool v0.0.0-20211006080735-07b1b58b8b09 // indirect
```
agnivade, line 10: Seems unrelated?

api4/post_test.go
```diff
@@ -2713,3 +2713,29 @@ func TestSetPostUnreadWithoutCollapsedThreads(t *testing.T) {
...
 }
+func TestGetPostsByIds(t *testing.T) {
+	th := Setup(t).InitBasic()
+	defer th.TearDown()
+	client := th.Client
+
+	post1 := th.CreatePost()
+	post2 := th.CreatePost()
+	// post3 := th.CreatePost()
+	// post4 := th.CreatePost()
```
agnivade, line 2724: Let's remove the debug lines?

api4/post.go
```diff
@@ -396,6 +397,54 @@ func getPost(c *Context, w http.ResponseWriter, r *http.Request) {
...
+		posts = append(posts, post)
+	}
+
+	b, jsonErr := json.Marshal(posts)
+	if jsonErr != nil {
+		c.Err = model.NewAppError("getPostsByIds", "api.post.search_posts.invalid_body.app_error", nil, jsonErr.Error(), http.StatusBadRequest)
+		return
+	}
+
+	w.Write(b)
```
agnivade, line 445: It's better re-written as:

```go
if err := json.NewEncoder(w).Encode(posts); err != nil {
	mlog.Warn("Error while writing response", mlog.Err(err))
}
```

to make it consistent with others.

api4/post.go
```diff
@@ -396,6 +397,54 @@ func getPost(c *Context, w http.ResponseWriter, r *http.Request) {
...
+	if err != nil {
+		c.Err = err
+		return
+	}
+
+	var posts = []*model.Post{}
+
+	for _, post := range postsList {
+
+		channel, err := c.App.GetChannel(post.ChannelId)
```
agnivade, line 418: This will call the same channelID repeatedly for every post even if they are from the same channel. We should create a map for this, and make the call only for new channels.

api4/post.go
```diff
@@ -396,6 +397,54 @@ func getPost(c *Context, w http.ResponseWriter, r *http.Request) {
...
+		if err != nil {
+			c.Err = err
+			return
+		}
+
+		if !c.App.SessionHasPermissionToChannel(*c.AppContext.Session(), channel.Id, model.PermissionReadChannel) {
+			if channel.Type == model.ChannelTypeOpen {
+				if !c.App.SessionHasPermissionToTeam(*c.AppContext.Session(), channel.TeamId, model.PermissionReadPublicChannel) {
```
agnivade, line 426: Seems like both these conditions can be combined with an `&&`?

BenCookie95: I don't think they can be combined with `&&`:

If a channel is set to open and `hasPermissionToTeam` is evaluated to `true`, the post would not be included in the list if they conditions are combined with `&&`

api4/post_test.go
BenCookie95: Removed

go.tools.mod
BenCookie95: removed

app/post.go
BenCookie95: Add a not found error to the store if no posts are found

store/sqlstore/post_store.go
BenCookie95: Enforced on the api layer now

api4/post.go
BenCookie95: Updated

api4/post.go
BenCookie95: I like this idea, added the map

api4/post.go
agnivade: Ah my apologies, it's because of the `else` condition. Would something like this work?

```
if !open || (open && !hasPermission()) {
  continue
}
```

0/5.

BenCookie95: That works, updated

agnivade (approved):
Great stuff Ben, thank you!

PR #18945, opened 2021-11-04
MM-39635: Migrate gorp to sqlx #18873

mr-aboutin:
#### Summary

This pull request migrate gorp to sqlx as demonstrated in #18873.

#### Ticket Link

Fixes https://github.com/mattermost/mattermost-server/issues/18873

#### Release Note

```release-note
Migrated GetReplica() to GetReplicaX() to use sqlx instead of gorp in remote_cluster_store.go
```


agnivade:
Hey @mr-aboutin - there are other queries here which need to be migrated as well.

store/sqlstore/remote_cluster_store.go
```diff
@@ -54,7 +79,22 @@ func (s sqlRemoteClusterStore) Update(remoteCluster *model.RemoteCluster) (*mode
...
+			  SET Name = :Name,
+			  Token = :Token,
+			  RemoteToken = :RemoteToken,
+			  CreatorId = :CreatorId,
+			  DisplayName = :DisplayName,
+			  SiteURL = :SiteURL,
+			  Topics = :Topics
+			  WHERE	RemoteId = :RemoteId`
+
+	err := s.ValidSiteURL(*remoteCluster)
```
agnivade, line 92: There's no need for this. The update should always happen by the primary key, which in this case is a compound key. So let's move the `Name` as well in the `WHERE` clause.

store/sqlstore/remote_cluster_store.go
```diff
@@ -42,7 +42,32 @@ func (s sqlRemoteClusterStore) Save(remoteCluster *model.RemoteCluster) (*model.
 		return nil, err
 	}
 
-	if err := s.GetMaster().Insert(remoteCluster); err != nil {
+	query := `INSERT INTO RemoteClusters
```
agnivade, line 45: Nit: may be good to coalesce the column names and values into their separate rows.

store/sqlstore/remote_cluster_store.go
```diff
@@ -140,7 +180,7 @@ func (s sqlRemoteClusterStore) GetAll(filter model.RemoteClusterQueryFilter) ([]
 	}
 
 	var list []*model.RemoteCluster
```
agnivade, line 182: `list := []*model.RemoteCluster{}`

store/sqlstore/remote_cluster_store.go
agnivade: Yeah, I meant to have a single line for column names and their values. But this is okay too.

agnivade:
Hey @mr-aboutin - just one small thing (https://github.com/mattermost/mattermost-server/pull/18945#discussion_r746645697) pending on this one.

mr-aboutin:
@agnivade Sorry for te delay, I've had a couple tings on my plate the last week but I'll be fixing before the end of the week.

store/sqlstore/remote_cluster_store.go
mr-aboutin: By removing the validation, the tests still fails because the Pkey is correct. In the `remote_cluster_test`, We test by having a model with the same `RemoteTeamId` and `SiteUrl` which does not go against the Primary Key (Name, SiteUrl)

agnivade: That is because we were not updating all the columns :)

If we fix the query to

```diff
        query := `UPDATE RemoteClusters
-                         SET Token = :Token,
-                         RemoteToken = :RemoteToken,
-                         CreatorId = :CreatorId,
+                         SET RemoteTeamId = :RemoteTeamId,
                          DisplayName = :DisplayName,
                          SiteURL = :SiteURL,
+                         CreateAt = :CreateAt,
+                         LastPingAt = :LastPingAt,
+                         Token = :Token,
+                         RemoteToken = :RemoteToken,
+                         CreatorId = :CreatorId,
                          Topics = :Topics
                          WHERE RemoteId = :RemoteId AND Name = :Name`

```

the tests pass

isacikgoz (approved):
LGTM, thanks for all the effort @mr-aboutin 🙏

mr-aboutin:
Thanks both of you for the help along the way 😃

agnivade (approved):
Thanks @mr-aboutin !

PR #18959, opened 2021-11-08
MM-39928: Fix for: MaximumPluginFileSize is too small after adding arm64 target

cpoile:
#### Summary
- When adding the arm64 target, the Playbooks plugin bundle crosses the 50MB threshold (51MB). Increase the upload file limit for us and other plugins that will run into this in the future.

#### Ticket Link
- https://mattermost.atlassian.net/browse/MM-39928
 
#### Release Note

-->
```release-note
NONE
```

### DIFF
--- a/api4/plugin.go
+++ b/api4/plugin.go
@@ -21,7 +21,7 @@ import (
 )
 
 const (
-	MaximumPluginFileSize = 50 * 1024 * 1024
+	MaximumPluginFileSize = 75 * 1024 * 1024
 )
 
 func (api *API) InitPlugin() {


api4/plugin.go
```diff
@@ -21,7 +21,7 @@ import (
 )
 
 const (
-	MaximumPluginFileSize = 50 * 1024 * 1024
+	MaximumPluginFileSize = 75 * 1024 * 1024
```
iomodo, line 24: Should we be increasing the size just for the developers or when testing is enabled or for everyone?

lieut-data, line 24: Hmm, I don't /think/ this is the tunable we want to configure -- IIRC, this doesn't constrain the amount of data `ParseMultipartForm` ingests, just how much it puts into memory vs. storing in temporary files. There's no need to store more in memory for quick access, since the interesting form parts should already be available in the first 50Mb (or something in the request is probably messed up).

When I fixed this locally, I had to bump `FileSettings.MaxFileSize` in the configuration, but I didn't dig deeper to understand the path through which this tunable vs. that one takes effect.

lieut-data (changes requested):
I wasn't expecting the code change to take place in this particular location -- comment below.

api4/plugin.go
cpoile: Yep, I ran into this and was just about to set the issue as WIP. Will fix.

cpoile: Nevermind, this whole thing is unnecessary -- Claudio bumped up the default to 100MB in August, we're just running old configs. But we should add to the next release's release notes. (for servers that have a similar old config)

PR #18964, opened 2021-11-09
Ordering clarifications

agnivade:
Clearly outline the dependency graph
during server bootup.

https://community-daily.mattermost.com/boards/workspace/zyoahc9uapdn3xdptac6jb69ic/285b80a3-257d-41f6-8cf4-ed80ca9d92e5/495cdb4d-c13a-4992-8eb9-80cfee2819a4?c=87df1e15-588e-49ff-8bd1-ffa9651b8c82

### DIFF
--- a/app/server.go
+++ b/app/server.go
@@ -222,6 +222,11 @@ func NewServer(options ...Option) (*Server, error) {
 		}
 	}
 
+	// Following outlines the specific set of steps
+	// performed during server bootup. They are sensitive to order
+	// and has dependency requirements with the previous step.
+	//
+	// Step 1: Config.
 	if s.configStore == nil {
 		innerStore, err := config.NewFileStore("config.json")
 		if err != nil {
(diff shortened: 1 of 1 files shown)


agnivade:
/release-note-none

@streamer45 - Could you re-check this? The plugin initialization is now moved out from NewServer to the products.Start phase. So the imaging initialization isn't really dependent on this any more. I tested with some plugins and it seems to work fine. Although I couldn't find any plugins that update user profiles on startup.

streamer45:
> @streamer45 - Could you re-check this? The plugin initialization is now moved out from NewServer to the products.Start phase. So the imaging initialization isn't really dependent on this any more. I tested with some plugins and it seems to work fine. Although I couldn't find any plugins that update user profiles on startup.

Sure. Any plugin that calls `EnsureBot` with an image should trigger that I believe (e.g. playbooks).

agnivade:
Yep, confirmed. Works fine.

streamer45 (approved):
Looks good :+1: 

My only comment is that we may consider placing imaging initializations further up given they are completely decoupled from the rest. I can see some of the logic that happens before potentially requiring those at some point (e.g. in app migrations, jobs). Not a blocker though.

agnivade:
No worries, let me move up.

PR #18966, opened 2021-11-09
MM-39341: Sentry crash: nil dereference in *User.ClearNonProfileFields

agnivade:
In some cases, an invalid participant id like an empty string
might creep into the database. It's not exactly clear
how can this happen, but if it does, then it breaks the logic
of assuming that the users slice will contain all users
in the participants slice.

To prevent this, we check if a match was found before
adding it in the slice.

https://community-daily.mattermost.com/boards/workspace/zyoahc9uapdn3xdptac6jb69ic/285b80a3-257d-41f6-8cf4-ed80ca9d92e5/495cdb4d-c13a-4992-8eb9-80cfee2819a4/7sixm4t7c6rsr3d8ydxbtfgy9gr

```release-note
NONE
```

### DIFF
--- a/store/sqlstore/thread_store.go
+++ b/store/sqlstore/thread_store.go
@@ -460,7 +460,9 @@ func (s *SqlThreadStore) GetThreadForUser(teamId string, threadMembership *model
 				break
 			}
 		}
-		participants = append(participants, participant)
+		if participant != nil {
+			participants = append(participants, participant)
+		}
 	}
 
 	result := &model.ThreadResponse{
--- a/store/storetest/thread_store.go
+++ b/store/storetest/thread_store.go
@@ -451,6 +451,25 @@ func testThreadStorePopulation(t *testing.T, ss store.Store) {
 		require.NoError(t, err)
 		require.Equal(t, int64(0), th.UnreadReplies)
 	})
+
+	t.Run("Empty participantID should not appear in thread response", func(t *testing.T) {
+		newPosts := makeSomePosts()
+		opts := store.ThreadMembershipOpts{
+			Following:             true,
+			IncrementMentions:     false,
+			UpdateFollowing:       true,
+			UpdateViewedTimestamp: false,
+			UpdateParticipants:    true,
+		}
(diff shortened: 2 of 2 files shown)


agnivade:
@ashishbhate @koox00 - This is just a WIP to start a conversation. This PR attempts to fix a sentry crash, but I am not able to find the real root cause here.


The issue happens when `GetThreadForUser` returns a nil user in the `Participants` slice here:

https://github.com/mattermost/mattermost-server/blob/3595a229a88fd1584210529bd70593feb60be97f/app/user.go#L2229-L2237

which eventually goes to call https://github.com/mattermost/mattermost-server/blob/3595a229a88fd1584210529bd70593feb60be97f/app/user.go#L947 and crashes.

Now the question is, why is the user pointer nil in the first place? On looking at how the participants slice is populated, it is possible if the main user slice does not contain the participant.

https://github.com/mattermost/mattermost-server/blob/ba5f1f3f5c5ccb4fef269969a7771617665220ab/store/sqlstore/thread_store.go#L454-L464

because we are not checking if the participant got assigned or not. But then again, the deeper question is why should the the users slice not have the participant in the first place?

This leads to the cache layer and store layer, both of which looks okay to me.

Since both of you have a better understanding of this, I am wondering if you would be able to shed any light here as to how can this happen?

agnivade:
My bad, I didn't realize Ashish was still on PTO.

koox00 (approved):

agnivade:
I can see the crash happening if an empty string appears in the participants slice. I haven't yet reproduced how can this happen, but I think this fix should be good to go in any case to make the logic foolproof.

agnivade:
Figured it out. If you delete a user, we don't remove the participant id from the threads. That's why it fails to find the user struct from the id. I will file a ticket.

/update-branch

ashishbhate (approved):

amyblais:
@agnivade Does this repro on 5.37 ESR? We're working on a fix for 5.37 and I'm wondering if it makes sense to include this or not.

agnivade:
It does not.

PR #18982, opened 2021-11-11
MM-39612: Make acquiring and removing connections atomic

agnivade:
The reconnect phase of a websocket was split into two parts:
one where we check if a connection with a given connectionID
exists or not. And second, where we remove that connection
and insert the new connection again in the index.

This would lead to a race where it would be possible
for 2 concurrent requests for the same connectionID to go through
which would lead to separate goroutines working on the same dead queue.

We simplify this by removing the connection from the index
in the check connection stage itself. And then just add that
during register phase.

And to distinguish between a fresh and an old connection, we add
a new field called reuseCount.

While here, we also cleanup some old comments and add more
in some places.

https://mattermost.atlassian.net/browse/MM-39612

```release-note
NONE
```

### DIFF
--- a/app/web_conn.go
+++ b/app/web_conn.go
@@ -57,6 +57,7 @@ type WebConnConfig struct {
 	Locale       string
 	ConnectionID string
 	Active       bool
+	ReuseCount   int
 
 	// These aren't necessary to be exported to api layer.
 	sequence         int
@@ -94,7 +95,13 @@ type WebConn struct {
 	// to this webConn or not.
 	// It is not used as an atomic, because there is no need to.
 	// So do not use this outside the web hub.
-	active       bool
+	active bool
+	// reuseCount indicates how many times this connection has been reused.
+	// This is used to differentiate between a fresh connection and
+	// a reused connection.
+	// It's theoretically possible for this number to wrap around. But we
+	// leave that as an edge-case.
+	reuseCount   int
 	sessionToken atomic.Value
 	session      atomic.Value
 	connectionID atomic.Value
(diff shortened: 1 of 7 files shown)


agnivade:
I don't think there is a way to unit test this. But if you find one, do let me know :)

app/web_conn.go
```diff
@@ -94,7 +95,13 @@ type WebConn struct {
...
 	// It is not used as an atomic, because there is no need to.
 	// So do not use this outside the web hub.
-	active       bool
+	active bool
+	// reuseCount indicates how many times this connection has been reused.
+	// This is used to differentiate between a fresh connection and
+	// a reused connection.
+	// It's theoretically possible for this number to wrap around. But we
+	// leave that as an edge-case.
+	reuseCount   int
```
streamer45, line 104: Sounds like a bool would do just fine. Do we need the actual value?

agnivade: Yep, bool would have done fine. Just keeping it there in case we need the number in future.

agnivade:
@streamer45 - I added a new method to prevent the model API from breaking. Everything else is same. PTAL. Thanks for the help :)

streamer45 (approved):
Looks great, thanks :+1:

isacikgoz (approved):
LGTM 👍

amyblais:
@agnivade Does this repro on 5.37 ESR? We're working on a fix for 5.37 and I'm wondering if it makes sense to include this or not.

agnivade:
It does, but `EnableReliableWebsockets` isn't turned on by default in 5.37, so I don't see an immediate need to backport. If a customer enables it and encounters this bug, they can always turn it off.

Issue #18965, opened 2021-11-09
Pass expiry time from outside in `*SqlTokenStore.Cleanup`


agnivade:
#### Summary
The `Cleanup` method calculates the expiry time from inside the method.

https://github.com/mattermost/mattermost-server/blob/ba5f1f3f5c5ccb4fef269969a7771617665220ab/store/sqlstore/tokens_store.go#L77-L83

This does not give control to caller, and makes the method hard to test. Apparent from the fact that this store does not have any tests at all.

The objective of this ticket is to calculate and pass the expiry time from the upper layers, and also add a unit test.

Take a look at how jobs are cleaned for example:

https://github.com/mattermost/mattermost-server/blob/ba5f1f3f5c5ccb4fef269969a7771617665220ab/app/server.go#L1557-L1562

agnivade:
cc @isacikgoz

krotovkk:
Hello! I'd like to pick this up.

agnivade:
Go for it!

PR #19008, opened 2021-11-16
[GH-18965] - Pass expire time from outside Cleanup method

krotovkk:
#### Summary
Pass expire time of token from outside in *SqlTokenStore.Cleanup.

#### Ticket Link
Fixes https://github.com/mattermost/mattermost-server/issues/18965

#### Release Note

```release-note
NONE
```


store/sqlstore/tokens_store.go
```diff
@@ -74,10 +74,10 @@ func (s SqlTokenStore) GetByToken(tokenString string) (*model.Token, error) {
 	return &token, nil
 }
 
-func (s SqlTokenStore) Cleanup() {
+func (s SqlTokenStore) Cleanup(expireTime int64) {
```
agnivade, line 77: Nit: `expiryTime` may be better.

store/sqlstore/tokens_store.go
```diff
@@ -74,10 +74,10 @@ func (s SqlTokenStore) GetByToken(tokenString string) (*model.Token, error) {
...
 	mlog.Debug("Cleaning up token store.")
```
agnivade, line 78: Let's move this debug line to the `doTokenCleanup` function before calling this method.

store/storetest/tokens_store.go
```diff
@@ -0,0 +1,42 @@
...
+func TestTokensStore(t *testing.T, ss store.Store) {
+	t.Run("TokensCleanup", func(t *testing.T) { testTokensCleanup(t, ss) })
+}
+
+func testTokensCleanup(t *testing.T, ss store.Store) {
+	now := model.GetMillis()
+
+	for i := 0; i < 10; i++ {
+		err := ss.Token().Save(&model.Token{
+			Token:    model.NewRandomString(model.TokenSize),
```
agnivade, line 25: Use `NewTestId` for this.

agnivade (commented):
Nice work @krotovkk. Sorry for the delay on this, please feel free to ping me on the PRs so that I get quickly notified.

store/storetest/tokens_store.go
krotovkk: Thanks for review! 
Why i could't use `model.NewRandomString`? This method is using by `NewToken` function.

store/sqlstore/tokens_store.go
krotovkk: Done.

store/sqlstore/tokens_store.go
krotovkk: Done.

store/storetest/tokens_store.go
agnivade: You definitely can. And in this case it's probably fine. The reason for `NewTestId` is that during tests, we have seen that randomly generated strings clash with user defined strings when searching for objects, leading to flaky tests. That's why `NewTestId` is specifically built to replace every odd character with a digit. In this case, it's okay to use this because we aren't searching by the string.

agnivade (approved):
Looking good @krotovkk !

isacikgoz (approved):
LGTM. Thanks @krotovkk 🚀

agnivade:
FTR: Only the `ci/circleci: Build Error ` stage was failing and the CI passed. Merged this to avoid getting blocked further.

PR #19016, opened 2021-11-18
[MM-40167] Check for file deleted before download

Willyfrog:
#### Summary

Trying to access a soft-deleted file should end in a 404 not found

Screen Shot 2021-11-18 at 14 34 26

#### Ticket Link
[MM-40167](https://mattermost.atlassian.net/browse/MM-40167)

#### Release Note
```release-note
Prevented downloading a file when the post that was attached to was deleted
```

### DIFF
--- a/api4/file.go
+++ b/api4/file.go
@@ -483,6 +483,10 @@ func getFile(c *Context, w http.ResponseWriter, r *http.Request) {
 		c.Err = err
 		return
 	}
+	if info.DeleteAt != 0 {
+		c.Err = model.NewAppError("getFile", "api.file.get_file.no_file.app_error", nil, "file_id="+info.Id, http.StatusNotFound)
+	}
+
 	auditRec.AddMeta("file", info)
 
 	if info.CreatorId != c.AppContext.Session().UserId && !c.App.SessionHasPermissionToChannelByPost(*c.AppContext.Session(), info.PostId, model.PermissionReadChannel) {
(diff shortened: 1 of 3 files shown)


api4/file.go
```diff
@@ -483,6 +483,10 @@ func getFile(c *Context, w http.ResponseWriter, r *http.Request) {
 		c.Err = err
 		return
 	}
+	if info.DeleteAt != 0 {
+		c.Err = model.NewAppError("getFile", "api.file.get_file.no_file.app_error", nil, "file_id="+info.Id, http.StatusNotFound)
+	}
```
jupenur, line 488: Shouldn't there be a `return` here?

Willyfrog: yes! 🤦 
it still works as the error is carried around, but it's better if we exit early.

api4/file.go
```diff
@@ -483,6 +483,11 @@ func getFile(c *Context, w http.ResponseWriter, r *http.Request) {
 		c.Err = err
 		return
 	}
+	if info.DeleteAt != 0 {
+		c.Err = model.NewAppError("getFile", "api.file.get_file.no_file.app_error", nil, "file_id="+info.Id, http.StatusNotFound)
+		return
+	}
+
 	auditRec.AddMeta("file", info)
```
jupenur, line 491: Let's move this above the `DeleteAt` check since we have the info already.

jupenur (approved):
A unit test would be nice, but otherwise LGTM!

Willyfrog:
> A unit test would be nice, but otherwise LGTM!

agreed, added test! Thanks for pointing it out

Willyfrog:
This is labeled as a false positive, it is checked in a different layer.

PR #19018, opened 2021-11-18
Change default log rotation size to 100MB 

wiggin77:
#### Summary
The default log rotation size was mistakenly set to 10GB for V6.  The rotation size is supposed to be specified in megabytes but I think the default was mistakenly changed thinking it was KB.

This PR reverts the default to 100MB. 

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-40191

#### Release Note
```release-note
Default log rotation file size was mistakenly set to 10GB for V6, and is now reverted back to 100MB.
```

### DIFF
--- a/config/logger.go
+++ b/config/logger.go
@@ -15,7 +15,7 @@ import (
 )
 
 const (
-	LogRotateSize           = 10000
+	LogRotateSize           = 100
 	LogCompress             = true
 	LogRotateMaxAge         = 0
 	LogRotateMaxBackups     = 0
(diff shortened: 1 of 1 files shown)


config/logger.go
```diff
@@ -15,7 +15,7 @@ import (
 )
 
 const (
-	LogRotateSize           = 10000
+	LogRotateSize           = 100
```
streamer45, line 18:
```suggestion
	LogRotateSizeMB           = 100
```

Should we add the unit as a suffix like we do for other similar constants (e.g. time related)? It may help to prevent this from happening again.

streamer45 (approved):
A small suggestion but otherwise looks good :+1:

PR #19040, opened 2021-11-24
Remove reliable websockets configuration

agnivade:
The feature is stable enough to be removed
as a config knob.

For backwards compatiblity with older mobile
clients, we keep sending the config param
as true for client config requests.

https://community-daily.mattermost.com/boards/workspace/zyoahc9uapdn3xdptac6jb69ic/285b80a3-257d-41f6-8cf4-ed80ca9d92e5/495cdb4d-c13a-4992-8eb9-80cfee2819a4/7m95frqbk9o8zxin6xo9c7iusoh

```release-note
The config setting ServiceSettings.EnableReliableWebSockets has been removed, and for compatiblity with older clients, the server will always return true for the /v4/config/client endpoint.
```

Mobile PR: https://github.com/mattermost/mattermost-mobile/pull/5843
Webapp PR: https://github.com/mattermost/mattermost-webapp/pull/9430

### DIFF
--- a/api4/websocket.go
+++ b/api4/websocket.go
@@ -46,20 +46,18 @@ func connectWebSocket(c *Context, w http.ResponseWriter, r *http.Request) {
 		Active:    true,
 	}
 
-	if *c.App.Config().ServiceSettings.EnableReliableWebSockets {
-		cfg.ConnectionID = r.URL.Query().Get(connectionIDParam)
-		if cfg.ConnectionID == "" || c.AppContext.Session().UserId == "" {
-			// If not present, we assume client is not capable yet, or it's a fresh connection.
-			// We just create a new ID.
-			cfg.ConnectionID = model.NewId()
-			// In case of fresh connection id, sequence number is already zero.
-		} else {
-			cfg, err = c.App.PopulateWebConnConfig(c.AppContext.Session(), cfg, r.URL.Query().Get(sequenceNumberParam))
-			if err != nil {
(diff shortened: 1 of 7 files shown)


streamer45:
I have to ask, does this mean [MM-38327](https://mattermost.atlassian.net/browse/MM-38327) is getting addressed as well?

agnivade:
Hah, I was about to mention that. I guess that has to be done :) I will see what I can do. Would you prefer to block this PR until that gets ready? Let me know. 0/5.

streamer45:
> Hah, I was about to mention that. I guess that has to be done :) I will see what I can do. Would you prefer to block this PR until that gets ready? Let me know. 0/5.

That would be probably ideal as it would force us to do it. Either way I think we should make sure the load-test gets updated prior to whatever release this goes into gets cut.

agnivade:
Sounds good to me. :+1:

streamer45:
Just to confirm, this change would still allow non-compatible clients to successfully connect?

agnivade:
Yes, it would fall back to creating a new connection every time. Basically the old load-test case where we saw high mem consumption due to dead queues not getting reused and clogging up until they were cleaned.

streamer45 (approved):

agnivade:
Gentle ping @mgdelacroix

mgdelacroix (approved):
LGTM! :+1:

PR #19049, opened 2021-11-25
[MM-40345] - Fix telemetry in resend invite email job

AGMETEOR:
#### Summary
The telemetry 'interval' for the re-invite after 24 hours and the second re-invites (48 and 72) is matching these durations properly.

#### Ticket Link
[MM-40345](https://mattermost.atlassian.net/browse/MM-40345)

#### Release Note

```release-note
NONE

```

### DIFF
--- a/jobs/resend_invitation_email/worker.go
+++ b/jobs/resend_invitation_email/worker.go
@@ -98,7 +98,7 @@ func (rseworker *ResendInvitationEmailWorker) DoJob_24_72(job *model.Job) {
 	rseworker.Execute(job, elapsedTimeSinceSchedule, DurationInMillis_24, DurationInMillis_72, "24", "72")
 }
 
-func (rseworker *ResendInvitationEmailWorker) Execute(job *model.Job, elapsedTimeSinceSchedule, firstDuration, secondDuration int64, firstInterval, secondInterval string) {
+func (rseworker *ResendInvitationEmailWorker) Execute(job *model.Job, elapsedTimeSinceSchedule, firstDuration, secondDuration int64, firstDurationTelemetryValue, secondDurationTelemetryValue string) {
 	systemValue, sysValErr := rseworker.App.Srv().Store.System().GetByName(job.Id)
 	if sysValErr != nil {
 		if _, ok := sysValErr.(*store.ErrNotFound); !ok {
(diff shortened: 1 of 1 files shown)


jobs/resend_invitation_email/worker.go
```diff
@@ -90,15 +90,15 @@ func (rseworker *ResendInvitationEmailWorker) DoJob_24(job *model.Job) {
...
 }
 
 func (rseworker *ResendInvitationEmailWorker) DoJob_24_72(job *model.Job) {
 	elapsedTimeSinceSchedule, DurationInMillis_24, _, DurationInMillis_72 := rseworker.GetDurations(job)
-	rseworker.Execute(job, elapsedTimeSinceSchedule, DurationInMillis_24, DurationInMillis_72)
+	rseworker.Execute(job, elapsedTimeSinceSchedule, DurationInMillis_24, DurationInMillis_72, "24", "72")
 }
 
-func (rseworker *ResendInvitationEmailWorker) Execute(job *model.Job, elapsedTimeSinceSchedule, firstDuration, secondDuration int64) {
+func (rseworker *ResendInvitationEmailWorker) Execute(job *model.Job, elapsedTimeSinceSchedule, firstDuration, secondDuration int64, firstInterval, secondInterval string) {
```
marianunez, line 101: Having `firstDuration` and `firstInterval` is a bit confusing and not readable here. Do we need these two and if we do, could there be something we can do to differentiate them with a more meaningful name?

AGMETEOR: Yes we need both. `firstDuration` and `secondDuration` just represent the durations after which the reminders should be sent. For example, when we have flag set to `48` , when the job runs, it'll first send invitation after firstDuration, in this case we decided it would be 24 hours and then it'll run again after  secondDuration which in this case is 48 hours before stopping the job for this particular invite.

For `firstInterval` and `secondInterval` args, they're just for telemetry. For example, knowing that we resend invitations after 24 hours and again after 48 hours their values would `firstInterval="24"` and `secondInterval=48`. We use these values in the email template invite link to know which of the reminders the user used to join. 24? or 48? 

For readability, I can find a more differentiating name.

marianunez (approved):

jobs/resend_invitation_email/worker.go
AGMETEOR: Agreed.

nickmisasi (approved):

PR #19090, opened 2021-11-26
build: add mmctl to arm64 packages

cpanato:
#### Summary
- now we are generating and making that public the amr64 binary for linux/osx for mmctl
so now we can include that in the arm64 package as well

/cc @jasonblais another step to have the mattermost available in arm64 with the same components we have in the amd64

#### Release Note

```release-note
build: add mmctl to arm64 packages
```

### DIFF
--- a/build/release.mk
+++ b/build/release.mk
@@ -159,13 +159,14 @@ else
 	cp $(GOBIN)/$(CURRENT_PACKAGE_ARCH)/$(MM_BIN_NAME) $(DIST_PATH_GENERIC)/bin # from cross-compiled bin dir
 endif
 
-ifeq ("darwin_arm64","$(CURRENT_PACKAGE_ARCH)")
-	echo "No plugins or mmctl yet for $(CURRENT_PACKAGE_ARCH) platform, skipping..."
-else ifeq ("linux_arm64","$(CURRENT_PACKAGE_ARCH)")
-	echo "No plugins or mmctl yet for $(CURRENT_PACKAGE_ARCH) platform, skipping..."
-else
 	#Download MMCTL for $(MMCTL_PLATFORM)
 	scripts/download_mmctl_release.sh $(MMCTL_PLATFORM) $(DIST_PATH_GENERIC)/bin
+
+ifeq ("darwin_arm64","$(CURRENT_PACKAGE_ARCH)")
+	echo "No plugins yet for $(CURRENT_PACKAGE_ARCH) platform, skipping..."
+else ifeq ("linux_arm64","$(CURRENT_PACKAGE_ARCH)")
+	echo "No plugins yet for $(CURRENT_PACKAGE_ARCH) platform, skipping..."
+else
 	@# Prepackage plugins
 	@for plugin_package in $(PLUGIN_PACKAGES) ; do \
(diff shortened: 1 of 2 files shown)


build/release.mk
```diff
@@ -159,13 +159,14 @@ else
 	cp $(GOBIN)/$(CURRENT_PACKAGE_ARCH)/$(MM_BIN_NAME) $(DIST_PATH_GENERIC)/bin # from cross-compiled bin dir
 endif
 
+	#Download MMCTL for $(MMCTL_PLATFORM)
+	scripts/download_mmctl_release.sh $(MMCTL_PLATFORM) $(DIST_PATH_GENERIC)/bin
```
ashishbhate, line 163: I'm not sure I understand, this script only downloads the `amd64` version of mmctl, but we want the arm64 version ?

jasonblais:
cc @nab-77 for visibility

build/release.mk
cpanato: @ashishbhate we are working to start releasing linux arm64 packages as well, so this is needed

ashishbhate (approved):

mgdelacroix (approved):

mtardy:
Just for information, it broke the Makefile when it try to install `mmctl` because it relies on the `uname` output, and uname without specific flags only return `Linux` or `Darwin`.

cpanato:
> Just for information, it broke the Makefile when it try to install `mmctl` because it relies on the `uname` output, and uname without specific flags only return `Linux` or `Darwin`.
thanks for this, are you willing to open a pr to fix that?

mtardy:
I don't really understand your whole Makefile logic because it's quite complex, I was just using `make test-data` and I understood that this change broke the installation of `mmctl`. I avoid the issue by calling the install script manually with the right arguments that override the output of uname in the `$PLATFORM` var before running the Makefile in my script.

We could actually just fix the auto detection by processing a little bit the `uname` output but it won't be very nice. If you think it's a good idea, I can help you :).

But I lack a bit of context maybe, why did you choose the `Linux-AMD64` string for example? And you made this change because you call this script elsewhere with arguments?

cpanato:
i will take a look on that

mtardy:
Spin a Linux virtual machine, or just use your computer if you are running Linux or macOS and start the script without arguments. It should directly output:
```
Downloading prepackaged binary: https://releases.mattermost.com/mmctl/master
error downloading mmctl: can't detect OS
```
because naked `uname` on Linux will return `Linux` and `Darwin` on macOS, so the switch case will default. You can make use of `uname -m` that typically return `x86_64` on OS running on AMD64 processors I guess!

If you need any help I would be glad to participate, I just discovered this issue because I'm developing a plugin and was writing a vagrant provision script to install the environment and by `git cloning` the Mattermost server repo, it broke the Makefile rule `make test-data` that was downloading `mmctl` :)!

cpanato:
@mtardy start which script?

mtardy:
I'm so sorry, I mean `scripts/download_mmctl_release.sh`. Maybe we can have a chat somewhere, I would be glad to help you on that :)!

cpanato:
ok, fixing, just need to confirm one thing for mac M1

sorry for that breaking your workflow

PR #19096, opened 2021-11-29
MM-38150 Adding playbooks permissions

crspeller:
#### Summary
Adding roles and permissions for playbooks. None of these are used on the server. They will be used in an upcoming PR for the playbooks product.

This also adds a function to the plugin API to handle checking if a set of roles grant a specific permission. `RolesGrantPermission`

Permissions added:
- PermissionPublicPlaybookCreate
- PermissionPublicPlaybookManageProperties
- PermissionPublicPlaybookManageMembers
- PermissionPublicPlaybookView
- PermissionPublicPlaybookMakePrivate
- PermissionPrivatePlaybookCreate
- PermissionPrivatePlaybookManageProperties
- PermissionPrivatePlaybookManageMembers
- PermissionPrivatePlaybookView
- PermissionPrivatePlaybookMakePublic
- PermissionRunCreate
- PermissionRunManageProperties
- PermissionRunManageMembers
- PermissionRunView

Roles added:
- PlaybookAdmin
- PlaybookMember
- RunAdmin
- RunMember

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-38504

#### Release Note
```release-note
NONE
```

### DIFF
--- a/app/permissions_migrations.go
+++ b/app/permissions_migrations.go
@@ -915,6 +915,20 @@ func (a *App) getAddTestEmailAncillaryPermission() (permissionsMap, error) {
 	return transformations, nil
 }
 
+func (a *App) getAddPlaybooksPermissions() (permissionsMap, error) {
+	transformations := []permissionTransformation{}
+
+	transformations = append(transformations, permissionTransformation{
+		On: isRole(model.SystemUserRoleId),
+		Add: []string{
+			model.PermissionPublicPlaybookCreate.Id,
+			model.PermissionPrivatePlaybookCreate.Id,
+		},
+	})
+
+	return transformations, nil
+}
+
 // DoPermissionsMigrations execute all the permissions migrations need by the current version.
 func (a *App) DoPermissionsMigrations() error {
 	return a.Srv().doPermissionsMigrations()
@@ -953,6 +967,7 @@ func (s *Server) doPermissionsMigrations() error {
 		{Key: model.MigrationKeyAddAboutSubsectionPermissions, Migration: a.getAddAboutSubsectionPermissions},
 		{Key: model.MigrationKeyAddReportingSubsectionPermissions, Migration: a.getAddReportingSubsectionPermissions},
 		{Key: model.MigrationKeyAddTestEmailAncillaryPermission, Migration: a.getAddTestEmailAncillaryPermission},
+		{Key: model.MigrationKeyAddPlaybooksPermissions, Migration: a.getAddPlaybooksPermissions},
 	}
 
 	roles, err := s.Store.Role().GetAll()
--- a/app/permissions_migrations.go
+++ b/app/permissions_migrations.go
@@ -929,6 +929,24 @@ func (a *App) getAddPlaybooksPermissions() (permissionsMap, error) {
 		},
 	})
 
+	transformations = append(transformations, permissionTransformation{
+		On: isRole(model.SystemAdminRoleId),
+		Add: []string{
+			model.PermissionPublicPlaybookManageProperties.Id,
+			model.PermissionPublicPlaybookManageMembers.Id,
+			model.PermissionPublicPlaybookView.Id,
+			model.PermissionPublicPlaybookMakePrivate.Id,
+			model.PermissionPrivatePlaybookManageProperties.Id,
+			model.PermissionPrivatePlaybookManageMembers.Id,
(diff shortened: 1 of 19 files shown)


app/permissions_migrations.go
```diff
@@ -915,6 +915,41 @@ func (a *App) getAddTestEmailAncillaryPermission() (permissionsMap, error) {
...
+
+	transformations = append(transformations, permissionTransformation{
+		On: permissionOr(
+			permissionExists(model.PermissionCreatePublicChannel.Id),
+			permissionExists(model.PermissionCreatePrivateChannel.Id),
+		),
+		Add: []string{
+			model.PermissionPublicPlaybookCreate.Id,
+			model.PermissionPrivatePlaybookCreate.Id,
+		},
```
cpoile, line 929: This means: if you have permissions to create a public/private channel, you have permissions to create a public or private playbook?

cpoile (approved):
👍

model/role.go
```diff
@@ -807,6 +817,61 @@ func MakeDefaultRoles() map[string]*Role {
...
+			PermissionRunManageProperties.Id,
+		},
+		SchemeManaged: true,
+		BuiltIn:       true,
+	}
+
+	roles[RunMemberRoleId] = &Role{
+		Name:        RunMemberRoleId,
+		DisplayName: "authentication.roles.playbook_member.name",
+		Description: "authentication.roles.playbook_member.description",
```
mkraft, line 867:
```suggestion
		Description: "authentication.roles.run_member.description",
```

store/sqlstore/scheme_store.go
```diff
@@ -163,6 +166,58 @@ func (s *SqlSchemeStore) createScheme(scheme *model.Scheme, transaction *sqlxTxW
...
+		}
+		scheme.DefaultRunAdminRole = savedRole.Name
+
+		// run member role
+		runMemberRole := &model.Role{
+			Name:          model.NewId(),
+			DisplayName:   fmt.Sprintf("Run Member Role for Scheme %s", scheme.Name),
+			Permissions:   defaultRoles[model.RunMemberRoleId].Permissions,
+			SchemeManaged: true,
+		}
```
mkraft, line 215: I think these all need the `BuiltIn` field set to `true`.

mkraft (changes requested):

app/permissions_migrations.go
```diff
@@ -915,6 +915,41 @@ func (a *App) getAddTestEmailAncillaryPermission() (permissionsMap, error) {
...
+			model.PermissionPublicPlaybookMakePrivate.Id,
+			model.PermissionPrivatePlaybookManageProperties.Id,
+			model.PermissionPrivatePlaybookManageMembers.Id,
+			model.PermissionPrivatePlaybookView.Id,
+			model.PermissionPrivatePlaybookMakePublic.Id,
+			model.PermissionRunCreate.Id,
+			model.PermissionRunManageProperties.Id,
+			model.PermissionRunManageMembers.Id,
+			model.PermissionRunView.Id,
+		},
```
mkraft, line 947:
```suggestion
		Add: model.PlaybookScopedPermissions,
```

mkraft (changes requested):

app/permissions_migrations.go
crspeller: Yup. That's the intent.

store/sqlstore/scheme_store.go
crspeller: Do they? The existing ones here don't have that set. I set it for the system scheme ones because the others there where set. Checking usage it doesn't seem to actually effect anything anyway so I am not sure the intent of the field.

app/permissions_migrations.go
crspeller: That's missing run scoped. Also these sets of permissions could diverge in the future. An additional migration would then be added.

store/sqlstore/scheme_store.go
mkraft: Actually, my mistake, I take it back. When you're creating a scheme via the store it should be `false`. I was thinking of a different context when I made that comment.

app/permissions_migrations.go
mkraft: I see, the concern is adding a new permission to the slice in the future but that you don't want added as part of this migration here. That's valid.

mkraft (approved):

PR #19103, opened 2021-11-29
MM-39433: Fixes (*App).getEmbedsAndImages parameter.

mkraft:
#### Summary

Fixes a parameter used with an invocation of `(*App).getEmbedsAndImages` which led to the re-retrieval of OpenGraph data from external servers because [the database lookup](https://github.com/mattermost/mattermost-server/blob/master/app/post_metadata.go#L505-L510) was being skipped.

#### Ticket Link

https://mattermost.atlassian.net/browse/MM-39433

#### Release Note

```release-note
Fixed slow channel loading for instances with website link previews enabled.
```

### DIFF
--- a/app/post_metadata.go
+++ b/app/post_metadata.go
@@ -132,7 +132,7 @@ func (a *App) PreparePostForClient(originalPost *model.Post, isNewPost, isEditPo
 
 func (a *App) PreparePostForClientWithEmbedsAndImages(originalPost *model.Post, isNewPost, isEditPost bool) *model.Post {
 	post := a.PreparePostForClient(originalPost, isNewPost, isEditPost)
-	post = a.getEmbedsAndImages(post, true)
+	post = a.getEmbedsAndImages(post, isNewPost)
 	return post
 }
 


enahum (approved):
Yay! Great find

cpoile (approved):

mkraft:
I went ahead and merged this in so that we can see the effect on community-daily. Deferring QA.

agnivade (commented):
Hey Martin, thanks for the PR!

- Would be good to add a unit test for this.
- Also it would be good to pre-compile the regex to improve performance further.

mkraft:
> Hey Martin, thanks for the PR!
> 
> * Would be good to add a unit test for this.
> * Also it would be good to pre-compile the regex to improve performance further.

I started trying to unit test this but it seems like it'll require some refactoring to make that feasible, which makes more sense to happen separately. I'm adding a "test that opengraph data isn't re-requested" ticket, though.

agnivade:
Thanks Martin. Let's also add the part about the regex improvement in that ticket.

mkraft:
> 

That should be a separate story: https://mattermost.atlassian.net/browse/MM-40405.

PR #19122, opened 2021-12-03
Remove dead struct ManifestExecutables

hanzei:
#### Summary
`ManifestExecutables` is not longer used and has caused some confusion: https://github.com/mattermost/mattermost-server/pull/18865

This PR removes it. While this changes does break semver, it's very unlikely that a consumer of `model` uses this type.

#### Ticket Link
None

#### Release Note
```release-note
Removed dead struct ManifestExecutables
```


larkox (approved):

lieut-data (approved):
cc @agnivade on the semver interpretation (which makes sense to me)

agnivade (commented):
Is the problem here that it just causes confusion? In that case, could we just mark this struct as deprecated with a comment? I'd be leaning towards not removing it unless necessary. 

It's unfortunate that we missed this but it is what it is.

hanzei:
Yes, the reason for deleting is to avoid confusion. The current comment already marks is as `legacy`: `ManifestExecutables is a legacy structure capturing a subet of the known platform executables.`. And still it caused confusion as in https://github.com/mattermost/mattermost-server/pull/18865.

Do you have a suggestion for an every stronger deprecation warning?

agnivade:
We could prefix the comment with the string `// Deprecated: ` as that is a known convention and godoc will detect that and hide the struct. Example: https://pkg.go.dev/net/http#ProtocolError.

model/manifest.go
```diff
@@ -206,7 +206,8 @@ type ManifestServer struct {
 	Executable string `json:"executable" yaml:"executable"`
 }
 
-// ManifestExecutables is a legacy structure capturing a subet of the known platform executables.
+// Deprecated: ManifestExecutables is a legacy structure capturing a subet of the known platform executables.
```
agnivade, line 209: Nit: typo on "subset".

agnivade (approved):
Thanks Ben

amyblais:
@hanzei Is this needed for v6.2, or can it go into v6.3? (I don't have context on this PR)

hanzei:
@amyblais This can go into v6.3.

Issue #9822, opened 2018-11-12
Improve email headers.


MrSorcus:
#### Summary
Rspamd assign some spam points based on few email headers.

#### Steps to reproduce
Version 5.4.0.
Just get notification and look at source of email.

#### Expected behavior
`MIME-version` and `Message-ID` must be correct.

#### Observed behavior (that appears unintentional)
![screen shot 2018-11-12 at 09 00 41](https://user-images.githubusercontent.com/23472035/48337114-f57eb600-e659-11e8-8b3f-6cbef8714964.png)
Screenshot from Rspamd. Here 2.90 / 15. If it be 5.90 / 15, soft reject will be applied.
If 15 / 15 - hard reject.
`Mime-Version: 1.0` - here MIME must be capitalized (see https://tools.ietf.org/html/rfc4021#section-2.2.1)
`Message-ID` - must be received from Mattermost server (see https://tools.ietf.org/html/rfc4021#section-2.1.8)

amyblais:
Hi @MrSorcus,

Giving you an update that I'm working on investigating this,

amyblais:
I created a ticket for this here: https://mattermost.atlassian.net/browse/MM-13012.

amyblais:
Hi @MrSorcus it looks like a PR was already opened in the gomail library for the MIME issue: https://github.com/go-gomail/gomail/pull/127.

For any further issues, it seems it would make sense to open an Issue in the gomail repo: https://github.com/go-gomail/gomail/issues. Would you be able to help with that?

MrSorcus:
> Would you be able to help with that?

Looks that here issue exists - https://github.com/go-gomail/gomail/issues/36
But it closed.

MrSorcus:
If i understand correctly, there is only one way to implement it. In mattermost-server.

jamiew0w:
Is there any movement on this? Looked at the jira but appears dead. Having issues with everything arriving in spam despite _everything_ being setup correctly.

compilenix:
Here is an updated state:
![Screenshot from 2022-02-19 14-56-28](https://user-images.githubusercontent.com/6356150/154803871-cb556641-686f-4cbc-9b44-2ae1c15163b1.png)

amyblais:
Closing as this is an old issue and can be tracked via https://github.com/go-gomail/gomail/pull/127.

PR #19124, opened 2021-12-03
[MM-13012][GH-9822] add Message-ID to Emails

Garionion:
#### Summary
This PR adds the necessary Message-ID to emails as requested in #9822

#### Ticket Link
[MM-13012](https://mattermost.atlassian.net/browse/MM-13012)

#### Release Note

```
Added a Message-ID to Emails
```


jespino (approved):
Looks good to me. Thanks @Garionion! :tada:

shared/mail/mail.go
```diff
@@ -350,3 +354,15 @@ func SendMail(c smtpClient, mail mailData, date time.Time) error {
 
 	return nil
 }
+
+func (mail mailData) generateMessageID() string {
+	id := uuid.NewV4()
```
mkraft, line 359: Can we switch this to use the post id rather than a random id? I believe that'll help with proper threading, as per https://github.com/mattermost/mattermost-server/pull/18232.

mkraft (changes requested):

shared/mail/mail.go
Garionion: but where would i get the post id from?

mkraft: It would need to be passed in from the nearest context that contains it. You can see it done in the PR I hyperlinked.

PR #19128, opened 2021-12-06
[MM-40485] Enable receiving binary websocket messages

streamer45:
#### Summary

Our websocket handler currently expects JSON only messages which are text based by nature. This can be problematic and inefficient if we need to send binary data (e.g. to allow for compression). PR adds support for receiving binary (messagepack encoded) content in websocket requests.

#### Ticket Link

https://mattermost.atlassian.net/browse/MM-40485

#### Release Note

```release-note
Added server support for receiving binary (messagepack encoded) websocket messages.
```

### DIFF
--- a/api4/websocket_test.go
+++ b/api4/websocket_test.go
@@ -234,6 +234,50 @@ func TestWebSocketReconnectRace(t *testing.T) {
 	wg.Wait()
 }
 
+func TestWebSocketSendBinary(t *testing.T) {
+	th := Setup(t).InitBasic()
+	defer th.TearDown()
+
+	client := th.CreateClient()
+	th.LoginBasicWithClient(client)
+	WebSocketClient, err := th.CreateWebSocketClientWithClient(client)
+	require.NoError(t, err)
+	defer WebSocketClient.Close()
+	WebSocketClient.Listen()
+	resp := <-WebSocketClient.ResponseChannel
+	require.Equal(t, resp.Status, model.StatusOk)
+
+	client2 := th.CreateClient()
+	th.LoginBasic2WithClient(client2)
+	WebSocketClient2, err := th.CreateWebSocketClientWithClient(client2)
+	require.NoError(t, err)
+	defer WebSocketClient2.Close()
+	WebSocketClient2.Listen()
+	resp = <-WebSocketClient2.ResponseChannel
+	require.Equal(t, resp.Status, model.StatusOk)
+
+	time.Sleep(1000 * time.Millisecond)
+
+	WebSocketClient.SendBinaryMessage("get_statuses", nil)
+	resp = <-WebSocketClient.ResponseChannel
+	require.Nil(t, resp.Error, resp.Error)
+	require.Equal(t, resp.SeqReply, WebSocketClient.Sequence-1)
+
+	status, ok := resp.Data[th.BasicUser.Id]
+	require.True(t, ok)
+	require.Equal(t, model.StatusOnline, status)
+	status, ok = resp.Data[th.BasicUser2.Id]
+	require.True(t, ok)
+	require.Equal(t, model.StatusOnline, status)
+
+	WebSocketClient.SendBinaryMessage("get_statuses_by_ids", map[string]interface{}{
+		"user_ids": []string{th.BasicUser2.Id},
+	})
+	status, ok = resp.Data[th.BasicUser2.Id]
+	require.True(t, ok)
+	require.Equal(t, model.StatusOnline, status)
+}
+
 func TestWebSocketStatuses(t *testing.T) {
 	th := Setup(t).InitBasic()
 	defer th.TearDown()
--- a/api4/websocket_test.go
+++ b/api4/websocket_test.go
@@ -252,9 +252,6 @@ func TestWebSocketSendBinary(t *testing.T) {
 	WebSocketClient2, err := th.CreateWebSocketClientWithClient(client2)
 	require.NoError(t, err)
 	defer WebSocketClient2.Close()
-	WebSocketClient2.Listen()
-	resp = <-WebSocketClient2.ResponseChannel
-	require.Equal(t, resp.Status, model.StatusOk)
 
 	time.Sleep(1000 * time.Millisecond)
 
--- a/app/web_conn.go
+++ b/app/web_conn.go
@@ -17,6 +17,7 @@ import (
 	"time"
 
 	"github.com/gorilla/websocket"
+	"github.com/vmihailenco/msgpack/v5"
 
 	"github.com/mattermost/mattermost-server/v6/model"
 	"github.com/mattermost/mattermost-server/v6/plugin"
@@ -43,6 +44,10 @@ const (
 )
 
 const websocketMessagePluginPrefix = "custom_"
+
+type msgDecoder interface {
+	Decode(v interface{}) error
+}
 
 type pluginWSPostedHook struct {
 	connectionID string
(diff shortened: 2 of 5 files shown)


app/web_conn.go
```diff
@@ -343,8 +348,20 @@ func (wc *WebConn) readPump() {
 	})
 
 	for {
+		msgType, rd, err := wc.WebSocket.NextReader()
+		if err != nil {
+			wc.logSocketErr("websocket.read", err)
+			return
+		}
+
+		var decoder msgDecoder
```
agnivade, line 357: Nit: would prefer this to be an anonymous interface to avoid this being declared outside.

api4/websocket_test.go
```diff
@@ -234,6 +234,50 @@ func TestWebSocketReconnectRace(t *testing.T) {
...
+	require.Equal(t, resp.Status, model.StatusOk)
+
+	client2 := th.CreateClient()
+	th.LoginBasic2WithClient(client2)
+	WebSocketClient2, err := th.CreateWebSocketClientWithClient(client2)
+	require.NoError(t, err)
+	defer WebSocketClient2.Close()
+	WebSocketClient2.Listen()
+	resp = <-WebSocketClient2.ResponseChannel
+	require.Equal(t, resp.Status, model.StatusOk)
```
agnivade, line 257: Wondering what is the purpose of this section.

app/web_conn.go
```diff
@@ -343,8 +348,20 @@ func (wc *WebConn) readPump() {
 	})
 
 	for {
+		msgType, rd, err := wc.WebSocket.NextReader()
+		if err != nil {
+			wc.logSocketErr("websocket.read", err)
```
agnivade, line 353: let's have a different error message?

model/websocket_client_test.go
```diff
@@ -149,3 +150,53 @@ func TestWebSocketClose(t *testing.T) {
...
+
+	err = cli.SendBinaryMessage("binaryAction", map[string]interface{}{
+		"func": func() {},
+	})
+	require.Error(t, err)
+
+	err = cli.SendBinaryMessage("binaryAction", clientData)
+	require.NoError(t, err)
+
+	<-cli.writeChan
```
agnivade, line 201: Can you clarify why this is needed? Also, shouldn't we do `cli.Close()`?

streamer45: I was trying to make sure that the message was handled prior to exiting the test. The close happens on the server side handler upon reception of the binary message. I've updated this now.

api4/websocket_test.go
streamer45: That was needed to make sure the status for the second user is present in the response. I've simplified a bit removing the listen which isn't necessary.

agnivade (approved):
Great stuff. Thanks!

model/websocket_client.go
```diff
@@ -182,6 +184,11 @@ func (wsc *WebSocketClient) writer() {
 			switch msg.msgType {
 			case msgTypeJSON:
 				wsc.Conn.WriteJSON(msg.data)
+			case msgTypeBinary:
+				data, ok := msg.data.([]byte)
+				if ok {
```
isacikgoz, line 189:
```suggestion
				if data, ok := msg.data.([]byte); ok {
```

isacikgoz (approved):
LGTM, just a suggestion 🤗

streamer45:
@agnivade , @isacikgoz Thanks for the quick review. 

I was testing this for a while and I was still getting conversion issues. Luckily, after some tedious debugging, I have finally figured out what's going on. Prior to send the websocket request to a plugin we clone the struct which essentially does a JSON marshal/unmarshal, which means `[]byte` gets converted to a base64 string (slightly defeating the original purpose of this PR). I can add a separate method/function to call only in case of a binary message although it seems to me like doing the clone with MessagePack may just be better for both cases. Please let me know what you think:

```
goos: linux
goarch: amd64
pkg: github.com/mattermost/mattermost-server/v6/model
cpu: Intel(R) Core(TM) i7-8565U CPU @ 1.80GHz
BenchmarkWebSocketRequestClone/json_based_clone_empty-8         	         5350746	      1110 ns/op	     472 B/op	       6 allocs/op
BenchmarkWebSocketRequestClone/json_based_clone_text_based-8    	          362662	     16433 ns/op	    6082 B/op	      44 allocs/op
BenchmarkWebSocketRequestClone/json_based_clone_binary_based-8  	          321562	     18513 ns/op	    8716 B/op	      33 allocs/op
BenchmarkWebSocketRequestClone/messagepack_based_clone_empty-8  	         7564702	     799.9 ns/op	     368 B/op	       4 allocs/op
BenchmarkWebSocketRequestClone/messagepack_based_clone_text_based-8         	 1890672	      3180 ns/op	    5655 B/op	      19 allocs/op
BenchmarkWebSocketRequestClone/messagepack_based_clone_binary_based-8       	 2032279	      2939 ns/op	    5618 B/op	      15 allocs/op
PASS
ok  	github.com/mattermost/mattermost-server/v6/model	44.393s
```


```go
func BenchmarkWebSocketRequestClone(b *testing.B) {
	bcs := []struct {
		name string
		req  WebSocketRequest
	}{
		{
			name: "empty",
			req:  WebSocketRequest{},
		},
		{
			name: "text based",
			req: WebSocketRequest{
				Seq:    45,
				Action: "some_action_name",
				Data: map[string]interface{}{
					"key1": 0,
					"key2": "some text",
					"key3": "some more text",
					"key4": strings.Repeat("AA", 1024),
				},
			},
		},
		{
			name: "binary based",
			req: WebSocketRequest{
				Seq:    45,
				Action: "some_action_name",
				Data: map[string]interface{}{
					"type": "some string",
					"data": []byte(strings.Repeat("AA", 1024)),
				},
			},
		},
	}

	for _, ct := range []string{"json", "messagepack"} {
		for _, bc := range bcs {
			b.Run(ct+"_based_clone_"+bc.name, func(b *testing.B) {
				for i := 0; i < b.N; i++ {
					if ct == "json" {
						bc.req.Clone()
					} else {
						bc.req.CloneBinary()
					}
				}
			})
		}
	}
}
```

agnivade:
Sounds good!

isacikgoz:
Agreed, seems like a good improvement 👍

agnivade:
Just ensure to run an e2e on this as a sanity check.

streamer45:
> Just ensure to run an e2e on this as a sanity check.

https://automation-dashboard.vercel.app/cycles/7dee5810-58c6-45af-a85c-1d3757a4b13c :stuck_out_tongue:

streamer45:
@furqanmlk This change should not affect functionality. I've added tests to make sure everything works as expected and also ran E2E tests which passed at 99% ([full results](https://automation-dashboard.vercel.app/cycles/7dee5810-58c6-45af-a85c-1d3757a4b13c)).

image

furqanmlk (approved):
Thanks @streamer45

streamer45:
> @streamer45 - Are product documentation updates needed for this PR for System Admins or end users?

@cwarnermm 
Probably neither. This would be useful for developers looking to extend (e.g. plugins) or work on top of server's API.

cwarnermm:
Great context, @streamer45. Thank you! Given the value to developers, would you be open to suggesting an update to developers.mattermost.com?

agnivade:
IMO this is an exposed method which already appears in the package documentation at https://pkg.go.dev/github.com/mattermost/mattermost-server/v6/model@master#WebSocketClient.SendBinaryMessage. We can probably add one line in  https://api.mattermost.com/#tag/WebSocket saying we support binary websocket messages, but for everything else, refer to this documentation.

cwarnermm:
@agnivade I agree. Thank you!

cwarnermm:
@agnivade - Can you or a member of your team create the documentation PR for the api repository, please?

agnivade:
Can do!

PR #19135, opened 2021-12-07
MM-40143: Out of bounds error in *Channel.GetOtherUserIdForDM

agnivade:
We check if there are indeed 2 user ids after splitting the string
at "__".

https://mattermost.atlassian.net/browse/MM-40143

```release-note
NONE
```

### DIFF
--- a/model/channel.go
+++ b/model/channel.go
@@ -290,6 +290,10 @@ func (o *Channel) GetOtherUserIdForDM(userId string) string {
 
 	userIds := strings.Split(o.Name, "__")
 
+	if len(userIds) != 2 {
+		return ""
+	}
+
 	var otherUserId string
 
 	if userIds[0] != userIds[1] {


agnivade:
This seems to come from a manual DB edit.

The only scenario this can happen is if the user manually changed the "town-square" channel type to a DM channel. I am willing to cancel this PR since this seems like a corner case not worth handling.

Let me know what you guys think.

streamer45:
> This seems to come from a manual DB edit.
> 
> The only scenario this can happen is if the user manually changed the "town-square" channel type to a DM channel. I am willing to cancel this PR since this seems like a corner case not worth handling.
> 
> Let me know what you guys think.

It's almost a coin toss for me. On one hand I understand that messing with DB can cause anything to happen and we can't realistically prevent that but from a purely code perspective if the input is not what we expect it to be we should probably return an error. I understand this is even more work given the current function doesn't really do that but if we decide to handle this case it would be worth logging a warning to inform the admin something is really wrong with their data.

agnivade:
Yep, understood. I think I have a slightly different perspective here :stuck_out_tongue: 

I don't believe our code should try to handle cases which it's not "meant to". The database is meant to be only managed by the software, so it should only handle scenarios that comes by data written from the software, not by manual editing. There can be tons of other edge-cases where our software might break if you tweak the DB in a certain way.

Coming to the point of alerting the admins, I think throwing a panic is a better way of letting the admin know that there's something seriously wrong, rather than logging just a warning, which in 99% of the times, they just ignore.

So for edge-cases like this, I'd prefer to just let it crash and have the admin come to us, rather than us trying to handle all possible sorts of data manipulations.

Thoughts?

streamer45:
That's fair reasoning and again I don't think there's the perfect solution here so I guess the best we can hope for is consistency.
As long as panic only causes the single request to fail meaning the impact on users is minimal/none I think it's fine to leave it as is. Let's just try to apply the same approach to similar occurrences in the future (maybe something to document?).

agnivade:
Yep, AFAIR we haven't explicitly tried to handle bad DB data before. So we should be consistent.

I'll queue an item in the dev meeting to discuss more.

hmhealey:
I could see taking this since checking the size of that array is standard enough. I think we even do that on the web app even though this won't usually cause a crash because JavaScript.

That said, I do agree that we probably shouldn't handle every single way someone could mess up their database

agnivade:
Yep, there's no right answer for sure. Let's wait until the dev meeting to get more feedback around this.

agnivade:
Closing this as per discussion in meeting. This seems like an application logic which can be easily added to integrity checker. I can create an HW ticket for this.

PR #19136, opened 2021-12-07
MM-40469: Handle invalid response codes from plugins

agnivade:
This was an interesting crash detected via Sentry.

Typically, any HTTP status code outside 100-999 range
will cause a crash in the HTTP server. And `ExecuteCommand`
is the only plugin hook which returns a model.AppError
instead of error.

So an incorrect plugin implementation could return
a status code of 0, and crash the server. We handle that
by rewriting any illegal response code to 500.

https://mattermost.atlassian.net/browse/MM-40469

```release-note
NONE
```

### DIFF
--- a/app/plugin_commands.go
+++ b/app/plugin_commands.go
@@ -13,6 +13,7 @@ import (
 
 	"github.com/mattermost/mattermost-server/v6/app/request"
 	"github.com/mattermost/mattermost-server/v6/model"
+	"github.com/mattermost/mattermost-server/v6/shared/mlog"
 )
 
 type PluginCommand struct {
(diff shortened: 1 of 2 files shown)


app/plugin_commands.go
```diff
@@ -164,5 +164,11 @@ func (a *App) tryExecutePluginCommand(c *request.Context, args *model.CommandArg
 		errMessage := fmt.Sprintf("err= Plugin %s crashed due to /%s command", matched.PluginId, trigger)
 		return matched.Command, nil, model.NewAppError("ExecutePluginCommand", "model.plugin_command_crash.error.app_error", map[string]interface{}{"Command": trigger, "PluginId": matched.PluginId}, errMessage, http.StatusInternalServerError)
 	}
+	// This is a response from the plugin, which may set an incorrect status code;
+	// e.g setting a status code of 0 will crash the server. So we always bucket everything under 500.
+	if appErr != nil && (appErr.StatusCode < 100 || appErr.StatusCode > 999) {
+		appErr.StatusCode = http.StatusInternalServerError
```
lieut-data, line 170: Should we log a warning here for potential debugging?

lieut-data (approved):
Nice -- and lovely unit tests :)

This reminds me of a similar crash in the past: https://github.com/mattermost/mattermost-server/pull/11276.

hanzei (approved):
Thanks for the fix, Agniva :+1:

app/plugin_commands.go
agnivade: done

amyblais:
@agnivade Is it recommended to add this to v6.2?

agnivade:
Not needed. Thanks Amy.

PR #19145, opened 2021-12-10
Server support for collapse consecutive preference

3ach:
#### Summary

This adds support for a new preference (`collapse_consecutive_messages`) which the client will use in a separate pull request to change how it displays several consecutive messages from the same user. This is push-compatible in all cases *except* when a new client sends the setting to an old server. 

#### Ticket Link

I don't have a ticket link, just a feature request: https://mattermost.uservoice.com/forums/306457-general/suggestions/44543601-add-a-ui-option-to-always-show-the-username-of-a-p

#### Release Note

```release-note
Added a new preferences which can be updated from the API, `collapse_consecutive_messages`. 
```

### DIFF
--- a/app/export.go
+++ b/app/export.go
@@ -55,6 +55,9 @@ var exportablePreferences = map[ComparablePreference]string{{
 }: "MessageDisplay", {
 	Category: model.PreferenceCategoryDisplaySettings,
 	Name:     "channel_display_mode",
+}: "CollapseConsecutive", {
+	Category: model.PreferenceCategoryDisplaySettings,
+	Name:     "collapse_consecutive_messages",
 }: "ChannelDisplayMode", {
 	Category: model.PreferenceCategoryTutorialSteps,
 	Name:     "",
--- a/app/import_functions.go
+++ b/app/import_functions.go
@@ -635,6 +635,15 @@ func (a *App) importUser(data *UserImportData, dryRun bool) *model.AppError {
 		})
 	}
 
+	if data.CollapseConsecutive != nil {
+		preferences = append(preferences, model.Preference{
+			UserId:   savedUser.Id,
+			Category: model.PreferenceCategoryDisplaySettings,
+			Name:     model.PreferenceNameCollapseConsecutive,
+			Value:    *data.CollapseConsecutive,
+		})
+	}
+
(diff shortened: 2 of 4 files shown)


esethna:
@3ach thanks for the PR! Do you have a screenshot of the before and after of what you're trying to accomplish here? 

I'm wondering if this needs to be a preference or whether this could just be a change we make to compact mode for everyone. Do you know of users on both sides of the fence who would prefer a setting?

3ach:
Sorry for the delay!
Here is the current behavior:


And here is what I'd like to happen:


I can't speak to all compact-mode users. I know that most of the folks I know who have opted into compact mode are coming from IRC clients which usually behave this way, and would prefer to have the display name every time. That's definitely not a representative sample.

3ach:
If you want to make this the normal behavior of compact mode, that's fine by me -- I'll close this PR and put up a revised one for webapp (it's *so* much simpler to always do this)

3ach:
https://github.com/3ach/mattermost-webapp/commit/9e44f76d4219fe9ac5fdc1b772ab4e6720747a4b is the commit to do this always, versus this change, which needs two client side commits: [modifying](https://github.com/3ach/mattermost-webapp/commit/1b6c5c02b5b3d7fbd06ce26049d08ebfcb8d43a5) and [using](https://github.com/3ach/mattermost-webapp/commit/311f844dc2084bc475d60b7b7a453f528daf00c8) the preference.

esethna:
@3ach I'd propose we just make this change to compact view always (no setting). I think you're right that the audience who prefers compact view will expect IRC style chat. I'll open a ticket and you can make the change to the webapp only

esethna:
@3ach thoughts on colorizing usernames in compact view?

3ach:
@esethna colorizing usernames would be great (also an expectation of most IRC users ;) ) and I'm happy to put in the patch.

3ach:
https://github.com/mattermost/mattermost-webapp/pull/9668 for that. I'll close this PR.

esethna:
> @esethna colorizing usernames would be great (also an expectation of most IRC users ;) ) and I'm happy to put in the patch.

Ticket opened here! https://mattermost.atlassian.net/browse/MM-41115

PR #19156, opened 2021-12-14
[MM-40560] Add Server locale to link preview request

BenHargreaves:
#### Summary

This PR adds the `Accept-Language` header to the Link Preview request, with its value set to the Default Server Language. If the external server allows this header, it should return content in the language specified.

There aren't any UI changes involved in this PR - but heres a screenshot of the end result for reference:

|  Link preview with server localization settings  | 
|----|
| mattermost-link-preview-localized | 

#### Ticket Link

https://mattermost.atlassian.net/browse/MM-40560

#### Release Note

```release-note
Add Accept-Language header to generate link previews in the Default Server language
```

### DIFF
--- a/app/post_metadata.go
+++ b/app/post_metadata.go
@@ -549,6 +549,7 @@ func (a *App) getLinkMetadata(requestURL string, timestamp int64, isNewPost bool
 		} else {
 			request.Header.Add("Accept", "image/*")
 			request.Header.Add("Accept", "text/html;q=0.8")
+			request.Header.Add("Accept-Language", *a.Config().LocalizationSettings.DefaultServerLocale)
 
 			client := a.HTTPService().MakeClient(false)
 			client.Timeout = time.Duration(*a.Config().ExperimentalSettings.LinkMetadataTimeoutMilliseconds) * time.Millisecond
--- a/app/post_metadata_test.go
+++ b/app/post_metadata_test.go
@@ -754,12 +754,24 @@ func TestGetEmbedForPost(t *testing.T) {
 	server := httptest.NewServer(http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
 		if r.URL.Path == "/index.html" {
 			w.Header().Set("Content-Type", "text/html")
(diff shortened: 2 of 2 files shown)


app/post_metadata.go
```diff
@@ -549,6 +549,7 @@ func (a *App) getLinkMetadata(requestURL string, timestamp int64, isNewPost bool
 		} else {
 			request.Header.Add("Accept", "image/*")
 			request.Header.Add("Accept", "text/html;q=0.8")
+			request.Header.Add("Accept-Language", *a.Config().LocalizationSettings.DefaultServerLocale)
```
mkraft, line 552: There's also a `DefaultClientLocale`... I'm just wondering the difference.

mkraft: @hmhealey This would cause the open graph data to be persisted permanently in the given language, right? And then it would be a separate feature request if we wanted to some day support multi-language preview?

BenHargreaves: From what I can tell, `DefaultClientLocale` is the default language set for new users added to server - whereas `DefaultServerLocale` is the language which system messages are displayed in. These 2 properties seem to correspond to the settings listed in `System Console -> Site Configuration -> Localization`

hmhealey: Yeah, that's correct about how `DefaultClientLocale` and `DefaultServerLocale` are used. I don't remember why we didn't just use a single value for both, but that doesn't change much here. I think the server locale would be the correct one to use here.

I suppose we could have this use the post author's locale or even the reader's locale in the future, but yeah, I agree that that would be a separate feature. Doing this much should solve the incorrect preview locale problem for the vast majority of users.

hmhealey (approved):
This looks good to me!

mkraft (approved):
Looks great @BenHargreaves!

furqanmlk (approved):
Working as expected.
image

hmhealey:
Awesome! Thanks again for the PR, @BenHargreaves. This'll be available then in 6.4 coming out in February.

PR #19178, opened 2021-12-15
Update session cache if a plugin extends a session

hanzei:
#### Summary
If a plugin uses the ExtendSessionExpiry RPC API method to update the session expiry, the new session doesn't get pushed to the cache and hence the cache returns the old one.

This PR prevents the issue from being introduce again by adding a new `ExtendSessionExpiry` to the UserService that takes care of extending a session.

The PR also adds a plugin e2e test that covers the fix.

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-40652

#### Release Note
```release-note
NONE
```


hanzei:
I just realized that the project uses 1.16, which doesn't support `UnixMilli`. CI will fail for now. I've opened a discussion on the community server: https://community-daily.mattermost.com/core/pl/q5abug4sqjboin6qaxo1t7ct3o

app/plugin_api.go
```diff
@@ -289,7 +289,16 @@ func (api *PluginAPI) CreateSession(session *model.Session) (*model.Session, *mo
...
-	return api.app.extendSessionExpiry(sessionID, expiresAt)
+	session, err := api.app.ch.srv.userService.GetSessionByID(sessionID)
+	if err != nil {
+		return model.NewAppError("extendSessionExpiry", "app.session.get_sessions.app_error", nil, err.Error(), http.StatusInternalServerError)
+	}
+
+	if err := api.app.ch.srv.userService.ExtendSessionExpiry(session, expiresAt); err != nil {
+		return model.NewAppError("extendSessionExpiry", "app.session.extend_session_expiry.app_error", nil, err.Error(), http.StatusInternalServerError)
+	}
+
```
agnivade, line 300: Nit: might be better to keep the `*App.extendSessionExpiry` method and move these 2 calls inside that. So that from the plugin api side, things remain unchanged, but the internal app logic changes.

agnivade (approved):
Thanks Ben. Left a non-blocking comment.

isacikgoz (approved):
LGTM 👍

app/plugin_api.go
hanzei: I like of like the plugin api method as that ensure `ExtendSessionExpiry` is the only method named that way and there is no confusion on whether the app or the userService method should be called.

hanzei:
@DHaussermann The only way I could see you testing this PR is by testing is alongside https://github.com/mattermost/mattermost-plugin-apps/pull/291. The PR also contains a plugin-e2e test and hence maybe it can be merged without manually testing. What do you think?

DHaussermann:
@hanzei can you please clarify the scope or test steps here? I was able to get some testing done. But testing to the point where an installation succeeds was not possible https://github.com/mattermost/mattermost-app-zendesk/pull/90#issuecomment-1006138135   

I'm not sure what the simplest way to test just the scope of this PR is.

hanzei:
@DHaussermann You can test this PR with any app expect zendesk

DHaussermann:
@hanzei I had an issue as well running Service Now through the happy path. 
Please see here https://github.com/mattermost/mattermost-app-servicenow/pull/36#issuecomment-1007679858

hanzei:
@DHaussermann Could you please test again with the test app?

PR #19194, opened 2021-12-17
MM-40537: Bump bleve dependency to fix Sentry crash

agnivade:
We observed a sentry crash, which is now fixed upstream.
So we bump the dependency accordingly.

```release-note
NONE
```


isacikgoz (approved):
Nice 👍

PR #19208, opened 2021-12-20
used silver searcher docs and regex to make todo target more concise

mxschumacher:
(This is my first ever contribution to Mattermost, please let me know how my PR can be improved. I am trying to follow the process to the best of my understanding)

[The silver searcher](https://github.com/ggreer/the_silver_searcher), a grep/ack alternative with a CLI named "ag", is used to go through most of the files in this repo and identify lines that contain any of the following strings: TODO, XXX, FIXME, "FIX ME".

Instead of calling ag once per search term, we can use some Regex syntax to search for them during the same execution.

> (TODO|XXX|FIXME|"FIX ME")

To avoid matching on hashes that contain "XXX" (there is at least one within the current HEAD of this project, see [here](https://github.com/mattermost/mattermost-server/blob/master/tests/development-public-key.asc#L19), I have also added **"[: ]+"** to the end of the regex - meaning that any of the search terms should be either followed by  ":" or a space.

The second change I made is to make use of the fact that the **--ignore** flag takes several arguments and does not have to be repeated, please a Stack Overflow thread discussing the topic [here](https://stackoverflow.com/a/48631459/2823986). 

Furthermore, it is not necessary to differentiate between **--ignore** and **--ignore-dir**, because they are the same for ag, see relevant source [here](https://github.com/ggreer/the_silver_searcher/blob/master/doc/ag.1.md?plain=1#L93
). 

QA steps:

With the version before the change contained in this PR, run:

> make todo > ~/make-todo-output-pre-refactoring

Then with the change, run:

> make todo > ~/make-todo-output-post-refactoring

(be sure to write the output of the "make todo" command to a file outside of the mattermost-server directory, because newly created file will otherwise show up in the changes)

Then, using a diffing tool like [delta](https://github.com/dandavison/delta), cmp or diff compare the two files:

> delta make-todo-output-pre-refactoring make-todo-output-post-refactoring

#### Ticket Link

Following the rules at [Contributing to Mattermost without a Ticket](https://developers.mattermost.com/contribute/getting-started/contributions-without-ticket/) for contributions of less than 20 lines, this PR is not linked to a ticket.

#### Release Note

```release-note
NONE
```


agnivade (changes requested):
The `--ignore` pattern doesn't seem to be working. I am getting results from the vendor directory as well. Can we look into that? My `ag` version is 2.2.0.

mxschumacher:
Hello @agnivade - I cannot reproduce, running 

```bash
ag --ignore={Makefile,vendor,runtime} '(TODO|XXX|FIXME|"FIX ME")[: ]+'
```

In the shell directly ignores the listed directories correctly, if I omit the --ignore flag, I get the results from /vendor. As expected.
If you run the command above in the shell, what result do you get?

My ag version is also 2.2.0

It's the same for when I execute "make todo"

Something we could look into is aliases that might have been set in the shell environment (for example to run grep and ag commands in ripgrep, rg: https://github.com/BurntSushi/ripgrep

Relatedly - maybe the default shell that make picks is different for your and my machine? I don't think it is explicitly defined in the Makefile.

agnivade:
Yeah, there may be an issue with the shell.

If I run `ag --ignore={Makefile,vendor,runtime} '(TODO|XXX|FIXME|"FIX ME")[: ]+'` directly, it works fine. But `make todo` fails to apply the ignore pattern. So it looks like make is messing up the special characters `{}` while spawning the command. Because the earlier version of `make todo` works fine in my system.

Could we revert the `ignore` changes to use `--ignore Makefile --ignore-dir vendor --ignore-dir runtime`? That fixes the output in my case.

mxschumacher:
I have reverted back to "--ignore Makefile --ignore-dir vendor --ignore-dir runtime"

agnivade (approved):
Thank you @mxschumacher !

mxschumacher:
Hello - is there anything else required from my side to get this PR to be merged?

agnivade:
Apologies @mxschumacher. @cpanato is on holidays. I'll get it merged. Thanks for your patience!

/update-branch

agnivade:
Sorry there seems to be an issue with CI. I'll follow up and try to merge this.

agnivade:
@mxschumacher - I have raised this with our Build team here: https://community-daily.mattermost.com/core/pl/sq9hkd6m9j8kugra9mt5ox6hze. Hopefully we'll be able to merge your PR soon!

cpanato (approved):

cpanato:
@agnivade we can merge this PR the upload part is not much important in this case, the build and test are more and those passed

agnivade:
@cpanato - Yes of course. Could you merge it please? I don't have the necessary permissions :stuck_out_tongue:

PR #19239, opened 2021-12-21
Focalboard compliance export initial support

jespino:
#### Summary
Focalboard compliance export initial support

#### Ticket Link
https://community-daily.mattermost.com/boards/workspace/qgsck6cts3fwpqwyjiupjm5cde/b4ctmm917q3rzdfi451qtuxjzcw/vuswsftt1rb873datosmmbtwkpa/cy7zk5jewk3da8yechjufg3pc7r

#### Release Note
```release-note
Add initia support for compliance export of focalboard data
```


store/sqlstore/compliance_store.go
```diff
@@ -342,3 +342,55 @@ func (s SqlComplianceStore) MessageExport(cursor model.MessageExportCursor, limi
...
 }
+
+func (s SqlComplianceStore) BlocksExport(cursor model.BlockExportCursor, limit int) ([]*model.BlockExport, model.BlockExportCursor, error) {
+	var version int
+	schemaVersionRow := s.GetReplicaX().QueryRowX("select version from focalboard_schema_migrations")
+	if schemaVersionRow.Err() != nil {
+		return nil, cursor, schemaVersionRow.Err()
+	}
+	schemaVersionRow.Scan(&version)
+	if version != 16 {
```
harshilsharma63, line 353: Should this be `version < 16`? The export will continue to work in schema versions above 16 as well, right?

store/sqlstore/compliance_store.go
```diff
@@ -342,3 +342,55 @@ func (s SqlComplianceStore) MessageExport(cursor model.MessageExportCursor, limi
...
+		return nil, cursor, fmt.Errorf("invalid focalboard version to export compliance, current=%d, expected=%d", version, 16)
```
harshilsharma63, line 354: Nit: making it clearer that any version above 16 is fine.

```suggestion
		return nil, cursor, fmt.Errorf("invalid focalboard version to export compliance, current=%d, expected at least=%d", version, 16)
```

harshilsharma63: If the user sees this error, how would they know which Focalboard plugin version to upgrade to? Should we use the Focalboard version in this error message? We can use the Focalboard plugin version we're expecting to ship this feature in.

harshilsharma63 (changes requested):

sbishel (dismissed):
LGTM - But I think this will need to be updated once the PR moving blocks to their own table is merged.

store/sqlstore/compliance_store.go
wiggin77: I'm wondering the same thing.

wiggin77 (approved):
LGTM :+1:   Harshil's chg request looks right to me.

store/sqlstore/compliance_store.go
jespino: The idea is to make it work only for the right version, in this case, is the last existing version, I'm not really happy with this part, I have to rethink this.

lieut-data:

PR #19249, opened 2021-12-24
[MM-40179] Go into directories when trying to export attachments

Willyfrog:
#### Summary

Attachments are nested into directories, but ListDir was unable to add anything that was not on the same folder.

#### Ticket Link
[MM-40179](https://mattermost.atlassian.net/browse/MM-40179)

#### Release Note
```release-note
Fixed zip creation failing when adding attachments
```


ashishbhate (approved):

shared/filestore/localstore.go
```diff
@@ -188,19 +188,43 @@ func (b *LocalFileBackend) RemoveFile(path string) error {
...
 	}
-	for _, fileInfo := range fileInfos {
-		paths = append(paths, filepath.Join(path, fileInfo.Name()))
+	for _, dirEntry := range dirEntries {
+		entryName := dirEntry.Name()
+		entryPath := filepath.Join(path, entryName)
+		if entryName == "." || entryName == ".." || entryPath == path {
+			continue
+		}
+		if dirEntry.IsDir() {
```
jespino, line 209: Should we add a test case for this recursion? I Think is not tested in the existing test case and maybe Is something that we should add.

jespino:
Sorry for the late reply, but I'm struggling to understand why this change makes sense. This is an abstract storage service, so, when I list a directory I don't expect a recursive behavior (unless I request for that explicitly). The current behavior is the one that I expect from this kind of service, I can see sensible to add a `recursive` parameter to that, but I don't see sensible to change the default behavior from not-recursive to recursive. That is a huge change, that can be generate performance and functionality problems in other parts of the code.

My suggestion is to respect the current behavior, and maybe, add another method like `RecursiveListDir` or a parameter to the `ListDir`, and keep the rest of the code using the non-recursive version.

app/file.go
```diff
@@ -217,15 +217,27 @@ func (s *Server) removeFile(path string) *model.AppError {
 }
 
 func (a *App) ListDirectory(path string) ([]string, *model.AppError) {
-	return a.Srv().listDirectory(path)
+	return a.Srv().listDirectory(path, false)
 }
 
-func (s *Server) listDirectory(path string) ([]string, *model.AppError) {
+func (a *App) ListDirectoryRecursively(path string) ([]string, *model.AppError) {
```
jespino, line 223: You now added this ListDirectoryRecursively, but looks like you are not using it anywhere, I guess you need to add this to the place where you were trying to fix the lack of recursion problem.

jespino (changes requested):
Now is need to fix the underneath problem that you were trying to solve before.

app/file.go
Willyfrog: yes, I have to add it to the Enterprise repo, will try to do it today, sadly I get other priorities in the way and I'm not moving as fast as I'd like

Willyfrog:
Usage of the function provided in https://github.com/mattermost/enterprise/pull/1131

jespino:
@Willyfrog the tests are failing

Willyfrog:
sorry for the long delay in a simple change 😔 , feel free to review again

jgilliam17:
For some reason e2es didn't trigger, starting again

Willyfrog:
I'm not really sure what's broken in here. Seems it is failing to build with enterprise. I'll update both branches and see if that helps

jgilliam17:
@Willyfrog There is an error when trying to upgrade to Enterprise, something about digital signature. I asked `@sresupport` [here](https://community-daily.mattermost.com/private-core/pl/n6h3smjpsff8bb4385tqzcg61y) about it

jgilliam17:
@Willyfrog `sresupport` suggests the failed step is fixed first to see if that unblocks the build with enterprise.

Willyfrog:
@jgilliam17 seems it finally passed the tests, so it should be ready to test

jgilliam17:
I am unable to run any compliance exports on the test server, all are stuck in pending state. 
Per @furqanmlk suggestion I've also set config `MessageExportSettings.DownloadExportResults "true"` on the test server - exports are still not working.
We'll need to fix the build/troubleshoot further  - testing is blocked due to this issue.

jgilliam17:
Revisited the test server today and I was able to run compliance export, download and open the file.
Looks good to merge.

jgilliam17 (approved):

Willyfrog:
@saturninoabril is it ok to ignore the e2e ldapsync fails? are they known to fail?

saturninoabril:
Did you mean the unit test at https://app.circleci.com/pipelines/github/mattermost/enterprise/15292/workflows/f327983a-2b50-4864-ac46-cebf4b73e3e0/jobs/91492? I honestly don't know if it's a known issue since it's my first time to see such failures. Maybe @agnivade could help validate?

saturninoabril:
@Willyfrog Feel free to merge. I ran E2E for sanity check only on latest update. No issue found; same as previously validated by Jelena (💯)

agnivade:
> is it ok to ignore the e2e ldapsync fails? are they known to fail?

Obviously not :) It just needed an update-branch as you can see. :wink:

Willyfrog:
> Obviously not :) It just needed an update-branch as you can see. 😉

Haha, not that obvious, the first update-branch was what got me into that problem so I wasn't hoping reissuing it would solve it. But it did,  great :)

PR #19256, opened 2021-12-28
[MM-39583] - Intermediary web login page: Workspace Cookies

AGMETEOR:
#### Summary
The CWS has this page https://customers.mattermost.com/cloud/connect-workspace . We want to be able to autofill the company domain field based on a cookie set by a user's workspace after they logged into it. Atm, the cookie will hold a value of the last workspace the user interacted with.

#### Ticket Link
[MM-39583](https://mattermost.atlassian.net/browse/MM-39583)

#### Related
https://github.com/mattermost/customer-web-server/pull/639

#### Release Note

```release-note
NONE

```

### DIFF
--- a/api4/user.go
+++ b/api4/user.go
@@ -1854,7 +1854,7 @@ func login(c *Context, w http.ResponseWriter, r *http.Request) {
 
 	// For context see: https://mattermost.atlassian.net/browse/MM-39583
 	if c.App.Srv().License() != nil && *c.App.Srv().License().Features.Cloud {
-		c.App.AttachCWSIntermediaryLoginCookie(c.AppContext, w, r)
+		c.App.AttachCloudSessionCookie(c.AppContext, w, r)
 	}
 
 	userTermsOfService, err := c.App.GetUserTermsOfService(user.Id)
--- a/api4/user.go
+++ b/api4/user.go
@@ -1853,7 +1853,7 @@ func login(c *Context, w http.ResponseWriter, r *http.Request) {
 	}
 
 	// For context see: https://mattermost.atlassian.net/browse/MM-39583
-	if c.App.Srv().License() == nil || !*c.App.Srv().License().Features.Cloud {
+	if c.App.Srv().License() != nil && *c.App.Srv().License().Features.Cloud {
 		c.App.AttachCWSIntermediaryLoginCookie(c.AppContext, w, r)
 	}
 
--- a/api4/user.go
+++ b/api4/user.go
@@ -1853,7 +1853,9 @@ func login(c *Context, w http.ResponseWriter, r *http.Request) {
 	}
 
 	// For context see: https://mattermost.atlassian.net/browse/MM-39583
-	c.App.AttachCWSIntermediaryLoginCookie(c.AppContext, w, r)
+	if c.App.Srv().License() == nil || !*c.App.Srv().License().Features.Cloud {
+		c.App.AttachCWSIntermediaryLoginCookie(c.AppContext, w, r)
+	}
 
 	userTermsOfService, err := c.App.GetUserTermsOfService(user.Id)
 	if err != nil && err.StatusCode != http.StatusNotFound {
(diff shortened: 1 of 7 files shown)


api4/user.go
```diff
@@ -1852,6 +1852,11 @@ func login(c *Context, w http.ResponseWriter, r *http.Request) {
 		c.App.AttachSessionCookies(c.AppContext, w, r)
 	}
 
+	// For context see: https://mattermost.atlassian.net/browse/MM-39583
+	if c.App.Srv().License() != nil && *c.App.Srv().License().Features.Cloud {
+		c.App.AttachCWSIntermediaryLoginCookie(c.AppContext, w, r)
```
marianunez, line 1857: Would suggest we name this based on the actual cookie content (instead of our usage in CWS), maybe `CloudSessionCookie`

```suggestion
		c.App.AttachCloudSessionCookie(c.AppContext, w, r)
```

model/session.go
```diff
@@ -15,6 +15,7 @@ const (
 	SessionCookieToken            = "MMAUTHTOKEN"
 	SessionCookieUser             = "MMUSERID"
 	SessionCookieCsrf             = "MMCSRF"
+	SessionCookieCWSIntermediary  = "MMCWSINTERMEDIARYLOGINCOOKIE"
```
marianunez, line 18:
```suggestion
	SessionCookieCloudUrl  = "MMCLOUDURL"
```

marianunez (commented):
Could we add this new cookie case in the tests here: https://github.com/mattermost/mattermost-server/blob/master/api4/user_test.go#L3574 covering both cloud and non-cloud scenarios?

app/login.go
```diff
@@ -238,6 +238,40 @@ func (a *App) DoLogin(c *request.Context, w http.ResponseWriter, r *http.Request
...
+		secure = true
+	}
+
+	maxAge := *a.Config().ServiceSettings.SessionLengthWebInDays * 60 * 60 * 24
+	domain := a.GetCookieDomain()
+	subpath, _ := utils.GetSubpathFromConfig(a.Config())
+	expiresAt := time.Unix(model.GetMillis()/1000+int64(maxAge), 0)
+
+	var val string
+	if strings.Contains(domain, "localhost") {
```
nickmisasi, line 253: Should we check for cloud here, and only drop the cookie in the event it is cloud? I don't think it's a big deal not to have it, but some more security-conscious on-prem users may appreciate it

jamiehurewitz:
This is obviously not a necessary cookie so can this cookie workflow process accommodate local requirements that require notice and the ability to opt out of cookies?

nickmisasi (commented):
One non-blocking comment. Wanted to ask @AGMETEOR, do you know what happens when your cookie is set, but from a domain that's not mattermost.com? For example, if someone is hosting on-prem mattermost at mattermost.somecompany.com, and it attempts to set a cookie for mattermost.com, does that cause any errors?

app/login.go
AGMETEOR: The cloud check is already there but in the api layer

nickmisasi:
> This is obviously not a necessary cookie so can this cookie workflow process accommodate local requirements that require notice and the ability to opt out of cookies?

@jamiehurewitz I've been poking around definitions on the internet of "Strictly Necessary Cookies" and I think this specific cookie could be argued. Just want to give a little context/reasoning to see if that changes your mind.

>Strictly necessary cookies are classified as cookies that must be present for the website to provide the basic functions of the website. They are essential to be able to access features of the website and could include signing in, adding items to a cart, or e-billing.

This cookie will allow us to automatically redirect you to your workspace from customers.mattermost.com without a login to customers.mattermost.com. If this cookie doesn't exist, then a user who doesn't have a portal account (only one person per workspace has one, so 99% of cloud users won't actually have a portal account) wouldn't be able to get back to their workspace, and thus would be "unable to sign in".

Given this information, do you think this cookie is still not necessary?

jamiehurewitz (approved):

marianunez (approved):
Thanks @AGMETEOR!

stevemudie (approved):
Will test this post merge.

PR #19273, opened 2022-01-04
MM-40801: Use replica for getPostsAround

agnivade:
This query deals with posts already created,
and it was originally already querying replica.

In recent performance investigations, this came up
a lot of times. I believe the change to query master
was unintentional.

```release-note
NONE
```

### DIFF
--- a/store/sqlstore/post_store.go
+++ b/store/sqlstore/post_store.go
@@ -1404,7 +1404,7 @@ func (s *SqlPostStore) getPostsAround(before bool, options model.GetPostsOptions
 	if err != nil {
 		return nil, errors.Wrap(err, "post_tosql")
 	}
-	err = s.GetMasterX().Select(&posts, queryString, args...)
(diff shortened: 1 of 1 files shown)


agnivade:
@jespino / @lieut-data - The commit history points to this PR https://github.com/mattermost/mattermost-server/pull/11980 making the change from replica to master. However, I don't see anything in the PR message to indicate why this change was made. I believe this was unintentional and hence reverting it back to replica. But let me know if there is something I am missing.

jespino (approved):
I don't see any reason to move that back to Master in the PR code, looks like an accidental change.

streamer45 (approved):
Nice catch :+1:

lieut-data (approved):
I'm afraid I can't recall any context that might help validate this change, but I trust our perf/soak tests will quickly uncover any unexpected issues here.

agnivade:
Thanks all!

PR #19275, opened 2022-01-04
Support 'postgresql' schema designator

Szymongib:
#### Summary

Stumbled upon this when troubleshooting issue reported to Mattermost Operator https://github.com/mattermost/mattermost-operator/issues/269.

When `postgresql` schema designator is used in the connection string passed with `MM_CONFIG` env the Mattermost server treats it as a file path, not a database connection string.

According to Postgres docs, both designators are valid https://www.postgresql.org/docs/current/libpq-connect.html#LIBPQ-CONNSTRING so I think it would make sense to support both.
> The URI scheme designator can be either postgresql:// or postgres://.

#### Release Note

```release-note
Support `postgresql` schema designator
```

### DIFF
--- a/config/database.go
+++ b/config/database.go
@@ -153,7 +153,7 @@ func parseDSN(dsn string) (string, string, error) {
 		// Strip off the mysql:// for the dsn with which to connect.
 		dsn = s[1]
 
-	case "postgres":
+	case "postgres", "postgresql":
 		// No changes required
 
 	default:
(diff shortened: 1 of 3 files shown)


agnivade (approved):
Nice. Ideally, we'd just leave the parsing part to the respective libraries and see which one does not error and just accept that one. But okay to add this since this is a small change.

mgdelacroix (approved):
LGTM! :+1:

PR #19279, opened 2022-01-05
Restored `ToJSON` encoding (sans Integration) for Post APIs

levb:
#### Summary
https://github.com/mattermost/mattermost-server/pull/17999 unintentionally removed `StripActionIntegrations` from some of the JSON communications. This masked another problem, that _Edit Post_ was attempting to overwrite the entire post, rather than just the `message` part see https://github.com/mattermost/mattermost-webapp/pull/9616.

#### Ticket Link
Fixes https://mattermost.atlassian.net/browse/MM-40509

#### Release Note
```
NONE
```


model/post.go
```diff
@@ -227,6 +228,12 @@ func (o *Post) ToJSON() (string, error) {
 	return string(b), err
 }
 
+func (o *Post) EncodeJSON(w io.Writer) error {
+	copy := o.Clone()
```
agnivade, line 232: Let's remove this line. The original object is not used, and clone is costly.

model/post_list.go
```diff
@@ -80,6 +81,12 @@ func (o *PostList) ToJSON() (string, error) {
 	return string(b), err
 }
 
+func (o *PostList) EncodeJSON(w io.Writer) error {
+	copy := *o
```
agnivade, line 85: This is a shallow copy. So it doesn't really do what you want. Let's just remove this one and the one below like above.

agnivade (changes requested):
Thanks Lev. Please also add a unit test to ensure this does not regress again.

api4/post.go
```diff
@@ -451,7 +451,7 @@ func getPostsByIds(c *Context, w http.ResponseWriter, r *http.Request) {
 		}
 
 		post = c.App.PreparePostForClient(post, false, false)
-
+		post.StripActionIntegrations()
```
hanzei, line 454: Should `StripActionIntegrations` be called as part of `PreparePostForClient`? Are there cases where `PreparePostForClient` would _not_ be followed up by a `StripActionIntegrations`?

api4/post.go
```diff
@@ -733,7 +733,7 @@ func patchPost(c *Context, w http.ResponseWriter, r *http.Request) {
 	auditRec.Success()
 	auditRec.AddMeta("patch", patchedPost)
 
-	if err := json.NewEncoder(w).Encode(patchedPost); err != nil {
+	if err := patchedPost.EncodeJSON(w); err != nil {
```
hanzei, line 736: Instead of using a separate method like `EncodeJSON` it might be cleaner to implement the [`json.Marshaler`](https://pkg.go.dev/encoding/json#Marshaler) interface and keep using `Encode`.

hanzei (changes requested):

api4/post.go
agnivade: This will be the best solution IMO, if it can be done. Totally avoids any further api level logic.

agnivade (approved):
Nicely done :+1:

api4/post.go
levb: I can easily make the change, but 2/5 thorough verification/tests would be out-of-scope for this PR.

streamer45 (commented):
Maybe I misunderstood the problem here but what was the issue with using the existing `ToJSON` methods as opposed to creating almost identical new ones?

levb:
@streamer45 `Encode...` write to stream, as opposed to `[]byte`->`string`->`Write` before.

streamer45 (approved):
Looks good :+1:

api4/post.go
hanzei: I see your concerns about overwriting the `MarshalJSON`. Would it make sense to rename `EncodeJSON` to something that makes it clear that this methods should be used for communication with clients?

hanzei (approved):

DHaussermann (approved):
Tested and passed
- Discussed this change with Lev. 
- Confirmed the expected behavior and the valid test approach. 
- This change is working as expected. 
- No issues or regressions found.

LGTM!

I will document test steps in Zephyr.

PR #19285, opened 2022-01-06
MM-40594: Fix nil retention policies

agnivade:
Due to the store method having named returns, the variable
was getting set to nil instead of being an empty slice.
We missed this during code review.

https://mattermost.atlassian.net/browse/MM-40594

```release-note
NONE
```


isacikgoz (approved):
LGTM 👍

PR #19291, opened 2022-01-07
MM-40852: Fix make gen-serialized breakage

agnivade:
- Fixed the Makefile command so that it can be run in CI
- Changed to go install to make it compatible with >1.16 versions.
- Removed unnecessary lines in go.tools.mod now that we aren't using that.

```release-note
NONE
```

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -17,6 +17,12 @@ else
 	export IS_LINUX =
 endif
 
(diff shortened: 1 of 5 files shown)


isacikgoz (approved):
LGTM 👍

PR #19295, opened 2022-01-07
MM-40677: Pre-package Playbooks v1.23.0

lieut-data:
#### Release Note
```release-note
Pre-package Playbooks v1.23.0.
```
#### Ticket Link
Fixes: https://mattermost.atlassian.net/browse/MM-40677


lieut-data:
Let's hold on actually merging this until we confirm if we need to bump the minimum server version in the plugin.

lieut-data:
Confirmed the minimum server version is already bumped. We should be good to go once we have confidence in the build in question.

crspeller (approved):

Phrynobatrachus:
Screenshot from 2022-01-07 11-18-13

:+1:

Phrynobatrachus (approved):

PR #19301, opened 2022-01-07
[MM-35471] Add a field to ConfirmPaymentMethodRequest type

nickmisasi:
#### Summary
Adding the SubscriptionID to the ConfirmPaymentMethodRequest field, so that the CWS can attach this confirmed method to a specific subscription, instead of setting it as the default payment method for the whole customer. 

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-35471

#### Release Note

```release-note
None
```


AGMETEOR (approved):

marianunez (approved):
Is there any changes needed in the webapp for the area where they can update their CC?

PR #19331, opened 2022-01-14
Migrate tool install to the new pattern

agnivade:
Installing tools via go get is deprecated since Go 1.17,
and directly installing a version via go install is recommended.

We remove the go.tools.mod file as it's no longer the right
pattern.

```release-note
NONE
```


hanzei (approved):
LGTM :+1:

PR #19339, opened 2022-01-14
[MM-41022] - Fix panic when AllowCookiesSubdomain setting is false

AGMETEOR:
#### Release Note

```release-note

```

### DIFF
--- a/app/login.go
+++ b/app/login.go
@@ -260,8 +260,6 @@ func (a *App) AttachCloudSessionCookie(c *request.Context, w http.ResponseWriter
 			domain = strings.SplitN(domain, ".", 3)[2]
 		}
 
-		fmt.Println("DOM", val, domain)
-
 		cookie := &http.Cookie{
 			Name:     model.SessionCookieCloudUrl,
 			Value:    val,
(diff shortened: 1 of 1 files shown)


marianunez (approved):
LGTM, non-blocking comments below

app/login.go
```diff
@@ -245,30 +246,37 @@ func (a *App) AttachCloudSessionCookie(c *request.Context, w http.ResponseWriter
...
-		Secure:   secure,
-	}
+	if domain != "" {
+		var val string
+		if strings.Contains(domain, "localhost") {
+			val = "localhost"
+		} else {
+			val = strings.SplitN(domain, ".", 2)[0]
+			domain = strings.SplitN(domain, ".", 3)[2]
+			domain = "." + domain
```
nickmisasi: If you pass `test.com` as the domain value, this will panic. Maybe we should wrap this in a cloud check?

nickmisasi (changes requested):

app/login.go
AGMETEOR: There's a cloud check in the api layer to ensure this function only gets called for cloud. For this, all our cloud workspace URLs in the form of `example.cloud.mattermost.com`.

nickmisasi: Here's a little test for panic: https://goplay.tools/snippet/_abLj1J_VNj

AGMETEOR: Yes, it's true but our cloud URLs can never be in the form `example.com` or `test.com`

nickmisasi (approved):
Thanks @AGMETEOR

PR #19349, opened 2022-01-17
MM-40148: threadsOnly query param for user threads

koox00:
#### Summary

Currently we always calculate counts when fetching user threads.
Those counts include total, unread replies, and unread mentions,
and are potentially expensive to calculate.

This commit adds a new query param 'threadsOnly' which won't calculate
any counts and just return threads.

#### Ticket Link

https://mattermost.atlassian.net/browse/MM-40148

#### Release Note

```release-note
NONE
```


koox00:
I've attempted a refactor to separate counts from threads fetching into 2 methods.
Pushing the commit RN. 
If you like you can review both commits separately.

agnivade (approved):
Finally it's here! Would be good to have a small test in the api layer for this, but it should not block this PR.

Great work, and excited to see this implemented in mobile and webapp.

koox00:
> Would be good to have a small test in the api layer for this

tests added :)

ashishbhate (approved):
LGTM!

There is a minor gotcha in that with both `ThreadsOnly` and `TotalsOnly` set to `true` only threads are returned. I don't think its a big deal or a blocker.

agnivade:
> There is a minor gotcha in that with both ThreadsOnly and TotalsOnly set to true only threads are returned. I don't think its a big deal or a blocker.

I think we should deny that from the API and return an error.

esethna:
@koox00 will this have any impact on the UX? ie you open a channel and unread threads don't have a dot on them, or you open the threads view and unread badges are missing

koox00:
it shouldn't affect UX

koox00:
> > There is a minor gotcha in that with both ThreadsOnly and TotalsOnly set to true only threads are returned. I don't think its a big deal or a blocker.
> 
> I think we should deny that from the API and return an error.

Since we support returning both ATM, wouldn't it make sense to return both in that case?

agnivade:
> Since we support returning both ATM, wouldn't it make sense to return both in that case?

I think that case should be both set to "false". It's more about aesthetics really, so 0/5 from my side. These params are "xxx_Only" which means they should be used to _only_ return that data and nothing else. Setting 2 "xxx_only" params is a bit confusing IMO. But again, totally 0/5 and not a big deal either ways.

koox00:
> I think that case should be both set to "false". It's more about aesthetics really, so 0/5 from my side. These params are "xxx_Only" which means they should be used to _only_ return that data and nothing else. Setting 2 "xxx_only" params is a bit confusing IMO. But again, totally 0/5 and not a big deal either ways.

@agnivade, @ashishbhate https://github.com/mattermost/mattermost-server/pull/19349/commits/550804dc2e3b2cb79c1058b0e57b5b321010f999

agnivade:
You already have my approval. :slightly_smiling_face:  The rest was just non-blocking stuff. It's good to go!

ashishbhate:
Closing this PR. Its been superseded by https://github.com/mattermost/mattermost-server/pull/19833

PR #19366, opened 2022-01-18
Growth spike: Get latest MM version

BenCookie95:
#### Summary
New endpoint to fetch the latest MM version

#### Ticket Link
https://community.mattermost.com/boards/workspace/7dqusmf8stn3dens5ckhyatbdo/boattb15z9fr4m85etwrdekkduh/v37ny5c5eafnk3dypexwzagip7a/ct5wex8sfxinojb64hqysz4oe7w

#### Release Note

```release-note
Added new endpoint GET api/v4/latest_version
```


jupenur:
Looks like the GitHub API sets `Access-Control-Allow-Origin: *`. Is there any reason this whole feature couldn't (or shouldn't) be implemented on fully the client, with no server component at all?

mkraft:
> Looks like the GitHub API sets `Access-Control-Allow-Origin: *`. Is there any reason this whole feature couldn't (or shouldn't) be implemented fully on the client, with no server component at all?

The reasons we had considered were:

* Rate limiting by Github
* Performance (we'll cache the response)
* Potential access limitations configured on prem (albeit, they could also affect the Mattermost server on prem)

jupenur:
> Most instances only have 1 or 2 people hitting this page, and not really that often even.

If this is implemented client-side those people will have *separate* rate limits, whereas now they share a single rate limit applied to the server.

mkraft:
> > Most instances only have 1 or 2 people hitting this page, and not really that often even.
> 
> If this is implemented client-side those people will have _separate_ rate limits, whereas now they share a single rate limit applied to the server.

In an office environment, that's not true. In the covid-era of everyone working from home it likely would be the case.

BenCookie95:
> > Looks like the GitHub API sets `Access-Control-Allow-Origin: *`. Is there any reason this whole feature couldn't (or shouldn't) be implemented fully on the client, with no server component at all?
> 
> The reasons we had considered were:
> 
> * Rate limiting by Github (this was the main one)... however @BenCookie95 on further consideration how much traffic will this actually get on the system console?
> * Performance... however Github should be about as fast as us.
> * Potential access limitations configured... however those limitations would likely affect the server equally?
> 
> @BenCookie95 I'm wondering if upon further consideration we were overestimating the traffic this would receive. Most instances only have 1 or 2 people hitting this page, and not really that often even.

My major reason was CORs but since GitHub API sets `Access-Control-Allow-Origin: *` that should be fine. I've always generally done external api requests from server though, my only worry would be a sudden changes on the github side with regards CORs.

In terms of rate limiting, having the cache on the server is an advantage. We don't know if this endpoint will end up being used elsewhere in the app in future, it could end up being used in a place where rate limiting could become a factor.

api4/system.go
```diff
@@ -402,6 +403,27 @@ func getAnalytics(c *Context, w http.ResponseWriter, r *http.Request) {
 	}
 }
 
+func getLatestVersion(c *Context, w http.ResponseWriter, r *http.Request) {
+	if *c.App.Config().ExperimentalSettings.RestrictSystemAdmin {
+		c.Err = model.NewAppError("latestVersion", "api.restricted_system_admin", nil, "", http.StatusForbidden)
+		return
+	}
```
mkraft, line 410: What was the requirement around this `RestrictSystemAdmin` again? I need my memory jogged here.

api4/system.go
```diff
@@ -402,6 +403,27 @@ func getAnalytics(c *Context, w http.ResponseWriter, r *http.Request) {
...
+
+	resp, err := c.App.GetLatestVersion("https://api.github.com/repos/mattermost/mattermost-server/releases/latest")
```
mkraft, line 412: Since this is Github-specific anyway, can we limit the parameter to just the org/repo path like this?
```suggestion
	resp, err := c.App.GetLatestVersion("mattermost/mattermost-server")
```

app/admin.go
```diff
@@ -14,11 +15,16 @@ import (
...
 	"github.com/mattermost/mattermost-server/v6/config"
 	"github.com/mattermost/mattermost-server/v6/model"
+	"github.com/mattermost/mattermost-server/v6/services/cache"
 	"github.com/mattermost/mattermost-server/v6/shared/i18n"
 	"github.com/mattermost/mattermost-server/v6/shared/mail"
 	"github.com/mattermost/mattermost-server/v6/shared/mlog"
 )
 
+var latestVersionCache = cache.NewLRU(cache.LRUOptions{
+	Size: 1000,
```
mkraft, line 25: Is this 1000 bytes or 1000 entries?

i18n/en.json
```diff
@@ -8375,6 +8391,10 @@
     "id": "model.file_info.is_valid.user_id.app_error",
     "translation": "Invalid value for user_id."
   },
+  {
+    "id": "model.github_release_info.is_valid.id.app_error",
+    "translation": "Invalid value for github ID."
+  },
```
mkraft, line 8397: If any of these are only used in errors you can trick the CI by setting the translation string to `" "` and then that'll be a few less entries that people need to verify.

pablovelezvidal (approved):
besides Martin questions, it LGTM. :shipit:

jupenur (approved):

api4/system.go
BenCookie95: https://docs.mattermost.com/configure/configuration-settings.html?highlight=restrictsystemadmin#restrict-system-admin 
```
Restricts the System Admin from viewing and modifying a subset of server configuration settings from the System Console. Not recommended for use in on-prem installations. This is intended to support Mattermost Private Cloud in giving the System Admin role to users but restricting certain actions only for Cloud Admins.
```

api4/system.go
BenCookie95: I wanted to pass in the whole URL so I can mock the GET request in my tests, I'll take a look and see if there's a way for me to incorporate this

app/admin.go
BenCookie95: It's entries, reduced it to 1

api4/system.go
mkraft: I'm 0/5 on this. I was thinking it might make the method easier to invoke, but if as a result it's less testable then it's probably not worth it.

mkraft (approved):

furqanmlk (approved):
Working as expected. Test case have been updated.

image

PR #19370, opened 2022-01-19
MM-40880: Sentry crash: MarkChannelAsUnread

agnivade:
Use the correct error variable to prevent crash

The errors.As check has been removed because ThreadStore.Get
returns a nil error if no thread is found and that's how
the application logic is written. So the check was redundant.

```release-note
NONE
```


app/channel_test.go
```diff
@@ -1992,6 +1993,78 @@ func TestMarkChannelsAsViewedPanic(t *testing.T) {
...
+	require.NoError(t, err)
+
+	mockPreferenceStore := mocks.PreferenceStore{}
+	mockPreferenceStore.On("Get", "userID", model.PreferenceCategoryDisplaySettings, model.PreferenceNameCollapsedThreadsEnabled).Return(&model.Preference{Value: "on"}, nil)
+
+	mockThreadStore := mocks.ThreadStore{}
+	mockThreadStore.On("MarkAllAsReadInChannels", "userID", []string{"channelID"}).Return(nil)
+	mockThreadStore.On("GetMembershipForUser", "userID", "rootID").Return(nil, nil)
+	mockThreadStore.On("MaintainMembership", "userID", "rootID", mock.AnythingOfType("store.ThreadMembershipOpts")).Return(&model.ThreadMembership{}, nil)
+	mockThreadStore.On("Get", "rootID").Return(nil, errors.New("bad error"))
```
streamer45, line 2031: Should we highlight this somehow? Most of the test seems to be boilerplate so it gets hard to follow what is that we are really testing for.

streamer45 (approved):
Looks good :+1:

app/channel_test.go
agnivade: Good point :+1:

agnivade:
@amyblais - This is a good candidate for 6.3 ESR.

isacikgoz (approved):
LGTM 👍

amyblais:
@agnivade Does this need QA review?

agnivade:
@amyblais - No need for that. There's already a unit test.

PR #19374, opened 2022-01-19
[MM-40917] - Inactive Server Email Notification

AGMETEOR:
#### Summary
How inactivity (in terms of use) is determined?  Run a job every 24 hours to determine following:
- If a no posts have been made for an arbitrary time (now 100 hours) since last post made, server is inactive
- If no posts have been made, check for the last time we had a session and if more than an arbitrary time (now 100 hours), server is inactive
- After job has run atleast once, we store a system value with a time stamp of when it last sent out an inactivity email, if that time and the last time a post or session was made exceeds an arbitrary time (now 100 hours), server is inactive

This means that a user will occasionally get inactivity emails sent to them if them take an arbitrary time (now 100 hours) , without interacting with their workspace

Screenshot 2022-01-20 at 13 31 29

#### Ticket Link

[MM-40917](https://mattermost.atlassian.net/browse/MM-40917)
#### Release Note

```release-note
NONE

```


app/server_inactivity.go
```diff
@@ -0,0 +1,134 @@
...
+	inactivityDurationHours, parseError := strconv.ParseFloat(inactivityDurationHourseEnv, 64)
+	if parseError != nil {
+		// default to 100 hours
+		inactivityDurationHours = serverInactivityHours
+	}
+
+	systemValue, sysValErr := s.Store.System().GetByName("INACTIVITY")
+	if sysValErr != nil {
+		if _, ok := sysValErr.(*store.ErrNotFound); !ok {
+			mlog.Error("An error occurred while getting INACTIVITY from system store", mlog.Err(sysValErr))
```
marianunez, line 30: Should we downgrade this to a warn or info given that is a non-essential check ?

app/server_inactivity.go
```diff
@@ -0,0 +1,134 @@
...
+			lastSessionAt := session.CreateAt
+			sesT := time.Unix(lastSessionAt/1000, 0)
+			timeForLastSession := time.Since(sesT).Hours()
+
+			if timeLastSentInativityEmail > inactivityDurationHours && timeForLastSession > inactivityDurationHours {
+				s.takeInactivityAction()
+			}
+			return
+		}
+	}
```
marianunez, line 67: Wondering what is the value of checking both? If there is post activity for sure there had been session inactivity, and if there wasn't post inactivity it never checks for session inactivity. Would checking both be redundant here?

marianunez (changes requested):

app/server_inactivity.go
AGMETEOR: From both the post and session, we're interested in when the *time* they last happened. Also, since MM is a chat app, checking whether people are sending posts or not should be the first indicator of inactivity/activity. But it's possible that no posts were ever made and therefore we wouldn't have a *time* reference to determine inactivity. In such a case, we use session as a fallback.

templates/inactivity_body.mjml
```diff
@@ -0,0 +1,62 @@
...
+            {{.Props.Title}}
+          </mj-text>
+          <mj-text align="left" color="#3F4350" font-size="16px" font-weight="normal" line-height="24px" padding-bottom="16px">
+            {{.Props.SubTitle}}
+          </mj-text>
+          <mj-text align="left" color="#3F4350" font-size="16px" font-weight="normal" line-height="24px" padding-bottom="16px">
+            <ul>
+              <li>{{.Props.InfoBullet}}<b>Channels</b></li>
+              <li>{{.Props.InfoBullet1}}<b>Playbooks</b></li>
+              <li>{{.Props.InfoBullet2}}<b>Boards</b></li>
```
nickmisasi, line 21: These might not translate well since the subject (Channels, Playbooks, Boards) could move around in different languages.

store/sqlstore/post_store.go
```diff
@@ -2145,6 +2145,17 @@ func (s *SqlPostStore) AnalyticsPostCount(teamId string, mustHaveFile bool, must
 	return v, nil
 }
 
+func (s *SqlPostStore) GetLastPostRow() (lastPost *model.Post, _ error) {
+	query := `SELECT * FROM Posts ORDER BY CREATEAT DESC LIMIT 1`
```
nickmisasi, line 2149: I know we discussed ORDER BY DESC on an indexed field as being "ok" - but think it would be good to test this functionality in a db with a few million posts, just to be safe. Wouldn't want this to lock up the DB or anything.

store/sqlstore/session_store.go
```diff
@@ -229,6 +229,16 @@ func (me SqlSessionStore) UpdateExpiresAt(sessionId string, time int64) error {
 	return nil
 }
 
+func (me SqlSessionStore) GetLastSessionRow() (lastSession *model.Session, _ error) {
+	query := `SELECT * FROM Sessions ORDER BY CREATEAT DESC LIMIT 1`
```
nickmisasi, line 233: Also want to verify, is this table's CREATEAT indexed?

nickmisasi (commented):
Code LGTM with a few nits.

One thing I want to call out around the content of the email: 
`Guest access to specified Channels` 
Might be a weird one to have in the list since it's not available in the Starter plan (I don't see any logic around tailoring these list items to the plan they're on) - only Enterprise and Professional have guest accounts.

marianunez (changes requested):

app/server_inactivity.go
```diff
@@ -0,0 +1,134 @@
...
+
+	systemValue, sysValErr := s.Store.System().GetByName("INACTIVITY")
+	if sysValErr != nil {
+		if _, ok := sysValErr.(*store.ErrNotFound); !ok {
+			mlog.Error("An error occurred while getting INACTIVITY from system store", mlog.Err(sysValErr))
+		}
+	}
+
+	post, _ := s.Store.Post().GetLastPostRow()
+	session, _ := s.Store.Session().GetLastSessionRow()
```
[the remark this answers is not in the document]
AGMETEOR: Agreed. Thanks for pointing this out

app/server_inactivity.go
```diff
@@ -0,0 +1,134 @@
...
+
+func (s *Server) takeInactivityAction() {
+	siteURL := *s.Config().ServiceSettings.SiteURL
+	properties := map[string]interface{}{
+		"SiteURL": siteURL,
+	}
+	s.GetTelemetryService().SendTelemetry("inactive_server", properties)
+	users, err := s.Store.User().GetSystemAdminProfiles()
+	if err != nil {
+		mlog.Error("Failed to get system admins for license expired message from Mattermost.")
```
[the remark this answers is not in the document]
AGMETEOR: oops

templates/inactivity_body.mjml
AGMETEOR: Since these are product names, I am wondering if they should be translated. @cwarnermm thoughts?

app/server.go
```diff
@@ -1455,6 +1456,12 @@ func runJobsCleanupJob(s *Server) {
 	}, time.Hour*24)
 }
 
+func (s *Server) runInactivityCheckJob() {
+	model.CreateRecurringTask("Server inactivity Check", func() {
+		s.doInactivityCheck()
+	}, time.Hour*24)
```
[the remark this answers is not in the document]
AGMETEOR: Added

templates/inactivity_body.mjml
cwarnermm: @AGMETEOR - Great catch and thanks for asking! Official product names, including Channels, Playbooks, and Boards aren't translated into other languages.

app/server.go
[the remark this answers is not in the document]
marianunez: Should we move the feature flag inside this method given that this is the one run that is running daily? the top level would be run only on server startup and if we change the feature flag we would need a server restart to have it not schedule the job?

store/sqlstore/post_store.go
AGMETEOR: It is safe. Also, changed the queries to run even faster

app/server_inactivity.go
```diff
@@ -0,0 +1,136 @@
...
+	if parseError != nil {
+		// default to 100 hours
+		inactivityDurationHours = serverInactivityHours
+	}
+
+	systemValue, sysValErr := s.Store.System().GetByName("INACTIVITY")
+	if sysValErr != nil {
+		if _, ok := sysValErr.(*store.ErrNotFound); !ok {
+			mlog.Warn("An error occurred while getting INACTIVITY from system store", mlog.Err(sysValErr))
+		}
```
marianunez, line 36: Is this the case if there is no store value for `INACTIVITY`?

app/server_inactivity.go
```diff
@@ -0,0 +1,136 @@
...
+		lastSessionAt, _ := s.Store.Session().GetLastSessionRowCreateAt()
+		if lastSessionAt != 0 {
+			sesT := time.Unix(lastSessionAt/1000, 0)
+			timeForLastSession := time.Since(sesT).Hours()
+
+			if timeLastSentInativityEmail > inactivityDurationHours && timeForLastSession > inactivityDurationHours {
+				s.takeInactivityAction()
+			}
+			return
+		}
```
marianunez, line 68: This seems to be the same functionality as below, any way we can extract this from this `if` and consolidate below?

app/server_inactivity.go
```diff
@@ -0,0 +1,136 @@
...
+		return
+	}
+}
+
+func (s *Server) takeInactivityAction() {
+	siteURL := *s.Config().ServiceSettings.SiteURL
+	properties := map[string]interface{}{
+		"SiteURL": siteURL,
+	}
```
marianunez, line 100: Do we need to default this to something if no SiteURL is set?

app/server_inactivity.go
```diff
@@ -0,0 +1,136 @@
...
+	s.GetTelemetryService().SendTelemetry("inactive_server", properties)
+	users, err := s.Store.User().GetSystemAdminProfiles()
+	if err != nil {
+		mlog.Error("Failed to get system admins for inactivity check from Mattermost.")
+		return
+	}
+
+	for _, user := range users {
+		user := user
```
marianunez, line 109: Is this assignment necessary?

templates/inactivity_body.mjml
```diff
@@ -0,0 +1,62 @@
...
+          <mj-text css-class="emailFooter" font-family="Arial" font-size="12px" line-height="16px" color="#3F4350">
+            {{.Props.Organization}}
+            {{.Props.FooterV2}}
+          </mj-text>
+        </mj-column>
+      </mj-section>
+
+    </mj-wrapper>
+  </mj-body>
+</mjml>
```
marianunez, line 62: nit: Missing new line at EOF

marianunez:
> One thing I want to call out around the content of the email:
> `Guest access to specified Channels`
> Might be a weird one to have in the list since it's not available in the Starter plan (I don't see any logic around tailoring these list items to the plan they're on) - only Enterprise and Professional have guest accounts.

@dontoisme see comment above regarding the email content

app/server_inactivity.go
```diff
@@ -0,0 +1,134 @@
...
+		}
+	}
+
+	post, _ := s.Store.Post().GetLastPostRow()
+	session, _ := s.Store.Session().GetLastSessionRow()
+
+	// If we have a system value, it means this job already ran atleast once.
+	// we then check the last time the job ran plus the last time a post was made to determine if we
+	// can remind the user to use workspace again. If no post was made, we check the last time they logged in (session)
+	// and determine whether to send them a reminder.
```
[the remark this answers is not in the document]
dontoisme: @marianunez yes I would say that is the intended behavior.

dontoisme:
@marianunez @nickmisasi the original idea for that copy was to bring people back to each of the products. In hindsight referencing a professional feature might not be the best thing. It could be confused with the other email that was created. I'm checking back on our notes.

> Code LGTM with a few nits.
> 
> One thing I want to call out around the content of the email: `Guest access to specified Channels` Might be a weird one to have in the list since it's not available in the Starter plan (I don't see any logic around tailoring these list items to the plan they're on) - only Enterprise and Professional have guest accounts.

templates/inactivity_body.mjml
nickmisasi: @cwarnermm to clarify, while the product names don't get translated, they should be translated as part of the entire string, right? 

For example:
"Guest access to specified **Channels**"
Could be something like:
"accès invité aux **Channels** publiques" in french - where "Channels" is no longer at the end of the sentence for it to make sense?

app/server_inactivity.go
AGMETEOR: It's not exactly the same, below, there's no `timeLastSentInativityEmail `

app/server_inactivity.go
AGMETEOR: Since SiteURL is critical to the telemetry and sending of the email, we can consider this an error
```go
if siteURL == "" {
		mlog.Error("SiteURL needs to be set to run inactivity job")
		return
	}
```

app/server_inactivity.go
AGMETEOR: Took it out

app/server_inactivity.go
AGMETEOR: I am going to consider logging for any error encountered regardless of whether it is a not found. I think it's safe since we have a nil check down for this `systemValue`

```go
systemValue, sysValErr := s.Store.System().GetByName("INACTIVITY")
	if sysValErr != nil {
		mlog.Warn("An error occurred while getting INACTIVITY from system store", mlog.Err(sysValErr))
	}
```

AGMETEOR:
Hello @dontoisme . Any final comments?

app/server_inactivity.go
marianunez: We can maybe unify this duplicate code by extracting the one different condition into a var (i.e `shouldSendEmail`) which defaults to `true` and can be overriden if it was sent before and meets the condition. What do you think?

```suggestion

	shouldSendEmail := true;

	// Check if the email was sent before and we shouldn't send again 
	if systemValue != nil {
		sysT, _ := strconv.ParseInt(systemValue.Value, 10, 64)
		tt := time.Unix(sysT/1000, 0)
		timeLastSentInativityEmail := time.Since(tt).Hours()

		shouldSendEmail :=  timeLastSentInativityEmail > inactivityDurationHours
	}
	
	// We check if the user has not made any posts and remind them to use the workspace.
	// If no posts have been made. We check the last time
	// they logged in (session) and send a reminder.

	lastPostAt, _ := s.Store.Post().GetLastPostRowCreateAt()
	if lastPostAt != 0 {
		posT := time.Unix(lastPostAt/1000, 0)
		timeForLastPost := time.Since(posT).Hours()
		if shouldSendEmail && timeForLastPost > inactivityDurationHours {
			s.takeInactivityAction()
		}
		return
	}

	lastSessionAt, _ := s.Store.Session().GetLastSessionRowCreateAt()
	if lastSessionAt != 0 {
		sesT := time.Unix(lastSessionAt/1000, 0)
		timeForLastSession := time.Since(sesT).Hours()
		if  shouldSendEmail && timeForLastSession > inactivityDurationHours {
			s.takeInactivityAction()
		}
		return
	}
}
```

marianunez (approved):
LGTM aside the possible refactor suggestion to reduce code duplication. Thanks @agmeteor!

app/server_inactivity.go
marianunez: Thanks for clarifying!

app/server_inactivity.go
```diff
@@ -0,0 +1,142 @@
...
+		return
+	}
+}
+
+func (s *Server) takeInactivityAction() {
+	siteURL := *s.Config().ServiceSettings.SiteURL
+	if siteURL == "" {
+		mlog.Error("SiteURL needs to be set to run inactivity job")
+		return
+	}
```
marianunez, line 103: Given that the telemetry does have a server id that identifies the server regardless of SiteURL, I think we don't need to error out if we don't have this information. We should be able to continue without this value ? It is possible that early stage servers may not have this configured and they may have inactivity happen - we can probably benefit of knowing this even though we don't have the SiteURl. Thoughts @dontoisme ?

AGMETEOR:
Hello @nickmisasi gentle ping here

AGMETEOR:
Hello @nickmisasi gentle ping again

templates/inactivity_body.mjml
nickmisasi: @AGMETEOR this hasn't been addressed. 

You either need to: 
Include the subject (ie, Channels, Playbooks, Boards, etc directly into the translation ["Guest access to specified Channels"](https://github.com/mattermost/mattermost-server/pull/19374/files#diff-1a490f08504158838befaa816ee8d5c08dff9e694d5a568ce4b0a4398dcc563eR3572) You can [bold like this](https://github.com/mattermost/mattermost-server/blob/daa19220cce18409f3f2aff6086d6d4e45ba4fd5/i18n/en.json#L2560)

or

Pass the subject in as a variable ["Guest access to specified {item}"](https://github.com/mattermost/mattermost-server/pull/19374/files#diff-1a490f08504158838befaa816ee8d5c08dff9e694d5a568ce4b0a4398dcc563eR3572) and then passing in "Channel" like you do with `Name` [here](https://github.com/mattermost/mattermost-server/pull/19374/files#diff-9eea96507e6338e4e5e7e0bb9a00794fcf4062ef1b710c68fc300377823f4233R771)

nickmisasi (changes requested):

AGMETEOR:
@nickmisasi in the email templates it's not possible to bold things like so "example **i am bold**". I'll give them their own translation strings.

AGMETEOR:
> @AGMETEOR did you try this (`**wrap in **`)?
> 
> https://github.com/mattermost/mattermost-server/blob/daa19220cce18409f3f2aff6086d6d4e45ba4fd5/i18n/en.json#L2560

I did. If the that translation is going into MM it'll bold but on the email templates it doesn't

nickmisasi (approved):
Other than the one issue with translations all LGTM here. Sorry for being so difficult @AGMETEOR, if that last attempt at bolding the translated string doesn't work, I can live with it as is

AGMETEOR:
> Other than the one issue with translations all LGTM here. Sorry for being so difficult @AGMETEOR, if that last attempt at bolding the translated string doesn't work, I can live with it as is

No worries. Thanks for the thorough review

app/server_inactivity.go
marianunez: Thoughts on the above? @AGMETEOR

AGMETEOR:
@stevemudie gentle nudge on this. QA steps are in the ticket. Let me know in case you would like us to QA test together

stevemudie (approved):
This is all set.

I received inactivity emails as expected after a time.

Nice work @AGMETEOR!

app/server_inactivity.go
```diff
@@ -0,0 +1,141 @@
...
+		timeLastSentInativityEmail := time.Since(tt).Hours()
+
+		lastPostAt, _ := s.Store.Post().GetLastPostRowCreateAt()
+		if lastPostAt != 0 {
+			posT := time.Unix(lastPostAt/1000, 0)
+			timeForLastPost := time.Since(posT).Hours()
+
+			if timeLastSentInativityEmail > inactivityDurationHours && timeForLastPost > inactivityDurationHours {
+				s.takeInactivityAction()
+			}
```
marianunez, line 57: @AGMETEOR can you confirm is the inactivity emails is sent again after the first pass of 100 hours of inactivity? cc/ @dontoisme

PR #19384, opened 2022-01-21
PostStore: Remove `db:"-"` tag and some other hacks

agnivade:
Before gorp, to use sqlx we had to resort to using
internal structs without the `db:"-"` tag so that
it doesn't interfere with the gorp table generation.

Now that it's gone, we can use the original model structs.

```release-note
NONE
```


isacikgoz (approved):
Code simplification, I'm sold!

PR #19386, opened 2022-01-21
Initial replace of circleci to use github actions

cpanato:
#### Summary
Initial work to replace CircleCI to use GitHub Actions.
In this PR we are moving the Checks jobs and golangci-lint

#### Ticket Link
n/a

#### Release Note

```release-note
NONE
```

### DIFF
--- a/.github/workflows/checks.yml
+++ b/.github/workflows/checks.yml
@@ -1,0 +1,160 @@
+name: Checks
+
+on:
+  push:
+    branches:
+      - master
+      - cloud
+      - "release-*"
+  pull_request:
+
+permissions:
+  contents: read
+
+jobs:
+  check-app-layers:
+    runs-on: ubuntu-latest
+
+    steps:
+    - name: Checkout code
+      uses: actions/checkout@v2
+
+    - name: Extract version of Go to use
+      run: echo "GOVERSION=$(cat go.mod | grep go | head -1 | awk ' { print $2 } ' | cut -d '@' -f 1 | cut -d ':' -f 2)" >> $GITHUB_ENV
+
+    - name: Install Go
+      uses: actions/setup-go@v2
+      with:
+        go-version: ${{ env.GOVERSION }}
+
+    - name: Run check-app-layers
+      run: |
+        make app-layers
+        if [[ -n $(git status --porcelain) ]]; then echo "Please update the app layers using 'make app-layers'"; exit 1; fi
+
+  check-store-layers:
+    runs-on: ubuntu-latest
+
+    steps:
+    - name: Checkout code
+      uses: actions/checkout@v2
+
+    - name: Extract version of Go to use
+      run: echo "GOVERSION=$(cat go.mod | grep go | head -1 | awk ' { print $2 } ' | cut -d '@' -f 1 | cut -d ':' -f 2)" >> $GITHUB_ENV
+
+    - name: Install Go
+      uses: actions/setup-go@v2
+      with:
+        go-version: ${{ env.GOVERSION }}
+
+    - name: Run check-store-layers
+      run: |
+        make store-layers
+        if [[ -n $(git status --porcelain) ]]; then echo "Please update the store layers using 'make store-layers'"; exit 1; fi
+
+  check-mocks:
+    runs-on: ubuntu-latest
+
+    steps:
+    - name: Checkout code
+      uses: actions/checkout@v2
+
+    - name: Extract version of Go to use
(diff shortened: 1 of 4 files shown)


cpanato:
@agnivade looks like there are some lints to fix and the `check-migrations` is failing, is it okay for you that I update those?

cpanato:
tagging @isacikgoz @lieut-data @streamer45 for visibility, tag others if needed

cpanato:
@amyblais this should be cherry-picked to the active release branches

amyblais:
> @amyblais this should be cherry-picked to the active release branches when approved and merged

The current active branches are: ``release-6.3``, ``release-6.2``, ``release-6.1``, ``release-5.37``, and ``cloud``.

agnivade:
@cpanato - It seems like there is no way to set up dependent steps like we did in CircleCI is there? This seems to be just running all jobs in parallel.

.github/workflows/checks.yml
```diff
@@ -0,0 +1,160 @@
...
+      with:
+        go-version: ${{ env.GOVERSION }}
+
+    - name: Run check-mocks
+      run: |
+        make einterfaces-mocks store-mocks
+        if [[ -n $(git status --porcelain) ]]; then echo "Please update the mocks using 'make einterfaces-mocks store-mocks'"; exit 1; fi
+
+  check-migrations:
+    runs-on: ubuntu-latest
```
agnivade, line 76: This does not seem to run on the mattermost docker image we have.

.github/workflows/checks.yml
```diff
@@ -0,0 +1,160 @@
...
+      with:
+        go-version: ${{ env.GOVERSION }}
+
+    - name: Run check-gen-serialized
+      run: |
+        make gen-serialized
+        if [[ -n $(git status --porcelain) ]]; then echo "Please update the serialized files using 'make gen-serialized'"; exit 1; fi
+
+  check-mattermost-vet:
+    runs-on: ubuntu-latest
```
agnivade, line 140: same here as well.

agnivade (commented):
@cpanato - I'd like to understand this change very well since this is the main repo. Please give me some time to go through the changes thoroughly.

cc @jupenur as well.

.github/workflows/checks.yml
cpanato: for those we just need go to be available, there is not extra dependency

cpanato:
> @cpanato - I'd like to understand this change very well since this is the main repo. Please give me some time to go through the changes thoroughly.
> 
> cc @jupenur as well.

sure take your time, this is the initial migration from circleci to use github actions, that I mention to you in our call and also gave a psa in the dev meeting this week

cpanato:
> @cpanato - It seems like there is no way to set up dependent steps like we did in CircleCI is there? This seems to be just running all jobs in parallel.

those check were running in parallel before (in circleci)

.github/workflows/checks.yml
cpanato: in circleci there is no easy way to install go or have that available in the env, that is why we run over a container.
we can do that here if you want as well

agnivade:
> sure take your time, this is the initial migration from circleci to use github actions, that I mention to you in our call and also gave a psa in the dev meeting this week

Yep, I would like to understand this well before merging anything.

> those check were running in parallel before (in circleci)

Yes, I understand "some" checks were in parallel. But as you can see, they were dependent on `setup`, and the following stages are dependent on those. Is that possible here?

cpanato:
> > sure take your time, this is the initial migration from circleci to use github actions, that I mention to you in our call and also gave a psa in the dev meeting this week
> 
> Yep, I would like to understand this well before merging anything.
> 
> > those check were running in parallel before (in circleci)
> 
> Yes, I understand "some" checks were in parallel. But as you can see, they were dependent on `setup`, and the following stages are dependent on those. Is that possible here?

the setup is needed for the testing / build part, not for the checks because the setup fetches the Webapp repository, only if the those checks needs the Webapp I understood wrong the checks and will refactor, otherwise it is not needed

agnivade:
This is needed for early exit. If setup fails, no other steps will run. Otherwise, every time all stages will run and increase overall CI execution time even if we know all other steps will fail.

cpanato:
> This is needed for early exit. If setup fails, no other steps will run. Otherwise, every time all stages will run and increase overall CI execution time even if we know all other steps will fail.

setup just run the checkout there is not other checks there

cpanato:
thanks for your review and feedback @agnivade based on our conversation over zoom I've made some changes

also, when we migrate the tests/build they will depends on the checks

cpanato:
closing per @spirosoik request

PR #19387, opened 2022-01-21
[MM-41200] Add OnSendDailyTelemetry() plugin hook

hanzei:
#### Summary
This PR implements a `OnSendDailyTelemetry()` hook that allows plugin to send there telemetry when ever the server does.

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-41200

#### Release Note
```release-note
Added OnSendDailyTelemetry() plugin hook
```


hanzei:
CI will fail until https://github.com/mattermost/mattermost-server/pull/19400 is merged.

services/telemetry/telemetry.go
```diff
@@ -902,6 +903,13 @@ func (ts *TelemetryService) trackPlugins() {
 		"plugins_with_settings":         settingsCount,
 		"plugins_with_broken_manifests": brokenManifestCount,
 	})
+
+	ts.srv.Go(func() {
+		pluginsEnvironment.RunMultiPluginHook(func(hooks plugin.Hooks) bool {
```
larkox, line 908: As of today, there seems not to be any use for the return value, neither here nor in the hook. Can we make both functions that return no value?

larkox (approved):
LGTM

services/telemetry/telemetry.go
hanzei: I'm not sure which function are you referring. Do you mean the return value of the functional argument of `RunMultiPluginHook`? If so, in e.g. https://github.com/mattermost/mattermost-server/blob/ee92bda912b78318e120e7efa46e72b71b991877/app/login.go#L165 it's used.

larkox: Oh, I see. Then dismiss my comment.

services/gomanager/gomanger.go
```diff
@@ -0,0 +1,40 @@
...
+
+// GoManager ensure the complete execution of go rountes managed by it.
+type GoManager struct {
+	goroutineCount      int32
+	goroutineExitSignal chan struct{}
+}
+
+func New() *GoManager {
+	return &GoManager{
+		goroutineExitSignal: make(chan struct{}, 1),
```
hanzei, line 15: I would have loved to get around a `New` method, but I don't think there is a good way to create a buffer channel by default. Maybe using `sync.Once` is an idea. 0/5 on the performance impact using that.

hanzei: I've went with this idea to make the package easier to use.  @agnivade Please let me know what you think,

agnivade:
Sorry, this is a bit too complicated workaround for a small thing. I was okay with the Go method on the *App, but this seems to be unnecessarily complicated. Might I suggest to just copy-paste the logic to the telemetry package? 

> A little copying is better than a little dependency.

http://go-proverbs.github.io/

Just add a note that this logic is copied from *Server.Go to avoid adding an adapter method to App, and it should be good to go.

hanzei:
> Might I suggest to just copy-paste the logic to the telemetry package?

While that is a possible way, I'm wondering if the `gomanager` package might be useful for other services. At least for plugin it's definitely useful. Even products could use this package. Copying the code to each plugin that wants to use that functionality seems a bit to much to me.

Could you elaborate what you mean with " unnecessarily complicated"? One thing I like about the package is that a `GoManager` has a useful zero value, while the current server code doesn't have one. That should reduce complexity.

agnivade:
>  At least for plugin it's definitely useful. Even products could use this package. Copying the code to each plugin that wants to use that functionality seems a bit to much to me.

Sure, but there's no need for that yet. We can do that if the need arises.

> Could you elaborate what you mean with " unnecessarily complicated"? 

It's an unnecessarily complicated solution to the problem we have at hand (IMO of course :)). We just want to access to Go method of the server. Your original solution of adding the method to the App package was fine to me.

Would you be fine with the original solution, and then if we see that other plugins need this, then move this to a separate package if needed?

hanzei:
@agnivade I have no objections to reverting my changes, but I'm wondering what the downside of them are. It's an equal number of code then the current implementation (Maybe even less) and the code is there now. If we plan to have that kind of package anyway in the future, why not do it now?

agnivade:
IMO, we shouldn't be aggresively refactoring things unless there's a need. As Corey mentioned in his original multi-product architecture vision, we should only de-duplicate when there are 3 or more places where the code is reused.

Personally, I am not a big fan of this pattern of randomly spinning up goroutines and having a global `Stop` method which waits for all the goroutines to finish. There are several problems with this:
- it encourages users to create goroutines without much thought on backpressure or resource constraints.

This has led to real problems. See my recent PR for example https://github.com/mattermost/mattermost-server/pull/19498 on the dangers of randomly spinning up goroutines.

A much better way is to cut down concurrency into much granular parts - like for `RunMultiPluginHook` - we can make the method synchronous, but run all the plugin hooks in separate goroutines and wait for all of them to finish. That's a much better model and limits the concurrency to a single part of the system rather than spawning goroutines globally.

Concurrency is a difficult problem, and we don't realize it unless it's too late. Some of it is also touched in our [Performance Review](https://docs.google.com/document/d/1Uzt3XHyKhDKipkuCmESkHoPio7vz7VYS4N_5_9ffgNU/edit) doc.

There's definitely an argument there that - so then should we change all of the code? I'd say that would be too much work :stuck_out_tongue: but I don't want to further encourage this pattern. Given that we have always been running `RunMultiPluginHook` in a goroutine, it's OK. But refactoring the code into a package has implications that this is a pattern that we want others to use.

Hopefully I was able to clarify my reasoning Ben. I fully realize that this discussion has gone for far too long, that's why I don't want to hold this up for any longer. Let me know your thoughts, and thanks for your patience with the review process! :hugs:

agnivade:
Looks like you might have accidentally removed the original code @hanzei, while fixing the merge conflicts. There's nothing in services/telemetry/telemetry.go anymore.

hanzei:
@agnivade Thank you very much for the insights :+1: I agree now that the pattern a `Gomanager` suggests it not one, that we should encourage in the future.

I messed up the merge from `master`, thanks for pointing that out. All the code is in place now. `RunMultiPluginHook` is run synchronous without it's own go routine in the telemetry service. The whole `sendDailyTelemetry` method runs in an own go routine, so only sending out the other telemetry data is blocked by the plugin hook. If we really want to speedup sending telemetry data, we can move all the calls in https://github.com/mattermost/mattermost-server/blob/c503b1230193e90067627ebc8db71af742936301/services/telemetry/telemetry.go#L155-L164 into there own go routines.

hanzei:
> A much better way is to cut down concurrency into much granular parts - like for `RunMultiPluginHook` - we can make the method synchronous, but run all the plugin hooks in separate goroutines and wait for all of them to finish. That's a much better model and limits the concurrency to a single part of the system rather than spawning goroutines globally.

OT to this PR: That seems like a great idea to me. All of the `Has` hooks invocations, e.g. `UserHasLoggedIn`, could be run async from each. I'm not so sure about the `Will` one's. E.g. `MessageWillBePosted` is chained between the plugin. That is a hook that can't be run async.

agnivade (approved):
Looks great Ben! And thank you for the discussion as well! 

> OT to this PR: That seems like a great idea to me. All of the Has hooks invocations, e.g. UserHasLoggedIn, could be run async from each. I'm not so sure about the Will one's. E.g. MessageWillBePosted is chained between the plugin. That is a hook that can't be run async.

Nice, yeah those that shouldn't be run async are already solved. But for the other ones, we could granularize them. :+1:

hanzei:
@DHaussermann Let's talk offline about how we can test this PR.

DHaussermann (approved):
@hanzei This would be far more efficient to test post merge.
For now there is not much we can do for functional testing on this.

We should test this post-merge.

PR #19393, opened 2022-01-21
[MM-41206] Fix: Cloud workspaces unable to use OIDC Authentication

nickmisasi:
#### Summary
A similar change to this one was introduced in https://github.com/mattermost/mattermost-server/pull/15747/files
That one was only introduced to the cloud branch. Last week, the cloud branch was deleted and recreated, and this change was lost in the process.

Without this change, the redirect URI is built based off of the `X-Forwarded-For` header, which in the cloud context is _always_ `http` - but the redirect URI the installation actually needs requires https, so there was a mismatch. Now, if in the context of a cloud installation, it will fall back on the siteURL instead.

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-41206

#### Release Note

```release-note

```

### DIFF
--- a/web/handlers.go
+++ b/web/handlers.go
@@ -207,6 +207,9 @@ func (h Handler) ServeHTTP(w http.ResponseWriter, r *http.Request) {
 
 	subpath, _ := utils.GetSubpathFromConfig(c.App.Config())
 	siteURLHeader := app.GetProtocol(r) + "://" + r.Host + subpath
+	if c.App.Srv().License() != nil && *c.App.Srv().License().Features.Cloud {
+		siteURLHeader = *c.App.Config().ServiceSettings.SiteURL + subpath
+	}
 	c.SetSiteURLHeader(siteURLHeader)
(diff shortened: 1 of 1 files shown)


amyblais:
@nickmisasi I wonder how to avoid this in the future. The Cloud Fast-Forward Process is now so that Ibrahim re-creates the cloud branch https://docs.google.com/document/d/1p54GySjFirwrTBbOc1jqXwfKE_XPJV0r6aNbVyZ1zV4/edit#heading=h.ey5x8s26ugvb.

nickmisasi:
@amyblais this one will be going into master, so this specific case shouldn't happen in the future. We're now checking for cloud and overriding the URL if it is, otherwise it'll remain as it was prior to this PR

amyblais:
Thanks!

marianunez (approved):
Great find! Mistery solved 👏

neallred (approved):

PR #19398, opened 2022-01-23
add support to store users recent emojis in user.props

### LINKED ISSUE #18603: Quick reactions in the hover menu and emoji picker should be based on most frequently used emoji instead of recently used
Currently quick reactions on mobile (long press menu) and webapp (hover menu) are based on recently used emoji.

image

We could improve this by basing the logic for which emoji appear based on "most frequently used" instead of "recently used". ~~The logic could be based on # reactions in the last week, with the top three showing in the hover menu.~~ It should be synced on the server so the most frequently used emoji are the same across all devices and sessions. 

*______* 

We should also add an option to **account settings** to set custom emoji. 

\< x ] Most frequently used
\[   ] Custom
Select the emoji you'd like to use

\[ :+1: ]  \[ :smile: ]  \[ :white*check*mark: ]   

\[   ] Off

 

Example:

 

----

JIRA: https://mattermost.atlassian.net/browse/MM-39150

darkLord19:
Fixes #18603

### DIFF
--- a/api4/status.go
+++ b/api4/status.go
@@ -17,6 +17,7 @@ func (api *API) InitStatus() {
 	api.BaseRoutes.User.Handle("/status", api.APISessionRequired(updateUserStatus)).Methods("PUT")
 	api.BaseRoutes.User.Handle("/status/custom", api.APISessionRequired(updateUserCustomStatus)).Methods("PUT")
 	api.BaseRoutes.User.Handle("/status/custom", api.APISessionRequired(removeUserCustomStatus)).Methods("DELETE")
+	api.BaseRoutes.User.Handle("/recentEmojis", api.APISessionRequired(removeUserCustomStatus)).Methods("PUT")
 
 	// Both these handlers are for removing the recent custom status but the one with the POST method should be preferred
 	// as DELETE method doesn't support request body in the mobile app.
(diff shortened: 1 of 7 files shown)


esethna:
Thanks @darkLord19! To clarify, is this storing recently used emoji or "most frequently used" emoji based on logic?

darkLord19:
@esethna it stores recently used emojis with their usage count. And then we render in the decreasing order of the usage count. This is as per my discussion with @hmhealey

esethna (approved):

esethna:
@darkLord19 great, thanks!

@jgilliam17 FYI this should be tested with the webapp PR

api4/status.go
```diff
@@ -17,6 +17,7 @@ func (api *API) InitStatus() {
 	api.BaseRoutes.User.Handle("/status", api.APISessionRequired(updateUserStatus)).Methods("PUT")
 	api.BaseRoutes.User.Handle("/status/custom", api.APISessionRequired(updateUserCustomStatus)).Methods("PUT")
 	api.BaseRoutes.User.Handle("/status/custom", api.APISessionRequired(removeUserCustomStatus)).Methods("DELETE")
+	api.BaseRoutes.User.Handle("/recentEmojis", api.APISessionRequired(updateUserRecentEmojis)).Methods("PUT")
```
hmhealey, line 20: I'm double checking with a server team, but I think the naming convention for this API might be `/recent_emojis`. I asked [here](https://community-daily.mattermost.com/core/pl/bxq6yh4guib9djr3cecttw6fby) on community

hmhealey (changes requested):
Glad to see this finally coming together!

One problem with this though is that `User.Props` actually isn't the correct spot for this since that field is more used for things that have to be accessible to other users. These should actually be stored in the `Preferences` table since that's where private settings and such for the user should be stored. Does that make sense?

Good/bad news with that though is that the `Preferences` table is designed to store any sort of data so it doesn't need server changes to accomplish that though.

darkLord19:
changed approach for feature which now requires only webapp changes so closing this

PR #19400, opened 2022-01-24
[MM-41202] Fix make apply if an app method has a function argument

hanzei:
#### Summary
As part of working on https://github.com/mattermost/mattermost-server/pull/19387, I've added a ` Go(f func())` method to `App`, which forwards to `Server`. When running `make app-layers`, `app/app_iface.go` get's correctly updated to include ` Go(f func())`, but `./app/layer_generators` fails with:
```
go run ./app/layer_generators -in ./app/app_iface.go -out ./app/opentracing/opentracing_layer.go -template ./app/layer_generators/opentracing_layer.go.tmpl
2022/01/21 14:06:33 ./app/opentracing/opentracing_layer.go:11671:46: expected 'IDENT', found ')'
exit status 1
make: *** [Makefile:263: app-layers] Error 1
```

This PR fixes the issue by not trying to fix an package name if the argument passed in a functional one. The fix surely doesn't count in for all cases e.g. there is still a extra case for `UploadFileX` which has `...func(*UploadFileTask)` as one argument. But it seemed to be, that the code should be re-written using AST anyway, so this fix follows the existing, already hacky approach.

There are no tests, so I didn't add another one. CI works as the test, as the old "target" for the tool is `app_iface.go` anyway.

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-41202

#### Release Note
```release-note
NONE
```

### DIFF
--- a/app/layer_generators/main.go
+++ b/app/layer_generators/main.go
@@ -85,7 +85,7 @@ func fixTypeName(t string) string {
 	if t == "...func(*UploadFileTask)" {
 		t = "...func(*app.UploadFileTask)"
 	}
-	if strings.Contains(t, ".") || strings.Contains(t, "{}") {
+	if strings.Contains(t, ".") || strings.Contains(t, "{}") || strings.Contains(t, "()") {
 		return t
 	}
 	typeOnly := textRegexp.FindString(t)


streamer45:
I am not sure I am totally convinced by the linked `Go(f func())` interface addition. I see that method more as an internal helper at this point, wondering if we could expose something less generic that accomplishes the same (e.g. through a callback on telemetry service initialization)?

hanzei:
Is there much of a difference between adding `Go(f func())` as a method to `telemetry.New` vs. adding it to `telemtry.ServerIface`?

streamer45:
> Is there much of a difference between adding `Go(f func())` as a method to `telemetry.New` vs. adding it to `telemtry.ServerIface`?

I meant avoiding to call that directly from the telemetry package. We could expose a server method to run a plugin hook or pass a callback during telemetry service initialization that essentially does the same. It's the server then that prepares the call to whatever is needed (`Go()` included) in here:

https://github.com/mattermost/mattermost-server/blob/9e9f0d2c44d2d18888b0b8f3205082aac0f748e2/app/server.go#L460

hanzei:
> It's the server then that prepares the call to whatever is needed (`Go()` included) in here:

What is needs is only the `Go()` method. The other things is a `plugin.Environment` and that is already there.

@streamer45 Would you be fine if I simply add `Go()` as a functional argument?

streamer45:
> > It's the server then that prepares the call to whatever is needed (`Go()` included) in here:
> 
> What is needs is only the `Go()` method. The other things is a `plugin.Environment` and that is already there.
> 
> @streamer45 Would you be fine if I simply add `Go()` as a functional argument?

Interesting, if telemetry has everything already do we even need to call it through this goroutine wrapper? 

@agnivade Asking for your input at this point as I don't want to block this further.

agnivade:
I am okay with this.

hanzei:
> Interesting, if telemetry has everything already do we even need to call it through this goroutine wrapper?

What is missing for `telemtry` is the `Go(f func())` method to run an "exit save" go routine. Alternatively, we could run the plugin hooks in the same go routine as the telemetry and block sending the server telemetry until the hook is called for all plugins. 0/5

hanzei:
@streamer45 What do you think about the following:
- We add a new service called `GoManager` which provides a `Go(f func())` and a `WaitForGoroutines()` method. 
- Both the server and the `TelemetryService` have there own `GoManager` which handlers go routines for them.
- In `TelemetryService.Shutdown()` it's ensure that all go routines have shut down for the `TelemetryService`

streamer45:
> @streamer45 What do you think about the following:
> 
> * We add a new service called `GoManager` which provides a `Go(f func())` and a `WaitForGoroutines()` method.
> * Both the server and the `TelemetryService` have there own `GoManager` which handlers go routines for them.
> * In `TelemetryService.Shutdown()` it's ensure that all go routines have shut down for the `TelemetryService`

You are taking it to the next level :) Honestly I'd defer to @agnivade as this is getting more involved and touching the server as well. I didn't mean to turn this into a completely different piece of work for you. We can always create a follow-up PR if necessary. I am not against the proposed changes.

hanzei:
@streamer45 The changes ended up being quite strait forward. You can find them in https://github.com/mattermost/mattermost-server/pull/19387. @agnivade I've added you as a reviewer there.

hanzei:
Closing this PR as it's not longer needed.

PR #19404, opened 2022-01-24
MM-41066: fixes replies when binary_parameters=yes

koox00:
#### Summary

Migrating post_store from gorp to sqlx breaks replies on threads.
Why? Participants is a jsonb field and when binary_parameters is set to
'yes' it is failing to insert because it needs the version (1) to be
prepended to the bytes array ([]byte{0x01}), see https://github.com/mattermost/mattermost-cloud/pull/477#issuecomment-878834792.

The easiest fix is to cast to a string when inserting.
There is a drawback though, non utf8 characters would be replaced by the
question mark icon �.

This commit does exactly that, casts StringArray to a string.

#### Ticket Link

https://mattermost.atlassian.net/browse/MM-41066

Closes https://github.com/mattermost/mattermost-server/issues/19420

#### Release Note

```release-note
Fixed inserting a `jsonb` field into the DB when `binary_parameters=yes`
```


model/utils.go
```diff
@@ -76,7 +76,12 @@ func (sa StringArray) Equals(input StringArray) bool {
 
 // Value converts StringArray to database value
 func (sa StringArray) Value() (driver.Value, error) {
-	return json.Marshal(sa)
+	j, err := json.Marshal(sa)
+	if err != nil {
+		return nil, err
+	}
+	// non utf8 characters are not supported
+	return driver.Value(string(j)), err
```
agnivade, line 84: There's no need to convert to `driver.Value`. It's an `interface{}`, so will allow anything.

agnivade (approved):
Thanks!

ashishbhate (approved):

stafot:
cc @koox00 Tests passed, looks ok to merge.

Piusgodw (commented):
<spam>

PR #19408, opened 2022-01-24
use custom types StringInterface and StringArray

mxschumacher:
#### Summary

The utils.go file in the models package defines two custom types:

```go
type StringInterface map[string]interface{}
type StringArray []string
```

In this PR, those two types are used in all type definitions and function signatures of the utils.go file

In a next step, we can consider using the custom types in all files of the model package

#### Ticket Link

no ticket link, it is a small PR.

#### Release Note

```
NONE
```

### DIFF
--- a/model/utils.go
+++ b/model/utils.go
@@ -162,7 +162,7 @@ type AppError struct {
 	StatusCode    int    `json:"status_code,omitempty"` // The http status code
 	Where         string `json:"-"`                     // The function where it happened in the form of Struct.Func
 	IsOAuth       bool   `json:"is_oauth,omitempty"`    // Whether the error is OAuth specific
-	params        map[string]interface{}
+	params        StringInterface
 }
 
 func (er *AppError) Error() string {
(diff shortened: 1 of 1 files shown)


agnivade:
Hey @mxschumacher - I'd suggest not to do this unless necessary. The custom types are mainly used to access their methods in some places. Other than that, they don't serve much purpose. I agree that this will make things inconsistent, but they should only be used if we need to access their methods. Otherwise, it becomes confusing and hard to read.

mxschumacher:
Hello @agnivade - okay, understood. Thanks for the fast feedback. While reading the code I noticed that sometimes these custom types are used and sometimes they are spelled out anew - so I thought my PR would bring some clarity (eventually applying this change to the entire model package). I'll close the PR!

PS: Do you know why the "release-note-label-needed" tag is has been put on this PR? I used the "NONE" section from the PR template

agnivade:
Thanks @mxschumacher :) Please feel free to contribute in other parts of the code, and before sending a PR, just drop a comment in Contributors or Ask R&D channel in our community server.

The PR template seems to be incorrect. The format needs to be:

```
```release-note
NONE
```
```

PR #19410, opened 2022-01-25
Disable sentry tracing completely

agnivade:
We don't do instrumentation of Sentry transactions
anywhere in the code. Only opentracing code is present.

One has to explicitly add custom instrumentation for it
to work: https://docs.sentry.io/platforms/go/performance/instrumentation/custom-instrumentation/

It's likely someone is running custom code and sending
transaction data to Sentry. It's not super hard to pick
off the sentry DSN string from the binary.

The PR is more of a best effort to stop any future
transaction events to be sent.

```release-note
NONE
```

### DIFF
--- a/app/server.go
+++ b/app/server.go
@@ -343,6 +343,9 @@ func NewServer(options ...Option) (*Server, error) {
 					}
 					return event
 				},
+				TracesSampler: sentry.TracesSamplerFunc(func(ctx sentry.SamplingContext) sentry.Sampled {
+					return sentry.SampledFalse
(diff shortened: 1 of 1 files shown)


agnivade:
Wait my bad, the errors are also transactions. Hmm .. sending a proper fix.

agnivade:
Verified. All good. We actually have a test for sending event to a dummy sentry endpoint. I am suddenly impressed :)

streamer45 (approved):
Nice :+1:

PR #19411, opened 2022-01-25
MM-41350] All CRT server operations should only happen if ThreadAutoFollow is true

ashishbhate:
#### Summary

- CRT server book-keeping should only happen if `ThreadAutoFollow` is true.
 
#### Ticket Link

- https://mattermost.atlassian.net/browse/MM-41350
#### Release Note

```release-note
On the server, CRT Threads will only be kept up to date if ThreadAutoFollow config is set to true
```


app/channel.go
```diff
@@ -2912,7 +2912,7 @@ func (a *App) MarkChannelsAsViewed(channelIDs []string, userID string, currentSe
 		a.clearPushNotification(currentSessionId, userID, channelID, "")
 	}
 
-	if !collapsedThreadsSupported || !a.IsCRTEnabledForUser(userID) {
+	if *a.Config().ServiceSettings.ThreadAutoFollow && (!collapsedThreadsSupported || !a.IsCRTEnabledForUser(userID)) {
 		if err := a.Srv().Store.Thread().MarkAllAsReadInChannels(userID, channelIDs); err != nil {
 			return nil, model.NewAppError("MarkChannelsAsViewed", "app.channel.update_last_viewed_at.app_error", nil, err.Error(), http.StatusInternalServerError)
```
ashishbhate, line 2917: @Willyfrog @koox00 

Here we mark every thread in a channel as read if user doesn't have CRT enabled (or client is not supported). 

My suggestion is that we don't do this if `ThreadAutoFollow` is false. 
A long time ago `ThreadAutoFollow` only decided whether a thread should be auto-followed for a user. Now we're using `ThreadAutoFollow` as the check for whether the server should do the backend book-keeping. 

Therefore, if `ThreadAutoFollow` is false, I think we shouldn't update a user's threads. 

Pros:
1. Better performance. We wont be running a heavy query on every channel view (i.e. every channel switch)

Cons:
1. The user's threads will not be marked as read on viewing a channel. This might lead to a huge number of unread threads when the server finally turns on CRT. I don't think this is huge problem, given that backend bookkeeping is expected not to happen when `ThreadAutoFollow` is false.

Willyfrog: I agree we are treating `ThreadAutoFollow` as the bookeeping flag, and if so it makes sense the proposal.

Regarding the cons: we should prepare a migration/job when we approach GA for easing the change on big installs.

esethna: > The user's threads will not be marked as read on viewing a channel. This might lead to a huge number of unread threads when the server finally turns on CRT. I don't think this is huge problem, given that backend bookkeeping is expected not to happen when ThreadAutoFollow is false.

@ashishbhate @Willyfrog if ThreadAutoFollow is off are we even populating the threadmemberships and related tables? I thought the ThreadAutoFollow was the switch that turns off all CRT related backend work - and if that's the case then shouldn't disabling it mean there's no threads to mark as read anyway?

ashishbhate: @esethna No new CRT threads will be created so the the number of unread threads will not increase. But all older unread threads will remain unread.

koox00 (approved):
💪

ashishbhate:
This fix needs to go into release 5.37, 6.3 and 6.4.

// cc @amyblais

Willyfrog (approved):

Willyfrog:
Great job finding those 🚀

amyblais:
> This fix needs to go into release 5.37, 6.3 and 6.4.
> 
> // cc @amyblais

Yes, also 6.1 and 6.2.

amyblais:
@ashishbhate Let me know if there is anything I can help with this and the other cherry-pick candidate PR.

amyblais:
@ashishbhate I believe this wasn't cherry-picked to the default ``release-6.3``, ``release-6.2``, ``release-6.1`` and ``release-5.37`` branches yet.

ashishbhate:
@amyblais I have PRs open for the cherry-picks. I'll merge then as soon as they pass CI

amyblais:
I posted too fast again, I see the PRs now. :)

ashishbhate:
@amyblais all cherry-picks done (6.3, 6.2, 6.1, 5.37)

PR #19416, opened 2022-01-26
MM-41209: Fix flaky test CreateRecurringTaskFromNextIntervalTime

agnivade:
We were noting down the time of execution of a function and
expected that to match with the exact interval that we set.

However, sometimes in very busy CI environments, the goroutine
does not get a chance to be scheduled at the right time for
the interval to match exactly. Sometimes, we see a delay of
1 second.

To account for this, we increase the interval to 3 seconds,
and consider a 1 second buffer for this. If we see lags of
more than a second, then this test becomes useless and we
would need to remove it entirely.

https://mattermost.atlassian.net/browse/MM-41209

```release-note
NONE
```

### DIFF
--- a/model/scheduled_task_test.go
+++ b/model/scheduled_task_test.go
@@ -76,8 +76,8 @@ func TestCancelTask(t *testing.T) {
 }
 
 func TestCreateRecurringTaskFromNextIntervalTime(t *testing.T) {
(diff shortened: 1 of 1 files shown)


streamer45 (approved):
Looks good :+1:

PR #19417, opened 2022-01-26
MM-41236: Sentry crash: Fix nil reference to token

agnivade:
The AND condition would mean that it would try to dereference
token.Valid if there was an error. And there's no guarantee
to always have a non-nil token in case of an error. We need
to track those conditions separately.

https://mattermost.atlassian.net/browse/MM-41236

```release-note
NONE
```


nickmisasi (approved):

AGMETEOR (approved):
👍🏿

PR #19418, opened 2022-01-26
MM-41184: Remove /config/migrate API endpoint

agnivade:
This was introduced to let the functionality happen
via mmctl, but it does not differentiate between
sysadmins vs people having physical access to machines.

Therefore, we remove it and only keep the local endpoint
accessible from mmctl --local.

https://mattermost.atlassian.net/browse/MM-41184

```release-note
The api/v4/config/migrate API endpoint has been removed
in favor of the mmctl --local endpoint. API clients
won't be able to access this endpoint without having
physical access to the server.
```


srkgupta (approved):
Thanks @agnivade. Changes LGTM.

isacikgoz (commented):
Looks good, @agnivade before merging this one shall we also create a companion PR for mmctl to remove `config migrate` command from non-local mode? I think we can just return an error that this command will be used local mode only. And also we may want to update documentation too.

agnivade:
Sure

isacikgoz (approved):
LGTM 👍

srkgupta:
Hi @amyblais 
Can this be cherry picked to 6.4 release?

amyblais:
Seems to need a manual cherry-pick

PR #19424, opened 2022-01-26
Adding the Scorecard analysis

oh6hay:
Github Action for Scorecard
https://github.com/ossf/scorecard

The Scorecard provides a (somewhat opinionated) view on the security posture of the repository -- we (the product security team) would like to adopt this on our key repositories to identify possible security improvements

### DIFF
--- a/.github/workflows/scorecards-analysis.yml
+++ b/.github/workflows/scorecards-analysis.yml
@@ -1,0 +1,55 @@
+name: Scorecards supply-chain security
+on:
+  # Only the default branch is supported.
+  branch_protection_rule:
+  schedule:
+    - cron: '44 6 * * 4'
+  push:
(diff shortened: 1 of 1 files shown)


.github/workflows/scorecards-analysis.yml
```diff
@@ -0,0 +1,55 @@
+name: Scorecards supply-chain security
+on:
+  # Only the default branch is supported.
+  branch_protection_rule:
+  schedule:
+    - cron: '44 6 * * 4'
```
cpanato, line 6: nit: worth make a comment in human reading to let people know when this will be schedule to run

cpanato (approved):
looks great, just a nit which is nonblocking, and add some explanation why we need that in the pr summary

thanks!

PR #19433, opened 2022-01-27
[MM-41330] Fix JSON type check in `hasMissingMigrationsVersion600`

streamer45:
#### Summary

MySQL returned type for JSON would be lowercase but we checked against an uppercase string, hence failing the check and re-triggering the migration every time.

/cc @neillcollie @isacikgoz 

@amyblais This goes directly into 6.3 release and will need to be included in a dot release. From what I can see bug also affects 6.0.x,6.1.x,6.2.x releases.

#### Ticket Link

https://mattermost.atlassian.net/browse/MM-41330

#### Release Note

```release-note
Fixed an issue that on MySQL installations would re-trigger 6.0 migration on server restart.
```

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -310,7 +310,6 @@ jobs:
       - run:
           name: Run Tests
           command: |
-            ulimit -n 8096
             mkdir -p mattermost-server/client/plugins
             cd mattermost-server/build
             docker run -it --net circleci_mm-test \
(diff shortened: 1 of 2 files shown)


neillcollie:
Thanks Claudio. Will this affect the 5.37.x branch as well perhaps? I have another customer reporting similar issue

streamer45:
> Thanks Claudio. Will this affect the 5.37.x branch as well perhaps? I have another customer reporting similar issue

This specific issue is 6.0 onward only as the affected migration was introduced in 6.0 .

agnivade (approved):
Sorry, I somehow thought that I had approved this.

amyblais:
@streamer45 Do you recommend also cherry-picking this to 6.2 / 6.1? (I'm ok with it).

streamer45:
> @streamer45 Do you recommend also cherry-picking this to 6.2 / 6.1? (I'm ok with it).

@amyblais That should be done already.

PR #19444, opened 2022-01-28
[MM-40582] Send unsanitized user to source on user_updated

plant99:
#### Summary

On `user_update` event,  unsanitized user is sent to source of the event, to prevent extra ping to get user. Sanitized user data is sent to other users.

#### Ticket Link

Fixes https://github.com/mattermost/mattermost-server/issues/19361

#### Release Note

```release-note
Added feature to send unsanitized user to source user on `user_updated` event.
```


plant99:
cc: @agnivade 

Changes to webapp, mobile code are pending from my end. Meanwhile, please have a look if this logic looks right to you.

app/user.go
```diff
@@ -1025,17 +1025,28 @@ func (a *App) UpdateUserAuth(userID string, userAuth *model.UserAuth) (*model.Us
...
+	omitUsers[user.Id] = true
 	a.SanitizeProfile(&user, false)
-	message := model.NewWebSocketEvent(model.WebsocketEventUserUpdated, "", "", "", nil)
+	message := model.NewWebSocketEvent(model.WebsocketEventUserUpdated, "", "", "", omitUsers)
 	message.Add("user", &user)
 	message.GetBroadcast().ContainsSanitizedData = true
 	a.Publish(message)
+
+	// send unsanitized user to event creator
+	sourceUserMessage := model.NewWebSocketEvent(model.WebsocketEventUserUpdated, "", "", unsanitizedCopyOfUser.Id, nil)
```
enahum, line 1046: correct me if I'm wrong, but I'm seeing 3 events being published here, shouldn't we ensure that the user only receives the one event.

What I see here is
`adminMessage` being broadcasted to everyone (this seems wrong, shouldn't this omit the source creator as well?)
`message` being broadcasted excluding the source creator (this looks ok to me)
`sourceUserMessage` being broadcasted only to the source creator (this looks ok to me)

plant99: I think https://github.com/mattermost/mattermost-server/blob/master/app/web_conn.go#L721 forwards to all admins, so I get two events (one `sourceUserMessage ` and one `adminMessage`) when admin is the data source, thanks for pointing this out. I'll omit `sourceUser` from adminMessage.

plant99: Please let me know if the latest commit fixes it?

agnivade: Ah my apologies on that one. You are correct @enahum. Great catch!

app/web_conn.go
```diff
@@ -732,7 +732,8 @@ func (wc *WebConn) shouldSendEvent(msg *model.WebSocketEvent) bool {
 			hasReadPrivateDataPermission = model.NewBool(wc.App.RolesGrantPermission(wc.GetSession().GetUserRoles(), model.PermissionManageSystem.Id))
 		}
 
-		if !*hasReadPrivateDataPermission {
+		// sensitive data should be allowed to be emitted back to source-user event.
+		if !*hasReadPrivateDataPermission && (wc.UserId != msg.GetBroadcast().UserId) {
```
agnivade, line 736: We can revert this change. We only need to set `ContainsSanitizedData`, when we need to differentiate between users and admins. It's not perfectly made, but `ContainsSensitiveData` goes together with `ContainsSanitizedData` to separate users and admins.

Since we are directly sending the message to the user, we can ignore this part and it will match the condition in L742.

agnivade (commented):
Looking great @plant99 ! Just some small changes needed.

app/web_conn.go
plant99: Understood the context, thanks @agnivade for the explanation. I've reverted these changes. Please review again?

hmhealey:
@agnivade FYI, this is probably something that we just accept with this, but this could cause some weirdness with the cross-version cluster we've got on community. Theoretically, a user on community-daily could receive a santized version of themselves if they were updated by someone (themselves or an admin) on community. I don't think that would cause anything other than a temporary issue though, and I doubt it'll occur often anyway though, but I wanted to point out that it could happen in case someone else thinks up a way that could cause any other issues.

agnivade:
The cross-version cluster is something that's going to go away soon. So I wouldn't be too worried about that. But yes, definitely a good point. :+1:

streamer45 (approved):
Looks great :+1:

agnivade:
@hmhealey - Are we okay to merge, or you would prefer the webapp PR to be merged along with this? 0/5 from my side. LMK.

hmhealey:
This is good to be merged first. It won't really have any effect on the front end until the web app changes are merged anyway.

PR #19484, opened 2022-02-02
[MM-41343] Password reset must be 24hour

JulienTant:
#### Summary
Password reset message says the password reset link is good for 24 hours when it expires after one hour.

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-41343

#### Release Note
```release-note
* Fixed a bug causing the password reset link to be valid for 1 hour instead of 2 4hours.
```

### DIFF
--- a/app/user.go
+++ b/app/user.go
@@ -35,7 +35,7 @@ const (
 	TokenTypeTeamInvitation    = "team_invitation"
 	TokenTypeGuestInvitation   = "guest_invitation"
 	TokenTypeCWSAccess         = "cws_access_token"
-	PasswordRecoverExpiryTime  = 1000 * 60 * 60      // 1 hour
+	PasswordRecoverExpiryTime  = 1000 * 60 * 60 * 24 // 24 hours
 	InvitationExpiryTime       = 1000 * 60 * 60 * 48 // 48 hours
 	ImageProfilePixelDimension = 128
 )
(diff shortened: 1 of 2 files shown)


app/user.go
```diff
@@ -1251,11 +1251,15 @@ func (a *App) UpdateHashedPassword(user *model.User, newHashedPassword string) *
 }
 
 func (a *App) ResetPasswordFromToken(userSuppliedTokenString, newPassword string) *model.AppError {
+	return a.resetPasswordFromToken(userSuppliedTokenString, newPassword, model.GetMillis())
```
JulienTant, line 1254: A word on this, because `go` does not allow monkey-patching, the easiest way* to "mock" the time is to have it as an argument. SO the "Public" method is unchanged but having the code in a private method allows us to mock the time.

* the hard way/"don't touch my production code way" would be to have the `App` struct hold a `Clock` interface that depends on `time` by default but would be mockable... which is quite a huge refactor 😅

neallred (approved):
LGTM. Nice tests!

app/user.go
neallred: > Sometimes, the elegant implementation is just a function. Not a method. Not a class. Not a framework. Just a function.

-- John Carmack

pablovelezvidal (approved):
LGTM, nice work @JulienTant

stevemudie (approved):
Tested and passed. Great work @JulienTant!

The reset email I first sent out was expired after 24 hours and then a second one worked as expected after a few hours as did one as soon as I sent it out.

PR #19485, opened 2022-02-03
MM-38164: Paginate the GetPostThread API

agnivade:
We implement a cursor based pagination model
to page through the posts in a given thread.

The cursor is a combination of the post.CreateAt+
post.Id to differentiate multiple posts in a given
timestamp.

Some additional parameters like direction, fromPost,
fromCreateAt and perPage were introduced to implement
this.

```release-note
NONE
```


agnivade:
cc @michelengelen as well

api4/post.go
```diff
@@ -503,10 +504,56 @@ func getPostThread(c *Context, w http.ResponseWriter, r *http.Request) {
...
+		fromCreateAt, err = strconv.ParseInt(fromCreateAtStr, 10, 64)
+		if err != nil {
+			c.SetInvalidParam("fromCreateAt")
+			return
+		}
+	}
+
+	fromPost := r.URL.Query().Get("fromPost")
+	// Either both have to be set, or none have to be set.
+	// Setting one and not setting the other is an error.
```
koox00, line 533: out of curiosity why is that? couldn't we just support one or the other?

koox00: nvm I read the query!

michelengelen (approved):
LGTM ... although i would advise to get one more opinion/review from someone more proficient with Golang to review as well

agnivade:
Thanks @michelengelen :) This was more of a cc since you will be implementing the client side of this. Kyriakos and Ashish will take a deeper look at this.

michelengelen:
> Thanks @michelengelen :) This was more of a cc since you will be implementing the client side of this. Kyriakos and Ashish will take a deeper look at this.

ok, thanks for the explanation! 🙇🏼

koox00 (approved):
This is great, and it opens the possibility to ask for replies newer than the last one, each time we open the RHS!

agnivade:
> This is great, and it opens the possibility to ask for replies newer than the last one, each time we open the RHS!

Yep, without loading the _entire_ thread all over again!

cc @enahum as well for visibility. Elias - does mobile use this API? If so, now it's paginated and you can get replies after a given point.

enahum:
Ok, does not seem to be a breaking change, so that is good I guess. ESR or not we should be really careful when introducing a breaking change not only cause of mobile ;)

As for this API being used in mobile, well the same team implementing this on web should do the implementation across clients including mobile and take care of backwards compatibility with older servers :)

Thanks @agnivade for the shout.

cc: @Willyfrog to increase visibility on this

Willyfrog:
@enahum yes, I'm not concerned about v2, but only if we need to do this change in v1 as well to ensure the benefits are there as soon as possible

agnivade:
> Ok, does not seem to be a breaking change, so that is good I guess. ESR or not we should be really careful when introducing a breaking change not only cause of mobile ;)

Of course, no breaking changes :). Mobile is always in my mind :heart:

saturninoabril:
@agnivade Looks like something is wrong with the mattermod itself, though I can't check further as I don't have access to it even logging. I triggered the test manually, see https://automation-dashboard.vercel.app/cycles/9bcc0260-bfb9-4dd1-a9cb-ffbd1c1305a6

@metanerd Please help check if you get a chance. Thanks!

agnivade:
@saturninoabril - Are there any unexpected failures in the e2e test?

api4/post_test.go
```diff
@@ -2183,6 +2183,22 @@ func TestGetPostThread(t *testing.T) {
...
+	})
+	require.Error(t, err)
+	CheckBadRequestStatus(t, resp)
+
+	_, resp, err = client.GetPostThreadWithOpts(th.BasicPost.Id, "", model.GetPostsOptions{
+		CollapsedThreads: true,
+		Direction:        "sideways",
+	})
+	require.Error(t, err)
+	CheckBadRequestStatus(t, resp)
```
saturninoabril, line 2200: Is there an existing test that verifies exact returned values, depending on perPage option?

saturninoabril:
Thanks @agnivade, this is great and looking forward to this improvement! 

Passed on Cypress test. No regression while 8 failed tests were known issues - https://automation-dashboard.vercel.app/cycles/f79527a8-e223-4d75-bf62-79930255addc

Just curious about the following:
1. Is there a load tests result for our reference with highlights on improvement from this PR?
2. I wonder if the same should be made for the following APIs as well
  - `GET /api/v4/channels/<channel-id>/posts`
  - `GET /api/v4/users/<user-id>/channels/<channel-id>/posts/unread`
3. Once all good, is this going to be merged to master, or should it wait for corresponding change in client (at least in webapp) to fully tests the performance improvement?

api4/post_test.go
agnivade: Nice, I see you are starting to look into tests :)

Yes, all the pagination tests are in the store layer in storetest/post_store.go. The api layer just tests the api layer restrictions.

agnivade:
1. That's on my agenda to add. :)
2. This was originally raised from a security issue by the security team. Ideally all endpoints should be paginated, and that's how it will be once we completely move to GraphQL. Speaking of those, please feel free to raise this with the Apollo team to get it scheduled.
3. I'll block this until I finish the load-tests. @michelengelen can start work on the client side in parallel. If the load-tests finish first, we can merge this.

saturninoabril:
> @agnivade Looks like something is wrong with the mattermod itself, though I can't check further as I don't have access to it even logging. I triggered the test manually, see https://automation-dashboard.vercel.app/cycles/9bcc0260-bfb9-4dd1-a9cb-ffbd1c1305a6
> 
> @metanerd Please help check if you get a chance. Thanks!

Thanks Elisabeth for fixing the slash command for E2E testing! Finally it's working in this server repo!

ashishbhate (approved):

agnivade:
@ashishbhate has graciously accepted to add the load-test coverage as part of his initiative to bring CRT coverage at par with the current feature set. Thank you Ashish for the help!

Willyfrog:
@ashishbhate how is that loadtest going for this?

ashishbhate:
>@ashishbhate how is that loadtest going for this?

The review fixes for that PR is back to being my main focus. In any case, I think we should merge this PR. The load test PR will take a little longer to finish reviews, let's not block this PR from going in. The changes here have been incorporated into that PR, and keeping this PR up to date is unnecessary work.

//cc @agnivade

agnivade:
@ashishbhate - My original idea [here](https://github.com/mattermost/mattermost-server/pull/19485#issuecomment-1029815544) was to block the PR until the load-tests are done. But as long as the client side PR isn't merged without finishing the load-tests, I am fine with it.

I'd want explicit sign off from @saturninoabril on this.

In general, client side work tend to always get merged as soon as reviews are done. If this goes in without load-tests, I'd want an explicit owner who would be responsible for ensuring that the load-tests are tested and finished before the feature is merged.

cc @Willyfrog

saturninoabril (approved):
> But as long as the client side PR isn't merged without finishing the load-tests, I am fine with it.

+1 on this.

Thanks for checking! I won't block this PR. I initially thought to look at load test results primarily for QA/Release quick check and reference since there were changes made to store/queries.

Feel free to merge as you see fit.

agnivade:
Ok let's do that. If this means prioritizing load-tests higher up, I am fine with it.

ashishbhate:
The loadtest PR - https://github.com/mattermost/mattermost-load-test-ng/pull/477 has been approved. I'll merge this PR once CI passes and then update and merge the loadtest PR.

agnivade:
Nice work! I'd like to see the results of the load-test so that we can quantify the impact of the changes in this PR.

agnivade:
I've just observed major performance regressions due to the way the SQL queries are written. This won't work without additional indexes in the posts table. I'll be looking into this.

Just one more reason to always run load-tests :)

PR #19486, opened 2022-02-03
GraphQL (part 2): Model changes and resolvers

agnivade:
This PR adds the necessary resolvers to support
the graphQL front-end. Some additional methods
need to be added to the model structs to support
this. Comments have been made to clarify their purpose.

There is a slight duplication of code in the api layer.
But eventually that's going to go away when we move
away from the REST API entirely.

The schema is in api4/schema.graphqls and gets compiled
into the binary as an asset.

The GraphiQL editor url is at GET /api/v5/graphql.

Everything is behind the feature flag MM_FEATUREFLAGS_GRAPHQL.

```release-note
NONE
```


agnivade:
cc @jupenur @hmhealey @M-ZubairAhmed 

@agarciamontoro (in case you want to take a deeper look)

agnivade:
@isacikgoz - I thought of splitting this further into the model changes, but they were fairly small and relevant to the graphql changes. Let me know if you want to split this further.

isacikgoz:
@agnivade It's alright, let's go as is.

jupenur:
Thanks Agniva, Osku will take a look :)

oh6hay (commented):
LGTM. Would it make sense to also write tests for queries for resources for which the current user shouldn't have access to, and check that the queries return errors?

agnivade:
All of the queries are at a userId level. There are some tests where I pass invalid ids, and test that we get an error. But I'll do another pass to see if I can add some more negative tests. Thanks Osku!

api4/resolver.go
```diff
@@ -0,0 +1,281 @@
...
+	if err != nil {
+		return nil, err
+	}
+
+	if args.UserID == model.Me {
+		args.UserID = c.AppContext.Session().UserId
+	}
+
+	// If it's a single channel
+	if args.ChannelID != "" {
```
jupenur, line 227: This is missing permission checks for the case where `args.ChannelID == ""`. Pass in an arbitrary user ID and get info about their channel memberships.

agnivade: Ohh dear me, what a bug. :cry:  Amazing eye there @jupenur :pray:

jupenur: I'm actually pretty happy with how GraphQL works even when there's a bug like this one. Because the permission checks still kick in when you hit any other resolver, the impact is very limited. With an equivalent bug in the REST APIs, the impact would be similarly limited thanks to the scoping of the individual endpoints, but there's one major difference: with REST you could still read the channel IDs of the channels you don't have access to. That would allow you to uniquely identify the different channel memberships -- which would then allow you to accurately track an arbitrary user joining and leaving private channels. With this bug you can still do some of that tracking, but to a much lesser extent because you can't reliably distinguish one channel membership from another.

So for a positive angle on this, it was actually valuable for me to see a bug like this play out. It gives me confidence that adopting GraphQL could actually end up improving our security resilience 🙂

agnivade:
Added some more negative tests and fixed the issue @jupenur and @oh6hay. PTAL. Thanks for the solid review!

jupenur (approved):
LGTM!

api4/apitestlib.go
```diff
@@ -45,6 +49,7 @@ type TestHelper struct {
 
 	Context              *request.Context
 	Client               *model.Client4
+	GraphQLClient        *model.Client4
```
isacikgoz, line 52: If we are going to use `v5` for the API, would it make sense this be a new Client? I mean would it confuse the users of the library?

api4/graphql.go
```diff
@@ -0,0 +1,156 @@
...
+	w.Header().Set("Content-Type", "text/html")
+	w.Write(graphiqlPage)
+}
+
+var graphiqlPage = []byte(`
+<!DOCTYPE html>
+<html>
+	<head>
+		<title>GraphiQL editor | Mattermost</title>
+		<link href="https://cdnjs.cloudflare.com/ajax/libs/graphiql/0.11.11/graphiql.min.css" rel="stylesheet" />
```
isacikgoz, line 116: Hmm, why we are not hosting these with the app server?

isacikgoz (commented):
This is awesome @agnivade 🚀🚀 Did a quick pass, I have some high level questions before diving into the implementation.

api4/apitestlib.go
agnivade: Yeah, good point. I'll create a new one. :+1:

api4/graphql.go
agnivade: Not a bad idea, it's just that this was the simplest way to implement it. Created: https://mattermost.atlassian.net/browse/MM-41614

api4/apitestlib.go
agnivade: @isacikgoz - I created a separate graphql client in the api4 package itself to prevent it from polluting the model API. I'd want to move it out there when we have a stable API and the methods are more fleshed out.

api4/resolver_channel.go
```diff
@@ -0,0 +1,142 @@
...
+		}
+	case "full_name":
+		displayName = user.GetFullName()
+		if displayName == "" {
+			displayName = user.Username
+		}
+	case "username":
+		displayName = user.Username
+	default:
+		displayName = user.Username
```
isacikgoz, line 138: Can these two be merged?

isacikgoz (commented):
Haven't go into test files yet. Before doing the last pass, let me understand things a bit more. So I have some questions below :)

api4/graphql.go
```diff
@@ -0,0 +1,156 @@
...
+			if err := json.NewEncoder(w).Encode(response); err != nil {
+				mlog.Warn("Error while writing response", mlog.Err(err))
+			}
+		}
+	}()
+
+	// Limit bodies to 100KiB.
+	// We need to enforce a lower limit than the file upload size,
+	// to prevent the library doing unnecessary parsing.
+	r.Body = http.MaxBytesReader(w, r.Body, 102400)
```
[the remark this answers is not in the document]
agnivade: Sure, I can add a test case. If we pass the limit, it returns a "request body too large" error - https://cs.opensource.google/go/go/+/refs/tags/go1.17.6:src/net/http/request.go;l=1178.

I'll change it to a constant as well.

api4/graphql.go
```diff
@@ -0,0 +1,156 @@
...
+	reqCtx := r.Context()
+	reqCtx = context.WithValue(reqCtx, ctxKey{}, c)
+
+	response = api.schema.Exec(reqCtx,
+		params.Query,
+		params.OperationName,
+		params.Variables)
+
+	if len(response.Errors) > 0 {
+		mlog.Error("Error executing request", mlog.String("operation", params.OperationName),
```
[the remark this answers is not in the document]
agnivade: Nice one. I'll update it.

api4/graphql_client.go
```diff
@@ -0,0 +1,87 @@
...
+	"encoding/json"
+	"io"
+	"net/http"
+	"strings"
+
+	"github.com/mattermost/mattermost-server/v6/model"
+)
+
+// graphQLClient is an internal test client to run the tests.
+// When the API matures, we will expose it to the model package.
```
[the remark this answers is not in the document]
agnivade: Yes, move and expose.

api4/resolver.go
```diff
@@ -0,0 +1,286 @@
...
+		c.SetPermissionError(model.PermissionViewTeam)
+		return nil, c.Err
+	}
+
+	limit := int(args.First)
+	// ensure args.First limit
+	if limit == 0 {
+		limit = web.PerPageDefault
+	} else if limit > web.PerPageMaximum {
+		return nil, fmt.Errorf("first parameter %d higher than allowed maximum of %d", limit, web.PerPageMaximum)
```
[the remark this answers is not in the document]
agnivade: Good one. Will change.

api4/resolver.go
```diff
@@ -0,0 +1,286 @@
...
+	}
+
+	// ensure args.After format
+	var afterChannel string
+	var ok bool
+	if args.After != "" {
+		afterChannel, ok = parseChannelCursor(args.After)
+		if !ok {
+			return nil, fmt.Errorf("after cursor not in the correct format: %s", args.After)
+		}
```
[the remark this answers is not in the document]
agnivade: Hmm .. I think this is okay. Does it look scattered because of the empty string check? I'm not overly opinionated either ways, but wanted to check if there's any other reason the logic seems scattered to you.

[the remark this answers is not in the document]
isacikgoz: Yeah, I though the empty string check can be done in `parseChannelCursor`. 0/5 though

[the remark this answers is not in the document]
agnivade: Ok, keeping it as is for now. Let me know if you feel otherwise.

api4/resolver.go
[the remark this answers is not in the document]
agnivade: Actually, I now remember why I did it this way. Our errors do not convey enough information as to what really went wrong. The `c.SetInvalidParam` will just tell you that parameter x was invalid. No further information on what/how to correct it. That's why I went with my own error with properly mentioned reason for the error.

crspeller (commented):
This is great 😄 
Just wanted to ask a question.

api4/schema.graphqls
```diff
@@ -0,0 +1,157 @@
...
+scalar SidebarCategoryType
+
+scalar SidebarCategorySorting
+
+scalar StringMap
+
+scalar Time
+
+type Channel {
+	id           : String!
```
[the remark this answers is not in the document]
agnivade: Yeah, good point. Actually these ids are basically the ids we have in the DB for an object. But the `ID` scalar type is meant as an opaque value not intended to be human-readable, and moreover, it's meant to serve as a globally unique identifier for the object and satisfy the node interface.

```
# An object with a Globally Unique ID
interface Node {
  # The ID of the object.
  id: ID!
}
```

So essentially, this is going to be the cursor field for an object, because the cursor is a base64 encoded value containing the object name and it's ID, so that on the server, we can decode the value, to get the object type, and then query the appropriate table for that value.

So this `id` field now will need to change to `objectId` or something and the cursor field will be replaced by the `id`. I'll do it after this PR.

api4/graphql_client.go
```diff
@@ -0,0 +1,87 @@
+// Copyright (c) 2015-present Mattermost, Inc. All Rights Reserved.
+// See LICENSE.txt for license information.
+
+package api4
```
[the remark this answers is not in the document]
agnivade: This didn't work because the non-test code references the client. Reverting it.

api4/resolver.go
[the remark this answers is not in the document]
isacikgoz: I see, makes sense. But, what would be the response code here? `http.StatusBadRequest`?

api4/resolver_channel_member_test.go
```diff
@@ -0,0 +1,355 @@
...
+
+		numPrivate := 0
+		numPublic := 0
+		numOffTopic := 0
+		numTownSquare := 0
+		for _, ch := range q.ChannelMembers {
+			assert.NotEmpty(t, ch.Channel.ID)
+			assert.NotEmpty(t, ch.Channel.Name)
+			assert.NotEmpty(t, ch.Channel.CreateAt)
+			assert.NotEmpty(t, ch.Channel.UpdateAt)
```
isacikgoz, line 126: Wondering if we could leverage `require.NotZero` here to avoid many assertions. Eg:
```suggestion
			require.NotZero(t, ch.Channel)
```

api4/resolver_channel_member_test.go
```diff
@@ -0,0 +1,355 @@
...
+	  }
+	}
+	`,
+		}
+
+		resp, err := th.MakeGraphQLRequest(&input)
+		require.NoError(t, err)
+		require.Len(t, resp.Errors, 0)
+		require.NoError(t, json.Unmarshal(resp.Data, &q))
+		assert.Len(t, q.ChannelMembers, 9)
```
isacikgoz, line 116: Instead of using magic numbers here, would you think it'd be more robust to get these numbers upfront? Any change on the testlib regarding to users may break these tests right?

api4/resolver_channel_member_test.go
```diff
@@ -0,0 +1,355 @@
...
+				assert.Equal(t, float64(2), ch.MsgCount)
+			case ch1.Id:
+				assert.Equal(t, float64(1), ch.MsgCount)
+			}
+		}
+
+		assert.Equal(t, 2, numPrivate)
+		assert.Equal(t, 7, numPublic)
+		assert.Equal(t, 2, numOffTopic)
+		assert.Equal(t, 2, numTownSquare)
```
isacikgoz, line 165: Same here, maybe we can get these numbers with the v4 API and check against those?

isacikgoz (commented):
Thanks for your patience @agnivade I think we are almost there, I have some minor improvement suggestions. I think this is almost good to go.

api4/resolver.go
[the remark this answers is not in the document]
agnivade: This is no longer REST any more. Status codes don't have a meaning :) They are always 200. Errors are collected in a separate json array. They will appear there. What we could have is some sort of structure there for clients to parse. We can add that as a next step for improvement.

api4/schema.graphqls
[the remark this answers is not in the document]
crspeller: Our IDs are globally unique (hopefully) so we could use our `id` as the graphql `ID` except that we need the cursor to be scoped to the object type? Why does the cursor need to be scoped? Doesn't context always tell you what object type you are dealing with?

api4/resolver_channel_member_test.go
```diff
@@ -0,0 +1,355 @@
...
+	"github.com/mattermost/mattermost-server/v6/model"
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
+)
+
+func TestGraphQLChannelMembers(t *testing.T) {
+	os.Setenv("MM_FEATUREFLAGS_GRAPHQL", "true")
+	defer os.Unsetenv("MM_FEATUREFLAGS_GRAPHQL")
+	th := Setup(t).InitBasic()
+	defer th.TearDown()
```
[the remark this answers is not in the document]
agnivade: Hmm no don't think so. All we need is to login the graphQL client, and that happens in `LoginBasic()`. But I think there are a few places where we don't need users and channels, but we are doing InitBasic. I'll replace them with `loginBasicWithGraphQL` and expose the method as well.

api4/resolver_channel_member_test.go
agnivade: Hmm .. how do you mean "get these numbers upfront"? These are actually the same API tests for REST. See `TestGetChannelsForUser` in api4/channel_test.go.

This basically comes from the 2 initial channels (town-square+off-topic) and then the additional teams and channels we create in L23-28. Unless we change the initial set of channels, these shouldn't break. And even if we do change the initial channels, I'd expect some tests to break.

api4/resolver_channel_member_test.go
agnivade: That would be slightly incorrect because in GraphQL, we are specifically fetching a set of fields. So that's why the test needs to check _exactly_ whether those fields are set or not. The `NotZero`  check will only verify whether it's a zero value or not, but won't check exactly which fields are set.

api4/resolver_channel_member_test.go
isacikgoz: It was actually meant that maybe we should validating these against the REST API. The flow would be like: getChannelCount (1) -> do the test (2) -> compare with step 1 (3). But if that is a common pattern in our codebase, you can ignore my comment 😅

api4/schema.graphqls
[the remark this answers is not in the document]
agnivade: > Our IDs are globally unique (hopefully)

There's no guarantee of that :stuck_out_tongue: 

> Why does the cursor need to be scoped? Doesn't context always tell you what object type you are dealing with?

So if an id is global, what a client can essentially do is ask a server to give an object for an id. In that case, the server won't know which table to pick from. That's why we need to add the object name to the id.

api4/resolver_channel_member_test.go
agnivade: Ahh I see. Well, then the question is who guarantees the REST API is correct? Because the exact same test is for the REST API as well :stuck_out_tongue: If you are relying on the REST API being correct, then that shouldn't make much difference because both use the same test.

api4/schema.graphqls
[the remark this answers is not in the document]
crspeller: Ah so graphql clients can ask for objects just by global id. That makes sense. 
Thanks!

api4/resolver_channel_member_test.go
isacikgoz: Ehe, we should rely on REST API at some point 😅 But it's not a big deal anyway.

isacikgoz (approved):
Good stuff @agnivade and nicely done. Let's go!

PR #19494, opened 2022-02-03
Skip exporting channels for deleted teams

gabrieljackson:
This corrects a bug where exports that contained deleted teams
could fail to import. Channels must be skipped that belong to these
deleted teams in order to produce a valid import bundle.

Fixes https://mattermost.atlassian.net/browse/MM-41564

```release-note
Export deleted teams to allow for valid imports
```

### DIFF
--- a/app/export.go
+++ b/app/export.go
@@ -83,17 +83,18 @@ func (a *App) BulkExport(writer io.Writer, outPath string, opts BulkExportOpts) 
 	}
 
 	mlog.Info("Bulk export: exporting teams")
-	if err := a.exportAllTeams(writer); err != nil {
+	teamNames, err := a.exportAllTeams(writer)
+	if err != nil {
 		return err
 	}
 
 	mlog.Info("Bulk export: exporting channels")
-	if err := a.exportAllChannels(writer); err != nil {
+	if err = a.exportAllChannels(writer, teamNames); err != nil {
 		return err
 	}
 
 	mlog.Info("Bulk export: exporting users")
-	if err := a.exportAllUsers(writer); err != nil {
+	if err = a.exportAllUsers(writer); err != nil {
 		return err
 	}
 
@@ -165,12 +166,13 @@ func (a *App) exportVersion(writer io.Writer) *model.AppError {
 	return a.exportWriteLine(writer, versionLine)
 }
 
-func (a *App) exportAllTeams(writer io.Writer) *model.AppError {
+func (a *App) exportAllTeams(writer io.Writer) (map[string]bool, *model.AppError) {
 	afterId := strings.Repeat("0", 26)
+	teamNames := make(map[string]bool)
 	for {
 		teams, err := a.Srv().Store.Team().GetAllForExportAfter(1000, afterId)
 		if err != nil {
-			return model.NewAppError("exportAllTeams", "app.team.get_all.app_error", nil, err.Error(), http.StatusInternalServerError)
+			return nil, model.NewAppError("exportAllTeams", "app.team.get_all.app_error", nil, err.Error(), http.StatusInternalServerError)
 		}
 
 		if len(teams) == 0 {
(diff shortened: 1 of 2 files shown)


agnivade:
> These team resources are needed as there are dependent resources
such as channels that are still included in the export.

Can you show us where is it? From what I am seeing, we are also ignoring soft-deleted channels:

```go
// Skip deleted.
if channel.DeleteAt != 0 {
	continue
}
```

And in a lot of other places we are doing this. Ignoring `DeleteAt` only for teams will make things inconsistent. If there are other resources from the deleted team still included in the export, we should rather not include them as well instead of including the team.

gabrieljackson:
@agnivade thanks for your thoughts. I actually realized that my initial change would introduce a new issue anyway.

I have modified the PR to use a new in-memory team validation when exporting channels. It's hard for me to know for sure that there are no new edge case issues with this, but my testing looked positive. Let me know what you think.

app/export.go
```diff
@@ -83,17 +83,18 @@ func (a *App) BulkExport(writer io.Writer, outPath string, opts BulkExportOpts)
...
 	mlog.Info("Bulk export: exporting teams")
-	if err := a.exportAllTeams(writer); err != nil {
+	teamNames, err := a.exportAllTeams(writer)
+	if err != nil {
 		return err
 	}
 
 	mlog.Info("Bulk export: exporting channels")
-	if err := a.exportAllChannels(writer); err != nil {
+	if err = a.exportAllChannels(writer, teamNames); err != nil {
```
agnivade, line 92: It's good to not shadow the error again. So we can bring back the `:=` unless you specifically didn't want to?

app/export.go
```diff
@@ -83,17 +83,18 @@ func (a *App) BulkExport(writer io.Writer, outPath string, opts BulkExportOpts)
...
 		return err
 	}
 
 	mlog.Info("Bulk export: exporting users")
-	if err := a.exportAllUsers(writer); err != nil {
+	if err = a.exportAllUsers(writer); err != nil {
```
agnivade, line 97: same here

app/export_test.go
```diff
@@ -703,3 +703,40 @@ func TestBuildPostReplies(t *testing.T) {
...
+	defer th2.TearDown()
+	err, i := th2.App.BulkImport(th2.Context, &b, nil, false, 5)
+	assert.Nil(t, err)
+	assert.Equal(t, 0, i)
+
+	teams1, err := th1.App.GetAllTeams()
+	assert.Nil(t, err)
+	teams2, err := th2.App.GetAllTeams()
+	assert.Nil(t, err)
+	assert.Equal(t, len(teams1), len(teams2))
```
agnivade, line 733: Maybe just do a quick spot check to see that `team1.Id` is not there in the response?

agnivade (approved):
Looks great! Left some non-blocking comments.

app/export.go
gabrieljackson: Without this change CI/CD was complaining. Isn't this change good so we are not shadowing the error?

agnivade: The fix is to rename `err` to `err2` so as to not shadow.

app/export.go
```diff
@@ -215,6 +218,10 @@ func (a *App) exportAllChannels(writer io.Writer) *model.AppError {
 			if channel.DeleteAt != 0 {
 				continue
 			}
+			// Skip channels on deleted teams.
+			if ok := teamNames[channel.TeamName]; !ok {
```
streamer45, line 222: Is this going to work properly with DMs/GMs (as they are teamless)?

gabrieljackson: Hmm, DMs are fine as those are handled separately. I am checking into GMs though as I am not sure how those are handled.

gabrieljackson: Yep, I confirmed that group channels are also handled separately @streamer45

streamer45 (approved):
Looks good :+1:

app/export.go
gabrieljackson: Ah I see what you are proposing. I will keep this method for now as it is already used later on line 114:

```
	if err = a.exportAllDirectChannels(writer); err != nil {
```

PR #19495, opened 2022-02-03
[MM-41408] Remove autotour feature flag

JulienTant:
#### Summary
The result of the A/B testing shows that we have a better conversion rate without the autotour. This PR drops the Autotour feature flag.

QA will be done on the webapp ticket

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-41408

#### Related Pull Requests
- Has webapp changes https://github.com/mattermost/mattermost-webapp/pull/9752

#### Release Note
```release-note
NONE
```

### DIFF
--- a/model/feature_flags.go
+++ b/model/feature_flags.go
@@ -47,9 +47,6 @@ type FeatureFlags struct {
 	// Enable Calls plugin support in the mobile app
 	CallsMobile bool
 
-	// Start A/B tour tips automatically, possible values = ("none", "auto")
-	AutoTour string
-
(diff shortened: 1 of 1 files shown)


neallred (approved):

pablovelezvidal (approved):
thanks @JulienTant

neallred:
@JulienTant I think we still need this code until we decide what should be done with autotour from the frontend side. Removing this has the effect of turning the feature flag off for all instances that pull in this code because this feature flag will be missing from the list returned to the front end.

So I think we should do another PR that reverts the commit of this to master.

PR #19498, opened 2022-02-04
MM-41085: Use a counting semaphore in sendNotifications

agnivade:
We throttle the concurrency limit in maintaining thread membership
for a given thread using a counting semaphore. The limit is currently
8 which is a decent number to start with.

On some more thinking, it would be even better if the sql query
could be modified to support batch updates.

https://mattermost.atlassian.net/browse/MM-41085

```release-note
NONE
```

### DIFF
--- a/app/notification.go
+++ b/app/notification.go
@@ -219,11 +219,22 @@ func (a *App) SendNotifications(post *model.Post, team *model.Team, channel *mod
 		for id := range mentions.Mentions {
 			threadParticipants[id] = true
 		}
+
+		// sema is a counting semaphore to throttle the number of concurrent DB requests.
+		// A concurrency of 8 should be sufficient.
+		// We don't want to set a higher limit which can bring down the DB.
+		sema := make(chan struct{}, 8)
 		// for each mention, make sure to update thread autofollow (if enabled) and update increment mention count
 		for id := range threadParticipants {
 			mac := make(chan *model.AppError, 1)
+			// Get token.
+			sema <- struct{}{}
 			go func(userID string) {
-				defer close(mac)
+				defer func() {
+					close(mac)
+					// Release token.
+					<-sema
+				}()
 				mentionType, incrementMentions := mentions.Mentions[userID]
 				// if the user was not explicitly mentioned, check if they explicitly unfollowed the thread
 				if !incrementMentions {


agnivade:
This is a bit hard to write a unit test for, since we have to verify the number of goroutines currently in-flight and that's a hazy number. But I have tested manually by adding sleep and verified that it works. Please let me know if there's a way to write a test, and I can do that :)

koox00 (approved):
awesome 💯 ! (didn't know about semaphore)

koox00:
Don't know how we can unit test it, but we could test the results in an api test

agnivade:
> Don't know how we can unit test it, but we could test the results in an api test

I mean that would verify the correctness (which is not part of this PR), but wouldn't test the semaphore concurrency. There lies the challenge.

jwilander (approved):
LGTM

ashishbhate (approved):
LGTM. 

Do we know if this change leads to an increase in API response times, especially for a mention-heavy posts?

agnivade:
Yes it should. Although it's a bit hard to quantify by how much. Let me see if I can create some artificial data and run some numbers.

agnivade:
Ok, so in my highly unscientific test, I did not notice any performance drop. Actually, all of the goroutines return their errors in a channel, and then later we range from a slice of `mentionAutofollowChans` to check if any of them had an error. Therefore, essentially, we start all of the goroutines and wait for them to finish. This means we are overloading the DB at once, and waiting for the responses to come. With the semaphore approach, we are throttling the requests at a set concurrency of 8. This does not overload the DB.

I wrote this test to do a rough calculation of how much time it takes:
```go
func TestSendNotificationsSemaphore(t *testing.T) {
	th := Setup(t).InitBasic()
	defer th.TearDown()

	th.App.UpdateConfig(func(cfg *model.Config) {
		*cfg.ServiceSettings.ThreadAutoFollow = true
		*cfg.ServiceSettings.CollapsedThreads = model.CollapsedThreadsDefaultOn
		*cfg.TeamSettings.MaxUsersPerTeam = 500
		*cfg.EmailSettings.SendEmailNotifications = false
	})

	rootPost, appErr1 := th.App.CreatePostMissingChannel(th.Context, &model.Post{
		UserId:    th.BasicUser.Id,
		ChannelId: th.BasicChannel.Id,
		Message:   "rootPost",
	}, true)
	require.Nil(t, appErr1)

	users := []*model.User{}
	for i := 0; i < 100; i++ {
		user := th.CreateUser()
		th.LinkUserToTeam(user, th.BasicTeam)
		th.App.AddUserToChannel(user, th.BasicChannel, false)

		post := &model.Post{
			UserId:    user.Id,
			ChannelId: th.BasicChannel.Id,
			Message:   "hello from " + user.Id,
			RootId:    rootPost.Id,
		}
		_, appErr := th.App.CreatePost(th.Context, post, th.BasicChannel, false, false)
		require.Nil(t, appErr)
		users = append(users, user)
	}

	now := time.Now()
	post := &model.Post{
		UserId:    th.BasicUser.Id,
		ChannelId: th.BasicChannel.Id,
		Message:   "@channel how " + th.BasicUser.Id,
		RootId:    rootPost.Id,
	}
	_, appErr := th.App.CreatePost(th.Context, post, th.BasicChannel, false, false)
	t.Log("time taken: ", time.Since(now))
	require.Nil(t, appErr)
}

```

And both in master and this branch, it takes roughly 800ms.

Here are my results: 
Optimized: 
- 761ms
- 829ms

Master
- 830ms

So it seems like a good improvement without any latency hit.

agnivade:
@saturninoabril - Need a quick check on https://automation-dashboard.vercel.app/cycles/b9c892e0-1dcd-4dd8-b681-7f8a00d3f76a :slightly_smiling_face: Does everything look good?

saturninoabril:
Thanks @agnivade! Yup, all good, including the recent known issues with onboarding.

app/notification.go
```diff
@@ -219,11 +219,22 @@ func (a *App) SendNotifications(post *model.Post, team *model.Team, channel *mod
...
 			mac := make(chan *model.AppError, 1)
+			// Get token.
+			sema <- struct{}{}
 			go func(userID string) {
-				defer close(mac)
+				defer func() {
+					close(mac)
+					// Release token.
+					<-sema
+				}()
```
saturninoabril, line 237: Quick question: is it possible to add a test for this change, maybe unit test?

agnivade: Unfortunately, I don't think so. See the discussion above in the PR to see the challenges to test this.

saturninoabril: I see, noted and thanks!

agnivade:
Thanks @saturninoabril for the help!

PR #19499, opened 2022-02-04
[MM-40935] Add OnInstall() plugin hook

hanzei:
#### Summary
This PR implements a `OnInstall()` hook that is triggered when the plugin is installed as part of the onboarding flow. A new type `model.OnInstallEvent` is used as an argument that allows future extension.

In the future, other install method like the Marketplace should also trigger the hook.

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-40935

#### Release Note

```release-note
Added OnInstall() plugin hook
```

### DIFF
--- a/model/plugin_on_install_event.go
+++ b/model/plugin_on_install_event.go
@@ -1,0 +1,9 @@
+// Copyright (c) 2015-present Mattermost, Inc. All Rights Reserved.
+// See LICENSE.txt for license information.
+
+package model
+
+// OnInstallEvent is send to the plugin when it gets installed.
+type OnInstallEvent struct {
+	UserId string // The user who installed the plugin
+}
--- a/model/plugin_on_install_event.go
+++ b/model/plugin_on_install_event.go
@@ -3,7 +3,7 @@
 
 package model
 
-// OnInstallEvent is send to the plugin when it gets installed.
+// OnInstallEvent is sent to the plugin when it gets installed.
 type OnInstallEvent struct {
 	UserId string // The user who installed the plugin
 }
--- a/plugin/hooks.go
+++ b/plugin/hooks.go
@@ -251,10 +251,10 @@ type Hooks interface {
 	// Minimum server version: 6.4
 	RunDataRetention(nowTime, batchSize int64) (int64, error)
 
-	// OnInstall is invoked after the installation of a plugin as part of the unboarding.
+	// OnInstall is invoked after the installation of a plugin as part of the onboarding.
 	// It's called on every installation, not only once.
 	//
-	// In the future, other plugin installation method will trigger this hook, e.g. an installation via the Marketplace.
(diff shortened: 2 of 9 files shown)


hanzei:
If the approach of this PR gets approval, I'm going to create tickets for the other install methods.

plugin/hooks.go
```diff
@@ -249,4 +250,12 @@ type Hooks interface {
 	//
 	// Minimum server version: 6.4
 	RunDataRetention(nowTime, batchSize int64) (int64, error)
+
+	// OnInstall is invoked after the installation of a plugin as part of the unboarding.
```
catalintomai, line 254: unboarding - > onboarding

model/plugin_on_install_event.go
```diff
@@ -0,0 +1,9 @@
+// Copyright (c) 2015-present Mattermost, Inc. All Rights Reserved.
+// See LICENSE.txt for license information.
+
+package model
+
+// OnInstallEvent is send to the plugin when it gets installed.
+type OnInstallEvent struct {
```
catalintomai, line 7: do we expect to have more events similar to this in the future - if yes, would it make sense to have an `events` file instead?

catalintomai (approved):

model/plugin_on_install_event.go
```diff
@@ -0,0 +1,9 @@
+// Copyright (c) 2015-present Mattermost, Inc. All Rights Reserved.
+// See LICENSE.txt for license information.
+
+package model
+
+// OnInstallEvent is send to the plugin when it gets installed.
```
catalintomai, line 6: send -> sent

plugin/hooks.go
```diff
@@ -249,4 +250,12 @@ type Hooks interface {
...
+	// It's called on every installation, not only once.
+	//
+	// In the future, other plugin installation method will trigger this hook, e.g. an installation via the Marketplace.
```
catalintomai, line 257: method -> method(s)

model/plugin_on_install_event.go
hanzei: Yes, I would expect more plugin events to be added. https://github.com/mattermost/mattermost-server/blob/97ccf0bdf6f407d7308d58763f7687d3f079db67/model/plugin_cluster_event.go is the other existing one.

0/5 if it makes sense to consolidate into a single file. They are "stand alone" events and have nothing in common with each other besides that they are events.

levb (approved):
👍

DHaussermann (approved):
@hanzei This would be far more efficient to test post merge. 
For now there is not much we can do for functional testing on this.

We should test this post-merge.

model/plugin_on_install_event.go
catalintomai: the idea was to avoid creating one (few lines) file per event if we expect (many) more events in the future.

PR #19505, opened 2022-02-07
Pin GolangCI to a specific version

agnivade:
Go 1.16+ allows installing a tool directly from a specific version.
Therefore, we don't need to rely on the tool being pre-installed
and can directly install it using standard method.

This keeps the same version of the tool for a repo and limits the
binary placement to the GOBIN of the repo.

```release-note
NONE
```


agnivade:
Hopefully, this gives some respite from the endless upgrade woes of golangci.

lieut-data (approved):
:100:

PR #19507, opened 2022-02-07
MM-41499: Optimize TeamsUnread

agnivade:
The GetTeamsUnreadForUser call would be called for every team switch.
In CRT mode, it would make a separate store call for every team, which
would run the 3 aggregate SQL queries in GetThreadsForUser.

This is suboptimal because the complexity is linearly proportional
to the number of teams.

We make the following optimizations:
1. Change the query to a single one which aggregates all teams.
2. The query originally used just 2 out of the 3 queries, so one
query was fully redundant. We remove that query in the new one.
3. Further analysis was done whether it makes sense to run the 2
queries synchronously or not. The load-tests didn't show any degradation
in running them concurrently, so we keep the same behavior.
4. Lastly, we query the replica instead of master to reduce the load on the writer DB.
Since this query is run mainly on team switch, there isn't any major downsides on querying
the replica.

```release-note
NONE
```

https://mattermost.atlassian.net/browse/MM-41499

### DIFF
--- a/app/team.go
+++ b/app/team.go
@@ -1565,11 +1565,13 @@ func (a *App) GetTeamsUnreadForUser(excludeTeamId string, userID string, include
 		return tu
 	}
 
+	teamIDs := make([]string, 0, len(data))
 	for i := range data {
 		id := data[i].TeamId
 		if mu, ok := membersMap[id]; ok {
 			membersMap[id] = unreads(data[i], mu)
 		} else {
+			teamIDs = append(teamIDs, id)
 			membersMap[id] = unreads(data[i], &model.TeamUnread{
 				MsgCount:           0,
 				MentionCount:       0,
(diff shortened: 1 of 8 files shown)


agnivade:
Following are the load-test results:

Serial:
bounded-teamsunread


Concurrent:

teamsUnread_concurrent

In both cases, we see the expected reduction in calls to ThreadStore because we have fewer calls, but there isn't much different in synchronous and concurrent.

Overall, the load-tests aren't showing much changes primarily because we don't have too many teams in our seed data set, which makes the changes slightly less conspicuous. This should be much more apparent with unreads in >5 teams. And even more where `GetThreadsForUser` is taking a long time for customers.

jwilander (approved):
LGTM - nice work and write-up

ashishbhate (approved):

koox00 (approved):

agnivade:
@jgilliam17 - This PR touches the teams unread functionality in the sidebar. So some testing around unreads and mentions and verifying that the badge number in the teams sidebar is correct should suffice. Thank you :slightly_smiling_face:

agnivade:
Hi @jgilliam17 - wondering if you got a chance to look into this?

jgilliam17:
Thanks for the ping @agnivade and for running e2es, apologies for delay 
I will take a look today.

jgilliam17 (commented):
Thanks @agnivade 
Manually tested - looks good.
- Verified team unread functionality on the sidebar. Tested mentions and new & manually marked as unread posts - number of mentions/unreads as expected.
- [E2E report](https://automation-dashboard.vercel.app/cycles/c0689634-ad25-48f9-9075-c2df6441dd27) looks good, one failure may be related 
"should collapse channels that are not unread channels" failed, when testing this manually, that functionality is working as expected on the test server. Can you please take a look? Thanks

agnivade:
I took a look, and indeed seems to be working fine. It's unlikely that this is related to this PR because the test failure seems to be that the off-topic should not show up in the sidebar when collapsed, but it still does. This points to some client side issue where in the off-topic could be in focus and that's why it still show up when collapsed.

jgilliam17:
Thanks for checking @agnivade 
LGTM

jgilliam17 (approved):

PR #19516, opened 2022-02-07
Skip and warn on emoji import when name conflicts with system emoji

gabrieljackson:
This changes import behavior related to emoji imports when the name
conflicts with the name of a system emoji. Previously, the import
would fail, but now a warning is logged and the conflicting emoji
is skipped.

Fixes https://mattermost.atlassian.net/browse/MM-41613

```release-note
Skip and warn on emoji import when name conflicts with system emoji
```

### DIFF
--- a/app/import_validators.go
+++ b/app/import_validators.go
@@ -562,28 +562,24 @@ func validateDirectPostImportData(data *DirectPostImportData, maxPostSize int) *
 
 // validateEmojiImportData validates emoji data and returns if the import name
 // conflicts with a system emoji.
-func validateEmojiImportData(data *EmojiImportData) (bool, *model.AppError) {
+func validateEmojiImportData(data *EmojiImportData) *model.AppError {
 	if data == nil {
-		return false, model.NewAppError("BulkImport", "app.import.validate_emoji_import_data.empty.error", nil, "", http.StatusBadRequest)
+		return model.NewAppError("BulkImport", "app.import.validate_emoji_import_data.empty.error", nil, "", http.StatusBadRequest)
 	}
 
 	if data.Name == nil || *data.Name == "" {
-		return false, model.NewAppError("BulkImport", "app.import.validate_emoji_import_data.name_missing.error", nil, "", http.StatusBadRequest)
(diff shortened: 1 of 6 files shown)


amyblais:
@agnivade @Willyfrog This is planned for a Cloud hotfix.

crspeller (approved):
Seems fine but how does an import even have an emoji that conflicts?

gabrieljackson:
@crspeller I went back and looked and we have changed the system emoji list before, so I am guessing there may be some workspaces that had custom emojis that later became system emojis? Not exactly sure, but definitely seeing this issue in an import.

app/import_validators.go
```diff
@@ -560,24 +560,30 @@ func validateDirectPostImportData(data *DirectPostImportData, maxPostSize int) *
 	return nil
 }
 
-func validateEmojiImportData(data *EmojiImportData) *model.AppError {
+// validateEmojiImportData validates emoji data and returns if the import name
+// conflicts with a system emoji.
+func validateEmojiImportData(data *EmojiImportData) (bool, *model.AppError) {
```
agnivade, line 565: This seems to be a slightly cumbersome way to achieve it. IMO, a more idiomatic way would be to use the error value to detect a system emoji case and appropriately log it.

So essentially, instead of clubbing the system emoji check with the other conditions (` || inSystemEmoji(name)`), what we could do is split that out to a separate check, with a separate AppError ID.

```go
if inSystemEmoji(name) {
	return NewAppError("Emoji.IsValid", "some_other_name", nil, "", http.StatusBadRequest)	
}
```

And then in `importEmoji`, we could check if `err.Id == "some_other_name"` and then just skip with a warning. This way, we don't need to add an additional bool variable to differentiate between the case, and `InSystemEmoji` can remain unexported. And this also gives us the flexibility to keep adding more edge-cases by using errors, without returning more values.

Willyfrog:
@crspeller when we upgraded emojis to the v13.0 standard, we added quite a lot of emojis, so it is very likely that some had a name conflict, that already happened in community when we merged the change as well.

Willyfrog (approved):
Other than @agnivade comments, looks good

agnivade (approved):
Nicely done!

PR #19523, opened 2022-02-08
MM-41349: CRT, fix LastUpdated semantics

lieut-data:
#### Summary
This is a long-winded solution to https://mattermost.atlassian.net/browse/MM-41349. The ask in that ticket was relatively easy: don't update `LastUpdated` to the past, but to justify that decision, I needed to dig into the overall semantics of the feature, peeling away various layers of `deadcode` and otherwise trying to simplify things before I could convince myself the change was "safe."

Along the way, I took a stab at trying to document various structs and methods, breaking things up into discrete commits to make the journey of validating correctness more palatable. I'd love your eyes on this! Here's a summary of the changes:
* Remove various unused methods to help clarify semantics, namely `App.UpdateChannelLastViewedAt`, as well as `ThreadStore.Save`, `ThreadStore.SaveMultiple`, `ThreadStore.Update`, and `ThreadStore.Delete` (callers do things manually instead of using these helper store methods)
* Remove support for `followThread` from `App.MarkChannelAsUnreadFromPost` since it's always passed as `false`.
* Fix various `ThreadStore` methods to actually update `LastUpdated` when updating other columns on the thread membership.
* Drop the `updateThreads` parameter from `ChannelStore.UpdateLastViewedAtPost` and `ChannelStore.IncrementMentionCount`, after simplifying the calling code to clarify that these aren't needed.
* Finally rename `ThreadStore.UpdateUnreadsByChannel` to `ThreadStore.UpdateLastViewedByThreadIds`, unconditionally writing `LastViewed` (and now `LastUpdated`).

I strongly suggest reviewing commit by commit vs. the overall PR.

#### Ticket Link
Fixes: https://mattermost.atlassian.net/browse/MM-41349

#### Release Note
```release-note
NONE
```


store/sqlstore/thread_store.go
```diff
@@ -521,35 +477,29 @@ func (s *SqlThreadStore) MarkAllAsRead(userId, teamId string) error {
...
 }
 
+// MarkAsRead marks the given thread for the given user as unread from the given timestamp.
 func (s *SqlThreadStore) MarkAsRead(userId, threadId string, timestamp int64) error {
 	query, args, _ := s.getQueryBuilder().
 		Update("ThreadMemberships").
 		Where(sq.Eq{"UserId": userId}).
 		Where(sq.Eq{"PostId": threadId}).
 		Set("LastViewed", timestamp).
+		Set("LastUpdated", timestamp).
```
ashishbhate, line 495: Here `LastUpdated` should be set to `model.GetMillis()` instead of the `timestamp` argument that was passed in.

app/channel.go
```diff
@@ -2578,7 +2578,7 @@ func (a *App) MarkChannelAsUnreadFromPost(postID string, userID string, collapse
 		}
 	}
 
-	channelUnread, nErr := a.Srv().Store.Channel().UpdateLastViewedAtPost(post, userID, unreadMentions, unreadMentionsRoot, *a.Config().ServiceSettings.ThreadAutoFollow, true)
```
ashishbhate, line 2612: The buggy code that you removed was hiding the bug here.

 When `ThreadAutoFollow` is `true` and if the code you removed in `UpdateUnreadsByChannel` actually set the `LastViewed` instead of `LastUpdated`, then we'd see threads with replies newer than the post being marked as unread, being marked as unread. I checked the spreadsheet defining CRT behaviour and I dont think this was ever expected behaviour.

ashishbhate (commented):
Awesome work @lieut-data ! 
The additional clean-up and documentation work is really helpful!

ashishbhate (approved):

koox00 (approved):
awesome! thank you @lieut-data 💯

store/storetest/thread_store.go
```diff
@@ -693,7 +592,7 @@ func testThreadStorePermanentDeleteBatchThreadMembershipsForRetentionPolicies(t
 	// Delete team policy and thread
 	err = ss.RetentionPolicy().Delete(teamPolicy.ID)
 	require.NoError(t, err)
-	err = ss.Thread().Delete(post.Id)
+	_, err = s.GetMaster().Exec("DELETE FROM Threads WHERE PostId='" + post.Id + "'")
```
agnivade, line 595: One quick comment - let's please use `GetMasterX()` instead. :) We are slowly moving away from gorp.

lieut-data: Awesome to see us moving away from gorp! One challenge here is that the `SqlStore` interface used in the tests only defines `GetMaster`, and it looks like `GetMasterX` returns a `sqlxDBWrapper` not presently exported by the `sqlstore` package. Should we tackle this holistically in the tests?

app/channel_test.go
```diff
@@ -2418,13 +2390,13 @@ func TestMarkUnreadWithThreads(t *testing.T) {
...
 
-			_, appErr = th.App.MarkChannelAsUnreadFromPost(rootPost.Id, th.BasicUser.Id, true, true)
+			_, appErr = th.App.MarkChannelAsUnreadFromPost(rootPost.Id, th.BasicUser.Id, true)
 			require.Nil(t, appErr)
 
 			threadMembership, appErr := th.App.GetThreadMembershipForUser(th.BasicUser.Id, rootPost.Id)
 			require.Nil(t, appErr)
 			require.NotNil(t, threadMembership)
-			assert.Equal(t, int64(1), threadMembership.UnreadMentions)
+			assert.Zero(t, threadMembership.UnreadMentions)
```
lieut-data, line 2399: @ashishbhate, just want to get your eyes on this -- and the comment newly introduced at the top of this method.

In short, the coverage here was excellent -- but of an unused code path. I didn't want to refactor `MarkChannelAsUnreadFromPost` further at this time -- in which I /think/ we can rip out all the `ThreadStore.MaintainMembership` calls -- so opted to "correct" this test accordingly.

lieut-data:
@jgilliam17, wondering if you would be open to testing this PR for correct collapsed reply thread semantics?

jgilliam17:
Thanks @lieut-data I will take a look tomorrow.

store/storetest/thread_store.go
agnivade: Oh yeah, this PR https://github.com/mattermost/mattermost-server/pull/19545 hasn't gone in yet. I'll let you know if that goes in first. Otherwise, we can keep this be for now.

lieut-data: Done!

jgilliam17:
Thanks @lieut-data 
E2E report [without feature flag for CRT](https://automation-dashboard.vercel.app/cycles/ccb64300-b273-409f-afd3-955fa8ea7b91) and [with CRT flag](https://automation-dashboard.vercel.app/cycles/61581607-77bf-49a7-93b5-9a8685324a97) both look good, no PR related failures
Tested, looks good to merge.
- Verified threading behavior - as expected.
There was one issue with marking the root post with mention as unread (on RHS) - global threads item included the mention badge which is not per design - also repros on daily. I'll be filing that separately.

jgilliam17 (approved):

app/channel_test.go
ashishbhate: @lieut-data Sorry about the delay in getting around to this. 

I think the change here breaks the expected behaviour. `App.MarkChannelAsUnreadFromPost`  is called from the `rhs_root_post` and `rhs_comment` components in the webapp (amongst other places). 

Looking at our spreadsheet of expected behaviour - https://docs.google.com/spreadsheets/d/1Rp2UN-TDW75zccfFWrco2Pqog2ktoa784GSvAVGTfgA/edit?pli=1#gid=1722303699

The test marks the root post as unread - this is covered by lines 17 and 18, which states that a new messages line should be inserted before the oldest post.

The reply in the test has a mention, so inserting a new messages line should also increment the mention count for the user. 

@koox00 could you please confirm if this is correct? Two eyes are better than one :)

esethna:
@amyblais I discussed this with Jesse, can we cherry pick to v6.5?

app/channel_test.go
ashishbhate: @lieut-data I think this comment was missed in the review. This PR appears to break existing  expected behaviour. Jelena opened a ticket for issue, but if this PR is the cause of the regression then we ought to prioritize the fix. 

// @Willyfrog

PR #19526, opened 2022-02-09
Pin the FROM in dockerfiles

oh6hay:
This prevents supply chain attacks where the latest image is replaced with a malicious version.

Pinning done by using https://github.com/Jille/dockpin (pins to the latest hash when run)

### DIFF
--- a/build/Dockerfile
+++ b/build/Dockerfile
@@ -1,4 +1,4 @@
-FROM alpine:3.14
+FROM alpine:3.14@sha256:635f0aa53d99017b38d1a0aa5b2082f7812b03e3cdb299103fe77b5c8a07f1d2
 
 # Some ENV variables
 ENV PATH="/mattermost/bin:${PATH}"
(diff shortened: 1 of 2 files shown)


oh6hay:
Hi, sorry for a delayed response. I can't seem to access the playbook?

> @oh6hay - Can you please outline the process that's needed for us to do this? Since our team owns bumping the Go version, I'l like to document this.
> 
> We have a playbook here that we follow: https://community-daily.mattermost.com/playbooks/playbooks/k6xrudry9fgkxk3mktffyhiega/preview.
> 
> Please let us know the steps that need to be updated. Thanks!

agnivade:
Added you now, can you check?

oh6hay:
> Added you now, can you check?

I can, thanks! In this case: You can simply remove the hash from the Dockerfile, then carry on with tasks from the playbook and pin the base image afterwards. (Pinning only protects against a tagged docker image being replaced with a malicious one between runs that pull a tagged image.)

agnivade (approved):
Looks great @oh6hay ! Please also pin the versions in https://git.internal.mattermost.com/mattermost/ci/images/mattermost-build-server/-/blob/master/Dockerfile to keep things consistent. Thanks!

PR #19536, opened 2022-02-11
MM-38419: Split UPSERT statement into 2

agnivade:
The UPSERT statement was taking unexplained gap
locks, due to which the import was somehow failing.
I still don't know exactly how this is happening.

But splitting the UPSERT into 2 different statements
in the same transaction changes the locking semantics,
and avoids the deadlock.

Naturally, this proves that even our retrylayer
is sometimes ineffective against some problems.

```release-note
NONE
```


agnivade:
@streamer45 - Would you mind doing a peer testing just to confirm that the bug has been fixed? I'd prefer a second set of eyes on this.

streamer45 (approved):
Great 🙌 

Tested locally, was able to reproduce the issue on master. Using this branch the import proceeds without issues.

PR #19545, opened 2022-02-14
Replace remaining GetMaster() queries

agnivade:
There are still some remaining which
access the raw DB struct. There will be
a separate PR to clean those up.

```release-note
NONE
```


isacikgoz (approved):
Awesome

PR #19547, opened 2022-02-14
MM-41796: Sentry crash: use the correct error variable

agnivade:
We use the correct error variable while checking
for the error

https://mattermost.atlassian.net/browse/MM-41796

```release-note
NONE
```


streamer45 (approved):
:+1: I wonder how many more are there, hiding around? :grimacing:

agnivade:
Yeah, we plan to write a vet check to catch these.

isacikgoz (approved):
Nice one!

PR #19560, opened 2022-02-15
MM-41600

koox00:
#### Summary

Adds a new prop on the `ThreadUpdated` websocket event called `new_thread` which depending if the thread membership is new or not it's set as `true` or `false`.

#### Ticket Link

https://mattermost.atlassian.net/browse/MM-41600

#### Release Note
```release-note
NONE
```


app/notification.go
```diff
@@ -281,6 +283,7 @@ func (a *App) SendNotifications(post *model.Post, team *model.Team, channel *mod
 
 				membershipsMutex.Lock()
 				participantMemberships[userID] = threadMembership
+				newParticipants[userID] = newMembership
 				membershipsMutex.Unlock()
```
koox00, line 287: can I do that?

store/sqlstore/thread_store.go
```diff
@@ -702,22 +702,23 @@ func (s *SqlThreadStore) MaintainMembership(userId, postId string, opts store.Th
...
 			}
 		}
 
 		if err = trx.Commit(); err != nil {
-			return nil, errors.Wrap(err, "commit_transaction")
+			return nil, false, errors.Wrap(err, "commit_transaction")
 		}
 
-		return membership, err
+		return membership, newMembership, err
```
ashishbhate, line 713: The reason this works is because of named return parameters. `newMembership` is defined in the return parameters. 
I feel this is confusing, we should just explicitly return `true` or `false`, same as in all the other returns in the function.

app/notification.go
ashishbhate: Do you mean "reusing" the same mutex lock? Yep!

ashishbhate (commented):
Looks good to me. Why are we re-using the thread updated event instead of sending a separate "new thread" event?

PR #19562, opened 2022-02-15
MM-41834 Add EnableClientPerformanceDebugging setting

hmhealey:
This setting is going to enable some options for users on the web app to enable and disable different parts of the web app to help us isolate performance issues while debugging. The specific settings will all be user-specific, but having those hidden behind a server setting should hopefully prevent people from finding them accidentally.

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-41835

#### Release Note
```release-note
Add EnableClientPerformanceDebugging setting
```


agnivade:
Could we make this as a flag under `ServiceSettings.DeveloperFlags`? I think it's a good use for that setting since we especially designed that to be a series of key=value pairs.

devinbinnie (approved):

cwarnermm:
Documented via https://github.com/mattermost/docs/pull/5438

PR #19566, opened 2022-02-15
MM-40302: CRT, fix performance of MarkAllAsReadInChannels

lieut-data:
#### Summary
`MarkAllAsReadInChannels` was the subject of a significant performance regression in v5.37 and is known to be very inefficient, by virtue of always writing to an ever increasing number of rows, and doing so on common events like simply viewing a channel.

Fortunately, `ChannelStore.UpdateLastViewedAt` already had the kernel of an improved algorithm: query the set of threads with newer posts, and then update only /those/. Missing was the need to reset the `UnreadMentions`, but thanks to the previous simplifications in #19523, we can make this change largely without impacting other semantics.

I've written up a more detailed performance analysis over at https://docs.google.com/document/d/1N01s486pWT2rKu4VaRaRZSI_40nWkHu-wmpOSIcJvxk/edit#. The following three graphs highlight performance at load from an implementation running `master` (mid-February, v6.4), `solution` (this PR), and `baseline` (this PR but with `ServiceSettings.ThreadAutoFollow` set to `false`):

CleanShot 2022-02-15 at 18 15 19@2x
CleanShot 2022-02-15 at 18 13 41@2x
CleanShot 2022-02-15 at 18 15 34@2x

Comparing the results:

| | master | solution | baseline |
| --- | -- | -- | -- |
| p50 AVG | 812ms | 423ms | 343ms |
| p99 AVG | 7.75s | 7.14s | 4.44s |

Enabling `ServiceSettings.ThreadAutoFollow` still has a non-trivial impact on the system, but `viewChannel` average performance is much closer to the feature being disabled. These performance improvements are expected to be even better on larger installations.

One specific thing to call out: my original implementation of `MarkAllAsReadByChannels` used an `UPDATE ... FROM` syntax:

```sql
UPDATE ThreadMemberships
SET LastViewed = NOW(), UnreadMentions = 0, LastUpdated = NOW()
FROM Threads
WHERE ThreadMemberships.UserId = :userId
AND Threads.PostId = ThreadMemberships.PostId
AND Threads.ChannelID IN (:channelIds)
AND Threads.LastReplyAt > ThreadMemberships.LastViewed
```

but with worse performance characteristics at the same stable load:
CleanShot 2022-02-15 at 18 09 50@2x

The solution I landed on was to instead perform a subquery:
```sql
UPDATE ThreadMemberships
SET LastViewed = NOW(), UnreadMentions = 0, LastUpdated = NOW()
WHERE UserId = :userId
AND PostId IN (
	SELECT Threads.PostId
	FROM Threads
	JOIN ThreadMemberships ON ThreadMemberships.PostId = Threads.PostId
	WHERE Threads.ChannelId IN (:channelIds)
	AND ThreadMemberships.UserId = :userId
	AND Threads.LastReplyAt > ThreadMemberships.LastViewed
)
```

I chalk this up to the fact that the expected set of threads (on average) is very small, so running a query and indexing purely by primary key is likely to be faster than the database's attempts to optimize the simpler query's direct `JOIN`s.

#### Ticket Link
Fixes: https://mattermost.atlassian.net/browse/MM-40302

#### Release Note
```release-note
Improve performance of CRT when backend enabled, but frontend disabled.
```


lieut-data:
In light of the failure with MySQL, I'm thinking it would be valuable to repeat these loadtests (or something comparable) with MySQL as well. I won't be able to get to that for at least two weeks.

agnivade:
@lieut-data - Could you mention the query plans for before/after for both MySQL and Postgres? I am interested to see the theoretical improvements.

store/sqlstore/thread_store.go
```diff
@@ -419,44 +419,68 @@ func (s *SqlThreadStore) GetThreadForUser(teamId string, threadMembership *model
...
 	}
 
-	timestamp := model.GetMillis()
-	query, args, _ = s.getQueryBuilder().
+	query, args, _ := s.getQueryBuilder().
 		Update("ThreadMemberships").
-		Where(sq.Eq{"PostId": threadIDs}).
 		Where(sq.Eq{"UserId": userID}).
+		Where(sq.Expr("PostId IN ("+subquery+")", subqueryArgs...)).
+		Set("LastViewed", now).
```
ashishbhate, line 453: Should we set this to `Threads.LastReplyAt` ? `Now` could be a lot after `Threads.LastReplyAt`. If there are network/lag issues (mobile client), could we conceivably be marking a thread as read for posts that are still in-flight? 

This is probably not an issue for a user with CRT turned off. Users with two clients, one of which has CRT and the other is unsupported, could see these still in flight posts as read after switching from the CRT off to the CRT on client. 

Analogous to why the the channelMembership.LastViewedAt isn't set to the anything greater than the CreateAt of the latest post. 

See - channel_store:UpdateLastViewedAt and https://community-daily.mattermost.com/core/pl/ff87f31mfb85pfipnw1sqgx81e

lieut-data:
> @lieut-data - Could you mention the query plans for before/after for both MySQL and Postgres? I am interested to see the theoretical improvements.

Absolutely, @agnivade! I have "real" queries for Postgres, but replicated the same locally for MYSQL.

Here's the Postgres `UPDATE ... FROM`:
```
EXPLAIN UPDATE ThreadMemberships
SET LastViewed = 500, UnreadMentions = 0, LastUpdated = 500
FROM Threads
WHERE ThreadMemberships.UserId = 'ra57yff7u7y49ra34cftj3hp7a'
AND Threads.PostId = ThreadMemberships.PostId
AND Threads.ChannelID = 'fi3cjy33rff5jnmjq3gfxsz9ge'
AND Threads.LastReplyAt > ThreadMemberships.LastViewed;
                                                            QUERY PLAN
-----------------------------------------------------------------------------------------------------------------------------------
 Update on threadmemberships  (cost=14.90..3166.16 rows=1 width=91)
   ->  Nested Loop  (cost=14.90..3166.16 rows=1 width=91)
         ->  Bitmap Heap Scan on threadmemberships  (cost=14.47..973.38 rows=264 width=69)
               Recheck Cond: ((userid)::text = 'ra57yff7u7y49ra34cftj3hp7a'::text)
               ->  Bitmap Index Scan on idx_thread_memberships_user_id  (cost=0.00..14.41 rows=264 width=0)
                     Index Cond: ((userid)::text = 'ra57yff7u7y49ra34cftj3hp7a'::text)
         ->  Index Scan using threads_pkey on threads  (cost=0.42..8.30 rows=1 width=41)
               Index Cond: ((postid)::text = (threadmemberships.postid)::text)
               Filter: (((channelid)::text = 'fi3cjy33rff5jnmjq3gfxsz9ge'::text) AND (lastreplyat > threadmemberships.lastviewed))
(9 rows)
```

and Postgres `UPDATE` via a subquery:
```
EXPLAIN UPDATE ThreadMemberships
SET LastViewed = 500, UnreadMentions = 0, LastUpdated = 500
WHERE UserId = 'ra57yff7u7y49ra34cftj3hp7a'
AND PostId IN (
    SELECT Threads.PostId
    FROM Threads
    JOIN ThreadMemberships ON ThreadMemberships.PostId = Threads.PostId
    WHERE Threads.ChannelID IN ('fi3cjy33rff5jnmjq3gfxsz9ge')
    AND ThreadMemberships.UserId = 'ra57yff7u7y49ra34cftj3hp7a'
    AND Threads.LastReplyAt > ThreadMemberships.LastViewed
);
                                                                   QUERY PLAN
-------------------------------------------------------------------------------------------------------------------------------------------------
 Update on threadmemberships  (cost=3166.72..3174.03 rows=1 width=97)
   ->  Nested Loop  (cost=3166.72..3174.03 rows=1 width=97)
         Join Filter: ((threadmemberships.postid)::text = (threadmemberships_1.postid)::text)
         ->  HashAggregate  (cost=3166.17..3166.18 rows=1 width=66)
               Group Key: (threads.postid)::text
               ->  Nested Loop  (cost=14.90..3166.16 rows=1 width=66)
                     ->  Bitmap Heap Scan on threadmemberships threadmemberships_1  (cost=14.47..973.38 rows=264 width=41)
                           Recheck Cond: ((userid)::text = 'ra57yff7u7y49ra34cftj3hp7a'::text)
                           ->  Bitmap Index Scan on idx_thread_memberships_user_id  (cost=0.00..14.41 rows=264 width=0)
                                 Index Cond: ((userid)::text = 'ra57yff7u7y49ra34cftj3hp7a'::text)
                     ->  Index Scan using threads_pkey on threads  (cost=0.42..8.30 rows=1 width=41)
                           Index Cond: ((postid)::text = (threadmemberships_1.postid)::text)
                           Filter: (((channelid)::text = 'fi3cjy33rff5jnmjq3gfxsz9ge'::text) AND (lastreplyat > threadmemberships_1.lastviewed))
         ->  Index Scan using threadmemberships_pkey on threadmemberships  (cost=0.55..7.84 rows=1 width=61)
               Index Cond: (((postid)::text = (threads.postid)::text) AND ((userid)::text = 'ra57yff7u7y49ra34cftj3hp7a'::text))
(15 rows)
```

I don't have a MySQL loadtesting instance setup at present, but running the equivalent queries locally, for `UPDATE ... FROM`:
```
EXPLAIN UPDATE ThreadMemberships, Threads
SET ThreadMemberships.LastViewed = 500, ThreadMemberships.UnreadMentions = 0, ThreadMemberships.LastUpdated = 500
WHERE ThreadMemberships.UserId = 'ra57yff7u7y49ra34cftj3hp7a'
AND Threads.PostId = ThreadMemberships.PostId
AND Threads.ChannelID IN ('fi3cjy33rff5jnmjq3gfxsz9ge')
AND Threads.LastReplyAt > ThreadMemberships.LastViewed;
           id: 1
  select_type: UPDATE
        table: ThreadMemberships
   partitions: NULL
         type: ref
possible_keys: PRIMARY,idx_thread_memberships_last_view_at,idx_thread_memberships_user_id
          key: idx_thread_memberships_user_id
      key_len: 106
          ref: const
         rows: 1
     filtered: 100.00
        Extra: NULL
*************************** 2. row ***************************
           id: 1
  select_type: SIMPLE
        table: Threads
   partitions: NULL
         type: eq_ref
possible_keys: PRIMARY,idx_threads_channel_id_last_reply_at
          key: PRIMARY
      key_len: 106
          ref: mattermost_test.ThreadMemberships.PostId
         rows: 1
     filtered: 100.00
        Extra: Using where
2 rows in set (0.00 sec)
```

and `UPDATE` with a subquery:
```
EXPLAIN UPDATE ThreadMemberships
SET LastViewed = 500, UnreadMentions = 0, LastUpdated = 500
WHERE UserId = 'ra57yff7u7y49ra34cftj3hp7a'
AND PostId IN (SELECT PostId FROM (
    SELECT Threads.PostId
    FROM Threads
    JOIN ThreadMemberships ON ThreadMemberships.PostId = Threads.PostId
    WHERE Threads.ChannelID IN ('fi3cjy33rff5jnmjq3gfxsz9ge')
    AND ThreadMemberships.UserId = 'ra57yff7u7y49ra34cftj3hp7a'
    AND Threads.LastReplyAt > ThreadMemberships.LastViewed
) as t );

           id: 1
  select_type: UPDATE
        table: ThreadMemberships
   partitions: NULL
         type: range
possible_keys: idx_thread_memberships_user_id
          key: idx_thread_memberships_user_id
      key_len: 106
          ref: const
         rows: 1
     filtered: 100.00
        Extra: Using where
*************************** 2. row ***************************
           id: 2
  select_type: DEPENDENT SUBQUERY
        table: <derived3>
   partitions: NULL
         type: index_subquery
possible_keys: <auto_key0>
          key: <auto_key0>
      key_len: 106
          ref: func
         rows: 2
     filtered: 100.00
        Extra: Using index
*************************** 3. row ***************************
           id: 3
  select_type: DERIVED
        table: Threads
   partitions: NULL
         type: ref
possible_keys: PRIMARY,idx_threads_channel_id_last_reply_at
          key: idx_threads_channel_id_last_reply_at
      key_len: 107
          ref: const
         rows: 1
     filtered: 100.00
        Extra: Using index
*************************** 4. row ***************************
           id: 3
  select_type: DERIVED
        table: ThreadMemberships
   partitions: NULL
         type: eq_ref
possible_keys: PRIMARY,idx_thread_memberships_last_view_at,idx_thread_memberships_user_id
          key: PRIMARY
      key_len: 212
          ref: mattermost_test.Threads.PostId,const
         rows: 1
     filtered: 100.00
        Extra: Using where
4 rows in set (0.00 sec)
```

store/sqlstore/thread_store.go
lieut-data: Hmm, this would be an option, but I expect it to cost us significantly more -- likely need a CTE a pair of queries like we do for the MySQL case.

How often do we expect this to happen? Are there still official clients that don't support CRT, or was this mostly historical? Are there third-party clients (like Matterhorn) that we're considering here?

I'd be inclined to keep this query as simple as possible if we want to preserve the observed performance gains, and wonder if we might be willing to sacrifice the risk of an in-flight message in such cases.

agnivade:
Ok, just focussing on Postgres for now, the `UPDATE .. FROM` looks much more optimized than the `UPDATE .. WHERE` query. Were you testing with Postgres or MySQL when you noticed worse results with it? From your graphs it looks like an improvement with p50, and near about the same times with p99.

I think we need to dive a bit deep into this and understand why we cannot use `UPDATE .. FROM`, because clearly there's a 2 order of magnitude cost difference in the queries:
` Update on threadmemberships  (cost=14.90..3166.16 rows=1 width=91)` compared to ` Update on threadmemberships  (cost=3166.72..3174.03 rows=1 width=97)`.

Can you also run `EXPLAIN ANALYZE VERBOSE` to get the actual run times and compare them? I'd be curious to see another load-test (since you already have the setup for Postgres) using `UPDATE .. FROM` and just focus on the Grafana chart for the store call and see the timings for that.

ashishbhate:
@agnivade 

>I think we need to dive a bit deep into this and understand why we cannot use UPDATE .. FROM, because clearly there's a 2 order of magnitude cost difference in the queries:
Update on threadmemberships (cost=14.90..3166.16 rows=1 width=91) compared to Update on threadmemberships (cost=3166.72..3174.03 rows=1 width=97).


I'm trying to make sense of both the query plans. I'm not sure I understand how there's the 2 order of magnitude cost difference. 

The "total cost"  is nearly the same (3166.16 vs 3174.03). 

The major difference is in the "startup cost" (14.90 vs 3166.72) - which IIUC is the cost before that node outputs its first row, which only matters significantly in queries that include statements like `EXISTS`, so that the engine can end execution as soon as it gets the first row.

The "total cost" (the cost before the node finishes writing its output) of the both root nodes is nearly the same, which seems to explain why the graphs show very similar p50 and p99 information?

store/sqlstore/thread_store.go
ashishbhate: The oldest supported server version is the current ESR 5.37. The oldest compatible mobile client is v1.47. CRT was added to mobile in v1.46.
The CRT user setting is a user level setting, not a client level setting. This means the user has CRT on on all clients or no clients. Since this code will only run with CRT off, it's not going to be a problem when switching clients. 
Third party clients that don't support CRT could be a problem. 

Given all of the above, I'm ok with keeping this query simple for performance reasons.

agnivade:
Sorry you are right, I was only looking at the startup costs. My point is that Jesse was observing degraded performance with the `UPDATE .. FROM` query, which should not be the case. From the query plan, it is visible that there are 2 nested loops in the other one, and the `UPDATE .. FROM` query is much simpler with just one nested loop.

So practically both should be the same in the worst case. But Jesse was observing degraded performance with the better one. That's what I wanted to understand.

If both perform same, I'd prefer to use `UPDATE .. FROM` to be as optimal as possible.

lieut-data:
> Were you testing with Postgres or MySQL when you noticed worse results with it? From your graphs it looks like an improvement with p50, and near about the same times with p99.

I was testing with Postgres. I'll looking into recreating these comparisons, including one for MySQL as well.

lieut-data:
@agnivade, thanks for the prompt to re-evaluate the performance of the two approaches discussed above. When I was originally testing the `UPDATE FROM` variant, my methodology was to stop the running server, restart with the updated binary, and run for an hour. I then repeated that with the original subquery approach (lest the cache reset should disadvantage one over the other).

This time, I opted to spin up 4x load test instances, (Subquery, UPDATE FROM) x (MySQL, Postgres). After running for the same amount of time (about two days), they show similar performance profiles for `ThreadStore.MarkAllAsReadByChannels`, with `UPDATE FROM` being faster on average for MySQL, but about the same for Postgres:

CleanShot 2022-03-02 at 09 07 45@2x
CleanShot 2022-03-02 at 09 07 55@2x
CleanShot 2022-03-02 at 09 08 05@2x
CleanShot 2022-03-02 at 09 08 16@2x

Examining the overall `viewChannel` API performance shows similar results for Postgres, but something wildly off between the two MySQL instances:
CleanShot 2022-03-02 at 09 27 58@2x
CleanShot 2022-03-02 at 09 28 12@2x
CleanShot 2022-03-02 at 09 28 23@2x
CleanShot 2022-03-02 at 09 28 31@2x

Given that the two MySQL instances are identical except for the implementation of `ThreadStore.MarkAllAsReadByChannels` (without any observed volatility therein), I conclude there there must be some non-determinism at play in the larger load test setup. Indeed, the two MySQL instances vary significantly across total store latency:
CleanShot 2022-03-02 at 09 41 42@2x
CleanShot 2022-03-02 at 09 41 51@2x

At this point, I don't have a theory as to /why/ there is an observed difference and would appreciate any insight you (and @streamer45?) might have.

Given the comparable performance of the `UPDATE FROM`, I'm going to go ahead and push those changes at least as we continue the review of these particular changes.

store/sqlstore/thread_store.go
```diff
@@ -511,44 +511,75 @@ func (s *SqlThreadStore) GetThreadForUser(teamId string, threadMembership *model
...
+// than the subquery with the extra JOIN below. Since the query constrains itself only to threads
+// which have been updated since the last time the channel was viewed, the subquery outperforms
+// the UPDATE FROM as it forces a lookup of a small number of post ids (on average) instead of
+// using other indexes less efficiently to reach the same conclusion.
+func (s *SqlThreadStore) MarkAllAsReadByChannels(userID string, channelIDs []string) error {
+	now := model.GetMillis()
+
+	// TODO: Fork squirrel to include https://github.com/Masterminds/squirrel/pull/256 and
+	// support FROM in an UPDATE query.
+	channelIDsSql, channelIDsArgs := constructArrayArgs(channelIDs)
```
agnivade, line 528: Ahh what a shame, I experienced the same while looking for join support in the DELETE statement.

store/sqlstore/thread_store.go
```diff
@@ -511,44 +511,75 @@ func (s *SqlThreadStore) GetThreadForUser(teamId string, threadMembership *model
...
 
-// MarkAllAsReadInChannels marks all threads for the given user in the given channels as read from
-// the current time.
-func (s *SqlThreadStore) MarkAllAsReadInChannels(userID string, channelIDs []string) error {
-	threadIDs := []string{}
+// MarkAllAsReadByChannels marks thread membership for the given users in the given channels
+// as read. This is used by the application layer to keep threads up-to-date when CRT is disabled
+// for the enduser, avoiding an influx of unread threads when first turning the feature on.
+//
+// Note that an alternate implementation of this method using UPDATE FROM actually performed worse
```
agnivade, line 518: Outdated comment?

agnivade (approved):
Great stuff @lieut-data ! No further concerns from my end. The irregularity with MySQL should not be related to the changes. We should be able to have a laser focus on improving load tests when we hire somebody. Till then, I'm afraid we have to live with this.

store/sqlstore/thread_store.go
lieut-data: Good catch! Fixed :)

jgilliam17:
@lieut-data I ran both [E2Es with CRT ](https://automation-dashboard.vercel.app/cycles/e290d33e-79b6-4581-a0e8-bad278ea3cf6)and [Without CRT](https://automation-dashboard.vercel.app/cycles/cab3d261-f730-45e6-9184-cff45d17f9e7), both reports look good, no PR related failures

esethna:
@amyblais I discussed this with Jesse, can we cherry pick to v6.5 as soon as it's merged? 

@ashishbhate please help review asap

ashishbhate (approved):
Fantastic work, @lieut-data !

amyblais:
@lieut-data Seems to need a manual cherry-pick.

lieut-data:
Will do, thanks @amyblais! Note that I'd like to see this hit `community` proper for at least a day before we complete the cherry pick, but will make sure this gets done.

PR #19573, opened 2022-02-16
Change log level for missing plugin

agnivade:
As discussed offline, this is not an actionable error message
and should therefore be demoted to debug.

```release-note
NONE
```


DSchalla (approved):
Thanks Agniva!

spirosoik (approved):
💯

PR #19575, opened 2022-02-16
Pin Docker base images in .circleci/config.yml

oh6hay:
Pin the Docker base images to their latest hashes to make supply chain attacks including malicious docker images less likely.


oh6hay:
I accidentally opened this PR on an outdated master -- closed and reopened against current master

PR #19577, opened 2022-02-16
Harden CodeQL analysis workflow

oh6hay:
Harden CodeQL analysis workflow
- declare token permissions
- dependency-pin actions

### DIFF
--- a/.github/workflows/codeql-analysis.yml
+++ b/.github/workflows/codeql-analysis.yml
@@ -9,8 +9,15 @@ on:
   schedule:
     - cron: '0 0 * * 0'
 
+permissions:
+  contents: read
+
 jobs:
   analyze:
+    permissions:
+      actions: read # for github/codeql-action/init to get workflow details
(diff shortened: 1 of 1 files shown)


.github/workflows/codeql-analysis.yml
```diff
@@ -21,12 +28,12 @@ jobs:
 
     steps:
     - name: Checkout repository
-      uses: actions/checkout@v2
+      uses: actions/checkout@629c2de402a417ea7690ca6ce3f33229e27606a5 # v2
```
srkgupta, line 31: @oh6hay hardcoding a specific commit means it will not pull the latest CodeQL queries added. We are currently leveraging this and any new CodeQL queries committed to the CodeQL repo helps us identify issues on existing codebase. I am afraid we will lose this advantage if we specify a specific commit.

oh6hay: that's right

PR #19578, opened 2022-02-16
Add TermsOfServiceId to UserTermsOfService table primary key.

larkox:
#### Summary
TermsOfService table has as primary key the userID, which impedes storing the acceptance of different terms of services in the system. This is needed for auditing.

This PR makes the primary key compounded between userID and termsOfServiceId.

Note: In order to backport this, we would need to update the old migration system, so we cannot simply cherry pick this PR.

IMPORTANT: The downgrade path is destructive, so we should make note of that.

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-41611

#### Release Note
```release-note
Fix problem that did not allow users to accept new Custom Terms of Service after they are changed or re-enabled.
```


larkox:
@isacikgoz please add any other reviewer that should take a look at this.

db/migrations/postgres/000076_alter_terms_of_service_primary_key.up.sql
```diff
@@ -0,0 +1,26 @@
...
+    where
+        pg_class.oid = 'usertermsofservice'::regclass AND
+        indrelid = pg_class.oid AND
+        nspname = 'public' AND
+        pg_class.relnamespace = pg_namespace.oid AND
+        a.attrelid = pg_class.oid AND
+        a.attnum = ANY(ix.indkey) AND
+        indisprimary;
+
+    IF COALESCE (column_name, '') != text('userid') THEN
```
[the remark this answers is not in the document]
larkox: Regarding COALESCE, I have to admit this has been mostly copy pasted from a different place without understanding 100% what is going on, so 0/5 whether it is necessary.

Regarding the `=`, you are completely right.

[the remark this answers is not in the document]
agnivade: Let's remove `COALESCE` then. I don't believe it's necessary.

agnivade (approved):

db/migrations/mysql/000076_alter_terms_of_service_primary_key.up.sql
```diff
@@ -0,0 +1,18 @@
...
+BEGIN
+    DECLARE columnName varchar(26) default '';
+
+    SELECT IFNULL(GROUP_CONCAT(column_name ORDER BY seq_in_index), '') INTO columnName
+    FROM information_schema.statistics
+    WHERE table_schema = DATABASE()
+    AND table_name = 'UserTermsOfService'
+    AND index_name = 'PRIMARY'
+
+    IF columnName = 'UserId' THEN
```
isacikgoz, line 11: Like we do in the `down` script, can we check if the primary keys are not `(UserId, TermsOfServiceId)`?

db/migrations/postgres/000076_alter_terms_of_service_primary_key.up.sql
```diff
@@ -0,0 +1,26 @@
...
+    where
+        pg_class.oid = 'usertermsofservice'::regclass AND
+        indrelid = pg_class.oid AND
+        nspname = 'public' AND
+        pg_class.relnamespace = pg_namespace.oid AND
+        a.attrelid = pg_class.oid AND
+        a.attnum = ANY(ix.indkey) AND
+        indisprimary;
+
+    IF column_name = text('userid') THEN
```
isacikgoz, line 21: Same here

isacikgoz (commented):
Looking very good! Have a small suggestion though. Let me know what you think

db/migrations/mysql/000076_alter_terms_of_service_primary_key.up.sql
larkox: Sure. Do you know what is the expected value of columnName in that case? Is it `(UserId, TermsOfServiceId)`, or is it any other string?

PR #19594, opened 2022-02-18
[MM-41712] Allow sending intra-cluster WS events reliably

streamer45:
#### Summary

The current implementation of plugin's API `PublishWebSocketEvent()` will eventually send messages across the cluster using the unreliable (UDP backed) channel. This can lead to unexpected message loss or ordering issues. It also imposes some hard limitations on maximum payload size (~64KB).

PR adds an option to allow for broadcasting WS events through the cluster using the reliable, TCP backed channel.

I've opted to extend `model.WebsocketBroadcast` to avoid either breaking the plugin's API or adding a new almost identical method just for this case. Please let me know if it makes sense or if you prefer a new method.

#### Ticket Link

https://mattermost.atlassian.net/browse/MM-41712

#### Release Note

```release-note
Added a new ReliableClusterSend field to model.WebsocketBroadcast to allow sending events through the cluster using the reliable channel.
```

### DIFF
--- a/app/web_hub.go
+++ b/app/web_hub.go
@@ -178,7 +178,8 @@ func (s *Server) Publish(message *model.WebSocketEvent) {
 			message.EventType() == model.WebsocketEventPostEdited ||
 			message.EventType() == model.WebsocketEventDirectAdded ||
 			message.EventType() == model.WebsocketEventGroupAdded ||
-			message.EventType() == model.WebsocketEventAddedToTeam {
+			message.EventType() == model.WebsocketEventAddedToTeam ||
+			message.GetBroadcast().ReliableClusterSend {
 			cm.SendType = model.ClusterSendReliable
 		}
 
--- a/app/web_hub_test.go
+++ b/app/web_hub_test.go
@@ -19,6 +19,7 @@ import (
 	"github.com/mattermost/mattermost-server/v6/model"
 	"github.com/mattermost/mattermost-server/v6/shared/i18n"
 	"github.com/mattermost/mattermost-server/v6/store/storetest/mocks"
+	"github.com/mattermost/mattermost-server/v6/testlib"
 )
 
 func dummyWebsocketHandler(t *testing.T) http.HandlerFunc {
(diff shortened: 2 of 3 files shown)


app/web_hub.go
```diff
@@ -178,7 +178,8 @@ func (s *Server) Publish(message *model.WebSocketEvent) {
 			message.EventType() == model.WebsocketEventPostEdited ||
 			message.EventType() == model.WebsocketEventDirectAdded ||
 			message.EventType() == model.WebsocketEventGroupAdded ||
```
agnivade, line 180: Maybe we can add the broadcast type as reliable for all of these messages and then remove these from the if condition altogether? In a separate HW ticket of course.

app/web_hub.go
```diff
@@ -178,7 +178,8 @@ func (s *Server) Publish(message *model.WebSocketEvent) {
...
-			message.EventType() == model.WebsocketEventAddedToTeam {
+			message.EventType() == model.WebsocketEventAddedToTeam ||
+			message.GetBroadcast().ReliableClusterSend {
```
agnivade, line 182: Just to confirm, this shouldn't break old plugins sending messages because by default this field will be false correct?

app/web_hub.go
streamer45: That's a nice idea and it probably makes sense to let the higher layer decide how they should get sent.

agnivade (approved):

metanerd:
/e2e-test

Sorry, I am just using this PR to test something on my end.

hanzei (approved):
LGTM :+1:

streamer45:
@amyblais We'd like to have this in 6.5.0 ideally. It's been merged for a while now. Let me know if possible and I can cherry-pick to cloud.

amyblais:
> @amyblais We'd like to have this in 6.5.0 ideally. It's been merged for a while now. Let me know if possible and I can cherry-pick to cloud.

Please share more details on why it's urgent and whether it's risky. It's too late to cherry-pick for tomorrow's cloud release as QA is wrapping up testing now, so it would need to be cherry-picked to 6.5 branch separately.

streamer45:
> Please share more details on why it's urgent and whether it's risky. It's too late to cherry-pick for tomorrow's cloud release as QA is wrapping up testing now, so it would need to be cherry-picked to 6.5 branch separately.

Okay, cherry-picking to 6.5 works as well. I can wait until the branch is ready.

It's nothing extremely urgent, it would be nice to have it so we can use this new functionality in Calls starting with the next version as it helps avoiding potential intra-cluster message loss which can make calls crash or not work as expected. Risk is minimal as the changes don't impact in any way existing app functionality.

amyblais:
Thanks, I'll cherry-pick to 6.5 separately.

PR #19596, opened 2022-02-18
MM-41752: Batch update mention increment

agnivade:
Previously, we were incrementing mentions one-by-one
all concurrently in an unbounded fashion.

This would cause a big spike in memory usage if there
were an `@all` mention in a large channel.

We fix this by changing the SQL query to take all userIDs
at once.

https://mattermost.atlassian.net/browse/MM-41752

```release-note
NONE
```

### DIFF
--- a/store/storetest/user_store.go
+++ b/store/storetest/user_store.go
@@ -2332,10 +2332,7 @@ func testUserUnreadCount(t *testing.T, ss store.Store) {
 	// Post one message with mention to open channel
 	_, nErr = ss.Post().Save(&p1)
 	require.NoError(t, nErr)
-	nErr = ss.Channel().IncrementMentionCount(c1.Id, []string{u2.Id}, false)
-	require.NoError(t, nErr)
-
-	nErr = ss.Channel().IncrementMentionCount(c1.Id, []string{u3.Id}, false)
+	nErr = ss.Channel().IncrementMentionCount(c1.Id, []string{u2.Id, u3.Id}, false)
 	require.NoError(t, nErr)
 
 	// Post 2 messages without mention to direct channel
--- a/store/storetest/user_store.go
+++ b/store/storetest/user_store.go
@@ -2285,6 +2285,15 @@ func testUserUnreadCount(t *testing.T, ss store.Store) {
 	_, nErr = ss.Team().SaveMember(&model.TeamMember{TeamId: teamId, UserId: u2.Id}, -1)
 	require.NoError(t, nErr)
 
+	u3 := &model.User{}
+	u3.Email = MakeEmail()
+	u3.Username = "user3" + model.NewId()
+	_, err = ss.User().Save(u3)
+	require.NoError(t, err)
+	defer func() { require.NoError(t, ss.User().PermanentDelete(u3.Id)) }()
+	_, nErr = ss.Team().SaveMember(&model.TeamMember{TeamId: teamId, UserId: u3.Id}, -1)
+	require.NoError(t, nErr)
+
 	_, nErr = ss.Channel().Save(&c1, -1)
 	require.NoError(t, nErr, "couldn't save item")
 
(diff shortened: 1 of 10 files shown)


store/storetest/user_store.go
```diff
@@ -2336,7 +2336,7 @@ func testUserUnreadCount(t *testing.T, ss store.Store) {
 	_, nErr = ss.Post().Save(&p3)
 	require.NoError(t, nErr)
 
-	nErr = ss.Channel().IncrementMentionCount(c2.Id, u2.Id, false, false)
+	nErr = ss.Channel().IncrementMentionCount(c2.Id, []string{u2.Id}, false, false)
```
larkox, line 2339: I don't see tests with more than one id (they should work fine, but still, would be nice to have).

Also, just as a sanity check, it would be nice to have a test with the same id repeated. Looking at the code it should be solved fine, but seems a nice to have too.

larkox (approved):
LGTM, just one small proposal for tests.

store/storetest/user_store.go
agnivade: Yep, good point. I am just waiting for Jesse to return and merge the other PR on which it's blocked.

lieut-data: I've resolved the merge conflicts and should auto-merge https://github.com/mattermost/mattermost-server/pull/19523 shortly.

agnivade: Tests added!

agnivade:
@ashishbhate @koox00 - ready for review.

store/storetest/user_store.go
```diff
@@ -2310,12 +2327,15 @@ func testUserUnreadCount(t *testing.T, ss store.Store) {
...
+	p1.Message = "this is a message for @" + u2.Username + " and " + "@" + u3.Username
 
 	// Post one message with mention to open channel
 	_, nErr = ss.Post().Save(&p1)
 	require.NoError(t, nErr)
-	nErr = ss.Channel().IncrementMentionCount(c1.Id, u2.Id, false)
+	nErr = ss.Channel().IncrementMentionCount(c1.Id, []string{u2.Id}, false)
+	require.NoError(t, nErr)
+
+	nErr = ss.Channel().IncrementMentionCount(c1.Id, []string{u3.Id}, false)
```
koox00, line 2338: Wouldn't the actual scenario be to increment both at the same time?

```suggestion
	nErr = ss.Channel().IncrementMentionCount(c1.Id, []string{u2.Id, u3.Id}, false)
```

store/storetest/user_store.go
larkox: Maybe I am missing something, but I don't see how the new tests cover the two cases I noted (more than one id, and repeated id).

store/storetest/user_store.go
agnivade: Oh, brain.exe stopped working for a while there.

store/storetest/user_store.go
agnivade: PTAL :)

koox00 (approved):

lieut-data (approved):

jgilliam17 (approved):
Thanks @agnivade 
Tested, looks good to merge.
- Verified mentions are working as expected.
- [E2E report](https://mattermost-cypress-report.s3.amazonaws.com/157780-5b19d29-onprem-ent-server-pr-19596-mattermost/mm-ee-test5b19d29/mochawesome.html) looks good, no PR related failures.

agnivade:
@ashishbhate - Going to merge this since I have 4 dev approvals.

ashishbhate (commented):
LGTM! 

Sorry about the delay in getting to this, @agnivade

amyblais:
@agnivade This can be manually cherry-picked to ``release-6.3``.

agnivade:
Ahh sorry @amyblais - this PR is built on top of some refactoring work that already went in for CRT. To make this work with 6.3 would mean to apply all of those previous work as well. Let me know if you think it's important enough to do that.

amyblais:
> Ahh sorry @amyblais - this PR is built on top of some refactoring work that already went in for CRT. To make this work with 6.3 would mean to apply all of those previous work as well. Let me know if you think it's important enough to do that.

Sounds good. I think it's not needed to be done.

PR #19610, opened 2022-02-22
[Gekidou] Add push notification check to ping

larkox:
#### Summary
Add push notification check to ping.

Changes in the API require doc changes.

#### Related PRs
Mobile (Gekidou): https://github.com/mattermost/mattermost-mobile/pull/6002
Push Proxy: https://github.com/mattermost/mattermost-push-proxy/pull/96

#### Ticket Link
None

#### Release Note
```release-note
Ping endpoint now can receive a device ID, which will report whether the device is able to receive push notifications.
```


enahum (approved):

app/notification_push.go
```diff
@@ -390,6 +391,8 @@ func (a *App) sendToPushProxy(msg *model.PushNotification, session *model.Sessio
 		mlog.String("status", model.PushSendPrepare),
 	)
 
+	fmt.Printf("%s %s\n", msg.DeviceId, msg.Platform)
```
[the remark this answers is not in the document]
larkox: Forgot to remove. Was there to understand the(and test with a real) deviceID.

api4/system.go
```diff
@@ -183,6 +183,17 @@ func getSystemPing(c *Context, w http.ResponseWriter, r *http.Request) {
 		w.Header().Set(filestoreStatusKey, s[filestoreStatusKey])
 	}
 
+	if deviceID := r.FormValue("device_id"); deviceID != "" {
+		canSend, err := c.App.TestPushNotification(deviceID)
```
enahum, line 187: Forgot to change it here?

enahum (changes requested):

api4/system.go
larkox: Trusted too much in the "rename symbol".

enahum (approved):

jupenur (approved):

api4/system.go
```diff
@@ -183,6 +183,17 @@ func getSystemPing(c *Context, w http.ResponseWriter, r *http.Request) {
 		w.Header().Set(filestoreStatusKey, s[filestoreStatusKey])
 	}
 
+	if deviceID := r.FormValue("device_id"); deviceID != "" {
+		canSend, err := c.App.SendTestPushNotification(deviceID)
+		if err != nil {
+			s["CanReceiveNotifications"] = err.Error()
```
agnivade, line 189: We don't do the same for other keys in the map. To remain consistent, I'd suggest to log a warning and return false instead.

app/notification_push.go
```diff
@@ -568,6 +568,47 @@ func (a *App) BuildPushNotificationMessage(contentsConfig string, post *model.Po
 	return msg, nil
 }
 
+func (a *App) SendTestPushNotification(deviceID string) (canSend bool, err error) {
+	var msg *model.PushNotification = &model.PushNotification{
```
agnivade, line 572: Nit: the more idiomatic way is `msg := ...`

app/notification_push.go
```diff
@@ -568,6 +568,47 @@ func (a *App) BuildPushNotificationMessage(contentsConfig string, post *model.Po
...
+	if jsonErr := json.NewDecoder(resp.Body).Decode(&pushResponse); jsonErr != nil {
+		return false, errors.Wrap(jsonErr, "failed to decode from JSON")
+	}
+
+	switch pushResponse[model.PushStatus] {
+	case model.PushStatusRemove:
+		return false, nil
+	case model.PushStatusFail:
+		return false, errors.New(pushResponse[model.PushStatusErrorMsg])
+	}
```
agnivade, line 607: This entire part is mostly copied over from `(a *App) sendToPushProxy`. What I'd suggest is to refactor that code to handle nil session as well and name that `sendToPushProxyRaw`, and have 2 high-level methods - `sendToPushProxy` and `sendToPushProxyNoSession`, and call `sendToPushProxyNoSession` for this purpose.

agnivade (commented):
Lastly, let's also add a test please :)

agnivade:
Also, this needs changes to the mattermost-api-reference repo.

api4/system.go
larkox: The problem is that returning false will mean "You cannot receive push notifications from this server". And that is a value that will be stored in the client "forever". This error may be something temporal. That will translate into the 3 states possible in the client: Can receive, cannot receive or unknown. We could hide the error from the client, by sending directly unknown, but not sure if we want to give more information to the user.

About logging or not a warning, I have no problem in doing that.

agnivade: > We could hide the error from the client, by sending directly unknown,

I like it, let's do that :+1:

agnivade (commented):
This isn't quite what I was suggesting. There is still some repetition going on here. What I had in mind was to have a single method that can handle a nil session and avoid doing anything session related in that.

So far the session is only used in the Info log method and in the `model.PushStatusRemove` case. For the first one, we can have a slice of mlog.Fields and conditionally append the session field. And then for the second one, an if condition will suffice.

This makes a single method work with and without session. After this, we have 2 top-level methods - one having session in it's signature, the other doesn't, and passes a nil session to this underlying method.

Sorry for the double work here, but it's good to keep things clean in this notification flow.

agnivade (approved):
Discussed offline. Looks good.

larkox:
@enahum Should this PR wait for https://github.com/mattermost/mattermost-push-proxy/pull/96 before merge?

enahum:
@larkox no, this can be merged now.. no need to wait for the proxy

PR #19619, opened 2022-02-23
Fixed MM-42027

AGMETEOR:
#### Summary
Fixes MM-42027

#### Ticket Link
[MM-42027](https://mattermost.atlassian.net/browse/MM-42027)

#### Release Note

```release-note
NONE
```

### DIFF
--- a/app/team.go
+++ b/app/team.go
@@ -1913,7 +1913,10 @@ func (a *App) InvalidateAllEmailInvites() *model.AppError {
 }
 
 func (a *App) InvalidateAllResendInviteEmailJobs() *model.AppError {
-	jobs, _ := a.Srv().Jobs.GetJobsByTypeAndStatus(model.JobTypeResendInvitationEmail, model.JobStatusPending)
+	jobs, appErr := a.Srv().Jobs.GetJobsByTypeAndStatus(model.JobTypeResendInvitationEmail, model.JobStatusPending)
+	if appErr != nil {
+		return appErr
+	}
 
 	for _, j := range jobs {
 		a.Srv().Jobs.SetJobCanceled(j)
--- a/app/team.go
+++ b/app/team.go
@@ -1906,6 +1906,21 @@ func (a *App) InvalidateAllEmailInvites() *model.AppError {
 	if err := a.Srv().Store.Token().RemoveAllTokensByType(TokenTypeGuestInvitation); err != nil {
 		return model.NewAppError("InvalidateAllEmailInvites", "api.team.invalidate_all_email_invites.app_error", nil, err.Error(), http.StatusBadRequest)
 	}
+	if err := a.InvalidateAllResendInviteEmailJobs(); err != nil {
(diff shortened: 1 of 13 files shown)


app/team.go
```diff
@@ -1906,6 +1906,24 @@ func (a *App) InvalidateAllEmailInvites() *model.AppError {
 	if err := a.Srv().Store.Token().RemoveAllTokensByType(TokenTypeGuestInvitation); err != nil {
 		return model.NewAppError("InvalidateAllEmailInvites", "api.team.invalidate_all_email_invites.app_error", nil, err.Error(), http.StatusBadRequest)
 	}
+	if err := a.InvalidateAllResendInviteEmailJobs(); err != nil {
+		return model.NewAppError("InvalidateAllEmailInvites", "api.team.invalidate_all_email_invites.app_error", nil, err.Error(), http.StatusBadRequest)
+	}
```
marianunez, line 1911: @AGMETEOR it seems this change is disabling any pending reminders correct? However, based on the ticket the invitation link still worked in the reminder email? Shouldn't the line above had invalidated the token of the invitation link of the reminder as well?

AGMETEOR: Not really. In fact what we’re doing here is to stop any reminder jobs from running again. The only reason, the token from reminder is valid is because everytime that job runs, it creates a new token.

marianunez: Oh ok, that makes sense! Thanks for clarifying 👍

app/team.go
```diff
@@ -1906,6 +1906,24 @@ func (a *App) InvalidateAllEmailInvites() *model.AppError {
 	if err := a.Srv().Store.Token().RemoveAllTokensByType(TokenTypeGuestInvitation); err != nil {
 		return model.NewAppError("InvalidateAllEmailInvites", "api.team.invalidate_all_email_invites.app_error", nil, err.Error(), http.StatusBadRequest)
 	}
+	if err := a.InvalidateAllResendInviteEmailJobs(); err != nil {
+		return model.NewAppError("InvalidateAllEmailInvites", "api.team.invalidate_all_email_invites.app_error", nil, err.Error(), http.StatusBadRequest)
```
marianunez, line 1910: Should we return 500 here instead?

marianunez (approved):
Great work @AGMETEOR! 💯 with the tests here👍

srkgupta (approved):
Thanks @AGMETEOR. The changes LGTM. Also tested and the issue is now working fine as expected.

nickmisasi (approved):

stevemudie (approved):
Invited, invalidated and no email went out after the invalidation as expected. Looks good @AGMETEOR!

app/team.go
AGMETEOR: Yes, I think I should change even the ones above. @marianunez what do you think?

marianunez: Agreed, those doesn't seem bad request but server error

PR #19621, opened 2022-02-24
MM-36668: Trim large slash commands before logging

agnivade:
If a slash command is not found, we would print the input
as-is in the logs. To keep the logs clean, we trim the message
beyond a threshold length.

https://mattermost.atlassian.net/browse/MM-36668
```release-note
NONE
```


isacikgoz (approved):

srkgupta (approved):
Thanks @agnivade. Changes LGTM.

PR #19633, opened 2022-02-24
[MM-42068] Add check to stop custom groups linking to teams and channels

BenCookie95:
#### Summary
Should not be able to link custom groups to any teams or channels

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-42069
https://mattermost.atlassian.net/browse/MM-42071

#### Release Note

```release-note
Fixing a bug that would allow you to link a custom group to a team or channel
```


api4/group_test.go
```diff
@@ -328,6 +328,19 @@ func TestLinkGroupTeam(t *testing.T) {
...
+		DisplayName: "dn_" + gid,
+		Name:        model.NewString("name" + gid),
+		Source:      model.GroupSourceCustom,
+		Description: "description_" + gid,
+		RemoteId:    model.NewString(model.NewId()),
+	})
+	assert.Nil(t, app2Err)
+
+	_, _, err = th.Client.LinkGroupSyncable(g2.Id, th.BasicTeam.Id, model.GroupSyncableTypeTeam, patch)
+	require.Error(t, err)
```
mylonsuren, line 343: For this test, I think it would be good to also verify that the correct error is returned, otherwise this test could give a false positive if the implementation changes in the future.

api4/group_test.go
```diff
@@ -370,6 +383,19 @@ func TestLinkGroupChannel(t *testing.T) {
...
+		DisplayName: "dn_" + gid,
+		Name:        model.NewString("name" + gid),
+		Source:      model.GroupSourceCustom,
+		Description: "description_" + gid,
+		RemoteId:    model.NewString(model.NewId()),
+	})
+	assert.Nil(t, app2Err)
+
+	_, _, err = th.Client.LinkGroupSyncable(g2.Id, th.BasicChannel.Id, model.GroupSyncableTypeChannel, patch)
+	require.Error(t, err)
```
mylonsuren, line 398: Same as my other comment:

> For this test, I think it would be good to also verify that the correct error is returned, otherwise this test could give a false positive if the implementation changes in the future.

mylonsuren (changes requested):

api4/group.go
```diff
@@ -297,6 +297,17 @@ func linkGroupSyncable(c *Context, w http.ResponseWriter, r *http.Request) {
 		return
 	}
 
+	group, groupErr := c.App.GetGroup(c.Params.GroupId, nil)
+	if groupErr != nil {
+		c.Err = groupErr
+		return
+	}
+
+	if group.Source == model.GroupSourceCustom {
```
mkraft, line 306: We want to reject everything except Ldap rather than only rejecting custom.

```suggestion
	if group.Source != model.GroupSourceLdap {
```

api4/group_test.go
mkraft: I also think based on the change I've recommended in the conditional we should test:
1. source is custom
2. source is something random

mkraft (changes requested):

api4/group_test.go
BenCookie95: Unfortunately we can't test with a random source because whenever we create a group it ensures the source is either `ldap` or `custom`

BenCookie95: > For this test, I think it would be good to also verify that the correct error is returned, otherwise this test could give a false positive if the implementation changes in the future.

Done

mkraft (approved):

mylonsuren (approved):

furqanmlk (approved):
LGTM
**TEAM**

image

**CHANNEL**

image

PR #19641, opened 2022-02-25
[MM-42022] Don't break importing process on missing display name of a channel

Willyfrog:
#### Summary
Import is failing when display name is missing or empty, this will allow it adding the value of name to it. 

The reason for substituting the value is because that's what the channel will show in the LHS, so we might as well provide a value instead of allowing something that might fail somewhere else (the webapp doesn't allow you to create an empty display name)

#### Ticket Link
[MM-42022](https://mattermost.atlassian.net/browse/MM-42022)

#### Release Note
```release-note
Fixed error when importing a channel missing its display name. It now uses name as display name if it is missing or empty
```

### DIFF
--- a/app/import_validators.go
+++ b/app/import_validators.go
@@ -165,9 +165,9 @@ func validateChannelImportData(data *ChannelImportData) *model.AppError {
 		return model.NewAppError("BulkImport", "app.import.validate_channel_import_data.name_characters.error", nil, "", http.StatusBadRequest)
 	}
 
-	if data.DisplayName == nil {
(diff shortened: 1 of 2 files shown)


gabrieljackson (approved):
LGTM

amyblais:
@hanzei FYI, we're aiming to get this in for next week's cloud release.
@Willyfrog Who would be the best QA for this?

hanzei (approved):
LGTM :+1:

saturninoabril:
@Willyfrog Thanks for checking! I'll take a look.

Willyfrog:
@amyblais let me know if this needs to be cherry-picked

PR #19642, opened 2022-02-25
fix string constants

neallred:
#### Summary
Mismatched key name between the migration code and the status fetching code meant that even the migration completed successfully, when the client tried to fetch the status of whether admin onboarding had been completed, they falsely saw it reported as `false`.

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-42103

#### Release Note

```release-note
NONE
```

### DIFF
--- a/app/migrations.go
+++ b/app/migrations.go
@@ -17,7 +17,7 @@ const GuestRolesCreationMigrationKey = "GuestRolesCreationMigrationComplete"
 const SystemConsoleRolesCreationMigrationKey = "SystemConsoleRolesCreationMigrationComplete"
 const ContentExtractionConfigDefaultTrueMigrationKey = "ContentExtractionConfigDefaultTrueMigrationComplete"
 const PlaybookRolesCreationMigrationKey = "PlaybookRolesCreationMigrationComplete"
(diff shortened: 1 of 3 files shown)


app/migrations.go
```diff
@@ -17,7 +17,7 @@ const GuestRolesCreationMigrationKey = "GuestRolesCreationMigrationComplete"
 const SystemConsoleRolesCreationMigrationKey = "SystemConsoleRolesCreationMigrationComplete"
 const ContentExtractionConfigDefaultTrueMigrationKey = "ContentExtractionConfigDefaultTrueMigrationComplete"
 const PlaybookRolesCreationMigrationKey = "PlaybookRolesCreationMigrationComplete"
-const FirstAdminSetupCompleteKey = "FirstAdminSetupComplete"
+const FirstAdminSetupCompleteKey = model.SystemFirstAdminSetupComplete
```
marianunez, line 20: Ah nice catch, all the above would probably benefit from having them defined as constants too

marianunez (approved):

JulienTant (approved):
LGTM!

PR #19644, opened 2022-02-25
[MM-41907] Don't return custom groups when feature is disabled

BenCookie95:
#### Summary
When custom groups feature flag is disabled or config is disabled, don't return any custom groups in the GET response

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-41907

#### Release Note

```release-note
No longer return custom groups for GET /api/v4/groups when feature is disabled
```


api4/group_test.go
```diff
@@ -1201,6 +1201,28 @@ func TestGetGroups(t *testing.T) {
...
+	opts.Source = model.GroupSourceLdap
+	groups, _, err = th.Client.GetGroups(opts)
+	assert.NoError(t, err)
+	assert.Len(t, groups, 1)
+
+	// don't include source and should only get ldap groups in response
+	opts.Source = ""
+	groups, _, err = th.Client.GetGroups(opts)
+	assert.NoError(t, err)
+	assert.Len(t, groups, 1)
```
mylonsuren, line 1225: Nit: for each of these can we also add:
```go
assert.Equal(t, groups[0].id, group.id)
```
to verify that the group returned is an ldap group.

mylonsuren (approved):

furqanmlk (approved):
Working as expected

image

image

image

image

image

image

PR #19653, opened 2022-03-01
[MM-40962] - Remove log when checking renewal token validity

AGMETEOR:
#### Summary
Following the report from the ticket attached to this PR, decision was made to remove the log in the diff

#### Ticket Link
[MM-40962](https://mattermost.atlassian.net/browse/MM-40962)

#### Release Note

```release-note
NONE

```

### DIFF
--- a/app/license.go
+++ b/app/license.go
@@ -313,10 +313,7 @@ func (s *Server) GenerateRenewalToken(expiration time.Duration) (string, *model.
 
 	currentToken, _ := s.Store.System().GetByName(model.SystemLicenseRenewalToken)
 	if currentToken != nil {
-		tokenIsValid, err := s.renewalTokenValid(currentToken.Value, license.Customer.Email)
-		if err != nil {
-			mlog.Warn("error checking license renewal token validation", mlog.Err(err))
-		}
+		tokenIsValid, _ := s.renewalTokenValid(currentToken.Value, license.Customer.Email)
 		if currentToken.Value != "" && tokenIsValid {
 			return currentToken.Value, nil
 		}


marianunez (approved):

JulienTant:
LGTM Thanks Allan!

agnivade:
I ran into this PR because we are seeing the same "Error validating JWT token: token is expired" log while running our load-tests.

Ignoring errors is a bad practice and makes debugging things harder. If we don't want to check the error, we should remove all error handling in the method altogether. There is no point in returning an error and ignoring it. We should always be mindful of these things.

But more importantly, do we really want to ignore the error? The error distinguishes between various failures states of parsing error, invalid token and token expired. Is it okay to completely ignore the error distinction here?

From the original issue which Harrison created, it seems like he landed up in a weird situation where he had an EE instance with a license and then got downgraded to a TE instance. I am wondering why we cannot detect the case explicitly and mention a nice error message saying "you have a license attached to a TE edition"? Or detect the case and ignore it specifically. I think we should look into the root cause here rather than just ignore the error. And if ignoring the error is indeed the right solution here, we should remove all error handling and make the method return just a boolean.

AGMETEOR:
> I ran into this PR because we are seeing the same "Error validating JWT token: token is expired" log while running our load-tests.
> 
> Ignoring errors is a bad practice and makes debugging things harder. If we don't want to check the error, we should remove all error handling in the method altogether. There is no point in returning an error and ignoring it. We should always be mindful of these things.
> 
> But more importantly, do we really want to ignore the error? The error distinguishes between various failures states of parsing error, invalid token and token expired. Is it okay to completely ignore the error distinction here?
> 
> From the original issue which Harrison created, it seems like he landed up in a weird situation where he had an EE instance with a license and then got downgraded to a TE instance. I am wondering why we cannot detect the case explicitly and mention a nice error message saying "you have a license attached to a TE edition"? Or detect the case and ignore it specifically. I think we should look into the root cause here rather than just ignore the error. And if ignoring the error is indeed the right solution here, we should remove all error handling and make the method return just a boolean.

Thanks @agnivade , reopened in order to solve as suggested

PR #19658, opened 2022-03-01
MM-40818: Refactor product initialization to happen late

agnivade:
This is more of a general refactor of the initialization
process which should allow us to pass services more
easily.

The changes are minimal to keep the scope limited.
For now, the objective is to pass the file service
to the Channels product. For that, it was required
to move some of the enterprise interfaces under Channels
from Server.

We also create a filestore field in the server to
avoid creating filestore reference every time
we make a filestore operation. This will be later
passed on to the Channels product.

```release-note
NONE
```

### DIFF
--- a/api4/ldap_test.go
+++ b/api4/ldap_test.go
@@ -153,7 +153,7 @@ func TestSyncLdap(t *testing.T) {
 		includeRemovedMembers = args[1].(bool)
 		ready <- true
 	}
-	th.App.Srv().Ldap = ldapMock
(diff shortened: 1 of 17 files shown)


agnivade:
@isacikgoz - This is what I came up with. I just sent the refactor PR as a first step. But with this, we should be able to pass the filestore service to Channels during the product init phase. LMK what you think.

isacikgoz (approved):
I think this looks neat, we are initializing the filestore before the product initialization 💯

PR #19662, opened 2022-03-01
fix plugin help example - method arguments order

davidboto:
#### Summary

I've noticed a bug while I was trying to test the "Help Plugin" example available at: https://developers.mattermost.com/integrate/plugins/server/reference/#Examples._helpPlugin 

Apparently, the order in which the arguments are passed doesn't match the order expected by the method. 

This PR fixes the order in which these arguments are being passed.

#### Release Note

```release-note
NONE
```


catalintomai:
@davidboto - looks like the signature in the server repo is the one used by other plugins as well (see image). What issue did you encounter and how did you trace it to the invalid argument order?
Screen Shot 2022-04-26 at 7 30 42 AM

catalintomai (approved):

hanzei (approved):

PR #19668, opened 2022-03-02
[MM-42121] - Inactive Server Email Notification sends duplicate emails of every admin to original admin

AGMETEOR:
#### Summary
We send inactivity emails to all system admins concurrently using go routines spun up in a for loop over a slice of these system admins. However, due to how golang works, we shall have only one value of the system admin at the time when go routines are spun up, hence the duplication of emails

#### Ticket Link
[MM-42121](https://mattermost.atlassian.net/browse/MM-42121)

#### Release Note

```release-note
NONE

```


app/server_inactivity.go
```diff
@@ -112,6 +112,7 @@ func (s *Server) takeInactivityAction() {
 	}
 
 	for _, user := range users {
+		user := user
```
nickmisasi, line 115: Do you know why this actually fixes it?

Doesn't seem like it changes anything.

AGMETEOR: Yes, it does fix it.

nickmisasi: Yeah but _why_ does this work, when not having it doesn't? 

I'm assuming there's a race between the go routines being launched and the value changing for `user` but I'm concerned that this is just masking the problem, since all this does is re sets the user value. 

I think the proper way to do this is editing the actual go routine to look like this:

```
s.Go(func(user *User) {
			if err := s.EmailService.SendLicenseInactivityEmail(user.Email, name, user.Locale, siteURL); err != nil {
				mlog.Error("Error while sending inactivity reminder email.", mlog.String("user_email", user.Email), mlog.Err(err))
			}
		}(user))
```

So that when the go routine is invoked, the user is passed as an argument to it

AGMETEOR: There’s two ways to solve that problem of goroutines and closures. The one you’re suggesting and the one I have added.

marianunez (commented):
Can you add a comment for readability on why this assignment is needed? Maybe a different variable name would reduce confusion on what is happening

AGMETEOR:
Hey @nickmisasi waiting for your response

AGMETEOR:
> Can you add a comment for readability on why this assignment is needed? Maybe a different variable name would reduce confusion on what is happening

@marianunez added suggested documentation

marianunez (approved):

app/server_inactivity.go
nickmisasi: The comment with link to the doc was helpful.
image
You can see that the doc mentions my solution as well as yours, I've never seen your approach before so it just felt off to me. I'm cool with it though, thanks for clarifying!

nickmisasi (approved):

app/server_inactivity.go
agnivade: I ran into this after Allan pointed me to the issue.

> I've never seen your approach before so it just felt off to me. 

This is quite idiomatic and in fact, used in a lot of places in our codebase. See https://github.com/mattermost/mattermost-server/pull/13969.

AGMETEOR:
Hey @stevemudie gentle ping here

stevemudie (approved):
Will test this post merge.

PR #19692, opened 2022-03-04
Fix MM-42270

AGMETEOR:
#### Summary
Fix MM-42270

#### Release Note

```release-note
NONE

```

### DIFF
--- a/api4/system_test.go
+++ b/api4/system_test.go
@@ -209,9 +209,6 @@ func TestGenerateSupportPacket(t *testing.T) {
 	})
 
 	t.Run("As a System Administrator but with RestrictSystemAdmin true", func(t *testing.T) {
-		l := model.NewTestLicense()
-		th.App.Srv().SetLicense(l)
-
 		originalRestrictSystemAdminVal := *th.App.Config().ExperimentalSettings.RestrictSystemAdmin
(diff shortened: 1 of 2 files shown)


srkgupta (approved):
Thanks @AGMETEOR. Changes LGTM

api4/system_test.go
```diff
@@ -208,6 +208,23 @@ func TestGenerateSupportPacket(t *testing.T) {
 		require.NotZero(t, len(file))
 	})
 
+	t.Run("As a System Administrator but with RestrictSystemAdmin true", func(t *testing.T) {
+		l := model.NewTestLicense()
```
marianunez, line 212: Is this needed ?

marianunez (approved):
Thanks @AGMETEOR!

api4/system_test.go
AGMETEOR: Not really because on the endpoint, the line that checks for permissions comes before the check for the license. But no harm in setting up the test with a license

AGMETEOR:
Hello @nickmisasi gentle ping here :)

api4/system_test.go
marianunez: If it's not relevant for the test we should remove it to keep the code cleaner

AGMETEOR: Removed it

marianunez: Thanks!

PR #19709, opened 2022-03-04
break up getThreadsForUser, leverage errgroup

lieut-data:
#### Summary
This change breaks up `GetThreadsForUser` in the `ThreadStore` into its constituent `GetTotalUnreadThreads`, `GetTotalThreads`, `GetTotalUnreadMentions`, and the original `GetThreadsForUser` but now solely returning the thread structures. Instead of a monolithic method at the store level, the application layer now handles calling bulk requests, leveraging `errgroup` for simpler parallelization.

This change brings with it a few benefits:
* Simpler code, including more idiomatic usage of squirrel
* Simpler SQL, joining tables only when configured conditions require same. (No performance benefit here, since an unused LEFT JOIN generally has no overhead.)
* Discrete metrics for each store method, giving us better insight into the performance characteristics in play.
* **Performance boost**: reduced overhead when clearing push notifications.

This last point is what prompted the re-re-refactoring in this PR. As I broke things up, I realized that `clearPushNotificationSync` only used the `TotalUnreadMentions`, but asked for the count of total threads and total unread threads. By exposing the discrete methods, this code path now avoids two aggregate queries. We clear notifications when marking a thread as read, and when marking a channel with unread mentions as viewed, so I expect we'll see at least a modest boost to performance from simply not wasting these cycles anymore.

No performance improvements are expected from this PR for the general case of using `GetThreadsForUser` to populate the threads view.

#### Ticket Link
None.

#### Release Note
```release-note
Improved performance when clearing notifications with CRT enabled.
```


lieut-data:
(Sorry for the force pushes here: I hope I got them in before anyone "saw" the review! Context was that I thought I had missed a change, then realized it was meant to be that way, so I reverted it, but didn't want to spam the PR.)

lieut-data:
I'm going to pause on this PR while I pursue some additional changes. I realized it would be tremendously useful to have a more detailed breakdown of the API times in Grafana for the individual portions of these queries, and we really should just aggregate them in the app layer anyway.

agnivade:
Using errgroups and separating the totals method was absolutely necessary. Nice work!

lieut-data:
@agnivade & @ashishbhate: apologies for inviting you, then removing you, then force pushing a half dozen times, though the history is at least interesting here. TLDR; I've rewritten the summary and you should re-read what these changes do. It's still the same basic breaking up of `GetThreadsForUser`, but with an (unmeasured) performance boost for the case when we clear notifications.

I'm opting not to do a perf test pass since I'm just removing SQL calls here, not tweaking semantics. I'm also looking forward to the granular Grafana charts to help us better pinpoint how we want to proceed in tackling the performance issues with this (previously) monolithic query.

app/notification_push.go
```diff
@@ -236,11 +236,11 @@ func (a *App) clearPushNotificationSync(currentSessionId, userID, channelID, roo
 	msg.Badge = int(unreadCount)
 
 	if msg.IsCRTEnabled {
-		data, err := a.Srv().Store.Thread().GetThreadsForUser(userID, "", model.GetUserThreadsOpts{TotalsOnly: true})
+		totalUnreadMentions, err := a.Srv().Store.Thread().GetTotalUnreadMentions(userID, "", model.GetUserThreadsOpts{})
```
lieut-data, line 239: This is the performance boost here -- we no longer count thread memberships nor unread threads only to discard this metadata.

app/user.go
```diff
@@ -2264,15 +2266,61 @@ func (a *App) ConvertBotToUser(bot *model.Bot, userPatch *model.UserPatch, sysad
 }
 
 func (a *App) GetThreadsForUser(userID, teamID string, options model.GetUserThreadsOpts) (*model.Threads, *model.AppError) {
-	threads, err := a.Srv().Store.Thread().GetThreadsForUser(userID, teamID, options)
-	if err != nil {
+	var result model.Threads
```
lieut-data, line 2269: I really like how this played out, with the app layer doing the parallelization, and keeping the sql store "simple". Something to aspire for more broadly?

store/sqlstore/thread_store.go
```diff
@@ -52,247 +52,259 @@ func (s *SqlThreadStore) Get(id string) (*model.Thread, error) {
...
-	type JoinedThread struct {
-		PostId         string
-		ReplyCount     int64
-		LastReplyAt    int64
-		LastViewedAt   int64
-		UnreadReplies  int64
-		UnreadMentions int64
-		Participants   model.StringArray
-		model.Post
+// GetTotalUnreadThreads counts the number of unread threads for the given user, optionally
```
lieut-data, line 55: Don't forget to review this file!

store/sqlstore/thread_store.go
```diff
@@ -52,247 +52,259 @@ func (s *SqlThreadStore) Get(id string) (*model.Thread, error) {
...
+		return 0, errors.Wrapf(err, "failed to count unread threads for user id=%s", userId)
+	}
+
+	return totalUnreadThreads, nil
+}
+
+// GetTotalUnreadThreads counts the number of threads for the given user, optionally constrained
+// to the given team + DMs/GMs.
+//
+// TODO: Why do we support an Unread flag here? It's basically the same as GetTotalUnreadThreads,
```
lieut-data, line 95: I've added this as a TODO without addressing for now, but in short, there's probably another minor opportunity for improvement here.

store/storetest/thread_store.go
```diff
@@ -679,7 +681,35 @@ func testGetTeamsUnreadForUser(t *testing.T, ss store.Store) {
 	assert.Equal(t, int64(1), teamsUnread[team2.Id].ThreadMentionCount)
 }
 
-func testGetThreadsForUser(t *testing.T, ss store.Store) {
+func testVarious(t *testing.T, ss store.Store) {
```
lieut-data, line 684: I wanted to reuse the same "data structure" for these tests, while breaking them up, and found a helper function would make me return half a dozen created objects. Open to feedback for how to organize this better!

store/storetest/thread_store.go
```diff
@@ -773,61 +817,206 @@ func testGetThreadsForUser(t *testing.T, ss store.Store) {
...
-		opts := store.ThreadMembershipOpts{
-			Following:             true,
-			IncrementMentions:     false,
-			UpdateFollowing:       true,
-			UpdateViewedTimestamp: false,
-			UpdateParticipants:    false,
+	err = ss.Post().Delete(team2channel1post2deleted.Id, model.GetMillis(), user1ID)
+	require.NoError(t, err)
+
+	t.Run("GetTotalUnreadThreads", func(t *testing.T) {
```
lieut-data, line 848: All three aggregate routines are now tested in their entirety.

store/storetest/thread_store.go
```diff
@@ -773,61 +817,206 @@ func testGetThreadsForUser(t *testing.T, ss store.Store) {
...
-		assert.Len(t, threads.Threads, 4)
+				assert.EqualValues(t, int64(len(testCase.ExpectedThreads)), totalUnreadMentions)
+			})
+		}
 	})
 
-	t.Run("team2 specified, user1", func(t *testing.T) {
-		threads, err := ss.Thread().GetThreadsForUser(user1ID, team2.Id, model.GetUserThreadsOpts{})
-		require.NoError(t, err)
+	t.Run("GetThreadsForUser", func(t *testing.T) {
```
lieut-data, line 962: This isn't 100% coverage yet, but broader than before. (I know we have the `api4` tests already, but I found it helpful to test holistically here too.)

app/user.go
```diff
@@ -2264,15 +2266,61 @@ func (a *App) ConvertBotToUser(bot *model.Bot, userPatch *model.UserPatch, sysad
...
+	var eg errgroup.Group
+
+	eg.Go(func() error {
+		totalUnreadThreads, err := a.Srv().Store.Thread().GetTotalUnreadThreads(userID, teamID, options)
+		if err != nil {
+			return errors.Wrapf(err, "failed to count unread threads for user id=%s", userID)
+		}
+		result.TotalUnreadThreads = totalUnreadThreads
```
lieut-data, line 2277: @agnivade, can you sanity check that it's safe to write to /different/ struct properties from multiple goroutines. Intuitively, I think this is safe, but want to make sure!

agnivade: Yes, it's safe. To confirm, you can run mattermost in `-race` mode and see if there are any issues.

app/user.go
agnivade: Agree.

store/sqlstore/thread_store.go
```diff
@@ -52,247 +52,259 @@ func (s *SqlThreadStore) Get(id string) (*model.Thread, error) {
...
 	}
 
-	fetchConditions := sq.And{
-		sq.Eq{"ThreadMemberships.UserId": userId},
-		sq.Eq{"ThreadMemberships.Following": true},
+	if !opts.Deleted {
+		query = query.Where(sq.Eq{"COALESCE(Posts.DeleteAt, 0)": 0})
 	}
 
+	sql, args, _ := query.ToSql()
```
agnivade, line 83: Let's check for error from these methods to be thorough :)

app/user.go
```diff
@@ -2264,15 +2266,61 @@ func (a *App) ConvertBotToUser(bot *model.Bot, userPatch *model.UserPatch, sysad
 }
 
 func (a *App) GetThreadsForUser(userID, teamID string, options model.GetUserThreadsOpts) (*model.Threads, *model.AppError) {
-	threads, err := a.Srv().Store.Thread().GetThreadsForUser(userID, teamID, options)
-	if err != nil {
+	var result model.Threads
+	var eg errgroup.Group
+
+	eg.Go(func() error {
```
agnivade, line 2272: I am wondering if we should wait (or perhaps coordinate) with the `ThreadsOnly` PR https://github.com/mattermost/mattermost-server/pull/19349? This will slightly change the semantics of that PR. 0/5 from my side, I just want both PRs to get in with minimal re-work.

store/sqlstore/thread_store.go
```diff
@@ -52,247 +52,259 @@ func (s *SqlThreadStore) Get(id string) (*model.Thread, error) {
...
+			})
 	}
+
 	if !opts.Deleted {
-		fetchConditions = sq.And{
-			fetchConditions,
-			sq.Eq{"COALESCE(Posts.DeleteAt, 0)": 0},
-		}
+		query = query.
+			LeftJoin("Posts ON Posts.Id = ThreadMemberships.PostId").
```
agnivade, line 120: I like that joins are conditional as well now. Nice work

store/sqlstore/thread_store.go
```diff
@@ -52,247 +52,259 @@ func (s *SqlThreadStore) Get(id string) (*model.Thread, error) {
...
+	if !opts.Deleted {
+		unreadRepliesQuery = unreadRepliesQuery.Where(sq.Eq{"Posts.DeleteAt": 0})
+	}
 
-			unreadRepliesQuery, _ := sq.
-				Select("COUNT(Posts.Id)").
-				From("Posts").
-				Where(unreadRepliesFetchConditions).
-				MustSql()
+	unreadRepliesSql, unreadRepliesArgs := unreadRepliesQuery.MustSql()
```
agnivade, line 203: No panics please :)

store/sqlstore/thread_store.go
```diff
@@ -52,247 +52,259 @@ func (s *SqlThreadStore) Get(id string) (*model.Thread, error) {
...
-				Select(`Threads.*,
+	query := s.getQueryBuilder().
+		Select(`Threads.*,
 				` + postSliceCoalesceQuery() + `,
 				ThreadMemberships.LastViewed as LastViewedAt,
 				ThreadMemberships.UnreadMentions as UnreadMentions`).
-				From("Threads").
-				Column(sq.Alias(sq.Expr(unreadRepliesQuery), "UnreadReplies")).
-				LeftJoin("Posts ON Posts.Id = Threads.PostId").
-				LeftJoin("Channels ON Posts.ChannelId = Channels.Id").
```
agnivade, line 200: This is interesting. It seems like we were joining with the Channels table unnecessarily?

store/sqlstore/thread_store.go
```diff
@@ -52,247 +52,259 @@ func (s *SqlThreadStore) Get(id string) (*model.Thread, error) {
...
-		threads := threadsResult.Data.([]*JoinedThread)
-		for _, thread := range threads {
-			for _, participantId := range thread.Participants {
-				if _, ok := userIdMap[participantId]; !ok {
-					userIdMap[participantId] = true
-					userIds = append(userIds, participantId)
-				}
-			}
+	}
+	participantUserIds = model.RemoveDuplicateStrings(participantUserIds)
```
agnivade, line 268: Just to confirm, we are trading off memory for CPU time?

agnivade (commented):
Great work with the tests!

store/sqlstore/thread_store.go
lieut-data: Yes, good call -- I've pushed a change to address this.

Though I admit this used to be easier, when `GetMaster().Exec()` would accept the squirrel statement builder and do the `ToSQL` itself. Would we consider moving back to that model with the new `GetMasterX()`, and providing a back door as it were to run raw SQL?

app/user.go
lieut-data: Agreed, @isacikgoz, should we coordinate more closely on this front? Adding `ThreadsOnly` should be trivial on top of these changes, but I also know we discussed possibly reworking those changes to expose a new endpoint instead (which would still be easy with these changes).

store/sqlstore/thread_store.go
lieut-data: Agreed, I've now purged `MustSql` from this entire file. (See comment above for how error handling "used to be easier.")

store/sqlstore/thread_store.go
lieut-data: Indeed, we only needed the `Channels` table if we had to resolve the team id in question. But I believe a `LEFT JOIN` has no performance impact (beyond parsing the query) if it's never actually used. A `JOIN` would have been problematic, of course.

store/sqlstore/thread_store.go
lieut-data: I admit I didn't consider the performance characteristics of this change: I was focussed on simplifying the code, and noticed we already had a helper method to achieve the desired outcome and make the semantics of this block easier to understand. Thoughts?

agnivade: Ah no worries, I was just clarifying. This is fine. :+1:

app/user.go
agnivade: Did you mean to mention someone else? :wink: I don't think Ibrahim is working on this.

lieut-data: Haha, yes, I'm mixing up my work on CRT with my work with @isacikgoz yesterday on the mobile outage :D -- apologies, @isacikgoz! cc @ashishbhate :)

store/sqlstore/thread_store.go
agnivade: `GetMasterX` still has the same `Exec` method where you can pass raw sql. But AFAIR, `GetMaster().Exec()` didn't accept squirrel statement builder, only raw sqls. While converting the squirrel builder to raw sql, the error checking needs to happen _somewhere_. 0/5 if you want to move it somewhere else. But 4/5 on checking the error.

app/user.go
ashishbhate: @lieut-data this PR lays the ground work for splitting up things up even further. Lets get this PR in and I can rework the other https://github.com/mattermost/mattermost-server/pull/19349 and go from there.

store/sqlstore/thread_store.go
ashishbhate: This PR is great on so many levels. It sets the stage for more a prudent API and performance fixes, while also making the code much easier to understand. 

For example - this refactored code highlights how often we join with channels just to get the teamId. I wager the teamId is non-empty for most calls, and that denormalizing threadMemberships with teamId (with an appropriate index) might give us a decent performance boost.

store/sqlstore/thread_store.go
```diff
@@ -52,247 +52,278 @@ func (s *SqlThreadStore) Get(id string) (*model.Thread, error) {
...
+	if !opts.Deleted {
+		query = query.Where(sq.Eq{"COALESCE(Posts.DeleteAt, 0)": 0})
+	}
+
+	sql, args, err := query.ToSql()
+	if err != nil {
+		return 0, errors.Wrapf(err, "failed to build query to count unread threads for user id=%s", userId)
 	}
 
+	err = s.GetMasterX().Get(&totalUnreadThreads, sql, args...)
```
ashishbhate, line 88: Should we use the replica here? 

Same for other methods as well.

store/sqlstore/thread_store.go
lieut-data: Sounds good -- I'll likely follow up separately with a proposal that preserves the error handling but gets rid of the boilerplate.

store/sqlstore/thread_store.go
lieut-data: Ooh, good catch! I don't see any reason we can't start using replicas here.

lieut-data: However, I wonder if we might make this change holistically in a separate PR? I'd like to be able to isolate unexpected changes of behaviour, and it looks like much of `ThreadStore` will need to be touched.

agnivade:
@lieut-data - feel free to request a review when it's ready :)

lieut-data:
> @lieut-data - feel free to request a review when it's ready :)

@agnivade, ah thank you for the prompt. Because you had only commented before and not approved / rejected, I had assumed it was still on your radar and was just waiting patiently :p -- what's your process for distinguishing PRs on which you're expecting changes vs. those awaiting your review?

agnivade:
>  Because you had only commented before and not approved / rejected, I had assumed it was still on your radar and was just waiting patiently

Oh I see. I always "comment" or "approve". I don't like to use the "reject" option unless in rare cases because it somehow feels rude to me. :)

So whenever I have submitted my review, I consider the ball to be out of my court, and usually wait to be re-requested another review. Does that make sense?

lieut-data:
> Oh I see. I always "comment" or "approve". I don't like to use the "reject" option unless in rare cases because it somehow feels rude to me. :)
> 
> So whenever I have submitted my review, I consider the ball to be out of my court, and usually wait to be re-requested another review. Does that make sense?

Aha, yes, I understand the potential implications of appearing rude. Ironically, I also hesitate to "re-request" reviews before I perceive the reviewer has "finished."

At least for my own process, I rely heavily on the "Pull requests needing review" counter in our GitHub integration. My goal is to get that to zero quickly -- unblocking authors of PRs -- and I can safely ignore it while it stays at zero vs. having to check daily to see if I'm expected to re-review. Since only approving or rejecting moves it "off my radar", I've adopted a pattern of "kindly rejecting," usually with a comment of thanks and explaining the scope of my requested changes. Or I approve if I only have suggestions on which I'd be fine taking no action. In my experience, this appears to consistently pass ownership between the author and the reviewer very explicitly.

Anyway, I now know to happily re-request your review without worrying about seeming pushy :)

lieut-data:
@jgilliam17, I think we're close to completing dev review, so wondering if you'd be willing to help kick off another round of e2e testing for CRT on this PR? There should be no semantic changes in this PR.

agnivade (approved):
Fantastic!

jgilliam17:
Run with CRT feature flag enabled took hours to finish and has some skipped tests. Updating and will restart the cycle.

store/sqlstore/thread_store.go
ashishbhate: >However, I wonder if we might make this change holistically in a separate PR? I'd like to be able to isolate unexpected changes of behaviour, and it looks like much of ThreadStore will need to be touched.


SGTM

ashishbhate (approved):

lieut-data:
@jgilliam17, do you know if the CRT tests passed?

lieut-data:
No worries, thanks @jgilliam17! I'll go ahead and merge and we can continue evaluating on community :)

PR #19713, opened 2022-03-04
Exclude bots when restricting DMs to team members.

wiggin77:
#### Summary
This PR excludes bots when restricting DMs to members of a team.

#### Background
https://mattermost.atlassian.net/browse/MM-7968 provided a setting that allowed administrators to restrict DMs to team members only.  A side affect of this is that bots cannot create DM channels with users unless the bots are added to the user’s team.

It was decided (https://community.mattermost.com/core/pl/57ia5zdm7i8epyrbgyyx5xr8fo) to exclude bots from this restriction.

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-42318

#### Release Note
```release-note
When restricting direct messages to users on the same team, bots are now excluded from that restriction.
```

### DIFF
--- a/app/channel.go
+++ b/app/channel.go
@@ -315,12 +315,18 @@ func (a *App) GetOrCreateDirectChannel(c *request.Context, userID, otherUserID s
 
 	if *a.Config().TeamSettings.RestrictDirectMessage == model.DirectMessageTeam &&
 		!a.SessionHasPermissionTo(*c.Session(), model.PermissionManageSystem) {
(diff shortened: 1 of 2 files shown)


mkraft (changes requested):
I guess we should unit test this somehow.

wiggin77:
> I guess we should unit test this somehow.

@mkraft I don't disagree there should be some tests, however that will require creating tests for direct channel creation in general for which there are none.  IMO that is out of scope for this simple bug fix.  I can create a ticket to get unit tests created for this and the other channel API for which they are missing.

mkraft (approved):
🎉 🎉

PR #19718, opened 2022-03-07
Mitigate Docker Compose Race in Test Stages

DSchalla:
#### Summary
With the current code, if Docker Compose hasn't created the network within 5 seconds the pipeline will crash as the Docker commands are erroring out. This replaces the static sleep with a loop that checks for the network to be created yet.

#### Ticket Link
N/A

#### Release Note
```release-note
NONE
```

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -318,7 +318,7 @@ jobs:
       - run:
           name: Wait for docker compose
           command: |
-            sleep 5
+            until docker network inspect circleci_mm-test; do echo "Waiting for Docker Compose Network..."; sleep 1; done;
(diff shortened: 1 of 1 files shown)


metanerd (approved):
Thank you! LGTM

metanerd:
Maybe we can do some empty commits to check if the change consistently doesn't fail. Not sure though. We can always fix again.

agnivade (approved):
Let's go! Nice sleuthing @DSchalla !

amyblais:
@agnivade Does this need to be cherry-picked to v6.5 branch to fix failing cherry-picks, e.g. https://github.com/mattermost/mattermost-server/pull/19702?

PR #19722, opened 2022-03-07
MM-42345: Include stdlog level in default mlog config

wiggin77:
#### Summary
During server startup, the std lib logger is redirected to mlog to capture any logging from third-party libraries.  These log records are emitted using a specific log level so that they can be output to specific log targets, mixed with MM logs or separated.

It was intended this log level be included in the default mlog config but that was missed.  This PR is to add that to the default config such that all stdlib logs are included.

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-42345

#### Release Note
```release-note
Logs from third-party libraries are now included in default logging configuration.
```

### DIFF
--- a/shared/mlog/levels.go
+++ b/shared/mlog/levels.go
@@ -14,7 +14,7 @@ var (
 	LvlInfo  = logr.Info  // ID = 4
 	LvlDebug = logr.Debug // ID = 5
 	LvlTrace = logr.Trace // ID = 6
-	StdAll   = []Level{LvlPanic, LvlFatal, LvlError, LvlWarn, LvlInfo, LvlDebug, LvlTrace}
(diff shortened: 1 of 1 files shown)


mkraft (approved):

mkraft (changes requested):
We should probably unit test this.

isacikgoz (approved):
LGTM, 0/5 adding tests.

mkraft (approved):

wiggin77:
> @wiggin77 - Can I get your help identifying the changes required for the product documentation for this feature, please?

@cwarnermm I doubt this requires any documentation changes.  A bit more log output will now happen *if* they are at the debug log level, which is not standard for production.

PR #19723, opened 2022-03-07
MM-19334 Ported AutocompleteInTeamForSearch to Squirrel

noxer:
#### Summary

This PR ports "AutocompleteInTeamForSearch" and the methods it depends on to Squirrel. It also includes a small optimization for "buildFulltextClause" and moves "spaceFulltextSearchChar" to the file it's actually used in.

I've added buildLIKEClauseX and buildFulltextClauseX which generate the respective conditions for the WHERE clause. I've not removed the original methods, they seem to be used in other, not yet ported methods as well. Be aware that the arguments for these methods have slighly changed from their non-X counterparts. Instead of passing the columns as one string, they expect a list of strings. This simplifies the code.

As an optimization, "buildFulltextClause" and "buildFulltextClauseX" use the `strings.Map` function to replace the characters in "spaceFulltextSearchChars" with spaces, which itself is now just a string and not a list of one-character strings.

Squirred does not support UNIONs, so the subqueries are generated and then put into the UNION using `fmt.Sprintf`. The placeholders for Postgres need to be manually applied afterwards.

#### Ticket Link

Fixes https://github.com/mattermost/mattermost-server/issues/19334

#### Release Note

```release-note
NONE
```

### DIFF
--- a/store/sqlstore/channel_store.go
+++ b/store/sqlstore/channel_store.go
@@ -3300,7 +3300,7 @@ func (s SqlChannelStore) buildLIKEClauseX(term string, searchColumns ...string) 
 	}
 
 	// add a placeholder at the beginning and end
-	likeTerm = "%" + likeTerm + "%"
+	likeTerm = wildcardSearchTerm(likeTerm)
 
 	// Prepare the LIKE portion of the query.
 	var searchFields sq.Or
--- a/store/sqlstore/channel_store.go
+++ b/store/sqlstore/channel_store.go
@@ -2873,7 +2873,7 @@ func (s SqlChannelStore) AutocompleteInTeamForSearch(teamID string, userID strin
 	}
 
 	var (
-		channels model.ChannelList
+		channels = model.ChannelList{}
 		sql      string
 		args     []interface{}
 	)
--- a/store/sqlstore/channel_store.go
+++ b/store/sqlstore/channel_store.go
@@ -3247,6 +3247,33 @@ func (s SqlChannelStore) buildLIKEClause(term string, searchColumns string) (lik
 	return
 }
 
+func (s SqlChannelStore) buildLIKEClauseX(term string, searchColumns ...string) sq.Sqlizer {
+	// escape the special characters with *
+	likeTerm := sanitizeSearchTerm(term, "*")
+
+	if likeTerm == "" {
+		return nil
+	}
+
+	// add a placeholder at the beginning and end
+	likeTerm = "%" + likeTerm + "%"
+
+	// Prepare the LIKE portion of the query.
+	var searchFields sq.Or
+
+	for _, field := range searchColumns {
+		if s.DriverName() == model.DatabaseDriverPostgres {
+			expr := fmt.Sprintf("LOWER(%s) LIKE LOWER(?) ESCAPE '*'", field)
+			searchFields = append(searchFields, sq.Expr(expr, likeTerm))
+		} else {
+			expr := fmt.Sprintf("%s LIKE ? ESCAPE '*'", field)
+			searchFields = append(searchFields, sq.Expr(expr, likeTerm))
+		}
+	}
+
+	return searchFields
+}
+
 const spaceFulltextSearchChars = "<>+-()~:*\"!@"
 
 func (s SqlChannelStore) buildFulltextClause(term string, searchColumns string) (fulltextClause, fulltextTerm string) {
--- a/store/sqlstore/channel_store.go
+++ b/store/sqlstore/channel_store.go
@@ -2945,39 +2945,34 @@ func (s SqlChannelStore) AutocompleteInTeamForSearch(teamID string, userID strin
 }
 
 func (s SqlChannelStore) autocompleteInTeamForSearchDirectMessages(userID string, term string) ([]*model.Channel, error) {
-	qb := s.getQueryBuilder()
+	// create the main query
+	query := s.getQueryBuilder().Select("C.*", "OtherUsers.Username as DisplayName").
+		From("Channels AS C").
+		Join("ChannelMembers AS CM ON CM.ChannelId = C.Id").
+		Where(sq.Eq{
+			"C.Type":    model.ChannelTypeDirect,
+			"CM.UserId": userID,
+		}).
+		Limit(50)
 
-	// create the subquery first
+	// create the subquery
 	subQuery := s.getSubQueryBuilder().Select("ICM.ChannelId AS ChannelId", "IU.Username AS Username").
 		From("Users AS IU").
-		Join("ChannelMembers AS ICM ON ICM.UserId = IU.Id")
-
-	// this AND will be extended, depending on the LIKE clause
-	subWhere := sq.And{
-		sq.NotEq{"IU.Id": userID},
-	}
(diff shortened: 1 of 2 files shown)


noxer:
Squirrel is really picky when it comes to numbering the placeholders. Only the outer most query may define a placeholder format.

store/sqlstore/channel_store.go
```diff
@@ -2900,47 +2946,56 @@ func (s SqlChannelStore) AutocompleteInTeamForSearch(teamId string, userId strin
...
+		Limit(50)
+
+	// create the SQL query and argument list
+	sql, args, err := query.ToSql()
+	if err != nil {
+		return nil, errors.Wrap(err, "autocompleteInTeamForSearchDirectMessages_InnerJoin_Tosql")
+	}
+
+	// query the channel list from the database using SQLX
+	var channels model.ChannelList
```
agnivade, line 2995: `channels := model.ChannelList{}`

store/sqlstore/channel_store.go
```diff
@@ -2900,47 +2946,56 @@ func (s SqlChannelStore) AutocompleteInTeamForSearch(teamId string, userId strin
...
+	}
+	subQuery = subQuery.Where(subWhere)
+
+	// put the subquery into an INNER JOIN
+	innerJoin := subQuery.
+		Prefix("INNER JOIN (").
+		Suffix(") AS OtherUsers ON OtherUsers.ChannelId = C.Id")
+
+	// create the main query
+	query := qb.Select("C.*", "OtherUsers.Username as DisplayName").
```
agnivade, line 2978: Can we build the main query first, and then add the join clauses like `query = query.JoinClause(...)`. It becomes more readable that way and consistent with the style used.

store/sqlstore/channel_store.go
```diff
@@ -3247,14 +3302,46 @@ func (s SqlChannelStore) buildLIKEClause(term string, searchColumns string) (lik
...
+func (s SqlChannelStore) buildLIKEClauseX(term string, searchColumns ...string) sq.Sqlizer {
+	// escape the special characters with *
+	likeTerm := sanitizeSearchTerm(term, "*")
+
+	if likeTerm == "" {
+		return nil
+	}
+
+	// add a placeholder at the beginning and end
+	likeTerm = "%" + likeTerm + "%"
```
agnivade, line 3314: This seems like the functionality of `wildcardSearchTerm`. Any reason we are not calling that function?

store/sqlstore/channel_store.go
```diff
@@ -3286,6 +3373,51 @@ func (s SqlChannelStore) buildFulltextClause(term string, searchColumns string)
...
+		fulltextTerm = strings.ReplaceAll(fulltextTerm, "|", "")
+
+		// split the search term and append :* to each part
+		splitTerm := strings.Fields(fulltextTerm)
+		for i, t := range splitTerm {
+			splitTerm[i] = t + ":*"
+		}
+
+		// join the search term with &
+		fulltextTerm = strings.Join(splitTerm, " & ")
```
agnivade, line 3400: Seems like we could use this optimization in the other method as well?

store/sqlstore/channel_store.go
```diff
@@ -3286,6 +3373,51 @@ func (s SqlChannelStore) buildFulltextClause(term string, searchColumns string)
...
+		if strings.ContainsRune(spaceFulltextSearchChars, r) {
+			return ' '
+		}
+		return r
+	}, fulltextTerm)
+
+	// Prepare the FULLTEXT portion of the query.
+	if s.DriverName() == model.DatabaseDriverPostgres {
+		// remove all pipes |
+		fulltextTerm = strings.ReplaceAll(fulltextTerm, "|", "")
```
agnivade, line 3391: Let's use `ReplaceAll` in the other method as well?

store/sqlstore/channel_store.go
```diff
@@ -2900,47 +2946,56 @@ func (s SqlChannelStore) AutocompleteInTeamForSearch(teamId string, userId strin
...
-				WHERE
-					IU.Id != :UserId
-					%v
-				) AS OtherUsers ON OtherUsers.ChannelId = C.Id
-			WHERE
-			    C.Type = :ChannelType
-				AND CM.UserId = :UserId
-			LIMIT 50`
+func (s SqlChannelStore) autocompleteInTeamForSearchDirectMessages(userID string, term string) ([]*model.Channel, error) {
+	qb := s.getQueryBuilder()
```
agnivade, line 2954: This is used only once. Let's inline it at the query site?

agnivade (commented):
Great start @noxer ! Just some small comments to take care of.

isacikgoz:
I did a quick pass, didn't spot anything right away on top of the initial review by @agnivade. Going to look deeper once @agnivade's comments are addressed.

store/sqlstore/channel_store.go
```diff
@@ -2850,47 +2850,93 @@ func (s SqlChannelStore) AutocompleteInTeam(teamID, userID, term string, include
...
+		args = append(likeArgs, fullArgs...)
+	}
+
+	var err error
 
-		if _, err := s.GetReplica().Select(&channels, query, map[string]interface{}{"TeamId": teamId, "UserId": userId, "LikeTerm": likeTerm, "FulltextTerm": fulltextTerm, "ChannelType": model.ChannelTypeGroup}); err != nil {
-			return nil, errors.Wrapf(err, "failed to find Channels with term='%s'", term)
+	// since the UNION is not part of squirrel, we need to assemble it and then update
+	// the placeholders manually
+	if s.DriverName() == model.DatabaseDriverPostgres {
```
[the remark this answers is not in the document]
noxer: Sorry, I don't understand what the issue is here.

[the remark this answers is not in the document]
agnivade: Aha! I missed that you are using `subQueryBuilder` instead of `queryBuilder`. `subQueryBuilder` will always generate in Question format but `queryBuilder` will generate in question or dollar format depending on the DB. If we used `queryBuilder`, it wouldn't have worked because replacePlaceholders will only work on question format.

store/sqlstore/channel_store.go
agnivade: Just this one got missed out

agnivade (commented):
Everything looks great! Just one comment got missed out.

store/sqlstore/channel_store.go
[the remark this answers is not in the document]
noxer: Yes, I ran into that problem when testing. Then I discovered the `subQueryBuilder` and it works like a charm.

store/sqlstore/channel_store.go
noxer: Yes, sorry.

noxer: Fixed, thanks.

noxer:
> Everything looks great! Just one comment got missed out.

Fixed now. Sorry I missed that one.

isacikgoz (approved):
Nicely done 🙏

agnivade (approved):
Brilliant stuff @noxer !

PR #19728, opened 2022-03-07
MM-19336 Migrated GetMembersByIds and GetMembersByChannelIds to Squirrel

noxer:
#### Summary

This PR ports GetMembersByIds and GetMembersByChannelIds to Squirrel.

#### Ticket Link

Fixes https://github.com/mattermost/mattermost-server/issues/19336

#### Release Note

```release-note
NONE
```

### DIFF
--- a/store/sqlstore/channel_store.go
+++ b/store/sqlstore/channel_store.go
@@ -1843,7 +1843,7 @@ func (s SqlChannelStore) UpdateMultipleMembers(members []*model.ChannelMember) (
 
 		// TODO: Get this out of the transaction when is possible
 		var dbMember channelMemberWithSchemeRoles
-		if err := transaction.Get(&dbMember, sqlSelect, args); err != nil {
+		if err := transaction.Get(&dbMember, sqlSelect, args...); err != nil {
 			if err == sql.ErrNoRows {
 				return nil, store.NewErrNotFound("ChannelMember", fmt.Sprintf("channelId=%s, userId=%s", member.ChannelId, member.UserId))
 			}
--- a/store/sqlstore/channel_store.go
+++ b/store/sqlstore/channel_store.go
@@ -1816,7 +1816,7 @@ func (s SqlChannelStore) UpdateMultipleMembers(members []*model.ChannelMember) (
 			return nil, errors.Wrapf(err, "UpdateMultipleMembers_Update_ToSql ChannelID=%s UserID=%s", member.ChannelId, member.UserId)
 		}
 
-		if _, err := transaction.Exec(sqlUpdate, args...); err != nil {
+		if _, err = transaction.Exec(sqlUpdate, args...); err != nil {
 			return nil, errors.Wrap(err, "failed to update ChannelMember")
 		}
 
--- a/store/sqlstore/channel_store.go
+++ b/store/sqlstore/channel_store.go
@@ -1882,32 +1882,49 @@ func (s SqlChannelStore) UpdateMemberNotifyProps(channelID, userID string, props
 				"channelid": channelID,
 			}).ToSql()
 		if err != nil {
-			errors.Wrapf(err, "UpdateMemberNotifyProps_Update_ToSql channelID=%s and userID=%s", channelID, userID)
+			return nil, errors.Wrapf(err, "UpdateMemberNotifyProps_Update_Postgres_ToSql channelID=%s and userID=%s", channelID, userID)
 		}
 
 		_, err = tx.Exec(sql, args...)
-	} else {
+	} else if len(props) > 0 {
 		// It's difficult to construct a SQL query for MySQL
 		// to handle a case of empty map. So we just ignore it.
-		if len(props) > 0 {
-			// unpack the keys and values to pass to MySQL.
-			args, argString := constructMySQLJSONArgs(props)
-			args = append(args, userID, channelID)
 
-			// Example: UPDATE ChannelMembers
-			// SET NotifyProps = JSON_SET(NotifyProps, '$.mark_unread', '"yes"' [, ...])
-			// WHERE ...
-			_, err = tx.Exec(`UPDATE ChannelMembers
-				SET NotifyProps = JSON_SET(NotifyProps, `+argString+`)
-				WHERE UserId=? AND ChannelId=?`, args...)
+		// unpack the keys and values to pass to MySQL.
+		jsonArgs, jsonSQL := constructMySQLJSONArgs(props)
+		jsonExpr := sq.Expr(fmt.Sprintf("JSON_SET(NotifyProps, %s)", jsonSQL), jsonArgs...)
+
+		// Example: UPDATE ChannelMembers
+		// SET NotifyProps = JSON_SET(NotifyProps, '$.mark_unread', '"yes"' [, ...])
+		// WHERE ...
+		sql, args, err := s.getQueryBuilder().
+			Update("ChannelMembers").
+			Set("NotifyProps", jsonExpr).
+			Where(sq.Eq{
+				"UserId":    userID,
+				"ChannelId": channelID,
+			}).ToSql()
+		if err != nil {
+			return nil, errors.Wrapf(err, "UpdateMemberNotifyProps_Update_MySQL_ToSql channelID=%s and userID=%s", channelID, userID)
 		}
+
+		_, err = tx.Exec(sql, args...)
 	}
 	if err != nil {
 		return nil, errors.Wrapf(err, "failed to update ChannelMember with channelID=%s and userID=%s", channelID, userID)
 	}
 
+	selectSQL, args, err := s.channelMembersForTeamWithSchemeSelectQuery.
+		Where(sq.Eq{
+			"ChannelMembers.ChannelId": channelID,
+			"ChannelMembers.UserId":    userID,
+		}).ToSql()
+	if err != nil {
+		return nil, errors.Wrapf(err, "UpdateMemberNotifyProps_Select_ToSql channelID=%s and userID=%s", channelID, userID)
+	}
+
 	var dbMember channelMemberWithSchemeRoles
-	if err2 := tx.Get(&dbMember, channelMembersForTeamWithSchemeSelectQuery+"WHERE ChannelMembers.ChannelId = ? AND ChannelMembers.UserId = ?", channelID, userID); err2 != nil {
+	if err2 := tx.Get(&dbMember, selectSQL, args...); err2 != nil {
 		if err2 == sql.ErrNoRows {
 			return nil, store.NewErrNotFound("ChannelMember", fmt.Sprintf("channelId=%s, userId=%s", channelID, userID))
 		}
--- a/store/sqlstore/channel_store.go
+++ b/store/sqlstore/channel_store.go
@@ -1801,33 +1801,43 @@ func (s SqlChannelStore) UpdateMultipleMembers(members []*model.ChannelMember) (
 
 	updatedMembers := []*model.ChannelMember{}
 	for _, member := range members {
-		if _, err := transaction.NamedExec(`UPDATE ChannelMembers
-			SET Roles=:Roles,
-			LastViewedAt=:LastViewedAt,
-			MsgCount=:MsgCount,
-			MentionCount=:MentionCount,
-			NotifyProps=:NotifyProps,
-			LastUpdateAt=:LastUpdateAt,
-			SchemeUser=:SchemeUser,
-			SchemeAdmin=:SchemeAdmin,
-			SchemeGuest=:SchemeGuest,
-			MentionCountRoot=:MentionCountRoot,
-			MsgCountRoot=:MsgCountRoot
-			WHERE ChannelId=:ChannelId AND UserId=:UserId`, NewChannelMemberFromModel(member)); err != nil {
+		update := s.getQueryBuilder().Update("ChannelMembers").SetMap(map[string]interface{}{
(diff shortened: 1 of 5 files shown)


store/sqlstore/channel_store.go
```diff
@@ -1558,6 +1558,21 @@ func (s SqlChannelStore) GetDeleted(teamId string, offset int, limit int, userId
 	return channels, nil
 }
 
+var channelMembersForTeamWithSchemeSelectQueryX = sq.Select(
```
agnivade, line 1561: We need to use `s.getQueryBuilder()`. Then we don't need to do the 

```go
if s.DriverName() == model.DatabaseDriverPostgres {
		query = query.PlaceholderFormat(sq.Dollar)
	}
```

business any more :wink:

store/sqlstore/channel_store.go
```diff
@@ -3428,29 +3443,59 @@ func (s SqlChannelStore) SearchGroupChannels(userId, term string) (model.Channel
 	return groupChannels, nil
 }
 
-// TODO: rewrite in squirrel (https://github.com/mattermost/mattermost-server/issues/19336)
-func (s SqlChannelStore) GetMembersByIds(channelId string, userIds []string) (model.ChannelMembers, error) {
-	var dbMembers channelMemberWithSchemeRolesList
+func (s SqlChannelStore) GetMembersByIds(channelID string, userIDs []string) (model.ChannelMembers, error) {
+	if len(userIDs) == 0 {
+		return nil, errors.New("no user IDs were provided")
+	}
```
agnivade, line 3449: This wasn't there previously. Let's remove it to keep the behavior same as earlier. In general, all input validations should be done in the app layer.

store/sqlstore/channel_store.go
```diff
@@ -3428,29 +3443,59 @@ func (s SqlChannelStore) SearchGroupChannels(userId, term string) (model.Channel
...
+	if s.DriverName() == model.DatabaseDriverPostgres {
+		query = query.PlaceholderFormat(sq.Dollar)
+	}
+
+	sql, args, err := query.ToSql()
+	if err != nil {
+		return nil, errors.Wrap(err, "GetMembersByIds_ToSql")
+	}
+
+	var dbMembers channelMemberWithSchemeRolesList
```
agnivade, line 3466: `dbMembers := channelMemberWithSchemeRolesList{}`

store/sqlstore/channel_store.go
```diff
@@ -3428,29 +3443,59 @@ func (s SqlChannelStore) SearchGroupChannels(userId, term string) (model.Channel
...
 	return dbMembers.ToModel(), nil
 }
 
-// TODO: rewrite in squirrel (https://github.com/mattermost/mattermost-server/issues/19336)
-func (s SqlChannelStore) GetMembersByChannelIds(channelIds []string, userId string) (model.ChannelMembers, error) {
-	var dbMembers channelMemberWithSchemeRolesList
+func (s SqlChannelStore) GetMembersByChannelIds(channelIDs []string, userID string) (model.ChannelMembers, error) {
+	if len(channelIDs) == 0 {
+		return nil, errors.New("no channel IDs were provided")
+	}
```
agnivade, line 3478: Same as earlier

store/sqlstore/channel_store.go
```diff
@@ -3428,29 +3443,59 @@ func (s SqlChannelStore) SearchGroupChannels(userId, term string) (model.Channel
...
+
+	sql, args, err := query.ToSql()
+	if err != nil {
+		return nil, errors.Wrap(err, "GetMembersByChannelIds_ToSql")
+	}
 
-	keys, props := MapStringsToQueryParams(channelIds, "Channel")
-	props["UserId"] = userId
+	var dbMembers channelMemberWithSchemeRolesList
```
agnivade, line 3495: Same as above

store/sqlstore/channel_store.go
```diff
@@ -1558,6 +1558,21 @@ func (s SqlChannelStore) GetDeleted(teamId string, offset int, limit int, userId
...
+	"ChannelScheme.DefaultChannelUserRole ChannelSchemeDefaultUserRole",
+	"ChannelScheme.DefaultChannelAdminRole ChannelSchemeDefaultAdminRole",
+).
+	From("ChannelMembers").
+	InnerJoin("Channels ON ChannelMembers.ChannelId = Channels.Id").
+	LeftJoin("Schemes ChannelScheme ON Channels.SchemeId = ChannelScheme.Id").
+	LeftJoin("Teams ON Channels.TeamId = Teams.Id").
+	LeftJoin("Schemes TeamScheme ON Teams.SchemeId = TeamScheme.Id")
+
 var channelMembersForTeamWithSchemeSelectQuery = `
```
agnivade, line 1576: I know this is out of scope, but I am wondering if we can take this opportunity to completely replace this string with the sqlx query everywhere?

agnivade (commented):
Great stuff @noxer ! Just some small comments to take care of.

store/sqlstore/channel_store.go
noxer: The problem here is, that `getQueryBuilder` is part of the `SqlChannelStore`. Would it make sense to move `channelMembersForTeamWithSchemeSelectQueryX` into `SqlChannelStore` and initialize it in `newSqlChannelStore`?

agnivade: Yes!

store/sqlstore/channel_store.go
noxer: This is here to make the method conform to previous behavior. GORP seems to error when an argument is an empty list. One of the tests (I don't remember if for `GetMembersByIds` or `GetMembersByChannelIds`, I can look it up for you) explicitly checks that the method errors when provided an empty list.

store/sqlstore/channel_store.go
noxer: This would slightly change the behavior, since it initializes the list instead of just leaving it `nil`. I'll change it if required.

store/sqlstore/channel_store.go
noxer: See my previous comment.

store/sqlstore/channel_store.go
noxer: See my previous comment.

store/sqlstore/channel_store.go
noxer: Sure, I'll get right to it.

noxer:
Thank you for the review @agnivade, I still have a few questions about some of the comments.

store/sqlstore/channel_store.go
agnivade: Yes, this is intentional. Squirrel throws an error in some cases if we pass nil. An empty slice is better to use.

store/sqlstore/channel_store.go
agnivade: I am wondering if this is due to the nil slice we are passing instead of the empty slice. Can you make the change to empty slice and see if you still get an error? If we do, we might need to change the returned error to pass the test.

store/sqlstore/channel_store.go
noxer: Done, I've migrated all methods using this query to Squirrel.

store/sqlstore/channel_store.go
noxer: The problem is this test (storetest/channel_store.go:6324):
```go
_, nErr = ss.Channel().GetMembersByIds(m1.ChannelId, []string{})
require.Error(t, nErr, "empty user ids - should have failed")
```
it expects `GetMembersByIds` to return an error when userIDs is empty. I could simply remove this test but I suspects it's there for a reason.

store/sqlstore/channel_store.go
noxer: Since there is no test enforcing the behavior, I've removed the check here.

store/sqlstore/channel_store.go
agnivade: Oh dear, the test was actually testing the wrong thing. The old code did 

```go
if _, err := s.GetReplica().Select(&dbMembers, channelMembersForTeamWithSchemeSelectQuery+"WHERE ChannelMembers.ChannelId = :ChannelId AND ChannelMembers.UserId IN "+keys, 
```

which means that when keys was empty, then it would essentially be an incorrect SQL statement, and indeed, when I print the error, that's what I see :)

```
=== RUN   TestChannelStore/MySQL/GetMembersByIds
    channel_store.go:6325: failed to find ChannelMembers with channelId=64zgnfped3rr8c8kjc3epe9yqo and userId in []: Error 1064: You have an error in your SQL syntax; check the manual that corresponds to your MySQL server version for the right syntax to use near ')' at line 19
=== RUN   TestChannelStore/PostgreSQL
=== RUN   TestChannelStore/PostgreSQL/GetMembersByIds
    channel_store.go:6325: failed to find ChannelMembers with channelId=ogi3kwf1zpymzms61qk6wedcda and userId in []: pq: syntax error at or near ")"
```

So we are good to just remove the test. Please apply this patch:

```diff
diff --git a/store/sqlstore/channel_store.go b/store/sqlstore/channel_store.go
index 6346e43ed..72e4c3c4e 100644
--- a/store/sqlstore/channel_store.go
+++ b/store/sqlstore/channel_store.go
@@ -3509,10 +3509,6 @@ func (s SqlChannelStore) SearchGroupChannels(userId, term string) (model.Channel
 }

 func (s SqlChannelStore) GetMembersByIds(channelID string, userIDs []string) (model.ChannelMembers, error) {
-       if len(userIDs) == 0 {
-               return nil, errors.New("no user IDs were provided")
-       }
-
        query := s.channelMembersForTeamWithSchemeSelectQuery.Where(
                sq.Eq{
                        "ChannelMembers.ChannelId": channelID,
@@ -3526,7 +3522,6 @@ func (s SqlChannelStore) GetMembersByIds(channelID string, userIDs []string) (mo
        }

        dbMembers := channelMemberWithSchemeRolesList{}
-
        if err := s.GetReplicaX().Select(&dbMembers, sql, args...); err != nil {
                return nil, errors.Wrapf(err, "failed to find ChannelMembers with channelId=%s and userId in %v", channelID, userIDs)
        }
diff --git a/store/storetest/channel_store.go b/store/storetest/channel_store.go
index 1449ce739..3e7343af7 100644
--- a/store/storetest/channel_store.go
+++ b/store/storetest/channel_store.go
@@ -6321,8 +6321,9 @@ func testChannelStoreGetMembersByIds(t *testing.T, ss store.Store) {
        require.NoError(t, nErr, nErr)
        require.Len(t, members, 2, "return wrong number of results")

-       _, nErr = ss.Channel().GetMembersByIds(m1.ChannelId, []string{})
-       require.Error(t, nErr, "empty user ids - should have failed")
+       members, nErr = ss.Channel().GetMembersByIds(m1.ChannelId, []string{})
+       require.NoError(t, nErr)
+       require.Len(t, members, 0)
 }
```

noxer: I've applied the patch and the tests pass :)

store/sqlstore/channel_store.go
```diff
@@ -1136,15 +1161,22 @@ func (s SqlChannelStore) GetChannelsByUser(userId string, includeDeleted bool, l
...
 	if err != nil {
 		return nil, errors.Wrapf(err, "failed to get ChannelMembers with channelID=%s", channelID)
 	}
 
-	res := make([]string, 0, len(dbMembers))
-	for _, member := range dbMembers.ToModel() {
-		res = append(res, member.UserId)
+	res := make([]string, len(dbMembers))
+	for i, member := range dbMembers.ToModel() {
+		res[i] = member.UserId
```
agnivade, line 1179: Nit: Both are the same, but this is fine as well.

store/sqlstore/channel_store.go
```diff
@@ -1841,31 +1862,57 @@ func (s SqlChannelStore) UpdateMemberNotifyProps(channelID, userID string, props
...
-				SET NotifyProps = JSON_SET(NotifyProps, `+argString+`)
-				WHERE UserId=? AND ChannelId=?`, args...)
+		// unpack the keys and values to pass to MySQL.
+		jsonArgs, jsonSQL := constructMySQLJSONArgs(props)
+		jsonExpr := sq.Expr(fmt.Sprintf("JSON_SET(NotifyProps, %s)", jsonSQL), jsonArgs...)
+
+		// Example: UPDATE ChannelMembers
+		// SET NotifyProps = JSON_SET(NotifyProps, '$.mark_unread', '"yes"' [, ...])
+		// WHERE ...
+		sql, args, eerr := s.getQueryBuilder().
```
agnivade, line 1888: Our convention is to use `err2` to avoid the shadowing errors.

agnivade (approved):
Nicely done @noxer ! Just left some non-blocking comments. But this is good to go.

store/sqlstore/channel_store.go
noxer: This is a bit more efficient, since we know the position of each element beforehand. So we can skip the `cap`-checks in `append`.

store/sqlstore/channel_store.go
noxer: Fixed, thanks.

isacikgoz (approved):
Looks good, nice job @noxer 🚀

store/sqlstore/channel_store.go
agnivade: That's true as well :) I like your low-level thinking!

PR #19686, opened 2022-03-03
MM-42049 - license endpoint not working

pablovelezvidal:
#### Summary
Initial investigation for the saving license issue. Workers not starting/stopping correctly when prevLicense/nextLicense features are validated.

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-42049

#### Release Note
```release-note
NONE
```

PR #19759, opened 2022-03-09
MM-42412 - guard watcher stop/stopped channels with mutex to prevent race condition

pvev:
#### Summary
After fixing the license save flow with PR https://github.com/mattermost/mattermost-server/pull/19686 the watcher stop/stopped channels initialization was moved to the *Watcher.Start() function which allow reuse but it was causing some race condition for the tests. With this PR we guard the channels with mutex.

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-42412

#### Release Note
```release-note
NONE
```


streamer45 (commented):
I am wondering if it wouldn't be simpler to just re-initialize the channels on Stop rather than on Start. I feel that's the safer approach when implementing reusable services.

agnivade:
I want to understand the issue properly first. Can you explain the problem in https://github.com/mattermost/mattermost-server/pull/19686?

pvev:
> I want to understand the issue properly first. Can you explain the problem in #19686?

Ok, I will try to explain as best as I can, so initially there was a problem on the [saveLicense method](https://github.com/mattermost/mattermost-server/pull/19686/files#diff-2dbd62d6fbcb1aa161c4aa3b0892d1afbb75c9cf1d78a09e03bf5b5a5739e5eaR180), that was trying to stop workers that were not initially started and that process was blocking the execution so the api request was not completing (that is why my [initial solution ](https://github.com/mattermost/enterprise/pull/1150) was to make sure  that the workers were actually started before trying to stop them). This problem was reflected in the UI by the uploading license flow being broken. So the solution was placed in the saveLicense method by moving the stop workers/schedulers before setting the new license and after that starting the workers and schedulers again. The other part of the code found in the PR was to make private the initWorker/Schedulers functions because they were not widely used so add them as part of the NewJobServer function. The other part for improving the reuse was to init the channels (stop/stopped) for the watcher/scheduler in the Start method, so everytime the Start method was called a new channel was created. Not sure if I covered everything, maybe @jespino could give more details of the intention of these changes. Thanks.

agnivade:
Understood! I don't think I mind either way re-initializing the channels on stop, or using a mutex on start. So let's go ahead with @streamer45 's suggestions.

streamer45 (approved):
Looks good :+1:

agnivade (approved):
Nicely done!

pvev:
hey @agnivade could you please trigger the e2e tests again? since I changed my github username it is not allowing me to do it.

PR #19765, opened 2022-03-10
[MM-38615] Permalink Previews for DMs/GMs

mylonsuren:
#### Summary

Adds permalink previews for DMs and GMs.

#### Ticket Link

Fixes [MM-38615](https://mattermost.atlassian.net/browse/MM-38615)

#### Release Note

```release-note
NONE
```


app/post.go
```diff
@@ -726,15 +726,23 @@ func (a *App) publishWebsocketEventForPermalinkPost(post *model.Post, message *m
...
+	permalinkPreviewedChannel, err := a.GetChannel(previewedPost.ChannelId)
+	if err != nil {
+		if err.StatusCode == http.StatusNotFound {
+			mlog.Warn("channel containing permalinked post not found", mlog.String("referenced_channel_id", previewedPost.ChannelId))
+			return false, nil
+		}
+		return false, err
+	}
+
+	permalinkPreviewedPost := post.GetPreviewPost()
```
BenCookie95, line 738: Wondering how this could impact performance with the extra GetChannel and GetPreviewPost calls? This ticket already had some concerns https://mattermost.atlassian.net/browse/MM-42810 around this function

mylonsuren: @BenCookie95 This actually reduces the number of `GetPreviewPost` and `GetChannel` calls by doing them here instead of in `SanitizePostMetadataForUser` in `post_metadata.go`. Before, we were making these calls for each user in a channel before publishing the message, but now these calls are made once here and their values are passed to the new function `sanitizePostMetadataForUserAndChannel`. I think this should help with some of the performance issues in [MM-42810](https://mattermost.atlassian.net/browse/MM-42810).

BenCookie95 (approved):

app/post_metadata.go
```diff
@@ -533,9 +544,14 @@ func (a *App) getLinkMetadata(requestURL string, timestamp int64, isNewPost bool
 			return nil, nil, nil, appErr
 		}
 
-		referencedTeam, appErr := a.GetTeam(referencedChannel.TeamId)
-		if appErr != nil {
-			return nil, nil, nil, appErr
+		referencedTeam := &model.Team{}
```
mkraft, line 547:
```suggestion
		var referencedTeam *model.Team
```

app/post_metadata.go
```diff
@@ -533,9 +544,14 @@ func (a *App) getLinkMetadata(requestURL string, timestamp int64, isNewPost bool
...
+		if referencedChannel.Type == model.ChannelTypeDirect || referencedChannel.Type == model.ChannelTypeGroup {
+			referencedTeam.Name = ""
```
mkraft, line 549:
```suggestion
			referencedTeam = &model.Team{}
```

app/post_metadata.go
```diff
@@ -533,9 +544,14 @@ func (a *App) getLinkMetadata(requestURL string, timestamp int64, isNewPost bool
...
+		} else {
+			referencedTeam, appErr = a.GetTeam(referencedChannel.TeamId)
```
mkraft, line 551: Are we doing the required authorization checks to see if the recipient is allowed to see the team?

app/post_metadata.go
mylonsuren: Done :+1:

app/post_metadata.go
mylonsuren: Done :+1:

app/post_metadata.go
mylonsuren: Yes those checks are still done as before to verify that the recipient is allowed to view the channel and team. Before the message is broadcast to each user, the metadata is removed by `sanitizePostMetadataForUserAndChannel` here: https://github.com/mattermost/mattermost-server/blob/63dfd61b96d7d719b2a97a4999813ebfaab1e8f9/app/post_metadata.go#L166

which is called from `publishWebsocketEventForPermalinkPost` here: https://github.com/mattermost/mattermost-server/blob/63dfd61b96d7d719b2a97a4999813ebfaab1e8f9/app/post.go#L752

The only modification made to this is that we now get the previewed post and channel once before sanitizing the post for each user, which should help with performance.

mkraft (approved):

furqanmlk (approved):
Updated webapp [PR](https://github.com/mattermost/mattermost-webapp/pull/9992#pullrequestreview-935575771)

PR #19771, opened 2022-03-10
Mm 42465

neallred:
#### Summary
When plugins are disabled, the onboarding/complete endpoint should still succeed because it represents onboarding completion and not just plugins status. A related webapp PR (TBD) takes care of hiding the plugins selection screen if plugins are disabled.

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-42465

#### Release Note
No release note because this does not affect cloud and it should be cherry-picked to the latest release candidate branch before its release.
```release-note
NONE
```

### DIFF
--- a/api4/system_test.go
+++ b/api4/system_test.go
@@ -914,4 +914,20 @@ func TestCompleteOnboarding(t *testing.T) {
 		}
 
 	})
+
+	t.Run("as a system admin when plugins are disabled", func(t *testing.T) {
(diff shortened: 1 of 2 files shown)


hanzei (approved):
Nit pick: I would like to see the code to store the state be DRY, but that's not blocking.

Thanks for adding a test

marianunez (approved):

PR #19781, opened 2022-03-14
Adds support for clients subscribing to websocket subjects.

mkraft:
#### Summary

Tech. spec: https://gist.github.com/mkraft/f3ec8c933f37f103fe7573ad081c55a2

#### Related

https://github.com/mattermost/mattermost-webapp/pull/10006

#### Ticket Link

n/a

#### Release Note

```release-note
Added support for WebSocket subscriptions.
```


app/web_conn.go
```diff
@@ -710,77 +725,21 @@ func (wc *WebConn) shouldSendEventToGuest(msg *model.WebSocketEvent) bool {
...
-		return false
-	}
-
-	// If the event contains sanitized data, only send to users that don't have permission to
-	// see sensitive data. Prevents admin clients from receiving events with bad data
-	var hasReadPrivateDataPermission *bool
-	if msg.GetBroadcast().ContainsSanitizedData {
-		hasReadPrivateDataPermission = model.NewBool(wc.App.RolesGrantPermission(wc.GetSession().GetUserRoles(), model.PermissionManageSystem.Id))
-
-		if *hasReadPrivateDataPermission {
```
mkraft, line 724: Below is just a refactor.

agnivade:
@streamer45 - To add some context behind this, Martin wanted an efficient way to implement the new activity feed that's coming. We didn't want to send every single update down to the client unless the client is on the activity page. Hence the concept of adding subscriptions to the server. This follows from a similar requirement we had for status updates where the client needed to subscribe for status updates only for users that are currently visible in the UI.

agnivade:
@mkraft - Looks very neatly done. Just a small question - wondering if we need to add the `isSubscribed` logic to the web_hub side? I was thinking that it should be possible by just storing a map of subscription ids per web_conn struct.

I'd prefer to keep web_hub as much detached from application logic as possible and only keep it responsible for routing events to webconns.

Something like this perhaps:

```diff
 func (wc *WebConn) Subscribe(id model.WebsocketSubjectID) {
-       hub := wc.App.GetHubForUserId(wc.UserId)
-       hub.subscribe(id, wc)
+       wc.subscriptions[id] = true
 }

 func (wc *WebConn) Unsubscribe(id model.WebsocketSubjectID) {
-       hub := wc.App.GetHubForUserId(wc.UserId)
-       hub.unsubscribe(id, wc)
+       delete(wc.subscriptions, id)
 }

 func (wc *WebConn) IsSubscribed(id model.WebsocketSubjectID) bool {
-       hub := wc.App.GetHubForUserId(wc.UserId)
-       return hub.isSubscribed(id, wc)
+       return wc.subscriptions[id]
 }
```

With this, the hub is totally unaware of what's happening and there's no need to add anything to the connection index. Re: multiple user sessions, I think those should be handled by the individual webconns. Different sessions might/might not have the component loaded. So it'll be good to have the subscription at a webconn level.

Any thoughts?

mkraft:
> @mkraft - Looks very neatly done. Just a small question - wondering if we need to add the `isSubscribed` logic to the web_hub side? I was thinking that it should be possible by just storing a map of subscription ids per web_conn struct.
> 
> I'd prefer to keep web_hub as much detached from application logic as possible and only keep it responsible for routing events to webconns.
> 
> Something like this perhaps:
> 
> ```diff
>  func (wc *WebConn) Subscribe(id model.WebsocketSubjectID) {
> -       hub := wc.App.GetHubForUserId(wc.UserId)
> -       hub.subscribe(id, wc)
> +       wc.subscriptions[id] = true
>  }
> 
>  func (wc *WebConn) Unsubscribe(id model.WebsocketSubjectID) {
> -       hub := wc.App.GetHubForUserId(wc.UserId)
> -       hub.unsubscribe(id, wc)
> +       delete(wc.subscriptions, id)
>  }
> 
>  func (wc *WebConn) IsSubscribed(id model.WebsocketSubjectID) bool {
> -       hub := wc.App.GetHubForUserId(wc.UserId)
> -       return hub.isSubscribed(id, wc)
> +       return wc.subscriptions[id]
>  }
> ```
> 
> With this, the hub is totally unaware of what's happening and there's no need to add anything to the connection index. Re: multiple user sessions, I think those should be handled by the individual webconns. Different sessions might/might not have the component loaded. So it'll be good to have the subscription at a webconn level.
> 
> Any thoughts?

Your proposal made me realize that I had forgotten to add an optimization that was the whole point of the structure I had chosen: avoiding iterating all web connections in cases where the event is not targeting a specific user.

I've just added a commit containing the optimiztion: https://github.com/mattermost/mattermost-server/pull/19781/commits/df5495b762909f447ec8c5fa0281e504d9079428

If we don't think that optimization is worthwhile, then absolutely the `*WebConn` instances could hold all of the knowledge of their subscriptions and we could avoid adding the `bySubscription map[model.WebsocketSubjectID]map[*WebConn]bool` field to the `hubConnectionIndex`.

Thoughts?

agnivade:
> Your proposal made me realize that I had forgotten to add an optimization that was the whole point of the structure I had chosen: avoiding iterating all web connections in cases where the event is not targeting a specific user.

Aha! There you go, now all of this makes sense.

I'd love to see a load-test comparison of both approaches. But in general, I am in favor of this. Reducing the iteration cost is certainly going to help at scale.

cc @streamer45 for your thoughts.

Nice work, and it's great to see some new work done in the web_conn/web_hub. Hopefully the comments that I added in the `hubConnectionIndex` helped :)

mkraft:
@agnivade I'm just wondering if this is going to give us correct results. You can lookup connections by user id in a hub because hubs are sharded by user id, but subjects are not. Wouldn't we would only get a subset of subscriber connections (the intersection of users in the hub and subject subscribers)?

agnivade:
This will return subscriber connections per hub, and the web_hub broadcast is done to all hubs. So we should be able to send to all subscribed connections in all hubs?

Btw, there's an important point that needs to be addressed: the connection index is deliberately kept as a local variable for a reason. Because concurrent access will create race conditions. The way to achieve this would be to have channels to send signals to subscribe/unsubscribe, and check for `isSubscribed` using a return channel.

streamer45:
The proposed approach looks good to me. Also very nice refactor on the conditional logic. Performance wise I don't think there would be a huge difference between the two ways of storing/retrieving subscriptions so I'd be leaning towards Agniva's suggestion as it feels logically simpler overall.

agnivade:
Sounds good. Thanks Claudio.

@mkraft - Can we then go ahead with the proposed approach here https://github.com/mattermost/mattermost-server/pull/19781#issuecomment-1068785779? Since websockets are probably the most complicated and sensitive part of the codebase, I'd be leaning towards simplicity first as well. We can always take the next step and add a new index to skip iterating unnecessary connections as a next step.

mkraft:
> Sounds good. Thanks Claudio.
> 
> @mkraft - Can we then go ahead with the proposed approach here [#19781 (comment)](https://github.com/mattermost/mattermost-server/pull/19781#issuecomment-1068785779)? Since websockets are probably the most complicated and sensitive part of the codebase, I'd be leaning towards simplicity first as well. We can always take the next step and add a new index to skip iterating unnecessary connections as a next step.

Just to be 100% clear, if the future Calls firehose of ws events (for example) is publishing join, leave, start, stop events and only 5% of users are subscribed to "calls" events, the websocket backend would still need to check every connection per publish. You have a better sense than me to know if that would be a concern, I'm cool to go with whichever strategy.

agnivade:
Correct. It's not a primary concern because the rejection of that event happens in the goroutine for that connection. It does not block the primary web_hub goroutine. The primary goroutine just pushes the event down the buffered channel and moves on. Yes, there is a non-zero cost obviously, but something we can work on later.

Lastly, I think it goes without saying that such an important change needs to be load-tested and verified as well. :)

app/web_conn.go
```diff
@@ -380,6 +382,18 @@ func (wc *WebConn) readPump() {
...
+	wc.subscriptions[id] = true
+}
+
+func (wc *WebConn) Unsubscribe(id model.WebsocketSubjectID) {
+	delete(wc.subscriptions, id)
+}
+
+func (wc *WebConn) IsSubscribed(id model.WebsocketSubjectID) bool {
+	return wc.subscriptions[id]
+}
```
agnivade, line 395: Now these need to be protected with a mutex as well

model/websocket_message.go
```diff
@@ -92,7 +97,8 @@ type WebsocketBroadcast struct {
 	ContainsSensitiveData bool            `json:"-"`
 	// ReliableClusterSend indicates whether or not the message should
 	// be sent through the cluster using the reliable, TCP backed channel.
-	ReliableClusterSend bool `json:"-"`
+	ReliableClusterSend bool               `json:"-"`
+	SubjectID           WebsocketSubjectID `json:"-"`
```
agnivade, line 101: Curious to know the reason behind not showing the result in the json output. This will prevent it from being sent in cluster messages so the subject will be entirely lost when broadcasted to other nodes.

wsapi/subscription.go
```diff
@@ -0,0 +1,57 @@
...
+
+	return nil, nil
+}
+
+func subjectIDFromRequest(req *model.WebSocketRequest) (model.WebsocketSubjectID, *model.AppError) {
+	const paramKey = "subscription_id"
+
+	paramVal, has := req.Data[paramKey]
+	if !has {
+		mlog.Debug("missing JSON field", mlog.String("key", paramKey))
```
agnivade, line 46: This will generate double logs since the error will also get logged at the upper layer. Let's remove this and add any custom info to the error.

wsapi/subscription.go
```diff
@@ -0,0 +1,57 @@
...
+		return "", NewInvalidWebSocketParamError(req.Action, paramKey)
+	}
+
+	id := model.WebsocketSubjectID(fmt.Sprintf("%s", paramVal))
+	if !id.IsValid() {
+		mlog.Debug("invalid JSON field", mlog.String("key", paramKey), mlog.Any("value", paramVal))
```
agnivade, line 52: Same as in L46

wsapi/subscription.go
```diff
@@ -0,0 +1,57 @@
...
+func subjectIDFromRequest(req *model.WebSocketRequest) (model.WebsocketSubjectID, *model.AppError) {
+	const paramKey = "subscription_id"
+
+	paramVal, has := req.Data[paramKey]
+	if !has {
+		mlog.Debug("missing JSON field", mlog.String("key", paramKey))
+		return "", NewInvalidWebSocketParamError(req.Action, paramKey)
+	}
+
+	id := model.WebsocketSubjectID(fmt.Sprintf("%s", paramVal))
```
agnivade, line 50: Instead of coercing to a string, I'd suggest to do a `paramVal.(string)` and check for `ok`.

model/websocket_message.go
```diff
@@ -75,6 +75,11 @@ const (
 	WebsocketEventThreadFollowChanged                 = "thread_follow_changed"
 	WebsocketEventThreadReadChanged                   = "thread_read_changed"
 	WebsocketFirstAdminVisitMarketplaceStatusReceived = "first_admin_visit_marketplace_status_received"
+	WebsocketEventSubscribe                           = "subscribe"
+	WebsocketEventUnsubscribe                         = "unsubscribe"
+
+	WebsocketSubjectActivityFeed WebsocketSubjectID = "activity_feed"
+	WebsocketSubjectActiveCalls  WebsocketSubjectID = "active_calls"
```
agnivade, line 82: Let's add a comment here to add any new subject to the `IsValid` method as well to keep both in sync.

app/websocket_event_conditions.go
```diff
@@ -0,0 +1,102 @@
...
+
+func wsConditionAuth(wc *WebConn, _ *model.WebSocketEvent) bool {
+	return wc.IsAuthenticated()
+}
+
+func wsConditionSubject(wc *WebConn, evt *model.WebSocketEvent) bool {
+	if subjectID := evt.GetBroadcast().SubjectID; subjectID != "" {
+		if !wc.IsSubscribed(subjectID) {
+			return false
+		}
```
agnivade, line 22: Can we just do `return wc.IsSubscribed(subjectID)`?

agnivade (commented):
Looking great Martin. Just a couple of things to take care of.

model/websocket_message.go
mkraft: Ok, I'm totally unfamiliar with that cluster sending behaviour. My purpose was to avoid responding with that new field in the WebSocket message response payload since clients don't need to be aware of the subject ID associated to messages. Seems these two concerns are coupled?

wsapi/subscription.go
mkraft: Actually I think we can do convert directly to the target type.

model/websocket_message.go
mkraft: I removed these to support dynamic subject ids.

app/web_conn.go
```diff
@@ -102,13 +102,15 @@ type WebConn struct {
...
-	pluginPosted chan pluginWSPostedHook
+	reuseCount    int
+	sessionToken  atomic.Value
+	session       atomic.Value
+	connectionID  atomic.Value
+	endWritePump  chan struct{}
+	pumpFinished  chan struct{}
+	pluginPosted  chan pluginWSPostedHook
+	subscriptions map[model.WebsocketSubjectID]bool
+	mu            sync.Mutex
```
agnivade, line 113: Let's use an `RWMutex`

app/web_conn.go
```diff
@@ -380,6 +383,24 @@ func (wc *WebConn) readPump() {
...
+
+func (wc *WebConn) Unsubscribe(id model.WebsocketSubjectID) {
+	wc.mu.Lock()
+	defer wc.mu.Unlock()
+	delete(wc.subscriptions, id)
+}
+
+func (wc *WebConn) IsSubscribed(id model.WebsocketSubjectID) bool {
+	wc.mu.Lock()
+	defer wc.mu.Unlock()
```
agnivade, line 400: We can use R(Lock/Unlock) here.

model/websocket_subject.go
```diff
@@ -0,0 +1,26 @@
...
+// An example of which is "insights", however, this could support more complex strings like "channels/wds7jxtetjgjue9yca5i5r1cjc".
+type WebsocketSubjectID string
+
+func (si WebsocketSubjectID) IsValid() bool {
+	if si == "" {
+		return false
+	}
+	patterns := []*regexp.Regexp{
+		regexp.MustCompile(`^activity_feed$`),
+		regexp.MustCompile(`channels/[a-z,0-9]{26}/typing`),
```
agnivade, line 18: Let's pre-compile the regexes outside the method. Regex compilation is costly and is mentioned in our performance guide: https://developers.mattermost.com/contribute/server/style-guide/#performance-sensitive-areas.

mkraft: Yeah I have a note for myself to do that.

app/web_conn.go
```diff
@@ -380,6 +383,24 @@ func (wc *WebConn) readPump() {
 	}
 }
 
+func (wc *WebConn) Subscribe(id model.WebsocketSubjectID) {
+	wc.mu.RLock()
+	defer wc.mu.RUnlock()
```
agnivade, line 388: Btw, this is modifying the map, so we needed Lock/Unlock here :) Only the read case should use R(Lock/Unlock). Anyways, no longer needed now.

agnivade (commented):
Only removing the mutex remains.

mkraft:
I'm running a load test comparison.

mkraft:
@agnivade @streamer45 Were there any other pending questions or concerns?

streamer45 (commented):
Looks great overall, my only question is why was the mutex added and removed?

streamer45:
> > Looks great overall, my only question is why was the mutex added and removed?
> 
> I realized that it wasn't needed and @agnivade confirmed.

How are we making sure that, for example, a call to `unsubscribe` doesn't happen concurrently with sending an event to the same connection (running `wsConditionSubject`)?

agnivade (approved):
Code looks good. Only thing pending is the load-test.

mkraft:
> Code looks good. Only thing pending is the load-test.

I thought we were in agreement that the load tests will occur when a feature using this is submitted?

agnivade:
> I thought we were in agreement that the load tests will occur when a feature using this is submitted?

I think there was some confusion regarding that. I'd suggest to wait merging this until we have the feature implementation ready, and then port it to the load-tests. Run the load-test and verify everything looks good.

> How are we making sure that, for example, a call to unsubscribe doesn't happen concurrently with sending an event to the same connection (running wsConditionSubject)?

Oooh good catch! I think I missed this ..

mkraft:
Putting this PR back in WIP until either Activity Feed or Typing or another feature is ready to use it at which point I'll loadtest and get a new round of reviews.

PR #19783, opened 2022-03-14
[GraphQL POC] Omit self from pretty display name

larkox:
#### Summary
When we are creating the pretty display name for a DM or GM, we omit the requester name, unless it is a DM to himself.

For the POC this is enough but:
- The names are not ordered based on the locale. This can be done in the client, but then we have to send the names in a format that the client can easily separate (we could separate by commas, but if the name has commas, then it is not possible...).
- We probably want to add the "(you)" string to the pretty display name of a DM to himself.

#### Ticket Link
None

#### Release Note
```release-note
NONE
```

### DIFF
--- a/api4/resolver_channel_test.go
+++ b/api4/resolver_channel_test.go
@@ -374,12 +374,14 @@ func TestGetPrettyDNForUsers(t *testing.T) {
 	t.Run("nickname_full_name", func(t *testing.T) {
 		users := []*model.User{
 			{
+				Id:        "user1",
 				Nickname:  "nick1",
 				Username:  "user1",
 				FirstName: "first1",
 				LastName:  "last1",
 			},
 			{
+				Id:        "user2",
 				Nickname:  "nick2",
 				Username:  "user2",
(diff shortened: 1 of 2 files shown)


api4/resolver_channel_test.go
```diff
@@ -386,7 +386,7 @@ func TestGetPrettyDNForUsers(t *testing.T) {
 				LastName:  "last2",
 			},
 		}
-		assert.Equal(t, "nick1, nick2", getPrettyDNForUsers("nickname_full_name", users))
+		assert.Equal(t, "nick2", getPrettyDNForUsers("nickname_full_name", users, "user1"))
```
agnivade, line 389: We need to pass the userId, not username :)

agnivade (approved):
Looking great!

M-ZubairAhmed (approved):

larkox:
@agnivade Tests fixed. Not sure if this has to go through QA testing of any sort, or can be merged.

PR #19826, opened 2022-03-23
[MM-40962-2] - Fix unnecessary logging in license expiration check

AGMETEOR:
#### Summary
This PR is a redo of https://github.com/mattermost/mattermost-server/pull/19653 and does the following
- Puts back the log that was removed after [this suggestion](https://github.com/mattermost/mattermost-server/pull/19653#issuecomment-1073680601)
- Skips check for TE
- Does a refactor to generate one renewal link for all admins when sending a "SendRemoveExpiredLicenseEmail". This means that in the case of an error in the renewal JWT token stored, we don't log this error for all admins present on the system

#### Ticket Link
[MM-40962](https://mattermost.atlassian.net/browse/MM-40962)

#### Release Note

```release-note
NONE
```


app/server.go
```diff
@@ -1721,7 +1721,15 @@ func (s *Server) sendLicenseUpForRenewalEmail(users map[string]*model.User, lice
 }
 
 func (s *Server) doLicenseExpirationCheck() {
-	s.LoadLicense()
```
AGMETEOR, line 1724: Hey @agnivade do you have any idea why this was added in the first place?

app/server.go
```diff
@@ -1745,40 +1753,41 @@ func (s *Server) doLicenseExpirationCheck() {
...
+			s.Go(func() {
+				if err := s.SendRemoveExpiredLicenseEmail(user.Email, renewalLink, user.Locale, *s.Config().ServiceSettings.SiteURL); err != nil {
+					mlog.Error("Error while sending the license expired email.", mlog.String("user_email", user.Email), mlog.Err(err))
+				}
+			})
+		}
 	}
 
-	//remove the license
-	s.RemoveLicense()
```
AGMETEOR, line 1772: Hey @agnivade , this too

app/server.go
AGMETEOR: This produces a weird effect due to a listener that was added, the listener then sends a websocket to the client with a sanitized license, what happens then is that, on the license page in the console, we have some details missing

app/server.go
```diff
@@ -1745,40 +1753,41 @@ func (s *Server) doLicenseExpirationCheck() {
...
 }
 
 // SendRemoveExpiredLicenseEmail formats an email and uses the email service to send the email to user with link pointing to CWS
 // to renew the user license
-func (s *Server) SendRemoveExpiredLicenseEmail(email string, locale, siteURL string) *model.AppError {
-	renewalLink, _, err := s.GenerateLicenseRenewalLink()
-	if err != nil {
-		return err
-	}
```
marianunez, line 1781: What do we gain extracting this?

app/license.go
```diff
@@ -379,7 +379,10 @@ func (s *Server) GenerateRenewalToken(expiration time.Duration) (string, *model.
 
 	currentToken, _ := s.Store.System().GetByName(model.SystemLicenseRenewalToken)
 	if currentToken != nil {
-		tokenIsValid, _ := s.renewalTokenValid(currentToken.Value, license.Customer.Email)
+		tokenIsValid, err := s.renewalTokenValid(currentToken.Value, license.Customer.Email)
+		if err != nil {
+			mlog.Warn("error checking license renewal token validation", mlog.Err(err))
+		}
```
marianunez, line 385: I guess the question here comes down to:
1. Why do we store the token in the System Store? It seems that we want to generate a new token for a link on demand and here if there is an error validating - we ignore and generate a new one anyways.

app/server.go
agnivade: I don't think I am aware of the context behind this. I think this part is owned by your team so I'll leave it to your best judgement.

app/license.go
agnivade: I don't mind either ways. Technically, it seems fine to return an existing token if it hasn't expired. I guess the main question we had was "Why are we getting a spam of these token renewal requests during load-tests"? Do we know why is that?

marianunez: +1, this should only be called if we are sending the license expired email or in the front end calling the endpoint to get renewal link. Are any of these actions happening in load testing?

AGMETEOR: @agnivade the spam of tokens was due an attempt to generate a token for each admin when sending the email. So if multiple admins exist then we get several of those logs.

@marianunez my thought also was we don’t need to store the token in the db. We can always have one generated on the fly

app/server.go
AGMETEOR: In case of logs having in that function call, it doesn’t log for each admin

app/license.go
marianunez: >  the spam of tokens was due an attempt to generate a token for each admin when sending the email. So if multiple admins exist then we get several of those logs.

Why is the load test triggering these emails?

isacikgoz: I think it was getting triggered by `ExpiryNotify` job.

AGMETEOR: @isacikgoz not sure I understand how it does. We trigger this job as a recurring task

agnivade: We need to get to the bottom of this as this is blocking the release. @isacikgoz - The question I had is do we see the logs being repeated or is it a one time spam of logs and it stops? I am wondering if we can create a separate Terraform instance and hand it off to @AGMETEOR to debug it further?

I guess you could not reproduce this locally right?

AGMETEOR: Yes, I couldn't reproduce it locally.

isacikgoz: @AGMETEOR actually you can first try the load test agent on your local instance, maybe we wouldn't need to go over the aws path? WDYT?

agnivade: Oh, can we reproduce this with a local load test? Then it should be fairly easy to debug.

app/server.go
marianunez: Same here https://github.com/mattermost/mattermost-server/pull/19826#discussion_r836438453

app/server.go
agnivade: @AGMETEOR - do you want to bring this back or are you confident this is good to be removed?

app/license.go
agnivade: @AGMETEOR / @isacikgoz -  Do we want to just move on with the review of the PR given that we weren't able to reproduce this a second time?

AGMETEOR: Yes.

app/server.go
```diff
@@ -1721,7 +1721,15 @@ func (s *Server) sendLicenseUpForRenewalEmail(users map[string]*model.User, lice
 }
 
 func (s *Server) doLicenseExpirationCheck() {
-	s.LoadLicense()
+
+	// This takes care of a rare edge case reported here https://mattermost.atlassian.net/browse/MM-40962
+	// To reproduce that case locally, attach a license to a server that was started with enterprise enabled
+	// Then restart using BUILD_ENTERPRISE=false make restart-server to enter Team Edition
+	if model.BuildEnterpriseReady != "true" {
+		mlog.Debug("Skipping license expiraion check because no license is expected on Team Edition")
```
agnivade, line 1729: expiration

app/server.go
```diff
@@ -1745,40 +1753,41 @@ func (s *Server) doLicenseExpirationCheck() {
...
 		return
 	}
 
-	//send email to admin(s)
-	for _, user := range users {
-		user := user
-		if user.Email == "" {
-			mlog.Error("Invalid system admin email.", mlog.String("user_email", user.Email))
-			continue
+	if license.IsPastGracePeriod() {
```
agnivade, line 1760: Hmm .. any reason to invert the check? This creates more indented code. In Go, we prefer early returns.

app/server.go
AGMETEOR: I am 1/5 removing it

agnivade (changes requested):

app/server.go
```diff
@@ -1745,13 +1754,20 @@ func (s *Server) doLicenseExpirationCheck() {
 		if appErr != nil {
 			mlog.Debug(appErr.Error())
 		}
+		return
 	}
 
-	if !license.IsPastGracePeriod() {
+	if license.IsPastGracePeriod() {
```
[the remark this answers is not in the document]
AGMETEOR: Screenshot 2022-03-30 at 18 43 24
I wonder why github is not reflecting that commit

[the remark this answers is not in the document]
AGMETEOR: my bad I had not pushed

app/license.go
marianunez, line 385: Looking more into the errors reported in the load testing `token is expired by`, this is actually returned from the [JWT library here](https://github.com/mattermost/mattermost-server/blob/baf9f69c273c67d8c1569c6bab9376d21971d8cf/vendor/github.com/dgrijalva/jwt-go/claims.go#L40). Which means that we are reporting an error for the token being expired which we know it will eventually happen given that it has a time period that it is valid and we are saving it for some reason. 

Again, I don't think there is any value on logging a warning for something that is expected and we should just be generating a new token each time - like I mentioned above. 

I'm 4/5 that we should **not** keep this error as is. We should either:

1. Stop storing the token in the System Store and just generate the token on demand. This would avoid storing something that will expire and produce these unnecessary errors/checks. 
2. Surface the error type [ValidationErrorExpired](https://github.com/mattermost/mattermost-server/blob/baf9f69c273c67d8c1569c6bab9376d21971d8cf/vendor/github.com/dgrijalva/jwt-go/claims.go#L41) from our `renewalTokenValid` and check here so we can ignore

marianunez (changes requested):

app/license.go
agnivade: Fantastic analysis Maria! Yeah, fully agree we can ignore the expired errors.

agnivade (approved):
Good to go after Maria's comments are addressed

app/license.go
AGMETEOR: @marianunez went with option 1

marianunez (approved):
Thanks @AGMETEOR!

AGMETEOR:
Got 2 dev reviews, so I am going to go ahead and merge

marianunez:
> Got 2 dev reviews, so I am going to go ahead and merge

We should also do a quick smoke test on the renew now link and email to make sure there is no regression from these. O/5 if we want to do this post merge cc/@steve.mudie

AGMETEOR:
Hey @stevemudie . Post merge?

stevemudie:
> Hey @stevemudie . Post merge?

Yup, sounds good.

PR #19841, opened 2022-03-25
MM-41260: Revamp ES/Bleve batching logic

agnivade:
The older method used to reply completely on timestamps
to take batches of items in a timestamp range and then
just incrementing the timestamp. This led to handling
edge-cases such as more items than the batch count, all
having the same timestamp.

Additionally, relying on timestamp as the page cursor
meant that indexing was not very efficient if you had
several items spread out across large spans of time.

To get away from all of that we use a proper cursor-based
approach consisting of createAt+Id. With this, we move
completely to a constant page size where we can fetch
a given number of objects irrespective of when they
were created. This makes indexing much more faster and
efficient.

https://mattermost.atlassian.net/browse/MM-41260

```release-note
Elasticsearch and Bleve indexing have been revamped to be much
more efficient and faster. The config parameter BulkIndexingTimeWindowSeconds
for both elasticsearch and bleve is now deprecated and no longer used.
A new config parameter called BatchSize has been introduced instead.
This parameter controls the number of objects that
can be indexed in a single batch. This makes things
more efficient and maintains a constant workload.

The following should go in upgrade notes:

This PR introduces schema changes in the form of a new index.
The following notes the test results for the schema changes:
MySQL 7M Posts - ~17s (Instance: db.r5.xlarge)
MySQL 9M Posts - 2min 12s (Instance: db.r5.large)
Postgres 7M Posts - ~9s  (Instance: db.r5.xlarge)

For customers wanting a zero downtime upgrade, they are welcome
to apply this index prior to doing the upgrade. This is fully backwards
compatible and will not acquire any table lock or affect any existing
operations on the table.

For MySQL: `CREATE INDEX idx_posts_create_at_id on Posts(CreateAt, Id) LOCK=NONE;`
For Postgres: `CREATE INDEX CONCURRENTLY IF NOT EXISTS idx_posts_create_at_id on posts(createat, id);`
```

### DIFF
--- a/model/config.go
+++ b/model/config.go
@@ -2475,7 +2475,7 @@ type ElasticsearchSettings struct {
 	PostsAggregatorJobStartTime   *string `access:"environment_elasticsearch,write_restrictable,cloud_restrictable"` // telemetry: none
 	IndexPrefix                   *string `access:"environment_elasticsearch,write_restrictable,cloud_restrictable"`
 	LiveIndexingBatchSize         *int    `access:"environment_elasticsearch,write_restrictable,cloud_restrictable"`
-	BulkIndexingTimeWindowSeconds *int    `json:",omitempty"`
+	BulkIndexingTimeWindowSeconds *int    `json:",omitempty"` // telemetry: none
 	BatchSize                     *int    `access:"environment_elasticsearch,write_restrictable,cloud_restrictable"`
 	RequestTimeoutSeconds         *int    `access:"environment_elasticsearch,write_restrictable,cloud_restrictable"`
 	SkipTLSVerification           *bool   `access:"environment_elasticsearch,write_restrictable,cloud_restrictable"`
@@ -2573,7 +2573,7 @@ type BleveSettings struct {
 	EnableIndexing                *bool   `access:"experimental_bleve"`
 	EnableSearching               *bool   `access:"experimental_bleve"`
 	EnableAutocomplete            *bool   `access:"experimental_bleve"`
-	BulkIndexingTimeWindowSeconds *int    `json:",omitempty"`
+	BulkIndexingTimeWindowSeconds *int    `json:",omitempty"` // telemetry: none
 	BatchSize                     *int    `access:"experimental_bleve"`
 }
 
--- a/model/config.go
+++ b/model/config.go
@@ -184,24 +184,24 @@ const (
 
 	TeamSettingsDefaultTeamText = "default"
 
-	ElasticsearchSettingsDefaultConnectionURL                 = "http://localhost:9200"
-	ElasticsearchSettingsDefaultUsername                      = "elastic"
-	ElasticsearchSettingsDefaultPassword                      = "changeme"
-	ElasticsearchSettingsDefaultPostIndexReplicas             = 1
-	ElasticsearchSettingsDefaultPostIndexShards               = 1
-	ElasticsearchSettingsDefaultChannelIndexReplicas          = 1
-	ElasticsearchSettingsDefaultChannelIndexShards            = 1
-	ElasticsearchSettingsDefaultUserIndexReplicas             = 1
-	ElasticsearchSettingsDefaultUserIndexShards               = 1
-	ElasticsearchSettingsDefaultAggregatePostsAfterDays       = 365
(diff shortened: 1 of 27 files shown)


agnivade:
Actually, removing the config field might not be such a good idea. Better to make it omitempty. I'll send another update.

store/storetest/channel_store.go
```diff
@@ -7633,23 +7632,20 @@ func testChannelStoreGetChannelsBatchForIndexing(t *testing.T, ss store.Store) {
...
 	require.NoError(t, nErr)
 
-	endTime := c6.CreateAt
-
 	// First and last channel should be outside the range
-	channels, err := ss.Channel().GetChannelsBatchForIndexing(startTime, endTime, 1000)
+	channels, err := ss.Channel().GetChannelsBatchForIndexing(c1.CreateAt, "", 4)
 	assert.NoError(t, err)
-	assert.ElementsMatch(t, []*model.Channel{c2, c3, c4, c5}, channels)
+	assert.Len(t, channels, 4)
```
isacikgoz, line 7638: Shall we do an assertion like this to make sure we ignore the `c1` and `c6`? Maybe we can use this assertion on cases below?
```suggestion
	assert.NotContains(t, channels, []*model.Channel{c1, c6})
	assert.Len(t, channels, 4)
```

model/config.go
```diff
@@ -184,24 +184,24 @@ const (
...
+	ElasticsearchSettingsDefaultChannelIndexReplicas        = 1
+	ElasticsearchSettingsDefaultChannelIndexShards          = 1
+	ElasticsearchSettingsDefaultUserIndexReplicas           = 1
+	ElasticsearchSettingsDefaultUserIndexShards             = 1
+	ElasticsearchSettingsDefaultAggregatePostsAfterDays     = 365
+	ElasticsearchSettingsDefaultPostsAggregatorJobStartTime = "03:00"
+	ElasticsearchSettingsDefaultIndexPrefix                 = ""
+	ElasticsearchSettingsDefaultLiveIndexingBatchSize       = 1
+	ElasticsearchSettingsDefaultRequestTimeoutSeconds       = 30
+	ElasticsearchSettingsDefaultBatchSize                   = 10000
```
isacikgoz, line 201: We are actually increasing the default batch size right? Why we do that?

isacikgoz (commented):
This is next level refactoring, great stuff @agnivade 💯 Regarding to deprecated config field, I think it should be safe to remove it right? What kind of side-effect you were expecting? Other than that, this is really nice! Also, do you have any performance results? I am smelling a new dev talk/blog post out of this :)

 I just have a question and some improvement suggestions below.

agnivade:
> Also, do you have any performance results?

It's like 10 times faster on my 488 posts local setup. The older version takes several minutes, but the new version takes just a second. Primarily because of the way it operates. The new version will batch all 488 in one shot. The old version iterates through small batches of time range where there are no posts, so everything just slows down.

In any case, I plan to run a proper test on my 13M setup.

store/storetest/channel_store.go
agnivade: With the new model, you cannot really guarantee exact order of elements. Because ids are random, so elements can be ordered randomly within the same timestamp. So we can only assert for number of elements, and verify that pagination works correctly till the end.

model/config.go
agnivade: This batch size is different from the older one. The older one was based on a time interval. This is based on the number of elements. This is much more efficient and faster. It's okay to load 10k posts in a single batch to ES. That's a decent enough number.

isacikgoz (approved):
Wow sounds like a massive improvement, looking forward to this 💪

agnivade:
There's just one small snag - I need to add an index to the posts table for this :)

agnivade:
Results from the load-test. A HUGE round of thanks to @lieut-data for doing this.

The following graphs compare the p50 and p99 only for `PostStore.*` metrics. Thereby confirming there is no regression.

mysql
postgres

streamer45:
@agnivade @isacikgoz 

Are the breaking changes to the config going to be an issue for other binaries (e.g. mmctl, load-test-ng, plugins)?

agnivade:
There are no breaking changes. The config param is there, just deprecated.

PR #19844, opened 2022-03-26
Add retry logic to initialize CodeQL

agnivade:
```release-note
NONE
```

### DIFF
--- a/.github/workflows/codeql-analysis.yml
+++ b/.github/workflows/codeql-analysis.yml
@@ -26,8 +26,6 @@ jobs:
       uses: Wandalen/wretry.action@v1.0.11
       with:
         action: github/codeql-action/init@v1
-        with: |
-          languages: ${{ matrix.language }}
         attempt_limit: 3
         attempt_delay: 5000        
     - name: Autobuild
--- a/.github/workflows/codeql-analysis.yml
+++ b/.github/workflows/codeql-analysis.yml
@@ -23,9 +23,13 @@ jobs:
     - name: Checkout repository
       uses: actions/checkout@v2
     - name: Initialize CodeQL
-      uses: github/codeql-action/init@v1
+      uses: Wandalen/wretry.action@v1.0.11
       with:
-        languages: ${{ matrix.language }}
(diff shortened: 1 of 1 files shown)


srkgupta:
Hi @agnivade 

I earlier tried with the Retry Action but it was failing as seen here. 

The issue with Github Actions failing which we had earlier encountered would not be resolved with the retry actions because as shown here in the failing the job, the Github action already has an internal retry logic:

https://github.com/mattermost/mattermost-server/runs/5481605618?check_suite_focus=true

```
    request: {
      agent: [Agent],
      hook: [Function: bound bound register],
      retryCount: 3,
      retries: 3,
      retryAfter: 16
    }
```

If you check the logs, the issue happened because the Action made a GET request to `https://api.github.com/repos/mattermost/mattermost-server/actions/runs/1957816678` API endpoint which responded back with a 502 Server Error. 

Also, as mentioned in the Github CodeQL action's troubleshooting [documentation here](https://docs.github.com/en/code-security/code-scanning/automatically-scanning-your-code-for-vulnerabilities-and-errors/troubleshooting-the-codeql-workflow#error-server-error), the only suggested approach is to contact their Support Team. 

```
If the run of a workflow for code scanning fails due to a server error, try running the workflow again. If the problem persists, contact [GitHub Support](https://support.github.com/contact?tags=docs-generic).
```

So the suggested approach would be to enable debug mode as done here and then try to contact the Github Support Team, if we are seeing the issues again. 
https://github.com/srkgupta/mattermost-server/blob/master/.github/workflows/codeql-analysis.yml#L29

Also please note that Github Actions was experiencing disruptions during that period as highlighted here:
https://www.githubstatus.com/incidents/cypv026dr23w

So considering all of these points, I would recommend to only enable the debug mode as of now and then relook at this if it is failing again.

agnivade:
Interesting, I missed that it has an internal retry logic. Agree, let's enable debug mode and see.

agnivade:
@srkgupta - PTAL

srkgupta (approved):
Thanks @agnivade. LGTM

PR #19849, opened 2022-03-28
MM-39058 - invite people to join team and channels

pvev:
#### Summary
This PR adds the option to invite members directly to channels. This creates a new set of functions to cover the functionality to being able to invite people to join a team and a channel directly. Also modifies the emails templates to reflect these invitations.

#### Related Pull Requests
Webapp: https://github.com/mattermost/mattermost-webapp/pull/10083

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-42543

#### Release Note
```release-note
NONE
```


neallred (commented):
Looking good! A few nits, but nothing major.

api4/team.go
```diff
@@ -1275,12 +1275,23 @@ func inviteUsersToTeam(c *Context, w http.ResponseWriter, r *http.Request) {
 		return
 	}
 
-	emailList := model.ArrayFromJSON(r.Body)
+	memberInvite := model.MemberInvite{}
```
JulienTant, line 1278: if you instantiate here as `&model.MemberInvite{}`, you could drop the `&` from the next line but also from function call `App.InviteNewUsersToTeamGracefully`

app/email/email.go
```diff
@@ -604,6 +604,147 @@ func (es *Service) SendGuestInviteEmails(team *model.Team, channels []*model.Cha
...
+
+	if rateLimited {
+		mlog.Error("rate limit exceeded", mlog.Duration("RetryAfter", result.RetryAfter), mlog.Duration("ResetAfter", result.ResetAfter), mlog.String("user_id", senderUserId),
+			mlog.String("team_id", team.Id), mlog.String("retry_after_secs", fmt.Sprintf("%f", result.RetryAfter.Seconds())), mlog.String("reset_after_secs", fmt.Sprintf("%f", result.ResetAfter.Seconds())))
+		return RateLimitExceededError
+	}
+
+	channelsLen := len(channels)
+
+	subject := i18n.T("api.templates.invite_team_and_channels_subject_a", map[string]interface{}{
```
JulienTant, line 635: we could use the keys `invite_team_and_channels_subject` and `invite_team_and_channel_subject` (no s at channel, or change with `single_channel`) to avoid the a/b suffixes and have more explicit names

app/email/email.go
```diff
@@ -604,6 +604,147 @@ func (es *Service) SendGuestInviteEmails(team *model.Team, channels []*model.Cha
...
+
+		title = i18n.T("api.templates.invite_team_and_channels_body.title_b", map[string]interface{}{
+			"SenderName":      senderName,
+			"ChannelName":     channelName,
+			"TeamDisplayName": team.DisplayName,
+		})
+	}
+
+	for _, invite := range invites {
+		if invite != "" {
```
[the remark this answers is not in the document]
JulienTant: ^ 👍🏻

app/team.go
```diff
@@ -1207,7 +1207,7 @@ func (a *App) postRemoveFromTeamMessage(c *request.Context, user *model.User, ch
 	return nil
 }
 
-func (a *App) prepareInviteNewUsersToTeam(teamID, senderId string) (*model.User, *model.Team, *model.AppError) {
+func (a *App) prepareInviteNewUsersToTeam(teamID string, senderId string, memberInvite *model.MemberInvite) (*model.User, *model.Team, []*model.Channel, *model.AppError) {
```
JulienTant: `teamID string, senderId string`; I don't know if you added the first string on purpose, but it's not required in go to declare a parameter  type if the following one has the same type.

app/team.go
```diff
@@ -1222,14 +1222,36 @@ func (a *App) prepareInviteNewUsersToTeam(teamID, senderId string) (*model.User,
...
+			channels, err := a.Srv().Store.Channel().GetChannelsByIds(memberInvite.Channels, false)
+			cchan <- store.StoreResult{Data: channels, NErr: err}
+			close(cchan)
+		}()
+
+		result := <-cchan
+		if result.NErr != nil {
+			return nil, nil, nil, model.NewAppError("prepareInviteNewUsersToTeam", "app.channel.get_channels_by_ids.app_error", nil, result.NErr.Error(), http.StatusInternalServerError)
+		}
+		channels := result.Data.([]*model.Channel)
```
JulienTant: having **one** go routine immediately followed by a blocking channel read seems unnecessary.

I'm pretty sure this will produce the same result:, and might even be faster
```suggestion
		channels, err := a.Srv().Store.Channel().GetChannelsByIds(memberInvite.Channels, false)
		err != nil {
			return nil, nil, nil, model.NewAppError("prepareInviteNewUsersToTeam", "app.channel.get_channels_by_ids.app_error", nil, result.NErr.Error(), http.StatusInternalServerError)
		}
```

JulienTant (commented):
Good job @pvev! I left a few remarks and nits

model/member_invite.go
```diff
@@ -0,0 +1,37 @@
...
+	"net/http"
+)
+
+type MemberInvite struct {
+	Emails   []string `json:"emails"`
+	Channels []string `json:"channels,omitempty"`
+	Message  string   `json:"message"`
+}
+
+// IsValid validates the user and returns an error if it isn't configured correctly.
```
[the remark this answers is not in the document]
pvev: changed!

app/email/email.go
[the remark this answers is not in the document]
pvev: done!

app/team.go
```diff
@@ -1222,14 +1222,33 @@ func (a *App) prepareInviteNewUsersToTeam(teamID, senderId string) (*model.User,
 		close(uchan)
 	}()
 
-	result := <-tchan
+	cchan := make(chan store.StoreResult, 1)
+	go func() {
+		channels, err := a.Srv().Store.Channel().GetChannelsByIds(memberInvite.Channels, false)
```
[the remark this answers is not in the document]
pvev: great advices! thanks.

app/email/email.go
pvev: make sense, thanks!

api4/team.go
pvev: ok, thanks.

app/team.go
pvev: ok

app/team.go
pvev: going to give it a try.

app/team.go
pvev: thanks for the comment! solved

app/email/email.go
pvev: nice catch! solved

api4/team.go
pvev: done

api4/team.go
```diff
@@ -1275,12 +1275,23 @@ func inviteUsersToTeam(c *Context, w http.ResponseWriter, r *http.Request) {
...
+	emailList := memberInvite.Emails
 
 	for i := range emailList {
 		emailList[i] = strings.ToLower(emailList[i])
 	}
 
+	if err := memberInvite.IsValid(); err != nil {
+		c.Err = err
+		return
+	}
```
marianunez, line 1293: The `isValid` seems to be checking if all emails are valid and failing if any are. I believe this would break the send gracefuly functionality given that it will not send any if there is one that is not valid. Before it would send all that it could and return the error of which ones failed.

api4/team.go
```diff
@@ -1275,12 +1275,23 @@ func inviteUsersToTeam(c *Context, w http.ResponseWriter, r *http.Request) {
...
+
 	if len(emailList) == 0 {
 		c.SetInvalidParam("user_email")
 		return
```
marianunez, line 1297: Additionally, it seems that this is included in the valid check above so we should converge in checking in one place.

api4/team.go
```diff
@@ -1275,12 +1275,23 @@ func inviteUsersToTeam(c *Context, w http.ResponseWriter, r *http.Request) {
...
 	}
 
-	emailList := model.ArrayFromJSON(r.Body)
+	memberInvite := &model.MemberInvite{}
+	if jsonErr := json.NewDecoder(r.Body).Decode(memberInvite); jsonErr != nil {
```
marianunez, line 1279: Would this be considered a breaking change to the api?

marianunez (changes requested):
Great work @pvev! Some comments regarding the new model for the api request and verification below 👇

api4/team.go
pvev: yes, because the api call was expecting before an array of strings and now expect a struct defining at least the emails property which is that array of strings. I think is a better idea to support both data structures so we don't break the api current behavior.

api4/team.go
pvev: good catch! thanks Maria, with this validation the "gracefully" behavior is broken.

JulienTant (commented):
Thanks for the fixes. I left a comment around the unmarshal.

api4/team.go
```diff
@@ -1275,7 +1275,14 @@ func inviteUsersToTeam(c *Context, w http.ResponseWriter, r *http.Request) {
 		return
 	}
 
-	emailList := model.ArrayFromJSON(r.Body)
+	memberInvite := &model.MemberInvite{}
+	bf, _ := io.ReadAll(r.Body)
+	if jsonErr := json.Unmarshal(bf, &memberInvite); jsonErr != nil {
```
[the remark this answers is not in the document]
pvev: the idea with this function https://github.com/mattermost/mattermost-server/pull/19849/commits/684c987015de73a4b84fbe7dae2c2fdffe8d6cdc#diff-0b9cc7e6646187be630de194b83272a908f89db4fbe824b5c336fca57e938840R34 is to cover both scenarios, when you send a plain array of emails ["email1@email", "email2@email"] or a data structure that matches with the structure of memberInvite struct {emails: ["...], channels: ["..."]}.

api4/team.go
```diff
@@ -1275,7 +1275,18 @@ func inviteUsersToTeam(c *Context, w http.ResponseWriter, r *http.Request) {
...
 	}
 
-	emailList := model.ArrayFromJSON(r.Body)
+	bf, err := io.ReadAll(r.Body)
+	if err != nil {
+		c.Err = model.NewAppError("Api4.inviteUsersToTeams", "api.team.invite_members_to_team_and_channels.invalid_body.app_error", nil, err.Error(), http.StatusBadRequest)
+		return
+	}
+	memberInvite := &model.MemberInvite{}
+	if jsonErr := json.Unmarshal(bf, &memberInvite); jsonErr != nil {
```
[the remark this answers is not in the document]
pvev: ok!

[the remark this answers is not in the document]
pvev: what you are suggesting is what it does with the overwrite of the Unmarshal method in the inviteMembers struct. With that Unmarshal function we are able to process both inputs, the array of emails and the memberInvite struct. In case of array of emails, we turn it into a memberInvite struct. If you check the webapp code, the current members invite modal, is still sending the array of emails and it is being parsed correctly, the same as when you open the add members channel and try to invite a member, it will send the memberInvite structure, and both work fine.

Captura de pantalla 2022-03-29 a las 12 55 17
Captura de pantalla 2022-03-29 a las 12 55 46
Captura de pantalla 2022-03-29 a las 12 56 00

api4/team_local.go
```diff
@@ -76,7 +77,15 @@ func localInviteUsersToTeam(c *Context, w http.ResponseWriter, r *http.Request)
 		return
 	}
 
-	emailList := model.ArrayFromJSON(r.Body)
+	memberInvite := &model.MemberInvite{}
```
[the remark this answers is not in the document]
pvev: got it

model/member_invite.go
```diff
@@ -0,0 +1,49 @@
...
+			if len(channel) != 26 {
+				return NewAppError("MemberInvite.IsValid", "model.member.is_valid.channel.app_error", nil, "channel="+channel, http.StatusBadRequest)
+			}
+		}
+	}
+
+	return nil
+}
+
+func (i *MemberInvite) UnmarshalJSON(b []byte) error {
```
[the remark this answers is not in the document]
neallred: ahh, yep, missed that! This also addresses my other comment about parsing both ways to limit API breakage. Nicely done!

api4/team_test.go
```diff
@@ -2982,7 +2982,8 @@ func TestInviteUsersToTeam(t *testing.T) {
 	user1 := th.GenerateTestEmail()
 	user2 := th.GenerateTestEmail()
 
-	emailList := []string{user1, user2}
+	memberInvite := &model.MemberInvite{Emails: []string{user1, user2}}
+	emailList := memberInvite.Emails
```
marianunez, line 2986: Could we add another test to make sure both inputs are still valid?

model/client4.go
```diff
@@ -2608,7 +2608,14 @@ func (c *Client4) ImportTeam(data []byte, filesize int, importFrom, filename, te
 
 // InviteUsersToTeam invite users by email to the team.
 func (c *Client4) InviteUsersToTeam(teamId string, userEmails []string) (*Response, error) {
-	r, err := c.DoAPIPost(c.teamRoute(teamId)+"/invite/email", ArrayToJSON(userEmails))
+	memberInvite := MemberInvite{
+		Emails: userEmails,
+	}
+	buf, err := json.Marshal(memberInvite)
```
marianunez, line 2614: We should have another option to invite to team with email + channel to make sure we are not breaking compatabiility

model/member_invite.go
```diff
@@ -0,0 +1,49 @@
...
+package model
+
+import (
+	"encoding/json"
+	"net/http"
+)
+
+type MemberInvite struct {
+	Emails   []string `json:"emails"`
+	Channels []string `json:"channels,omitempty"`
```
marianunez, line 13: Should we rename this to `ChannelIds` so that it's clear that is not an array of channels?

marianunez:
Adding @AGMETEOR for visibility on changes related to email invites reminders

model/client4.go
pvev: compatibility is working. The current member email invite flow keeps working the same, sending a plain email-strings array. Only when the member invite modal is opened from the add member to channel modal (by clicking the 'invite them to the team' new link), the data structure changes to the new memberInvite model structure model.

api4/team_test.go
pvev: the test is already there https://github.com/mattermost/mattermost-server/pull/19849/files#diff-f157eb201938e9e26eff6958772bfa684b18152b6697e7cc94fb2261ef4f3003R1326

model/member_invite.go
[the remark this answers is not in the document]
JulienTant: 👏🏻

JulienTant (approved):
LGTM @pvev, smart move to use a custom unmarshaller to handle both types of input.
Not sure how to reflect those two options on the API documentation tho 🤔

api4/team_test.go
marianunez: In that test you are passing also a MemberInvite model with only emails. We should test that the api endpoint can receive just an array of strings like it did before

model/client4.go
```diff
@@ -2637,7 +2644,14 @@ func (c *Client4) InviteGuestsToTeam(teamId string, userEmails []string, channel
...
 func (c *Client4) InviteUsersToTeamGracefully(teamId string, userEmails []string) ([]*EmailInviteWithError, *Response, error) {
-	r, err := c.DoAPIPost(c.teamRoute(teamId)+"/invite/email?graceful="+c.boolString(true), ArrayToJSON(userEmails))
+	memberInvite := MemberInvite{
+		Emails: userEmails,
+	}
+	buf, err := json.Marshal(memberInvite)
+	if err != nil {
+		return nil, nil, NewAppError("InviteMembersToTeam", "api.marshal_error", nil, err.Error(), http.StatusInternalServerError)
+	}
+	r, err := c.DoAPIPostBytes(c.teamRoute(teamId)+"/invite/email?graceful="+c.boolString(true), buf)
```
marianunez, line 2654: Same here, it should still work without this change

marianunez (approved):
Great work @pvev 💯

pvev:
maybe @AGMETEOR can take a look at this after merge. @marianunez ^^

marianunez:
> maybe @AGMETEOR can take a look at this after merge. @marianunez ^^

Sure, no need to hold the PR with all the approvals complete. 👍

PR #19850, opened 2022-03-28
[MM-42742] Insights - Top Reactions API Endpoint

mylonsuren:
#### Summary

Adds new endpoint to get top reactions for a team and user.

Reference: [tech-specs/activity-and-insights/widgets/top-reactions.md#rest-apis](https://github.com/mkraft/suite-users/blob/main/tech-specs/activity-and-insights-widgets/top-reactions.md#rest-apis)

This adds two new endpoints:
 `GET /api/v4/teams/:team_id/top/reactions` and `GET /api/v4/users/me/top/reactions`.
 
Top reactions for a user can also be scoped to a team by adding the `team_id` as a parameter to the request.

This also introduces a new parameter `time_range`, which scopes the request to reactions from a certain time range.
- `today`: Today
- `7_day`: Last 7 days
- `28_day`: Last 28 days

#### Sample Requests

Top Reactions for a Team:
```
http://localhost:8065/api/v4/teams/{{team_id}}/top/reactions?time_range=28_day&page=0&per_page=5
```
```
{
"has_next": false,
"items:": [
    {"rank": 1, "emoji_name":"blush","count":74},
    {"rank": 2, "emoji_name":"-1","count":66},
    {"rank": 3, "emoji_name":"+1","count":62},
    {"rank": 4, "emoji_name":"heart","count":55}
  ]
}
```
Top Reactions for a User (Scoped to Team):
```
http://localhost:8065/api/v4/users/me/top/reactions?time_range=28_day&page=0&per_page=5&team_id={{team_id}}
```
```
{
"has_next": false,
"items:": [
    {"rank": 1, "emoji_name":"blush","count":10},
    {"rank": 2, "emoji_name":"-1","count":9},
    {"rank": 3, "emoji_name":"+1","count":7},
    {"rank": 4, "emoji_name":"heart","count":3}
  ]
}
```

- `has_next` indicates if there are more responses that can be fetched by incrementing the page. 
- `rank` is the rank of the emoji (by `count`) and will track across pages (i.e. if you fetch the second page with a limit of 5, the ranks would be 6 through 10).

#### Ticket Link

Fixes [MM-42742](https://mattermost.atlassian.net/browse/MM-42742)

#### Release Note

```release-note
Added new API endpoints:  GET /api/v4/teams/:team_id/top/reactions and GET /api/v4/users/me/top/reactions
```


BenCookie95 (changes requested):

api4/reaction.go
```diff
@@ -16,6 +16,9 @@ func (api *API) InitReaction() {
 	api.BaseRoutes.Post.Handle("/reactions", api.APISessionRequired(getReactions)).Methods("GET")
 	api.BaseRoutes.ReactionByNameForPostForUser.Handle("", api.APISessionRequired(deleteReaction)).Methods("DELETE")
 	api.BaseRoutes.Posts.Handle("/ids/reactions", api.APISessionRequired(getBulkReactions)).Methods("POST")
+
+	api.BaseRoutes.Reactions.Handle("/top/team/{team_id:[A-Za-z0-9]+}", api.APISessionRequired(getTopReactionsForTeamSince)).Methods("GET")
```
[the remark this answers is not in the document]
mylonsuren: Yeah not sure either, but looking at it now I think you're right in putting `top` at the end as it matches some of our other endpoints better. I'll make the change :+1:

api4/reaction.go
```diff
@@ -16,6 +16,9 @@ func (api *API) InitReaction() {
...
+	api.BaseRoutes.Reactions.Handle("/top/user/{user_id:[A-Za-z0-9]+}", api.APISessionRequired(getTopReactionsForUserSince)).Methods("GET")
```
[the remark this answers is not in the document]
mylonsuren: Yes you're right, I wasn't entirely sure if we were scoping to only allow the user to view their own reactions so I had left it open and just added the required permission checks for the user. I'll change the endpoint to use `me` instead.

web/params.go
```diff
@@ -254,6 +255,8 @@ func ParamsFromRequest(r *http.Request) *Params {
 		params.Timestamp = val
 	}
 
+	params.TimeRange = query.Get("time_range")
```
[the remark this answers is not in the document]
mylonsuren: Yeah, good idea. I've modified it now to get the time range here.

store/sqlstore/reaction_store.go
```diff
@@ -228,6 +228,132 @@ func (s *SqlReactionStore) PermanentDeleteBatch(endTime int64, limit int64) (int
 	return rowsAffected, nil
 }
 
+func (s *SqlReactionStore) GetTopForTeamSince(teamID string, userID string, since int64, offset int, limit int) ([]*model.TopReactions, error) {
+	var reactions []*model.TopReactions
+
+	// Get reactions from all public channels within the team, and private channels that the user is a member of
+	query := `
```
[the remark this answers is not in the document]
mylonsuren: These queries still need to be optimized. Right now in Postgres it takes about 250ms for a user and 4s for the team (on the 15M post database over 28 days). I was thinking of just getting the endpoints setup so we can test and play around with them, and then optimize the queries (and performance elsewhere) once all the tests are in place so that they're easier to validate. I can create a ticket in JIRA for that as well.

store/sqlstore/reaction_store.go
```diff
@@ -228,6 +228,132 @@ func (s *SqlReactionStore) PermanentDeleteBatch(endTime int64, limit int64) (int
...
+				AND Reactions.UserId = ?
+				AND Reactions.CreateAt > ?
+			GROUP BY
+				Reactions.EmojiName
+			ORDER BY
+				Count DESC
+			LIMIT ?
+			OFFSET ?`
+		args = []interface{}{userID, since, limit, offset}
+	}
```
mkraft, line 349: Was the `UNION ALL` necessary for this one or could we simplify it to something like:

```sql
SELECT
    EmojiName,
    count(EmojiName) AS Count
        FROM
            Reactions
            INNER JOIN Posts ON Reactions.PostId = Posts.Id
            INNER JOIN Channels ON Posts.ChannelId = Channels.Id
        WHERE
            Reactions.DeleteAt = 0
            AND Reactions.UserId = ?
            AND (Channels.TeamId = ? OR Channels.Type = 'D' OR Channels.Type = 'G')
            AND Reactions.CreateAt > 0
GROUP BY
    EmojiName
ORDER BY
    Count DESC
LIMIT ?
OFFSET ?
```

api4/reaction.go
```diff
@@ -16,6 +16,9 @@ func (api *API) InitReaction() {
 	api.BaseRoutes.Post.Handle("/reactions", api.APISessionRequired(getReactions)).Methods("GET")
 	api.BaseRoutes.ReactionByNameForPostForUser.Handle("", api.APISessionRequired(deleteReaction)).Methods("DELETE")
 	api.BaseRoutes.Posts.Handle("/ids/reactions", api.APISessionRequired(getBulkReactions)).Methods("POST")
+
+	api.BaseRoutes.Reactions.Handle("/team/{team_id:[A-Za-z0-9]+}/top", api.APISessionRequired(getTopReactionsForTeamSince)).Methods("GET")
+	api.BaseRoutes.Reactions.Handle("/user/me/top", api.APISessionRequired(getTopReactionsForUserSince)).Methods("GET")
```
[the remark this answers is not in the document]
mylonsuren: Yeah I was thinking the same thing after looking through channels, I've updated the paths to `teams/<team_id>/reactions/top` and `users/me/reactions/top`

mylonsuren:
@mkraft @BenCookie95 Sorry I'm doing some rewrites here after some investigation, and will incorporate your comments and let you know once it's ready for another review.

store/sqlstore/reaction_store.go
mylonsuren: You're right, the `UNION ALL` and additional joins were unnecessary and slowed down the query. Updated to use `OR` instead.

store/sqlstore/reaction_store.go
```diff
@@ -228,6 +228,132 @@ func (s *SqlReactionStore) PermanentDeleteBatch(endTime int64, limit int64) (int
...
+			Count DESC
+		LIMIT ?
+		OFFSET ?`
+
+	if err := s.GetReplicaX().Select(&reactions, query, userID, teamID, since, teamID, since, limit, offset); err != nil {
+		return nil, errors.Wrap(err, "failed to get top Reactions")
+	}
+	return reactions, nil
+}
+
```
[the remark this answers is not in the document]
mylonsuren: Done 👍

store/sqlstore/reaction_store.go
```diff
@@ -228,6 +228,132 @@ func (s *SqlReactionStore) PermanentDeleteBatch(endTime int64, limit int64) (int
 	return rowsAffected, nil
 }
```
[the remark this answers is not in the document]
mylonsuren: Done 👍

mylonsuren:
@BenCookie95 @mkraft Sorry for the messy commits, should be good for second review 😄

i18n/en.json
```diff
@@ -5863,6 +5867,14 @@
     "id": "app.reaction.get_for_post.app_error",
     "translation": "Unable to get reactions for post."
   },
+  {
+    "id": "app.reaction.get_top_for_team_since.app_error",
+    "translation": "Unable to get top reactions for team."
+  },
+  {
+    "id": "app.reaction.get_top_for_user_since.app_error",
+    "translation": "Unable to get top reactions for user."
```
mkraft, line 5876: Are these all seen by users in the UI? The reason I ask is we've updated our method of operating regarding non-user-facing errors. They're not longer translated. You can trick the CI for now by committing a single blank space `" "` for anything that won't be user-facing. A proper path forward is pending from the Server Platform team.

model/insights.go
```diff
@@ -0,0 +1,41 @@
...
+)
+
+const (
+	TimeRange1Day  string = "1_day"
+	TimeRange7Day  string = "7_day"
+	TimeRange28Day string = "28_day"
+)
+
+type InsightsOpts struct {
+	TimeRange int64
```
mkraft, line 18: Instead of `TimeRange` being `int64` could we make it a `time.Duration` and add a custom method `(i *InsightsOpts),MarshalJSON` that can convert the parameter we receive from the API request (ex `1_day`) to a `time.Duration` value (ex  `time.Duration(time.Hour*24)`)?

mkraft (changes requested):

model/insights.go
mkraft: Sorry, just remembered it's not coming in via JSON but rather via a query param, so the `MarshalJSON` method wouldn't make sense, but same idea.

mylonsuren: Right now we're converting the string directly to the timestamp that's used in the query (which is of type `int64`). 
https://github.com/mattermost/mattermost-server/blob/8b39b8be808c9e7ed08d7b1f89815031a727741d/model/insights.go#L36-L47

I'm not sure I see the benefit of storing it as a `time.Duration` if we'd have to convert that into the timestamp before passing it to the query. I might be misunderstanding what you mean though.

Just to clarify as well, `TimeRange` here holds the timestamp of the earliest reaction that we would fetch. It might be a misunderstanding of the name, in which case I can rename it to something more clear.

i18n/en.json
mylonsuren: I don't think any of these strings will appear in the UI, I'll replace them with the blank space :+1:

model/insights.go
mkraft: Right, now I realize it's not representing a range at all and I agree that another option is to rename the field to something like `StartUnixMilli` (making sure it includes the units).

What do you prefer from the perspective of a client using `(*App).GetTopReactionsForUserSince`, do you think it's better to say get me reactions for the past `time.Duration(time.Hour)` or to pass in the unix timestamp?

mkraft:
I didn't dig in specifically, do the tests check the exclusion case where there are reactions outside of the requested time range that should not appear in the sums?

mkraft (approved):
Nice one! I just had one question about a test case but otherwise no other questions from me.

mylonsuren:
> Nice one! I just had one question about a test case but otherwise no other questions from me.

@mkraft I've just added some additional test data to cover that scenario, so should be good. I'm just double checking everything and simplifying some duplicated code. No major changes expected but I will let you know and @BenCookie95 know once it's ready for a final pass 🤞

mkraft (approved):

BenCookie95 (approved):

furqanmlk (approved):
Working as expected.
image

image

image

PR #19856, opened 2022-03-29
api4/license: Return 400 instead of 501 if format is not passed

agnivade:
```release-note
NONE
```


spirosoik (approved):
thanks for the fast PR :)

agnivade:
Good to see some tests are failing :)

PR #19835, opened 2022-03-24
[MM-42713] - Remove cloud user limit restrictions

AGMETEOR:
#### Summary
This PR removes cloud user limitation number from the webapp

#### Ticket Link
[MM-42713](https://mattermost.atlassian.net/browse/MM-42713)

#### Related Pull Requests

- Has webapp changes (https://github.com/mattermost/mattermost-webapp/pull/10061)
- Has enterprise changes (https://github.com/mattermost/enterprise/pull/1174)

#### Release Note

```release-note
Remove cloud user limit count
```

PR #19874, opened 2022-03-30
[MM-41993]: fixed counting thread mentions in unread root post

michelengelen:
#### Summary
the root post got appended to the post list which led to it being counted towards mentions in the thread, which is considered a bug according to the specs.

The second change (`CreateAt` instead of `UpdateAt` in threads_store.go) was done in collaboration with @koox00 

#### Ticket Link
[MM-41993](https://mattermost.atlassian.net/browse/MM-41993)

#### Release Note
```release-note
NONE
```


koox00 (commented):
LGTM 💪 Can you fix the tests as well?

michelengelen:
@koox00 thanks for your time and help ... I actually had to add a new assertion to the test as we were defining a var `reply` and the test script did not like that. I took the chance to just test marking the first reply as unread as well and check if the correct value is returned, so we can be sure that this is not failing anymore. 🙇🏼

koox00 (approved):

ashishbhate:
>The ThreadStore().GetPosts seems to be called only from count mentions, I don't know the reasoning behind using update_at instead of create_at, do you have any idea @ashishbhate?

I'm not sure. I looked in the git history but I couldn't find anything. 

>Using create_at seemed to me more intuitive and intentional that's why I proposed the change.
>PS: since we exclude posts having a CreateAt less than timestamp when we count mentions. keeping update_at in the query would not affect functionality.

Since it's only being called from the `countThreadMentions` method, I think it's fine to change it. You're that it wont affect functionality.

ashishbhate (approved):
LGTM!

jgilliam17:
Thanks @michelengelen 
Tested, looks good to merge - pending E2E report
- Verified thread mention is not counted when marking root post with mention as unread on rhs.

jgilliam17:
@michelengelen Can you please take a look at the failure - blocked on running e2es. Thanks 🙂

michelengelen:
@ashishbhate or @koox00 could you have a look why that job is failing? I don't have that much experience with the server repo

ashishbhate:
@michelengelen 

It looks like the i18n tool is trying to re-insert a translation (ent.cloud.subscription.error) that was removed in #19835 

I'm not sure why its trying to do that.

jgilliam17:
Thanks for fixing failed checks and running E2Es @michelengelen , [report](https://automation-dashboard.vercel.app/cycles/053c6aaa-352d-4456-b008-4012548cf7c8) looks good - no PR related failures.

jgilliam17 (approved):

PR #19878, opened 2022-03-30
MM-42919: avoid sequential scan on Posts

lieut-data:
#### Summary
On Postgres, `GetTeamsUnreadForUser` triggers a sequential scan on `Posts`. We can avoid this by querying the `Threads` table directly and only joining to `Posts` to eliminate deleted threads. (We could avoid the latter if we later denormalize `DeleteAt` onto `Threads`.)

#### Ticket Link
Fixes: https://mattermost.atlassian.net/browse/MM-42919

#### Release Note
```release-note
Improve performance of GetTeamsUnreadForUser when collapsed reply threads is enabled.
```

### DIFF
--- a/store/sqlstore/thread_store.go
+++ b/store/sqlstore/thread_store.go
@@ -358,12 +358,13 @@ func (s *SqlThreadStore) GetTeamsUnreadForUser(userID string, teamIDs []string) 
 	go func() {
 		defer wg.Done()
 		repliesQuery, repliesQueryArgs, err := s.getQueryBuilder().
-			Select("COUNT(DISTINCT(Posts.RootId)) AS Count, TeamId").
-			From("Posts").
-			LeftJoin("ThreadMemberships ON Posts.RootId = ThreadMemberships.PostId").
-			LeftJoin("Channels ON Posts.ChannelId = Channels.Id").
+			Select("COUNT(Threads.PostId) AS Count, TeamId").
+			From("Threads").
+			LeftJoin("ThreadMemberships ON Threads.PostId = ThreadMemberships.PostId").
+			LeftJoin("Channels ON Threads.ChannelId = Channels.Id").
+			LeftJoin("Posts ON Posts.Id = Threads.PostId").
 			Where(fetchConditions).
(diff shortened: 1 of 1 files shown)


agnivade:
@lieut-data - Just to confirm here - you wanted to add the rewinding functionality as part of this PR itself to maintain correct unreads count?

lieut-data:
@agnivade, actually, given the impact of these changes, I was inclined to take the lazy approach and file a bug alongside the other columns we don't currently maintain properly. Thoughts?

esethna:
@lieut-data can you help me understand what this PR is changing in terms of current behaviour vs expected behaviour after this PR goes in?

lieut-data:
> @lieut-data can you help me understand what this PR is changing in terms of current behaviour vs expected behaviour after this PR goes in?

@esethna, in short, if a user posts a reply to a thread and then deletes it, this API endpoint will still "count" that thread as unread because we're currently not "rewinding" `Threads.LastReplyAt` when deleting a reply comment. This self-resolves once a new reply is added, or the thread is viewed.

As part of this change, I'll file a bug to properly maintain `Threads.LastReplyAt` on deleting a post so we can circle back and address this holistically. The proposal with not doing that right now is that the user-facing impact of the sequential scan on Posts is objectively worse than this edge case.

esethna:
Okay thanks @lieut-data, agree. Let's please file the bug to track it.

agnivade (approved):

ashishbhate (approved):

lieut-data:
@Phrynobatrachus, can you help kick off an e2e run for this PR so we can compare if anything new breaks?

Phrynobatrachus (approved):

lieut-data:
@agnivade, any concerns with merging this one now, or do we want to see more data from Grafana for @ashishbhate's changes?

agnivade:
Unfortunately community deployments are frozen due to an issue. I am working on that right now. So ashish's changes haven't been deployed yet. I don't want to block this one, so go ahead with the merge. Ashish's changes will only change the rate, and yours will change the duration. So I am hoping to still be able to separate the two changes.

PR #19879, opened 2022-03-31
Update Scorecards action

oh6hay:
Update the action to [v1.0.4](https://github.com/ossf/scorecard-action/releases/tag/v1.0.4)
this should fix the odd failures on periodic runs


srkgupta (approved):
Thanks @oh6hay. LGTM

PR #19895, opened 2022-04-01
[MM-42474] OAuth redirect issue

BenCookie95:
#### Summary
I was not able to reproduce the exact issue from the ticket but I can see why the redirect was incorrect.

The URL for redirecting to the `/complete` endpoint was being generated using `c.GetSiteURLHeader()`. If you track down what the value of that will be, it would look like this: `app.GetProtocol(r) + "://" + r.Host + subpath`. The issue here is that `GetProtocol's` value would be set to `http` if `X-Forwarded-Proto` is set to http. According to this old [ticket](https://github.com/mattermost/mattermost-server/issues/6489#issuecomment-305917377), Gitlab sets `X-Forwarded-Proto` to http by default. 

Thanks to this comment by @apollo13 https://github.com/mattermost/mattermost-server/issues/19750#issuecomment-1076309739 we can see there is a discrepancy as to how we generate the  `/complete` endpoint. This PR fixes the mismatch.

#### Ticket Link
https://github.com/mattermost/mattermost-server/issues/19750
https://mattermost.atlassian.net/browse/MM-42474

#### Release Note

```release-note
NONE
```


jwilander:
For some context, the historical reason we were using `c.GetSiteURLHeader()` is because some customers would have two different URLs they were hosting the same MM at (one internal, one external). If we only used the configured SiteURL then those customers would not be able to login with one of their URLs.

We ran into that causing an issue for Cloud a while back and when we had a `cloud` branch we had a "temporary" fix in there to use the SiteURL rather than `c.GetSiteURLHeader()`. See https://github.com/mattermost/mattermost-server/pull/16677.

@nickmisasi I recall this coming up as an issue when we deleted the `cloud` branch and started basing the Cloud releases off `master`. How'd we solve this for that case?

Also, if we do start using the configured SiteURL, there's a ton of other places in `web/oauth.go` where we should update the usage too.

web/oauth.go
```diff
@@ -273,7 +273,11 @@ func completeOAuth(c *Context, w http.ResponseWriter, r *http.Request) {
 
 	state := r.URL.Query().Get("state")
 
-	uri := c.GetSiteURLHeader() + "/signup/" + service + "/complete"
+	siteURL := *c.App.Config().ServiceSettings.SiteURL
```
nickmisasi, line 276: Full disclosure: I'm not super familiar with this flow

Wondering if this single change is enough since there are quite a few other places in this file that seemingly use `c.GetSiteURLHeader`. Probably need to update all of them?

nickmisasi:
@jwilander our fix for cloud (https://github.com/mattermost/mattermost-server/pull/19393) was to use the SiteURL - but we set it in the app/handler.go, but if you did it the same way you'd lose that multiple deployment functionality.

BenCookie95:
@jwilander Another approach could be to update the setup docs to note this bit of information about the header? And leave the existing implementation?

nickmisasi:
+1 on Ben's solution - I think everything here is working "As Designed"

apollo13:
Can you explain what working "as designed" means here? Producing different URLs in /oauth/token and /oauth/authorize is hardly something I'd call working at all. Don't get me wrong, I honestly do not care if there are two URLs for internal and external. But the URLs under /oauth have to consistently use one option to work at all.

Or did your working as designed refer to something else completely?

nickmisasi:
Hi @apollo13!

Before I clarify just want to shed some context on the issue.

In some instances when you host Mattermost behind a load balancer or other sort of ingress, SSL terminates at that point. Traffic between the ingress and Mattermost is often done over regular http. The fix you mention [https://github.com/mattermost/mattermost-server/issues/19750#issuecomment-1076309739](here) is actually what I would consider to be the correct solution to this problem. We plan to add more information about the `X-Forwarded-Proto` header  to the docs to make that more clear.

The reason we're hesitant to merge this change is due to the nature of Mattermost being self-hosted. We don't have control over the infrastructure that Mattermost is deployed to, so we need to be sure we can support the widest range of deployments as possible.

It is a valid use case to support multiple redirect URI's - one that we have a few customers leveraging. That's what I meant by "As Designed"

apollo13:
Hi @nickmisasi I understand that multiple redirect URLs are a valid usecase. But are they actually a valid usecase within one authorization flow? After all Gitlab for instance seems to be blocking the request immediately if the authorize and token requests do not use the same parameters. I fully understand that two users might have different redirect URLs (even different domains etc) but shouldn't it be consistent for a single user?

nickmisasi:
@apollo13 

>I understand that multiple redirect URLs are a valid usecase. But are they actually a valid usecase within one authorization flow?

You're correct that there is no valid use case for the redirect URI changing mid-authorization flow. But like I said earlier, the proper solution to this inconsistency is setting that header - once you've done so, everything should be consistent end to end.

>After all Gitlab for instance seems to be blocking the request immediately if the authorize and token requests do not use the same parameters.

You're absolutely correct. The thing is, most OAuth providers allow you to attribute multiple redirect URI's to a single client ID/Client Secret pair - meaning multiple environments (or the same environment, using different DNS addresses) could authenticate with it, so long as the client ID + Secret are valid (they can be the same each time), and the passed redirect URI is one of the ones that has been configured in the provider.

This is where our (ie, the current) implementation comes in. Some customers have (for various reasons) multiple DNS entries mapped to _the same mattermost instance_ (it's something even we sort of leverage internally with the community, community-release, and community-daily servers). Meaning for example, some users may access the instance via `https://community.mattermost.com` while others may access over `http://community.internal.mattermost.com` (likely over some sort of VPN or internal network)

Since both DNS entries map to the same server, you can't necessarily use the SiteURL to pass as the redirect URI, because which SiteURL would you choose? Right now we build the redirect URI based off the request itself, so auth flows from community.internal.mattermost.com will pass that as the redirect URI (which gitlab accepts so long as its configured), and flows from community.mattermost.com will work as well.

BenCookie95:
> @BenCookie95 Was this PR being closed in favour of a docs update or did I misunderstand?

Yeah I'm going to close this and update docs instead

PR #19896, opened 2022-04-01
adopt forked squirrel supporting from UPDATE FROM

lieut-data:
#### Summary
Fork https://github.com/Masterminds/squirrel as https://github.com/lieut-data/squirrel with the following changes:
* https://github.com/Masterminds/squirrel/pull/256 supporting FROM clause to update builder (Postgres)
* Extension of above to support multiple FROM in UPDATE (Postgres)
* Support for multiple tables in UPDATE (MySQL)

This PR then leverages those changes to simplify a query that previously had to be coded by hand and duplicate for each of MySQL and Postgres.

#### Ticket Link
None.

#### Release Note
```release-note
NONE
```


go.mod
```diff
@@ -126,6 +126,8 @@ require (
 	gopkg.in/yaml.v2 v2.4.0
 )
 
+replace github.com/Masterminds/squirrel v1.5.2 => github.com/lieut-data/squirrel v1.5.4
```
lieut-data, line 129: I've forked this into my own organization for now, but open to promoting it to `mattermost/squirrel` if we think that makes sense!

agnivade: I think we should do that. There are more improvements in squirrel to make. :) But need not block this PR.

agnivade (approved):

isacikgoz (approved):
LGTM!

PR #19899, opened 2022-04-04
[MM-42522] - The reply-to email address in our Cloud trial emails is no-reply@mattermost.com instead of feedback-cloud@mattermost.com

AGMETEOR:
#### Summary
We need to customize the reply to address for some cloud emails like the welcome email. This PR makes it possible.

#### Ticket Link
[MM-42522](https://mattermost.atlassian.net/browse/MM-42522)

#### Release Note

```release-note
NONE
```

### DIFF
--- a/app/email/email.go
+++ b/app/email/email.go
@@ -244,9 +244,10 @@ func (es *Service) SendCloudTrialEndWarningEmail(userEmail, name, trialEndDate, 
 		return err
 	}
 
-	if err := es.sendMail(userEmail, subject, body); err != nil {
+	if err := es.sendEmailWithCustomReplyTo(userEmail, subject, body, *es.config().SupportSettings.SupportEmail); err != nil {
 		return err
 	}
+
 	return nil
 }
 
@@ -271,9 +272,10 @@ func (es *Service) SendCloudTrialEndedEmail(userEmail, name, locale, siteURL str
 		return err
 	}
 
-	if err := es.sendMail(userEmail, subject, body); err != nil {
+	if err := es.sendEmailWithCustomReplyTo(userEmail, subject, body, *es.config().SupportSettings.SupportEmail); err != nil {
 		return err
 	}
+
 	return nil
 }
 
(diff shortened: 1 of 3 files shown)


app/email/utils.go
```diff
@@ -8,9 +8,14 @@ import (
...
 )
 
-func (es *Service) mailServiceConfig() *mail.SMTPConfig {
+func (es *Service) mailServiceConfig(replyToAddress string) *mail.SMTPConfig {
 	emailSettings := es.config().EmailSettings
 	hostname := utils.GetHostnameFromSiteURL(*es.config().ServiceSettings.SiteURL)
+
+	if replyToAddress == "" {
+		replyToAddress = *emailSettings.ReplyToAddress
+	}
```
marianunez, line 17: If the replyToAddress is part of the config, why can we default this to the address we want for cloud and there would be no need to do a special case override here https://github.com/mattermost/enterprise/blob/6b3accafc357e037105fafe14643396708eb90b9/cloud/config/cloud_defaults.json#L16

marianunez (changes requested):

app/email/utils.go
AGMETEOR: @marianunez I thought so too, but that means all cloud emails reply to would be that value set in the config. But based on this conversation https://mattermost.atlassian.net/browse/MM-42522?focusedCommentId=125238, Jason mentioned that it shouldn't be that way

marianunez: I see, but if we are overriding it here at the highest level of the Mail Service, wouldn't this change affect all emails anyways?

marianunez: I see now that we instantiate this for each email. I thought this was a shared config for all emails. 👍

marianunez (approved):
Can we also do this change for all emails coming from CWS webhook? Trial and payment failure emails

app/email/email.go
```diff
@@ -310,7 +310,7 @@ func (es *Service) SendCloudWelcomeEmail(userEmail, locale, teamInviteID, workSp
 		return err
 	}
 
-	if err := es.sendMail(userEmail, subject, body); err != nil {
+	if err := es.sendEmailWithCustomReplyTo(userEmail, subject, body, *es.config().SupportSettings.SupportEmail); err != nil {
```
marianunez, line 313: Any tests we can add to cover this?

neallred (approved):

app/email/email.go
AGMETEOR: Added a unit test

PR #19910, opened 2022-04-05
MM-42880: Fix TestRunServerSuccess

agnivade:
https://mattermost.atlassian.net/browse/MM-42880

```release-note
NONE
```

### DIFF
--- a/cmd/mattermost/commands/server_test.go
+++ b/cmd/mattermost/commands/server_test.go
@@ -17,7 +17,7 @@ import (
 )
 
 const (
-	UnitTestListeningPort = ":0"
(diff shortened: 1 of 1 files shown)


isacikgoz (commented):
Now checked the file and saw that there are similar issues. Could we add this to similar occurrences in the test file? Eg:

https://github.com/mattermost/mattermost-server/blob/c4b4e1bc38eb7ff9fcc77af93873c1386392089e/cmd/mattermost/commands/server_test.go#L117

isacikgoz (approved):
Nice one!

PR #19917, opened 2022-04-05
MM-43077  Allow inline json for advanced logging config

wiggin77:
#### Summary
This PR addresses an issue that several customers and internal folks have tripped over.  Advanced logging allows for configuration to be inlined in config.json but the inlined JSON had to be escaped into a string.  

A better experience is to allow non-escaped inline JSON as people expect.  This requires backwards compatibility for any config using escaped JSON.  Note, inline JSON is also important for cloud config where ENV vars are used.

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-43077

#### Release Note
NONE

### DIFF
--- a/app/audit.go
+++ b/app/audit.go
@@ -108,14 +108,14 @@ func (s *Server) configureAudit(adt *audit.Audit, bAllowAdvancedLogging bool) er
 	adt.OnError = s.onAuditError
 
 	var logConfigSrc config.LogConfigSrc
-	dsn := *s.Config().ExperimentalAuditSettings.AdvancedLoggingConfig
-	if bAllowAdvancedLogging && dsn != "" {
+	dsn := s.Config().ExperimentalAuditSettings.AdvancedLoggingConfig
+	if bAllowAdvancedLogging && len(dsn) > 2 {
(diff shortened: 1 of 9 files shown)


wiggin77:
Ugh.  I don't think I can change config field type like this without causing a breaking change we plugins etc, that compile against model.

streamer45:
> Ugh. I don't think I can change config field type like this without causing a breaking change with plugins etc, that compile against model.

Indeed, this is a breaking change which will cause issues with dependant code (plugins, mmctl, etc), unless recompiled with the latest version.

wiggin77:
> > Ugh. I don't think I can change config field type like this without causing a breaking change with plugins etc, that compile against model.
> 
> Indeed, this is a breaking change which will cause issues with dependant code (plugins, mmctl, etc), unless recompiled with the latest version.

Thanks @streamer45.  Disappointing.  I thought I could get away with it because no plugins reference the field I changed, but the plugin API, particularly the plugin API wrapper, is structured such that even unreferenced fields become dependencies.

My fault for picking the wrong data type in the first place. :-(

PR #19929, opened 2022-04-06
Pre-package Playbooks v1.23.1

Phrynobatrachus:
```release-note
Pre-package Playbooks v1.23.1
```


lieut-data (approved):
Thanks, @Phrynobatrachus! cc @amyblais, this isn't "urgent", but we'd love to get this change into a patch for ESR, possibly in coordination with any other fixes.

PR #19930, opened 2022-04-06
[MM-42355] Prepackage Calls v0.4.8

streamer45:
#### Summary

Adding Calls as pre-packaged plugin. PR is purposely targeting `release-6.6` branch as this is only expected to land on self-managed deployment for the time being in order to unblock some of our closed beta customers.

Had to add a couple of additional checks since Calls currently supports Linux platforms only as backend.

#### Ticket Link

https://mattermost.atlassian.net/browse/MM-42355

#### Release Note

```release-note
Pre-package Calls v0.4.8
```

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -119,6 +119,7 @@ TEMPLATES_DIR=templates
 PLUGIN_PACKAGES ?= mattermost-plugin-antivirus-v0.1.2
 PLUGIN_PACKAGES += mattermost-plugin-autolink-v1.2.2
 PLUGIN_PACKAGES += mattermost-plugin-aws-SNS-v1.2.0
+PLUGIN_PACKAGES += mattermost-plugin-calls-v0.4.8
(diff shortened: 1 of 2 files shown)


cpoile (approved):
👍

lieut-data (approved):

lieut-data:
cc @amyblais

lieut-data:
Propose we rely on a cloud test server to verify the pre-packaging.

lieut-data:
Calls successfully installs from the marketplace, and can be enabled. (Can't actually get data to transmit when running, but I'm assuming that's a `/cloud` specific thing.) LGTM!

lieut-data:
@Phrynobatrachus, anything else you'd recommend we test before merging? (The primary goal of this effort isn't that calls is supposed to be functional out of the box, but that we can ship it via pre-packaging to customers who want to be part of a closed beta but can't actually install remote plugins.)

Phrynobatrachus (approved):

PR #19935, opened 2022-04-07
MM-43137: Fix racy test TestUpdateActiveBotsSideEffect

agnivade:
This was introduced with https://github.com/mattermost/mattermost-server/commit/2e027ae9278d60f318e7c286fcbe5c88895cc23f.

We fix it by deep copying the user struct before sending
it to the websocket hub.

Sample failures:
https://app.circleci.com/pipelines/github/mattermost/mattermost-server/32721/workflows/50d2ac52-4ff8-4116-ab50-7eae7a3306e8/jobs/257286
https://app.circleci.com/pipelines/github/mattermost/mattermost-server/32721/workflows/50d2ac52-4ff8-4116-ab50-7eae7a3306e8/jobs/257287

https://mattermost.atlassian.net/browse/MM-43137

```release-note
NONE
```

### DIFF
--- a/app/user.go
+++ b/app/user.go
@@ -912,7 +912,7 @@ func (a *App) UpdateActive(c *request.Context, user *model.User, active bool) (*
 			return nil, model.NewAppError("UpdateActive", "app.user.update.finding.app_error", nil, err.Error(), http.StatusInternalServerError)
 		}
 	}
-	ruser := userUpdate.New
+	ruser := userUpdate.New.DeepCopy()
 
 	if !active {
 		if err := a.RevokeAllSessions(ruser.Id); err != nil {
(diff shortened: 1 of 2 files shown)


app/user.go
```diff
@@ -912,7 +912,7 @@ func (a *App) UpdateActive(c *request.Context, user *model.User, active bool) (*
 			return nil, model.NewAppError("UpdateActive", "app.user.update.finding.app_error", nil, err.Error(), http.StatusInternalServerError)
 		}
 	}
-	ruser := userUpdate.New
+	ruser := userUpdate.New.DeepCopy()
```
isacikgoz, line 915: This seem to solve the issue but, I think returning new user in the old users pointer isn't correct. Not sure if it would be part of this PR anyway. Because, semantically I would expect the new one to have a fresh struct and I probably wouldn't do anything with the old one. WDYT?

agnivade: Agree. Makes sense.

agnivade:
Interesting, got a flaky test on fixing another racy test :P

isacikgoz:
bryan-cranston-replace-the-bulb

isacikgoz (approved):

PR #19939, opened 2022-04-07
MM-42725: exclude channels from teams user has left

koox00:
#### Summary

Currently the channel switcher will display channels even from the teams you have left.

This commit excludes channels from "left" teams to be shown in the switcher.

PS: When a user leaves a team we soft-delete the TeamMembers entry for user-team.

#### Ticket Link

https://mattermost.atlassian.net/browse/MM-42725

#### Release Note

```release-note
Fixes channel switcher showing channels from team the user is no longer part of
```

### DIFF
--- a/store/sqlstore/channel_store.go
+++ b/store/sqlstore/channel_store.go
@@ -2893,6 +2893,7 @@ func (s SqlChannelStore) Autocomplete(userID, term string, includeDeleted bool) 
 		Where(sq.And{
 			sq.Expr("c.TeamId = t.id"),
 			sq.Expr("t.id = tm.TeamId"),
+			sq.Eq{"tm.DeleteAt": 0},
 			sq.Eq{"tm.UserId": userID},
 			sq.Or{
 				sq.NotEq{"c.Type": model.ChannelTypePrivate},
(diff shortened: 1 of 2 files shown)


koox00:
ha... I just saw that we have a flag for that....

amyblais:
@agnivade @ashishbhate Maybe we can get this in for today's code freeze.

agnivade (commented):
Let's add a test and this is good to go :)

ashishbhate (approved):
LGTM!

Approving to not block PR incase this needs to go in. As Agniva said, a test would be good to have.

koox00:
> Hmm .. check your DB? The query seems straighforward enough

Yes I know... :/ Does `Team().RemoveMembers()` soft delete?

koox00:
> Yes I know... :/ Does `Team().RemoveMembers()` soft delete?

I'll answer my own question :), no! So I use `Team().UpdateMember()` to soft delete, and I can fail/pass the test

koox00:
> I am thinking it might be good to match with that flag and accordingly add c.DeleteAt=0 and tm.DeleteAt=0 as well.

I'm not against, I'm not sure what the expectations are when `includeDeleted` is `true`.

agnivade (approved):
Nicely done! And thanks for adding the test :heart:

PR #19953, opened 2022-04-08
[MM-42739] Insights - Top Channels API Endpoint

mylonsuren:
#### Summary

Adds new endpoint to get top channels for a team and user.

This adds two new endpoints:
 `GET /api/v4/teams/:team_id/top/channels` and `GET /api/v4/users/me/top/channels`.
 
Top channels for a user can also be scoped to a team by adding the `team_id` as a parameter to the request.

##### Notes: 
- This also removes the `rank` field that was added as part of https://github.com/mattermost/mattermost-server/pull/19850 as it is no longer required.
- Another PR will be created to support top channels by date/time to populate the line graph of top channels over a given period of time.

##### Merge After: 
- https://github.com/mattermost/mattermost-server/pull/19850

#### Sample Requests

Top Channels for a Team:
```
http://localhost:8065/api/v4/teams/{{team_id}}/top/channels?time_range=today&page=0&per_page=5
```
```
{
    "has_next": false,
    "items": [
        {
            "id": "w8nmi8a6mfdf8nogy5h39i585o",
            "type": "O",
            "display_name": "ut",
            "name": "ratione-1",
            "team_id": "bwxhd41titrujmoqyo1ei457wh",
            "message_count": 142
        },
        {
            "id": "nkaxhuxoffd978xqhk9f458wka",
            "type": "P",
            "display_name": "nesciunt",
            "name": "sequi-7",
            "team_id": "bwxhd41titrujmoqyo1ei457wh",
            "message_count": 136
        },
        {
            "id": "esf99mqbsbb7bre8jo6ahpe4be",
            "type": "O",
            "display_name": "Town Square",
            "name": "town-square",
            "team_id": "bwxhd41titrujmoqyo1ei457wh",
            "message_count": 19
        }
    ]
}
```
Top Channels for a User (Scoped to Team):
```
http://localhost:8065/api/v4/users/me/top/channels?time_range=today&page=0&per_page=5&team_id={{team_id}}
```
```
{
    "has_next": false,
    "items": [
        {
            "id": "esf99mqbsbb7bre8jo6ahpe4be",
            "type": "O",
            "display_name": "Town Square",
            "name": "town-square",
            "team_id": "bwxhd41titrujmoqyo1ei457wh",
            "message_count": 19
        },
        {
            "id": "w9urfcfxat8ju81n9na7i4mmuy",
            "type": "O",
            "display_name": "nostrum",
            "name": "minus-6",
            "team_id": "bwxhd41titrujmoqyo1ei457wh",
            "message_count": 2
        },
        {
            "id": "nkaxhuxoffd978xqhk9f458wka",
            "type": "P",
            "display_name": "nesciunt",
            "name": "sequi-7",
            "team_id": "bwxhd41titrujmoqyo1ei457wh",
            "message_count": 1
        },
        {
            "id": "o3asuydzdbnnzkinzox7f48m5h",
            "type": "O",
            "display_name": "sint",
            "name": "suscipit-4",
            "team_id": "bwxhd41titrujmoqyo1ei457wh",
            "message_count": 1
        }
    ]
}
```

#### Ticket Link

Fixes [MM-42739](https://mattermost.atlassian.net/browse/MM-42739)

#### Release Note

```release-note
Added new API endpoints:  GET /api/v4/teams/:team_id/top/channels and GET /api/v4/users/me/top/channels
```


api4/channel.go
```diff
@@ -2094,3 +2097,87 @@ func moveChannel(c *Context, w http.ResponseWriter, r *http.Request) {
...
+		team, teamErr := c.App.GetTeam(c.Params.TeamId)
+		if teamErr != nil {
+			c.Err = teamErr
+			return
+		}
+
+		if (!team.AllowOpenInvite || team.Type != model.TeamOpen) && !c.App.SessionHasPermissionToTeam(*c.AppContext.Session(), team.Id, model.PermissionViewTeam) {
+			c.SetPermissionError(model.PermissionViewTeam)
+			return
+		}
```
mkraft, line 2162: Is there a reason for a different authz check here compared to [in the `getTopChannelsForTeamSince ` function](https://github.com/mattermost/mattermost-server/pull/19953/files#diff-036c0265e2c7ac93563884c5b029665ae090d47443df092263ce1cd619585246R2113-R2116)?

mkraft (changes requested):
Could we please add a test that checks the `view_team` permission for each of the new API endpoints?

Also, I don't think I saw a test that verifies channels that a user is not a member of are excluded, but I could have missed that.

api4/channel.go
mylonsuren: There's no reason other than I missed changing it here as well 😅, thanks for catching that.

mkraft (changes requested):
Could we put these into an api4/insights.go file?

mylonsuren:
> Could we put these into an api4/insights.go file?

Yes, I can do that now 👍

mkraft (changes requested):
I know [it was from a previous PR](https://github.com/mattermost/mattermost-server/blob/master/web/params.go#L258) but I think it's a bit confusing that the `time_range` parameter is being converted from a string to a timestamp as it's being marshalled. I think it would be more obvious to invoke `GetStartUnixMilliForTimeRange` on  `time_range` [at the point where it needs the conversion](https://github.com/mattermost/mattermost-server/pull/19953/files#diff-036c0265e2c7ac93563884c5b029665ae090d47443df092263ce1cd619585246R2119). I noticed it when I tried printing the value of `c.Params.TimeRange`.

mylonsuren:
> I know [it was from a previous PR](https://github.com/mattermost/mattermost-server/blob/master/web/params.go#L258) but I think it's a bit confusing that the `time_range` parameter is being converted from a string to a timestamp as it's being marshalled. I think it would be more obvious to invoke `GetStartUnixMilliForTimeRange` on `time_range` [at the point where it needs the conversion](https://github.com/mattermost/mattermost-server/pull/19953/files#diff-036c0265e2c7ac93563884c5b029665ae090d47443df092263ce1cd619585246R2119). I noticed it when I tried printing the value of `c.Params.TimeRange`.
> 
> What do you think? If you think it's best where it is I'm cool with that.

Yeah I had originally done it this way where the conversion only happens once we need it in that in that timestamp form. There was a comment on that PR suggesting that we do at the start where it's being marshalled to simplify the code and remove some duplication. 

I agree that for debugging purposes it could cause confusion as you would never be able to get the string value for `time_range` (only the `StartUnixMilli` timestamp that it's converted to).

store/sqlstore/channel_store.go
```diff
@@ -4097,3 +4097,159 @@ func (s SqlChannelStore) GetTeamForChannel(channelID string) (*model.Team, error
...
+	postgresPropQuery := `AND (Posts.Props ->> 'from_bot' IS NULL OR Posts.Props ->> 'from_bot' = 'false')`
+	mySqlPropsQuery := `AND (JSON_EXTRACT(Posts.Props, '$.from_bot') IS NULL OR JSON_EXTRACT(Posts.Props, '$.from_bot') = 'false')`
+
+	query := `
+		SELECT
+			ID,
+			Type,
+			DisplayName,
+			Name,
+			TeamID,
```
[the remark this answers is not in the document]
mylonsuren: I added it as part of the idea that the user could get top channels across the workspace by omitting the `team_id` parameter. I'm happy to remove it if you feel it's not required at this time.

store/sqlstore/channel_store.go
```diff
@@ -4097,3 +4097,159 @@ func (s SqlChannelStore) GetTeamForChannel(channelID string) (*model.Team, error
...
+	if s.DriverName() == model.DatabaseDriverMysql {
+		query += mySqlPropsQuery
+	} else if s.DriverName() == model.DatabaseDriverPostgres {
+		query += postgresPropQuery
+	}
+
+	query += `
+				AND Channels.DeleteAt = 0
+				AND Channels.TeamId = ?
+				AND Channels.Type = 'O'
```
mkraft, line 4144: I have to keep digging, but for some reason if I comment-out any one of these three `where` clauses on the `Channels` table the query time decreases from about 6s to <1s on MySQL in our large test dataset.

mkraft: For some reason putting the `WHERE DeleteAt = 0` in the outer query speeds thing up a lot:

```diff
SELECT
    ID,
    Type,
    DisplayName,
    Name,
    TeamID,
    MessageCount
FROM
    (
        (
            SELECT
                Posts.ChannelId AS ID,
                Channels.Type AS Type,
                Channels.DisplayName AS DisplayName,
                Channels.Name AS Name,
                Channels.TeamId AS TeamID,
                count(Posts.Id) AS MessageCount,
+                Channels.DeleteAt AS DeleteAt
            FROM
                Posts
                LEFT JOIN Channels on Posts.ChannelId = Channels.Id
            WHERE
                Posts.DeleteAt = 0
                AND Posts.CreateAt > 1635996172311
                AND Posts.Type = ''
                AND (
                    JSON_EXTRACT(Posts.Props, '$.from_bot') IS NULL
                    OR JSON_EXTRACT(Posts.Props, '$.from_bot') = 'false'
                )
-                AND Channels.DeleteAt = 0
                AND Channels.TeamId = 'd6z6ei9gmpdyfys5ay7opket8o'
                AND Channels.Type = 'O'
            GROUP BY
                Posts.ChannelId,
                Channels.Type,
                Channels.DisplayName,
                Channels.Name,
                Channels.TeamId,
                Channels.DeleteAt
        )
        UNION
        ALL(
            SELECT
                Posts.ChannelId AS ID,
                Channels.Type AS Type,
                Channels.DisplayName AS DisplayName,
                Channels.Name AS Name,
                Channels.TeamId AS TeamID,
                count(Posts.Id) AS MessageCount,
+                Channels.DeleteAt AS DeleteAt
            FROM
                Posts
                LEFT JOIN Channels on Posts.ChannelId = Channels.Id
                LEFT JOIN ChannelMembers on Posts.ChannelId = ChannelMembers.ChannelId
            WHERE
                Posts.DeleteAt = 0
                AND Posts.CreateAt > 1635996172311
                AND Posts.Type = ''
                AND (
                    JSON_EXTRACT(Posts.Props, '$.from_bot') IS NULL
                    OR JSON_EXTRACT(Posts.Props, '$.from_bot') = 'false'
                )
-                AND Channels.DeleteAt = 0
                AND Channels.TeamId = 'd6z6ei9gmpdyfys5ay7opket8o'
                AND Channels.Type = 'P'
                AND ChannelMembers.UserId = '116ny8fhbb8fbfdeauq73t591o'
            GROUP BY
                Posts.ChannelId,
                Channels.Type,
                Channels.DisplayName,
                Channels.Name,
                Channels.TeamId,
                Channels.DeleteAt
        )
    ) AS A
+WHERE DeleteAt = 0
ORDER BY
    MessageCount DESC,
    Name ASC
LIMIT
    61 OFFSET 0;
```

mkraft: btw, I'm totally OK with optimizing this query (if needed) later and moving forward as-is, assuming logical correctness.

store/sqlstore/channel_store.go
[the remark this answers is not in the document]
mkraft: No, that's cool. Just making sure we had a reason.

mylonsuren:
@mkraft I think I've addressed all your comments, should be good for another review :+1:

mylonsuren:
> Could we please add a test that checks the `view_team` permission for each of the new API endpoints?
> 
> Also, I don't think I saw a test that verifies channels that a user is not a member of are excluded, but I could have missed that.

These tests have been added as well, for both top reactions and top channels.

mkraft (approved):

BenCookie95:
Everything looks great but I wonder if we should return the full channel object? Would that require extra joins? Might make it easier for redux because I can pop the full channel object into `state.entities.channels.channels`

mylonsuren:
> Everything looks great but I wonder if we should return the full channel object? Would that require extra joins? Might make it easier for redux because I can pop the full channel object into `state.entities.channels.channels`

@BenCookie95 It wouldn't require extra joins or anything, but wouldn't all the channels already exist in state when the user first loads the page? Do you need the full object for linking or another reason?

BenCookie95:
> > Everything looks great but I wonder if we should return the full channel object? Would that require extra joins? Might make it easier for redux because I can pop the full channel object into `state.entities.channels.channels`
> 
> @BenCookie95 It wouldn't require extra joins or anything, but wouldn't all the channels already exist in state when the user first loads the page? Do you need the full object for linking or another reason?

Not necessarily because the team insights will return public channels that you are not a member of

BenCookie95:
> > > Everything looks great but I wonder if we should return the full channel object? Would that require extra joins? Might make it easier for redux because I can pop the full channel object into `state.entities.channels.channels`
> > 
> > 
> > @BenCookie95 It wouldn't require extra joins or anything, but wouldn't all the channels already exist in state when the user first loads the page? Do you need the full object for linking or another reason?
> 
> Not necessarily because the team insights will return public channels that you are not a member of

Looks like there would be a performance issue with grabbing all the channel data as we're now using the PublicChannels table, so leaving it as is.

Thanks @mylonsuren for taking a look

BenCookie95 (approved):

mylonsuren:
@furqanmlk This also has changes to top reactions as well, so if you can please test both top reactions and top channels for this PR.

furqanmlk (approved):
Thanks @mylonsuren. Tested ```Top Reaction```  and ```Top Channel```
Working as expected.

PR #19962, opened 2022-04-11
MM-41565,MM-42473: fixes incorrect unread thread count

koox00:
#### Summary

TL;DR
The infamous "-1 unread threads" bug, was due to incorrectly counting
previous unread mentions and replies.

When we calculate the previous counts, we didn't take into account if
a thread was newly followed one or not. So when a user gets added to a
thread via a mention receives a WS event with the previous count values
calculated by subtracting "1" from the current count values.

In result the webapp thinks that already has this unread thread calculated
in the `total_unread_threads` count, thus not incrementing it ([L70-L71](https://github.com/mattermost/mattermost-webapp/blob/83c54107edac8301e5359872253b6148d835743d/packages/mattermost-redux/src/reducers/entities/threads/counts.ts#L70-L71)).  
Essentially showing that you have NO unread threads (MM-42473).
Now if you view the received thread the webapp will correctly reduce
the `total_unread_threads` cout by 1 and essentially showing that you
have -1 unread threads (MM-41565). 

Since the user just followed the thread those previous values
should be 0, so then the thread will be correctly marked as unread
thus solving both bugs.

#### Ticket Link

https://mattermost.atlassian.net/browse/MM-41565
https://mattermost.atlassian.net/browse/MM-42473

#### Release Note

```release-note
Fixes a bug where it would appear that a user has a negative number of unread threads
```


app/notification.go
```diff
@@ -246,9 +247,6 @@ func (a *App) SendNotifications(post *model.Post, team *model.Team, channel *mod
 					}
 
 					if membership != nil && !membership.Following {
-						membershipsMutex.Lock()
-						participantMemberships[userID] = membership
-						membershipsMutex.Unlock()
```
koox00, line 251: It seems that these are not needed

ashishbhate: I think this was done to prevent an extra store call later in this function. 
https://github.com/mattermost/mattermost-server/blob/090c4242c0ab5a5f780a14b275b0516d1633ac5a/app/notification.go#L581-L588

The loop there loops over `followers` so the profiles here definitely wont be required there.

app/notification.go
```diff
@@ -595,11 +594,19 @@ func (a *App) SendNotifications(post *model.Post, team *model.Team, channel *mod
...
 				if userThread != nil {
-					previousUnreadMentions := userThread.UnreadMentions
-					previousUnreadReplies := max(userThread.UnreadReplies-1, 0)
-					if mentions.isUserMentioned(uid) {
-						previousUnreadMentions = max(userThread.UnreadMentions-1, 0)
+					previousUnreadMentions := int64(0)
+					previousUnreadReplies := int64(0)
+
+					// if it's not a newly followed thread, calculate previous unread values.
+					if _, ok := newParticipants[uid]; !ok {
```
agnivade, line 601: Nit: this can be just simplified to `if !newParticipants[uid] {`. A map value returns a default value of the key if the key doesn't exist. https://go.dev/blog/maps#exploiting-zero-values

agnivade (commented):
Possible to add a test for this? Seems like sensitive logic.

api4/user_test.go
```diff
@@ -6116,6 +6116,53 @@ func TestThreadSocketEvents(t *testing.T) {
...
+						require.Equal(t, int64(1), thread.UnreadMentions)
+					}
+				case <-time.After(1 * time.Second):
+					return
+				}
+			}
+		}()
+
+		require.Equalf(t, 1, count, "User should have received 1 %s event", model.WebsocketEventThreadUpdated)
+		require.Truef(t, caught, "User should have received %s event", model.WebsocketEventThreadUpdated)
```
agnivade, line 6164: We are doing count++ only for the `WebsocketEventThreadUpdated` event. Is `caught` necessary anymore?

agnivade (approved):

api4/user_test.go
koox00: 😅 not needed at all, will remove!

ashishbhate (approved):
💯

jgilliam17:
Thank you @koox00 
Tested, looks good.
- Verified global threads bolding, mentions and unread count on the inbox graphic - all as expected.
- [E2E report](https://automation-dashboard.vercel.app/cycles/715556e9-855e-4da5-8817-4be207713c2e) shows few CRT tests failing.

koox00:
> [E2E report](https://automation-dashboard.vercel.app/cycles/715556e9-855e-4da5-8817-4be207713c2e) shows few CRT tests failing. Can you please take a look. Thanks 🙂

@jgilliam17 locally they are all passing expect this one `MM-T4682 should show search guidance at the end of the list after scroll loading` however an enterprise CTA modal, or the thread tour tooltip, are interfering with the tests.

About `MM-T4682`, I would suspect that this fails on master as well. It should be related with the recent "decoupling of counts from threads" PR.

I'd like to fix those both of these independently.

jgilliam17:
Thanks @koox00 looks good to merge.
[Latest E2E report](https://mattermost-cypress-report.s3.amazonaws.com/183412-c4a3777-onprem-ent-server-pr-19962-mattermost/mm-ee-testc4a3777/mochawesome.html) looks good, no PR related failures.

jgilliam17 (approved):

PR #19988, opened 2022-04-14
[MM-43393]: replace .Decoder() method with .Token()

michelengelen:
#### Summary
title says it all.
This might need more polishing from someone more proficient in GO.

I commented out and left the block of code where it was checking with regexes for occurences of `viewbox` or `width`/`height` values as a fallback, since afaik a `SVG` with those attributes missing is broken anways.

#### Ticket Link
[MM-43393](https://mattermost.atlassian.net/browse/MM-43393)

#### Release Note
```release-note
NONE
```

### DIFF
--- a/app/imaging/svg.go
+++ b/app/imaging/svg.go
@@ -39,18 +39,17 @@ func ParseSVG(svgReader io.Reader) (SVGInfo, error) {
 					values := strings.Fields(attr.Value)
 					if len(values) == 4 {
 						width := 0
-						_, err := fmt.Sscan(values[2], &width)
-
-						if err != nil {
-							svgInfo.Width = width
-						}
+						_, widthErr := fmt.Sscan(values[2], &width)
 
 						height := 0
-						_, err = fmt.Sscan(values[3], &height)
+						_, heightErr = fmt.Sscan(values[3], &height)
 
-						if err != nil {
-							svgInfo.Height = height
+						if widthErr != nil || heightErr != nil {
+							return svgInfo, err
 						}
+
+						svgInfo.Width = width
+						svgInfo.Height = height
 
 						return svgInfo, nil
 					}
--- a/app/imaging/svg.go
+++ b/app/imaging/svg.go
@@ -42,7 +42,7 @@ func ParseSVG(svgReader io.Reader) (SVGInfo, error) {
 						_, widthErr := fmt.Sscan(values[2], &width)
 
 						height := 0
-						_, heightErr = fmt.Sscan(values[3], &height)
+						_, heightErr := fmt.Sscan(values[3], &height)
 
 						if widthErr != nil || heightErr != nil {
 							return svgInfo, err
--- a/app/imaging/svg.go
+++ b/app/imaging/svg.go
@@ -5,8 +5,8 @@ package imaging
 
 import (
 	"encoding/xml"
-	"io"
 	"fmt"
+	"io"
 	"strings"
 
 	"github.com/pkg/errors"
@@ -109,6 +109,4 @@ func ParseSVG(svgReader io.Reader) (SVGInfo, error) {
 
 	// commented out the above code since IMO this is not needed anymore with the approach using decoder.Token(),
 	// but for the case it is still valid to do it I just commented it out
-
-	return svgInfo, errors.New("unable to extract SVG dimensions")
 }
--- a/app/imaging/svg.go
+++ b/app/imaging/svg.go
@@ -6,8 +6,9 @@ package imaging
 import (
 	"encoding/xml"
 	"io"
-	"strconv"
+	"fmt"
 	"strings"
+
 	"github.com/pkg/errors"
 )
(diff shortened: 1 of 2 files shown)


jupenur (approved):
Can't comment on code quality or anything, but I verified this fixes the issue, so LGTM. Thanks Michel!

amyblais:
@mkraft @furqanmlk This is a backport candidate.

mkraft:
> @mkraft I updated the description to accommodate for the commented out fallback

If we're not going to use it can we remove the commented-out code?

mkraft:
Adding @hmhealey as a second set of eyes on whether we're OK to remove the viewbox stuff.

michelengelen:
> > @mkraft I updated the description to accommodate for the commented out fallback
> 
> If we're not going to use it can we remove the commented-out code?

We could, we still have `git` if we need to re-insert it, right? :D

mkraft:
Do we need the outer `for` loop looking for `xml.StartElement`? If it doesn't find that—with the current code—will it crash the server? Is `xml.StartElement` always the first element?

app/imaging/svg.go
```diff
@@ -20,39 +20,68 @@ type SVGInfo struct {
...
+		case xml.StartElement:
+			for _, attr := range t.Attr {
+				if attr.Name.Local == "viewBox" {
+					values := strings.Fields(attr.Value)
+					if len(values) == 4 {
+						width := 0
+						_, err := fmt.Sscan(values[2], &width)
+
+						if err != nil {
+							svgInfo.Width = width
```
mkraft, line 45: Are you meaning to not ignore the errors in the `"viewBox"` case?

jupenur:
> Do we need the outer for loop looking for xml.StartElement? If it doesn't find that—with the current code—will it crash the server? Is xml.StartElement always the first element?

Yes that's necessary, and no, not finding one shouldn't crash the server. The first *element* will always be a `StartElement`, but the first *token* is not necessarily an *element*. If there is no element, `decoder.Token()` will eventually return an EOF (or another error).

Might be a good idea to add a test case with an SVG file containing only a single `ProcInst` token, though!

app/imaging/svg.go
```diff
@@ -20,39 +20,67 @@ type SVGInfo struct {
...
+
+						svgInfo.Width = width
+						svgInfo.Height = height
+
+						return svgInfo, nil
+					}
+				}
+				if attr.Name.Local == "width" {
+					width := 0
+					_, err := fmt.Sscan(attr.Value, &width)
```
hmhealey, line 59: I think that stuff that was removed was because the width/height properties could theoretically be `100px` instead of `100`. I don't know if that's actually the case, but maybe we can confirm that? Otherwise, it'll probably return an error here and hopefully not panic

hmhealey: Oh, never mind. I didn't realize that using `fmt.Sscan` to parse the number just goes as long as it finds a number like the `parseInt` in JavaScript, so this should still work without the regex

Screen Shot 2022-04-21 at 9 29 30 AM

hmhealey (approved):
Looks good to me

michelengelen:
@mkraft I have added a test case for an svg with a single `ProcInst` Token. Could you have a look and see if that one is OK?

app/imaging/svg_test.go
```diff
@@ -74,3 +74,11 @@ func TestParseInvalidSVGData(t *testing.T) {
 		}
 	}
 }
+
+func TestParseProcInstOnlySVGData(t *testing.T) {
+	svg := strings.NewReader("<?xml version='1.0' encoding='utf-8'?>")
+	svgInfo, err := ParseSVG(svg)
+	if err == nil || svgInfo.Width != 0 || svgInfo.Height != 0 {
+		t.Errorf("Should not be able to parse SVG attributes, but was definitely able to!")
+	}
```
mkraft, line 83: Good. I would recommend:

```go
require.NoError(t, err)
require.Equal(t, 0, svgInfo.Width)
require.Equal(t, 0, svgInfo.Height)
```

mkraft (approved):
Just one recommendation for the new test's assertions, but otherwise LGTM.

amyblais:
@michelengelen Do we need help from devops or SRE to look at the failing CI?

michelengelen:
@amyblais I dont think so ... once i find out why the linter fails i think the others will become green as well

jupenur:
@michelengelen tabs vs spaces in the unit test :)

michelengelen:
> @michelengelen tabs vs spaces in the unit test :)

LOL ... the linter is pretty bitchy! 🐱

amyblais:
@furqanmlk This is ready for testing.

michelengelen:
@furqanmlk kind reminder to approve. :D

furqanmlk (approved):
Thanks @michelengelen. Working as expected and also E2E passed.

PR #19993, opened 2022-04-14
[MM-43253] Allow per connection WebSocket broadcasts

streamer45:
#### Summary

In Calls we have many messages that are really only meant to be sent to a single connection. Right now we are forced to broadcast to all connected sessions causing both bandwidth and processing time to go wasted. 

PR extends `model.WebsocketBroadcast` with a `ConnectionId` field to address this. On the hub side we are adding a `byConnectionId` map to the hub index to easily filter by connection id.

#### Ticket Link

https://mattermost.atlassian.net/browse/MM-43253

#### Release Note

```release-note
Added a new ConnectionId field to model.WebsocketBroadcast that allows broadcasting a message only to a specific connection.
```


agnivade:
> Right now we are forced to broadcast to all connected sessions

Just to clarify, all connected sessions for a given call or for that user?

app/web_hub.go
```diff
@@ -539,13 +539,20 @@ func (h *Hub) Start() {
 						}
 					}
 				}
-				if msg.GetBroadcast().UserId != "" {
+
+				if connID := msg.GetBroadcast().ConnectionId; connID != "" {
```
agnivade, line 543: I am wondering what is the difference between adding the connectionId as part of the connIndex versus adding the subscription topic to the index. Any reason to do this, but not do the other? Both seem to reduce iteration time, and both would be accepted/rejected anyways in the web_conn goroutine.

app/web_hub.go
```diff
@@ -539,13 +539,20 @@ func (h *Hub) Start() {
...
+					if webConn := connIndex.byConnectionId[connID]; webConn != nil {
+						broadcast(webConn)
```
agnivade, line 545: Maybe better to change the function name to `send`? It's somewhat misleading.

agnivade (commented):
Can we have a load-test for this? :)

streamer45:
> > Right now we are forced to broadcast to all connected sessions
> 
> Just to clarify, all connected sessions for a given call or for that user?

That would be to the same user, like having multiple tabs open or apps/devices connected to the same Mattermost installation.
Also, the way things are implemented now, we always have a dedicated websocket connection when in a call so even the same tab is two sessions but only one is meant to receive certain messages (e.g. signaling data).

app/web_hub.go
streamer45: Happy to change but how does this differ from previous occurrences? Even the channel is called `broadcast` and we are always dealing with a single `WebConn`.

app/web_hub.go
streamer45: I think the main difference is semantic. This is meant to be a lower level addition than topics. We are extending the existing API for comfort but technically we could go one step further and implement a `WebSocketSend()` method and allow the plugin side to send whatever data, not constrained by the broadcast struct format. 

The case I am trying to cover is one where I don't necessarily want clients to subscribe to anything but at the same time I need full control (server-side) on which connection the data is targeting.

app/web_hub.go
agnivade: Hmm .. fair enough :P

agnivade (approved):

cpoile (approved):
👍  Looks good!

streamer45:
Initial comparison looks good:

streamer45:
Second round has completed. As expected no change on overall performance. Websocket events slightly decreased:

image

The load-test was conducted with 2 sessions per user and a customized version of Calls that sent broadcasts every few seconds to every connection.

Full data at http://18.207.190.169:3000/d/fHVpkVQ7z/comparison-0-postgresql-bounded

agnivade:
Nice work!

streamer45:
@metanerd Do you know where I can find the report?

metanerd:
@streamer45 Yipyip: https://community-daily.mattermost.com/core/pl/amp7adac67ygueagmmek61q1ph

streamer45:
Awesome, thanks. Report looks good to me. The 6 failures look totally unrelated to the proposed changes.

streamer45:
@amyblais Do you know when we last merged master in cloud for 6.7 release? I somewhat expected this to be in.

amyblais:
> @amyblais Do you know when we last merged master in cloud for 6.7 release? I somewhat expected this to be in.

Monday, April 18th was the feature complete / final fast-forward date for v6.7 https://community-daily.mattermost.com/core/pl/cenmhk1zxpnc3q8f6aisjdn3fe.

PR #20014, opened 2022-04-18
[DOPS-920] Fix mmctl download bash script

phoinixgrr:
#### Summary

There are cases that instead of suffix, we get a SHA prefix for cherrypicking branch.
For example this [PR](https://github.com/mattermost/mattermost-server/pull/19906) in which a manual cherrypicking process has started. (because of Mattermod failure to auto-cherrypick due to Conflicts) 

We rewrite the regex to support this case

#### Ticket Link

Ticket: https://mattermost.atlassian.net/browse/DOPS-920

#### Release Note

```release-note

```

### DIFF
--- a/scripts/download_mmctl_release.sh
+++ b/scripts/download_mmctl_release.sh
@@ -19,7 +19,7 @@ BIN_PATH=${2:-bin}
 THIS_BRANCH=$(git rev-parse --abbrev-ref HEAD)
 if [[ "$THIS_BRANCH" =~ 'release-'[0-9] ]];
 then
-  RELEASE_TO_DOWNLOAD=$(echo $THIS_BRANCH | grep -Eo 'release-.*')
+  RELEASE_TO_DOWNLOAD=$(echo $THIS_BRANCH | grep -Eo 'release-([0-9](\.){0,1})\.([0-9](\.){0,1})')
 else
   RELEASE_TO_DOWNLOAD=master
 fi


pfltdv (approved):
🏅 nice work, LGTM.

Sorry I missed one test case, do we need to escape the dot ? (Will add comment to the line)
```
echo release-6a11 | grep -Eo 'release-([0-9](\.){0,1}).([0-9](\.){0,1})'
release-6a1
```

scripts/download_mmctl_release.sh
```diff
@@ -19,7 +19,7 @@ BIN_PATH=${2:-bin}
 THIS_BRANCH=$(git rev-parse --abbrev-ref HEAD)
 if [[ "$THIS_BRANCH" =~ 'release-'[0-9] ]];
 then
-  RELEASE_TO_DOWNLOAD=$(echo $THIS_BRANCH | grep -Eo 'release-.*')
+  RELEASE_TO_DOWNLOAD=$(echo $THIS_BRANCH | grep -Eo 'release-([0-9](\.){0,1}).([0-9](\.){0,1})')
```
pfltdv, line 22:
```suggestion
  RELEASE_TO_DOWNLOAD=$(echo $THIS_BRANCH | grep -Eo 'release-([0-9](\.){0,1})\.([0-9](\.){0,1})')
```
Do we need to escape the dot in the middle ?

phoinixgrr: It doesn't make a world of difference, but it is better escaping it. 👍 
Thanks for the comment @pfltdv 

Rebased and pushed the change

pfltdv (approved):
LGTM

spirosoik (approved):
great 🍻

spirosoik:
something though is odd with the CI here.

pfltdv:
@spirosoik @phoinixgrr I've added `EETests` label to initiate tests. Mattermod started tests after removing that label. Currently EE tests are in progress.

spirosoik:
@pfltdv I think for these cases probably we should not trigger `EETests` at all, right? I think we need to check that if we change bash scripts to not have these jobs as required.

pfltdv:
@spirosoik yes that is very good point, and for the unrelated changes at application. (Changes in bash scripts, md files, etc.). We do not need to trigger EEtests. And mattermod or CI should make related checks green. Shall we create a ticket to implement those. (We also need to get confirmation from developers about the exclusion files and folders)

spirosoik:
@pfltdv let's create a medium priority ticket

pfltdv:
Related ticket https://mattermost.atlassian.net/browse/DOPS-949

PR #20027, opened 2022-04-19
POC: Cross-team recent search

agnivade:
Just a quick POC to move fast :P

We use a search pointer to keep track of
the next row to inesrt to. For every new search
we increment the pointer and do modulo 5.
This means that the value will always remain
between 0-4. And that way, we will always overwrite
the oldest entry on every search.

And while getting the results, we get
all results for that user.

The search parameters are json marshalled
and stored as a JSON blob. This is because
there is no need to search/filter them
in the DB.

Pending items:
Tests obviously.

To improve:
The client needs to send the channel ids
instead of channel names.

```release-note
NONE
```


agnivade:
Sent it!

agnivade:
All tests are passing now.

store/sqlstore/post_store.go
```diff
@@ -2529,11 +2533,22 @@ func (s *SqlPostStore) SearchPostsForUser(paramsList []*model.SearchParams, user
...
+		// prevent surprises in UI.
+		buf, err := json.Marshal(params)
+		if err != nil {
+			return nil, err
+		}
+		err = s.LogRecentSearch(userID, buf, now)
+		if err != nil {
+			return nil, err
+		}
+
```
srkgupta, line 2551: Looks like this logic iterates through each parameter passed in the API and then stores it as Recent Search. Shouldn't the logic be once per API?

store/sqlstore/post_store.go
```diff
@@ -2565,6 +2580,103 @@ func (s *SqlPostStore) SearchPostsForUser(paramsList []*model.SearchParams, user
...
+	return nil
+}
+
+func (s *SqlPostStore) GetRecentSearchesForUser(userID string) ([]*model.SearchParams, error) {
+	params := [][]byte{}
+	err := s.GetReplicaX().Select(&params, `SELECT query
+		FROM RecentSearches
+		WHERE UserId=?
+		ORDER BY CreateAt DESC`, userID)
```
srkgupta, line 2665: Looks like we are performing a query on the database without any upper limits. This might be performance intensive if a user creates multiple searches. Can we have a limit on max number of rows?

db/migrations/postgres/000081_recent_searches.up.sql
```diff
@@ -0,0 +1,7 @@
+CREATE TABLE IF NOT EXISTS recentsearches (
```
srkgupta, line 1: Just curious to know: 
How do we handle a case when there is another feature which requires DB migrations and also has the sequence `000081` and gets merged before this PR?

store/sqlstore/post_store.go
agnivade: Discussed offline. ParamsList is a list of queries, and each item contains the struct containing all the params.

db/migrations/postgres/000081_recent_searches.up.sql
agnivade: The sequence number isn't really important. It's just a human readable number to distinguish. The library will see which all migrations haven't been applied and apply those. But only when you need to enforce order between multiple migrations, then first it sorts by sequence numbers, and then by the name. So if 2 people send 2 PRs, but there is a dependency between the 2 migrations, then one needs to wait for the other. The tool doesn't know this and is left at the user layer to resolve.

store/sqlstore/post_store.go
ashishbhate, line 2665: Can we use the query builder?

api4/post.go
```diff
@@ -31,6 +31,7 @@ func (api *API) InitPost() {
 
 	api.BaseRoutes.Team.Handle("/posts/search", api.APISessionRequiredDisableWhenBusy(searchPostsInTeam)).Methods("POST")
 	api.BaseRoutes.Posts.Handle("/search", api.APISessionRequiredDisableWhenBusy(searchPostsInAllTeams)).Methods("POST")
+	api.BaseRoutes.User.Handle("/recent_searches", api.APISessionRequiredDisableWhenBusy(getRecentSearches)).Methods("GET")
```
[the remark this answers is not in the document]
agnivade: Good point.

store/sqlstore/post_store.go
agnivade: The convention we follow is that for simple queries like the above, there is no need to use squirrel as raw SQLs are much more readable this way. For complicated SQLs where you need to form it with different conditions, there of course you need squirrel.

srkgupta (changes requested):
Hi @agnivade 
Your assumption of allowing versions with duplicate migration is incorrect. If I already have a migration with version 81, and then try to run this PR, it fails with this error:

`
{"timestamp":"2022-04-22 12:48:40.061 +05:30","level":"fatal","msg":"Failed to apply database migrations.","caller":"sqlstore/store.go:163","error":"driver: postgres, message: failed to execute migration, command: executing_query, originalError: pq: duplicate key value violates unique constraint \"db_migrations_pkey\", query: \n\nINSERT INTO db_migrations (version, name) VALUES (81, 'recent_searches')\n"}
exit status 1
`

Hence can you please check and fix this. There's already a migration with version 81 already checked in with the name `threads_deleteat`, which is causing this conflict.

agnivade:
Yep, you are right, my apologies. I'll bump the version.

agnivade:
@harshilsharma63 - Gentle ping for review. :)

harshilsharma63 (approved):
Send it 💯

agnivade:
Gentle ping @srkgupta

agnivade:
@srkgupta - Merging this. The code is hidden behind a feature flag anyways. Let me know if you have any concerns, and I can address them post-merge.

agnivade:
Woops, I cannot merge until you approve. :)

srkgupta (approved):
Thanks @agnivade. Changes LGTM. Sorry for the delay in review as I was in vacation last week :)

PR #20033, opened 2022-04-20
[MM-43444] Delete profile image and invalidate cache on permanent user deletion

plant99:
#### Summary

This change deletes profile image of a user when permanently deleting the account.
#### Ticket Link

Fixes https://github.com/mattermost/mattermost-server/issues/19996 
Jira: https://mattermost.atlassian.net/browse/MM-43444
#### Release Note

```release-note
Fixed user account deletion to delete profile images.
```

#### Related PRs
https://github.com/mattermost/mmctl/pull/508


plant99:
cc: @mkraft can't add reviewers as I'm not part of the GitHub organisation yet. Please review.

mkraft:
Yes good call on doing something like this with file uploads too.

plant99:
> One option for the HTTP status code is [202](https://developer.mozilla.org/en-US/docs/Web/HTTP/Status/202) and the error in the JSON response body. Do you think that makes sense or do you think 200 or 500 are better?

202 semantically makes better sense, I'll go with this. It'll make front-end changes easier too with just an extra if case. 

> Do you want to make a ticket for that?
Yes, will create a JIRA and GitHub ticket sometime soon.

mkraft (approved):

BenCookie95 (approved):

furqanmlk (approved):
Working as expected.
1. Login as Sysadmin
2. Add a user 
3. Add profile picture
4. Delete user via mmctl  `mmctl user delete <User's Email>
5. Go to MM-server/data/users 
6. Verify user folder is deleted.
7. Go to System Console and verify User is inactive

PR #20041, opened 2022-04-20
Mm 43084 add cloud workspace limits endpoint

neallred:
#### Summary
Add `/api/v4/cloud/limits` HTTP endpoint that is only accessible when:
* CloudFree feature flag is enabled
* User is authenticated
* Server has a cloud license

Also adds the `CloudFree` feature flag which can be used for other related feature work going forward.

This work supports https://github.com/mattermost/mattermost-webapp/pull/10217 , and the cloud implementation is provided by https://github.com/mattermost/enterprise/pull/1187

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-43084

#### Release Note
No release note: everything in this commit is gated behind a feature flag. If someone makes requests to this endpoint anyways, then all that gets returned is an unimplemented status code.

```release-note
NONE
```


api4/cloud.go
```diff
@@ -141,6 +143,33 @@ func getCloudProducts(c *Context, w http.ResponseWriter, r *http.Request) {
 	w.Write(json)
 }
 
+func getCloudLimits(c *Context, w http.ResponseWriter, r *http.Request) {
+	if c.App.Channels().License() == nil || !*c.App.Channels().License().Features.Cloud {
+		c.Err = model.NewAppError("Api4.getCloudLimits", "api.cloud.license_error", nil, "", http.StatusNotImplemented)
+		return
+	}
+
+	if !c.App.Config().FeatureFlags.CloudFree {
```
harshilsharma63, line 152: Does this feature flag indicate the user is on a cloud-free instance? If so, do we have the ability to change feature flags at runtime?

In Boards should we check for `License().Features.Cloud` and `FeatureFlags.CloudFree` before calling the `/limits` endpoint to avoid erros?

harshilsharma63, line 152: Does this feature flag indicate the user is on a cloud-free instance? If so, do we have the ability to change feature flags at runtime?

In Boards should we check for `License().Features.Cloud` and `FeatureFlags.CloudFree` before calling the `/limits` endpoint to avoid erros?

neallred: This flag only indicates whether the feature is enabled in general. The checks above are like what you mention checking in boards, and make it so that unlicensed or non-cloud licensed instances are still given the "Not Implemented" error response.

marianunez: nit: For ease of implementation in the clients, if the feature flag is not enabled, should we just return empty limits. This way the client doesn't have to check for the feature flag and just act like it has no limits case instead of having to check or handle the error 0/5

neallred: That sounds good to me for cloud instances. For on-prem instances, I think I'd rather return the not implemented error to encourage clients to not make the request in the first place.

marianunez: Agreed, clients should still check for cloud license which is a permanent check. However, the feature flag will be removed so one less case to remove. Again, 0/5 and possibly a preference for the clients

marianunez (approved):
LGTM, great test coverage 🚀

ashishbhate (approved):
LGTM, the changes in this PR should be added to the API reference github.com/mattermost/mattermost-api-reference/

PR #20057, opened 2022-04-22
Denormalize Post.ChannelId on FileInfo.ChannelId

lieut-data:
#### Summary
Following in the footsteps of the recent work to denormalize `Posts.DeleteAt` onto `Threads.DeleteAt`, this change adds a new `ChannelId` column on `FileInfos` and keeps it up-to-date with the corresponding post, avoiding these crazy API times:

CleanShot 2022-04-21 at 21 15 53@2x

There's no immediate urgency to merging this, but I was annoyed enough by it during loadtests that I thought I'd just propose a fix. In addition to the materialized column, we add an index on `FileInfo(ChannelId, CreateAt)` to streamline the common queries.

#### Postgres (before FileInfo.ChannelId & Index)

Query
```
EXPLAIN ANALYZE SELECT COUNT(*) FROM FileInfo LEFT JOIN Posts as P ON FileInfo.PostId=P.Id LEFT JOIN Channels as C on C.Id=P.ChannelId WHERE FileInfo.DeleteAt = 0 AND C.Id = 'nd9kjtpint8nzcuxifhffpn8ie';
```

Query Plan
```
 Aggregate  (cost=108623.51..108623.52 rows=1 width=8) (actual time=287.536..287.536 rows=1 loops=1)
   ->  Nested Loop  (cost=2523.81..108618.36 rows=2063 width=0) (actual time=30.654..287.245 rows=2419 loops=1)
         ->  Index Only Scan using channels_pkey on channels c  (cost=0.42..2.64 rows=1 width=27) (actual time=0.017..0.020 rows=1 loops=1)
               Index Cond: (id = 'nd9kjtpint8nzcuxifhffpn8ie'::text)
               Heap Fetches: 1
         ->  Gather  (cost=2523.38..108595.08 rows=2063 width=27) (actual time=30.635..289.050 rows=2419 loops=1)
               Workers Planned: 2
               Workers Launched: 2
               ->  Nested Loop  (cost=1523.38..107388.78 rows=860 width=27) (actual time=26.804..281.697 rows=806 loops=3)
                     ->  Parallel Bitmap Heap Scan on posts p  (cost=1522.96..59901.72 rows=23774 width=54) (actual time=25.977..93.872 rows=21549 loops=3)
                           Recheck Cond: ((channelid)::text = 'nd9kjtpint8nzcuxifhffpn8ie'::text)
                           Heap Blocks: exact=15823
                           ->  Bitmap Index Scan on idx_posts_channel_id_delete_at_create_at  (cost=0.00..1508.69 rows=57058 width=0) (actual time=18.462..18.462 rows=64646 loops=1)
                                 Index Cond: ((channelid)::text = 'nd9kjtpint8nzcuxifhffpn8ie'::text)
                     ->  Index Scan using idx_fileinfo_postid_at on fileinfo  (cost=0.42..1.98 rows=2 width=26) (actual time=0.008..0.008 rows=0 loops=64646)
                           Index Cond: ((postid)::text = (p.id)::text)
                           Filter: (deleteat = 0)
                           Rows Removed by Filter: 0
 Planning Time: 0.543 ms
 Execution Time: 289.753 ms
(20 rows)
```

#### Postgres (after FileInfo.ChannelId & Index)
Query
```
EXPLAIN ANALYZE SELECT COUNT(*) FROM FileInfo WHERE FileInfo.DeleteAt = 0 AND FileInfo.ChannelId = 'nd9kjtpint8nzcuxifhffpn8ie';
```

Query plan
```
 Aggregate  (cost=3044.63..3044.64 rows=1 width=8) (actual time=3.411..3.412 rows=1 loops=1)
   ->  Bitmap Heap Scan on fileinfo  (cost=40.55..3037.75 rows=2751 width=0) (actual time=0.667..3.285 rows=2419 loops=1)
         Recheck Cond: ((channelid)::text = 'nd9kjtpint8nzcuxifhffpn8ie'::text)
         Filter: (deleteat = 0)
         Rows Removed by Filter: 11
         Heap Blocks: exact=2188
         ->  Bitmap Index Scan on idx_fileinfo_channelid  (cost=0.00..39.86 rows=2765 width=0) (actual time=0.412..0.412 rows=2430 loops=1)
               Index Cond: ((channelid)::text = 'nd9kjtpint8nzcuxifhffpn8ie'::text)
 Planning Time: 0.162 ms
 Execution Time: 3.446 ms
(10 rows)
```

#### Postgres (before FileInfo.ChannelId & Index)

Query
```
EXPLAIN ANALYZE SELECT FileInfo.*, Coalesce(P.ChannelId, '') AS ChannelId FROM FileInfo LEFT JOIN Posts as P ON FileInfo.PostId=P.Id LEFT JOIN Channels as C ON C.Id=P.ChannelId LEFT JOIN ChannelMembers as CM ON C.Id=CM.ChannelId WHERE (C.TeamId = 'pzky6ogznj8h5dmqfn1ypzfhwh' OR C.TeamId = '') AND FileInfo.DeleteAt = 0 AND C.Id = 'nd9kjtpint8nzcuxifhffpn8ie' ORDER BY FileInfo.CreateAt DESC LIMIT 100;
```

Query Plan
```
 Limit  (cost=1.96..11.26 rows=100 width=1247) (actual time=3202.456..3202.456 rows=0 loops=1)
   ->  Nested Loop  (cost=1.96..765806.03 rows=8237559 width=1247) (actual time=3202.455..3202.455 rows=0 loops=1)
         ->  Nested Loop  (cost=0.98..662544.07 rows=2063 width=1242) (actual time=80.180..3201.707 rows=2419 loops=1)
               ->  Index Scan Backward using idx_fileinfo_create_at on fileinfo  (cost=0.42..60831.23 rows=361888 width=1215) (actual time=0.026..269.783 rows=361767 loops=1)
                     Filter: (deleteat = 0)
                     Rows Removed by Filter: 1988
               ->  Index Scan using posts_pkey on posts p  (cost=0.56..1.66 rows=1 width=54) (actual time=0.008..0.008 rows=0 loops=361767)
                     Index Cond: ((id)::text = (fileinfo.postid)::text)
                     Filter: ((channelid)::text = 'nd9kjtpint8nzcuxifhffpn8ie'::text)
                     Rows Removed by Filter: 1
         ->  Materialize  (cost=0.98..302.45 rows=3993 width=27) (actual time=0.000..0.000 rows=0 loops=2419)
               ->  Nested Loop Left Join  (cost=0.98..282.49 rows=3993 width=27) (actual time=0.017..0.017 rows=0 loops=1)
                     Join Filter: ((c.id)::text = (cm.channelid)::text)
                     ->  Index Scan using channels_pkey on channels c  (cost=0.42..2.65 rows=1 width=27) (actual time=0.017..0.017 rows=0 loops=1)
                           Index Cond: ((id)::text = 'nd9kjtpint8nzcuxifhffpn8ie'::text)
                           Filter: ((teamid)::text = 'pzky6ogznj8h5dmqfn1ypzfhwh'::text)
                           Rows Removed by Filter: 1
                     ->  Index Only Scan using channelmembers_pkey on channelmembers cm  (cost=0.56..229.93 rows=3993 width=27) (never executed)
                           Index Cond: (channelid = 'nd9kjtpint8nzcuxifhffpn8ie'::text)
                           Heap Fetches: 0
 Planning Time: 0.693 ms
 Execution Time: 3202.509 ms
(22 rows)
```

#### Postgres (after FileInfo.ChannelId & Index)
Query
```
EXPLAIN ANALYZE SELECT FileInfo.* FROM FileInfo LEFT JOIN Channels as C ON C.Id=FileInfo.ChannelId LEFT JOIN ChannelMembers as CM ON C.Id=CM.ChannelId WHERE (C.TeamId = 'pzky6ogznj8h5dmqfn1ypzfhwh' OR C.TeamId = '') AND FileInfo.DeleteAt = 0 AND C.Id = 'nd9kjtpint8nzcuxifhffpn8ie' ORDER BY FileInfo.CreateAt DESC LIMIT 100;
```

Query Plan
```
 Limit  (cost=1.40..2.68 rows=100 width=1242) (actual time=2.807..2.807 rows=0 loops=1)
   ->  Nested Loop  (cost=1.40..140679.56 rows=10984743 width=1242) (actual time=2.806..2.807 rows=0 loops=1)
         ->  Index Scan Backward using idx_fileinfo_channelid_createat on fileinfo  (cost=0.42..3077.80 rows=2751 width=1242) (actual time=0.033..2.447 rows=2419 loops=1)
               Index Cond: ((channelid)::text = 'nd9kjtpint8nzcuxifhffpn8ie'::text)
               Filter: (deleteat = 0)
               Rows Removed by Filter: 11
         ->  Materialize  (cost=0.98..302.46 rows=3993 width=27) (actual time=0.000..0.000 rows=0 loops=2419)
               ->  Nested Loop Left Join  (cost=0.98..282.49 rows=3993 width=27) (actual time=0.014..0.015 rows=0 loops=1)
                     Join Filter: ((c.id)::text = (cm.channelid)::text)
                     ->  Index Scan using channels_pkey on channels c  (cost=0.42..2.65 rows=1 width=27) (actual time=0.014..0.014 rows=0 loops=1)
                           Index Cond: ((id)::text = 'nd9kjtpint8nzcuxifhffpn8ie'::text)
                           Filter: (((teamid)::text = 'pzky6ogznj8h5dmqfn1ypzfhwh'::text) OR ((teamid)::text = ''::text))
                           Rows Removed by Filter: 1
                     ->  Index Only Scan using channelmembers_pkey on channelmembers cm  (cost=0.56..229.93 rows=3993 width=27) (never executed)
                           Index Cond: (channelid = 'nd9kjtpint8nzcuxifhffpn8ie'::text)
                           Heap Fetches: 0
 Planning Time: 0.275 ms
 Execution Time: 2.847 ms
(18 rows)
```

#### Ticket Link
None.

#### Release Note
```release-note
Adds new column, FileInfo.ChannelId, materialized from the corresponding Posts.ChannelId column.
Add new index on FileInfo(ChannelId, CreateAt)
```


agnivade:
First off, great work on this Jesse!

Not to spoil the party here, but I am wondering if we are slightly over-optimizing to make our load-test results look better? I took  a look at our Community data and even after I take the top 10 queries just for ChannelStore, `ChannelStore.GetFileCount` comes at 7th.

And also filtering by FileStore data, the queries that are getting optimized here and not getting called at all.

Maybe this points to the fact that the load-tests themselves need to be re-tuned again.

lieut-data:
Thanks for thinking holistically about this, @agnivade!

> I am wondering if we are slightly over-optimizing to make our load-test results look better? 

This change is absolutely focussed on addressing loadtest issues. I note that this API is called with the same volume as `viewChannel` itself, but the performance characteristics definitely depend on the number of file attachments on the server. I have about 400k file attachments on my loadtesting cluster for around 12m posts. If we think that ratio is not representative of a large install, then our simulated events definitely need to be tuned.

> I took a look at our Community data and even after I take the top 10 queries just for ChannelStore, ChannelStore.GetFileCount comes at 7th.

To confirm, this is for time taken? Can you confirm the volume claim above (i.e. `len(viewChannel) == len(getChannelStats)`)?

> And also filtering by FileStore data, the queries that are getting optimized here and not getting called at all.

The `FileStore` ones I highlighted are indeed invoked less frequently (only when users search manually, as near as I can tell). Perhaps I shouldn't have included those metrics in this PR. The change of interest to the loadtests at least is definitely `getChannelStats` that routes through the `ChannelStore`.

> Maybe this points to the fact that the load-tests themselves need to be re-tuned again.

Yes, it would be interesting to count the number of entities in my loadtest and compare the ratio to other installations. If, indeed, file attachments are an order of magnitude less frequent, this change may not be warranted.

agnivade:
> To confirm, this is for time taken?

This is for volume.

>  Can you confirm the volume claim above (i.e. len(viewChannel) == len(getChannelStats))?

Seems like it's inaccurate:

comp

> Yes, it would be interesting to count the number of entities in my loadtest and compare the ratio to other installations.

I wasn't referring to the dataset size, but more on the API call distribution. From real world data, it doesn't seem like this store method is called that heavily at all.

lieut-data:
Thanks for the quick reply, @agnivade!

> > Can you confirm the volume claim above (i.e. len(viewChannel) == len(getChannelStats))?
> 
> Seems like it's inaccurate:
> 

Hmm, those graphs -- despite the volume -- are remarkably similar. I can confirm the webapp calls these two API endpoints at least once when we switch channels. If I click through five channels on community-daily, I see the following events for `getChannelStats`:

CleanShot 2022-04-26 at 14 52 29@2x

while for `viewChannel`, I observe the known issue of this firing more often than seems necessary:

CleanShot 2022-04-26 at 14 52 16@2x

Are we sure we aren't /overcounting/ `viewChannel` in practice because of bugs in the product? 

> > Yes, it would be interesting to count the number of entities in my loadtest and compare the ratio to other installations.
> 
> I wasn't referring to the dataset size, but more on the API call distribution. From real world data, it doesn't seem like this store method is called that heavily at all.

Of course, let's figure out the above disparity first. But I'm curious if we know the number of posts and file attachments on community as a datapoint.

agnivade:
> Are we sure we aren't /overcounting/ viewChannel in practice because of bugs in the product?

Indeed, that's a good point. Yes the shapes of those graphs are the same, which means it's originating from the same UI action - viewing or switching channels.

In that case, it does make sense to add this. 

>  But I'm curious if we know the number of posts and file attachments on community as a datapoint.

Posts are 9M. @streamer45 - can you help us with the fileInfo count?

Unrelated: I'd love to have the viewChannel API multiple calls fixed. :P

streamer45:
These are the counts for `Posts` and `FileInfo` tables on community, both all time and last 30 days for more recent/accurate comparisons:

```
mysql> SELECT COUNT(*) FROM Posts;
--------------
SELECT COUNT(*) FROM Posts
--------------

+----------+
| COUNT(*) |
+----------+
|  7538429 |
+----------+
1 row in set (3.14 sec)

mysql> SELECT COUNT(*) FROM FileInfo;
--------------
SELECT COUNT(*) FROM FileInfo
--------------

+----------+
| COUNT(*) |
+----------+
|   171096 |
+----------+
1 row in set (0.45 sec)

mysql> SELECT COUNT(*) FROM Posts WHERE FROM_UNIXTIME(CreateAt/1000) >= NOW() - INTERVAL 30 DAY;
--------------
SELECT COUNT(*) FROM Posts WHERE FROM_UNIXTIME(CreateAt/1000) >= NOW() - INTERVAL 30 DAY
--------------

+----------+
| COUNT(*) |
+----------+
|   174599 |
+----------+
1 row in set (10.26 sec)

mysql> SELECT COUNT(*) FROM FileInfo WHERE FROM_UNIXTIME(CreateAt/1000) >= NOW() - INTERVAL 30 DAY;
--------------
SELECT COUNT(*) FROM FileInfo WHERE FROM_UNIXTIME(CreateAt/1000) >= NOW() - INTERVAL 30 DAY
--------------

+----------+
| COUNT(*) |
+----------+
|     4200 |
+----------+
1 row in set (0.61 sec)
```

agnivade:
Woops, I got the 9M number from an elasticsearch full post index. Not sure why it over-counted the number. Anyways, thanks! @lieut-data - Looks like your data set has much more file attachments than normal usage.

lieut-data:
Thanks @agnivade & @streamer45! 

> Anyways, thanks! @lieut-data - Looks like your data set has much more file attachments than normal usage.

If community has 7m posts and 171k fileinfos , and I have 12M posts and 390k posts, that's only about 15% difference for scale, right?

I sense a hesitation to effect a(nother) denormalization change, but it seems to be predicated on the conviction that this isn't a material issue on community. I suspect if we updated the loadtests to call `viewChannel` with the current (broken) multiplier, that would indeed dwarf this issue, and possibly explain the `ChannelStore` imbalance as well.

Basically, I'm not sure I'm ready to chalk this up to "loadtest data anomaly." The effects of fixing /this/ issue on my loadtests have been staggering, and so far: the data seems to line with potentially any real customer installation of this size.

> Unrelated: I'd love to have the viewChannel API multiple calls fixed. :P

I'm motivated to look into that :)

agnivade:
> If community has 7m posts and 171k fileinfos , and I have 12M posts and 390k posts, that's only about 15% difference for scale, right?

Indeed, that sounds about right.

> I suspect if we updated the loadtests to call viewChannel with the current (broken) multiplier, that would indeed dwarf this issue, and possibly explain the ChannelStore imbalance as well.

Yep, exactly.

I am not really against this change. But in the priority of things, I feel that fixing the multiple viewChannel call issue takes more priority than fixing this. Because that is what's causing the API skew and masking this problem. Once we have a 1:1 API call volume, then I'm pretty sure this will come as the next bottleneck. So after we fix that, I'd be much more comfortable denormalizing this.

> I'm motivated to look into that :)


ff

lieut-data:
> So after we fix that, I'd be much more comfortable denormalizing this.

Sounds good, thank you for clarifying @agnivade! I'm going to "hold onto" this change in current CRT loadtests, simply because it stabilizes the clusters at their current scale, but acknowledge that it isn't a CRT-specific fix. Let's also chat later today about how to take the recent unbounded user counts and interpret them without this change and without a proper fix (at the moment) for `viewChannel`.

agnivade:
@lieut-data - Now that https://github.com/mattermost/mattermost-webapp/commit/7e81400b9b0241f0ac7a2673875d65ac231b2952 is merged, I am open to reviewing and merging this. Is there anything you want to complete or should we just go ahead with the review?

lieut-data:
@agnivade, I /think/ this is good to review as-is. But, pursuant to your original concern, I'd also want to feel confident that "it's worth it." Unfortunately, I don't have the bandwidth to revisit the original loadtesting analysis right now. Thoughts on how to proceed?

agnivade:
The time has come! @agarciamontoro - This is a PR has Jesse submitted a while back but we decided to pause going ahead with this until we have a performance engineer hired. It might be good for you to take a look at this, run the load tests, verify the gain and merge it.

No rush on this whatsoever. We will prioritize it accordingly with the other items. Just something I wanted to let you know.

agarciamontoro:
Thanks, @agnivade, added a ticket to help us prioritize it https://mattermost.atlassian.net/browse/MM-45956

lieut-data:
Going to close this for now, but happy to resurrect if there's interest!

agnivade:
I definitely think there is interest. However Alejandro has been super busy with the high priority stuff lately :) We want to stabilize our load test first and then tackle further perf improvements.

lieut-data:
Resurrecting this PR as it's been seen in a large customer installation. cc @maxwellpower

agnivade:
@lieut-data - I am owning this task: https://mattermost.atlassian.net/browse/MM-45956 :) I am just waiting on a new data dump from Alejandro to validate the results better. Please close this one.

lieut-data:
Hey @agnivade! That JIRA ticket is about reviewing this PR, right? Is the plan to resubmit with other changes?

agnivade:
I meant that I would have opened a new PR anyways. But it's okay if you prefer to keep this one open till that time, totally fine with me :)

lieut-data:
Aha, ok! 0/5 on how it gets in, and super happy to know you're driving this next potential improvement. I'll close this out again, but feel free to resurrect if helpful :)

PR #20063, opened 2022-04-25
Respect send push proxy notification configuration on test notifications

larkox:
#### Summary
Comply with the configuration on push proxy when testing the notifications.

On the PR this functionality was included, we missed the case where the `SendPushNotifications` config was set to false. This has two implications:
- Test push notifications are sent to the push proxy even if the config or the license does not allow it. Test push notifications are invisible and not being used yet, so should not be a big deal, but still possible.
- V2 applications may receive information as if the notifications can be sent (the test push notifications goes to the Push Proxy without problems) when in reality no push notifications will be sent due to the configuration.

#### Ticket Link
None

#### Release Note
```release-note
Fix bug that allow to send test (empty) notifications even if the `SendPushNotifications` config is set to false.
```


enahum:
@larkox this needs back porting too right?

cc: @amyblais

enahum (approved):

app/notification_push.go
```diff
@@ -578,6 +578,21 @@ func (a *App) BuildPushNotificationMessage(contentsConfig string, post *model.Po
 }
 
 func (a *App) SendTestPushNotification(deviceID string) string {
+	canSendPushNotifications := false
+	if *a.Config().EmailSettings.SendPushNotifications {
+		pushServer := *a.Config().EmailSettings.PushNotificationServer
+		if license := a.Srv().License(); pushServer == model.MHPNS && (license == nil || !*license.Features.MHPNS) {
+			mlog.Warn("Push notifications are disabled. Go to System Console > Notifications > Mobile Push to enable them.")
```
jupenur, line 585: Nit: Is this a reasonable warning message? Shouldn't the call to action be to upgrade the license or to change the push server? Right now it reads like you just need to enable push notifications, but in reality they're already enabled, right?

jupenur (approved):
No security concerns, LGTM.

app/notification_push.go
agnivade: Agree. This is slightly confusing. We should say to either point to a different push server or upgrade license.

larkox: This piece of code is taken verbatim from  https://github.com/mattermost/mattermost-server/blob/9d37a6cc3d916e90ca8dd6fd4d61881227707437/app/notification.go#L405

Should the error change also there?

Also, I see some other places where we use slightly different checks (quick search I find it in `app/channels.go line 2907` and `app/expirynotify.go line 20`. Should we make a util function like `func (a *App) canSendPushNotifications() boolean` to unify everything?

agnivade: Great find. Yes, let's make a util method. Thanks Daniel.

app/channel.go
```diff
@@ -2904,7 +2904,7 @@ func (a *App) SearchChannelsUserNotIn(teamID string, userID string, term string)
 func (a *App) MarkChannelsAsViewed(channelIDs []string, userID string, currentSessionId string, collapsedThreadsSupported bool) (map[string]int64, *model.AppError) {
 	// I start looking for channels with notifications before I mark it as read, to clear the push notifications if needed
 	channelsToClearPushNotifications := []string{}
-	if *a.Config().EmailSettings.SendPushNotifications {
+	if a.canSendPushNotifications() {
```
larkox, line 2907: There is a minor behaviour change here. Before, we were just checking the "SendPushNotifications" setting, so it was not checking the license. Now, in case the license does not allow it, it will not send these notifications. I think the behaviour is expected, but still wanted to point that out.

app/expirynotify.go
```diff
@@ -17,12 +17,8 @@ const (
 
 // NotifySessionsExpired is called periodically from the job server to notify any mobile sessions that have expired.
 func (a *App) NotifySessionsExpired() error {
-	if *a.Config().EmailSettings.SendPushNotifications {
-		pushServer := *a.Config().EmailSettings.PushNotificationServer
-		if license := a.ch.srv.License(); pushServer == model.MHPNS && (license == nil || !*license.Features.MHPNS) {
-			mlog.Warn("Push notifications are disabled. Go to System Console > Notifications > Mobile Push to enable them.")
-			return nil
-		}
+	if !a.canSendPushNotifications() {
```
larkox, line 20: Another behavior change here. Before, we only checked the license if the sendPushNotifications setting was set. But if it was not set, we were sending the session expired notifications.

Now we don't send these notifications if the sendPushNotifications is set to false.

larkox: CC @enahum

app/channel.go
agnivade: Maybe good to add this in the release notes.

app/expirynotify.go
agnivade: Great find, I think this was a clear bug.

agnivade (approved):
Nice work!

amyblais:
@agnivade @larkox Is this safe to backport to 6.3 and who could QA review this?

larkox:
There are a few moving pieces here, so it is not trivial to test. Let me explain each one.

- Mark channel as viewed
  - Used to send the notification to clear the push notifications even if the license did not allow it (`SendPushNotifications = true`, `pushServer = MHPNS` and no license or license with `MHPNS = false`).
  - Now it respects the license to send "mark channel as viewed" notifications.
  - To test one should:
    - Receive message notifications on the mobile (for that, you would need a proper license).
    - Without clearing those notifications, remove the license from the server.
    - On a different client (web, for example), see those messages.
    - Expected result:
      - Before the fix: the notifications for the viewed channel get cleared.
      - After the fix: no notifications get cleared, because no push notification has been sent.
- Session expired notification
  - If `SendPushNotifications = true`, it was working fine (just send the notification if the license allows it). But if `SendPushNotifications = false` it was sending the notification eitherway.
  - To test one should:
    - Set `SendPushNotifications = false`.
    - (optional) Set the mobile expiry time to a very short time
    - Wait for the mobile session to expire
    - Expected result:
      - Before the fix: a notification regarding the expiry arrives to the mobile.
      - After the fix: no notifications regarding the expiry arrives to the mobile, since no push notifications are sent.
- Test notification
  - This was not respecting the license nor the setting at all, so if a proper push proxy was set, it would always check with the push proxy (and in general, return "true").
  - To test one should:
    - Try with no license or a license that does not allow push proxy, or the setting set to false.
    - Find out the device ID of the phone (can be done inspecting during login the communication between the mobile and the server, since there is a call to set this device ID).
    - Make a request to `api/v4/ping?device_id=YOUR_DEVICE_ID`.
    - Expected result:
      - Before the fix: the response should include a `canReceiveNotifications` field set to true.
      - After the fix: the response should include a `canReceiveNotifications` field set to false.
- Other notifications (new mentions)
  - No changes should be expected here.

I don't know who is the right person to test this.

josephbaylon:
@amyblais Testing this requires a local server. I wasn't able to update the `SendPushNotifications` config on an already running server (needs to be restarted)

However, I'm having issues. spinning up my local server. Can you please re-assign this ticket to either @saturninoabril or @furqanmlk to test against their local server?

saturninoabril (approved):
Thanks @larkox! Tested and passed.

saturninoabril:
Please take a look at the failed test. Btw, is there a corresponding Jira ticket for this?

larkox:
@amyblais This PR should be backported to anywhere https://github.com/mattermost/mattermost-server/pull/19610 is in (6.3 I think).

I think it should be safe to backport. Worst case scenario, if we find any conflict, we can backport only the part of the functionality that deals with #19610 .

amyblais:
@larkox This needs a manual cherry-pick to ``release-6.3``.

enahum:
@amyblais what about 6.4 and 6.5?

larkox:
I think (not sure) that https://github.com/mattermost/mattermost-server/pull/19610 hasn't been backported to 6.4 and 6.5 neither. Am I right? If so, it should be backported before this one.

amyblais:
6.4 is reaching end of lifecycle on May 15th, no need to cherry-pick there. I'll do cherry-picks for 6.5.

amyblais:
@larkox Kind reminder to manually cherry-pick this to ``release-6.3``.

PR #20075, opened 2022-04-27
Replace expired GPG public key which is used to upgrade to enterprise

ghost:
#### Summary
There is an embedded GPG Public key in the codebase. That key is used to verify enterprise binary after downloading during upgrade phase. That key is expired, we replaced it with current version

#### Release Note
```release-note
Replaced expired GPG key which is used to verify enterprise binary.
```


spirosoik (approved):

agnivade (approved):
Super fast work!

PR #20076, opened 2022-04-27
Mm 43084 cache refreshing

neallred:
#### Summary
Receive webhook of product limit updates (for example, because the plan subscription changed), and call enterprise method to broadcast the websocket event of the update.

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-43084

#### Release Note
Webhooks will not be sent until feature flag is on, so no release note
```release-note
NONE
```

### DIFF
--- a/api4/cloud.go
+++ b/api4/cloud.go
@@ -475,6 +475,12 @@ func handleCWSWebhook(c *Context, w http.ResponseWriter, r *http.Request) {
 			c.Err = appErr
 			return
 		}
+	case model.EventTypeProductLimitsChanged:
+		if err := c.App.Cloud().UpdateCloudLimitsFromHook(event.ProductLimits); err != nil {
+			c.Err = model.NewAppError("Api4.handleCWSWebhook", "api.cloud.limits.update_error", nil, err.Error(), http.StatusInternalServerError)
+			return
+		}
+		c.Logger.Info("Updated product limits")
 
 	default:
 		c.Err = model.NewAppError("Api4.handleCWSWebhook", "api.cloud.cws_webhook_event_missing_error", nil, "", http.StatusNotFound)
--- a/einterfaces/cloud.go
+++ b/einterfaces/cloud.go
@@ -10,6 +10,7 @@ import (
 type CloudInterface interface {
 	GetCloudProducts(userID string, includeLegacyProducts bool) ([]*model.Product, error)
 	GetCloudLimits(userID string) (*model.ProductLimits, error)
+	UpdateCloudLimitsFromHook(*model.ProductLimits) error
 
 	CreateCustomerPayment(userID string) (*model.StripeSetupIntent, error)
 	ConfirmCustomerPayment(userID string, confirmRequest *model.ConfirmPaymentMethodRequest) error
--- a/einterfaces/mocks/CloudInterface.go
+++ b/einterfaces/mocks/CloudInterface.go
@@ -292,3 +292,17 @@ func (_m *CloudInterface) UpdateCloudCustomerAddress(userID string, address *mod
 
 	return r0, r1
 }
+
+// UpdateCloudLimitsFromHook provides a mock function with given fields: _a0
+func (_m *CloudInterface) UpdateCloudLimitsFromHook(_a0 *model.ProductLimits) error {
+	ret := _m.Called(_a0)
+
+	var r0 error
+	if rf, ok := ret.Get(0).(func(*model.ProductLimits) error); ok {
+		r0 = rf(_a0)
+	} else {
+		r0 = ret.Error(0)
+	}
+
+	return r0
+}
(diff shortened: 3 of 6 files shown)


api4/cloud.go
```diff
@@ -475,6 +475,12 @@ func handleCWSWebhook(c *Context, w http.ResponseWriter, r *http.Request) {
 			c.Err = appErr
 			return
 		}
+	case model.EventTypeProductLimitsChanged:
```
neallred, line 478: Tested through test in enterprise.

neallred: Maybe? I see 3 options.

1) We could update the sub here now. It seems safe now while limits and subs are joined at the hip, but becomes brittle in the future if we do things like condition limits upon special one-off things we apply at a customer level (like incentivized trials). If we are setting ourselves up to support that now, I'd wan't to add more fields to the event like a nillable sub and the reason the limits changed.

2) If we leave this as implemented now, the subscription cache in mm-server is very short-lived, 5 minutes. In that window, an admin would see out of date info in the system console, but I think most other UI would be ok.

3) As an easy compromise could be to always evict the subscription cache without renewing it. Then the only time inconsistent data would be seen is if the admin is already in the system console at the time the subscription changes, and only until they refresh the page.

marianunez: Will review these once we have the CWS one again so I can get a complete picture

neallred: Tried adding the sub to the event and updating it here, too. Since all properties are nillable on the event, looks like it is as easy as updating the sub in the cache if its in the event, and leave it alone otherwise. Right now, it is always present because the only webhook updates we're sending are because the subscription change made it change. Opened the CWS PR and updated these 3 PRs to update the subscription.

api4/cloud.go
```diff
@@ -486,6 +487,22 @@ func handleCWSWebhook(c *Context, w http.ResponseWriter, r *http.Request) {
 			c.Err = appErr
 			return
 		}
+	case model.EventTypeSubscriptionChanged:
+		// event.ProductLimits is nil if there was no change
+		if event.ProductLimits != nil {
+			c.App.GetPluginsEnvironment()
+			if pluginsEnvironment := c.App.GetPluginsEnvironment(); pluginsEnvironment != nil {
```
marianunez, line 494: Is it intended to call `GetPluginsEnvironment` twice here?

api4/cloud.go
```diff
@@ -486,6 +487,22 @@ func handleCWSWebhook(c *Context, w http.ResponseWriter, r *http.Request) {
...
+				pluginsEnvironment.RunMultiPluginHook(func(hooks plugin.Hooks) bool {
```
marianunez, line 495: Interesting... TIL about this plugin hooks 👍

api4/cloud.go
```diff
@@ -486,6 +487,22 @@ func handleCWSWebhook(c *Context, w http.ResponseWriter, r *http.Request) {
...
+					hooks.OnCloudLimitsUpdated(event.ProductLimits)
+					return true
+				}, plugin.OnCloudLimitsUpdatedID)
+			}
+		}
+		if err := c.App.Cloud().UpdateSubscriptionFromHook(event.ProductLimits, event.Subscription); err != nil {
+			c.Err = model.NewAppError("Api4.handleCWSWebhook", "api.cloud.subscription.update_error", nil, err.Error(), http.StatusInternalServerError)
+			return
+		}
+		c.Logger.Info("Updated subscription")
```
marianunez, line 505: Was this intended to stay , maybe:
```suggestion
		c.Logger.Info("Updated subscription from webhook event")
```

app/plugin_api.go
```diff
@@ -1179,3 +1179,9 @@ func (api *PluginAPI) RequestTrialLicense(requesterID string, users int, termsAc
 
 	return api.app.Channels().RequestTrialLicense(requesterID, users, termsAccepted, receiveEmailsAccepted)
 }
+
+// GetCloudLimits requests a trial license and installs it in the server
```
marianunez, line 1183: We need to update the documentation

marianunez (approved):
Thanks @nathanaelhoun 🎉

api4/cloud.go
neallred: Me too! Except YIL (yesterday I learned)

api4/cloud.go
neallred: nope!

app/plugin_api.go
marianunez: I believe this one is still open

neallred: yep

neallred: updated

vish9812 (approved):

mgdelacroix (approved):
LGTM! :+1:

PR #20086, opened 2022-04-28
[MM-39918] System admins should be able to access all deleted channels

vivek-ng:
#### Summary

System admin users should be ale to access all deleted channels irrespective of their membership status. Right now , they aren't able to access archived private channels where they aren't members. This PR fixes this issue.

In addition to the fix , I also refactored the code to make our APIs backward compatible by changing the interface to a struct option. This makes future changes easy to make.

#### Ticket Link

  Fixes https://github.com/mattermost/mattermost-server/issues/19913

#### Release Note

```release-note
System admin users should be able to access all deleted channels

```

### DIFF
--- a/api4/channel.go
+++ b/api4/channel.go
@@ -780,7 +780,12 @@ func getDeletedChannelsForTeam(c *Context, w http.ResponseWriter, r *http.Reques
 		return
 	}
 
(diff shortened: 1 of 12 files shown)


vivek-ng:
Hi @amyblais ,  code reviews here take too much time. I'm no longer interested in contributing to mattermost project . Could you please add the `Up For Grabs` tag again for this issue? Thanks for understanding! I'll close this PR.

PR #20087, opened 2022-04-28
MM-43829: Default for user display_name preference

agnivade:
We fall back to TeammateNameDisplay if a user display_name
preference is not found. This makes the logic consistent
with other parts of the app.

https://mattermost.atlassian.net/browse/MM-43829

```release-note
NONE
```


larkox (approved):

isacikgoz (approved):
LGTM 👍

PR #20091, opened 2022-04-28
Remove code duplication and move utility methods to the store utils

mgdelacroix:
#### Summary
This PR removes code duplication around the `appendMultipleStatementsFlag` and `resetReadTimeout` methods, makes them public to be imported from other tools (namely focalboard) and moves them to the `sqlstore/utils.go` file to avoid import cycle.

#### Release Note
```release-note
NONE
```

### DIFF
--- a/config/database.go
+++ b/config/database.go
@@ -25,6 +25,7 @@ import (
 
 	"github.com/mattermost/mattermost-server/v6/model"
 	"github.com/mattermost/mattermost-server/v6/shared/mlog"
+	"github.com/mattermost/mattermost-server/v6/store/sqlstore"
 	"github.com/mattermost/morph"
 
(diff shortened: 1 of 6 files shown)


sbishel (approved):
LGTM

isacikgoz (commented):
LGTM but wondering if we store these utility functions on another package though. @mgdelacroix how did the cyclic dependency happened in your case?

mgdelacroix:
@isacikgoz I tried to leave them in the `config/utils` file (where one of the method copies was), and the cycle was `store -> config -> store`. Reviewing the other DB related method I've used recently, `DNSHasBinaryParams`, that is also related to the DSN, it's stored in the `sqlstore` package too, so it looked like a good candidate to me. 0/5 on this tho.

Do you have any other package in mind?

PR #20102, opened 2022-04-29
Doc extractor as service

jespino:
#### Summary
This is a draft of how we can do a more decouple doc extractor to isolate the system more and be able to scale this independently


app/server.go
```diff
@@ -670,6 +672,12 @@ func NewServer(options ...Option) (*Server, error) {
 
 	s.initPostMetadata()
 
+	docExtractorService, err := docextractor.NewLocalExtractorService(*s.Config().FileSettings.ArchiveRecursion)
```
jespino, line 675: Here we need to better decide when to use one or another based on the configuration, now we only use the local spawn process, that was the most interesting one for this case, the other to are just subcases of this one.

srkgupta:
Hi @jespino 

There are some open CodeQL alerts generated in this PR. Since this PR is still in a draft state, I am not looking at them currently. However, please include me as a reviewer once the PR is ready and has been dev reviewed. 

Thanks

lieut-data:

PR #20129, opened 2022-05-03
[MM-33123] Add a test for SetPostUnread-DM Channel

vish9812:
#### Summary

Add a test for SetPostUnread-DM Channel

#### Ticket Link

JIRA: https://mattermost.atlassian.net/browse/MM-33123

#### Release Note

```release-note
NONE
```


Willyfrog (approved):
LGTM! 🚀

api4/post_test.go
```diff
@@ -2644,6 +2644,45 @@ func TestSetChannelUnread(t *testing.T) {
 		assert.Equal(t, int64(2), unread.MsgCount)
 	})
 
+	t.Run("Unread on a direct channel", func(t *testing.T) {
+		dc := th.CreateDmChannel(u2)
+		th.CreateMessagePostNoClient(dc, "test1", now)
+		p := th.CreateMessagePostNoClient(dc, "test2", now+10)
+		th.CreateMessagePostNoClient(dc, "test3", now+20)
+		p1 := th.CreateMessagePostNoClient(dc, "test4", now+30)
+		require.NotNil(t, p)
```
ashishbhate, line 2653: This should be moved below where p is first assigned. 

Since we're checking that `p` is `nil`, we should do the same check for `p1`.

ashishbhate (changes requested):
The parent ticket https://mattermost.atlassian.net/browse/MM-32253
says,

>When trying to mark unread a DM message that is in a thread, I get this error:

So we need to add tests for posts in a thread as well.

api4/post_test.go
vish9812: Done

vish9812:
@ashishbhate Added test case for DM Thread.

api4/post_test.go
```diff
@@ -2643,6 +2643,93 @@ func TestSetChannelUnread(t *testing.T) {
...
+		require.Nil(t, err)
+		require.Equal(t, int64(4), unread.MsgCount)
+		require.Equal(t, int64(1), unread.MsgCountRoot)
+		cv := &model.ChannelView{ChannelId: dc.Id}
+		_, appErr = th.App.ViewChannel(cv, u1.Id, s2.Id, false)
+		require.Nil(t, appErr)
+		unread, err = th.App.GetChannelUnread(dc.Id, u1.Id)
+		require.Nil(t, err)
+		require.Equal(t, int64(0), unread.MsgCount)
+		require.Equal(t, int64(0), unread.MsgCountRoot)
```
ashishbhate, line 2708: This is bit strange - the thread contains only posts by `th.BasicUser`, (also referred to as `u1`), since posting a message and viewing a channel/thread are two separate operations, the user's own messages appear as unread in the channel/thread. 

I suggest changing references to `th.BasicUser` to `u1` in the CreatePost calls.

ashishbhate (approved):

api4/post_test.go
vish9812: Yes, at least on the backend, it seems to work this way.

> I suggest changing references to th.BasicUser to u1 in the CreatePost calls.

Done

PR #20130, opened 2022-05-03
MM-43929: Ignore archived team memberships in GraphQL

agnivade:
https://mattermost.atlassian.net/browse/MM-43929

```release-note
NONE
```

### DIFF
--- a/api4/emoji.go
+++ b/api4/emoji.go
@@ -53,7 +53,7 @@ func createEmoji(c *Context, w http.ResponseWriter, r *http.Request) {
 	defer c.LogAuditRec(auditRec)
 
 	// Allow any user with CREATE_EMOJIS permission at Team level to create emojis at system level
-	memberships, err := c.App.GetTeamMembersForUser(c.AppContext.Session().UserId)
+	memberships, err := c.App.GetTeamMembersForUser(c.AppContext.Session().UserId, true)
 
 	if err != nil {
 		c.Err = err
@@ -143,7 +143,7 @@ func deleteEmoji(c *Context, w http.ResponseWriter, r *http.Request) {
 	auditRec.AddMeta("emoji", emoji)
 
 	// Allow any user with DELETE_EMOJIS permission at Team level to delete emojis at system level
-	memberships, err := c.App.GetTeamMembersForUser(c.AppContext.Session().UserId)
+	memberships, err := c.App.GetTeamMembersForUser(c.AppContext.Session().UserId, true)
 
 	if err != nil {
 		c.Err = err
--- a/app/team.go
+++ b/app/team.go
@@ -893,8 +893,8 @@ func (a *App) GetTeamMember(teamID, userID string) (*model.TeamMember, *model.Ap
 	return teamMember, nil
 }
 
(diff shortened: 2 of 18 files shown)


app/team.go
```diff
@@ -893,8 +893,8 @@ func (a *App) GetTeamMember(teamID, userID string) (*model.TeamMember, *model.Ap
 	return teamMember, nil
 }
 
-func (a *App) GetTeamMembersForUser(userID string) ([]*model.TeamMember, *model.AppError) {
-	teamMembers, err := a.Srv().Store.Team().GetTeamsForUser(context.Background(), userID)
+func (a *App) GetTeamMembersForUser(userID string, includeDeleted bool) ([]*model.TeamMember, *model.AppError) {
```
agnivade, line 896: Since this was just one more param, I'd say this is the borderline beyond which a struct would be necessary. Because even a struct would need one additional param.

api4/emoji.go
```diff
@@ -53,7 +53,7 @@ func createEmoji(c *Context, w http.ResponseWriter, r *http.Request) {
 	defer c.LogAuditRec(auditRec)
 
 	// Allow any user with CREATE_EMOJIS permission at Team level to create emojis at system level
-	memberships, err := c.App.GetTeamMembersForUser(c.AppContext.Session().UserId)
+	memberships, err := c.App.GetTeamMembersForUser(c.AppContext.Session().UserId, true)
```
larkox, line 56: Out of the scope, but this seems weird. Looks like if we used to have permissions to create emojis in a team that we are no longer a member of, we could still create an emoji by hitting the API directly. Not sure if other API endpoints are suffering of similar behaviour, but I feel someone should take a look.

store/storetest/team_store.go
```diff
@@ -1347,14 +1347,26 @@ func testTeamMembers(t *testing.T, ss store.Store) {
...
+
+	ms, err = ss.Team().GetTeamsForUser(ctx, uid, false)
+	require.NoError(t, err)
+	require.Len(t, ms, 1)
+
 	nErr = ss.Team().RemoveAllMembersByUser(uid)
 	require.NoError(t, nErr)
 
-	ms, err = ss.Team().GetTeamsForUser(ctx, m1.UserId)
+	ms, err = ss.Team().GetTeamsForUser(ctx, m1.UserId, true)
```
larkox, line 1369: Should we add some tests for `GetTeamsForUser(x,y,false)`?

larkox (approved):
LGTM, just one nit and one thing out of the scope.

agnivade:
Hehe, you deleted your comment :)

api4/emoji.go
agnivade: Agree. @Willyfrog - should I create a ticket for this?

larkox:
Actually I deleted it before posting the review, but internet magic made it reappear XD

streamer45 (approved):

api4/emoji.go
Willyfrog: Added: https://mattermost.atlassian.net/browse/MM-44407

PR #20141, opened 2022-05-05
MM-38982: fixes SVG uploading

koox00:
#### Summary

Uploading SVG files was "almost" treated like an image in some places,
resulting in not showing the SVG on the client.
This is happening because when we fail to generate a mini preview for an
image, we prepend "invalid-" to the MimeType.

This commit adds a check to guard against SVG files in methods that make
no sense for SVGs.

#### Ticket Link

https://mattermost.atlassian.net/browse/MM-38982

#### Release Note

```release-note
Fixes uploading SVG files
```


streamer45:
Thanks @koox00, I am thinking the core issue here is that we are messing with the mime-type which we shouldn't. 

@cpoile  I assume that was a temporary solution and that we figured out the problem eventually so does it still make sense to keep that logic?

koox00:
> Thanks @koox00, I am thinking the core issue here is that we are messing with the mime-type which we shouldn't.

That's true, but also some methods are too specific for images, that assume a `thumb` file and a `preview` one, which don't exist for SVGs

koox00:
>  @koox00 Am I right in assuming that even if we removed the invalid- mime prefix we would still need these changes?

@cpoile not that the app is going to crash without the changes, but IMO, yes either way these changes would be good to have

agnivade:
Yeah, the original OOM issue was due to passing the image preview as part of *model.Post to all plugins. Therefore, serializing the huge byte slice caused a large allocation. Then there was another unrelated issue caused by a model API change, but that didn't cause any memory problem.

We can revert this one if it does not make sense any more.

koox00:
@agnivade, @cpoile what do you propose to do, 
so should I close this PR, 
revert this one https://github.com/mattermost/mattermost-server/pull/18445,
or just re-assign the ticket?

agnivade:
I'll defer to Claudio/CPoile on this.

streamer45:
@koox00 I think we can remove the "invalid" mime type fix on top of this PR and proceed.

koox00:
> @koox00 I think we can remove the "invalid" mime type fix on top of this PR and proceed.

@streamer45 done

cpoile (approved):
Thanks @koox00 !

streamer45 (approved):
There are a couple more places where we check against `image/svg+xml`. I'd suggest using the nice util you added here. Other than that, it looks great, thanks :+1:

jgilliam17:
@koox00 Sometimes when I add a single .svg file it is not shown, but if the file is part of multi-attachment post it is shown. 
On the image, the window svg was posted twice (see red lines) and it failed to be shown, but preview is there when it is posted later in a multi-attachment post 
Screen Shot 2022-05-16 at 6 02 08 PM

koox00:
> @koox00 Sometimes when I add a single .svg file it is not shown, but if the file is part of multi-attachment post it is shown.
On the image, the window svg was posted twice (see red lines) and it failed to be shown, but preview is there when it is posted later in a multi-attachment post

The previews are shown now though, right? I assume it's this (the timestamp matches):

Screenshot 2022-05-17 at 11 52 10 AM

jgilliam17:
@koox00 I still don't see the single svg posts on Chrome and Safari. They are visible on FirefoxScreen Shot 2022-05-17 at 8 47 54 AM

koox00:
@jgilliam17 OK the issue for not showing are CSS rules, (which this PR doesn't change).
I'll raise a webapp PR as well.

jgilliam17:
@koox00 do we have to wait for webapp PR before merging this one?

jgilliam17:
Just waiting for E2Es to finish, will merge after report posts and if everything looks good.

koox00:
> @koox00 do we have to wait for webapp PR before merging this one?

@jgilliam17 I've added this ticket for webapp https://mattermost.atlassian.net/browse/MM-44470

jgilliam17 (approved):

PR #20144, opened 2022-05-05
MM-43716: enforce size limits in slack importer

ashishbhate:
#### Summary

- Added a wrapped `io.LimitedReader` to enforce size limits in the slack importer. We wrap the `io.LimitedReader` to identify and bubble up the size exceeded error instead of returning EoF from the limited reader. We return a new error instead of wrapping the EoF to prevent users of the wrapped reader checking for EoF with a `Errors.Is` and assuming that the EoF was legitimate. 

#### Ticket Link

- https://mattermost.atlassian.net/browse/MM-43716

#### Release Note

```release-note
NONE
```


utils/file.go
```diff
@@ -112,3 +113,25 @@ func CopyDir(src string, dst string) (err error) {
...
+func NewLimitedReaderWithError(reader io.Reader, n int64) *LimitedReaderWithError {
+	return &LimitedReaderWithError{
+		limitedReader: &io.LimitedReader{R: reader, N: n},
+		maxBytes:      n,
+	}
+}
+
+func (l *LimitedReaderWithError) Read(p []byte) (int, error) {
+	n, err := l.limitedReader.Read(p)
+	if l.limitedReader.N <= 0 && err == io.EOF {
```
jupenur, line 133: In this specific case I think this is fine and will work as intended, but considering `NewLimitedReaderWithError` is laid out as a general-purpose utility that could end up being used elsewhere... this is from the `io.Reader` docs:

> When Read encounters an error or end-of-file condition after successfully reading n > 0 bytes, it returns the number of bytes read. It may return the (non-nil) error from the same call or return the error (and n == 0) from a subsequent call. An instance of this general case is that a Reader returning a non-zero number of bytes at the end of the input stream may return either err == EOF or err == nil. The next Read should return 0, EOF.

So -- in theory, for some `io.Reader` implementation -- we could see an EOF here even before there was any attempt to go over the limit. In that case this `Read` method should return the original EOF error and not the `SizeLimitExceeded` one. Right?

Some of Go's internal code solves this by setting the limit one byte *higher* than what's actually the intended maximum, then checking just the `N` value: https://cs.opensource.google/go/x/mod/+/refs/tags/v0.5.1:zip/zip.go;l=658-670

ashishbhate: Here `N` is the number of [**remaining** bytes](https://pkg.go.dev/io#LimitedReader), it will only be `<=0` after we've exceeded the limit.

ashishbhate: Do you mean the check should be `<` instead of `<=`?

jupenur: > Do you mean the check should be < instead of <=?

Take a look at the link. `<=` is right but `err == io.EOF` is wrong, *and* the initial `N` value should be set to `n + 1` instead of `n`.

> Here N is the number of [remaining bytes](https://pkg.go.dev/io#LimitedReader), it will only be `<=0` after we've exceeded the limit.

This is very subtly incorrect. It will only be `<= 0` after exceeding the limit *or exactly at the limit*. If the initial `N` value is set to `n + 1`, *then* it will only be `<= 0` after exceeding the limit. And in that case checking the error value is unnecessary.

ashishbhate: I think I understand what you mean:

If `l.limitedReader.N` starts off as the MaxByteLimit (it reduces as data is read from the reader), 

then the conditional

`if l.limitedReader.N <= 0 && err == io.EOF `

is incorrect, because `l.limitedReader.N == 0 && err == io.EOF` means that the file/stream was exactly MaxByteLimit size and so we've hit the EoF, and this should not be a `SizeLimitExceeded` error.



>And in that case checking the error value is unnecessary.

This could potentially hide an underlying error. https://cs.opensource.google/go/go/+/refs/tags/go1.8:src/io/io.go;l=429-439

On line 438, the std library returns the error from the underlying `Read`, after setting the N value. 
The EOF is only returned on the next call to the limited reader's `Read`.

If we don't check that `err = io.EOF` we could be masking an underlying error. Of course, we are still in a `SizeLimitExceeded` state, but I wanted to return any underlying error instead of overwriting it, and the next call to the `Read` would still raise the `SizeLimitExceeded` error as expected.



Interestingly, all of this is moot when https://github.com/golang/go/commit/fd6ef06296b55a234d4fd5ebf5dd3e5bfd8f4e25 is released in some future version of go :)

jupenur: Yup, exactly.

agnivade: I don't see why we can't just use our own `LimitedReader` implementation instead of trying to override the existing one just to return a custom error. It's near about the same lines of code anyways. We just need to return our custom error instead of `EOF`.

```go
func (l *LimitedReader) Read(p []byte) (n int, err error) {
	if l.N <= 0 {
		err := l.Err
		if err == nil {
			err = // OUR CUSTOM ERROR HERE
		}
		return 0, err
	}
	if int64(len(p)) > l.N {
		p = p[0:l.N]
	}
	n, err = l.R.Read(p)
	l.N -= int64(n)
	return
}
```

utils/file.go
```diff
@@ -112,3 +113,24 @@ func CopyDir(src string, dst string) (err error) {
...
+		limitedReader: &io.LimitedReader{R: reader, N: maxBytes + 1},
+	}
+}
+
+func (l *LimitedReaderWithError) Read(p []byte) (int, error) {
+	n, err := l.limitedReader.Read(p)
+	if l.limitedReader.N <= 0 && err == io.EOF {
+		return n, SizeLimitExceeded
+	}
+	fmt.Println(l.limitedReader.N)
```
agnivade, line 134: Remove?

agnivade (commented):
0/5 maybe better to place this in utils/fileutils.

utils/file.go
ashishbhate: >I don't see why we can't just use our own LimitedReader implementation instead of trying to override the existing one just to return a custom error. It's near about the same lines of code anyways. We just need to return our custom error instead of EOF.

As I said before, the code you quoted was merged into Go master just yesterday and will released in 1.19, at which time we can use it directly. 
For the version of `io.LimitedReader` in our supported version of Go (1.18), which I looked at, it seemed wrapping made more sense than duplicating, because I think the wrapper I've written is shorter, easier to understand, and expresses the intent better (all subjective, sure).

agnivade: 0/5 on it really, but I don't really think writing our own LimitedReader implementation is that complicated. Your conversation with Juho made me think that there maybe subtle edge-cases in trying to implement a correct wrapper. So I thought that just copying the implementation until Go 1.19 is available should do for now. But it's a non-blocker, so approving it.

agnivade (approved):
Great stuff!

utils/file.go
ashishbhate: >Your conversation with Juho made me think that there maybe subtle edge-cases in trying to implement a correct wrapper. 

This is true, but I think the wrapper is good now. I think Juho might prefer copying the Go standard library version, so I'll leave it to him to take the final call.

jupenur: I'm 0/5 on what's the right thing to do and in this specific use-case it doesn't seem to matter at all. Fine with either option.

jupenur (approved):
Tested locally and verified to work correctly. LGTM!

jgilliam17:
[E2E report](https://mattermost-cypress-report.s3.amazonaws.com/191548-1a3f177-onprem-ent-server-pr-20144-mattermost/mm-ee-test1a3f177/mochawesome.html) looks good, no PR related failures.
Reassigning to @furqanmlk as I'm having issues with mmctl and can't verify import manually.

furqanmlk (commented):
Hi @ashishbhate 

I followed these steps using two different zip files, but getting same error,
I used `import.zip` file attached with Jira ticket

```  
1. 
❯ ./mmctl import upload /Users/furqanmalik/Downloads/import.zip
    Upload session successfully created, ID: 4zt4d1txgpghip9gpau3cgnhqw
    Import file successfully uploaded, name: g6nmm6358tb8bk3spyan95ioma
```

```
2.
❯ ./mmctl import list available
    4zt4d1txgpghip9gpau3cgnhqw_import.zip
```

```
3.
❯ ./mmctl import process 4zt4d1txgpghip9gpau3cgnhqw_import.zip
    Import process job successfully created, ID: uf8u3dw7ubb9xe6jtuwznm46ma
```

```
4.  
❯ ./mmctl import job show uf8u3dw7ubb9xe6jtuwznm46ma --json
[
  {
    "id": "uf8u3dw7ubb9xe6jtuwznm46ma",
    "type": "import_process",
    "priority": 0,
    "create_at": 1652201108184,
    "start_at": 1652201118920,
    "last_activity_at": 1652201119027,
    "status": "error",
    "progress": -1,
    "data": {
      "error": "We could not count the users. — ImportProcessWorker: Unable to process import: JSONL file is missing., jsonFile was nil",
      "import_file": "4zt4d1txgpghip9gpau3cgnhqw_import.zip"
    }
  }
]
```

ashishbhate:
@furqanmlk could you share the import file with me on the community server please?

furqanmlk:
> @furqanmlk could you share the import file with me on the community server please? 

@ashishbhate this is the same import file attached to the ticket

https://mattermost.atlassian.net/browse/MM-43716

I will share it in community server too

furqanmlk (approved):
Thanks @ashishbhate for your help
working as expected
```
curl --location --request POST 'https://mattermost-server-pr-20144.test.mattermost.cloud/api/v4/teams/pd4dih1x4j837kwtwbb5t3idga/import' \
--header 'Authorization: Bearer xoc51zadbprqixhftqfg4owqrw' \
--form 'file=@"/Users/furqanmalik/Downloads/import.zip"' \
--form 'importFrom="slack"' \
--form 'filesize="4659203"'
{"results":"TWF0dGVybW9zdCBTbGFjayBJbXBvcnQgTG9nDQpjaGFubmVscy5qc29uIGluIHppcCBhcmNoaXZlIHRvbyBsYXJnZSB0byBwcm9jZXNzIGZvciBTbGFjayBpbXBvcnQNCg0KVXNlcnMgY3JlYXRlZDoNCj09PT09PT09PT09PT09PQ0KDQpUaGUgSW50ZWdyYXRpb24vU2xhY2sgQm90IHVzZXIgd2l0aCBlbWFpbCBzbGFja2ltcG9ydHVzZXJfbjgzZ2dwNTFwaThoZGp1M2FpdGRjZnJvM2NAbG9jYWxob3N0IGFuZCBwYXNzd29yZCB3aHFleGdpdzNicnI5anQzY3NtMTZnNWt3dyBoYXMgYmVlbiBpbXBvcnRlZC4NCg0KQ2hhbm5lbHMgYWRkZWQ6DQo9PT09PT09PT09PT09PT09PQ0KDQoNCk5vdGVzOg0KPT09PT09PQ0KDQotIFNvbWUgbWVzc2FnZXMgbWF5IG5vdCBoYXZlIGJlZW4gaW1wb3J0ZWQgYmVjYXVzZSB0aGV5IHdlcmUgbm90IHN1cHBvcnRlZCBieSB0aGlzIGltcG9ydGVyLg0KLSBTbGFjayBib3QgbWVzc2FnZXMgYXJlIGN1cnJlbnRseSBub3Qgc3VwcG9ydGVkLg0KLSBBZGRpdGlvbmFsIGVycm9ycyBtYXkgYmUgZm91bmQgaW4gdGhlIHNlcnZlciBsb2dzLg0K"}%
```
Decoded 

```
curl --location --request GET 'https://httpbin.org/base64/TWF0dGVybW9zdCBTbGFjayBJbXBvcnQgTG9nDQpjaGFubmVscy5qc29uIGluIHppcCBhcmNoaXZlIHRvbyBsYXJnZSB0byBwcm9jZXNzIGZvciBTbGFjayBpbXBvcnQNCg0KVXNlcnMgY3JlYXRlZDoNCj09PT09PT09PT09PT09PQ0KDQpUaGUgSW50ZWdyYXRpb24vU2xhY2sgQm90IHVzZXIgd2l0aCBlbWFpbCBzbGFja2ltcG9ydHVzZXJfZTg3enVya3oxM2I0NXhleW02Ym81dXB6ZG9AbG9jYWxob3N0IGFuZCBwYXNzd29yZCA4Yzdod2R6M2JpbjM3bWZ0MzQ1c3lvd3JweSBoYXMgYmVlbiBpbXBvcnRlZC4NCg0KQ2hhbm5lbHMgYWRkZWQ6DQo9PT09PT09PT09PT09PT09PQ0KDQoNCk5vdGVzOg0KPT09PT09PQ0KDQotIFNvbWUgbWVzc2FnZXMgbWF5IG5vdCBoYXZlIGJlZW4gaW1wb3J0ZWQgYmVjYXVzZSB0aGV5IHdlcmUgbm90IHN1cHBvcnRlZCBieSB0aGlzIGltcG9ydGVyLg0KLSBTbGFjayBib3QgbWVzc2FnZXMgYXJlIGN1cnJlbnRseSBub3Qgc3VwcG9ydGVkLg0KLSBBZGRpdGlvbmFsIGVycm9ycyBtYXkgYmUgZm91bmQgaW4gdGhlIHNlcnZlciBsb2dzLg0K'
Mattermost Slack Import Log
channels.json in zip archive too large to process for Slack import
```

amyblais:
@ashishbhate FYI, I will own cherry-picking this. We're at code freeze for 6.7, so I'll cherry-pick this to the needed release branches a bit later.

ashishbhate:
>@ashishbhate FYI, I will own cherry-picking this. We're at code freeze for 6.7, so I'll cherry-pick this to the needed release branches a bit later.

Thanks, let me know if I can help with anything.

PR #20146, opened 2022-05-05
GraphQL: Unlock goroutine throttle to match batch capacity

agnivade:
We were throttling the amount of concurrent resolvers
at a given time. The idea behind this was to avoid overloading
the database with too many requests.

However, with the introduction of dataloaders, this limitation
actually becomes a bottleneck because all DB calls are actually
batched, so we are unnecessarily throttling the amount of
items that can be processed in a single batch.

The only caveat with this is that now all resolvers
need to backed by dataloaders, or otherwise not be queried
as part of a loop.

In a subsequent PR, we will be removing channel stats
from under channel to be a top-level object to be returned
for a given channel.

```release-note
NONE
```


isacikgoz (approved):
LGTM

PR #20152, opened 2022-05-06
[MM-43917] Cloud Freemium limits API: messages/posts

vish9812:
#### Summary

Implement the API endpoint that returns the current usage limits for messages

#### Ticket Link

JIRA - https://mattermost.atlassian.net/browse/MM-43917

#### Release Note

```release-note
Added new API endpoint 
* GET /api/v4/usage/posts
* Response field: 
  * count: number - Rounded off number like 900 instead of 987
```


api4/cloud_test.go
```diff
@@ -141,3 +141,91 @@ func Test_getCloudLimits(t *testing.T) {
...
+	t.Run("no license returns not implemented", func(t *testing.T) {
+		th := Setup(t)
+		defer th.TearDown()
+
+		th.App.Srv().RemoveLicense()
+
+		usage, r, err := th.Client.GetCloudUsageForMessages()
+		require.Error(t, err)
+		require.Nil(t, usage)
+		require.Equal(t, http.StatusNotImplemented, r.StatusCode)
```
Willyfrog, line 155: given [mozilla's notes](https://developer.mozilla.org/en-US/docs/Web/HTTP/Status/501) on the status, I'm not 100% sure it is the right one, but 1/5 on this, status are usually very subjective.

Willyfrog (commented):
I'll continue the review later, but leaving the comment for now.

api4/cloud_test.go
vish9812: Since getCloudLimits already returns 501 for the same case, that's why I made it consistent.
Otherwise, I think 403 could be more appropriate, as the resource is forbidden for Non-Licensed/Non-Cloud users.
Whichever we choose, we should make it consistent in both APIs.
cc @neallred

model/client4.go
```diff
@@ -7755,6 +7755,19 @@ func (c *Client4) GetProductLimits() (*ProductLimits, *Response, error) {
...
 
+func (c *Client4) GetCloudUsageForMessages() (*MessagesUsage, *Response, error) {
+	r, err := c.DoAPIGet(c.cloudRoute()+"/usage/messages", "")
+	if err != nil {
+		return nil, BuildResponse(r), err
+	}
+	defer closeBody(r)
+
+	var usage *MessagesUsage
+	json.NewDecoder(r.Body).Decode(&usage)
```
ashishbhate, line 7766: We should check the error returned. Most of the other client code doesn't check marshalling/unmarshalling errors, which is something we need to fix eventually.

store/sqlstore/post_store.go
```diff
@@ -2124,25 +2124,29 @@ func (s *SqlPostStore) AnalyticsPostCountsByDay(options *model.AnalyticsPostCoun
...
 		query = query.Where(sq.Or{sq.NotEq{"p.FileIds": "[]"}, sq.NotEq{"p.Filenames": "[]"}})
 	}
 
-	if mustHaveHashtag {
+	if options.MustHaveHashtag {
 		query = query.Where(sq.NotEq{"p.Hashtags": ""})
 	}
 
+	if options.ExcludeDeleted {
+		query = query.Where(sq.Eq{"p.deleteat": 0})
```
ashishbhate, line 2147: This should be `p.DeleteAt`, otherwise it will fail on MySQL. 

We should add tests for this new option.

ashishbhate (changes requested):
Nicely done, @vish9812!

Some non-blocking and blocking comments inline.

api4/cloud.go
```diff
@@ -22,6 +22,8 @@ func (api *API) InitCloud() {
 	api.BaseRoutes.Cloud.Handle("/products", api.APISessionRequired(getCloudProducts)).Methods("GET")
 	// GET /api/v4/cloud/limits
 	api.BaseRoutes.Cloud.Handle("/limits", api.APISessionRequired(getCloudLimits)).Methods("GET")
+	// GET /api/v4/cloud/usage/messages
+	api.BaseRoutes.Cloud.Handle("/usage/messages", api.APISessionRequired(getCloudUsageForMessages)).Methods("GET")
```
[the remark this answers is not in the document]
vish9812: Yes, I discussed it with @Willyfrog. And it does seem more appropriate here as it's specific to cloud.

api4/cloud_test.go
neallred: Ditto :laughing: I did it because most of the other endpoints in this file did it that way. I lean slightly more towards keeping it consistent with the majority of the other handlers but otherwise agree with @vish9812 . If either of you feel strongly enough to change this endpoint and `getCloudLimits` to return 403 and update tests, I wouldn't oppose it, either.

neallred (changes requested):
Overall looks good! The biggest soundness change is checking for nil in limits.Messages. As a convenience for the front end, it would also be nice to return a (truncated) number instead of a percent.

app/cloud.go
```diff
@@ -156,3 +157,30 @@ func (a *App) SendCloudTrialEndedEmail() *model.AppError {
...
+	}
+
+	// Fetch messages count
+	c, err := a.Srv().Store.Post().AnalyticsPostCount(&model.PostCountOptions{ExcludeDeleted: true})
+	if err != nil {
+		return 0, model.NewAppError("GetCloudUsageForMessages", "app.channel.AnalyticsPostCount.internal_error", nil, err.Error(), http.StatusInternalServerError)
+	}
+
+	// Return usage %
+	max := float64(*limits.Messages.History)
```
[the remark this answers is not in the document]
ashishbhate: @neallred What's the difference between a limit of `0` and a limit of `nil`?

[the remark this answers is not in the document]
neallred: There shouldn't be a limit of 0 for messages, but if there were, it would mean they're not allowed to post any messages. I think "make 0 meaningful" by having `0` mean no limit could be useful for message history, but I could see 0 being a valid limit for other resources now or in the future, so I didn't want to have an inconsistency where for some limits 0 means none allowed but for others it means no limit.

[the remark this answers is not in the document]
ashishbhate: Thanks for clarifying! SGTM!

api4/cloud_test.go
Willyfrog: as I said, no strong feelings, it usually ends up being a bit subjective, as long as we are consistent it's all good with me :)

api4/cloud_test.go
```diff
@@ -141,3 +141,91 @@ func Test_getCloudLimits(t *testing.T) {
...
+				History: &max,
+			},
+		}
+		cloud.On("GetCloudLimits", mock.Anything).Return(mockLimits, nil)
+		th.App.Srv().Cloud = cloud
+
+		// The following three calls create total of 4 posts
+		th.BasicTeam = th.CreateTeam()
+		th.BasicChannel = th.CreatePublicChannel()
+		th.BasicPost = th.CreatePost()
```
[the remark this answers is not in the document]
Willyfrog: good call! no, system messages shouldn't count. But I don't think it will matter much in the first months. Can we create a ticket to adress it properly later on? right now we are not enforcing, but once we enforce them people is likely to complain.

app/cloud.go
```diff
@@ -156,3 +157,30 @@ func (a *App) SendCloudTrialEndedEmail() *model.AppError {
...
+	count := float64(c)
+
+	if count >= max {
+		return 100, nil
+	}
+
+	usage := (count / max) * 100
+	return utils.FloorToNearest10(usage), nil
```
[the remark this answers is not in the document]
vish9812: As per this change, I don't need to call the getCloudLimits at all. I could just return the roundOff count of posts, right?
And that would mean, it could be just a regular posts/ api instead of anything to do with cloud.
Or should I just remove the getCloudLimits call and keep the rest files/api same?

api4/cloud.go
```diff
@@ -182,6 +184,37 @@ func getCloudLimits(c *Context, w http.ResponseWriter, r *http.Request) {
...
+		return
+	}
+
+	count, appErr := c.App.GetCloudUsageForMessages(c.AppContext.Session().UserId)
+	if appErr != nil {
+		c.Err = model.NewAppError("Api4.getCloudUsageForMessages", "api.cloud.request_error", nil, appErr.Error(), http.StatusInternalServerError)
+		return
+	}
+
+	json, err := json.Marshal(&model.MessagesUsage{Enabled: true, Status: count})
```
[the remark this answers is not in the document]
vish9812: If the code reaches this point, then it means this condition is already satisfied.
Although as per above comments, I may not event need to call the limits.

app/cloud.go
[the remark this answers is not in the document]
vish9812: I'm 3/5 to have a `/limits` path. Messages and storage both could fall under this path.
@Willyfrog @ashishbhate thoughts?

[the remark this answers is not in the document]
ashishbhate: Yeah, I think a `/limits` subpath is the way to go. 

It appears Mattermost routes are grouped up by what is being acted on, returned, or contained within a sub-router.

I think it makes sense for `limits/posts` to return the posts limit because we're returning limits, posts just happen to be one of the limits. 

`posts/limit` doesn't seem right because we aren't acting on posts, returning posts and limits aren't contained within a post.

api4/cloud_test.go
agnivade: Please avoid using 5xx codes if possible. 5xx codes are tracked by SRE and security as SLA violations and eventually will end up in your queue to fix :P There is a recent campaign by SRE to clean up any unnecessary 5xx codes to 4xx. I'd suggest to change everything to 403.

Ticket for reference: https://mattermost.atlassian.net/browse/MM-43202

cc @stylianosrigas @DSchalla

app/cloud.go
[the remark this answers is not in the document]
ashishbhate: This is bike-shedding but I think the subpath should probably be `/usage` instead of `limits`. I'm 0/5.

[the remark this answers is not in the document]
vish9812: * Alright, lets finalize it to `/usage/posts`. 
* And I'll remove the cloud call as well and just return the posts count. 
* That would also mean, no `enabled` field, just a `count` field will do.

Although, now posts count raise another question like up to what level do we want to round off?
* The current method rounds off only to nearest 10, i.e., 9876 to 9870.
  * Will this work? 
* Or should I make it round off to the last possible digit like 9876 to 9000?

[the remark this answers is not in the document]
ashishbhate: >(regular user view https://www.figma.com/file/Aob5ffHtfjKzQoebR93ycA/Cloud-Freemium?node-id=602%3A430680)


@neallred This mock up is under the "admin" view (Frame 957). 

From the regular user views (Frame  958), it looks like a regular user doesn't need access  to this API. The limits check can happen transparently in the app. 

Can we continue limiting access to this API to sysadmin users only?

api4/cloud_test.go
neallred: Noted, sounds good to me.

app/cloud.go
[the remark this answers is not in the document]
neallred: There's a number of cases where end users need the usage data, too, so I think anyone with a session needs to be able to access.

The usage & limits modal shows up in a number of non-admin user contexts, too. For example, in boards: https://www.figma.com/file/Aob5ffHtfjKzQoebR93ycA/Cloud-Freemium?node-id=784%3A453152 , and I think in channels, too, but I need to check more.

The channels main menu also sometimes needs usage data for end users: https://www.figma.com/file/Aob5ffHtfjKzQoebR93ycA/Cloud-Freemium?node-id=121%3A333428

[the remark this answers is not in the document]
neallred: > Although, now posts count raise another question like up to what level do we want to round off?

No strong opinion from me. The UI currently only shows in increments of 100, e.g. `5.5K messages`, so no benefit returning anything more granular.

model/post_count.go
```diff
@@ -0,0 +1,12 @@
...
+
+package model
+
+type PostCountOptions struct {
+	// Only include posts on a specific team. "" for any team.
+	TeamId          string
+	MustHaveFile    bool
+	MustHaveHashtag bool
+	ExcludeDeleted  bool
+}
```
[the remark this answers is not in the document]
vish9812: Done.

store/sqlstore/post_store.go
vish9812: Done.

utils/utils.go
```diff
@@ -215,3 +216,16 @@ func IsValidMobileAuthRedirectURL(config *model.Config, redirectURL string) bool
 	}
 	return false
 }
+
+// RoundOffToZeroes converts all digits to 0 except the 1st one.
+// Special case: If there is only 1 digit, then returns 0.
+func RoundOffToZeroes(n float64) int {
```
neallred, line 222: Works for me!

vish9812:
Now, only finalizing the Permission for the API remains.

neallred (approved):
LGTM, thanks @vish9812 !

agnivade:
Hey @vish9812 - no need for my review :) It was just a passing comment. You already have a lot of eyes on the PR. :)

api4/cloud_test.go
```diff
@@ -141,91 +141,3 @@ func Test_getCloudLimits(t *testing.T) {
...
-
-		th.Client.Logout()
-
-		usage, r, err := th.Client.GetCloudUsageForMessages()
-		require.Error(t, err)
-		require.Nil(t, usage)
-		require.Equal(t, http.StatusUnauthorized, r.StatusCode)
-	})
-
-	t.Run("good request with cloud server and feature flag returns response", func(t *testing.T) {
```
mickmister: @vish9812 Note that some unit tests were removed here while moving the `/cloud/usage/messages` endpoint to `/usage/posts`

app/cloud.go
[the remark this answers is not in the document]
mickmister: @neallred I'll follow suit with a `/usage/integrations` endpoint

api4/cloud_test.go
vish9812: Yes, they are no longer relevant in new code.
Only permission test remains, because it's been not finalized yet.

model/usage.go
```diff
@@ -0,0 +1,8 @@
+// Copyright (c) 2015-present Mattermost, Inc. All Rights Reserved.
+// See LICENSE.txt for license information.
+
+package model
+
+type PostsUsage struct {
+	Count int `json:"count"`
```
ashishbhate, line 7: Max int is about 2.14 Billion. That's a huge number, but I think we should probably change this to int64.

vish9812: Done.

Willyfrog (approved):

model/client4.go
```diff
@@ -8078,3 +8082,21 @@ func (c *Client4) GetAppliedSchemaMigrations() ([]AppliedMigration, *Response, e
...
+	}
+	defer closeBody(r)
+
+	var usage *PostsUsage
+	if err = json.NewDecoder(r.Body).Decode(&usage); err != nil {
+		return nil, BuildResponse(r), err
+	}
+
+	return usage, BuildResponse(r), nil
+}
```
ashishbhate, line 8102: Non-blocker: This can just be: 

```suggestion
	err = json.NewDecoder(r.Body).Decode(&usage)
	return usage, BuildResponse(r), err
}
```

model/post.go
```diff
@@ -268,6 +268,14 @@ type GetPostsOptions struct {
 	Direction                string // Only accepts up|down. Indicates the order in which to send the items.
 }
 
+type PostCountOptions struct {
+	// Only include posts on a specific team. "" for any team.
+	TeamId          string
+	MustHaveFile    bool
+	MustHaveHashtag bool
+	ExcludeDeleted  bool
```
ashishbhate, line 276: I suggest inverting this to "IncludeDeleted", I think the convention in our app and store code is to exclude deleted by default, and have an `IncludeDeleted` option.

ashishbhate (changes requested):
Looking great @vish9812! 

One minor change requested, and we're good to go!

Other comment is a non-blocker.

model/post.go
vish9812: This func **includes** the deleted posts by default. So, that's why I added the `ExcludeDeleted` option to keep the func default behaviour and to filter out for specific cases only.

ashishbhate: SGTM

ashishbhate (approved):

ashishbhate:
@Willyfrog @jgilliam17 Since this API endpoint doesn't have a consumer yet (hasn't been implemented on the client-size), I think we can defer QA on this PR.

amyblais:
@Willyfrog Is this needed for May 25th cloud release?

Willyfrog:
No, this is for v7

amyblais:
> > No, this is for v7
> 
> @Willyfrog v7.0 release branch will be cut based off Cloud May 25th cloud release, is this PR something that shouldn't yet go into a Cloud release and should be cherry-picked later to v7.0 release branch?

@Willyfrog @marianunez Just following up on this question to ensure I cherry-pick this to the correct place as needed.

marianunez:
> @Willyfrog @marianunez Just following up on this question to ensure I cherry-pick this to the correct place as needed.

This is not needed for v7.0 specifically but for Cloud release June 15th. I believe it can wait for the next fast-forward, unless this is needed in the `cloud` branch for early testing @neallred (it is low risk for May 25th release as any consumer of this API is behind feature flags at this point)

PR #20160, opened 2022-05-09
[MM-44056] Make Calls plugin's state on Cloud controllable by feature flag

streamer45:
#### Summary

Fortunately it looks like Apps solved our problem already so I am following a similar approach.

- Calls will be pre-packaged and set to enabled by default which means it gets installed automatically on first server start.
- Upon actual plugin's activation we check whether the `CallsEnabled` feature flag is set to true and only in that case we allow the plugin to be started.

None of this should impact on-prem as all the checks are guarded by the `IsCloud` util (leveraging `MM_CLOUD_INSTALLATION_ID`).

Since we are here, I also added the code to pre-package as it will make it easier to test this PR. Just bear in mind I may have to bump the plugin's version before merging to get the latest changes necessary for it to work properly on Cloud.

#### Ticket Link

https://mattermost.atlassian.net/browse/MM-44056

#### Release Note

```release-note
Prepackage Calls v0.4.9
```

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -119,6 +119,7 @@ TEMPLATES_DIR=templates
 PLUGIN_PACKAGES ?= mattermost-plugin-antivirus-v0.1.2
 PLUGIN_PACKAGES += mattermost-plugin-autolink-v1.2.2
 PLUGIN_PACKAGES += mattermost-plugin-aws-SNS-v1.2.0
+PLUGIN_PACKAGES += mattermost-plugin-calls-v0.4.9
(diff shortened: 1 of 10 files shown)


streamer45:
> The file `Makefile` is in the blocklist for external contributors. Hence, these changes are not tested by the CI pipeline active until the build is re-triggered by a core committer or the PR is merged. Please be careful when reviewing it. /cc @mattermost/core-security @mattermost/core-build-engineers

Oh well, I pushed to my fork by mistake really :grimacing:

PR #20161, opened 2022-05-09
[MM-44056] Make Calls plugin's state on Cloud controllable by feature flag

streamer45:
#### Summary

Fortunately it looks like Apps solved our problem already so I am following a similar approach.

- Calls will be pre-packaged and set to enabled by default which means it gets installed automatically on first server start.
- Upon actual plugin's activation we check whether the `CallsEnabled` feature flag is set to true and only in that case we allow the plugin to be started.

None of this should impact on-prem as all the checks are guarded by the `IsCloud` util (leveraging `MM_CLOUD_INSTALLATION_ID`).

Since we are here, I also added the code to pre-package as it will make it easier to test this PR. Just bear in mind I may have to bump the plugin's version before merging to get the latest changes necessary for it to work properly on Cloud.

#### Ticket Link

https://mattermost.atlassian.net/browse/MM-44056

#### Release Note

```release-note
Prepackage Calls v0.4.9
```


app/plugin.go
```diff
@@ -98,11 +98,8 @@ func (ch *Channels) syncPluginsActiveState() {
 				pluginEnabled = state.Enable
 			}
 
-			// Tie Apps proxy disabled status to the feature flag.
-			if pluginID == "com.mattermost.apps" {
-				if !ch.cfgSvc.Config().FeatureFlags.AppsEnabled {
-					pluginEnabled = false
-				}
+			if overrides, value := ch.getPluginStateOverride(pluginID); overrides {
```
lieut-data, line 101: nit:
```suggestion
			if hasOverride, value := ch.getPluginStateOverride(pluginID); hasOverride {
```

app/plugin_install.go
```diff
@@ -393,9 +393,10 @@ func (ch *Channels) installExtractedPlugin(manifest *model.Manifest, fromPluginD
 	// Activate the plugin if enabled.
 	pluginState := ch.cfgSvc.Config().PluginSettings.PluginStates[manifest.Id]
 	if pluginState != nil && pluginState.Enable {
-		if manifest.Id == "com.mattermost.apps" && !ch.cfgSvc.Config().FeatureFlags.AppsEnabled {
+		if overrides, value := ch.getPluginStateOverride(manifest.Id); overrides && !value {
```
lieut-data, line 396:
```suggestion
		if hasOverride, enabled := ch.getPluginStateOverride(manifest.Id); hasOverride && !enabled {
```

model/feature_flags.go
```diff
@@ -32,12 +32,16 @@ type FeatureFlags struct {
 	PluginPlaybooks  string `plugin_id:"playbooks"`
 	PluginApps       string `plugin_id:"com.mattermost.apps"`
 	PluginFocalboard string `plugin_id:"focalboard"`
+	PluginCalls      string `plugin_id:"com.mattermost.calls"`
```
lieut-data, line 35: I /think/ most of these are now unused, but no objection to extending.

lieut-data (approved):
Nice! Just one suggestion below.

cpoile (approved):
Thanks Claudio, looks great!

model/feature_flags.go
streamer45: Thanks, wasn't sure either but given there was a check that made use of those in `installFeatureFlagPlugins` I opted to include it although it shouldn't have any impact on the PR's scope.

streamer45:
@saturninoabril Could you suggest someone to have a quick look at this from a QA perspective? If possible, we are looking to merge this asap to get into tomorrow's Cloud release.

Main things to check would be:

- Calls is installed but not enabled by default. No UI related to Calls should be visible to the user at this point.
- Switching the `CallsEnabled` feature flag to `true` enables the plugin.
- Switching the `CallsEnabled` feature flag to `false` disables the plugin.
- Apps is enabled by default.
- Switching the `AppsEnabled` feature flag to `false` disables the plugin.

/cc @amyblais

streamer45:
@angeloskyratzakos This landing on Community will likely mean we need to have the flag on in order for Calls to remain enabled. So it's about setting `MM_FEATUREFLAGS_CALLSENABLED=true` . Let me know if you have questions/concerns.

saturninoabril:
Claudio, I'll take a look.

streamer45:
> Claudio, I'll take a look.

Amazing, thank you :raised_hands:

saturninoabril (approved):
Thanks @streamer45! Tested and passed.

PR #20163, opened 2022-05-09
MM-43529 - start freemium trial

pvev:
#### Summary
This are the web-server changes to support starting a trial from cloud freemium

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-43529

#### Related Pull Requests
MM Webapp: https://github.com/mattermost/mattermost-webapp/pull/10297
Enterprise: https://github.com/mattermost/enterprise/pull/1197

#### Release Note
```release-note
NONE
```


api4/cloud.go
```diff
@@ -40,6 +40,9 @@ func (api *API) InitCloud() {
 	api.BaseRoutes.Cloud.Handle("/subscription/invoices/{invoice_id:in_[A-Za-z0-9]+}/pdf", api.APISessionRequired(getSubscriptionInvoicePDF)).Methods("GET")
 	api.BaseRoutes.Cloud.Handle("/subscription", api.APISessionRequired(changeSubscription)).Methods("PUT")
 
+	// GET /api/v4/cloud/request-trial
+	api.BaseRoutes.Cloud.Handle("/request-trial", api.APISessionRequired(requestTrial)).Methods("GET")
```
neallred, line 44: Just noticed while reviewing the webapp PR that this route is currently a `GET`. Even though there's no body in the request, `GET` are typically used for requests that don't change the state of the system. Since this is altering the subscription by adding a trial to it, could we instead make this a `PUT` (and update the webapp client to also be a `PUT`)?

pvev: now that you mention that, yes, I agree that there is actually a change in the state of the license. Changing it. Thanks for the comment!

api4/cloud.go
```diff
@@ -120,6 +123,38 @@ func changeSubscription(c *Context, w http.ResponseWriter, r *http.Request) {
 	w.Write(json)
 }
 
+func requestTrial(c *Context, w http.ResponseWriter, r *http.Request) {
```
AGMETEOR, line 126: @pvev just a test for this endpoint

api4/cloud.go
```diff
@@ -121,6 +124,43 @@ func changeSubscription(c *Context, w http.ResponseWriter, r *http.Request) {
 	w.Write(json)
 }
 
+func requestCloudTrial(c *Context, w http.ResponseWriter, r *http.Request) {
+	if c.App.Channels().License() == nil || !*c.App.Channels().License().Features.Cloud {
+		c.Err = model.NewAppError("Api4.requestCloudTrial", "api.cloud.license_error", nil, "", http.StatusInternalServerError)
```
neallred, line 129: In a recent PR it was suggested that for non cloud instances that get cloud endpoint specific traffic, we should return `http.StatusForbidden`: https://github.com/mattermost/mattermost-server/pull/20152#discussion_r872013975

pvev: ok

AGMETEOR (approved):

PR #20165, opened 2022-05-10
MM-44045: Fix double plugin creation due to recursive plugin activation

agnivade:
The (*Environment).Activate method only guarded against multiple calls
and therefore wasn't properly re-entrant.

If a plugin calls an UpdateConfiguration in the OnActivate hook, it would
again call this method, resulting in 2 plugin processes getting created.

To prevent that from happening, we pre-emptively set the status to running
just after the supervisor is created.

I am not sure why starting the plugin normally from the server does not
cause this because the root cause is the same and there should be 2 plugin
processes created in the normal scenario as well. I have not spent time
looking into that. But let me know if you want me to.

Unit test: This might be a bit hard to test via a unit test. I think
an e2e test maybe written in the playbooks test itself which counts
the number of processes that the server spawns.

https://mattermost.atlassian.net/browse/MM-44045

```release-note
NONE
```

### DIFF
--- a/plugin/environment.go
+++ b/plugin/environment.go
@@ -274,6 +274,15 @@ func (env *Environment) Activate(id string) (manifest *model.Manifest, activated
 			return nil, false, errors.Wrapf(err, "unable to start plugin: %v", id)
 		}
 
+		// We pre-emptively set the state to running to prevent re-entrancy issues.
+		// The plugin's OnActivate hook can in-turn call UpdateConfiguration
+		// which again calls this method. This method is guarded against multiple calls,
+		// but fails if it is called recursively.
+		//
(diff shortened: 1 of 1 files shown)


trilopin (approved):
That's great!

I've just tested with playbooks e2e tests and no additional processes have been created.

agnivade:
I don't think there are any side-effects. Some sort of optimistic locking of the status seems like a right solution.

agnivade:
@saturninoabril - Could you please do a quick check on the e2e test and let me know if everything looks alright? :)

saturninoabril:
Found an error with one job. I've restarted and will keep you updated.

saturninoabril:
@agnivade all good with E2E. No related failed test.

PR #20176, opened 2022-05-10
MM-44088: Add teamID filter to channelMembers

agnivade:
We add 2 new params to channel members query.
1. Filter by teamId.
2. Negate that filter.

We include some more optimizations like:
- Moved the team role checks inside the dataloader.
- Moved the channel pretty name computation inside the loader.

Now that we load less data on initial load, we can reduce
the concurrency requirement to be a bit on the safer side.

```release-note
NONE
```

### DIFF
--- a/api4/graphql.go
+++ b/api4/graphql.go
@@ -34,7 +34,7 @@ const (
 	usersLoaderCtx    ctxKey = 4
 )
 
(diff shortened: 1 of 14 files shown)


isacikgoz (approved):
LGTM 💯

PR #20180, opened 2022-05-11
MM-43775: Use conn.NetConn() to get TCP conn from tls.Conn

agnivade:
We use the new API in Go 1.18 to set TCP_NO_DELAY for tls connections
as well.

Also took this chance to add the error handling while setting the read deadline.

https://mattermost.atlassian.net/browse/MM-43775

```release-note
NONE
```

### DIFF
--- a/app/web_conn.go
+++ b/app/web_conn.go
@@ -5,6 +5,7 @@ package app
 
 import (
 	"bytes"
+	"crypto/tls"
 	"encoding/json"
 	"errors"
 	"fmt"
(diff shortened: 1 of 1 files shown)


app/web_conn.go
```diff
@@ -334,7 +343,9 @@ func (wc *WebConn) readPump() {
 	wc.WebSocket.SetReadLimit(model.SocketMaxMessageSizeKb)
 	wc.WebSocket.SetReadDeadline(time.Now().Add(pongWaitTime))
 	wc.WebSocket.SetPongHandler(func(string) error {
-		wc.WebSocket.SetReadDeadline(time.Now().Add(pongWaitTime))
+		if err := wc.WebSocket.SetReadDeadline(time.Now().Add(pongWaitTime)); err != nil {
+			return err
```
streamer45, line 347: How is the error surfaced? Through the NextReader call?

streamer45 (commented):
Nice, just one question :+1:

app/web_conn.go
agnivade: Yep

PR #20197, opened 2022-05-12
Fixes a null pointer if the plugin requests the limits on a non-cloud server

mgdelacroix:
#### Summary
This PR checks if the app has the cloud interface set before trying to call it to avoid a null pointer when using the get limits plugin API endpoint in non-cloud mode.

```release-note
NONE
```


neallred (approved):
Thanks, @mgdelacroix !

PR #20198, opened 2022-05-12
MM-41569: Prevent compliance directory to be changed on cloud instances

agnivade:
https://mattermost.atlassian.net/browse/MM-41659

```release-note
NONE
```


agnivade:
@isacikgoz - I realized that there are more settings to block after finishing this. I gave a comment on https://mattermost.atlassian.net/browse/MM-41660 to extend the whole logic and make it generic. But the locations will be this two functions.

isacikgoz (approved):

PR #20217, opened 2022-05-17
MM-42379: Do not set TrustedProxyIPHeader for new instances

agnivade:
We move away with the update logic and just set an empty
slice always.

```release-note
The value of ServiceSettings.TrustedProxyIPHeader will default
to empty from now on. A previous bug prevented this from happening
in certain conditions. Customers are requested to check for these
values in their config and set them to nil if necessary.
```

### DIFF
--- a/model/config.go
+++ b/model/config.go
@@ -545,13 +545,7 @@ func (s *ServiceSettings) SetDefaults(isUpdate bool) {
 		s.Forward80To443 = NewBool(false)
 	}
 
-	if isUpdate {
-		// When updating an existing configuration, ensure that defaults are set.
-		if s.TrustedProxyIPHeader == nil {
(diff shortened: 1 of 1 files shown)


jupenur (approved):

srkgupta (approved):

amyblais:
@agnivade Is this needed for v7.0?

srkgupta:
Hi @amyblais
Yes, we should cherry pick this for 7.0 release as well. Details in the JIRA ticket
https://mattermost.atlassian.net/browse/MM-42379


Also this needs to be patched in previous 3 release versions along with the following announcement post release:

```
Customers are requested to check for these
values in their config and set them to nil if necessary.
```

PR #20225, opened 2022-05-17
MM-43918: freemium limit for storage

ashishbhate:
#### Summary

- Adds an API end point to return storage usage

#### Ticket Link

- https://mattermost.atlassian.net/browse/MM-43918

#### Release Note

```release-note
NONE
```


app/usage.go
```diff
@@ -5,11 +5,22 @@ package app
...
 
+var usagesCache = cache.NewLRU(cache.LRUOptions{
+	Size:          1,
+	DefaultExpiry: time.Hour,
+})
+
+const (
+	UsageStorageKey = "storage"
+)
+
```
ashishbhate, line 23: @agnivade This PR is currently WIP while I add tests and some polish. I wanted to get some early feedback.

Originally the idea was to calculate the storage usage on every file upload (through the api/app), but turns out plugins often just directly update the `fileInfo` table (e.g. https://github.com/mattermost/focalboard/pull/3052/files#r874707796)

I'm going to work on getting all Mattermost plugins to use the api/app, but right now we're missing some Plugin API methods that plugins could use. There could also be plugins which we don't control that could directly update the FileInfo table. 

This PR can't wait on those changes, so decided I to use a cache here (credit to @vish9812  for the idea)
I'm not sure this is a permanent change, so I didn't opt to make the extensive localcachelayer changes. 

What do you think?

api4/usage.go
```diff
@@ -30,3 +32,20 @@ func getPostsUsage(c *Context, w http.ResponseWriter, r *http.Request) {
...
 }
+
+func getStorageUsage(c *Context, w http.ResponseWriter, r *http.Request) {
+	usage, appErr := c.App.GetStorageUsage()
+	if appErr != nil {
+		c.Err = model.NewAppError("Api4.getPostsUsage", "app.post.analytics_posts_count.app_error", nil, appErr.Error(), http.StatusInternalServerError)
+		return
+	}
+
+	usage = utils.RoundOffToZeroes(float64(usage))
```
ashishbhate, line 43: Rounding off in the api layer instead of the app layer.

store/localcachelayer/file_info_layer.go
```diff
@@ -67,3 +67,33 @@ func (s LocalCacheFileInfoStore) InvalidateFileInfosForPostCache(postId string,
...
+		return usage, nil
+	}
+
+	usage, err := s.FileInfoStore.GetStorageUsage(allowFromCache, includeDeleted)
+	if err != nil {
+		return 0, err
+	}
+
+	s.rootStore.doStandardAddToCache(s.rootStore.fileInfoCache, storageUsageKey, usage)
+	return usage, nil
```
vish9812, line 98: This whole thing can be extracted into a local func.
Then you can just do 
`return getStorageUsage(allowFromCache, includeDeleted)`
And same return can be used to replace the `!allowFromCache` block.

store/sqlstore/file_info_store.go
```diff
@@ -734,3 +734,20 @@ func (fs SqlFileInfoStore) GetFilesBatchForIndexing(startTime int64, startFileID
 
 	return files, nil
 }
+
+func (fs SqlFileInfoStore) GetStorageUsage(allowFromCache, includeDeleted bool) (int64, error) {
```
vish9812, line 738: Why pass `allowFromCache` here?

app/usage.go
```diff
@@ -19,3 +20,13 @@ func (a *App) GetPostsUsage() (int64, *model.AppError) {
 
 	return utils.RoundOffToZeroes(float64(count)), nil
 }
+
+// GetStorageUsage returns the sum of files stored
+func (a *App) GetStorageUsage() (int64, *model.AppError) {
```
vish9812, line 25: Missing test.

store/localcachelayer/file_info_layer.go
```diff
@@ -67,3 +67,33 @@ func (s LocalCacheFileInfoStore) InvalidateFileInfosForPostCache(postId string,
 		s.rootStore.metrics.IncrementMemCacheInvalidationCounter("File Info Cache - Remove by PostId")
 	}
 }
+
+func (s LocalCacheFileInfoStore) GetStorageUsage(allowFromCache, includeDeleted bool) (int64, error) {
```
vish9812, line 71: Missing test.

vish9812 (changes requested):

api4/usage_test.go
```diff
@@ -37,3 +37,17 @@ func TestGetPostsUsage(t *testing.T) {
 		assert.Equal(t, int64(10), usage.Count)
 	})
 }
+
+func TestGetStorageUsage(t *testing.T) {
+	t.Run("unauthenticated users cannot access", func(t *testing.T) {
```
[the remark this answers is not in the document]
ashishbhate: I left it out because I don't think it adds anything valuable. The test would essentially be testing the session/auth code.

store/localcachelayer/file_info_layer.go
ashishbhate: I think it's easier to read this way. I'm not sure a small 4 line function is helpful in this case.

store/sqlstore/file_info_store.go
ashishbhate: Our struct embedding layer code needs it.

app/usage.go
ashishbhate: I left it out because I don't think it adds anything valuable. The test would essentially be doing the same exact thing as the store test.

store/localcachelayer/file_info_layer.go
ashishbhate: I left it out because I don't think it adds anything valuable. The test would essentially be testing the caching logic.

api4/usage.go
vish9812: The purpose of rounding here most likely is to actually hide the real figures from the api consumers.

app/app_iface.go
```diff
@@ -200,6 +201,8 @@ type AppIface interface {
 	// GetSessionLengthInMillis returns the session length, in milliseconds,
 	// based on the type of session (Mobile, SSO, Web/LDAP).
 	GetSessionLengthInMillis(session *model.Session) int64
+	// GetStorageUsage returns the sum of files stored
```
Willyfrog, line 204:
```suggestion
	// GetStorageUsage returns the sum of file's size stored in the instance
```

Willyfrog (approved):
no major blockers, so aproving :)

vish9812 (approved):

agnivade:
@ashishbhate - Anything in specific you wanted me to look at? Also checking on what's the plan to add load-test coverage for this?

ashishbhate:
>@ashishbhate - Anything in specific you wanted me to look at? Also checking on what's the plan to add load-test coverage for this?

The client-side consumer of these APIs haven't been implemented yet. The consumer and this method are meant specifically for cloud and will be called from behind feature flags (though the server API itself isn't behind feature flags). The caching is defensive for now, and  currently only for reporting.

Based on the way the API is used (especially when we enforce the limits in an upcoming cloud release), I will push for load testing, and possibly changes to the implementation.

ashishbhate:
This needs to go in for the June 15th release of cloud. Merging for now. We can make fixes in follow-up PRs.

PR #20238, opened 2022-05-18
MM-31353: Fix flaky TestNoticeValidation/notice_with_specific_date_check

agnivade:
We now generate the current time in UTC to avoid date transitions
when running the tests at different local times.

https://mattermost.atlassian.net/browse/MM-31353

```release-note
NONE
```


wiggin77 (approved):
LGTM 👍

PR #20239, opened 2022-05-18
cherrypick-50eaec6

larkox:
```release-note
Fix bug that allow to send test (empty) notifications even if the `SendPushNotifications` config is set to false.
```


amyblais:
There are some failing tests.

amyblais (approved):

PR #20247, opened 2022-05-18
[MM-44307] On Limits Change, archive or restore teams according to the limits

nickmisasi:
#### Summary
With the Freemium lifecycle, when a trial of Enterprise ends, the installation will be downgraded to Freemium. Freemium introduces certain limits to the workspace that need to be applied. This PR covers limits around the allowed number of _teams_ the installation can have. 

**AdjustTeamsFromProductLimits**

This function is called by the webhook handler for limit changes. It grabs the teams list, orders them by date of creation, and then archives any teams passed the allowed number of teams. When it archives them, it sets a new value in the db `CloudLimitsArchived` - which is a boolean value. This boolean value allows us to differentiate between installations archived explicitly by the user, and those that were archived by the automation.

When the teams limit _increases_ (for whatever reason) AdjustTeamsFromProductLimits will look through the list of archived teams, and restore teams up to the new limit, in order of creation (oldest first). This will ONLY restore teams that have a `CloudLimitsArchived` value of `true` - ie, if a user archives a team, and then the limits change, that team will not be restored. 

**softDeleteTeamsExcept**

In the future, users who choose to downgrade from Professional to Starter will be forced to choose which team they want to keep active, and the rest will be archived. The new endpoint mimics the current team deletion endpoint exactly, with a teamId query parameter. This endpoint will soft delete all teams _except_ for the teamId passed to the endpoint. 

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-44307

#### Release Note

```release-note
Teams Limit changes will propagate automatically
Added new endpoint DELETE /api/v4/teams/{team_id:[A-Za-z0-9]+}/except which will archive all teams except for the team_id passed to the endpoint
```

### DIFF
--- a/app/team_test.go
+++ b/app/team_test.go
@@ -553,6 +553,46 @@ func TestAdjustTeamsFromProductLimits(t *testing.T) {
 		}
 	})
 
+	t.Run("Should not do anything if the amount of teams is equal to the limit", func(t *testing.T) {
+
+		expectedTeamsList, err := th.App.GetAllTeams()
+
+		var expectedActiveTeams []*model.Team
+		var expectedCloudArchivedTeams []*model.Team
+		for _, team := range expectedTeamsList {
+			if team.DeleteAt == 0 {
+				expectedActiveTeams = append(expectedActiveTeams, team)
+			}
+			if team.DeleteAt > 0 && team.CloudLimitsArchived {
+				expectedCloudArchivedTeams = append(expectedCloudArchivedTeams, team)
+			}
+		}
+
+		require.Nil(t, err)
+
+		activeLimit := len(expectedActiveTeams)
+		teamLimits := &model.TeamsLimits{Active: &activeLimit}
+		err = th.App.AdjustTeamsFromProductLimits(teamLimits)
+		require.Nil(t, err)
+
+		actualTeamsList, err := th.App.GetAllTeams()
+
+		require.Nil(t, err)
+		var actualActiveTeams []*model.Team
+		var actualCloudArchivedTeams []*model.Team
+		for _, team := range actualTeamsList {
+			if team.DeleteAt == 0 {
+				actualActiveTeams = append(actualActiveTeams, team)
+			}
+			if team.DeleteAt > 0 && team.CloudLimitsArchived {
+				actualCloudArchivedTeams = append(actualCloudArchivedTeams, team)
+			}
+		}
+
+		require.Equal(t, len(expectedActiveTeams), len(actualActiveTeams))
+		require.Equal(t, len(expectedCloudArchivedTeams), len(actualCloudArchivedTeams))
+	})
+
 	t.Run("Should restore archived teams if limit increases", func(t *testing.T) {
 		activeLimit := 1
 		teamLimits := &model.TeamsLimits{Active: &activeLimit}


app/cloud.go
```diff
@@ -37,6 +37,17 @@ func (a *App) SendPaymentFailedEmail(failedPayment *model.FailedPayment) *model.
 	return nil
 }
 
+func (a *App) AdjustInProductLimits(limits *model.ProductLimits, subscription *model.Subscription) *model.AppError {
+	if limits.Teams != nil && *limits.Teams.Active > 0 {
```
neallred, line 41: I think we need to nil check `teamLimits.Active`, too. Small repro of what would go wrong if `Active` was nil:

```go
package main

import (
	"fmt"
)

type TeamsLimits struct {
	Active *int
}

func main() {
	teamLimits := &TeamsLimits{
		Active: nil,
	}
	if 3 > *teamLimits.Active {
		fmt.Println("greater")
	} else {
		fmt.Println("leq")
	}
}
```

```
go run tmp.go
[signal SIGSEGV: segmentation violation code=0x1 addr=0x0 pc=0x47df5a]

goroutine 1 [running]:
main.main()
	/home/n/tmp.go:15 +0x1a
exit status 2
```

app/team.go
```diff
@@ -29,6 +30,105 @@ import (
...
+	var cloudArchivedTeams []*model.Team
+	for _, team := range teams {
+		if team.DeleteAt == 0 {
+			activeTeams = append(activeTeams, team)
+		}
+		if team.DeleteAt > 0 && *team.CloudLimitsArchived {
+			cloudArchivedTeams = append(cloudArchivedTeams, team)
+		}
+	}
+
```
neallred, line 57: Nit: since we read `*teamLimits.Active` multiple times, could we make a variable of it, something like `maxActiveTeams := *teamLimits.Active`

app/team.go
```diff
@@ -29,6 +30,105 @@ import (
...
+	teams, appErr := a.GetAllTeams()
+	if appErr != nil {
+		return appErr
+	}
+
+	if teams == nil {
+		return nil
+	}
+	cloudLimitsArchived := false
+	patch := &model.TeamPatch{CloudLimitsArchived: &cloudLimitsArchived}
```
neallred, line 115: Since the intention of this is to mimic `AdjustTeamsFromProductLimits`, should `cloudLimitsArchived` be true so that if the admin upgrades again, their archived teams can be automatically unarchived?

app/team_test.go
```diff
@@ -466,6 +466,155 @@ func TestAddUserToTeamByTeamId(t *testing.T) {
...
+
+	err := th.App.SoftDeleteAllTeamsExcept(teamId)
+	assert.Nil(t, err)
+	allTeams, err := th.App.GetAllTeams()
+	require.Nil(t, err)
+	for _, team := range allTeams {
+		if team.Id == teamId {
+			require.Equal(t, int64(0), team.DeleteAt)
+		} else {
+			require.NotEqual(t, int64(0), team.DeleteAt)
```
neallred, line 496: Should we add asserts for `CloudLimitsArchived` to the assertions in these two branches?

app/team_test.go
```diff
@@ -466,6 +466,155 @@ func TestAddUserToTeamByTeamId(t *testing.T) {
...
+		}
+	})
+
+	t.Run("Should restore archived teams if limit increases", func(t *testing.T) {
+		activeLimit := 1
+		teamLimits := &model.TeamsLimits{Active: &activeLimit}
+
+		err := th.App.AdjustTeamsFromProductLimits(teamLimits)
+		require.Nil(t, err)
+		activeLimit = 10000 // make the limit extremely high to all are enabled
```
neallred, line 560: Nit: `to all are enabled` -> `so all are enabled` ?

app/team_test.go
```diff
@@ -466,6 +466,155 @@ func TestAddUserToTeamByTeamId(t *testing.T) {
...
+		err := th.App.AdjustTeamsFromProductLimits(teamLimits)
+		require.Nil(t, err)
+
+		cloudLimitsArchived := false
+		patch := &model.TeamPatch{CloudLimitsArchived: &cloudLimitsArchived}
+		team, err := th.App.PatchTeam(teamIds[0], patch)
+		require.Nil(t, err)
+		require.Equal(t, false, team.CloudLimitsArchived)
+
+		activeLimit = 10000 // make the limit extremely high to all are enabled
```
neallred, line 596: Nit: `to all are enabled` -> `so all are enabled` ?

neallred (changes requested):
Overall looks good to me; there were a couple conditions I thought needed adding or tweaking.

app/team_test.go
nickmisasi: Added

app/team.go
nickmisasi: Good catch, I did intend this to be true

app/team.go
nickmisasi: updated

app/cloud.go
nickmisasi: ah I had assumed that no change in this limit would mean that the TeamsLimits struct would be nil - but yeah good to be safe anyways

neallred: Ahh yeah that's true today. Could change if teams got more/different limits and not all of them were defined.

neallred (approved):
LGTM, thanks @nickmisasi !

db/migrations/mysql/000085_add_cloud_limits_archived.down.sql
```diff
@@ -0,0 +1,14 @@
+SET @preparedStatement = (SELECT IF(
+	EXISTS (
+		SELECT 1 FROM INFORMATION_SCHEMA.COLUMNS
+		WHERE table_name = 'Teams'
+		AND table_schema = DATABASE()
+		AND column_name = 'CloudLimitsArchived'
+	),
+	'ALTER TABLE Teams DROP COLUMN IF EXISTS CloudLimitsArchived;',
```
ashishbhate, line 8: `TestUpAndDownMigrations/Should_be_reversible_for_mysql`

is failing because MySQL doesn't support `IF EXISTS`. That's why we do the convoluted EXISTS check a few lines above. Its a workaround for MySQL not supporting the `IF EXISTS` check in-line.

nickmisasi: Thanks

app/team_test.go
```diff
@@ -466,6 +466,157 @@ func TestAddUserToTeamByTeamId(t *testing.T) {
...
+			Type:        model.TeamOpen,
+		},
+	}
+	teamIds := []string{}
+	for _, create := range teams {
+		team, err := th.App.CreateTeam(th.Context, create)
+		require.Nil(t, err)
+		teamIds = append(teamIds, team.Id)
+	}
+	t.Run("Should soft delete teams if there are more teams than the limit", func(t *testing.T) {
```
marianunez, line 533: Could we include a case of no team is archived if the limit is the same as the amount of teams (the case of trial ended and they only have one team - nothing should happen)

marianunez (approved):
Great work @nickmisasi 👍 Just a non-blocking request

nickmisasi:
@stevemudie this one requires a connection between Stripe > CWS > Installation so will be QA Deferred

PR #20254, opened 2022-05-20
[MM-44263] Cache Posts Usage

vish9812:
#### Summary

Enable caching of posts count for API - GET /api/v4/usage/posts

#### Ticket Link

JIRA - https://mattermost.atlassian.net/browse/MM-44265

#### Related Pull Requests
- Dependent on changes in https://github.com/mattermost/mattermost-server/pull/20250

#### Release Note

```release-note
Improved performance of API - GET /api/v4/usage/posts, by enabling caching of posts count
```


Willyfrog (approved):

vish9812:
This branch is cut-off from `MM-44263_user-posts`, so it includes the https://github.com/mattermost/mattermost-server/pull/20250 PR changes as well. You can ignore those changes.

store/localcachelayer/post_layer.go
```diff
@@ -131,3 +141,25 @@ func (s LocalCachePostStore) GetPosts(options model.GetPostsOptions, allowFromCa
 
 	return list, err
 }
+
+func (s LocalCachePostStore) AnalyticsPostCount(options *model.PostCountOptions) (int64, error) {
+	if !options.AllowFromCache || options.MustHaveFile || options.MustHaveHashtag || !options.UsersPostsOnly || !options.ExcludeDeleted || options.TeamId != "" {
```
ashishbhate, line 146: Could we add a docstring above this function that says it only looks at the cache in the case `UsersPostsOnly` and `ExcludeDeleted` are not set?

model/post.go
```diff
@@ -280,6 +280,8 @@ type PostCountOptions struct {
 	MustHaveFile    bool
 	MustHaveHashtag bool
 	ExcludeDeleted  bool
+	UsersPostsOnly  bool
+	AllowFromCache  bool
```
ashishbhate, line 284: Same as below, can we add a comment here that says `AllowCache` only works if `ExcludeDeleted` and `UserPostsOnly` are unset?

ashishbhate (commented):
In general this PR looks good, it caches the two main filter conditions we care about. 

Have we investigated tracking post counts in `app.CreatePost` and `app.DeletePost`? 

In #20225  I added a cache because plugins (Focalboard was) were directly writing to the FileInfo table (because our PlugIn API doesn't have the helper methods plugins need), so we couldn't hook into the Upload API to track file uploads.

Plugins and another extensions have API methods to create and delete Posts. AFAIK, none of the official Mattermost plugins (Focalboard and Playbooks) directly write to the Posts table. 
I don't see why they would either - by writing to the Posts table directly instead of using the API they're missing on important side-effects like notifications and mention count updates.

For post count atleast, I think it would be easier to just add the logic in the CreatePost and DeletePost app methods, along with a one-time migration to seed the initial count data. That way we wont have to think about the filter conditions and various combinations thereof.

ashishbhate:
The storage in the DB is not the main concern. I was suggesting tracking counts in `CreatePost` and `DeletePost` app methods. Those app methods are called every time a Post is created or deleted. They have an easy way to identify whether a Post is from a Bot.   

1. 
Create a row in the `Systems` table. The `key` could be `PostUsage` and the `value` column would be the actual post count.

2. 
In app.CreatePost:
- Check if message is a system message or from a bot. If it isn't, increment the count stored in `Systems` table (or whereever appropriate)

In app.DeletePost:
- Same as above, except decrement count.



The reason I think this merits investigation:

1. It amortizes the read cost. We have to run the `count (*) from post...` everytime we call the API method for usage. We're working around this by caching the result. But every hour or so, one API request will take a long time. Post writes happen rarely (relatively speaking), so it might be fine to pay a small performance penalty while creating (and deleting) a post to update the post count in some DB row some where. The added benefit is that we get realtime accurate post counts (might not be necessary to have). 

2. We won't have to bother with filter conditions. In the app methods we can check if the post is from a bot, or if its a system message. We don't have to use `AnalyticsPostCount` at all,. We simply have to SQL increment or decrement a row in the Systems table. There's no need for complicated filter logic for caching. 

4. The reason I implemented a cache for the FileInfo table instead of hooking into the app methods as suggested above is because plugins were directly accessing the FileInfo table (because there isn't an API method to update the FileInfo store through the PlugIn API). 
I don't think Plugins will directly write to the Posts table because (a) Plugin API methods exist to create Posts (b) Plugins need the side-effects of post creation, e.g. notifications, which they won't get if they write directly to the DB.

ashishbhate (approved):

PR #20260, opened 2022-05-22
Enable plugins for ARM64

bkmgit:
#### Summary
Enable building of ARM64 plugins

#### Release Note

```release-note
Added a new config setting linux-arm64 for building plugins.
```


bkmgit:
Related pull request: https://github.com/mattermost/mattermost-server/pull/18491

hanzei (commented):
This is not needed. Plugins can already declare, that they work for arm64. See https://github.com/mattermost/mattermost-plugin-starter-template/blob/2629a23a058b2b975fb47a366c7b23af7c15e3b3/plugin.json#L14 for an example.

PR #20263, opened 2022-05-23
MM-44412: correct usage of updateAt to createAt

ashishbhate:
#### Summary

- Fixes unread mention count under some edge cases. I discovered this while working on https://github.com/mattermost/mattermost-server/pull/20034. Since that PR is being punted for later, I've ported those fixes here, with tests. 
- Removed a lot of dead code. 

#### Ticket Link

- https://mattermost.atlassian.net/browse/MM-44412

#### Release Note

```release-note
NONE
```


app/channel.go
```diff
@@ -2585,66 +2585,6 @@ func (a *App) MarkChannelAsUnreadFromPost(postID string, userID string, collapse
...
-					mlog.Warn("Failed to encode thread to JSON")
-				}
-				message := model.NewWebSocketEvent(model.WebsocketEventThreadUpdated, channel.TeamId, "", userID, nil)
-				message.Add("thread", string(payload))
-				a.Publish(message)
-			}
-		}
-	}
-
 	channelUnread, nErr := a.Srv().Store.Channel().UpdateLastViewedAtPost(post, userID, unreadMentions, unreadMentionsRoot, true)
```
ashishbhate, line 2588: Discussed this with @koox00 on zoom

This is all dead code. If a threadMembership did not exist, we made one with `Following` set to false. Because `Following` was set to false, the condition on line 2618 was never true, and therefore counts and mentions are never updated.

Doesn't seem to be a need to create a ThreadMembership here, especially since `Following` is always false. 

I've removed all deprecated tests as well.

koox00 (approved):

Willyfrog (approved):

ashishbhate:
@jgilliam17 

This PR mostly removes dead code. It also fixes an edge case for mentions when a user (with CRT off) joins a channel and then marks a thread that he/she is mentioned in as unread. This edge-case is unit tested, so for QA here some smoke testing around marking threads an unread and checking mention counts should suffice.

jgilliam17:
Thanks @ashishbhate 
Tested, looks good to merge.
- Verified marking threads as unread with CRT on/off and mentions - as expected.
- [E2E report](https://mattermost-cypress-report.s3.amazonaws.com/202440-f67e985-onprem-ent-server-pr-20263-mattermost/mm-ee-testf67e985/mochawesome.html) looks good, no PR related failures ( 2 failed tests are crt related, but not failing when manually testing, possibly flaky)

jgilliam17 (approved):

PR #20282, opened 2022-05-24
[MM-44397] Restrict team creation based on subscription limits

nickmisasi:
#### Summary
This PR builds on: https://github.com/mattermost/mattermost-server/pull/20281

It does the same thing, except for the team creation endpoint rather than the restore endpoint.

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-44397

#### Release Note

```release-note
Users will be restricted from creating teams when at or above the teams limit for their plan.
```


api4/team.go
```diff
@@ -95,6 +95,29 @@ func createTeam(c *Context, w http.ResponseWriter, r *http.Request) {
 		return
 	}
 
+	// Freemium enabled, on a cloud license. We must check limits before allowing to restore
```
marianunez, line 98:
```suggestion
	// Freemium enabled, on a cloud license. We must check limits before allowing to create
```

api4/team.go
```diff
@@ -95,6 +95,29 @@ func createTeam(c *Context, w http.ResponseWriter, r *http.Request) {
...
+	if c.App.Config().FeatureFlags != nil && c.App.Config().FeatureFlags.CloudFree && (c.App.Channels().License() != nil && c.App.Channels().License().Features != nil && *c.App.Channels().License().Features.Cloud) {
```
marianunez, line 99: The feature flag check doesn't hurt but in theory if Freemium is not enabled limits are empty so it will just not guard below

api4/team_test.go
```diff
@@ -70,27 +70,78 @@ func TestCreateTeam(t *testing.T) {
...
+				Active: model.NewInt(1),
+			},
+		}, nil).Once()
+		team := &model.Team{Name: GenerateTestUsername(), DisplayName: "Some Team", Type: model.TeamOpen}
+		_, resp, err := th.Client.CreateTeam(team)
+		require.Error(t, err)
+		CheckBadRequestStatus(t, resp)
+	})
+
+	t.Run("cloud below limit returns 200", func(t *testing.T) {
```
marianunez, line 121: Same comment here if we could add case for non-cloud/no limits

marianunez (approved):
LGTM - non blocking comments

api4/team.go
nickmisasi: Going to leave it for explicitness - mainly so that the code gets double checked when the feature flag is removed down the line

neallred (approved):

stevemudie (commented):
Deferring for testing post merge to allow full on review alongside https://github.com/mattermost/mattermost-server/pull/20281

PR #20289, opened 2022-05-24
Audit logging v2 -- new schema

oh6hay:
This PR adds the new `event_data` key in the audit log record. Also, the [Channel](api4/channel.go) class demonstrates how populate the `event_data` record.

The proposed new audit log schema includes an `event_data` structure for every audit log record (that inserts, modifies, or deletes any data). The objective of this is to make processing audit logs easier. The `event_data` structure contains the following keys:

* `new_data` contains the content submitted to the API endpoint being logged. In most cases this would be the JSON payload deserialized into an `interface{}`.
* `prior_state` is the object affected by the API call, prior to applying whatever the API being called does.
* `resulting_state` is the result of the API call.
* `result_object_type` is the type of the result object.
* `parameters` (not sure this is necessary) would contain any parameters relevant to the API operation. For example, some state-modifying API endpoints do not expect an actual payload (in a POST body).

This new schema affects programming practices in a few ways:

* Instead of deserializing the request body (`r.Body` in most API endpoints) directly into a model object of interest, the payload should be put aside for audit logging. For example, before:
```
	var oauthApp model.OAuthApp
	if jsonErr := json.NewDecoder(r.Body).Decode(&oauthApp); jsonErr != nil {
```
would become:
```
	postBody, readErr := ioutil.ReadAll(r.Body)
	if readErr != nil {
		return
	}
	defer r.Body.Close()

	var postPayload interface{}
	_ = json.NewDecoder(bytes.NewBuffer(postBody)).Decode(&postPayload)

	var oauthApp model.OAuthApp
	if jsonErr := json.NewDecoder(bytes.NewBuffer(postBody)).Decode(&oauthApp); jsonErr != nil
```
* Similarly, put aside the model objects prior and after any changes within the API call so they can be added to the audit record object

Also please note that implementing the new audit log schema does not involve getting rid of any of the current audit log implementation (at least for now) in order not to introduce breaking changes.

### Note

The `prior_state` and `resulting_state` should be model objects but they should also be sanitized (eg. not include sensitive info like passwords or message content?) -- model objects need to implement the `Auditable` interface which requires objects to provide a copy or representation of itself suitable for audit logging. The `AuditableObject()` method can for example either return the model object as is, or filter out certain fields.


jupenur (commented):
Looks like we're not sanitizing any of the new stuff that goes into audit logs in any way. The channel APIs aren't *particularly* sensitive, but I'm still not entirely convinced that something like the full channel header markdown needs to end up in the logs.

Even if there's nothing to sanitize in these APIs specifically, what's our strategy for the APIs that *do* need sanitization? The existing schema uses `mode.AuditModelTypeConv` but that doesn't work here, at least not directly.

api4/channel.go
```diff
@@ -79,8 +81,17 @@ func (api *API) InitChannel() {
 }
 
 func createChannel(c *Context, w http.ResponseWriter, r *http.Request) {
+	postBody, readErr := ioutil.ReadAll(r.Body)
+	if readErr != nil {
+		return
+	}
+	defer r.Body.Close()
```
jupenur, line 88: Handlers don't need to close the body: https://pkg.go.dev/net/http#:~:text=//%20The%20Server%20will%20close%20the%20request%20body.%20The%20ServeHTTP%0A%09//%20Handler%20does%20not%20need%20to.

oh6hay:
> Looks like we're not sanitizing any of the new stuff that goes into audit logs in any way. The channel APIs aren't _particularly_ sensitive, but I'm still not entirely convinced that something like the full channel header markdown needs to end up in the logs.
> 
> Even if there's nothing to sanitize in these APIs specifically, what's our strategy for the APIs that _do_ need sanitization? The existing schema uses `mode.AuditModelTypeConv` but that doesn't work here, at least not directly.

I pushed a comment that addresses model object sanitization by requiring that they (as in `prior_state` and `resulting_state`) implement the `Auditable` interface which has an `AuditableObject() interface{}` method -- this means any arbitrary objects can't be logged. So there needs to be a conscious choice of what to include in audit logs.

The `new_data` needs another approach though.

agnivade (commented):
Some questions:
1. Why are we decoding the payload twice? This is creating double allocations and double decoding overhead. Any reason we can't just clone the parsed payload?
2. Are we okay with adding the entire request body to the audit log? In some cases, request bodies are considerably huge and also contain binary data. Serializing a large binary data into a log file can cause performance issues. Have we looking into the performance aspect of this?

oh6hay:
hello!

>     1. Why are we decoding the payload twice? This is creating double allocations and double decoding overhead. Any reason we can't just clone the parsed payload?

the justification for the double decoding was to write the payload as is as it's not necessarily at all the same thing as the model object it's assumed to represent. Both approaches (logging payloads as they come, vs as the model objects they are deserialized into) can be argued to make sense. In this context though, after some initial discussion we thought it's better to write the parsed (as a model object) payload instead.

>     2. Are we okay with adding the entire request body to the audit log? In some cases, request bodies are considerably huge and also contain binary data. Serializing a large binary data into a log file can cause performance issues. Have we looked into the performance aspect of this?

this is a valid concern too, addressed by not logging the payloads as they are

oh6hay:
> @oh6hay Thanks for the PR. Could you include:
> 
>     1. Example for the new incl. added data audit log entry

before
```
{
  "api_path": "/api/v4/posts",
  "event": "createPost",
  "status": "success",
  "user_id": "s6mbcn4b4fdhbpquuu6qxh7cme",
  "session_id": "9qaesjg4dffs7bqn3cnb38bbme",
  "client": "Mozilla/5.0 (Windows NT 10.0; Win64; x64) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/97.0.4692.71 Safari/537.36",
  "ip_address": "127.0.0.1",
  "cluster_id": "",
  "post": {
    "id": "ctyqhq8gdtrftercg1ciwhi3hy",
    "channel_id": "qwx7sd1xatdfprcwyoheympmqo",
    "type": "",
    "pinned": false
  }
}
```

> 
>     2. Example of a previous audit call with the new implementation log entry

new schema
```
{
  "timestamp": "2022-05-31 13:12:02.370 +03:00",
  "api_path": "/api/v4/posts",
  "event": "createPost",
  "status": "success",
  "user_id": "1thotweb1iyufegag78apwbkya",
  "session_id": "yby3kcrdefrmpyi1e1fz843ndy",
  "client": "Mozilla/5.0 (X11; Linux x86_64) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/100.0.4896.75 Electron/18.0.3 Safari/537.36 Mattermost/5.1.0",
  "ip_address": "127.0.0.1",
  "event_data": {
    "parameters": null,
    "new_data": null,
    "prior_state": null,
    "resulting_state": null,
    "result_object_type": ""
  },
  "cluster_id": "ezfafrtbttfjfr5ht15a6z6fia",
  "post": {
    "id": "epx87wgqwp8szq5kct3q9rwmcc",
    "channel_id": "53nbi6jfz78fipmheh3zdk15ro",
    "type": "",
    "pinned": false
  }
}
```

note that the new schema with the `event_data` value is empty, this is due to missing implementation on what to actually include in the audit logs as the review discussien identified that model objects in the default case can't be just logged as is

DSchalla:
Thanks! I think there was a slight misunderstanding, but I think what you are saying is that the latter snippet would also be what it looks like for any "old" audit call that doesn't get the new metadata included. Do you have an example what a fully-blown example would look like? (Doesn't have to be all metadata as that's still tbd, but to have an example for everyone (e.g. the Customer Engineering team.))

api4/channel.go
```diff
@@ -379,6 +387,8 @@ func restoreChannel(c *Context, w http.ResponseWriter, r *http.Request) {
 	}
 	teamId := channel.TeamId
 
+	channelForAudit, _ := c.App.GetChannel(c.Params.ChannelId)
```
agnivade, line 390: Here and other places below as well.

agnivade (commented):
Looks good overall. We also need to run a load-test to ensure there are no performance regressions.

DSchalla:
@agnivade Since there is some time pressure to get the schema changes in for 7.x, would you feel comfortable if we add the full audit calls to less performance critical endpoints and perform the full load testing in the upcoming, separate PR where it gets rolled out more broadly? 

The plan is to do this rollout staged, so if the schema changes itself are no concern that would help us.

jwilander:
I'm not familiar with the specific changes in this PR but if there's a reasonable chance that it could affect performance we should run a load test. As Agniva said, it's really simple to do just follow the steps here https://github.com/mattermost/mattermost-load-test-ng/blob/master/docs/terraform_loadtest.md and if you need any help the team will gladly support :)

PR #20290, opened 2022-05-25
[MM-42192] Include deleted posts in GetPost

santoniriccardo:
#### Summary

Introduced include_deleted query param on GetPost endpoint, allowing admins to retrieve contents of posts whether they are deleted or not.

https://api.mattermost.com/#operation/GetPost. Only system admin users have the permission to access deleted post through the ```include_deleted=true``` query param , other users get a permission error.

#### Ticket Link

Fixes https://github.com/mattermost/mattermost-server/issues/19867
JIRA: https://mattermost.atlassian.net/browse/MM-42192

#### Release Note

```release-note
Introduced include_deleted query param to GetPost endpoint.
```


model/client4.go
```diff
@@ -3780,6 +3780,24 @@ func (c *Client4) GetPost(postId string, etag string) (*Post, *Response, error)
...
+	var post Post
+	if r.StatusCode == http.StatusNotModified {
+		return &post, BuildResponse(r), nil
+	}
+	if jsonErr := json.NewDecoder(r.Body).Decode(&post); jsonErr != nil {
+		return nil, nil, NewAppError("GetPostIncludeDeleted", "api.unmarshal_error", nil, jsonErr.Error(), http.StatusInternalServerError)
+	}
+	return &post, BuildResponse(r), nil
+}
+
```
santoniriccardo, line 3800: Note, decided to implement completely seperate client function for the specific case for including deleted as this was the pattern that I noticed throughout the file ([ex1](https://github.com/mattermost/mattermost-server/blob/54a88fb53268874aad436c77c72586ac24eb3e70/model/client4.go#L1919-L1950), [ex2](https://github.com/mattermost/mattermost-server/blob/54a88fb53268874aad436c77c72586ac24eb3e70/model/client4.go#L3313-L3343), etc...), as well as the fact that changing the current function would deprecate current users of the library.

That being said, not sure that this is the best pattern for scalability down the line, especially if more options are anticipated.

santoniriccardo:
@amyblais Similarly tagging you here to help find reviewers as you assigned.

model/client4.go
koox00: It is fine as we don't want to have breaking changes :)

koox00 (approved):
Thank you @santoniriccardo 💯 !!

model/client4.go
```diff
@@ -3780,6 +3780,24 @@ func (c *Client4) GetPost(postId string, etag string) (*Post, *Response, error)
 	return &post, BuildResponse(r), nil
 }
 
+// GetPost gets a single post, including deleted.
```
vish9812, line 3783:
```suggestion
// GetPostIncludeDeleted gets a single post, including deleted.
```

vish9812 (changes requested):
LGTM 👍 
Just update the correct func name in the comment.

vish9812 (approved):

vish9812:
@santoniriccardo Good work on this one.

However you should also raise another PR to update the API docs in the repository - https://github.com/mattermost/mattermost-api-reference.

For reference, 
* you can look at the `bots.yaml` code - https://github.com/mattermost/mattermost-api-reference/blob/2e0d09892edaf7ce91f8196d1c02c55cc1b028d2/v4/source/bots.yaml#L250-L254
* also the doc itself - https://developers.mattermost.com/contribute/server/rest-api/#documenting-the-endpoint

vish9812:
There are a couple of PRs, in enterprise and api-reference, need to be raised before merging this one.

vish9812:
Great! I'll have a look at the Docs PR. As for the enterprise, there was just a small change, I've already updated that.

Will update once both PRs are ready.

vish9812:
Created a clone of this PR here - https://github.com/mattermost/mattermost-server/pull/20358.
So, closing this one.

PR #20309, opened 2022-05-27
MM-44638: Remove upgrade.go

agnivade:
We remove any remaining migrations from the upgrade.go file

```release-note
NONE
```


store/sqlstore/upgrade.go
```diff
@@ -1,815 +0,0 @@
...
-
-func upgradeDatabaseToVersion35(sqlStore *SqlStore) {
-	if shouldPerformUpgrade(sqlStore, Version340, Version350) {
-		sqlStore.GetMasterX().Exec("UPDATE TeamMembers SET Roles = 'team_user' WHERE Roles = ''")
-		sqlStore.GetMasterX().Exec("UPDATE TeamMembers SET Roles = 'team_user team_admin' WHERE Roles = 'admin'")
-		sqlStore.GetMasterX().Exec("UPDATE ChannelMembers SET Roles = 'channel_user' WHERE Roles = ''")
-		sqlStore.GetMasterX().Exec("UPDATE ChannelMembers SET Roles = 'channel_user channel_admin' WHERE Roles = 'admin'")
-
-		// The rest of the migration from Filenames -> FileIds is done lazily in api.GetFileInfosForPost
-		sqlStore.CreateColumnIfNotExists("Posts", "FileIds", "varchar(150)", "varchar(150)", "[]")
```
agnivade, line 335: This is not needed because `FileIds` is already varchar(300) and there is a migration for that.

store/sqlstore/upgrade.go
```diff
@@ -1,815 +0,0 @@
...
-
-		sqlStore.Session().RemoveAllSessions()
```
agnivade, line 337: Skipping this one. Don't feel it's necessary now.

store/sqlstore/upgrade.go
```diff
@@ -1,815 +0,0 @@
...
-		saveSchemaVersion(sqlStore, Version340)
-	}
-}
-
-func upgradeDatabaseToVersion35(sqlStore *SqlStore) {
-	if shouldPerformUpgrade(sqlStore, Version340, Version350) {
-		sqlStore.GetMasterX().Exec("UPDATE TeamMembers SET Roles = 'team_user' WHERE Roles = ''")
-		sqlStore.GetMasterX().Exec("UPDATE TeamMembers SET Roles = 'team_user team_admin' WHERE Roles = 'admin'")
-		sqlStore.GetMasterX().Exec("UPDATE ChannelMembers SET Roles = 'channel_user' WHERE Roles = ''")
-		sqlStore.GetMasterX().Exec("UPDATE ChannelMembers SET Roles = 'channel_user channel_admin' WHERE Roles = 'admin'")
```
agnivade, line 332: Pinged @mkraft's to clarify.

store/sqlstore/upgrade.go
```diff
@@ -1,815 +0,0 @@
...
-		if err := sqlStore.GetReplicaX().Select(&teams, "SELECT * FROM Teams WHERE InviteId = ''"); err != nil {
-			mlog.Error("Error fetching Teams without InviteID", mlog.Err(err))
-		} else {
-			for _, team := range teams {
-				team.InviteId = model.NewId()
-				if _, err := sqlStore.Team().Update(team); err != nil {
-					mlog.Error("Error updating Team InviteIDs", mlog.String("team_id", team.Id), mlog.Err(err))
-				}
-			}
-		}
```
agnivade, line 562: Moved to an app migration

store/sqlstore/upgrade.go
```diff
@@ -1,815 +0,0 @@
...
-}
-
-func upgradeDatabaseToVersion5281(sqlStore *SqlStore) {
-	if shouldPerformUpgrade(sqlStore, Version5280, Version5281) {
-		saveSchemaVersion(sqlStore, Version5281)
-	}
-}
-
-func precheckMigrationToVersion528(sqlStore *SqlStore) error {
-	teamsQuery, _, err := sqlStore.getQueryBuilder().Select(`COALESCE(SUM(CASE
```
agnivade, line 685: Ignored this. Wasn't doing anything

store/sqlstore/upgrade.go
```diff
@@ -1,815 +0,0 @@
...
-	}
-}
-
-func upgradeDatabaseToVersion531(sqlStore *SqlStore) {
-	if shouldPerformUpgrade(sqlStore, Version5300, Version5310) {
-		saveSchemaVersion(sqlStore, Version5310)
-	}
-}
-
-const RemoteClusterSiteURLUniqueIndex = "remote_clusters_site_url_unique"
```
agnivade, line 737: Moved to store.go file.

agnivade:
@lieut-data - Adding you for review because you probably were involved in some of these earlier migrations. Please check once if everything looks okay. I think a lot of it may not be applicable today so I have skipped some.

@isacikgoz - I have also deliberately not run any backward migrations because all of these are destructive things. So doesn't make much sense.

isacikgoz (approved):
Nice work!

app/migrations.go
```diff
@@ -479,6 +480,33 @@ func (s *Server) doFirstAdminSetupCompleteMigration() {
 	}
 }
 
+func (s *Server) doRemainingSchemaMigrations() {
+	// If the migration is already marked as completed, don't do it again.
+	if _, err := s.Store.System().GetByName(remainingSchemaMigrationsKey); err == nil {
+		return
+	}
+
+	if teams, err := s.Store.Team().GetByEmptyInviteID(); err != nil {
```
lieut-data, line 489: Since there's only "one remaining", should we name this function appropriately? Or is the idea to extend it in the future?

Also, it's not clear to me why this is at the application layer, vs. a migration script -- could we run the equivalent of `model.NewId()` in SQL?

db/migrations/mysql/000088_remaining_migrations.down.sql
```diff
@@ -0,0 +1 @@
+-- Skipping it
```
lieut-data, line 1: Might we expand in file on why we chose to skip? I sense a support engineer arriving here and needing additional context in the future.

lieut-data (approved):
Nice to see this wrapped up! Just two, non-blocking comments below.

app/migrations.go
agnivade: > could we run the equivalent of model.NewId() in SQL?

I don't think it's impossible, but I reckoned that it might be simpler to do it in the app layer.

PR #20326, opened 2022-05-31
websocket: drop less important events earlier

trilopin:
#### Summary
Discard the non-essential WebSocket events **before** they are queued.

Some events seem more essential to improve the user experience than to keep the platform running.  It's the case of  `WebsocketEventTyping`, `WebsocketEventStatusChange` and `WebsocketEventChannelViewed`.  Additionally, those events seem to be very frequent.

There is already a mechanism in place to drop those messages when the buffered send channel is at least at 50% capacity. Messages are dropped in web_conn once they have been already queued to be sent. 

The proposal consists of dropping those messages before they are queued, preventing the `wc.send` to be filled with messages that will be dropped anyway.

**LEN(webconn.send) < 50%**
webhub -> webconn.send -> websocket.writemessage 

**LEN(webconn.send) >= 50%**
BEFORE: webhub -> webconn.send -> DROPPED
PROPOSED: webhub -> DROPPED

#### Ticket Link
I came to this because of some WebSocket-related stuff, but not 100% related to the fix :)

```release-note
NONE
```


app/web_conn.go
```diff
@@ -737,6 +716,23 @@ func (wc *WebConn) shouldSendEvent(msg *model.WebSocketEvent) bool {
 		}
 	}
 
+	// When the pump starts to get slow we'll drop non-critical
```
agnivade, line 719: Let's make this as the second condition to avoid unnecessary computations.

agnivade (commented):
Looks like a decent improvement!

I'd be curious to see some load test results for this. Our load test repo is here: https://github.com/mattermost/mattermost-load-test-ng. The process to run a load test via terraform is mentioned here: https://github.com/mattermost/mattermost-load-test-ng/blob/master/docs/terraform_loadtest.md.

Give it a run and let me know how it goes.

trilopin:
avg-api-times
avg-store-times
cpu-utilization
goroutines-in-use
heap-in-use
p99-api-times
p99-store-times
rps
stack-in-use

agnivade (approved):
Nicely done and great work on the load tests!

streamer45 (approved):
Nice work, there's a decent CPU improvement as well :+1: 

cpu

trilopin:
🎉

PR #20338, opened 2022-06-01
[MM-43649] - Ability for end users to notify admin to upgrade workspace

AGMETEOR:
#### Summary
Add an endpoint that allows user to notify admins to upgrade plan. 
Test endpoint
```curl
curl --location --request POST 'localhost:8065/api/v4/cloud/notify-admin-to-upgrade' \
--header 'Authorization: Bearer < MMAUTHTOKEN >' \
--header 'Content-Type: application/json' \
--data-raw '{
    "current_team_id": "<user_team_id>"
}'
```
#### Ticket Link
[MM-43649](https://mattermost.atlassian.net/browse/MM-43649)

#### Related Pull Requests

- Has webapp changes (https://github.com/mattermost/mattermost-webapp/pull/10437)

#### Release Note
```release-note
NONE
```


api4/cloud.go
```diff
@@ -49,6 +49,25 @@ func (api *API) InitCloud() {
...
+
+func handleNotifyAdminToUpgrade(c *Context, w http.ResponseWriter, r *http.Request) {
+	var notifyAdminRequest *model.NotifyAdminToUpgradeRequest
+	err := json.NewDecoder(r.Body).Decode(&notifyAdminRequest)
+	if err != nil {
+		c.SetInvalidParam("notifyAdminRequest")
+		return
+	}
+
+	appErr := c.App.NotifySystemAdminsToUpgrade(c.AppContext, notifyAdminRequest.CurrentUserId, notifyAdminRequest.CurrentTeamId)
```
harshilsharma63, line 64: Should it not be picking the user ID from auth-token? In the current implementation someone can call the API providing in another user's ID in `notifyAdminRequest.CurrentUserId`.

app/cloud.go
```diff
@@ -4,14 +4,106 @@
...
+		}
+
+		post := &model.Post{
+			Message:   T("api.cloud.upgrade_plan_bot_message", map[string]interface{}{"TeamName": team.Name}),
+			UserId:    systemBot.UserId,
+			ChannelId: channel.Id,
+			Type:      fmt.Sprintf("%sup_notification", model.PostCustomTypePrefix), // webapp will have to create renderer for this custom post type
+		}
+
+		_, appErr = a.CreatePost(c, post, channel, false, true)
```
harshilsharma63, line 82: I think we should log the error and continue processing for other users. One user's error shouldn't stop the message from being sent to other users.

app/cloud.go
```diff
@@ -4,14 +4,106 @@
...
+	if sysVal != nil {
+		val := sysVal.Value
+
+		err = json.Unmarshal([]byte(val), alprnu)
+		if err != nil {
+			mlog.Error("Unable to Unmarshal", mlog.Err(err))
+		}
+
+		if alprnu.ContainsID(user.Id) {
+			return model.NewAppError("app.SendCloudUpgradeConfirmationEmail", "api.cloud.notify_admin_to_upgrade_error.already_notified", nil, "", http.StatusForbidden)
```
harshilsharma63, line 43: Can you please share if we have any requirements on this behaviour?

harshilsharma63 (changes requested):

harshilsharma63:
@AGMETEOR added a few comments but this overall this looks great! This makes things much easier!.

app/cloud.go
AGMETEOR: Agreed

app/cloud.go
AGMETEOR: Somewhere on the ticket for this PR it mentions `current thought is to just allow the admin to receive it once only`. @marianunez thoughts on this?

harshilsharma63 (approved):
LGTM 👍

harshilsharma63:
Looks good now @AGMETEOR ! Great work!

neallred (changes requested):
I think this implements a more limited, earlier version of the requirements that a user can notify the admin only once, but doesn't yet include only notifying at most every 30 days.

I think to minimize code churn and changes to how and when admins get notified, it makes sense to get the rest of the v2 requirements in https://docs.google.com/document/d/1_22mA3eATujDQl0ZrKeinQBt5eY6uhOGvgYSK48JQmQ/edit# and then implmement that in this PR.

marianunez (changes requested):

api4/cloud.go
```diff
@@ -49,6 +50,29 @@ func (api *API) InitCloud() {
 
 	// POST /api/v4/cloud/webhook
 	api.BaseRoutes.Cloud.Handle("/webhook", api.CloudAPIKeyRequired(handleCWSWebhook)).Methods("POST")
+
+	api.BaseRoutes.Cloud.Handle("/notify-admin-to-upgrade", api.APISessionRequired(handleNotifyAdminToUpgrade)).Methods("POST")
+}
+
+var notifyLock sync.Mutex
```
neallred, line 57: 3/5 I'd recommend placing the lock in the app layer. My thinking:
* A lock is a bottleneck; it should be as short as possible. Currently, it would lock and block other requests even if it is a bad request that would never be processed and never need to read or write the notify admin system data.
* No other code in this project locks in the api layer; they are placed at the lower layers in to limit the bottleneck. Other projects I've worked on are similar.
* Keeping the lock limited to the piece of code that can cause races makes it much clearer why the lock exists.

model/cloud.go
```diff
@@ -218,3 +225,55 @@ type ProductLimits struct {
...
+
+type AlreadyCloudNotifiedAdminUsersInfo struct {
+	Info []UserInfo
+}
+
+func (a *AlreadyCloudNotifiedAdminUsersInfo) CanNotify(ID string) bool {
+	coolOffPeriodDaysEnv := os.Getenv("MM_CLOUD_NOTIFY_ADMIN_COOL_OFF_DAYS")
+	coolOffPeriodDays, parseError := strconv.ParseFloat(coolOffPeriodDaysEnv, 64)
+	if parseError != nil {
+		fmt.Println("THE RROR", parseError.Error())
```
neallred, line 246: Nit remove debug statement

model/cloud.go
```diff
@@ -218,3 +225,55 @@ type ProductLimits struct {
...
+		coolOffPeriodDays = defaultCloudNotifyAdminCoolOffDays
```
neallred, line 247: I like the default if parsing fails.

model/cloud.go
```diff
@@ -218,3 +225,55 @@ type ProductLimits struct {
...
 	Teams        *TeamsLimits        `json:"teams,omitempty"`
 }
+
+type NotifyAdminToUpgradeRequest struct {
+	CurrentTeamId string `json:"current_team_id"`
+}
+
+type UserInfo struct {
+	UserID    string
+	TimeStamp int64
```
neallred, line 235: Nit: `Timestamp` instead of `TimeStamp`. It is a a single word, and most code in this codebase uses it as a single word.

model/cloud.go
```diff
@@ -218,3 +225,55 @@ type ProductLimits struct {
...
+
+	return true
+}
+
+func (a *AlreadyCloudNotifiedAdminUsersInfo) Upsert(ID string) []UserInfo {
+	for ind, i := range a.Info {
+		if i.UserID == ID {
+			currentUserInfo := a.Info[ind]
+			currentUserInfo.TimeStamp = GetMillis()
+			a.Info[ind] = currentUserInfo
```
neallred, line 268: Can we simplify
```go
			currentUserInfo := a.Info[ind]
			currentUserInfo.TimeStamp = GetMillis()
			a.Info[ind] = currentUserInfo
```
to 
```go
			a.Info[ind].Timestamp = GetMillis()
```

neallred (changes requested):
Functionality looks good to me. Requesting changes for moving the lock to the app layer and renaming `TimeStamp` to `Timestamp`.

i18n/en.json
```diff
@@ -491,6 +495,10 @@
     "id": "api.cloud.teams_limit_reached.restore",
     "translation": "Unable to restore team because teams limit has been reached"
   },
+  {
+    "id": "api.cloud.upgrade_plan_bot_message",
+    "translation": "A member of {{.TeamName}} has notified you to upgrade this workspace before the trial ends."
```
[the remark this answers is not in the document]
neallred: > `...to upgrade this workspace before the trial ends...`

@annelieseklein @justinegeffen This matches the design, but the text doesn't seem to match the workspace's state; the workspace may or may not be in trial at the time the click the button. Can we simplify the wording to something like `A member of {{.TeamName}} has notified you to upgrade this workspace`

app/cloud.go
```diff
@@ -4,14 +4,105 @@
...
 	"time"
 
+	"github.com/mattermost/mattermost-server/v6/app/request"
 	"github.com/mattermost/mattermost-server/v6/model"
+	"github.com/mattermost/mattermost-server/v6/shared/i18n"
 	"github.com/mattermost/mattermost-server/v6/shared/mlog"
 )
 
+func (a *App) NotifySystemAdminsToUpgrade(c *request.Context, currentUserTeamID string) *model.AppError {
+	userId := c.Session().Id
```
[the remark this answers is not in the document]
neallred: This is better being in the app layer! We only want one section of locking & unlocking though. Otherwise, arbitrarily many requests could read at the same time, and then whichever is the last to write will clobber the others, losing track of their info. Here's an example of what I mean. If you run this with `go run temp.go`, you will sometimes get different amounts of elements in different runs.

temp.go
```go
package main

import (
	"fmt"
	"sync"
	"time"
)

func main() {
	mu := sync.Mutex{}
	wg := sync.WaitGroup{}
	wg.Add(2)

	sharedData := []int{}

	go func() {
		for i := 0; i < 1000; i += 2 {
			// read
			mu.Lock()
			val := sharedData
			mu.Unlock()

			// do other works/checks
			time.Sleep(time.Nanosecond)
			newVal := append(val, i)
			fmt.Printf("goroutine 2 %v\n", val)

			// write
			mu.Lock()
			sharedData = newVal
			mu.Unlock()
		}
		wg.Done()
	}()

	go func() {
		for i := 1; i < 1000; i += 2 {
			// read
			mu.Lock()
			val := sharedData
			mu.Unlock()

			// do other works/checks
			time.Sleep(time.Nanosecond)
			newVal := append(val, i)
			fmt.Printf("goroutine 2 %v\n", val)

			// write
			mu.Lock()
			sharedData = newVal
			mu.Unlock()
		}
		wg.Done()
	}()
	wg.Wait()
	fmt.Printf("final result %v, %d elements\n", sharedData, len(sharedData))
}
```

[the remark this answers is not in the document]
neallred: Made a PR that changes it to one critical section: https://github.com/mattermost/mattermost-server/pull/20488/files , and makes sending the notifications async since we don't care about the result of that.

neallred (approved):
LGTM, thanks @AGMETEOR !

app/cloud.go
```diff
@@ -4,14 +4,123 @@
...
+			post := &model.Post{
+				Message:   T("api.cloud.upgrade_plan_bot_message", map[string]interface{}{"TeamName": team.Name}),
+				UserId:    systemBot.UserId,
+				ChannelId: channel.Id,
+				Type:      fmt.Sprintf("%sup_notification", model.PostCustomTypePrefix), // webapp will have to create renderer for this custom post type
+			}
+
+			_, appErr = a.CreatePost(c, post, channel, false, true)
+			if appErr != nil {
+				mlog.Error("Error creating post", mlog.Err(appErr))
```
marianunez, line 100: nit: add more context on what post
```suggestion
				mlog.Error("Error creating post for admin upgrade notification", mlog.Err(appErr))
```

i18n/en.json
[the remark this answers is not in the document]
marianunez: @AGMETEOR heads up that this  text update is pending

model/cloud.go
```diff
@@ -218,3 +224,51 @@ type ProductLimits struct {
...
+type UserInfo struct {
+	UserID    string
+	Timestamp int64
+}
+
+type AlreadyCloudNotifiedAdminUsersInfo struct {
+	Info []UserInfo
+}
+
+func (a *AlreadyCloudNotifiedAdminUsersInfo) CanNotify(ID string) bool {
```
marianunez, line 241: Can we add some tests for this new `CanNotify` and `Upsert` below?

api4/cloud_test.go
```diff
@@ -282,6 +284,56 @@ func Test_requestTrial(t *testing.T) {
 	})
 }
 
+func TestNotifyAdminToUpgrade(t *testing.T) {
+	t.Run("user can only notify admin once in cool off period", func(t *testing.T) {
```
marianunez, line 288: Any possibility to add a test that it will notify again after the cool off period? We can probably set a shorter time overriding the env var

marianunez (changes requested):
Thanks @AGMETEOR, LGTM - just requesting some additional tests

stevemudie (approved):
Will test post merge as part of https://github.com/mattermost/mattermost-webapp/pull/10437

app/cloud.go
```diff
@@ -4,14 +4,123 @@
...
+	// mark as done for current user until end of cool off period
+	alreadyNotifiedUsersInfo.Upsert(userId)
+
+	out, err := json.Marshal(alreadyNotifiedUsersInfo)
+	if err != nil {
+		mlog.Error("Unable to Unmarshal", mlog.Err(err))
+	}
+
+	sysVar := &model.System{Name: "NOTIFIED_ADMIN_TO_UPGRADE", Value: string(out)}
+	if err := a.Srv().Store.System().SaveOrUpdate(sysVar); err != nil {
```
AGMETEOR, line 115: @neallred I just looked at this table's migration
```sql
CREATE TABLE IF NOT EXISTS systems (
    name VARCHAR(64),
    value VARCHAR(1024),
    PRIMARY KEY (name)
);
and I don't think this solution will work
```

AGMETEOR: I just realized the preferences table might solve this well and we could avoid introducing a new migration on huge user tables

api4/cloud_test.go
AGMETEOR: Added a test

marianunez (approved):

app/cloud.go
```diff
@@ -4,14 +4,103 @@
...
+	}
+
+	if notificationPref != nil {
+		info := &model.AdminNotificationUserInfo{}
+		err = json.Unmarshal([]byte(notificationPref.Value), info)
+		if err != nil {
+			mlog.Warn("Unable to Unmarshal", mlog.Err(err))
+		}
+
+		if !model.CanNotify(info.LastNotificationTimestamp) {
```
neallred, line 34: Something is still ambiguous from the requirements and want to make sure we settle it before shipping it. Looks like this makes it so individual users can notify no more frequently than every 30 days. But if several users were to click notify admin within a 30 day user, the admin would get pinged, once per user. Is that acceptable? My guess from reading the ticket was we don't want that, but on closer reading I'm not sure its clear. @johndavidlugtu 

If we do want that, this looks good to me without further changes. If we don't want that, we'll need to read all the CloudNotifyAdminInfo entries in the preferences table to check for the latest timestamp (I don't think that's hard or too performance heavy).

AGMETEOR: That’s true. In the ticket it’s not stated whether only one user can ping per 30 days. @johndavidlugtu can verify

annelieseklein: No, we want to edit the system bot message to update the number of people that have requested. we should not ping the admin per user request @neallred

AGMETEOR: @neallred changed the logic. Only one user can notify within cool off period. Asking for a rereview

app/cloud.go
```diff
@@ -4,14 +4,107 @@
...
+	"github.com/mattermost/mattermost-server/v6/app/request"
 	"github.com/mattermost/mattermost-server/v6/model"
+	"github.com/mattermost/mattermost-server/v6/shared/i18n"
 	"github.com/mattermost/mattermost-server/v6/shared/mlog"
 )
 
+func (a *App) NotifySystemAdminsToUpgrade(c *request.Context, currentUserTeamID string) *model.AppError {
+	userId := c.Session().Id
+
+	prefId := strings.ReplaceAll(model.CloudNotifyAdminInfo, "_", "") + "123456"
```
neallred, line 22: Nit: Could we rename the fake user id suffix `123456` to something more descriptive like `fakeid`?

neallred (approved):

PR #20346, opened 2022-06-02
remove is paid tier

neallred:
#### Summary
mattermost-webapp [no longer uses](https://github.com/mattermost/mattermost-webapp/pull/10431) `IsPaidTier`, and the mattermost-server/enterprise layer doesn't use it either. Removing.

#### Ticket Link
side effect of implementing https://mattermost.atlassian.net/browse/MM-44440

#### Release Note

```release-note
Remove `IsPaidTier` from subscription struct.
```

### DIFF
--- a/model/cloud.go
+++ b/model/cloud.go
@@ -125,7 +125,6 @@ type Subscription struct {
 	Seats       int      `json:"seats"`
 	Status      string   `json:"status"`
 	DNS         string   `json:"dns"`
-	IsPaidTier  string   `json:"is_paid_tier"`
 	LastInvoice *Invoice `json:"last_invoice"`
 	IsFreeTrial string   `json:"is_free_trial"`
 	TrialEndAt  int64    `json:"trial_end_at"`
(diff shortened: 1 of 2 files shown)


nickmisasi (approved):

model/cloud.go
```diff
@@ -125,7 +125,6 @@ type Subscription struct {
 	Seats       int      `json:"seats"`
 	Status      string   `json:"status"`
 	DNS         string   `json:"dns"`
-	IsPaidTier  string   `json:"is_paid_tier"`
```
marianunez, line 128: We do want to keep passing this based on the discussion today given there are some checks we still need to keep until Nov. 1st. Correct?

marianunez (changes requested):

model/cloud.go
neallred: Correct. I can either leave this open until then, or close the three related PRs and re-open once we reach that date.

PR #20351, opened 2022-06-02
[MM-44329] Add a method to store.Store interface to get sql.DB object

noxer:
#### Summary
This PR adds two methods to the store.Store interface. `GetInternalMasterDB()` returns the `*sql.DB` for the master database. `GetInternalReplicaDBs()` returns a slice of `*sql.DB`s for the configured replica databases (or the master, when no replicas are configured). This behavior is implemented to emulate the functionality of `GetMasterX()` and `GetReplicaX()`.
This PR also updates the mocks to include the new methods.

As an internal change, this adds a method `"sqlStore".Store.hasLicense()` which returns `true` iff the license is not `nil`. Putting it into it's own method allows for simpler (and more efficient) code.

#### Ticket Link
[MM-44329](https://mattermost.atlassian.net/browse/MM-44329)

```release-note
NONE
```

### DIFF
--- a/store/store.go
+++ b/store/store.go
@@ -67,7 +67,11 @@ type Store interface {
 	GetDBSchemaVersion() (int, error)
 	GetAppliedMigrations() ([]model.AppliedMigration, error)
 	GetDbVersion(numerical bool) (string, error)
+	// GetInternalMasterDB allows access to the raw master DB
+	// handle for the multi-product architecture.
 	GetInternalMasterDB() *sql.DB
+	// GetInternalReplicaDBs allows access to the raw replica DB
+	// handles for the multi-product architecture.
 	GetInternalReplicaDBs() []*sql.DB
 	TotalMasterDbConnections() int
 	TotalReadDbConnections() int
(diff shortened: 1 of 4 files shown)


store/store.go
```diff
@@ -66,6 +67,8 @@ type Store interface {
 	GetDBSchemaVersion() (int, error)
 	GetAppliedMigrations() ([]model.AppliedMigration, error)
 	GetDbVersion(numerical bool) (string, error)
+	GetInternalMasterDB() *sql.DB
+	GetInternalReplicaDBs() []*sql.DB
```
agnivade, line 71: Maybe worth adding a comment here to explain that these are added for other products in the multi-product architecture to get raw access to the DB. Just for someone wondering why these are exposed.

agnivade (approved):
Nicely done!

store/store.go
noxer: Yeah, that makes sense.

noxer: I added comments to both methods in the interface.

isacikgoz (approved):
LGTM, well done!

PR #20352, opened 2022-06-02
MM-44509 - subscription change not reflected after trial

pvev:
#### Summary
This PR invalidates all the server caches after trial request.

#### Related PRs
Enterprise: https://github.com/mattermost/enterprise/pull/1229

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-44509

#### Release Note
```release-note
NONE
```


pvev:
@neallred @nickmisasi guys, could you please review this PR including the enterprise one and help me identify what is missing to fully invalidate the cache. Even with this changes, it is still intermittent and one of the requests to getSubscription keeps saying it is not in trial. I have open this PR too (https://github.com/mattermost/mattermost-webapp/pull/10457) where I have spin up the spinwick server to validate these changes, that as I mentioned, keep being intermittent. Thanks!

nickmisasi:
@pvev Happy to look into this a bit more but won't be able to until Monday. 

Smells like a race condition - maybe somewhere in the webapp is requesting the subscription before the trial request has actually completed (it looks to take around 5-8 seconds, lot's can happen in that time). Stripe sends webhooks to CWS which triggers cache invalidations in the installation - I would expect those to be enough for this whole flow.

The CWS will log when it pulls a subscription from cache vs from Stripe. I wasn't able to repro in the spinwick here, but if you can get it to happen tomorrow let me know and I can pull the logs for that time frame for you. 

You can also comment out these 2 lines in your enterprise PR, then start up a spinwick: 
https://github.com/mattermost/enterprise/blob/4d305ca2f40df92c557fbba281ecff62529f03a0/cloud/cloud.go#L408
https://github.com/mattermost/enterprise/blob/4d305ca2f40df92c557fbba281ecff62529f03a0/cloud/cloud.go#L137
This will remove the installation's cache from the equation to help you isolate if it's something on the CWS or installation side

api4/cloud.go
```diff
@@ -195,7 +191,30 @@ func validateBusinessEmail(c *Context, w http.ResponseWriter, r *http.Request) {
...
-	// validate current userAdmin email
+	// if an email was sent as a body param, validate it and return wether is valid or not
+	bodyBytes, err := ioutil.ReadAll(r.Body)
+	if err != nil {
+		c.Err = model.NewAppError("Api4.requestCloudTrial", "api.cloud.app_error", nil, err.Error(), http.StatusInternalServerError)
+		return
+	}
+
+	var emailToValidate *model.ValidateBusinessEmailRequest
+	if err := json.Unmarshal(bodyBytes, &emailToValidate); err != nil {
```
neallred, line 202: The `&` isn't needed here, thought it doesn't break anything.

neallred (approved):

nickmisasi (approved):

api4/cloud.go
```diff
@@ -195,7 +191,30 @@ func validateBusinessEmail(c *Context, w http.ResponseWriter, r *http.Request) {
...
 
-	// validate current userAdmin email
+	// if an email was sent as a body param, validate it and return wether is valid or not
+	bodyBytes, err := ioutil.ReadAll(r.Body)
+	if err != nil {
+		c.Err = model.NewAppError("Api4.requestCloudTrial", "api.cloud.app_error", nil, err.Error(), http.StatusInternalServerError)
+		return
+	}
+
+	var emailToValidate *model.ValidateBusinessEmailRequest
```
marianunez, line 201: nit: 
```suggestion
	var emailToValidateRequest *model.ValidateBusinessEmailRequest
```

api4/cloud.go
```diff
@@ -195,7 +191,30 @@ func validateBusinessEmail(c *Context, w http.ResponseWriter, r *http.Request) {
...
+	if emailToValidate.Email != "" {
+		errValidatingEmail := c.App.Cloud().ValidateBusinessEmail(user.Id, emailToValidate.Email)
+		if errValidatingEmail != nil {
+			c.Err = model.NewAppError("Api4.valiateBusinessEmail", "api.cloud.request_error", nil, errValidatingEmail.Error(), http.StatusInternalServerError)
+			return
+		}
+		ReturnStatusOK(w)
+		return
+	}
+
```
marianunez, line 216: If the email is not valid, we still need to fall back on the workspace owner's email check below. If seems here it just returns an error without the check that was done previously

Also, I'm wondering if we should:

- Invert (or extract) the logic of first checking if the workspace owner has a valid business email. If they do, then we don't need to check anyone else and continue the trial request. If not, then we move to check the email of the current admin. 
- If we do extract the logic, we should keep this api call as just a dumb passthrough of validate any business email given 
- If we do not extract the logic from here, I propose we change the name of this api call to something that indicates we are doing more than just validating the given email - maybe `validateWorkspaceBusinessEmail` or `validateWorkspaceBusinessContact` 0/5

marianunez (changes requested):

api4/cloud.go
pvev: the workflow is the following:
- The UI sends a first request to validate user admin email and customer creation used email: (no email is sent with the first api request), the user admin email is validated, in case of being an invalid email, the customer account used email is validated too. If both are invalid, the UI receives this response and it keeps that info in memory, so in case the user  wants to start the trial, the request email modal is prompted requesting for a valid business email, then the subsequent request to this api includes the entered email value, which will be validated too. Also, the changes included with this PR from the front-end add live validation of the business email providing inmediate feedback to the user.

marianunez: Do we have a ticket for this followup work? @pvev

pvev: yes, this one https://mattermost.atlassian.net/browse/MM-45174

marianunez (approved):
LGTM with the follow up ticket as discussed in the review.

PR #20353, opened 2022-06-02
GraphQL: Create dedicated top-level filter for sidebar categories

agnivade:
We add the excludeTeam switch to allow user to get
all sidebar categories from other teams excluding a given team.

```release-note
NONE
```


store/sqlstore/channel_store_categories.go
```diff
@@ -42,82 +43,133 @@ func (s SqlChannelStore) CreateInitialSidebarCategories(userId, teamId string) (
 	return oc, nil
 }
 
-func (s SqlChannelStore) createInitialSidebarCategoriesT(transaction *sqlxTxWrapper, userId, teamId string) error {
-	selectQuery, selectParams, _ := s.getQueryBuilder().
-		Select("Type").
+func (s SqlChannelStore) createInitialSidebarCategoriesT(transaction *sqlxTxWrapper, userId string, opts *store.SidebarCategorySearchOpts) error {
```
agnivade, line 46: @hmhealey - Please review this carefully. The major change here is that now this method works on multiple teams. The optimization that I have done is that instead of doing `transaction.NamedExec` for each team, I am batching up all the queries and executing them at one shot. Only exception remains `s.migrateFavoritesToSidebarT`. But since it's a one-time migration cost per user, I haven't gone that far. We can optimize that as well once we go further into GraphQL.

agnivade:
ping @hmhealey

store/sqlstore/channel_store_categories.go
hmhealey: Is it right to be handling the possibility of multiple teams in this function? I would've maybe thought we'd have `CreateInitialSidebarCategories` handle calling this for each team separately instead of making this function more complicated. I feel like something like

```go
t := startTransaction()

teams := getTeams(excludeTeam)
for team := range teams {
    createInitialSidebarCategoriesT(t, userId, team.Id)
}
```
would be easier and shouldn't require as many changes here. It would mean a few extra queries, but I think that'd be fine given that this doesn't happen too often

agnivade: The few extra queries is what I am concerned about. Some customers have thousands of teams, which would mean a single request will issue a lot of DB queries. It's a major change with GraphQL, and I'd like to keep the DB load as same as possible.

store/sqlstore/channel_store_categories.go
```diff
@@ -42,82 +43,131 @@ func (s SqlChannelStore) CreateInitialSidebarCategories(userId, teamId string) (
...
-			DisplayName: "Favorites", // This will be retranslated by the client into the user's locale
-			Id:          favoritesCategoryId,
-			UserId:      userId,
-			TeamId:      teamId,
-			Sorting:     model.SidebarCategorySortDefault,
-			SortOrder:   model.DefaultSidebarSortOrderFavorites,
-			Type:        model.SidebarCategoryFavorites,
-		}); err != nil {
-			return errors.Wrap(err, "createInitialSidebarCategoriesT: failed to insert favorites category")
+	getRequiredTeamIDs := func(category model.SidebarCategoryType, opts *store.SidebarCategorySearchOpts) []string {
```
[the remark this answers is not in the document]
agnivade: That makes sense. I can do that.

agnivade:
Looks much better now. PTAL.

noxer (approved):

store/sqlstore/channel_store_categories.go
hmhealey: Maybe the `getTeams` could be specialized for this case and just look for the user's teams that don't have any categories? I can't remember why I made it handle a user having only one or two default categories to begin with.
Then we'd only be making a single query most of the time and finding that the users have all of their initial categories created. It could be something like
```sql
select
    TeamMembers.TeamId
from
    TeamMembers
    left join SidebarCategories on
        TeamMembers.UserId = SidebarCategories.UserId
        and TeamMembers.TeamId = SidebarCategories.TeamId
where
    TeamMembers.UserId = :UserId
    and TeamMembers.DeleteAt = 0
    and SidebarCategories.Type is null;
```

hmhealey:
I replied to the one conversation above in case that gets missed because GitHub doesn't always make that clear

store/sqlstore/channel_store_categories.go
agnivade: @hmhealey

store/sqlstore/channel_store_categories.go
```diff
@@ -468,28 +522,35 @@ func (s SqlChannelStore) GetSidebarCategory(categoryId string) (*model.SidebarCa
...
 		Order:      make([]string, 0),
 	}
 
 	categories := []*sidebarCategoryForJoin{}
-	query, args, err := s.getQueryBuilder().
+	query := s.getQueryBuilder().
 		Select("SidebarCategories.*", "SidebarChannels.ChannelId").
 		From("SidebarCategories").
 		LeftJoin("SidebarChannels ON SidebarChannels.CategoryId=Id").
 		Where(sq.And{
```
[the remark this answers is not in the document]
agnivade: Going to leave this one as-is for now Tim. Hope that's okay :)

PR #20362, opened 2022-06-03
[MM-44488] Cloud limits: enforcing messages

vish9812:
#### Summary

Do not show posts, past the set cloud plan's limit.

#### Ticket Link

JIRA - https://mattermost.atlassian.net/browse/MM-44488

#### Release Note

```release-note
* Updated the Posts search and get APIs to filter out posts, past the set cloud plan's limit.
* Added a new response-header `Has-Inaccessible-Posts` for `getPost` and `getPostByIDs` APIs.
```


jobs/last_accessible_post/worker.go
```diff
@@ -0,0 +1,35 @@
...
+}
+
+func MakeWorker(jobServer *jobs.JobServer, app AppIface) model.Worker {
+	isEnabled := func(cfg *model.Config) bool {
+		return cfg.FeatureFlags != nil && cfg.FeatureFlags.CloudFree
+	}
+	execute := func(job *model.Job) error {
+		_, appErr := app.GetLastAccessiblePostTime(false)
+		if appErr != nil {
+			mlog.Error("Worker: Failed at GetLastAccessiblePostTime", mlog.String("worker", model.JobTypeLastAccessiblePost), mlog.String("job_id", job.Id), mlog.Err(appErr))
```
agnivade, line 27: No need to log here. You are already returning the error which will log it at the higher layer.

store/store.go
```diff
@@ -371,6 +371,8 @@ type PostStore interface {
 	GetOldestEntityCreationTime() (int64, error)
 	HasAutoResponsePostByUserSince(options model.GetPostsSinceOptions, userId string) (bool, error)
 	GetPostsSinceForSync(options model.GetPostsSinceForSyncOptions, cursor model.GetPostsSinceForSyncCursor, limit int) ([]*model.Post, model.GetPostsSinceForSyncCursor, error)
+	// GetRecentNthPostTime returns CreateAt time of most recent nth user-post
```
agnivade, line 374: `// GetNthRecentPostTime returns the CreateAt time of the nth most recent post.` and rename the method name as well.

store/sqlstore/post_store.go
```diff
@@ -1735,6 +1735,39 @@ var specialSearchChar = []string{
 	":",
 }
 
+// GetRecentNthPostTime returns CreateAt time of most recent nth user-post
+func (s *SqlPostStore) GetRecentNthPostTime(n int64) (int64, error) {
```
agnivade, line 1739: This needs to be unit tested as well.

store/sqlstore/post_store.go
```diff
@@ -1735,6 +1735,39 @@ var specialSearchChar = []string{
...
+		OrderBy("p.CreateAt DESC").
+		Limit(1).
+		Offset(uint64(n - 1))
+
+	query, queryArgs, err := builder.ToSql()
+	if err != nil {
+		return 0, errors.Wrap(err, "GetRecentNthPostTime_tosql")
+	}
+
+	query = "SELECT COALESCE((" + query + "), 0)"
```
agnivade, line 1761: I am not sure if this even works. Have you tested this locally? In any case, use COALESCE on the `createAt` column not on the whole query.

store/sqlstore/post_store.go
```diff
@@ -1735,6 +1735,39 @@ var specialSearchChar = []string{
...
+
+	var createAt int64
+	if err := s.GetMasterX().Get(&createAt, query, queryArgs...); err != nil {
+		return 0, errors.Wrapf(err, "failed to get the Nth Post=%v", n)
```
agnivade, line 1765: `n` is an integer, so the format specifier should be `%d`. Additionally, you need to pass the error as well.

app/post.go
```diff
@@ -1353,6 +1490,10 @@ func (a *App) SearchPostsForUser(c *request.Context, terms string, userID string
 		}
 	}
 
+	if appErr := a.filterInaccessiblePosts(postSearchResults.PostList); appErr != nil {
+		return nil, model.NewAppError("SearchPostsForUser", "app.post.filter_inaccessible_posts.app_error", nil, appErr.Error(), http.StatusInternalServerError)
```
agnivade, line 1494: Just return the error directly. No need to unwrap-wrap again. It's already an AppError.

app/post.go
```diff
@@ -1296,10 +1341,102 @@ func (a *App) convertUserNameToUserIds(usernames []string) []string {
...
+
+// filterInaccessiblePosts filters out the posts, past the cloud limit
+func (a *App) filterInaccessiblePosts(postList *model.PostList) *model.AppError {
+	if postList == nil || postList.Posts == nil || len(postList.Posts) == 0 {
+		return nil
+	}
+
+	lastAccessiblePostTime, appErr := a.GetLastAccessiblePostTime(true)
+	if appErr != nil {
+		return model.NewAppError("filterInaccessiblePosts", "app.last_accessible_post.app_error", nil, appErr.Error(), http.StatusInternalServerError)
```
agnivade, line 1407: Same here, return the error directly.

app/post.go
```diff
@@ -809,6 +810,10 @@ func (a *App) GetPostsPage(options model.GetPostsOptions) (*model.PostList, *mod
 		}
 	}
 
+	if appErr := a.filterInaccessiblePosts(postList); appErr != nil {
+		return nil, model.NewAppError("GetPostsPage", "app.post.filter_inaccessible_posts.app_error", nil, appErr.Error(), http.StatusInternalServerError)
```
agnivade, line 814: Return the error directly. No need to unwrap-wrap again. Same applies for other instances in this file.

agnivade (changes requested):
Left some comments. This PR needs some refactoring work and unit tests need to be added. I'd expect a unit test in the store layer, app layer, and in the main api layer which verifies that posts are truncated and `HasInaccessiblePosts` field is set.

store/sqlstore/post_store.go
vish9812: > format specifier should be %d

The %v already has the default format as %d for the int types. Is there any benefit of using %d explicitly?

> pass the error as well

I'm not sure if I follow correctly, but the first parameter is already the `err` field.

store/sqlstore/post_store.go
vish9812: Yes, the purpose of Coalesce here is to return 0 in case there is no row returned from the main query, which would happen when "offset" value is greater than the total no. of posts.
That would mean, all posts are still accessible/within the cloud limit.

i18n/en.json
```diff
@@ -9095,6 +9103,10 @@
     "id": "oauth.gitlab.tos.error",
     "translation": "GitLab's Terms of Service have updated. Please go to gitlab.com to accept them and then try logging into Mattermost again."
   },
+  {
+    "id": "parse.int64",
```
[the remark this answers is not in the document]
vish9812: Actually, I wanted it to be a generic error. As "parsing error" doesn't need to associated with any specific method. 
Maybe we can have a different format for generic errors like `app.common.parse_error`?

app/post.go
```diff
@@ -1296,10 +1341,102 @@ func (a *App) convertUserNameToUserIds(usernames []string) []string {
...
 }
 
+// GetLastAccessiblePostTime returns CreateAt time of last accessible post as per the cloud limit
+func (a *App) GetLastAccessiblePostTime(useCache bool) (int64, *model.AppError) {
+	cfg := a.Config()
+	if cfg.FeatureFlags == nil || !cfg.FeatureFlags.CloudFree {
+		return 0, nil
+	}
+
+	if useCache {
```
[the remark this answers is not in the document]
vish9812: Interestingly, I initially implemented the way you're suggesting🙂.
But then it felt a little off to me, as the app was now depending on a "helper" piece, in this "cache" to work correctly.

I generally try to make sure that app should keep running as long as its "core" features/logic are working. So, in this case even if the "cached" value doesn't work, the app is still able to work(even though slower), while the cache can get fixed.

And for the same reason, I logged the errors instead of returning as the core functionality is still working, so the app should also keep running.
Maybe I can change the Warn to Err, and I'm not sure, how these logs work on Prod, especially the .Err() logs. Does admin get notifications or has an access to some UI to see these error logs?

Also, I think it's a common pattern to fetch from the cache > if not found, fetch from the main source > then update the cache > and return the data.

Let me know your thoughts, I can revert it, if you feel we should go that way.

vish9812:
Thanks @agnivade for the review. It's still in WIP, so I haven't started on the test cases yet. First, I wanted to get the main functionality reviewed as it touches all the *core* funcs, returning the `PostList` model.

app/post.go
vish9812: Done.

store/sqlstore/post_store.go
agnivade: > The %v already has the default format as %d for the int types. Is there any benefit of using %d explicitly?

Typically we use %v when we don't know the type or for errors. For integers, it's more idiomatic to use %d.

> I'm not sure if I follow correctly, but the first parameter is already the err field.

Ah my bad there. :+1:

store/sqlstore/post_store.go
agnivade: Can you post the final query that's being generated? Maybe I missed something.

i18n/en.json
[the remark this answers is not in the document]
agnivade: Sounds good, let's do that.

app/post.go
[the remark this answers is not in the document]
agnivade: > I generally try to make sure that app should keep running as long as its "core" features/logic are working. So, in this case even if the "cached" value doesn't work, the app is still able to work(even though slower), while the cache can get fixed.

Yep, do all of that, but the method names should be separate, and there is no need for the `useCache` flag.

> And for the same reason, I logged the errors instead of returning as the core functionality is still working, so the app should also keep running.

I understand your intention, but in practice this leads to more harm than help. You want to keep the app running, but the app is already failing, and all we are doing is hiding the errors from the user. This doesn't really help and the user gets confused because no error is surfaced.

So here is my proposal:

App method: `GetLastAccessiblePostTime()`. (no extra param needed)
- check cache
- if found return
- if not found, compute and insert in cache. 
 
Job method: `ComputeLastAccessiblePostTime()`:
- compute and insert in cache.

What I was suggesting is that to refactor these small bits into separate small methods so that they can be called from these 2 high level methods. E.g. the "compute and insert in cache" could be its own method.

WDYT?

store/store.go
agnivade: You just changed the comment, but the method needs to be renamed as well :)

app/post.go
```diff
@@ -1297,10 +1342,102 @@ func (a *App) convertUserNameToUserIds(usernames []string) []string {
...
+	if err != nil {
+		mlog.Warn("GetLastAccessiblePostTime: failed to update cache.", mlog.Err(err))
+	}
+
+	return createdAt, nil
+}
+
+// filterInaccessiblePosts filters out the posts, past the cloud limit
+func (a *App) filterInaccessiblePosts(postList *model.PostList) *model.AppError {
+	if postList == nil || postList.Posts == nil || len(postList.Posts) == 0 {
```
neallred, line 1402: The main issue I see with this function is posts can be in time ascending or descending order, and this function needs to support both. In addition, while reviewing it, I realized it traverses the entire post list, but as far as I can tell we are guaranteed that the list is sorted either ascending or descending, so I think we can do a binary-search like operation. Since this is a fairly main path of the application, I think its worth optimizing for it.

I made a branch off yours that implements this search: https://github.com/mattermost/mattermost-server/pull/20439. What do you think?

app/post.go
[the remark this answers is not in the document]
vish9812: SGTM.

store/sqlstore/post_store.go
vish9812:
```sql
SELECT COALESCE((SELECT CreateAt FROM Posts p WHERE (p.Type = ? AND p.UserId NOT IN (SELECT UserId FROM Bots)) ORDER BY p.CreateAt DESC LIMIT 1 OFFSET 99), 0)
```

store/store.go
vish9812: Oops😅

app/post.go
```diff
@@ -1296,10 +1341,102 @@ func (a *App) convertUserNameToUserIds(usernames []string) []string {
...
+
+	limits, err := a.Cloud().GetCloudLimits("")
+	if err != nil {
+		return 0, model.NewAppError("GetLastAccessiblePostTime", "api.cloud.app_error", nil, err.Error(), http.StatusInternalServerError)
+	}
+
+	if limits == nil || limits.Messages == nil || limits.Messages.History == nil {
+		// Cloud limit is not applicable
+		return 0, nil
+	}
```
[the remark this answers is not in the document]
vish9812: That means if `cfg.FeatureFlags.CloudFree` is `false`, then it's sure that there are no limits.
But if it's `true`, then I'll have to double-check it with the `limits.Messages` to confirm whether limits exist or not?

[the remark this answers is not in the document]
neallred: Correct.

app/post.go
vish9812: Thanks for the PR @neallred.

However, I noticed a couple of things:
1. This func `getPostsAround` doesn't seem to return posts ordered by CreatedAt, so we may still need to support the linear search.

```go
query = query.From(table).
		Where(conditions).
		// Adding ChannelId and DeleteAt order columns
		// to let mysql choose the "idx_posts_channel_id_delete_at_create_at" index always.
		// See MM-24170.
		OrderBy("p.ChannelId", "p.DeleteAt", "p.CreateAt "+sort).
		Limit(uint64(options.PerPage)).
		Offset(uint64(offset))
```

2. There are a few other funcs as well like `GetParentsForExportAfter` which don't return ordered posts. But those are used by app/exports.go, although not sure, if these need to be limited as well.

neallred: > 1. ... getPostsAround doesn't ... return posts ordered by CreatedAt, so we may still need to support the linear search.

I missed that one, I think you're right. In that case I'd leave it up to others more familiar with this code whether we want to somehow support both the linear path as well as an optimized path for lists that are sorted by CreatedAt. Could be an extra boolean parameter or `filterOptions` struct parameter, that lets us add more optimizations to the limiting in the future.

> 2. ... other funcs ... like GetParentsForExportAfter ... used by app/exports.go

I don't think those should be limited. [Prior internal discussion](https://community.mattermost.com/private-core/pl/f9npw3nprbrszpe5id1wt9nuno) was that message export should always be unlimited so that we don't lock people into our platform (one of the reasons Mattermost was created in the first place: frustration with vendor lock-in).

neallred: I merged the PR and attempted to sort out everywhere that looked like it was sorted by CreateAt vs not. Went the route of giving `filterInaccessiblePosts` an options struct parameter so that the caller can choose what searching technique is appropriate.

vish9812:
@neallred after this PR-https://github.com/mattermost/mattermost-server/pull/20491, should I use `a.License() == nil || !*a.License().Features.Cloud` now to determine if it's a cloud or not and then use `limits api` to determine if it's `starter`?

neallred:
> *a.License().Features.Cl



> @neallred after this PR-#20491, should I use `a.License() == nil || !*a.License().Features.Cloud` now to determine if it's a cloud or not and then use `limits api` to determine if it's `starter`?

Yep! That is perfect.

app/post_helpers_test.go
```diff
@@ -0,0 +1,295 @@
...
+import (
+	"testing"
+
+	"github.com/mattermost/mattermost-server/v6/model"
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
+)
+
+func Test_getTimeSortedPostAccessibleBounds(t *testing.T) {
+	var p = func(at int64) *model.Post {
```
agnivade, line 15: `p` is a bad choice for a function name. A better name would be postFromCreateAt. Remember that - "further the variable is invoked from the declaration site, the longer it should be": https://github.com/golang/go/wiki/CodeReviewComments#variable-names

app/post_helpers_test.go
```diff
@@ -0,0 +1,295 @@
...
+				"post_e": p(4),
+			},
+			Order: []string{"post_e", "post_d", "post_c", "post_b", "post_a"},
+		}, 2)
+		require.Equal(t, postAccessibleBounds{accessible: 2, inaccessible: 3}, bounds)
+	})
+}
+
+func Test_filterInaccessiblePosts(t *testing.T) {
+	th := Setup(t).InitBasic()
```
agnivade, line 152: Just checking: do you need to call "InitBasic" here? Doesn't look like you are accessing any of BasicUser/Team/Channel etc.

Calling `InitBasic` has additional overhead which adds to the test runtime.

app/post_helpers_test.go
```diff
@@ -0,0 +1,295 @@
...
+
+		assert.Equal(t, []string{
+			"post_e",
+			"post_d",
+		}, postList.Order)
+	})
+}
+
+func Test_isInaccessiblePost(t *testing.T) {
+	th := Setup(t).InitBasic()
```
agnivade, line 277: Same comment as above

app/post_test.go
```diff
@@ -2767,6 +2769,32 @@ func TestGetPostIfAuthorized(t *testing.T) {
 	require.Nil(t, err)
 }
 
+func TestGetPostIfAuthorizedNew(t *testing.T) {
```
agnivade, line 2772: This is a verbatim copy of the previous test. We need to refactor much more here.

jobs/last_accessible_post/worker.go
```diff
@@ -0,0 +1,28 @@
...
+
+type AppIface interface {
+	ComputeLastAccessiblePostTime() *model.AppError
+}
+
+func MakeWorker(jobServer *jobs.JobServer, license *model.License, app AppIface) model.Worker {
+	isEnabled := func(cfg *model.Config) bool {
+		return license != nil && *license.Features.Cloud
+	}
+	execute := func(job *model.Job) error {
```
agnivade, line 23: Same comment as above.

app/post_helpers.go
```diff
@@ -0,0 +1,215 @@
...
+}
+
+// getTimeSortedPostAccessibleBounds returns what the boundaries for accessible and inaccessible posts.
+// It assumes that CreateAt time for posts is monotonically increasing or decreasing.
+// It could be either because posts can be returned in ascending or descending time order.
+// Because it returns the boundaries, it is necessary to check whether the accessible or
+// inaccessible post index is greater, and do the filtering accordingly.
+// Special values (which can be checked with methods `allAccessible` and `allInaccessible`)
+// denote if all or none of the posts are accessible
+func getTimeSortedPostAccessibleBounds(postList *model.PostList, earliestAccessibleTime int64) postAccessibleBounds {
```
agnivade, line 49: This entire implementation seems too complicated to me. I'll revisit this in the second pass.

agnivade (commented):
Gave it a first pass. I also posted a comment in the ~Developers channel.

app/post_helpers.go
```diff
@@ -0,0 +1,215 @@
...
+			for i := min(bounds.inaccessible, otherInaccessibleBound); i <= maxInt(bounds.inaccessible, otherInaccessibleBound); i++ {
+				delete(postList.Posts, order[i])
+			}
+		} else {
+			accessiblePosts := make(map[string]*model.Post, accessibleCount+1)
+			for i := min(bounds.accessible, otherAccessibleBound); i <= maxInt(bounds.accessible, otherAccessibleBound); i++ {
+				accessiblePosts[order[i]] = postList.Posts[order[i]]
+			}
+			postList.Posts = accessiblePosts
+		}
```
agnivade, line 194: I am failing to understand this piece of logic. From what I understand, we have an ordered slice of posts. And we need to exclude posts that are older than `earliestAccessibleTime`. I am not sure why we need to have this logic about `accessible` and `inAccessible` and doing a binary search. Can you clarify?

api4/post.go
```diff
@@ -416,7 +421,58 @@ func getPost(c *Context, w http.ResponseWriter, r *http.Request) {
...
+		return
+	}
+	postWrapper.Post = post
+
+	if c.HandleEtag(post.Etag(), "Get Post", w, r) {
+		return
+	}
+
+	w.Header().Set(model.HeaderEtagServer, post.Etag())
+	post.StripActionIntegrations()
```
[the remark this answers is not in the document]
vish9812: Actually, the old code already does it in this abstraction - `post.EncodeJSON(w)`. But since I'm using `PostWrapper` now, I need to write the same functionality explicitly here.

store/sqlstore/post_store.go
```diff
@@ -1735,6 +1735,41 @@ var specialSearchChar = []string{
...
+		return 0, errors.New("n can't be less than 1")
+	}
+
+	builder := s.getQueryBuilder().
+		Select("CreateAt").
+		From("Posts p").
+		// Consider users posts only for cloud limit
+		Where(sq.And{
+			sq.Eq{"p.Type": ""},
+			sq.Expr("p.UserId NOT IN (SELECT UserId FROM Bots)"),
```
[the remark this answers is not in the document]
vish9812: I remember in some conversation, that it was okay to include the deleted posts against the limit.
Maybe @Willyfrog can confirm it.

jobs/last_accessible_post/worker.go
```diff
@@ -0,0 +1,28 @@
...
+const (
+	JobName = "LastAccessiblePost"
+)
+
+type AppIface interface {
+	ComputeLastAccessiblePostTime() *model.AppError
+}
+
+func MakeWorker(jobServer *jobs.JobServer, license *model.License, app AppIface) model.Worker {
+	isEnabled := func(cfg *model.Config) bool {
```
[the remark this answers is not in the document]
vish9812: It's being used here `worker := jobs.NewSimpleWorker(JobName, jobServer, execute, isEnabled)`

app/post_helpers.go
vish9812: Since the posts are **sorted**, we don't need to find the `earliestAccessibleTime` _linearly_ and can optimize it via _binary search_.

jobs/last_accessible_post/worker.go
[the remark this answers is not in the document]
agnivade: I meant the `cfg` variable.

jobs/last_accessible_post/worker.go
agnivade: I meant the `job` variable

app/post_helpers.go
agnivade: Ok, got that bit. Then I am wondering why do we need a separate struct with `accessible` and `inaccessible` fields for that? Can we just not do a [sort.Search](https://pkg.go.dev/sort#Search) to find the index and then trim everything outside that?

neallred: Using `sort.Search` would work and I'd be down with that as long as we check for ascending and descending order and use the [right comparator](https://pkg.go.dev/sort#example-Search-DescendingOrder) accordingly. When I recommended the binary search, I didn't know `sort.Search` existed.

agnivade: >  I'd be down with that as long as we check for ascending and descending order

Yes of course, that goes without saying.

> When I recommended the binary search, I didn't know sort.Search existed.

No worries, that's what code reviews are for :)

app/post_helpers.go
```diff
@@ -0,0 +1,228 @@
...
+
+	newPostOrder := []string{}
+	for _, postId := range postList.Order {
+		if posts[postId].CreateAt < earliestAccessibleTime {
+			postList.HasInaccessiblePosts = true
+			delete(posts, postId)
+		} else {
+			newPostOrder = append(newPostOrder, postId)
+		}
+	}
```
agnivade, line 90: There is a trick to filter a slice in-place and not allocate a new one: https://github.com/golang/go/wiki/slicetricks#filter-in-place

agnivade (commented):
This looks so much readable and improved! Thank you for your patience @vish9812.

I am 0/5 on the slice trick, feel free to ignore that. But the "X-" is something that needs to be done.

model/client4.go
```diff
@@ -18,29 +18,30 @@ import (
...
+	HeaderToken                = "token"
+	HeaderCsrfToken            = "X-CSRF-Token"
+	HeaderBearer               = "BEARER"
+	HeaderAuth                 = "Authorization"
+	HeaderCloudToken           = "X-Cloud-Token"
+	HeaderRemoteclusterToken   = "X-RemoteCluster-Token"
+	HeaderRemoteclusterId      = "X-RemoteCluster-Id"
+	HeaderRequestedWith        = "X-Requested-With"
+	HeaderRequestedWithXML     = "XMLHttpRequest"
+	HeaderHasInaccessiblePosts = "Has-Inaccessible-Posts"
```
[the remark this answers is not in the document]
agnivade: Oh wow, I wasn't aware that it was deprecated. Thank you for letting me know! Guess that means I am getting old .. :joy:

agnivade (approved):
Looks good. Brilliant work on this Vishal!

vish9812:
@jgilliam17

jgilliam17 (approved):
Thanks @vish9812 
Tested, looks good to merge.
- Verified basic posting functions, posting, searching for the post and reading the post as other user - working as before.
-[ E2E report](https://mattermost-cypress-report.s3.amazonaws.com/214959-777d932-onprem-ent-server-pr-20362-mattermost/mm-ee-test777d932/mochawesome.html) looks good, no PR related faiures.

PR #20364, opened 2022-06-03
[MM-43831] removed the parallel marker from the subtests

noxer:
#### Summary
Removed the skip for the main test and the parallel markers for the sub-tests. I tried running it multiple times and multiple ways and it seems to work now. Running the sub-tests in parallel triggered a fail once but I couldn't reproduce it again.

#### Ticket Link
[MM-43831](https://mattermost.atlassian.net/browse/MM-43831)

#### Release Note
```release-note
NONE
```

### DIFF
--- a/app/post_test.go
+++ b/app/post_test.go
@@ -571,7 +571,6 @@ func TestImageProxy(t *testing.T) {
 }
 
 func TestMaxPostSize(t *testing.T) {
-	t.Skip("TODO: fix flaky test")
 	t.Parallel()
 
 	testCases := []struct {
@@ -599,8 +598,6 @@ func TestMaxPostSize(t *testing.T) {
 	for _, testCase := range testCases {
 		testCase := testCase
 		t.Run(testCase.Description, func(t *testing.T) {
-			t.Parallel()
-
 			mockStore := &storetest.Store{}
 			defer mockStore.AssertExpectations(t)
 


agnivade:
> Running the sub-tests in parallel triggered a fail once 

Did you see what the failure was?

Flaky tests are sometimes notorious that way that they'll never reproduce but you need to figure out why they happened. But if you did see the failure once, you should be able to backtrack and figure out what made it fail.

noxer:
Looking at the command that VSCode is running, there is a good chance that this was just a random failure due to the 30s timeout.

agnivade:
The test was skipped in this PR: https://github.com/mattermost/mattermost-server/pull/14576/files. But unfortunately, it's quite old so CircleCI won't show us the failure output. So this is a bit of an unfortunate case.

Could you give it another try without the timeout and see if you get any failure? Just to be safe. And did you verify with the `-race` flag as well that there are no race conditions?

noxer:
I ran it a about 20 times now with a timeout of 45 minutes. I couldn't reproduce the crash. While running the test I noticed that it's the only test marked as parallel in `post_test.go`. Since all the parallel tests run at the same time, `go test` runs all other, sequential tests first before letting the parallel ones run. This may explain why it was failing when the timeout hit. It was simply the last test to run. I checked the code itself and I don't see any access to a shared variable.
To be sure I also let the `pest_test` tests run with the `-race` flag more than 10 times, yielding no errors.

agnivade (approved):

isacikgoz (approved):
Not sure if this test brings any value as it only tests a simple if condition :) Anyway, looking forward to see why this fails if it flakes again on CI.

PR #20367, opened 2022-06-05
MM-44576 autocomplete names including utf 8 chars

dimoiko100:
#### Summary

User "@" autocomplete with ascii characters fetches utf8 variants. E.g. searching for "@Jer" fetches "Jérémie".

For **postgres** I added the "unaccent" extension
For **mysql** it worked out of the box which it seems to have default collation "latin1_swedish_ci"

#### Ticket Link
[Jira_ticket](https://mattermost.atlassian.net/browse/MM-44576)
[github_issue](https://github.com/mattermost/mattermost-server/issues/20327)

#### Webapp related PR
[webapp_pr](https://github.com/mattermost/mattermost-webapp/pull/10478)

#### Release Note
```release-note
Implemented search of names with utf8 characters by searching for a similar name with only ascii characters.
```


dimoiko100:
Hello @amyblais it's ready for review. Thank you very much

store/sqlstore/user_store.go
```diff
@@ -1471,7 +1471,7 @@ func generateSearchQuery(query sq.SelectBuilder, terms []string, fields []string
 		termArgs := []interface{}{}
 		for _, field := range fields {
 			if isPostgreSQL {
-				searchFields = append(searchFields, fmt.Sprintf("lower(%s) LIKE lower(?) escape '*' ", field))
+				searchFields = append(searchFields, fmt.Sprintf("lower(unaccent(%s)) LIKE lower(unaccent(?)) escape '*' ", field))
```
agnivade, line 1474: We don't enforce installing extensions on users. They are from the contrib package and it's not mandatory for it to be there. What you can do is change the `LIKE` clause to a `tsvector` search query. Then users can change the default search configuration to use a different dictionary and postgres will pick that up.

agnivade (changes requested):

store/sqlstore/user_store.go
dimoiko100: Ok. One question. How the user is going to change the default search configuration to use different dictionary? Do you mean that the user is going to change them during searching?

agnivade: They can changes the search configuration like this: https://www.postgresql.org/docs/current/unaccent.html#id-1.11.7.54.7. You just need to pass the search configuration name that's stored in `s.pgDefaultTextSearchConfig`.

dimoiko100: I have made the changes.
I also implemented another method which seemed to make the queries 20 - 50ms longer but is more flexible in the terms
that it always returns all the accented variants using the `translate `function.
```
translationString := "'âãäåāăąÁÂÃÄÅĀĂĄèééêëēĕėęěĒĔĖĘĚìíîïìĩīĭÌÍÎÏÌĨĪĬóôõöōŏőÒÓÔÕÖŌŎŐùúûüũūŭůÙÚÛÜŨŪŬŮ','aaaaaaaaaaaaaaaeeeeeeeeeeeeeeeiiiiiiiiiiiiiiiiooooooooooooooouuuuuuuuuuuuuuuu'"
```
```
searchFields = append(searchFields, fmt.Sprintf("translate(lower(%[1]s),%[2]s) LIKE lower(?) escape '*' OR lower(%[1]s) LIKE lower(?) escape '*'", field, translationString))
```

agnivade: > But the example in the https://github.com/mattermost/mattermost-server/issues/20327 Jérémie is not retrieved when s.pgDefaultTextSearchConfig is pg_catalog.english but when it is e.g. pg_catalog.french . I suppose it is the expected behavior.

Yes, that's the correct behavior.

> It returns more results as it's language agnostic, without using any extension or the default text search configuration.

We are trying to make it language sensitive, not language agnostic.

store/sqlstore/user_store.go
```diff
@@ -1465,13 +1465,13 @@ func (us SqlUserStore) SearchNotInGroup(groupID string, term string, options *mo
...
 
-func generateSearchQuery(query sq.SelectBuilder, terms []string, fields []string, isPostgreSQL bool) sq.SelectBuilder {
+func generateSearchQuery(query sq.SelectBuilder, terms []string, fields []string, isPostgreSQL bool, textSearchConfig string) sq.SelectBuilder {
 	for _, term := range terms {
 		searchFields := []string{}
 		termArgs := []interface{}{}
 		for _, field := range fields {
 			if isPostgreSQL {
-				searchFields = append(searchFields, fmt.Sprintf("lower(%s) LIKE lower(?) escape '*' ", field))
+				searchFields = append(searchFields, fmt.Sprintf("to_tsvector('%[1]s', lower(%[2]s)) @@ to_tsquery('%[1]s', concat( lower(?) ,':*'))", textSearchConfig, field))
```
agnivade, line 1474: Adding the `:*` prefix search seems like a diverging behavior from MySQL. Let's remove this.

agnivade (changes requested):
This requires tests. They are in store/storetest/user_store.go. 

You can test for different search strings, by setting different search configurations and verifying the output.

store/sqlstore/user_store.go
dimoiko100: @agnivade I have some questions about the tests. What direction is best?

- mock a function e.g. `GetDefaultTextSearchConfig` to return a giver language
- create a new function like `SetDefaultTextSearchConfig` and call it whenever I want to change language during the test
- add default text search config inside the model.UserSearchOptions to be accessible from the tests

agnivade: You can manually set a different search configuration with `SET default_text_search_config`. We can run couple of tests with different configurations and then verify the results. Every time you initialize the store, it will read the search config from the DB again.

dimoiko100: @agnivade I've made some changes but I am maybe missing something.  Specifically
E.g. name `Jérémie` queried by the term `jere` retrieves the correct result for `default_text_search_config` equal to `spanish` but not for `french`. I do not understand why, maybe it is the expected behavior.

agnivade: > . name Jérémie queried by the term jere retrieves the correct result for default_text_search_config equal to spanish but not for french. 

Can you manually test this in the database?

> Also about the :* MySQL adds a % right after the term so I suppose MySQL also checks for prefix, right?

Ah yes that's right. So we need to remove the `%` in case of Postgres, and add the prefix search param in tsquery.

dimoiko100: Yes I tested manually. Some of the results are below.
Using the query `to_tsvector({language}, firstname) @@ to_tsquery({language}, concat(lower('jere'),':*'))`

Users:
Firstname: `Jérémie` , Language: `french`, query_term: `jere` results = 0
Firstname: `Jérémie` , Language: `spanish`, query_term: `jere` results = 1
Firstname: `Jerémie` , Language: `french`, query_term: `jere` results = 1
Firstname: `Jerémie` , Language: `spanish`, query_term: `jere` results = 1

agnivade: Ok, there you go then. Seems to be expected behavior.

store/sqlstore/store.go
```diff
@@ -357,6 +357,20 @@ func (ss *SqlStore) IsBinaryParamEnabled() bool {
 	return ss.isBinaryParam
 }
 
+func (ss *SqlStore) SetPgDefaultTextSearchConfig(defaultTextSearchConfig string) {
```
agnivade, line 360: No need to create a separate method for this. You can directly call `s.GetMasterX().Exec` from the test function.

store/storetest/user_store.go
```diff
@@ -6010,3 +6011,160 @@ func testGetUsersWithInvalidEmails(t *testing.T, ss store.Store) {
...
+	require.NoError(t, err)
+	defer func() { require.NoError(t, ss.User().PermanentDelete(u4.Id)) }()
+
+	// The users returned from the database will have AuthData as an empty string.
+	nilAuthData := new(string)
+	*nilAuthData = ""
+	u1.AuthData = nilAuthData
+	u2.AuthData = nilAuthData
+	u3.AuthData = nilAuthData
+	u4.AuthData = nilAuthData
```
agnivade, line 6064: What is the reason for this?

store/storetest/user_store.go
```diff
@@ -6010,3 +6011,160 @@ func testGetUsersWithInvalidEmails(t *testing.T, ss store.Store) {
...
+			},
+			[]*model.User{},
+			[]*model.User{u4},
+			"english",
+		},
+	}
+
+	for _, testCase := range testCases {
+		ss.SetPgDefaultTextSearchConfig(testCase.Language)
+		time.Sleep(1000)
```
agnivade, line 6155: What's the reason for this?

store/storetest/user_store.go
```diff
@@ -6010,3 +6011,160 @@ func testGetUsersWithInvalidEmails(t *testing.T, ss store.Store) {
...
+				Limit:          model.UserSearchDefaultLimit,
+			},
+			[]*model.User{},
+			[]*model.User{u4},
+			"english",
+		},
+	}
+
+	for _, testCase := range testCases {
+		ss.SetPgDefaultTextSearchConfig(testCase.Language)
```
agnivade, line 6154: We should also revert back to the original search config after this to prevent the setting from polluting other tests.

store/storetest/user_store.go
dimoiko100: I removed it now. I had added this line in order to let the query take place but it is true, it is redundant.

store/sqlstore/store.go
dimoiko100: Ok I removed it. I created because I read the `pgDefaultTextSearchConfig` from the store variable every time I made the search query. I removed it now and let the query take the default_text_search_config from the database.

store/storetest/user_store.go
dimoiko100: It seems to be a necessary action in order for the test to correctly identify the user. That is because the `u` (user) objects that the tests are being validated against have `nil` as `AuthData` but the database objects have `AuthData` equal to empty string. So with this way the validating objects' `AuthData` are being initialized as empty strings too

store/storetest/user_store.go
dimoiko100: Sorry I forgot this part. I added it.

store/storetest/user_store.go
```diff
@@ -6010,3 +6011,174 @@ func testGetUsersWithInvalidEmails(t *testing.T, ss store.Store) {
...
+			},
+			[]*model.User{},
+			[]*model.User{u4},
+			"english",
+		},
+	}
+
+	var initialDefaultTextSearchConfig string
+	if s.DriverName() == model.DatabaseDriverPostgres {
+		error := s.GetMasterX().Get(&initialDefaultTextSearchConfig, `SHOW default_text_search_config`)
```
agnivade, line 6155: We usually use `err2` for these shadowing errors, `error` is a package name so might get confusing. But not a big deal.

agnivade (approved):
Nicely done!

jgilliam17:
Thanks @dimoiko100 and apologies for delay
This passes when tested manually, but last week I wasn't able to get the Cypress cycle to complete. Trying again after update.

dimoiko100:
No problem @jgilliam17 ! Thank you very much for all the help. Do you know how and where the test server creation error is related to the new feature?

jgilliam17:
There's a significant number of failures on the [E2E report](https://mattermost-cypress-report.s3.amazonaws.com/222850-3a3dc4d-onprem-ent-server-pr-20367-mattermost/mm-ee-test3a3dc4d/mochawesome.html). 
I'm looking into why one element in particular is causing a large number of tests to fail -  `advancedTextEditorCell`.

dimoiko100:
Hello @metanerd! I did the rebase and everything seems up to date. I made a mistake though and I did not squash the commits.

saturninoabril:
I've looked at the build hash and the webapp is still referencing the [old commit](https://github.com/mattermost/mattermost-webapp/commit/b395083755bd9e429a22be2038935c075fcb91b8) and not with master. @pfltdv Could you please take a look on why the build is not using the latest master? Thanks in advance!
Screen Shot 2022-07-31 at 7 06 35 PM. 

@dimoiko100 For the meantime, could you please delete your old webapp branch ([MM-44576_autocomplete_names_including_utf-8_chars](https://github.com/dimoiko100/mattermost-webapp/tree/MM-44576_autocomplete_names_including_utf-8_chars)), then let's update this server PR branch and see how the build works after?

dimoiko100:
@saturninoabril webapp branch is deleted

saturninoabril:
Build with latest `0f6b0c5` is now as expected with webapp/enterprise using master.
Screen Shot 2022-08-01 at 9 22 56 AM

saturninoabril:
Thanks @dimoiko100!

@jgilliam17 E2E should be as expected now. You may find the results [here](https://mattermost-cypress-report.s3.amazonaws.com/224550-0f6b0c5-onprem-ent-server-pr-20367-mattermost/mm-ee-test0f6b0c5/mochawesome.html) to be compared with latest daily run.

pfltdv:
> I've looked at the build hash and the webapp is still referencing the [old commit](https://github.com/mattermost/mattermost-webapp/commit/b395083755bd9e429a22be2038935c075fcb91b8) and not with master. @pfltdv Could you please take a look on why the build is not using the latest master? Thanks in advance! Screen Shot 2022-07-31 at 7 06 35 PM.
> 
> @dimoiko100 For the meantime, could you please delete your old webapp branch ([MM-44576_autocomplete_names_including_utf-8_chars](https://github.com/dimoiko100/mattermost-webapp/tree/MM-44576_autocomplete_names_including_utf-8_chars)), then let's update this server PR branch and see how the build works after?

@saturninoabril All pipelines uses same branch names on all projects if exits. At build time if there is a branch with same name exists at mattermost-webapp, server and enterprise pipeline will use that branch. Otherwise it will use default branch.

At build time there were a branch and pipelines used that branch. 4/5 sure that that branch was not updated. (After deleting the branch generated artifacts has the correct commits since pipelines will use master.)

Unfortunately it is not possible to check at the moment since branch was deleted.

jgilliam17:
Thanks @dimoiko100 and @saturninoabril for running e2es
Looks good to merge.
[E2E report ](https://mattermost-cypress-report.s3.amazonaws.com/224550-0f6b0c5-onprem-ent-server-pr-20367-mattermost/mm-ee-test0f6b0c5/mochawesome.html)looks good, no PR related failures when compared to [master](https://mattermost-cypress-report.s3.amazonaws.com/224556-onprem-ent-master-mm-ee-test:test/mochawesome.html)

jgilliam17 (approved):

dimoiko100:
Thanks @jgilliam17 for merging and all the review!

agarciamontoro:
@dimeko: Unfortunately we had to revert this PR, since the load-tests we run in every release surfaced a performance regression that was caused by this new way of searching. We should have caught this earlier, that's completely on me! If you want more details, you can see the full report in [community](https://community.mattermost.com/core/pl/hfufprbegprqbkwzsynzfad34a). I am also copying it here in case you're not registered there:

<details>
<summary>Investigation report</summary>

# Performance regression investigation in Postgres v7.2.0 - v7.3.0-RC1

The release load-tests that compared v7.3-RC1 against v7.2 showed a significant drop in performance when using Postgres. This report describes the investigation carried out to uncover the causes of such regression.

## Results

The cause of the performance regression was the change applied in [this PR](https://github.com/mattermost/mattermost-server/pull/20367), that introduced a new way to search and autocomplete users that was language-aware, so that for example looking for `andres` would also return `andrés`. There were two main problems:

1.  The new query was not optimized for using the already in-place GIN indexes, but even with [a fix](https://github.com/mattermost/mattermost-server/commit/4aca2929c2bc3233984557848cef6efe5e12973d) that would leverage such indexes in the new query, the performance regression was still in place—albeit smaller, with a number of supported users 1.12 times lower for the new build, instead of 1.2 times lower as reported in the first tests.
2.  The language of the GIN indexes are right now hard-coded to `english`, so the improved query would not be able to use the indexes unless the default text search configuration of the database would be set to `english`. In such case, though, looking for `andres` would no longer return `andrés`, so the original improvement of the PR would become moot.

For this, we have decided to revert [the original PR](https://github.com/mattermost/mattermost-server/pull/20367) and the subsequent changes ([one](https://github.com/mattermost/mattermost-server/pull/20863), [two](https://github.com/mattermost/mattermost-server/pull/20963)). A deeper investigation of other alternatives will be required to implement this feature.

## Methodology

Three different tests were carried out:

  - T0: The original release comparison, which used the usual setup in such tests. See the whole configuration in [the original thread](https://community.mattermost.com/core/pl/7jxjuen49pywjqfk6w85n5jg3h).
  - T1: An exact copy of 1. to ensure the results were reproducible.
  - T2: A copy of 1. with a fix for making sure that Postgres was using the index in the new query: before starting each test, the database parameter `default_text_search_config` was set to `english`, instead of the default `simple`, and [this patch](https://github.com/mattermost/mattermost-server/commit/4aca2929c2bc3233984557848cef6efe5e12973d) was applied to ensure that Postgres was using the GIN indexes in the modified queries.

This report focuses mostly on T1 and T2, since the Prometheus data was no longer available in T0 when this investigation was carried out.

## Symptoms

When the original test, T0, was first analyzed, the main conclusion drawn was:

> Postgres performance seems to have degraded significantly (the unbounded test returns a number of supported users 1.2 times lower for the v7.3.0-rc1 version (10367 vs 8675)). The cause of this regression is still unknown.

These symptoms were reproduced in test T1, and while T2 returned better results, a performance degradation was still in place:

|   | # users v7.2.0  | # users v7.3.0-rc1  | v7.2.0 / v7.3.0-rc1 |
|---|---|---|---|
| T0  | 10367  | 8675  | 1.20  |
| T1  |  10745 | 8850  | 1.21  |
| T2  |  9886 | 8827  | 1.12  |


## Investigation

### What is causing the regression?

After some querying the Grafana dasboards and the slow query logs, it was clear that the problem seemed to be caused by several slow queries queuing in the DB. To find out which DB methods were the slowest in the new build against the base one, one can simply get the average of each method and substract the one from the base build from the new one with a query such as the following:

``` promql
(
    sum(rate(mattermost_db_store_time_sum[1m])) by (method)
    /
    sum(rate(mattermost_db_store_time_count[1m])) by (method)
)
-
(
    sum(rate(mattermost_db_store_time_sum[1m] offset 6269s)) by (method)
    /
    sum(rate(mattermost_db_store_time_count[1m] offset 6269s)) by (method)
)
```

That is quite verbose, but it's just computing the average time spent in each DB method and, for each one, getting the difference between the new build minus the base one. The offset depends on the time difference between the start of the base test and the new test. For test T1, it returns this graph, where a line above zero corresponds to a DB method slower in `v7.3.0-rc1` and a line below zero to a DB method slower in `v7.2.0`:



The interpretation of this is clear: there is one single method that is taking significantly more time in `v7.3.0-rc1` than in `v7.2.0`. Looking at the data in Grafana, this method turns out to be `UserStore.AutocompleteUsersInChannel`. Other methods above zero are `PostStore.SearchPostsForUser` and `UserStore.Search`.

The same graph for test T2 follows. The results are similar, but less pronounced: `UserStore.AutocompleteUsersInChannel` is still the slowest method.



In order to give hard numbers, we can compute at each point the cumulative difference of the average of each DB method. Taking the value at the end of the tests, we can get the slowest ones. The query used is

``` promql
(
    sum(mattermost_db_store_time_sum) by (method)
    /
    sum(mattermost_db_store_time_count) by (method)
)
-
(
    sum(mattermost_db_store_time_sum offset 6269s) by (method)
    /
    sum(mattermost_db_store_time_count offset 6269s) by (method)
)
```

Getting the instant values at the end of the tests, we can get the six slowest methods for T1 and get those same times for T2. The set of the four slowest methods is the same in both tests; while `SearchPostsForUser` is at position 5 in T1 and position 9 in T2, and `Search` is at position 6 in T1 and position 5 in T2. The following table is ordered by the values for test T1.

|                                        | T1  | T2   |
| -------------------------------------- | --- | ---- |
| `UserStore.AutocompleteUsersInChannel` | 10  | 5.2  |
| `PostStore.AnalyticsPostCount`         | 4.5 | 2.2  |
| `ReactionStore.GetTopForUserSince`     | 2.2 | 1.1  |
| `ReactionStore.GetTopForTeamSince`     | 2.1 | 1.3  |
| `PostStore.SearchPostsForUser`         | 1.7 | 0.12 |
| `UserStore.Search`                     | 1.2 | 0.96 |

The `AnalyticsPostCount` query is, as suggested by its name, an analytic query, so it's expected that it is slowest when the DB is clogged. It happens something similar for the two queries in the `ReactionStore`, which are known to be slow and will be affected more in this setup. The other three are all search queries, but the difference in times in `SearchPostsForUser` and `Search` is negligible in both tests. The one that stands out is the first one, `UserStore.AutocompleteUsersInChannel`.

This points out to a change in the code to auto-complete users in a channel. A quick look to the commits between `v7.2.0` and `v7.3.0-rc1` is enough to find a potential cause: [PR \#20367](https://github.com/mattermost/mattermost-server/pull/20367). This PR changes the `generateSearchQuery` function to use [full text search queries](https://www.postgresql.org/docs/12/textsearch.html) instead of plain `LIKE` clauses only in Postgres.

The suggested fix introduced in the first section alleviates the pressure on the DB to half of the original regression, but it is still significant.

### Why is it causing a regression?

Once the change causing the performance regression was identified, we needed to understand why the regression was happening. When looking at the slow query log, one query was consistently being logged:

``` sql
SELECT u.Id, u.CreateAt, u.UpdateAt, u.DeleteAt, u.Username, u.Password, u.AuthData, u.AuthService, u.Email, u.EmailVerified, u.Nickname, u.FirstName, u.LastName, u.Position, u.Roles, u.AllowMarketing, u.Props, u.NotifyProps, u.LastPasswordUpdate, u.LastPictureUpdate, u.FailedAttempts, u.Locale, u.Timezone, u.MfaActive, u.MfaSecret, b.UserId IS NOT NULL AS IsBot, COALESCE(b.Description, '') AS BotDescription, COALESCE(b.LastIconUpdate, 0) AS BotLastIconUpdate, u.RemoteId
FROM Users u
LEFT JOIN Bots b ON ( b.UserId = u.Id )
WHERE u.DeleteAt = 0 AND (
      to_tsvector(lower(Username)) @@ to_tsquery(concat(lower('agmlt-agent-'),':*')) OR
      to_tsvector(lower(FirstName)) @@ to_tsquery(concat(lower('agmlt-agent-'),':*')) OR
      to_tsvector(lower(LastName)) @@ to_tsquery(concat(lower('agmlt-agent-'),':*')) OR
      to_tsvector(lower(Nickname)) @@ to_tsquery(concat(lower('agmlt-agent-'),':*')) OR
      to_tsvector(lower(Email)) @@ to_tsquery(concat(lower('agmlt-agent-'),':*'))
)
ORDER BY Username ASC
LIMIT 100;
```

The `WHERE` clause is the part generated by the new `generateSearchQuery` function, now using the full text search capabilities. The `EXPLAIN ANALYZE` from this query returns the
following:

``` sql
 Limit  (cost=41465.72..41465.73 rows=1 width=917) (actual time=9532.141..9532.232 rows=100 loops=1)
   ->  Sort  (cost=41465.72..41465.73 rows=1 width=917) (actual time=9532.140..9532.221 rows=100 loops=1)
         Sort Key: u.username
         Sort Method: top-N heapsort  Memory: 195kB
         ->  Nested Loop Left Join  (cost=1000.00..41465.71 rows=1 width=917) (actual time=0.475..9323.882 rows=11037 loops=1)
               Join Filter: ((b.userid)::text = (u.id)::text)
               Rows Removed by Join Filter: 88296
               ->  Gather  (cost=1000.00..41464.60 rows=1 width=876) (actual time=0.465..8597.685 rows=11037 loops=1)
                     Workers Planned: 1
                     Workers Launched: 1
                     ->  Parallel Seq Scan on users u  (cost=0.00..40464.50 rows=1 width=876) (actual time=2161.958..6243.189 rows=5518 loops=2)
                           Filter: ((deleteat = 0) AND ((to_tsvector(lower((username)::text)) @@ to_tsquery(concat('agmlt-agent-'::text, ':*'))) OR (to_tsvector(lower((firstname)::text)) @@ to_tsquery(concat('agmlt-agent-'::text, ':*'))) OR (to_tsvector(lower((lastname)::text)) @@ to_tsquery(concat('agmlt-agent-'::text, ':*'))) OR (to_tsvector(lower((nickname)::text)) @@ to_tsquery(concat('agmlt-agent-'::text, ':*'))) OR (to_tsvector(lower((email)::text)) @@ to_tsquery(concat('agmlt-agent-'::text, ':*')))))
                           Rows Removed by Filter: 8045
               ->  Seq Scan on bots b  (cost=0.00..1.05 rows=5 width=594) (actual time=0.037..0.040 rows=8 loops=11037)
 Planning Time: 1.208 ms
 Execution Time: 9532.305 ms
(16 rows)
```

From this output we see that the planner is not using any index, defaulting to a sequential scan on users. This looks wrong, since we do have the following GIN indexes in the `Users` table that this query should be leveraging:

``` sql
CREATE INDEX idx_users_all_no_full_name_txt ON public.users USING gin
             (to_tsvector('english', username || ' ' || nickname || ' ' || email));
CREATE INDEX idx_users_all_txt ON public.users USING gin
             to_tsvector('english', username || ' ' || firstname || ' ' || lastname || ' ' || nickname || ' ' || email));
CREATE INDEX idx_users_names_no_full_name_txt ON public.users USING gin
             (to_tsvector('english', username || ' ' || nickname));
CREATE INDEX idx_users_names_txt ON public.users USING gin
             (to_tsvector('english', username || ' ' || firstname || ' ' || lastname || ' ' || nickname));
```

So why aren't the query using these indexes? This can be explained by the way the query is built. For a GIN index to be used, the following two criteria must be met:

1.  The `tsvector` needs to use the same concatenated columns that appear in the index, in the same order.
2.  The `tsvector` needs to use the same `regconfig` parameter as the index.

The query does not meet any of these criteria (the second is met only if `default_text_search_config` is set to `english`), so the GIN index is not used.

### Potential solutions

#### Fixing the query

The most obvious solution here is to fix the query so that Postgres plans to use the GIN indexes we have in place. This can be done applying [this patch](https://github.com/mattermost/mattermost-server/commit/4aca2929c2bc3233984557848cef6efe5e12973d), but this still has a problem:

While the structure of the query would be fixed so that criterion number 1 is met, criterion number 2 still depends on the database configuration. In the GIN indexes the database has, the `'english'` value is hard-coded, so we'd need to either change these indexes in individual installations to match the language of the instance, or hard-code the `'english'` value in the queries. Both approaches have cons:

1.  If we change the language of the indexes, we just move the problem to another language. On top of the overhead of creating a new index, if the language of the instance is, for example, Swedish, and we try to look for a user with a French name, we won't get the benefit of the change.
2.  If we hard-code the value of the language in the queries to match the hard-coded `'english'` in the indexes, the original goal of the change becomes moot, since this configuration would not return accented words for unaccented inputs.

In both cases we would still suffer the performance regression.

#### Reverting

The other solution is to revert the original changes. We will lose the ability to match accented users with non-accented inputs, but we will avoid the performance regression measured in the tests.

### Decision

Taking into account all the previous discussion, we decided to revert the original changes. If we want to support this feature, a deeper investigation into other alternatives that do not degrade performance will be needed.

## Next steps

1. [MM-47265](https://mattermost.atlassian.net/browse/MM-47265): Set `default_text_search_config` to `english` in the load-test tool.
2. [MM-47374](https://mattermost.atlassian.net/browse/MM-47374): Revert PRs [#20367](https://github.com/mattermost/mattermost-server/pull/20367), [#20863](https://github.com/mattermost/mattermost-server/pull/20863) and [#20963](https://github.com/mattermost/mattermost-server/pull/20963) and cherry-pick to v7.3 and v7.4.
3. [MM-47375](https://mattermost.atlassian.net/browse/MM-47375): Collect useful Grafana queries to speed up future investigations.

----

</details>

Thank you for all your work on this feature, even if it was reverted we've learned a ton thanks to it. And as stated in the report, investigation into other alternatives to implement this is still welcome!

Thanks again :)

PR #20384, opened 2022-06-07
Simplify ParamsFromRequest

noxer:
#### Summary
This PR cleans up the `ParamsFromRequest` function. It simplifies the file by using two default behaviors of Go. Firstly the `map` type returns the default value for keys not in the map and secondly, struct properties are initialized with the default value for their respective type. Also `strconv.Parse*` functions return the default value when an parsing error has occurred. Putting this together allows us to replace
```go
if var, ok := props["key"]; ok {
    params.Key = val
}
```
with
```go
params.Key = props["key"]
```
This greatly simplifies the function and makes it more readable.

The PR also adds tests for the more complex transformations.

#### Release Note
```release-note
NONE
```

### DIFF
--- a/web/params.go
+++ b/web/params.go
@@ -101,29 +101,12 @@ func ParamsFromRequest(r *http.Request) *Params {
 	props := mux.Vars(r)
 	query := r.URL.Query()
 
-	if val, ok := props["user_id"]; ok {
-		params.UserId = val
-	}
-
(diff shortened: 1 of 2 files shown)


agnivade (approved):
Brilliant work and nice tests!

isacikgoz (approved):
Awesome stuff 💯

PR #20388, opened 2022-06-07
[MM-44512] Pre-package Calls v0.6.0 and enable it by default

streamer45:
#### Summary

Pre-packaging latest version of Calls (v0.6.0) for June 15th release.  
Also enabling the plugin by default from now on, although still controllable through the dedicated feature flag if needed.

/cc @amyblais 

#### Ticket Link

https://mattermost.atlassian.net/browse/MM-44512

#### Release Note

```release-note
Pre-packaged Calls v0.6.0 and enabled it by default.
```


cpoile (approved):

app/plugin.go
```diff
@@ -1121,8 +1121,8 @@ func (ch *Channels) getPluginStateOverride(pluginID string) (bool, bool) {
 			return true, false
 		}
 	case model.PluginIdCalls:
-		if model.IsCloud() {
-			return true, ch.cfgSvc.Config().FeatureFlags.CallsEnabled
+		if !ch.cfgSvc.Config().FeatureFlags.CallsEnabled {
```
lieut-data, line 1124: Given this feature flag existed before, and presumably defaulted to false, will existing installations that chose to enable it discover it disabled via this feature flag unexpectedly?

lieut-data (approved):

app/plugin.go
streamer45: Merging but please let's continue the discussion here if necessary.

amyblais:
@streamer45 Needs a manual cherry-pick to ``release-7.0`` and ``cloud``.

streamer45:
> @streamer45 Needs a manual cherry-pick to `release-7.0` and `cloud`.

Yeah, I am looking into this.

app/plugin.go
lieut-data: I think I was worried that `CallsEnabled: false` might have been written to `config.json` in a past version, and now it would "force it off," but looking at my own local installation, it seems we don't persist these to the actual config anyway, so it sounds like a non-issue. Thanks for the follow-up!

PR #20398, opened 2022-06-08
[MM-44925] simplified the ID generator

noxer:
#### Summary
This PR improves the ID generation. Instead of generating the base32 encoding with the padding and then stripping the padding off, we just generate it without the padding to begin with. The result is exactly the same but without the need to allocate a byte buffer and with much simpler code.

#### Ticket
[MM-44925](https://mattermost.atlassian.net/browse/MM-44925)

#### Release Note
```release-note
NONE
```

### DIFF
--- a/model/utils.go
+++ b/model/utils.go
@@ -4,7 +4,6 @@
 package model
 
 import (
-	"bytes"
 	"crypto/rand"
 	"database/sql/driver"
 	"encoding/base32"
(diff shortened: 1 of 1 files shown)


isacikgoz (approved):
Nice

agnivade (approved):
Great stuff! Feel free to add the improvement for random bytes generation with this PR or later. Up to you.

PR #20413, opened 2022-06-09
MM-43733: Set concurrency limits via configuration

agnivade:
```release-note
We create a new config option MaxImageDecoderConcurrency which
indicates how many images can be decoded concurrently at once.

The default is -1 which means number of CPUs present.

This affects the total memory consumption of the server. The
maximum memory of a single image is dictated by MaxImageResolution * 24 bytes.
Therefore, a good rule of thumb to follow is that MaxImageResolution
* MaxImageDecoderConcurrency * 24 should be less then the allocated memory for
image decoding.
```

https://mattermost.atlassian.net/browse/MM-43733

### DIFF
--- a/app/channels.go
+++ b/app/channels.go
@@ -197,8 +197,12 @@ func NewChannels(s *Server, services map[ServiceKey]interface{}) (*Channels, err
 	}
 
 	var imgErr error
+	decoderConcurrency := int(*ch.cfgSvc.Config().FileSettings.MaxImageDecoderConcurrency)
+	if decoderConcurrency == -1 {
+		decoderConcurrency = runtime.NumCPU()
+	}
 	ch.imgDecoder, imgErr = imaging.NewDecoder(imaging.DecoderOptions{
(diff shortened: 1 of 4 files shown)


jupenur (approved):

streamer45 (approved):
Nice improvement :+1:

amyblais:
@agnivade Since this needs to be backported, is this safe to cherry-pick to 7.0 now or should we let this soak for a bit?

agnivade:
Hmm .. no I don't think this needs to be backported. Unless @jupenur thinks otherwise.

agnivade:
We are having internal discussions with the security team on whether this needs to be backported or not. I'll keep you posted.

/e2e-test

agnivade:
@saturninoabril - Can you check this and let me know. Thanks!

amyblais:
The ticket originally was set to medium level, I see now that it was changed to low level so it doesn't need to backported.

furqanmlk:
@agnivade 
E2E Passed, no new issue found

PR #20429, opened 2022-06-10
MM-44805: Avoid pointers in JSON parsing in createCategoryForTeamForUser

agnivade:
Passing null is a valid JSON input for a pointer struct. So we avoid
passing pointer to pointer.

https://mattermost.atlassian.net/browse/MM-44805

```release-note
NONE
```

### DIFF
--- a/api4/channel_category_test.go
+++ b/api4/channel_category_test.go
@@ -5,6 +5,7 @@ package api4
 
 import (
 	"encoding/json"
+	"fmt"
 	"testing"
 	"time"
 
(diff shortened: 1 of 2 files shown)


api4/channel_category_test.go
```diff
@@ -90,6 +91,17 @@ func TestCreateCategoryForTeamForUser(t *testing.T) {
...
 	})
 
+	t.Run("should not crash with null input", func(t *testing.T) {
+		require.NotPanics(t, func() {
+			user, client := setupUserForSubtest(t, th)
+			payload := []byte(`null`)
+			route := fmt.Sprintf("/users/%s/teams/%s/channels/categories", user.Id, th.BasicTeam.Id)
+			r, err := client.DoAPIPostBytes(route, payload)
+			require.Error(t, err)
+			defer closeBody(r)
```
noxer, line 101: Why is there a `defer` on the last line of the function?

noxer (changes requested):

api4/channel_category_test.go
agnivade: heh good point. Copy pasta :)

isacikgoz (approved):
LGTM, same viewpoint with the other PR :)

noxer (approved):
LGTM, thank you for the change

PR #20439, opened 2022-06-10
Mm 44488 enforce posts support sorts

neallred:
#### Summary
`filterInaccessiblePosts` supports ascending and descending time.
`getPostAccessibleBounds` helper runs in logarithmic time with small optimizations for common cases where no posts are accessible or all posts are accessible.

#### Ticket Link
Supports work done for https://mattermost.atlassian.net/browse/MM-44488

#### Release Note
```release-note
NONE
```


vish9812:
Great PR @neallred 👍 
However, I've tested it a couple of times, and it seems, it is returning _inverted_ result. 
Like it's actually removing the posts.CreateAt >= lastAccessiblePostTime instead of returning them.
Can you please double check it?

neallred:
> it is returning *inverted* result.

Good catch! The helper function I added was ok, but my usage of it in `filterInaccessiblePosts` was wrong. Added a couple tests to that function to make sure it returns the correct posts. Also renamed the helper function to be clearer about the assumptions of it, since we have some post search results that aren't sorted primarily by time.

vish9812 (approved):

app/post.go
```diff
@@ -810,7 +810,13 @@ func (a *App) GetPostsPage(options model.GetPostsOptions) (*model.PostList, *mod
 		}
 	}
 
-	if appErr := a.filterInaccessiblePosts(postList); appErr != nil {
+	// we use the slower route if not a collapsed threads request;
+	// Even though we know the parent and child subqueries are ordered within themselves, they are not added to the list in an ordered fashion with respect to each other, and there is not a 1:1 correspondence between Order and Posts (root posts are not added to Order).
+	filterOptions := filterPostOptions{}
```
vish9812, line 815: It looks to me, that `rootPosts` is a superset of `parentsPosts`. And since we add `order` from ~~parentsPosts~~ rootPosts only, then posts must be sorted. So, I think in this case we can set `assumeSortedCreatedAt` to `true`.

Although _another topic_, I don't see why add parentsPosts at all, as the rootPosts(superSet) is already being added.

app/post.go
```diff
@@ -883,7 +889,7 @@ func (a *App) GetPostThread(postID string, opts model.GetPostsOptions, userID st
 		}
 	}
 
-	if appErr := a.filterInaccessiblePosts(posts); appErr != nil {
+	if appErr := a.filterInaccessiblePosts(posts, filterPostOptions{assumeSortedCreatedAt: true}); appErr != nil {
```
vish9812, line 892: For this one, sorting is dependent on `opts.Direction`, if it has no value then there is no sorting.

app/post_helpers.go
```diff
@@ -0,0 +1,193 @@
...
+func getTimeSortedPostAccessibleBounds(postList *model.PostList, earliestAccessibleTime int64) postAccessibleBounds {
+	if postList == nil || postList.Posts == nil || len(postList.Posts) == 0 {
+		return allAccessiblePosts
+	}
+	posts := postList.Posts
+	order := postList.Order
+	lenPosts := len(posts)
+
+	if lenPosts != len(order) {
+		return allAccessiblePosts
```
vish9812, line 58: In this case, shouldn't we fallback to linear?

app/post_helpers.go
```diff
@@ -0,0 +1,193 @@
...
+		if bounds.accessible > bounds.inaccessible {
+			otherInaccessibleBound = 0
+			otherAccessibleBound = len(postList.Order) - 1
+		} else {
+			otherInaccessibleBound = len(postList.Order) - 1
+			otherAccessibleBound = 0
+		}
+		order := postList.Order
+		for i := min(bounds.inaccessible, otherInaccessibleBound); i <= maxInt(bounds.inaccessible, otherInaccessibleBound); i++ {
+			delete(postList.Posts, order[i])
```
vish9812, line 184: Could it be beneficial to decide the `delete/make` operation conditionally, in case the `len` of inaccessible posts is much higher than the accessible posts. In those cases, we can create a new map instead of deleting keys?

PR #20454, opened 2022-06-13
File storage usage default 0

neallred:
#### Summary
If there are no entries in the FileInfo table, then null results, which can't be cast to a number. So we add a COALESCE expression to default to 0 in that case. Because the frontend handles errors gracefully, I don't think we need to cherry-pick this for June 15 release, although we may see a few error logs in brand new installations until a file is uploaded.

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-45057

#### Release Note
```release-note
Fix for cloud instances that had backend errors when fetching file storage usage when no files had been uploaded.
```


marianunez (approved):
Thanks for the quick fix on this! @neallred Agreed this is not required for launch tomorrow. 👍

agnivade (approved):
Nice work on the UT!

PR #20503, opened 2022-06-20
Remove go.work files in make nuke

agnivade:
This can cause some confusion while switching
to older branches which does not support module
workspaces.

```release-note
NONE
```


agnivade:
cc @sadohert

noxer (approved):
LGTM

PR #20508, opened 2022-06-20
MM-45209  Revert PR checking for GetUser returning nil,nil

wiggin77:
#### Summary
Reverts the PR https://github.com/mattermost/mattermost-server/pull/20501 which was added for SET escalation.

It was determined that the plugins call to GetUser never made it to the app layer.

#### Ticket Link
Fixes https://mattermost.atlassian.net/browse/MM-45209

#### Release Note
```release-note
NONE
```

### DIFF
--- a/app/user.go
+++ b/app/user.go
@@ -388,12 +388,6 @@ func (a *App) GetUser(userID string) (*model.User, *model.AppError) {
 		}
 	}
 
-	// Temporary check to diagnose a SET escalation; when GetUser is called by some plugins a nil,nil is returned.
-	if user == nil {
-		mlog.Error("Unexpected nil from GetUser", mlog.String("user_id", userID))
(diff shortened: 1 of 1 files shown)


ashishbhate (approved):

agnivade:
@ashishbhate - Did you confirm that the issue was fixed with the opengraph revert?

ashishbhate:
I've been keeping an eye on the logs - 

The currently deployed version is the revert PR #20508 (7793ae), which was deployed approximately at 22:00 IST yesterday (20th June). This PR hasn't been deployed yet. 

Since then I have seen only two 
`msg:panic: runtime error: invalid memory address or nil pointer dereference` entries in the log, and only from the "mattermost-plugin-newchannelnotify" plugin. According to the logs they're coming from - https://gitlab.com/thepill/mattermost-plugin-newchannelnotify/-/blob/master/server/plugin.go#L77 , so I'm not sure if they're related. 

Since the deployment, I'm not seeing any "RPC call to KVGet API failed" in the logs, which was a symptom of the the issue.

agnivade:
Great, thanks!

PR #20516, opened 2022-06-21
MM-40450: Follow redirects in image proxy

agnivade:
For the local proxy, we weren't following redirects.
We remove the CheckRedirect function and use the default
function, which is good enough for our purposes.

https://mattermost.atlassian.net/browse/MM-40450

```release-note
NONE
```

### DIFF
--- a/services/imageproxy/local.go
+++ b/services/imageproxy/local.go
@@ -64,9 +64,6 @@ func makeLocalBackend(proxy *ImageProxy) *LocalBackend {
 	}
 
 	client := proxy.HTTPService.MakeClient(false)
-	client.CheckRedirect = func(newreq *http.Request, via []*http.Request) error {
-		return http.ErrUseLastResponse
-	}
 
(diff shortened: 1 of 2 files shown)


noxer (approved):
Looks good to me. Why was is restricted anyway?

agnivade:
No idea. Initially we used the https://github.com/willnorris/imageproxy library, but later we removed the code and copied over the necessary bits. But this part https://github.com/willnorris/imageproxy/blob/6022f6a372c01b3e900a4d7343edc443ef4fc4f4/imageproxy.go#L207-L212 was copied over as-is.

hmhealey (approved):
Looks good to me. I don't think it was intentional that we didn't previously follow redirects with the local proxy.

Also, I noticed that the old proxy we were using also checked for a maximum number of redirects, but that's not needed with the default version of `CheckRedirect` (it limits it to 10 max)

PR #20521, opened 2022-06-22
MM-44953: Refactor EnsureBot to be used from plugin.API

agnivade:
The mattermost-plugin-api code also needs to use this.
Hence we make it available from plugin.API.

https://mattermost.atlassian.net/browse/MM-44953

```release-note
NONE
```


isacikgoz (approved):
Nice 👍

PR #20522, opened 2022-06-22
[MM-45250] Add indices to reactions, channelmembers 

plant99:
#### Summary

Adds indices on the following columns

- reactions - postid, userid
- channelmembers - channelid

#### Ticket Link

https://mattermost.atlassian.net/browse/MM-45250
#### Release Note

```release-note
NONE
```


agnivade:
@plant99 - Adding indices to these tables is a big task and requires several steps. Can you schedule some time with me on Monday? I want to go through some items with you. Please go through https://developers.mattermost.com/contribute/server/schema-migration-guide/#how-do-i-measure-the-impact-of-the-migration as well.

In the call, I'd like to go through these items:

**Releasing an index change**
- Making the MySQL and Postgres index additions non locking.
- Measuring the time the query takes on a large DB.
- Writing out the exact queries so that customers can run them before hand during off hours. This is done so that the actual upgrade doesn't block and happens instantenously.
- Adding that info to the release notes for our customers.

**Analysing the index change**
- Please prepare the query plan outputs for both MySQL and Postgres before hand (**on a large dataset**), and how does it change after adding those indexes. I'd like to go through them. If possible, you can post them here so that I can take a look before.

No rush if you don't have the query plan outputs handy. Prepare them and we can go over the bullet points.

plant99:
Thanks for looking at it @agnivade , I'll setup a call. And I'll prepare the query plans, and loadtest for you to look at.

PR #20526, opened 2022-06-22
:construction:  Audit logging -- convert audit logs to use the new schema

oh6hay:
#### Summary

Convert calls to audit logging API to use the [new schema](https://github.com/mattermost/mattermost-server/pull/20447)

#### Release Note

TBD


api4/bot.go
```diff
@@ -39,7 +39,7 @@ func createBot(c *Context, w http.ResponseWriter, r *http.Request) {
 
 	auditRec := c.MakeAuditRecord("createBot", audit.Fail)
 	defer c.LogAuditRec(auditRec)
-	auditRec.AddMeta("bot", bot)
+	auditRec.AddEventParameter("bot", bot)
```
DSchalla, line 42:
```suggestion
	auditRec.AddEventParameter("bot", bot.Auditable())
```

We should call `Auditable` on the struct one way or another. Either explicitly as above, or we could alternatively try to do a type cast to the interface in the `AddEventParameter` function and then call the function, that way the developer cannot can forget it. @agnivade may have thoughts on this as well.

api4/bot.go
```diff
@@ -89,7 +90,8 @@ func patchBot(c *Context, w http.ResponseWriter, r *http.Request) {
 
 	auditRec := c.MakeAuditRecord("patchBot", audit.Fail)
 	defer c.LogAuditRec(auditRec)
-	auditRec.AddMeta("bot_id", botUserId)
+	auditRec.AddEventParameter("bot_id", botUserId)
```
DSchalla, line 93:
```suggestion
	auditRec.AddEventParameter("id", botUserId)
```

Let's use `id` for all entities to make analysis easier. The context helps to understand which id it is.

api4/bot.go
```diff
@@ -89,7 +90,8 @@ func patchBot(c *Context, w http.ResponseWriter, r *http.Request) {
...
+	auditRec.AddEventParameter("bot_patch", botPatch)
```
DSchalla, line 94:
```suggestion
	auditRec.AddEventParameter("bot", botPatch)
```

As the event is called `Patch`, we can have a consistency across create and update events and use the same parameter. The question above re: `Auditable` remains.

api4/bot.go
```diff
@@ -205,8 +208,8 @@ func updateBotActive(c *Context, w http.ResponseWriter, active bool) {
 
 	auditRec := c.MakeAuditRecord("updateBotActive", audit.Fail)
 	defer c.LogAuditRec(auditRec)
-	auditRec.AddMeta("bot_id", botUserId)
-	auditRec.AddMeta("enable", active)
+	auditRec.AddEventParameter("bot_id", botUserId)
```
DSchalla, line 211:
```suggestion
	auditRec.AddEventParameter("id", botUserId)
```

api4/bot.go
```diff
@@ -238,8 +242,8 @@ func assignBot(c *Context, w http.ResponseWriter, _ *http.Request) {
 
 	auditRec := c.MakeAuditRecord("assignBot", audit.Fail)
 	defer c.LogAuditRec(auditRec)
-	auditRec.AddMeta("bot_id", botUserId)
-	auditRec.AddMeta("assign_user_id", userId)
+	auditRec.AddEventParameter("bot_id", botUserId)
+	auditRec.AddEventParameter("assign_user_id", userId)
```
DSchalla, line 246:
```suggestion
	auditRec.AddEventParameter("id", botUserId)
	auditRec.AddEventParameter("user_id", userId)
```

api4/channel.go
```diff
@@ -311,7 +316,8 @@ func patchChannel(c *Context, w http.ResponseWriter, r *http.Request) {
 
 	auditRec := c.MakeAuditRecord("patchChannel", audit.Fail)
 	defer c.LogAuditRec(auditRec)
-	auditRec.AddMeta("channel", oldChannel)
+	auditRec.AddEventParameter("channel_patch", patch)
```
DSchalla, line 319:
```suggestion
	auditRec.AddEventParameter("channel", patch)
```

api4/channel.go
```diff
@@ -439,7 +449,7 @@ func createDirectChannel(c *Context, w http.ResponseWriter, r *http.Request) {
 		otherUserId = userIds[1]
 	}
 
-	auditRec.AddMeta("other_user_id", otherUserId)
+	auditRec.AddEventParameter("other_user_id", otherUserId)
```
DSchalla, line 452:
```suggestion
	auditRec.AddEventParameter("user_id", otherUserId)
```

api4/channel.go
```diff
@@ -458,6 +468,8 @@ func createDirectChannel(c *Context, w http.ResponseWriter, r *http.Request) {
 		return
 	}
 
+	auditRec.AddEventResultState(sc)
+	auditRec.AddEventObjectType("channel")
 	auditRec.Success()
 	auditRec.AddMeta("channel", sc)
```
DSchalla, line 474: Should this be removed?

api4/channel.go
```diff
@@ -1211,8 +1225,9 @@ func deleteChannel(c *Context, w http.ResponseWriter, r *http.Request) {
 	}
 
 	auditRec := c.MakeAuditRecord("deleteChannel", audit.Fail)
+	auditRec.AddEventParameter("channel_id", c.Params.ChannelId)
```
DSchalla, line 1228:
```suggestion
	auditRec.AddEventParameter("id", c.Params.ChannelId)
```

api4/channel.go
```diff
@@ -1717,6 +1735,8 @@ func addChannelMember(c *Context, w http.ResponseWriter, r *http.Request) {
 	}
 
 	auditRec.Success()
+	auditRec.AddEventResultState(cm) // TODO what about the channel object? what's prior and resulting state... and type?
```
DSchalla, line 1738: I think the object_type you set makes sense. I don't think there is a prior state as this is a create operation, for resulting state that would be the channel membership (which we may not have here as a result and can be added later)

api4/channel.go
```diff
@@ -1630,9 +1649,8 @@ func addChannelMember(c *Context, w http.ResponseWriter, r *http.Request) {
 		return
 	}
 
-	auditRec := c.MakeAuditRecord("addChannelMember", audit.Fail)
-	defer c.LogAuditRec(auditRec)
-	auditRec.AddMeta("channel", channel)
+	// TODO if channel is prior state and resulting state is a channel_member then types are kind of mismatched
```
DSchalla, line 1652: Yes, but the `channel_id` could be added as a parameter.

api4/channel.go
```diff
@@ -1746,8 +1766,8 @@ func removeChannelMember(c *Context, w http.ResponseWriter, r *http.Request) {
 
 	auditRec := c.MakeAuditRecord("removeChannelMember", audit.Fail)
 	defer c.LogAuditRec(auditRec)
-	auditRec.AddMeta("channel", channel)
-	auditRec.AddMeta("remove_user_id", user.Id)
+	auditRec.AddEventParameter("channel_id", channel.Id)
+	auditRec.AddEventParameter("remove_user_id", user.Id)
```
DSchalla, line 1770:
```suggestion
	auditRec.AddEventParameter("user_id", user.Id)
```

api4/channel.go
```diff
@@ -1797,7 +1817,7 @@ func updateChannelScheme(c *Context, w http.ResponseWriter, r *http.Request) {
 
 	auditRec := c.MakeAuditRecord("updateChannelScheme", audit.Fail)
 	defer c.LogAuditRec(auditRec)
-	auditRec.AddMeta("new_scheme_id", *schemeID)
+	auditRec.AddEventParameter("new_scheme_id", *schemeID)
```
DSchalla, line 1820:
```suggestion
	auditRec.AddEventParameter("scheme_id", *schemeID)
```

model/config.go
```diff
@@ -3129,6 +3129,52 @@ type Config struct {
 	ExportSettings            ExportSettings
 }
 
+func (o *Config) Auditable() map[string]interface{} {
```
DSchalla, line 3132: I think we may want to return an empty interface for this one until we have it properly implemented. As you noted in the API handler it may make sense to just print the diff for the config in the Prior and Result state, but that would use a different mechanism. The risk here is if someone would use it they would also log all sensitive secrets.

DSchalla (changes requested):
Thanks for the work on this @oh6hay :tada: I did a first initial review and wanted to share some thoughts before walking through the remainder and raise the same flags. Let me know what you think.

- For the primary ID of the operation, e.g. a bot, let's use `id` for the ID and `bot` for the payload. As the event name has the entity in it (and the object_type too) ID refers to that entity, and the object_type being the name of the payload
- If it's an external ID, e.g. channel membership, let's just use `channel_id` as prefix. For other actions such as `create direct message` we can also just use `user_id` as the event name indicates what the ID is used for. The goal would be to have consistency across all event parameters to make writing alerting rules as easy as possible.
- We may need to look into a good solution for parameters added that are the object type we are auditing, e.g. how do we call auditable on it. I added a proposal to in my comment above.

model/bot.go
```diff
@@ -33,6 +33,20 @@ type Bot struct {
 	DeleteAt       int64  `json:"delete_at"`
 }
 
+func (b *Bot) Auditable() map[string]interface{} {
```
agnivade, line 36: I think it would be better if we auto-generate these methods. Keeping it up-to-date manually is fraught with danger, and it's very likely some dev will add a new field and forget to add that in this method.

It shouldn't be a big challenge technically. We just need to read the struct tags, and use them as the map keys and the struct fields as the values for the methods.

Not suggesting to do this right away, but something in the near future.

agnivade (approved):

DSchalla (approved):

PR #20534, opened 2022-06-23
MM-44863 - update trial details in product menu

pvev:
#### Summary
This PR make the getSubscription endpoint available with restricted and containing non-sensitive information to non-admin users so we can start implementing the functionalities where end users will be able to notify admins about "starting a trial" or "subscribing" depending on the subscription status.

#### Related Pull Requests
https://github.com/mattermost/mattermost-webapp/pull/10609

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-44863

#### Release Note
```release-note
NONE
```

### DIFF
--- a/api4/cloud_test.go
+++ b/api4/cloud_test.go
@@ -111,6 +111,90 @@ func Test_getCloudLimits(t *testing.T) {
 	})
 }
 
+func Test_GetSubscription(t *testing.T) {
+	subscription := &model.Subscription{
+		ID:          "MySubscriptionID",
+		CustomerID:  "MyCustomer",
+		ProductID:   "SomeProductId",
+		AddOns:      []string{},
+		StartAt:     1000000000,
+		EndAt:       2000000000,
+		CreateAt:    1000000000,
+		Seats:       10,
+		IsFreeTrial: "true",
+		DNS:         "some.dns.server",
+		IsPaidTier:  "false",
+		TrialEndAt:  2000000000,
+		LastInvoice: &model.Invoice{},
+	}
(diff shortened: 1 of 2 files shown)


marianunez (approved):

AGMETEOR:
@pvev and @marianunez  requesting we do something similar for the product end point `/api/v4/cloud/products` too. This PR https://github.com/mattermost/mattermost-webapp/pull/10437 will need to get products for non admin users

api4/cloud_test.go
```diff
@@ -111,6 +111,90 @@ func Test_getCloudLimits(t *testing.T) {
...
+		th := Setup(t).InitBasic()
+		defer th.TearDown()
+
+		th.Client.Login(th.BasicUser.Email, th.BasicUser.Password)
+
+		th.App.Srv().SetLicense(model.NewTestLicense("cloud"))
+
+		cloud := mocks.CloudInterface{}
+
+		cloud.Mock.On("GetSubscription", mock.Anything).Return(userFacingSubscription, nil)
```
AGMETEOR, line 157: This test doesn't achieve the intended purpose. The sanitizing happens within the handler and not from what the CWS returns

AGMETEOR (changes requested):

api4/cloud_test.go
pvev: you are right! nice catch @AGMETEOR

neallred (approved):

AGMETEOR (approved):
LGTM

PR #20557, opened 2022-06-28
[MM-45159] - Flaky test: TestCreateUserWithInitialPreferences

AGMETEOR:
#### Summary
Fix flaky TestCreateUserWithInitialPreferences test

#### Ticket Link
[MM-45159](https://mattermost.atlassian.net/browse/MM-45159)

#### Release Note

```release-note
NONE
```

### DIFF
--- a/app/user_test.go
+++ b/app/user_test.go
@@ -1725,7 +1725,6 @@ func TestUpdateThreadReadForUser(t *testing.T) {
 }
 
 func TestCreateUserWithInitialPreferences(t *testing.T) {
-	t.Skip("MM-45159")
 	th := Setup(t).InitBasic()
 	defer th.TearDown()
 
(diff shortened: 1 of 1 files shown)


agnivade:
Can you clarify what was the issue here?

AGMETEOR:
> Can you clarify what was the issue here?

I am not completely certain what was going on. I was also not able to reproduce it locally. However, my hunch was that it is possible for  the order of preferences in the slice to change and therefore indexing for a certain preference was not accurate.

agnivade (changes requested):
You are right @AGMETEOR - it is the order of elements in the slice. I can reproduce the exact failure by changing the order. This is expected because we are not ordering the results from the database, and the DB can return it in any order. Usually, it will always return in a certain order, so it will be hard to reproduce this locally.

@BenCookie95 - No I don't think it was flaky when it had 2 preferences because with just 2 items, this was hard to trigger. But now with 3 items, this is much more likely.

Since we are not returning the items in any order (and even doing that will not be deterministic because there is a userID in the preference name), so what we need to do is to check the presence of the 3 preferences in the slice instead of checking at specific slice indices.

Additionally, the order of arguments in the `assert.Equal` is incorrect. It should be (t, expected, actual). Right now it's (t, actual, expected).

@AGMETEOR - can you ping me when you have some time? I can walk you through this.

agnivade (approved):

